PR #102, opened 2014-12-31
Sort by downloads without letter restriction

### LINKED ISSUE #44: sorting by downloads
I don't know if it's only me or the sorting page for crates a bit weird. When we sort by downloads I think it would be nice that we remove the filter by alphabet above. Because it's rarely for someone to expect number of downloads by alphabets.

markberger:
* 'Browse All Crates' links to all crates instead of all
  crates that start with the letter A.
* Changing the 'Sort by' param returns the user to the first page.
* Crates pagination only shows the surrounding 9 pages instead
  of all possible pages.

Closes #44.

### DIFF
--- a/app/mixins/pagination.js
+++ b/app/mixins/pagination.js
@@ -1,14 +1,34 @@
 import Ember from 'ember';
 
+var VIEWABLE_PAGES = 9;
+
 export default Ember.Mixin.create({
+
+    // Gives page numbers to the surrounding 9 pages.
     pages: function() {
-        var availablePages = this.get('availablePages');
         var pages = [];
-        for (var i = 0; i < availablePages; i++) {
+        var currentPage = this.get('currentPage');
+        var availablePages = this.get('availablePages');
+        var lowerBound = 0;
+        var upperBound = 0;
+
+        // Always show the same number of pages even if we're
+        // at the beginning or at the end of the list.
+        if (availablePages - currentPage < Math.ceil(VIEWABLE_PAGES / 2)) {
+            lowerBound = Math.max(0, availablePages - VIEWABLE_PAGES);
+            upperBound = availablePages;
+        } else if (currentPage <= Math.ceil(VIEWABLE_PAGES / 2)) {
+            lowerBound = 0;
+            upperBound = Math.min(availablePages, VIEWABLE_PAGES);
+        } else {
+            lowerBound = currentPage - Math.ceil(VIEWABLE_PAGES / 2);
+            upperBound = currentPage + Math.floor(VIEWABLE_PAGES / 2);
+        }
+        for (var i = lowerBound; i < upperBound; i++) {
             pages.push(i + 1);
         }
         return pages;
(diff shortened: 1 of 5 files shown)


app/templates/crates.hbs
```diff
@@ -59,7 +60,7 @@
     {{#link-to 'crates' (query-params page=prevPage) class="prev"}}
         <img src="/assets/left-pag.png"/>
     {{/link-to}}
-    {{#each pages}}
+    {{#each surroundingPages}}
```
alexcrichton: I think that this should probably be done for all pagination, so you may just want to merge the `surroundingPages` method into `pages`. If we need to we can add an `allPages` method later on.

app/templates/crates.hbs
```diff
@@ -12,7 +14,6 @@
             {{ this }}
         {{/link-to}}
     {{/each}}
-    {{view Ember.Select content=alphabet value=letter}}
```
alexcrichton: This is used by the "mobile site" which you can see if you shrink the page by a good bit, so this needs to stay

app/mixins/pagination.js
```diff
@@ -10,6 +10,32 @@ export default Ember.Mixin.create({
...
+            upperBound = Math.min(availablePages, 9);
+        } else {
+            lowerBound = currentPage - 5;
+            upperBound = currentPage + 4;
+        }
+        for (var i = lowerBound; i < upperBound; i++) {
+            pages.push(i + 1);
+        }
+        return pages;
+    }.property('currentPage', 'availablePages'),
```
alexcrichton: Would it be possible to refactor this such that the number of pages is a constant listed in one location? In theory all the other numbers would be based off that one constant.

alexcrichton:
This looks great, thanks! Also, feel free to ping the PR whenever it's updated, sadly github doesn't send out notifications for force-pushes :(.

Would it be possible to have some `...` tokens in the nav to show that there are more pages to the left and the right? When looking at:



It's not super clear that there are pages to the right, or how many pages there are. It may be tough to get too fancy here, though, but curious what you think about that.

markberger:
Something like the Github pagination?

screen shot 2015-01-02 at 9 56 43 pm

I'm not opposed to having `...` to signify there are more pages. To be honest, I'm not much of a UI expert so I usually see what other sites are doing and copy them :stuck_out_tongue_winking_eye:. Google and Bing both use the system Crates.io has right now:

screen shot 2015-01-02 at 10 00 57 pm

screen shot 2015-01-02 at 10 01 29 pm

Also Crates.io does show the total number of packages in front of the user in the top left-hand corner:

screen shot 2015-01-02 at 10 03 12 pm

However, the `...` would definitely be clearer. Ultimately, I don't have a major preference for either one.

alexcrichton:
Eh, it looks like this has some good precedent, so I'm fine by that!

alexcrichton:
Thanks again!

Issue #99, opened 2014-12-31
Show reverse dependencies


winding-lines:
I see that the rust server already has an API for fetching reverse dependencies. Is there a plan/desire to add this information to the Ember app?

Also is it possible to get a dump of the postgres database to have some real data when playing with the application locally?

alexcrichton:
Now that we've got a more prominent location for links than we used to, I think I'd be fine just adding a link to the reverse dependencies page.

bjz:
:+1:

I constantly see 'got to the `/reverse_dependencies` page pointed out in IRC. Folks are always surprised and say, 'that's amazing!' or something. :)

alexcrichton:
cc @Turbo87 

an easy thing that could be added to the links sections for crates! Now that you've fixed the bugs on the page I think it'd be fine to just show it.

PR #104, opened 2015-01-02
Add a simple link to display reverse dependencies

winding-lines:
This is related to issue #99, it is the simplest approach I can think of. I can add more functionality if desired, for example display the link only if there are reverse dependencies.

Let me know what you think.

### DIFF
--- a/app/templates/crate/index.hbs
+++ b/app/templates/crate/index.hbs
@@ -143,9 +143,9 @@
       <li>None</li>
       {{/each}}
     </ul>
-    <div>
+    <div class=small>
       {{#link-to 'crate.reverse_dependencies' (query-params dependency=crate_id)}}
-      Reverse Dependencies
+      Reverse Dependencies {{reverse_dependencies.length}}
       {{/link-to}}
     </div>
   </div>
--- a/app/templates/crate/index.hbs
+++ b/app/templates/crate/index.hbs
@@ -143,6 +143,11 @@
       <li>None</li>
       {{/each}}
     </ul>
+    <div>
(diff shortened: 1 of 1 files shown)


alexcrichton:
This may want a bit of visual separation to set it a apart from the rest of the links:



cc @wycats, thoughts on the look-and-feel of this link?

winding-lines:
@alexcrichton here is the output using one of the existing css classes (small) and with the count of reverse dependencies.
screen shot 2015-01-03 at 9 32 39 am

A different option for the UI is to add inline the first 3 (x) reverse dependencies and then add an optional link to the rest.

alexcrichton:
Interesting! I'd probably put the numbers in parens, and I guess we probably don't want to show the link if no reverse dependencies exist.

My first thought was to not show the names of reverse dependencies, but perhaps there's precedent for it?

wycats:
I'll reply to this in more detail tomorrow. I definitely want something like this, but not QUITE this.

winding-lines:
@wycats, thx for your input :)

BTW, this is Marius, we met in Dec in Mountain View @ Linkedin.

huonw:
@wycats ping

amutake:
Just for your information, I wrote a simple chrome extension to see reverse dependencies.
https://github.com/amutake/crates.io-reverse-dependencies

wycats:
@amutake would be cool to get something like this in, but a fourth column is too narrow imo.

Maybe right below the three columns in full-width with the name "Dependents" instead of "Reverse-Dependencies"?

PR #105, opened 2015-01-02
Allows users to see download numbers on overlapping lines

markberger:
Allows users to see a box with the number of downloads for each crate version on a given day instead of examining each data point independently. This helps when multiple versions have 0 downloads on a given day and their chart lines overlap.

screen shot 2015-01-02 at 2 42 09 pm


alexcrichton:
Thanks!

Issue #107, opened 2015-01-03
Excessive api calls on crate page to versions end piont


markberger:
When I'm running `npm start -- --proxy https://crates.io` and I visit `http://0.0.0.0:4200/crates/openssl-sys`, there seems to be an excessive amount of calls to the versions end point.
```
GET /api/v1/crates/openssl-sys/owners?_=1420255799612 200 528.376 ms - 303
GET /api/v1/crates/openssl-sys/versions?_=1420255799611 200 662.884 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799614 200 231.526 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799615 200 245.911 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799613 200 262.085 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799616 200 261.535 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799617 200 243.079 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799618 200 248.667 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799619 200 77.900 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799621 200 140.690 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799623 200 105.282 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799622 200 127.032 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799624 200 101.641 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799625 200 81.966 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799620 200 274.322 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799629 200 143.284 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799630 200 114.798 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799626 200 217.548 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799631 200 101.379 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799627 200 232.704 ms - -
GET /api/v1/crates/openssl-sys/versions?_=1420255799632 200 102.858 ms - -
[...]

PR #109, opened 2015-01-05
Use computed property for versions length

BlakeWilliams:
A bug is causing `versions` to be fetched once for each version causing
excessive queries to slow down or freeze the page.

* Turn versions.length into a computed property to prevent extra
  requests

This only happens when you visit from the search. I found this by looking for "hyper" and clicking the first link and experiencing unresponsiveness. I'm not totally sure, but I think this might be a result of having both a `versions` attribute that is null from the search and having a `versions` attribute in `links`.

This could probably be avoided altogether by passing `version_ids` instead of the objects or links in addition to coalescing the requests. If you decide to go that way and need some help on the Ember side I'm more than happy to help implement that.

### DIFF
--- a/app/controllers/crate/index.js
+++ b/app/controllers/crate/index.js
@@ -36,6 +36,10 @@ export default Ember.ObjectController.extend({
              this.get('repository');
     }.property('homepage', 'wiki', 'mailing_list', 'documentation', 'repository'),
 
+    versionsCount: function() {
+      return this.get('versions.length');
+    }.property('versions.@each'),
+
     displayedAuthors: function() {
         var self = this;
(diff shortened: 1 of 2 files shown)


huonw:
It seems like this might fix https://github.com/rust-lang/crates.io/issues/107?

BlakeWilliams:
Yep! I should have checked the issues for it.

alexcrichton:
Awesome, thanks!

alexcrichton:
Hm, it looks like there's still 3 requests made for versions, but it's certainly better than N! I'll see if I can't find them.

PR #111, opened 2015-01-08
Change most used to most directly depended upon

### LINKED ISSUE #110: Change the front page "most downloaded" list to "most directly depended upon"
I think the latter list is likely to be full of projects which new users are more likely to want to play with, rather than just the "base" dependency packages of the ecosystem. 

I think it's important this happens before we get a massive influx of new users from 1.0, who will want to find highly usable libraries rather than just those which are important to the ecosystem (I'm thinking of things like gcc, openssl-sys, and newly libz-sys, which aren't needed directly by >90% of users but are currently on the front page) and I would be happy to work on making it happen if we agree this is desirable.

zzmp:
Fixes #110 

<s>This is very verbose (it calls it `Most Directly Depended Upon`).</s>
This calls the field `Most Depended Upon`, but is still verbose with the data field, calling the property `most_directly_depended_upon`.
Is this what we want?

@alexcrichton you mentioned in the issue to sort by `dependencies-on`.
Is this sufficient to count _only_ direct dependencies?

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -550,22 +550,22 @@ pub fn summary(req: &mut Request) -> CargoResult<Response> {
                                         WHERE updated_at::timestamp(0) !=
                                               created_at::timestamp(0)
                                         ORDER BY updated_at DESC LIMIT 10"));
-    let most_downloaded = try!(tx.prepare("SELECT * FROM crates \
-                                           ORDER BY downloads DESC LIMIT 10"));
+    let most_directly_depended_upon = try!(tx.prepare("SELECT * FROM crates \
+                                        ORDER BY dependencies-on DESC LIMIT 10"));
 
     #[derive(RustcEncodable)]
     struct R {
         num_downloads: i64,
         num_crates: i64,
         new_crates: Vec<EncodableCrate>,
-        most_downloaded: Vec<EncodableCrate>,
+        most_directly_depended_upon: Vec<EncodableCrate>,
         just_updated: Vec<EncodableCrate>,
     }
     Ok(req.json(&R {
         num_downloads: num_downloads,
         num_crates: num_crates,
         new_crates: try!(to_crates(new_crates)),
-        most_downloaded: try!(to_crates(most_downloaded)),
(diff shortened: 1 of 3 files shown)


reem:
I'd rather "Most Depended Upon"

zzmp:
@reem Updated the HTML to call it "Most Depended Upon" but leave the data as `most_directly_depended_upon` for verbosity.

src/krate.rs
```diff
@@ -550,22 +550,22 @@ pub fn summary(req: &mut Request) -> CargoResult<Response> {
                                         WHERE updated_at::timestamp(0) !=
                                               created_at::timestamp(0)
                                         ORDER BY updated_at DESC LIMIT 10"));
-    let most_downloaded = try!(tx.prepare("SELECT * FROM crates \
-                                           ORDER BY downloads DESC LIMIT 10"));
+    let most_directly_depended_upon = try!(tx.prepare("SELECT * FROM crates \
+                                        ORDER BY dependencies-on DESC LIMIT 10"));
```
huonw: I can't see this column defined in the "schema" (as in, it's not mentioned in the [migration code](https://github.com/rust-lang/crates.io/blob/743bfc980b76c914d545b41d6f4cb601a5c1d213/src/bin/migrate.rs)), am I missing something?

(If it turns out this isn't defined, one may wish to look at `reverse_dependencies` which does essentially exactly the same thing.)

alexcrichton:
Yes unfortunately as @huonw mentioned this will require some deeper changes to implement. We do not currently track the number of immediate dependencies in the row for each crate. Additionally, the reverse dependencies page is currently implemented by [joining many tables together](https://github.com/rust-lang/crates.io/blob/743bfc980b76c914d545b41d6f4cb601a5c1d213/src/krate.rs#L361-L371), but for a home-page, often used query I don't think this is gonna cut it.

To implement this a counter will need to be maintained on the `crates` table of all immediate dependencies, both when dependencies are added as well as removed. That way we can actually just use a `ORDER BY` clause to sort everything correctly.

brson:
Closing because very old.

PR #116, opened 2015-01-17
Add OpenSearch description

derkyjadex:
So we can search for crates easily from the address bar in Chrome/Firefox/etc.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -13,7 +13,7 @@
     <link rel="stylesheet" href="assets/cargo.css">
     <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon">
     <link rel="icon" href="/favicon.ico" type="image/x-icon">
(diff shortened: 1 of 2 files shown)


app/index.html
```diff
@@ -13,6 +13,7 @@
     <link rel="stylesheet" href="assets/cargo.css">
     <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon">
     <link rel="icon" href="/favicon.ico" type="image/x-icon">
+    <link rel="search" href="opensearch.xml" type="application/opensearchdescription+xml" title="Cargo">
```
alexcrichton: Should this be `/opensearch.xml` instead of `opensearch.xml` because this page is loaded from any path?

alexcrichton:
Nice! Thanks!

app/index.html
derkyjadex: I suppose it can't hurt to be more specific here

alexcrichton:
Thanks!

PR #132, opened 2015-02-08
Automatically install bower dependencies

pornel:
Makes `npm install` "just work" and doesn't require bower to be installed globally.

npm is clever enough to prepend `PATH` with `./node_modules/.bin` for the postinstall script, which ensures that the right version of bower is used.


alexcrichton:
Nice!

PR #140, opened 2015-03-20
Normalise keywords to lowercase when searching for crates.

### LINKED ISSUE #100: Keywords could be case-normalised
e.g. we have https://crates.io/keywords/FFT and https://crates.io/keywords/fft separate atm.

huonw:
Currently keyword search is case-sensitive, but this is often unhelpful,
as they are heuristic classification tools, not hard guarantees.  It's
not particularly helpful to search for
["fft"](https://crates.io/keywords/fft) and not get the crates tagged
["FFT"](https://crates.io/keywords/FFT).

Closes #100.

### DIFF
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -440,6 +440,14 @@ fn migrations() -> Vec<Migration> {
                              ON crates (lower(name))", &[]));
             Ok(())
         }),
+        Migration::new(20150320174400, |tx| {
+            try!(tx.execute("CREATE INDEX index_keywords_lower_keyword ON keywords (lower(keyword))",
+                            &[]));
+            Ok(())
+        }, |tx| {
+            try!(tx.execute("DROP INDEX index_keywords_lower_keyword", &[]));
+            Ok(())
+        }),
     ];
     // NOTE: Generate a new id via `date +"%Y%m%d%H%M%S"`
 
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -463,7 +463,7 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
                                 ON crates.id = crates_keywords.crate_id
                         INNER JOIN keywords
                                 ON crates_keywords.keyword_id = keywords.id
-                        WHERE keywords.keyword = $1";
+                        WHERE lower(keywords.keyword) = lower($1)";
             (format!("SELECT crates.* {} {} LIMIT $2 OFFSET $3", base, sort_sql),
              format!("SELECT COUNT(crates.*) {}", base))
         })
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -68,7 +68,9 @@ fn index_queries() {
     krate.readme = Some("readme".to_string());
     krate.description = Some("description".to_string());
     ::mock_crate(&mut req, krate);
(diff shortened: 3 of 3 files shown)


huonw:
I don't have a testing set-up set up, unfortunately.

src/bin/migrate.rs
```diff
@@ -418,6 +418,11 @@ fn migrations() -> Vec<Migration> {
                              crate_owners_unique_user_per_crate", &[]));
             Ok(())
         }),
+        Migration::new(20150320174400, |tx| {
+            try!(tx.execute("ALTER TABLE keyword SET keyword = lower(keyword)",
+                            &[]));
```
alexcrichton: I think that this should be `ALTER TABLE keywords` (although I also thought it would be `UPDATE keywords ...`?).

I think that this may also not land well in production as there's currently a uniqueness constraint on the `keyword` column, but this will generate duplicates (e.g. FFT => fft).

I think that this may want to iterate through all keywords, deleting those with a downcase equivalent and migrating all current members of that keyword to the downcase equivalent. This should also probably delete the old uniqueness constraint and add a new one on `lower(keyword)` to speed up queries and provide a db-level gurantee that we don't have duplicates.

alexcrichton:
Nice! Could you also add a test where two crates are uploaded with two different cases of a keyword, but they're canonicalized to just one?

For crates at least we also search case insensitively but *insert* with the case provided. In that case we may want to preserve the case of the keyword given to us when we insert into the database.

src/bin/migrate.rs
huonw: Oh, I totally mangled this.

huonw:
> Could you also add a test where two crates are uploaded with two different cases of a keyword, but they're canonicalized to just one?

I'm not sure I understand what this would look like.

huonw:
> For crates at least we also search case insensitively but insert with the case provided. In that case we may want to preserve the case of the keyword given to us when we insert into the database.

Updated with this approach.

alexcrichton:
> > Could you also add a test where two crates are uploaded with two different cases of a keyword, but they're canonicalized to just one?
> 
> I'm not sure I understand what this would look like.

Oh nevermind, I thought that this was going to prevent creation of keywords with the same name that differed only in case, but I think I prefer this approach instead.

Thanks!

src/bin/migrate.rs
```diff
@@ -440,6 +440,14 @@ fn migrations() -> Vec<Migration> {
                              ON crates (lower(name))", &[]));
             Ok(())
         }),
+        Migration::new(20150320174400, |tx| {
+            try!(tx.execute("CREATE INDEX index_keywords_lower_keyword ON (lower(keyword))",
```
alexcrichton: I think this is missing the table name

alexcrichton: This file also seems to have made itself executable?

huonw: I'm not doing a very good job of writing correct code... :/

Fixed both.

alexcrichton:
Looks good to me! I'll merge once it agrees with the nightlies again.

alexcrichton:
Merged as https://github.com/rust-lang/crates.io/commit/4658cf38cde8ddc5a35156e0150b766d26594766, botched the merge but oh well :(

PR #142, opened 2015-03-31
Wrap the search field in a <form> element

mbrubeck:
This should have no effect on the behavior or appearance of the search box, except that it enables features like right-click -> "Add a keyword for this search" in Firefox and other browsers.


alexcrichton:
Neta!

alexcrichton:
er, neat*!

PR #143, opened 2015-04-02
add a "badge" route for crates

seanmonstar:
The route /crates/foo.svg will redirect to shields.io, of an svg showing
the latest version of the crate on crates.io.

Example: [](https://crates.io/crates/hyper)

This could be shown in the UI somehow, but for now, exposing a route is quite cheap, and shields.io encourages use like this. A use case is including the badge in the README on github, so visitors can see what is the latest version published, even if master is a little ahead.

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -1019,3 +1019,12 @@ pub fn reverse_dependencies(req: &mut Request) -> CargoResult<Response> {
     struct Meta { total: i64 }
     Ok(req.json(&R{ dependencies: rev_deps, meta: Meta { total: total } }))
 }
+
+pub fn shield(req: &mut Request) -> CargoResult<Response> {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Neat!

seanmonstar:
Though, I couldn't test locally, since the rust version was older. It might be that `/crates/:crate_id.svg` may not parse to what I expect.

alexcrichton:
Ah unfortunately route-recognizer doesn't support `:foo.bar`, so I'm going to have to revert this.

seanmonstar:
If that's the case, it could be `/crates/:crate_id/badge.svg` instead?

alexcrichton:
Perhaps, but `/crates/foo.svg` is probably the right place for this to actually be, so it may be best to fix the underlying bug.

PR #147, opened 2015-04-21
Fix user links CSS

gentoo90:
Before | After
------------ | -------------
before | after

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -113,6 +113,9 @@ body {
 
     #current-user-links {
         display: none;
+        left: auto;
+        right: 0;
+        min-width: 200px;
 
         &.open { display: block; }
     }


tomaka:
Ah! This problem has been bothering me for weeks! :heart:

alexcrichton:
Hm when I shrink the size of the windows to be very skinny this causes the menu to look a little funny. The current rendering also looks good to me on Firefox at least. What browser are you using?

gentoo90:
Firefox. Fixed:
after2

alexcrichton:
Hm the text still looks a little scrunched to me on Firefox with a smaller screen:

Before:



After:



Is the `min-width` property still necessary? (removing that made it look ok). To clarify, what browser are you using to view the changes here?

gentoo90:
OK, now I found it, this small-screen menu already has min-width in its style, and it's `200px`.
> what browser are you using to view the changes here?

Firefox and Chrome.

gentoo90:
> Is the min-width property still necessary?

Yes, without it user menu still look like this
before

alexcrichton:
Ah ok, with the higher min-width this looks fine to me, thanks!

PR #157, opened 2015-05-21
Give the search results page a title

mbrubeck:
I'm an Ember novice, so please let me know if there's a better way to do this!


alexcrichton:
Thanks!

Issue #141, opened 2015-03-20
Redirecting 404 is unfriendly


andersk:
https://crates.io/crates/nonexistent redirects to https://crates.io, which makes it hard to correct a manually typed URL and try again. It also breaks the back button.

eternaleye:
Sourceforge has spent the last week demonstrating just how awful this is - it redirecting downloads while the site is broken rather than properly 404'ing has caused real issues for Exherbo's distfile mirroring setup.

steveklabnik:
:+1:

comex:
Bump.

steveklabnik:
This no longer redirects!

durka:
This has come back!

PR #167, opened 2015-07-13
Use `replaceWith` to transition from missing crate

BlakeWilliams:
When visiting a crate that doesn't exist the application uses
`transitionTo` to change the route to the index route. This allows the
user to navigate back to the same missing crate route which breaks the
browsers back button.

* Replace `transitionTo` with `replaceWith` to avoid adding history
  entry.

https://github.com/rust-lang/crates.io/issues/141

### DIFF
--- a/app/routes/crate.js
+++ b/app/routes/crate.js
@@ -10,17 +10,16 @@ export default Ember.Route.extend({
             version = parts[1];
         }
         var self = this;
-        var a = Ember.RSVP.hash({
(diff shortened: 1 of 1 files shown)


BlakeWilliams:
It might also be worth investigating an actual "missing crate" so typos can easily be fixed instead of having to be written from scratch like the issue says.

alexcrichton:
Thanks!

Issue #137, opened 2015-02-24
Groups for ownership *or* GitHub-based permissions


SimonSapin:
We (the Servo team) want to publish more of our libraries on crates.io. One thing this implies is dealing with crates ownership: who can publish new versions. Currently, AFAIU, we have to add or remove permissions for individual people for individual crates. Compare with GitHub where we have "teams" where people participate, and teams has a whole have push access to repositories. This is much easier to manage when the number of people and/or repositories/crates grows.

It would help us if cartes.io had "groups" or "teams" or something like it, to manage crates ownership.

Alternatively, I was going to suggest deferring permission management to GitHub: if a crate name is associated with a repository, anyone with permission to push to that repo can also publish a new version. Unfortunately, [that part of the GitHub API](https://developer.github.com/v3/repos/collaborators/#get) seems not to be public:

```
$ curl -si https://api.github.com/repos/rust-lang/crates.io/collaborators/alexcrichton | grep Status
Status: 401 Unauthorized
```

Manishearth:
Crates is anyway associated with a GH account, perhaps it can ask for those additional permissions?

alexcrichton:
Yes I'd definitely love to integrate the teams/organizations aspect of Github a bit more into crates.io for smoother management of these aspects.

Gankro:
+1 

I'm splitting a monolithic crate into a bunch of little ones and this is a headache.

Gankro:
discussed this more on IRC. Current plans is to tie this 100% to github orgs/teams as a starter, since that's what most of us just *want* anyway.

Proposed syntax:
`cargo owner --add github:ORG:TEAM`

Backend Implementation:
* Add a new column to the owners table specifying `owner_kind`, with two possible choices: `user`, and `github team`. 
* Change owner auth from "is in owners list" to "is in owners list, or is in a team in owners list". Critically this "is in team" lookup should be deferred totally to Github, so that we don't need to synchronize who's in what team (or make shadow users or anything).

I'm willing to implement but am honestly a bit swamped with school stuff and am not familiar with any of the related codebases or APIs. So basically if I manage to implement it won't be for a while.

SaberUK:
Tight integration with a proprietary code hosting service which may not be around in 10 years seems a bit concerning to me.

Manishearth:
[...]

PR #177, opened 2015-08-06
WIP: Teams

Gankro:
**This is completely untested**

I just got it to compile, and wanted to get this up for the bikesheddy part of review while I start setting up testing and stuff.

Some of the JSON schemas still probably need to be updated to yield duplicate fields for bridging legacy and clarity. I don't *think* Cargo needs to be modified at all (teams are Just Another Name as far as it should be concerned), but I might be forgetting something.


Gankro:
Oh, closes #137

Gankro:
All old tests pass locally now. Moving on to making new tests.

src/krate.rs
```diff
@@ -174,8 +175,9 @@ impl Crate {
         })));
 
         try!(conn.execute("INSERT INTO crate_owners
-                           (crate_id, user_id, created_at, updated_at, deleted)
-                           VALUES ($1, $2, $3, $3, FALSE)",
+                           (crate_id, owner_id, created_by, created_at,
```
Gankro: Note: I added the created_by field here while debugging a test failure cause by something else. Not sure why it wasn't being set and whether it should be set.

src/krate.rs
```diff
@@ -953,10 +972,10 @@ pub fn owners(req: &mut Request) -> CargoResult<Response> {
     let tx = try!(req.tx());
     let krate = try!(Crate::find_by_name(tx, crate_name));
     let owners = try!(krate.owners(tx));
-    let owners = owners.into_iter().map(|u| u.encodable()).collect();
+    let owners = owners.into_iter().map(|o| o.encodable()).collect();
 
     #[derive(RustcEncodable)]
-    struct R { users: Vec<EncodableUser> }
+    struct R { users: Vec<EncodableOwner> }
```
Gankro: Note: this is the one place that I'm not sure how to handle API-wise -- this is something that cargo sends *us*. I suppose we could have two duplicate fields, both of which are Options? Try to decode both, pick whichever has stuff first?

Gankro:
Status Update: I have tested this out locally by hand and everything seems to be working!!!

I just need to write some unit tests. (oh nooo mocking)

src/krate.rs
```diff
@@ -174,8 +175,9 @@ impl Crate {
         })));
 
         try!(conn.execute("INSERT INTO crate_owners
-                           (crate_id, user_id, created_at, updated_at, deleted)
-                           VALUES ($1, $2, $3, $3, FALSE)",
+                           (crate_id, owner_id, created_by, created_at,
+                             updated_at, deleted, owner_kind)
+                           VALUES ($1, $2, $2, $3, $3, FALSE, 0)",
```
alexcrichton: Could you refactor this literal 0 into something like `OwnerKind::User as i32`?

src/krate.rs
```diff
@@ -288,38 +290,53 @@ impl Crate {
         Ok(ret)
     }
 
-    pub fn owners(&self, conn: &Connection) -> CargoResult<Vec<User>> {
+    pub fn owners(&self, conn: &Connection) -> CargoResult<Vec<Owner>> {
+        // Users and Teams are differentiated by the owner_kind field:
+        // User = 0, Team = 1
```
alexcrichton: Could this be indicated through an `enum` instead?

src/krate.rs
alexcrichton: Yeah for now let's have two duplicate fields, both options. We can slowly phase out the older field later on down the road (could add a comment explaining what's up)

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+use pg::rows::Row;
+use rustc_serialize::json;
+use util::errors::NotFound;
+use std::str;
+
+
+/// Unifies the notion of a User or a Team.
+pub enum Owner {
+    User(User),
+    Team(Team)
```
alexcrichton: trailing comma

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+/// For now, just a Github Team. Can be upgraded to other teams
+/// later if desirable.
+pub struct Team {
+    /// Github annoyingly has some APIs talk about teams by name,
+    /// and others by some opaque id. I couldn't find any docs
+    /// suggesting these ids are stable, so let's just always
+    /// ask github for it. *shrug*
+    github_id: i32,
+    /// Unique table id
+    cargo_id: i32,
```
alexcrichton: Currently the model-like structs all have public fields, and the crates.io ID is just called `id`

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// We only query membership with github using the github_id, though.
+    name: String,
+}
+
+#[derive(RustcDecodable, RustcEncodable)]
+pub struct EncodableOwner {
+    pub id: i32,
+    // Login is deprecated in favour of name, but needs to be printed for back-compat
+    pub login: String,
+    pub kind: String,
```
alexcrichton: This'll need to be `Option<String>` to work with older Cargos

src/bin/migrate.rs
```diff
@@ -448,6 +448,21 @@ fn migrations() -> Vec<Migration> {
...
         }),
+        Migration::add_column(20150715170350, "crate_owners", "owner_kind",
+                              "INTEGER NOT NULL DEFAULT 0"),
+        Migration::run(20150804170127,
+            "ALTER TABLE crate_owners ALTER owner_kind DROP DEFAULT",
+            "ALTER TABLE crate_owners ALTER owner_kind SET DEFAULT 0",
+        ),
+        Migration::add_table(20150804170128, "teams", "
+            id            SERIAL PRIMARY KEY,
+            name          VARCHAR NOT NULL UNIQUE,
```
alexcrichton: Can you make sure there's an index on this field? You're doing a query on it sometimes so it'll need to be indexed, although I forget if the `UNIQUE` constraint adds an index or not...

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /// Tries to create the Team in the DB (assumes a `:` has already been found).
+    pub fn create(conn: &Connection, name: &str, req_user: &User) -> CargoResult<Self> {
+        // must look like system:xxxxxxx
+        let mut chunks = name.split(":");
+        match chunks.next().unwrap() {
+            // github:rust-lang:owners
+            "github" => {
+                // Ok to unwrap since we know one ":" is contained
```
alexcrichton: Odd comment? (no unwraps here)

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+        // GET orgs/:org/teams
+        // check that `team` is the `slug` in results, and grab its `id`
+
+        // "sanitization"
+        fn whitelist(c: &char) -> bool {
+            match *c {
+                'a'...'z' | 'A'...'Z' | '0'...'9' | '-' | '_' => false,
+                _ => true
+            }
+        }
```
alexcrichton: Could this reuse `Crate::valid_name`?

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+            return Err(human(format!("organization cannot contain special characters like {}",
+                                   c)));
+        }
+
+        let resp = try!(http::handle()
+                         .get(format!("https://api.github.com/orgs/{}/teams", org_name))
+                         .header("Accept", "application/vnd.github.v3+json")
+                         .header("User-Agent", "hello!")
+                         .header("Authorization", &format!("token {}", &req_user.gh_access_token))
+                         .exec());
```
alexcrichton: Could you refactor this between `src/user/mod.rs` to avoid duplication?

src/owner.rs
Gankro: This is actually going to revert back -- the try! on org is otherwise dead code.

src/owner.rs
Gankro: That doesn't seem very semantically clear. We're trying to filter out to only the valid set of GH team slugs. (actually I think A...Z could be omitted)

src/owner.rs
alexcrichton: Unfortunately this may also not be quite right as it looks like the teams listing for an organization has pagination, so we may have to make multiple requests to look up the teams in an organization.

Gankro: What's the duplication?

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+            return Err(human("only members of a team can add it as an owner"));
+        }
+
+        // insert into DB for reals
+        try!(conn.execute("INSERT INTO teams
+                           (name, github_id)
+                           VALUES ($1, $2)",
+                          &[&name, &github_id]));
+
+        // read it right back out:
```
alexcrichton: If you search for "RETURNING" you can see a method of inserting while also returning the row that was just inserted

src/owner.rs
Gankro: I think this is only ever encoded today. Cargo only ever sends us a list of names.

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+                // Yes, this is actually how "no membership" is signaled
+                return Ok(false);
+            }
+            403 => {
+                return Err(human("It looks like you don't have permission \
+                                  to query an organization that owns this \
+                                  crate. You may need to re-authenticate on \
+                                  crates.io to grant permission to read \
+                                  github org memberships. Just go to \
+                                  https://crates.io/login"));
```
alexcrichton: It'd be good to refactor this 403-checking logic with the above block to reduce duplication

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+                break;
+            }
+        }
+
+        let github_id = try!(github_id.ok_or_else(|| {
+            human(format!("could not find the github team {}/{}", org_name, team_name))
+        }));
+
+        // mock Team (only need ID to check team status)
+        let team = Team { github_id: github_id, cargo_id: 0, name: String::new() };
```
alexcrichton: Instead of making a dummy team could this just be a static method which takes only a github id?

src/owner.rs
Gankro: They have different messages, fwiw.

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// it will verify the req_user is on the team first.
+    ///
+    /// If the req_user is on the Team, it will create the team in the DB
+    /// if it is not already present. When this occurs, this will set the
+    /// One True Name of the team. All future references to the team must
+    /// use the exact casing provided here. If a different casing is provided
+    /// to this method, we may still succeed if Github returns us the same ID
+    /// as the One True Name. However in this case, the One True Name will
+    /// still be selected.
+    pub fn find_by_name_for_add(conn: &Connection, name: &str, req_user: &User)
```
alexcrichton: This is only called in one location, so perhaps the logic for checking whether the user is a member of a team can be inlined at that location? Either that or otherwise add a method to `Owner` querying whether a user is a member of that owner.

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+                }
+            }
+            Owner::Team(Team { cargo_id, name, .. }) => {
+                EncodableOwner {
+                    id: cargo_id,
+                    login: name,
+                    email: None,
+                    avatar: None,
+                    name: None,
+                    kind: String::from("owner"),
```
alexcrichton: s/owner/team/ ?

src/owner.rs
```diff
@@ -0,0 +1,357 @@
...
+                    kind: String::from("user"),
+                }
+            }
+            Owner::Team(Team { cargo_id, name, .. }) => {
+                EncodableOwner {
+                    id: cargo_id,
+                    login: name,
+                    email: None,
+                    avatar: None,
+                    name: None,
```
alexcrichton: Some(name) ?

src/owner.rs
alexcrichton: It is? It looks like cargo currently [requires a `login` field](https://github.com/rust-lang/cargo/blob/master/src/crates-io/lib.rs#L76-L83)

src/owner.rs
alexcrichton: Could we just leave this up to github?

src/owner.rs
alexcrichton: Setting up the `http::handle`, and various other headers are duplicated, and it'd be useful to just do something like passing in the URL and then getting the response.

src/owner.rs
alexcrichton: True, but it's probably not that important in this case

src/owner.rs
Gankro: Definitely want to sanitize to avoid `/`'s and what not -- easiest way I could think of was to just whitelist the stuff that's in slugs.

src/owner.rs
Gankro: Yes, cargo will decode this -- but we won't. So we say we have to provide the login, and downstream Cargo can say it's optional if it wants. But we have to for old Cargos.

src/bin/migrate.rs
Gankro: Per http://dba.stackexchange.com/questions/144/when-should-i-use-a-unique-constraint-instead-of-a-unique-index it appears that a unique constraint implies the creation of an index. The top answer argues you should favour an index if you actually intend to index off that value as a matter of clarity (the index won't go away if the uniqueness constraint goes away), but that doesn't seem like a concern here.

src/owner.rs
Gankro: So one thing is that Teams and Users auth in different ways. Users are using the oauth2 lib to inject the header, but Teams can't do that (as far as I can tell) because there isn't a way to create an oauth2 object from the string key we store. As such teams just write the appropriate header manually (which is trivial and will never change).

src/owner.rs
Gankro: :+1:

src/owner.rs
Gankro: I think it's kinda important from a UX perspective -- people are going to try this feature and get walled on out-of-date authentication. There's two places where this can happen, and they're sufficiently different that they warrant different messages IMO.

src/owner.rs
Gankro: What's the purpose of the ownership query? (Old drafts had such a method, but it ended up being useless).

src/owner.rs
Gankro: Wait I'm confused, are we talking about `login` or `kind`?

src/owner.rs
Gankro: ok I have this back in cache now. `name` is the pretty-printed-name. e.g. "Alexis Beingessner". So it doesn't super make sense to produce that here? (it will make cargo produce a silly output when you list owners).

We do want a duplicate of `login`, it's unfortunate that `name` is taken!

alexcrichton: Ah ok, this is fine in that case.

src/owner.rs
alexcrichton: Just in the sense that the caller of this method would do something like:

```rust
let owner = try!(Owner::find_by_name(name));
if !owner.contains(&user) {
    return Err(...)
}
```

src/owner.rs
alexcrichton: Hm ok, can there perhaps be a switch of some form? It'd just be nice to consolidate all this HTTP logic as much as possible.

src/owner.rs
alexcrichton: Hm ok, I'll probably do one final auditing pass but this is fine for now

src/bin/migrate.rs
alexcrichton: Ok, seems fine to me!

src/owner.rs
alexcrichton: Oh sorry I meant the `kind` field. Cargo doesn't currently give us a `kind` field, right? That means this'll have to be an `Option` if we want to continue to receive responses from Cargo?

Gankro: Yeah so due to the `github:` prefix owners are uniquely identified by their string name -- which is the only way Cargo talks to us about them. They just send a list of names (which isn't parsed with this struct at all).

src/owner.rs
Gankro: Is this not the moral equivalent of User::find_or_insert (but generalized)?

src/http.rs
```diff
@@ -0,0 +1,22 @@
+use curl;
+use oauth2::*;
+use app::App;
+
+pub fn github(app: &App, url: &str, auth: &Token)
+    -> Result<curl::http::Response, curl::ErrCode> {
+    println!("HTTP: {}", url);
```
alexcrichton: s/println/info/

src/http.rs
```diff
@@ -0,0 +1,22 @@
...
+    app.handle()
+     .get(url)
+     .header("Accept", "application/vnd.github.v3+json")
+     .header("User-Agent", "hello!")
+     .auth_with(auth)
+     .exec()
+}
+
+pub fn token(token: String) -> Token {
```
alexcrichton: Could you add a comment here indicating that this is intended to pass to the function above? (e.g. the scopes/token_type are blank)

src/krate.rs
alexcrichton: You can probably delete this comment now

src/krate.rs
```diff
@@ -288,38 +290,65 @@ impl Crate {
...
                         name: &str) -> CargoResult<()> {
-        let user = try!(User::find_by_login(conn, name).map_err(|_| {
+        let owner = try!(Owner::find_by_name(conn, name).map_err(|_| {
             human(format!("could not find user with login `{}`", name))
         }));
         try!(conn.execute("UPDATE crate_owners
                               SET deleted = TRUE, updated_at = $1
-                            WHERE crate_id = $2 AND user_id = $3",
-                          &[&::now(), &self.id, &user.id]));
+                            WHERE crate_id = $2 AND owner_id = $3 AND owner_kind = $4",
```
alexcrichton: You shouldn't need the `owner_kind` field here, `owner_id` should be unique and should directly imply what `owner_kind` should be

src/krate.rs
```diff
@@ -476,13 +505,15 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
             (format!("SELECT crates.* FROM crates
                        INNER JOIN crate_owners
                           ON crate_owners.crate_id = crates.id
-                       WHERE crate_owners.user_id = $1 {} \
+                       WHERE crate_owners.owner_id = $1
+                       AND crate_owners.owner_kind = 0 {}
```
alexcrichton: this 0 should be replaced with a symbolic constant

alexcrichton: (same below)

src/owner.rs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        if let Some(c) = org_name.chars().find(whitelist) {
+            return Err(human(format!("organization cannot contain special characters like {}",
+                                   c)));
+        }
+
+        let resp = try!(http::github(app,
+            &format!("http://api.github.com/orgs/{}/teams", org_name),
+            &http::token(req_user.gh_access_token.clone())));
```
alexcrichton: Could you stylistically break this up into:

```rust
let url = ...;
let token = ...;
let resp = try!(http::github(app, &url, &token));
```

src/owner.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+fn team_with_gh_id_contains_user(app: &App, github_id: i32, user: &User) -> CargoResult<bool> {
+    // GET teams/:team_id/memberships/:user_name
+    // check that "state": "active"
+
+    let resp = try!(http::github(app,
+        &format!("http://api.github.com/teams/{}/memberships/{}", &github_id, &user.gh_login),
+        &http::token(user.gh_access_token.clone())));
```
alexcrichton: Same comment about style here as above

src/owner.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// Phones home to Github to ask if this User is a member of the given team.
+    /// Note that we're assuming that the given user is the one interested in
+    /// the answer. If this is not the case, then we could accidentally leak
+    /// private membership information here.
+    pub fn contains_user(&self, app: &App, user: &User) -> CargoResult<bool> {
+        team_with_gh_id_contains_user(app, self.github_id, user)
+    }
+}
+
+fn team_with_gh_id_contains_user(app: &App, github_id: i32, user: &User) -> CargoResult<bool> {
```
alexcrichton: Can you be sure this wraps to 80 chars?

src/tests/team.rs
```diff
@@ -0,0 +1,234 @@
...
+    body.extend([
+        (json.len() >>  0) as u8,
+        (json.len() >>  8) as u8,
+        (json.len() >> 16) as u8,
+        (json.len() >> 24) as u8,
+    ].iter().cloned());
+    body.extend(json.as_bytes().iter().cloned());
+    body.extend([0, 0, 0, 0].iter().cloned());
+    body
+}
```
alexcrichton: These look duplicated with `krate.rs`, could they be consolidated?

src/krate.rs
Gankro: (discussed on irc -- this is false)

src/tests/team.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    ::mock_user(&mut req, mock_kitty());
+    let body = r#"{"users":["FlashCat"]}"#;
+    let json = bad_resp!(middle.call(req.with_path("/api/v1/crates/foo/owners")
+                                       .with_method(Method::Put)
+                                       .with_body(body.as_bytes())));
+    assert!(json.errors[0].detail.contains("don't have permission"),
+            "{:?}", json.errors);
+}
+
```
alexcrichton: Can you also add tests for:

* Owners can remove a team from a crate
* Team members cannot remove a team from a crate

src/version.rs
```diff
@@ -343,7 +344,7 @@ fn modify_yank(req: &mut Request, yanked: bool) -> CargoResult<Response> {
     let user = try!(req.user());
     let tx = try!(req.tx());
     let owners = try!(krate.owners(tx));
-    if !owners.iter().any(|u| u.id == user.id) {
+    if try!(rights(req.app(), &owners, &user)) < Rights::Publish {
```
alexcrichton: This means that team members can't yank? I'd be fine allowing this

src/owner.rs
```diff
@@ -0,0 +1,325 @@
...
+                    name: name,
+                    kind: String::from("user"),
+                }
+            }
+            Owner::Team(Team { id, name, .. }) => {
+                EncodableOwner {
+                    id: id,
+                    login: name,
+                    email: None,
+                    avatar: None,
```
alexcrichton: Can you make sure the frontend handles this nicely? I'm curious what the page looks like when it lists owners and one of the owners is a team.

src/http.rs
```diff
@@ -0,0 +1,22 @@
+use curl;
+use oauth2::*;
+use app::App;
+
+pub fn github(app: &App, url: &str, auth: &Token)
```
alexcrichton: This function might be a lot nicer to call if it was something along the lines of:

```rust
pub fn github<T: Decodable>(app: &App, url: &str, auth: &Token)
        -> Result<T, curl::ErrCode> { 
    // ...
}
```

Basically something that does the decoding for you automatically and fills in an error for bad JSON and bad utf-8-ness

alexcrichton:
OK, I just hooked up coveralls to this repo so in theory when you rebase you'll start seeing coverage information published, and it's a good check to see if at least most of the branches and such were taken during the tests perhaps?

src/owner.rs
Gankro: You get an output that looks like:

```
steveklabnik (Steve Klabnik <steve@steveklabnik.com>)
github:rust-lang:owners
```

Gankro: Ohhh you mean the actual crates.io website

src/version.rs
Gankro: Team members have Rights::Publish, so they should pass this check.

src/owner.rs
Gankro: The frontend really drops the ball on this, I'm not gonna lie. Broken image and a link to `github.com/github:rust-lang:owners`

src/http.rs
Gankro: The response-code handling isn't uniform, though. In particular when querying team membership, 404 is how Github reports "false". Do you want this to take a closure for (maybe) processing the error code?

Gankro: I made two fns, one for querying, one for parsing, so that you can intermediate evaluation of the request as desired.

alexcrichton:
Awesome, thanks so much @Gankro! I've added some extra things here and there and pushed to master.

PR #181, opened 2015-08-10
Make graph version numbers not wrap

mdinger:
Crates such as [regex](https://crates.io/crates/regex) have their version numbers wrapping in the graph which shouldn't be happening. Make the graph smaller so it stops happening.

Note: I couldn't test this because no test crates have version numbers high enough to trigger wrap-around. I'm not sure if I'm able to upload to it or if an admin has to do it. Either way, I'd upload a crate with a version number like `42.12.88` (admittedly, a crate with 6 digits might be unlikely...not sure) so the graph can be sized accordingly.

Then I'd change the size so it doesn't wrap.

### DIFF
--- a/app/controllers/crate/index.js
+++ b/app/controllers/crate/index.js
@@ -206,7 +206,7 @@ export default Ember.ObjectController.extend({
                 }
                 var chart = new google.visualization.AreaChart(el);
                 chart.draw(myData, {
-                    chartArea: {'width': '80%', 'height': '80%'},
+                    chartArea: {'width': '78%', 'height': '80%'},
                     hAxis: {
                         minorGridlines: { count: 8 },
                     },


alexcrichton:
Thanks!

mdinger:
@alexcrichton Bah, it didn't work. How do you upload a crate to the test crates? I need to create one with 6 digits in the version. Then I can actually set the size properly.

That last one was just a guess to try to make it small enough.

alexcrichton:
Ah ok, sorry I think I hit the merge button a little too quickly! I've reverted this in https://github.com/rust-lang/crates.io/pull/182.

You should be able to run an instance locally by following the instructions in the README and then you can publish a longer crate name. You can also probably just search on crates.io for those with a long version name as well though and just hook up to that (easier than running the whole server locally)

mdinger:
@alexcrichton The proxy works. I'll try to get it working.

PR #183, opened 2015-08-10
Second try making the graph legend fit without wrap

mdinger:
Should be good for at least a 100,000 downloads and 5 digit versions without resorting to wrapping. It's shifted because staying centered tends to keep it too far to the right (the diagram stays centered while the legend gets closer to the edge).

Regex crate:
regex

### DIFF
--- a/app/controllers/crate/index.js
+++ b/app/controllers/crate/index.js
@@ -206,7 +206,7 @@ export default Ember.ObjectController.extend({
                 }
                 var chart = new google.visualization.AreaChart(el);
                 chart.draw(myData, {
-                    chartArea: {'width': '80%', 'height': '80%'},
+                    chartArea: {'left': 85, 'width': '77%', 'height': '80%'},
                     hAxis: {
                         minorGridlines: { count: 8 },
                     },


alexcrichton:
What's the wrapping you see today? The screenshot looks like what I see today for regex and it looks ok.

mdinger:
Regex at crates.io taken just now:
regex_on_crates_io

I tried proxy directly to crates.io instead of heroku when testing my patch but I don't think it could have modified crates.io but I don't really know. What's the lag time in crates.io being updated by the repository? My previous fix might have fixed it and it just hadn't appeared when I looked last time. If the revert hasn't landed yet, it could be that.

mdinger:
I just confirmed that it still looks the same in a new firefox profile also.

alexcrichton:
Ok, may just be odd browser/size differences.

This site doesn't auto-deploy, I have to push manually, I'll probably do so once I've gotten through #177

mdinger:
Okay, that's cool.

mdinger:
@alexcrichton Did you deploy this? This is my current screenshot for Aug 17, 2015:

regex_aug17

If you deployed and it still doesn't fix it, I have no idea what the issue is. It fixed it when I tried locally.

alexcrichton:
Ah no I haven't deployed this just yet, I'm finishing up with @Gankro's work on teams and I hope to deploy this and that both soon

mdinger:
Ah, silly me. I should've looked at that PR. I thought for some reason that PR would've been finished the day you mentioned it. Don't know what I was thinking. Looking at it further didn't even occur to me.

Thanks!

PR #184, opened 2015-08-15
Accept SPDX license expressions in the license field.

withoutboats:
The [SPDX spec](https://spdx.org/SPDX-specifications/spdx-version-2.0) defines not only licenses but also "license expressions," which allow multiple licenses, or licenses and license exceptions, to be combined. My personal preference for licensing free software libraries is one such expression: `GPL-3.0+ WITH Classpath-exception-2.0`.

This PR replaces the body of the validate_license function in src/krate.rs with a thin wrapper around a function defined in [an external crate](https://github.com/withoutboats/license-exprs), which validates arbitrary license expressions. It does not eliminate the crates.io-specific `/` syntax for combining multiple licenses (or, now, license expressions). All currently valid licenses remain valid after this change (including licenses deprecated by the SPDX standard).


alexcrichton:
Does this still accept the old `/` syntax? Looking at the spdx crate it looks like it's no longer accepted? I'd be fine with some crates.io-specific logic for accepting that syntax as well.

withoutboats:
Oof, I meant for that not to change. I thought the splitting on `/` was applied outside of the validate_license function.

Fixed. The intent (now correctly expressed) is just that `/` can be used to divide license expressions.

alexcrichton:
Thanks!

PR #188, opened 2015-08-27
Add a tool to generate test data

achanda:
Sometimes, it's easier to work on UI changes with some test data.
This script can be modified later to add more functionality if
necessary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,10 @@ test = false
 name = "populate"
 test = false
 
+[[bin]]
+name = "generate-test-data"
+test = false
+
 [[test]]
 name = "all"
 path = "src/tests/all.rs"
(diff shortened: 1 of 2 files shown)


achanda:
@alexcrichton does this look useful?

alexcrichton:
Did this end up providing a useful data set to work off? In the past I've found that manually generating data like this is a little lossy in the sense that it's not very complete (e.g. lots of model connections would be missing here). It may be better to script hitting the API endpoints perhaps?

achanda:
The data this generated was fine for my use, I just wanted to see something for search results etc. But yeah, I do agree that the API should be used in general.

mschmo:
What about pulling a dump of the staging DB and importing it to the developer's DB?

carols10cents:
I've decided to close this since it's over a year old. The test data this generates isn't quite what gets generated on the live site, so I'm worried it wouldn't be close enough for accurate testing. I think the `yarn run start:staging` is a better option for working on the frontend locally with data that's more real.

PR #190, opened 2015-08-30
Added reserved keywords to the list of reserved crate names.

withoutboats:
Unfortunately for anyone hoping to name their project `override` or `virtual`, it is currently not possible to import crates whose names are reserved words, because the name of the crate is required to be an ident. This was recently discovered to be a problem thanks to the now-deprecated [proc crate](https://crates.io/crates/proc).

This just adds all of the reserved words documented in [the reference](https://doc.rust-lang.org/grammar.html#keywords) to the list of reserved crate names, so that users cannot upload crates to crates.io that other users would not be able to depend on.

I have not (yet) checked if there are any existing crates other than proc that would be "broken" by this change. If there are, I figure one would want to give them a chance to update their projects to announce a rename before enacting this change.


alexcrichton:
This seems like something that'll weed itself out naturally, not something we need to catch explicitly on crates.io. Not all packages may end up being imported as well, with `cargo-install` we'll perhaps start seeing a large number of binary-only packages in which case these crate names should be fine I think.

withoutboats:
All of that is true. Its a trade off between maintaining this list and a rare user issue; in either course the burden is not that great. Up to you.

alexcrichton:
Ok, in that case I'd prefer to let these kinds of cases weed themselves out, so I'm gonna close this. Thanks for the PR though!

PR #192, opened 2015-09-09
Add a table to track rust version compatibiltiy of crates

achanda:
The rust_version table will track what rust versions is a given
crate compatible with. There is a foreign key relationship
with the crates table on crate id.


alexcrichton:
Thanks! I prefer to land changes like this all in one go though instead of having extra tables lying around in the database (this may also need to change during development), so feel free to submit this as part of a larger PR!

Issue #162, opened 2015-06-23
Duplicate items in search results


gkoz:
Example:
```shell
$ for i in `seq 1 10`; do curl -s "https://crates.io/api/v1/crates?q=write&page=${i}&per_page=10" | jq '.crates[].name'; done | sort | uniq -cd
      2 "drossel-journal"
      2 "external_mixin_umbrella"
      2 "id3"
      2 "las"
      2 "leveldb"
      2 "pnet"
      2 "probe-c-api"
      2 "racc"
      2 "recycler"
      2 "rs-es"
      2 "rush"
      2 "rust_cli"
      2 "rust-turbo"
      2 "selecta"
      2 "show"
      2 "snappy_framed"
      2 "sysfs_gpio"
      2 "winreg"
$
```

gkoz:
Can't reproduce any more, hopefully because the fix has been deployed.

PR #199, opened 2015-10-11
In a crate search query, order the results by name in addition to rank

gkoz:
Since `ts_rank_cd()` doesn't provide unique ordering using it with `LIMIT ... OFFSET` will lead to inconsistent results, as is apparent from #162. Add ordering by name to break the ties.


alexcrichton:
Thanks!

Issue #197, opened 2015-10-05
[feature request] well-known URLs for crate links


durka:
I often type out a crates.io link, e.g. crates.io/crates/libc, with the sole intention of scrolling down and clicking "Documentation". But what if I could type crates.io/crates/libc/docs or crates.io/docs/libc and it would automatically redirect to the crate's doc link? Similarly for the homepage and repository links. A 404 is fine if the crate hasn't provided a link.

Does this sound like a useful feature?

[moved from rust-lang/cargo#2024]

alexcrichton:
Sounds like a good idea to me!

Turbo87:
@durka I guess this can be closed now that #207 is merged, right?

durka:
Yes. Though now that it's simple I might open a new PR to add "/home" and "/repo" :)
How often is the live site rebuilt?

Turbo87:
no idea, asked @alexcrichton the same thing in #208 just now

durka:
Ping?

alexcrichton:
Oh sorry, the site is deployed manually and I haven't done so yet (hit a few bugs and such), I hope to get it updated this week, however!

PR #207, opened 2015-11-03
Well-known routes

dirk:
Implementation of rust-lang/crates.io#197.

#### Things to discuss:

- [ ] Possible CSRF action if a crate has a malicious `documentation` URL?
- [x] This currently transitions to the crate's main page if there isn't a documentation URL; is that the behavior we want? (Update: No, showing a brief [flash message](#issuecomment-153287635) before redirecting.)
- [ ] Should we take this opportunity to add more well-known paths? (@durka suggested homepage and repository in their original issue.)

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,25 @@
+# Credentials for uploading packages to S3, these can be blank if you're not
+# publishing locally.
+S3_BUCKET=
+S3_ACCESS_KEY=
+S3_SECRET_KEY=
+# Not needed if the S3 bucket is in US standard
+S3_REGION=      
+
+# Credentials for talking to github, can be blank if you're not logging in.
+#
+# When registering a new application, be sure to set the callback url to the
+# address `http://localhost:4200/authorize/github`.
+GH_CLIENT_ID=
+GH_CLIENT_SECRET=
+
+# Key to sign and encrypt cookies with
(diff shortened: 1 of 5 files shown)


durka:
Awesome!

steveklabnik:
>  is that the behavior we want?

I would suggest at least a flash message.

dirk:
@steveklabnik: Done! Here's a little preview of it when the documentation URL is present:
image

alexcrichton:
This looks pretty good to me, thanks @dirk! I agree with others in that an instantaneous transition is probably the best way to go here, the page can still say "redirecting to ..." but scheduling the redirect ASAP seems best.

I also agree with @steveklabnik that a missing doc URL should probably just emit a flash message and redirect back to the crate itself.

dirk:
Cool, thanks for you guise input! I'll get this implemented today. (Sorry for not implementing this yesterday, was a bit under the weather.)

dirk:
@durka, @steveklabnik: Revised UX is implemented. :smile:

alexcrichton:
OK, looks good to me, thanks @dirk!

dirk:
@alexcrichton: Yup, glad to help! :tada:

PR #212, opened 2015-11-06
Deploy "cargo doc" to GitHub Pages

Turbo87:
This PR adds a deploy step to the `master` branch which will build the cargo docs and publish them to the `gh-pages` branch. There are two more steps needed to make this work, but I can only do them with write access to the repo:

1. Add a deploy key to the repo at https://github.com/rust-lang/crates.io/settings/keys

    ```
    ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC8j7Sx3Ons2ghbYwpXMIvJGB5aFAnKpgVyTupJM0hZbclUnHBbhjv3R/Z9c1cruS1wsgU+OdgQJZayphAieggheI/yBaNwsW+WDe2BLUQ1KZQ4w8EjSXNj/LRlGTJJUj8J5F2AE/S3hT6MyrAJBhboGDbxlRqZIa1p0ruE7wGYRVnucluAmnL3McKGCmAXH/Nv6CfAkB+PHrjyCIypgFfZBpt157e3n37vxbxwCXIVZayfBRpRdcH3Vu7NMSjCTz9xdtxxyvZrjf5qoSnvMA5JxJJw6MV8zJ0AD8absLHdfmNoJE26K8G9VAaaf96PJwKuHcDfAsE2JquQwziV3Oo9
    ```

2. Add the secret IV/Key environment variables to the Travis CI build that are needed to decrypt the private key

I've tested this out on my forked repository and it works pretty well: http://turbo87.github.io/crates.io/

For the second part I would need temporary write access to the repo or a way to privately tell you the values.

For reference the key was generated and encrypted like this:

```shell
ssh-keygen -f .travis/deploy_key.pem -N ""
travis encrypt-file .travis/deploy_key.pem .travis/deploy_key.pem.enc -p
```


alexcrichton:
Perhaps this could [use `travis-cargo`](https://github.com/alexcrichton/git2-rs/blob/master/.travis.yml#L22) to publish docs? That should take care a lot of the logic here and mean that I just need to encrypt `GH_TOKEN` to get the push to work.

Turbo87:
I've never worked with travis-cargo before. The issue with the token
approach is that you need to allow write access to **all** your repos while
a deploy key can be limited to a single repo.

alexcrichton:
Hm true, although that's currently in wide use so it's not too too bad (e.g. it's not making things any worse than they already are)

huonw:
travis-cargo could probably be tweaked to allow using a deploy key (although I don't know for sure).

PR #213, opened 2015-11-12
Forbid wildcard dependency constraints

sfackler:
This shouldn't be deployed until December 11th, but I'm opening it now so we don't forget.

### DIFF
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -228,6 +228,25 @@ fn new_krate_with_dependency() {
 }
 
 #[test]
+fn new_krate_with_wildcard_dependency() {
+    let (_b, app, middle) = ::app();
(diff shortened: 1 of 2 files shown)


huonw:
Maybe this could include a link to http://doc.crates.io/faq.html#can-libraries-use--as-a-version-for-their-dependencies ?

sfackler:
Updated

alexcrichton:
lgtm, r+ modulo a month passing

sfackler:
Update: we're pushing this back to January 22nd since the Cargo warning didn't make it into stable until 1.5.

sfackler:
@alexcrichton Time to merge!

alexcrichton:
Indeed!

PR #208, opened 2015-11-03
Update Ember.js

Turbo87:
This PR aims to continue the Ember.js update process that was started in #152. So far I've only merged the changes from the current `master` branch into the existing `updates` branch.

/cc @stefanpenner @alexcrichton 

resolves #156

PR #214, opened 2015-11-13
Update Ember (again)

Turbo87:
This PR updates the Ember app to the most recent v1.13 version and fixes the remaining bug mentioned in #208


alexcrichton:
Thanks!

I'll deploy this to staging now and then hopefully get around to prod tomorrow

alexcrichton:
Hm it seems like there's a bit of a more deep-seated issue here that may warrant some more investigation, for example if you go to [the all crates index page](https://staging-crates-io.herokuapp.com/crates) and click on `cargo` you'll get the same "Version `null` does not exist" error again

Turbo87:
That's what I feared... I guess this is related to https://github.com/rust-lang/crates.io/pull/208#issuecomment-155714465. In this case the crates index page returns the crates with `versions: null` which unfortunately gets converted to `versions: []` and appears as if it was sideloaded.

I suspected this to be a bug in Ember Data, but this seems to be expected behavior. What we need to do is implement an adapter/serializer which filters out `versions: null` and replaces it with undefined.

PR #215, opened 2015-11-13
Documentation

Turbo87:
This PR adds documentation comments to the `user` module.

bildschirmfoto 2015-11-14 um 00 48 58

### DIFF
--- a/src/bin/delete-crate.rs
+++ b/src/bin/delete-crate.rs
@@ -18,6 +18,7 @@ use std::io::prelude::*;
 
 use cargo_registry::Crate;
 
+#[allow(dead_code)]
 fn main() {
     let conn = postgres::Connection::connect(&env("DATABASE_URL")[..],
                                              &postgres::SslMode::None).unwrap();
@@ -97,4 +98,3 @@ fn delete(tx: &postgres::Transaction) {
     io::stdin().read_line(&mut line).unwrap();
     if !line.starts_with("y") { panic!("aborting transaction"); }
 }
-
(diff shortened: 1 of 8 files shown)


Gankro:
:confetti_ball: awesome!

src/bin/delete-crate.rs
```diff
@@ -18,6 +18,7 @@ use std::io::prelude::*;
 
 use cargo_registry::Crate;
 
+#[allow(dead_code)]
 fn main() {
```
Gankro: iirc the "proper" way to do this is `#[cfg(not(test))]`. May be misremembering for `doc`.

src/user/mod.rs
```diff
@@ -20,6 +20,8 @@ pub use self::middleware::{Middleware, RequestUser};
 
 pub mod middleware;
 
+/// The model representing a row in the `users` database table.
+///
```
Gankro: We don't add trailing ///'s

Gankro:
No opinion on the middleware changes. (I also don't have review rights anyway, so y'all'll need to wait on someone else for that)

src/bin/delete-crate.rs
Turbo87: if I replace this with `#[cfg(not(test))]` I'm getting the following soft of errors again:

```
src/bin/delete-version.rs:22:1: 31:2 error: function is never used: `main`, #[deny(dead_code)] on by default
```

src/user/mod.rs
Turbo87: I felt like it was much easier to read the doc if they have a little room to breathe, but if this is not desired I can remove the extra line

alexcrichton:
Thanks!

PR #216, opened 2015-11-16
Fix remaining Ember Data problem

Turbo87:
Resolves the remaining issue in #214

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -5,12 +5,7 @@ export default Ember.Route.extend({
   model() {
     function addCrates(store, crates) {
         for (var i = 0; i < crates.length; i++) {
-            const crate = crates[i];
-            if (crate.versions == null) {
-              // passing `null` will return an empty versions array
-              delete crate.versions;
-            }
-            crates[i] = store.push(store.normalize('crate', crate));
+            crates[i] = store.push(store.normalize('crate', crates[i]));
         }
     }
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
[You mention](https://github.com/rust-lang/crates.io/pull/214#issuecomment-156950624) that this is because we're sending `versions: null`, but perhaps we could omit the `versions` key entirely when sending data? That should in theory have the same effect, right?

Turbo87:
That would probably have the same effect, yes. I don't have enough experience with the codebase yet to implement it there though. I would suggest to use the current solution for now and when we later refactor the api we can revisit this problem.

Turbo87:
ah shit, hadn't seen that one yet... I've been writing too much Kotlin lately :wink:

Turbo87:
should be fixed now

alexcrichton:
Thanks!

alexcrichton:
Ok, I've now deployed this, (thanks for all the fixes @Turbo87!)

One issue I see is that the "other downloads" section of the graphs isn't updated when you navigate between crates, for example if you click on the winapi crate and then on the libc crate you see this graph:



but if you reload the page you see:

Turbo87:
awesome! :tada: 

I'll see what I can do about the remaining issue. There is also one more deprecation warning type that I have not fixed yet regarding request metadata. If that one is fixed we are on track to Ember v2.x.

PR #218, opened 2015-11-18
Allow per-crate max upload sizes

### LINKED ISSUE #40: Support *-sys packages larger than 10MB (somehow)
(Continuing a discussion started [here](http://www.reddit.com/r/rust/comments/2mwice/cratesio_has_shipped/cm8b0o7).)

The [`cld2` library](https://code.google.com/p/cld2/) is a natural-language detection library from Google, and it [does some pretty cool stuff](https://github.com/emk/rust-cld2/blob/master/examples/analyze_texts.rs). I've packaged it as [two Rust libraries, `cld2` and `cld2-sys`](https://github.com/emk/rust-cld2). But because the upstream `cld2` library is packaged by very few Linux distributions, I've chosen to distribute the source code with the `cld2-sys` package and build it using the Rust `gcc` library. So far, so good—all this works quite nicely.

But I can't upload the package to crates.io because it contains statistical language models, and those models are just too big:

```
$ du -sh target/package/cld2-sys-0.0.1.crate 
35M	target/package/cld2-sys-0.0.1.crate
```

I can shrink this down somewhat (by omitting everything I don't need for the build), but I almost certainly can't get it under the 10MB limit. I can think of a couple of ways to address this issue:

1. Accept that certain `*-sys` packages will be larger than 10MB, and provide some way to override the limit selectively.
2. Store compressed source code in an S3 bucket, and ask `build.rs` to download it. But this introduces a dependency on an outside data source that may go away.

Any thoughts on the best way to handle this? Thank you for your advice, and for a great package-management system!

### LINKED ISSUE #195: Raise the crate size limit?
Servo regularly gets bugs like https://github.com/servo/servo/issues/7687 filed because downloading 320 MB of git history for servo/skia takes such a long time. (This repository is used as a git dependency.) 

Shallow clones (https://github.com/rust-lang/cargo/issues/1171) would fix this (`git clone --depth=1` fetches about 11 MB worth of objects), but according to @Manishearth libgit2 support is unlikely to happen.

We’re playing with hacks like https://github.com/servo/skia-snapshots/blob/master/update.sh , but they seem brittle and require jumping through hoops to make updates.

Another option would be to publish on crates.io, which has a well-established process to make updates. However `cargo package` creates a 12 MB `.crate` file, which is over the current 10 MB limit. This is a big pile of C++ source code. Splitting it up seems challenging at best.

Could the limit be raised a bit, say to 15 or 20 MB? Or, could we add a way for crates.io administrators to grant a derogation to individual crates?

CC @metajack @larsbergstrom

alexcrichton:
There's still a global limit on the nginx server but each crate can now have its
own maximum limit as well which is larger than the standard limit.

Closes #40
Closes #195


src/krate.rs
```diff
@@ -687,6 +692,16 @@ pub fn new(req: &mut Request) -> CargoResult<Response> {
         return Err(human(format!("crate was previously named `{}`", krate.name)))
     }
 
+    let length = try!(req.content_length().chain_error(|| {
+        human("missing header: Content-Length")
+    }));
+    let max = krate.max_upload_size.map(|m| m as u64)
+                   .unwrap_or(app.config.max_upload_size);
+    if length > max {
+        println!("guth");
```
huonw: Maybe this could be slightly more... eloquent?

alexcrichton: oops

Issue #220, opened 2015-11-19
Downloads Graph missing data


Turbo87:
Migrated from #216:

/cc @alexcrichton 

---

One issue I see is that the "other downloads" section of the graphs isn't updated when you navigate between crates, for example if you click on the winapi crate and then on the libc crate you see this graph:

![](https://i.imgur.com/pBNlvn0.png)

but if you reload the page you see:

![](https://i.imgur.com/eOV2SOL.png)

Turbo87:
hmm... I've tried to reproduce this problem, but it doesn't seem to happen consistently. It worked as expected for me most of the time and I've only seen it happen once. Not sure yet what might be causing this or how to reliably trigger it.

PR #221, opened 2015-11-19
Resolved remaining Ember 1.13 deprecations

Turbo87:
After this PR is merged we should be alright to upgrade to Ember 2.x without any major issues.

### DIFF
--- a/app/controllers/crate/index.js
+++ b/app/controllers/crate/index.js
@@ -42,7 +42,7 @@ export default Ember.Controller.extend({
         return DS.PromiseArray.create({
             promise: this.get('currentVersion.authors').then((authors) => {
                 var ret = authors.slice();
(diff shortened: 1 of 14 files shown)


alexcrichton:
Nice! I wonder if this different handling of `meta` will auto-fix https://github.com/rust-lang/crates.io/issues/220?

Turbo87:
It shouldn't, but if it does that sure would be great ;)

alexcrichton:
Ah looks like it's still an issue, I went to the main page of crates.io, clicked libc, clicked the home page, clicked winapi, clicked home page, and clicked libc again and the graph was odd.

Turbo87:
that is such a weird problem... I've tried it with those instructions and could reproduce it on first try, but the second, third and fourth time it just worked

PR #225, opened 2015-11-20
routes/crate: Fix meta.extra_downloads handling

### LINKED ISSUE #220: Downloads Graph missing data

Turbo87:
This might resolve #220 but since I can't reliably reproduce it I'm not sure. In any case it fixes the currently broken download graphs for crate version routes like https://crates.io/crates/libc/0.2.2. 

Sorry about that!

### DIFF
--- a/app/routes/crate/index.js
+++ b/app/routes/crate/index.js
@@ -52,11 +52,7 @@ export default Ember.Route.extend({
             });
 
         }).then(() => {
-            if (controller.get('requestedVersion')) {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice!

PR #228, opened 2015-11-26
Rely on the database for data things

sgrif:
There's a lot of places in the code where there's duplicated hard coded default values for certain tables, or manual updating of things which the database can handle for us (like counter caches and `updated_at`). This updates as many of these cases as I could find to rely on the database when possible. 

We shouldn't require the person adding new code to know that they have to manually increment a counter on another table, or manually set `updated_at` to the right value when they change a row. (As an example of this point, I found that `versions.updated_at` was never actually changed after insert, and several places that it seems we would want `crates.updated_at` to be set, but weren't).

The trigger to update the counter cache on `keywords` will end up performing more queries than what we were doing before (one per deleted row instead of one per delete statement), however this should not cause any problems, as we're not paying a DB round trip cost, and the index should still only be updated once at the end of the trigger.

One other case that might make sense, but I did not do here is to move the setting of `max_version` to a database trigger as well. However, implementing proper semver comparison is arguably more logic than I think belongs in the database.

I've kept each change as a separate commit to ease reverting if there's any specific changes here where there's contention, but I'll be happy to squash if these changes look good.

/cc @alexcrichton

### DIFF
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -492,7 +492,8 @@ fn migrations() -> Vec<Migration> {
             ALTER TABLE version_downloads ALTER date SET DEFAULT current_date;
             ALTER TABLE version_downloads ALTER processed SET DEFAULT 'f';
 
-
+            ALTER TABLE keywords ALTER created_at SET DEFAULT current_timestamp;
+            ALTER TABLE keywords ALTER crates_cnt SET DEFAULT 0;
             "));
             Ok(())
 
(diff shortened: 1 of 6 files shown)


Turbo87:
are you planning on porting the backend to diesel?

sgrif:
@Turbo87 I've been using crates.io as my "real world" app to drive feature development until now. I'll likely submit a PR at some point in the future, but my main goal has just been to have a sandbox to explore the API in.

Turbo87:
Okay, cool. From the blog post it looked pretty great already. Nice work!

alexcrichton:
Looks fantastic to me, thanks @sgrif! 

I thought for a moment that there may be some oddness that arises between `::now()` vs `current_date` if there's some out-of-sync clocks between the db and the webserver, but this pattern is already all over the place so no worries really.

I originally liked having all columns explicit as silent defaults have bit me in the past, but the code here I think is much more readable and more amenable to extensions in the future, so I'm gonna merge.

PR #229, opened 2015-12-03
Ember: Use JSCS for code style checking

Turbo87:
This PR adds the [ember-suave](https://github.com/dockyard/ember-suave) plugin to the Ember.js app, which is using [JSCS](http://jscs.info/) for JavaScript code style checking. The resulting style inconsistencies were fixed afterwards.


alexcrichton:
Nice!

PR #231, opened 2015-12-07
Ember: Acceptance Tests

Turbo87:
This PR adds two basic acceptance tests for the index page and the search functionality. The implementation is probably not perfect yet but I wanted to get it out there before going on vacation next week. I'll continue improving it after I get back in January.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,5 +1,6 @@
 {
   "predef": [
+    "server",
     "document",
     "window",
     "-Promise"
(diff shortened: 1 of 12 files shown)


Turbo87:
@alexcrichton feel free to already merge this at any point

alexcrichton:
Oh this is awesome, thanks @Turbo87! Sorry for the delay here, been at a work week in Orlando

Turbo87:
right, I read about mozlando somewhere :wink:

this should hopefully make upgrading to Ember 2.0 a bit safer in the future. we are currently using a private API on the front page for pushing data into the store and if that breaks the test should be able to catch it.

PR #242, opened 2015-12-21
fix highlight bar going outside rounded corner

princemaple:
The problem:
untitled

With the fix:
untitled2


alexcrichton:
Thanks!

PR #243, opened 2015-12-21
Fixed unaligned Getting Started button in Safari.

Racum:
The "Getting Started" button was a bit lower compared to "Install Cargo" in Safari.

Tested in OS X 10.11.2 and Safari 9.0.2. Doesn't break in Chrome 47 nor 43.


alexcrichton:
Thanks!

Issue #254, opened 2016-01-29
Make keyword links case-independent


kud1ing:
- This gives a result: https://crates.io/keywords/Amazon
- This gives an error message: https://crates.io/keywords/amazon

It would be a pitty if "postgresql" and "Postgresql" or "PostgreSql" would be treated as different keywords.

I suggest to make the lower case keyword url match all variants of a keyword.

gsquire:
If this is desired behavior, I think I can tackle this. I'm not sure if crates are treated as case-sensitive.

alexcrichton:
Sure! Crates are currently case insensitive today, yeah, so there should be some example code handling this.

carols10cents:
I think this is fixed! Please reopen if I'm mistaken! ❤️

kud1ing:
Seems fixed. Thank you.

PR #255, opened 2016-01-30
Case-Insensitive Keyword

gsquire:
This is a simple fix to address #254


alexcrichton:
Thanks! I think that we may want to go a bit further here, however. There's a few things that I think this may want to do as well:

* Uniqueness in the database may want to be guaranteed to avoid duplicate keywords like `FOO` and `foo`
* The index on the database will likely need to be changed to account for the `LOWER` function.
* We may want to canonicalize all inserts of new keywords. For example we could lowercase all old ones and auto-lower-case all new ones.

alexcrichton:
Ah, and can you also be sure to add some tests for this change as well?

gsquire:
Sure, those are all important.
- Is uniqueness already enforced [here](https://github.com/rust-lang/crates.io/blob/da954b647fd51cad87cc982815e8fc28eaef2127/src/bin/migrate.rs#L372)?
- Is there an [index](https://github.com/rust-lang/crates.io/blob/da954b647fd51cad87cc982815e8fc28eaef2127/src/bin/migrate.rs#L372) with LOWER too?

Forgive me if I have misunderstood. I will work on adding a test too.

gsquire:
I will need to look at this further later as to why the uniqueness constraint is failing.

gsquire:
I had to change some tests because of a uniqueness constraint. I'm not sure if that was the most principled approach, however. I can squash my changes in the future too to get rid of some tiny commits.

alexcrichton:
Ah yeah I think the easiest thing to do here would be to insert the lowercase version into the database (as you're doing here), and then the existing uniqueness constraints and indexes should continue to work.

src/keyword.rs
```diff
@@ -33,7 +33,7 @@ impl Keyword {
     pub fn find_by_keyword(conn: &GenericConnection, name: &str)
                            -> CargoResult<Option<Keyword>> {
         let stmt = try!(conn.prepare("SELECT * FROM keywords \
-                                      WHERE keyword = $1"));
+                                      WHERE LOWER(keyword) = $1"));
```
alexcrichton: This might not hit the index because it's strictly on just the `keyword` column, but I think here we'll actually want `LOWER($1)` instead as we want to lowercase the user input

alexcrichton: (same below as well)

src/keyword.rs
```diff
@@ -164,8 +164,9 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 /// Handles the `GET /keywords/:keyword_id` route.
 pub fn show(req: &mut Request) -> CargoResult<Response> {
     let name = &req.params()["keyword_id"];
+    let lower_name = name.to_lowercase();
     let conn = try!(req.tx());
-    let kw = try!(Keyword::find_by_keyword(&*conn, &name));
+    let kw = try!(Keyword::find_by_keyword(&*conn, &lower_name[..]));
```
alexcrichton: The database already turns this into lowercase, so I don't think we need the explicit `to_lowercase` here

src/tests/keyword.rs
```diff
@@ -57,32 +68,32 @@ fn update_crate() {
...
-    Keyword::update_crate(tx(&req), &krate, &["kw1".to_string(),
+    Keyword::update_crate(tx(&req), &krate, &["kw3".to_string(),
                                               "kw2".to_string()]).unwrap();
-    assert_eq!(cnt(&mut req, "kw1"), 1);
+    assert_eq!(cnt(&mut req, "kw3"), 1);
     assert_eq!(cnt(&mut req, "kw2"), 1);
 
     Keyword::update_crate(tx(&req), &krate, &[]).unwrap();
-    assert_eq!(cnt(&mut req, "kw1"), 0);
+    assert_eq!(cnt(&mut req, "kw3"), 0);
```
alexcrichton: It looks like kw1 was just renamed to kw3 everywhere here, was that necessary?

src/tests/krate.rs
```diff
@@ -78,7 +78,7 @@ fn index_queries() {
     let mut req = ::req(app, Method::Get, "/api/v1/crates");
     let u = ::mock_user(&mut req, ::user("foo"));
     let mut krate = ::krate("foo");
-    krate.keywords.push("kw1".to_string());
+    krate.keywords.push("kw0".to_string());
```
alexcrichton: how come this rename was necessary?

alexcrichton:
Thanks @gsquire! Can you also prepare a migration to handle all the old keywords? It probably wants to do something along the lines of:

1. Iterate over all keywords by id ascending
2. Update the keyword name to be the lowercase version
3. If that failed, then connect everything up to the existing keyword

Does that make sense?

alexcrichton:
ah and:

4. Delete the duplicate row

src/keyword.rs
gsquire: That makes sense, fixed.

src/keyword.rs
gsquire: You're right, I overlooked this when I updated the queries.

src/tests/keyword.rs
gsquire: If you look at this [build](https://travis-ci.org/rust-lang/crates.io/builds/106411768) you can see that the test conflicts on the unique constraint causing it to fail. This is the reason for the rename here and in the other file.

I hope this actually tests that the keywords now are unique :)

src/tests/krate.rs
gsquire: See above.

gsquire:
I tried adding a migration. I get how to do the update, the rollback however is unclear to me.

src/tests/keyword.rs
alexcrichton: Hm I may not be following this, but it looks like every instance of kw3 was replaced with kw1? Does that mean that kw1 already existed in the database? (that's a bug)

src/bin/migrate.rs
```diff
@@ -619,6 +619,9 @@ fn migrations() -> Vec<Migration> {
             "));
             Ok(())
         }),
+        Migration::run(20160202210722,
+                       "UPDATE keywords SET keyword = LOWER(keyword)",
+                       "UPDATE keywords SET keyword = keyword"),
```
alexcrichton: Unfortunately this migration will trigger a violation when run over crates.io, for example the keyword `mouse` exists as well as `Mouse`, and this will cause the two to conflict.

I think the migration will have to basically handle each keyword in turn and custom merge it into another keyword if it exists.

gsquire: Ah, that's what you meant about the merging. I'll have to rethink the approach.

gsquire:
@alexcrichton I have been trying to come up with an idea for a migration but I think at this point it's beyond my DB knowledge. Before this PR goes to stale I figured I would let you know. If you have bandwidth, I would appreciate some help.

alexcrichton:
Sure, I'll see if I can't find some time to push this forward soon

tari:
I had a go at writing a migration for this (below). Didn't test it thoroughly, but it seems correct.

```diff
diff --git a/src/bin/migrate.rs b/src/bin/migrate.rs
index b0b6a3b..e43b265 100644
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -682,6 +682,52 @@ fn migrations() -> Vec<Migration> {
                 DROP FUNCTION touch_crate();
             ")
         }),
+        Migration::new(20160326123149, |tx| {
+            use postgres::error::{Error, SqlState};
+
+            for row in try!(tx.query("SELECT id FROM keywords ORDER BY id", &[])).iter() {
+                let kw_id: i32 = row.get(0);
+                let err = {
+                    let tx = try!(tx.transaction());
+                    let res = tx.execute("UPDATE keywords SET keyword = LOWER(keyword)
+                                          WHERE id = $1", &[&kw_id]);
+                    match res {
+                        Ok(n) => {
+                            assert_eq!(n, 1);
+                            try!(tx.commit());
+                            continue;
+                        },
+                        Err(e) => /* Rollback update, fall through */ e
+                    }
+                };
+
+                match err {
+                    Error::Db(ref e) if e.code == SqlState::UniqueViolation => {
+                        // There is already a lower-case version of this keyword.
+                        // Merge into the other one.
+                        let target_id: i32 = try!(tx.query("
+                            SELECT id FROM keywords WHERE keyword = LOWER((
+                                SELECT keyword FROM keywords WHERE id = $1
+                            ))
+                        ", &[&kw_id])).get(0).get(0);
+
+                        try!(tx.batch_execute(&format!("
+                            UPDATE crates_keywords SET keyword_id = {}
+                                WHERE keyword_id = {};
+                            UPDATE keywords SET crates_cnt = crates_cnt + (
+                                SELECT crates_cnt FROM keywords WHERE id = {}
+                            ) WHERE id = {};
+                            DELETE FROM keywords WHERE id = {};
+                        ", target_id, kw_id, kw_id, target_id, kw_id)));
+                    },
+                    e => return Err(e)
+                }
+            }
+
+            Ok(())
+        }, |_tx| {
+            Ok(())
+        }),
     ];
     // NOTE: Generate a new id via `date +"%Y%m%d%H%M%S"`
```

gsquire:
@tari Awesome! Thanks for adding this in. I am happy this can keep going forward :+1:

alexcrichton:
@tari that looks good to me as well, thanks! @gsquire could you add that to this PR and I'll look to merging/deploying afterwards.

gsquire:
@alexcrichton I added the migration, thanks again to @tari for adding this in. Would you like me to clean up the commits as well before you merge?

gsquire:
@alexcrichton I forgot this PR is still open. I'll see why the coverage decreased.

PR #258, opened 2016-02-01
another stab at fixing the trailing-slash redirect

### LINKED ISSUE #236: trailing slash redirect steals back button
Steps to reproduce:

1. Go to [https://crates.io/crates/libc](https://crates.io/crates/libc) (note the lack of trailing slash).
2. Observe the automatic redirect to [https://crates.io/crates/libc/](https://crates.io/crates/libc/).
3. Try the Back button -- you don't end up back here.

### LINKED ISSUE #252: clicking a search result removes the results page from history
Steps to reproduce:

1. Go to https://crates.io/
2. Search for "foo" and click on the first link ("tenacious").
3. Search for "bar" and click on the first link ("bar").
4. Click the back button. You end up back at the tenacious page, skipping the search results page.

Possibly related to #236 or its fix.

durka:
According to my calculations, this reverts #251 while still fixing #236 but _not_ causing #252.

I'm not sure this is the right way to do it (checking the actual URL) but it seems to work.

Fixes #236 (again).
Fixes #252.

I may have missed something, so someone else please check the back button behavior in all these use cases:

- External site links to a crate page (with or without a trailing slash)
- Type in the URL to a crate page
- Homepage links to a crate
- Link from a crate to dependency
- Do a search from a crate page

### DIFF
--- a/app/routes/crate/index.js
+++ b/app/routes/crate/index.js
@@ -2,6 +2,6 @@ import Ember from 'ember';
 
 export default Ember.Route.extend({
     redirect() {
-        this.replaceWith('crate.version', '');
+        this.transitionTo('crate.version', '');
     }
 });
--- a/app/routes/crate/index.js
+++ b/app/routes/crate/index.js
@@ -1,7 +1,12 @@
 import Ember from 'ember';
 
 export default Ember.Route.extend({
-    redirect() {
-        this.transitionTo('crate.version', '');
+    beforeModel() {
+        const url = this.get('router.url');
+        if (url.startsWith('/crates/') && !url.endsWith('/')) {
+            this.replaceWith('crate.version', '');
+        } else {
+            this.transitionTo('crate.version', '');
+        }
     }
 });


durka:
cc @ryanwilliams

alexcrichton:
Oh dear, I'm stretched basically to the limit of my already nonexistent ember knowledge... I'll see if I can't help investigate though.

durka:
Hey, I know nothing about Ember either!

I couldn't even figure out exactly why the trailing-slash redirect occurs -- it seems like it's redirecting from `/crates/<crate name>` to `/crates/<crate name>/<version>` where `<version>` is empty? But, for some reason this redirect isn't needed when coming from outside of crates.io?

alexcrichton:
So I talked with @wycats briefly about this on IRC, and the conclusion is that the best thing to do here would actually be to synthetically populate the history stack. I think that we may wish to always use `replaceWith` except that if this was also a link triggered from within the app itself we push a piece of history onto the stack.

That way external links won't push a history item on the stack (and just do the silent redirect) but internal links will push something on the stack for where they came from. Does that make sense?

durka:
Makes sense to me. I will look into how to manually push a history item. However, from my limited research it seems difficult to find out the previous Ember route, unless we add some infrastructure to remember it?

alexcrichton:
Unfortunately I don't think we have any infrastructure on our own to manage that.

durka:
If you search for how to manually maintain a route history in Ember you get a lot of hackish homegrown solutions. It seems like there is a lot of friction, so why exactly do we want to do it that way? cc @wycats ?

Turbo87:
@alexcrichton thanks for talking to @wycats about this. I meant to come back to that issue, but haven't found the time yet. 

I think in any case it would be good to have some acceptance tests for these scenarios. I'll see if I can figure that out, but it looks like using the history API in Ember tests isn't really supported out of the box.

Turbo87:
okay, so I've investigated this a bit more and I believe that the solution that @durka presents here is probably the best although I've modified it slightly like this:

```js
    redirect(model, transition) {
        if (transition.intent.url) {
            this.replaceWith('crate.version', '');
        } else {
            this.transitionTo('crate.version', '');
        }
    }
```

> That way external links won't push a history item on the stack (and just do the silent redirect) but internal links will push something on the stack for where they came from. Does that make sense?

Yes, essentially that is the logic that my code above follows too.

I admit though that I still haven't entirely understood what *exactly* is causing the different behavior. It looks like the `redirect` hook has quite different semantics based on the type of intent in the transition (external vs internal).

alexcrichton:
Thanks @Turbo87! That looks good to me, and I'd also love to have some Ember tests, I just never got around to writing any :(

(I also don't know how...)

nathanhammond:
Ember-familiar dev here. :) From what I can tell you just want the `index` route for `crates` to be `versions`.

Steps:
1. Modify https://github.com/rust-lang/crates.io/blob/master/app/router.js#L17 to include `{ path: '/' }`.
2. Remove the `route/crate/index.js` file.

That should prevent needing the `redirect` at all and make everything play nicely.

Turbo87:
hi @nathanhammond, thanks for looking into this, but unfortunately things are not quite as easy:

- `crates.version` shows the details for one version
- `crates.versions` shows the list of all versions
- `crates.index` should essentially show the details for the most recent version

`crates.index` is currently implemented using a redirect to `crates.version` with an empty `version_id`, which the route will then fill with the most recent version instead. Unfortunately though that `redirect()` hook is behaving in surprising ways (see #236 and #252).

Turbo87:
@nathanhammond I've tried implementing this pattern with your `ember-route-alias` library too, but using `this.alias('index', '/', 'version')` didn't seem to work.

Turbo87:
@alexcrichton looks like we're not the only ones hitting this "bug": https://github.com/emberjs/ember.js/issues/12824

nathanhammond:
Ah, my apologies, I didn't exactly understand what you were attempting. Hooray drive-by code drops. :stuck_out_tongue:

I'm going to take a look at this inside of Ember/router.js. What I believe is happening is that `replaceWith` is snagging the absolute first `transition` created and telling it not to create a history state.

In the mean time I might try `transition.abort()` to forcibly drop one on the floor, and then `this.transitionTo()`, but I'm not sure that actually solves the "first visit" situation. (https://github.com/emberjs/ember.js/issues/10831)

nathanhammond:
First, I just published a new version of `ember-route-alias` that will work for Ember 1.X. You won't get the `rel-link-to` helper but I don't think it matters in this case. With that change you should be able to use this approach:

```
this.route('version', { path: '/:version_num' });
this.alias('index', '/', 'version');
```

You'll need to add a bit of logic into the `version` route to support not having a `version_num`, but it should be relatively straightforward. I'm still looking into addressing what I see as the underlying issue per the referenced Ember issues but that has a longer time horizon. :smile:

Turbo87:
awesome, thanks for your help!

nathanhammond:
For reference: https://github.com/nathanhammond/crates.io/commit/578c8db2e2c802d13efcba89f428e3f36c8fa745

alexcrichton:
Superseded by https://github.com/rust-lang/crates.io/pull/264, but thanks regardless for the PR @durka!

PR #266, opened 2016-02-13
Use rust-dotenv to manage env vars in development

sgrif:
This repo includes a `.env` file, but it doesn't appear to be used (at
least not on the Rust side. I haven't looked at the ember side).

This adds the `dotenv` crate to use the .env file to automatically set
the environment variables from that file when set, without messing with
the acutal environment of the machine.

As part of this, I've taken the liberty of moving the heavily duplicated
`env` function to somewhere common. I've also moved `.env` to
`.env.sample`, and added `.env` to gitignore, as a person's specific
environment variables shouldn't be in the repo, and the `.env` file
provided isn't actually valid.


alexcrichton:
Thanks!

Issue #139, opened 2015-03-09
The "Documentation" drop down is not accessible from the home page


nham:
I think this is a poor design. I needed to look up how the features in Cargo.toml work. That documentation is located on the 'Manifest Format' page, but the link to that page is only accessible from the "Documentation" drop down on either the Getting Started page or the Guide page.

There was a previous issue for this, #121, that I believe was incorrectly closed. If I'm looking for documentation on the manifest format, then "Getting Started" is not what I want. I'm already started with Cargo, I don't need to get started. Also the Guide doesn't contain what I want. It is not great to have to navigate to an irrelevant intermediate page to get access to the drop down menu I need. 

Can this be added to the home page? Maybe there could be some kind of central documentation page that the home page links to?

TimNN:
I wholeheartedly agree with this! 

I especially think that both links currently at the top right (Login & List all crates) provide much less value, than a documentation dropdown would provide.

birkenfeld:
I think this can now be closed.

alexcrichton:
Ah I unfortunately had to roll back https://github.com/rust-lang/crates.io/pull/267 because the alignment was off (see screenshot on the PR)

carols10cents:
Looks like this has been fixed since by making the menu label say "Docs" :)

<img width="1072" alt="screen shot 2016-12-14 at 7 30 28 pm" src="https://cloud.githubusercontent.com/assets/193874/21207072/e4feb7f4-c233-11e6-9756-5bd9f36bbc82.png">

PR #267, opened 2016-02-14
Add Cargo guide link to the top nav bar.

birkenfeld:
(This is just a quick change; haven't tested locally if it looks like it is supposed to.)

I think that the links to the cargo docs are much too hard to find at the bottom of the page. (Especially since you can't find it by searching for "doc".) Neither is the "Getting started" button on the index page a good entry point, since it is only present on that page.

This is kind of what #139 suggests, but without the full drop-down.

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -98,6 +98,12 @@ body {
 
         &.open { display: block; }
     }
+    #doc-links {
+        display: none;
+        left: auto;
+
+        &.open { display: block; }
+    }
     @media only screen and (max-width: 820px) {
         .menu { display: block; }
         .nav { display: none; }
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -25,6 +25,25 @@
             Browse All Crates
         {{/link-to}}
         <span class="sep">|</span>
+        {{#rl-dropdown-container class="dropdown-container"}}
+            {{#rl-dropdown-toggle class="dropdown"}}
+                Documentation
+                <span class='arrow'></span>
+            {{/rl-dropdown-toggle}}
+
+            {{#rl-dropdown tagName="ul" id="doc-links" class="dropdown" closeOnChildClick="a:link"}}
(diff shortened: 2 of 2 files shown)


alexcrichton:
Can you provide a screenshot of what this ends up looking like?

birkenfeld:
Here you go:
2016-02-16-201132_1920x1080_scrot

alexcrichton:
Thanks! It may be good to keep doc.crates.io in sync with crates.io as much as possible, however. The current deviation is from the "log in" link (the links at the bottom match), so I'm somewhat hesitant to allow the two to deviate more?

birkenfeld:
Not sure that goal is more important than quick access to the page you want. But I wouldn't mind them to become *more* in sync by putting the "Documentation" popup from doc.crates.io on crates.io as well (instead of this patch).

BTW, if you want them to be in sync, you might also want to look at the stylesheet - for example, crates.io specifies `"Helvetica Neue",...,sans-serif` for font-face, while doc.crates.io specifies only `sans-serif`.

alexcrichton:
Yeah perhaps we can go that route of adding the whole documentation link instead? I think that's what #139 is basically asking for anyway

birkenfeld:
I'll see what I can do :)

birkenfeld:
Updated, new screenshot below. Note that doc.crates.io has a style for dropdowns that makes the "Documentation v" header look depressed and reverses the arrow while the dropdown is open. crates.io doesn't, currently, but neither does it for the user dropdown menu.

2016-02-26-103808_1920x1080_scrot

birkenfeld:
Reported the style inconsistencies in #273.

alexcrichton:
Thanks!

birkenfeld:
And it's already live. Great!

alexcrichton:
Aww looks like the spacing may be a bit off though?

birkenfeld:
I assume you mean the spacing around the separator? The same happens with the user dropdown though.

As long as you're not logged in, there is less space between separator and menu item (there's no dropdown).

PR #270, opened 2016-02-19
Remove mostly unused keywords column from crates table

sgrif:
This column was present both in the table and in the Rust model, but
was only ever used to update the search index on update. The actual
source of keyword info is the crates_keywords table, and it's actually
reasonably easy for this column to get out of sync.

Instead of relying on things calling the `update_crate` function every
time the keywords column is modified, we can instead populate the column
with a subselect which grabs the keywords associated. This also means we
need an additional trigger to touch the crate whenever the
crates_keywords table is modified.


alexcrichton:
Oh man, I have so much sql-fu I need to learn. Thanks!

PR #272, opened 2016-02-25
Properly sideload keywords in crates#show

### LINKED ISSUE #271: Keywords went missing
Keywords are no longer displayed on a crates profile page.
Can't tell at what point they went missing.

sgrif:
The data is sent properly in this endpoint, but they aren't actually
associated with the crate as the ids aren't present. This appears to be
the only endpoint where these are sideloaded.

I've added a test for this, but haven't visually verified the fix (as I
don't have everything set up to run the backend locally), so this should
be tested on staging before deployment.

Fixes #271


alexcrichton:
Thanks!

PR #277, opened 2016-03-01
Update ember and remove an outdated dep

steveklabnik:
We recently upgraded ember-data, but not ember itself.

This also removes an unneeded package; it was there for pre-2.0
reasons, but we've upgraded past 2.0.

@wycats , this introduces warnings:

```
DEPRECATION: Support for the `ember-legacy-controllers` addon will end soon, please remove it from your application. [deprecation id: ember-legacy-controllers] See http://emberjs.com/deprecations/v1.x/#toc_objectcontroller for more details.
```

One for views as well. As far as I can tell, none of our code (nor our dependencies' code) uses the deprecated stuff; is this an unconditional warning?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "cargo",
   "dependencies": {
-    "ember": "2.3.1",
+    "ember": "2.4.0",
     "ember-cli-shims": "0.1.0",
     "ember-cli-test-loader": "0.2.1",
     "ember-load-initializers": "0.1.7",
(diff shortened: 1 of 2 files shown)


alexcrichton:
Oh these should be updated in lockstep? Good to know!

steveklabnik:
They don't _have_ to be, but you know, it's easiest to stay abreast of changes if we keep up. :)

stefanpenner:
they can be lockstep, but generally not required. Unless you use some new feature in ember-data that depends on a more recent version of ember. But that should be rare, especially if usage isn't changing but deps are merely being bumped

steveklabnik:
@stefanpenner any comment about the warnings?

stefanpenner:
https://github.com/emberjs/ember.js/pull/13030 might be related

steveklabnik:
Ahhhhhh yes that would make sense.

PR #280, opened 2016-03-03
disable autocorrect for the mobile search field

romankl:
Right now, the textfield tells the keyboard (e.g. on mobile Safari) to provide autocompletion/ autocorrect. Most keyboards/phones don't know words/ terms like "https" or "zlib" which causes an autocorrect for these words. 
This commit disables the autocompletion/autocorrect for the mobile search textfield.


alexcrichton:
Thanks!

PR #285, opened 2016-03-13
Show which dependencies are optional on the crate page

### LINKED ISSUE #233: Show which dependencies are optional
Show which dependencies are optional, so that it's easier to judge what a package depends on in its default configuration.

bluss:
Same idea & style as the yanked marker (but not red).

Fixes #233

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -146,6 +146,10 @@ span.small {
     }
 }
 
+.optional {
+    font-size: 80%;
+}
+
 .yanked {
     font-size: 80%;
     color: rgb(166, 0, 0)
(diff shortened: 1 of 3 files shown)


bluss:
I'm failing to install & run ember, I have no idea how any of this works. So this is so far entirely untested!

alexcrichton:
Hm looks like some extra css may be needed to get it to align right?



I like the idea though!

bluss:
Thanks for taking a look at it! It seems to me the text needs to be inside the `<li>` generated by `{{link-to-dep dep=dep}}`

alexcrichton:
Looks good to me!

bluss:
Yay!

Issue #281, opened 2016-03-04
Search Results Ordering


tadman:
Searching for the [bytes](https://crates.io/crates/bytes) package produces a list of results that's bizarre.

Some repositories with just a hundred or so downloads crowd out this package which has 33K. What order are these results presented in? I can't see a pattern here.

You'd think an exact match on name would push it to the top, but apparently that's not the case.

While I'm happy I found the [nom](https://crates.io/crates/nom) package as an interesting alternative, there's a bunch of other stuff in there which really shouldn't be at the top.

g-k:
I noticed this as well.

I had to go to the third page (almost the fourth) to find the official cargo crate: https://crates.io/search?page=3&q=cargo

carols10cents:
This looks old-- exact matches have been getting pulled to the top for a while now, so searching for "bytes" or "cargo" has those crates listed first. Please reopen or file a new issue if you have examples of other searches that don't return what you'd expect!

tadman:
Yeah, things look much better now. Thanks!

PR #286, opened 2016-03-14
Add weights to search query

jespino:
This give weight to the fields and I think must fix the #281 issue.

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -474,7 +474,7 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
         args.insert(0, query);
         ("SELECT crates.* FROM crates,
                                plainto_tsquery($1) q,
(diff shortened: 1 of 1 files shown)


alexcrichton:
I believe in theory this is what the [`setweight` calls](https://github.com/rust-lang/crates.io/blob/master/src/bin/migrate.rs#L339-L346) were trying to do in the migrations. Now knowing basically nothing about postgres full text search, maybe that's wrong? Seems though that we should rely on the weights that are indexed?

jespino:
Yes you are right, I have rereaded the postgresql documentation and the default values looks ok, but then I don't understand why is exact match name isn't the first result normally.

jespino:
There are cases where the full-text search have strange behavior, cargo-<anything> have the same score than cargo (I think we can add something like ORDER crates.name == q) to ensure that if any exact match happens have precedence over not exact matches.

PR #293, opened 2016-03-23
add autofocus attribute to search field

hobofan:
Most people come to the site to search for a package, so why not save them a click or pressing tab once? ;)

On browsers that don't support the autofocus attribute it should just be ignored.


alexcrichton:
Thanks!

PR #295, opened 2016-03-25
Add a link to reverse dependencies for crates.

### LINKED ISSUE #99: Show reverse dependencies

tari:
Fixes #99.

I would have liked to include a number for how many total dependents there are, but couldn't figure out how to get the actual number. Taking `reverse_dependencies.length` is limited by pagination so tops out at 10.


alexcrichton:
Thanks!

winding-lines:
nice :-)

PR #296, opened 2016-03-26
Add a backlink to newest for yanked crate versions

### LINKED ISSUE #148: Version page for yanked crates should link to non-yanked versions
e.g. https://crates.io/crates/order-stat/0.1.1 could have a link back to https://crates.io/crates/order-stat

tari:
Fixes #148.

Both paragraphs need to be wrapped in another `div` under `crate-info` since the stylesheet for `crate-info` makes a column-based layout which is not desirable for this case.

### DIFF
--- a/app/templates/crate/version.hbs
+++ b/app/templates/crate/version.hbs
@@ -37,8 +37,13 @@
 
 {{#if currentVersion.yanked}}
     <div class='crate-info'>
-        This crate has been yanked, but it is still available for download for
-        other crates that may be depending on it.
+        <div>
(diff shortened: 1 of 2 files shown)


alexcrichton:
Perhaps this could be conditional to ensure that you don't back-link to the same page? (In the case that the newest version is yanked as well)

tari:
I changed it to link to all versions, which is still friendly for the case that the newest version has been yanked. Also fixed a bug from 73c2e12b where the crate versions list would never indicate a version had been yanked.

alexcrichton:
Thanks!

PR #316, opened 2016-04-17
Doc dropdown menu: name "Cargo.toml" explicitly

birkenfeld:
This is what most people are looking for.  Many might not have heard
of the term "manifest".

See also https://github.com/rust-lang/cargo/pull/2587 for the same in the cargo docs.


alexcrichton:
Thanks!

Issue #319, opened 2016-04-22
Retina Icons


octplane:
Hello,

Because I wanted to redesign the search result page a bit (I find that items take a lot of place and that the fact we can't sort by download/... is annoying), I realised most icons are not ready for High density displays. Would you be interested if I did a pass on these ?

I'm very crappy at design in general, but using Sketch, I was quickly able to do this (left is original, right is retina)
<img width="192" alt="screen shot 2016-04-22 at 13 49 20" src="https://cloud.githubusercontent.com/assets/22177/14740688/1c4a6f12-0891-11e6-90a6-a1407c4b136f.png">

I've only worked to design 2x icon. But the main work is to redo the icons using Sketch, so in vector format so that when anybody will want 3x (or 4x, who knows?), it's easy to generate.

Let me know if this seems interesting for the project.

alexcrichton:
What nice! I'd love to add those :)

alexcrichton:
Fixed in https://github.com/rust-lang/crates.io/pull/322

PR #322, opened 2016-04-25
Retina Icons

octplane:
- all retina icons in public/assets
- added source in assets/: Sketch file and checker source
- fixed CSS accordingly


octplane:
Oh, yeah, this is the fix for #319, but I should have mentionned that somewhere in the commit msg... 🔥

alexcrichton:
Nice! Thanks!

alexcrichton:
Ok, playing around locally, it looks like there's only two differences that stick out. The lock icon if you're logged out in the top right is a brighter gray and the "crate" icon (purple box in a white circle) has a much larger purple box than before. Do you know if those could get tweaked?

octplane:
Tweaked in #323 
I can probably work on more details if you find other dubious stuff.

PR #323, opened 2016-04-25
Fix lock, package and search icons

octplane:
- lock was transparent, cargo(!) culted until it looked ok
- crate icon was very much to large
- search icon too thin


alexcrichton:
Thanks!

PR #349, opened 2016-06-04
Add "Specifying Dependencies" in Docs drop down

kennytm:
This PR synchronizes the Docs drop down from doc.crates.io to crates.io:

* Added the missing "Specifying Dependencies" link 
* "Using crates.io" has been renamed to "Publishing on crates.io"

(This list should be converted to a shared template in future, I think.)


kennytm:
The Travis failure is because it is using 0.8.2 instead of 0.8.3, someone should update the [Cargo.lock](https://github.com/rust-lang/crates.io/blob/79a672b654915e8387be6f65b3603703dde07cfc/Cargo.lock#L6).

alexcrichton:
Thanks!

PR #357, opened 2016-06-14
Annotate next/prev page links.

Stebalien:
The titles help blind users and the rel attributes help search engines (in
theory at least).

These also allow users of extensions like vimperator to navigate forwards and
backwards using hotkeys (`[[` and `]]`).


alexcrichton:
Thanks!

PR #363, opened 2016-06-17
Greenkeeper ember suave 3.0.1

steveklabnik:
So, mostly what was blocking this update was new defaults in our linter tool. I turned them all back to false, and fixed some indentation.

@alexcrichton @wycats do you have opinions about these particular settings?


alexcrichton:
Sounds good to me!

PR #365, opened 2016-06-20
Improve example .env documentation

carols10cents:
Hi! I hit a few bumps along my path to getting everything set up and the rust tests passing, so I made these changes to make it a little easier for the next person ❤️ 

- Instead of recommending exporting a bunch of env vars, point out the
  convenient `.env.sample` file that can be customized into your own,
  personalized `.env` file.
- Reorganize `.env.sample` to have the vars you'd most likely need to
  change first.
- Move most of the documentation about the vars into `.env.sample` for
  easy reference while editing it.
- Document the need to have the s3 bucket set to `alexcrichton-test`
  when running `cargo test` in order to get them to pass.

I think it'd be nice to not have to set the s3 bucket to `alexcrichton-test` to get the tests to pass, but I see a bunch of different solutions possible and thought I'd just start with documenting the current state. If you'd like, I'd be happy to, in this PR or a separate one, do one of the following, or some other solution:

- Set the `S3_BUCKET` env var to `alexcrichton-test` in some test setup code to overwrite whatever it is or isn't set to in the environment running the tests
- Remove `alexcrichton-test` from the URLs saved in the `tests/http-data` files, effectively requiring that the S3_BUCKET env var be set to empty string for the tests to pass
- Create a `TEST_S3_BUCKET` env var that works similarly to the `TEST_DATABASE_URL` env var
- Extend the functionality of the test request/response recorder to be able to substitute in variables [like vcr can](https://relishapp.com/vcr/vcr/v/3-0-1/docs/cassettes/dynamic-erb-cassettes) and use whatever the `S3_BUCKET` env var is set or not set to when matching the requests up
- ❓ ❓ ❓

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,25 +1,34 @@
-# Credentials for uploading packages to S3, these can be blank if you're not
-# publishing locally.
-S3_BUCKET=
-S3_ACCESS_KEY=
-S3_SECRET_KEY=
-# Not needed if the S3 bucket is in US standard
-S3_REGION=      
+# Location of the *postgres* database. For example, if you have created a
+# blank database locally named `cargo_registry`, this would be
+# `postgres://postgres:@localhost/cargo_registry`.
+export DATABASE_URL=""
 
-# Credentials for talking to github, can be blank if you're not logging in.
-#
-# When registering a new application, be sure to set the callback url to the
-# address `http://localhost:4200/authorize/github`.
-GH_CLIENT_ID=
-GH_CLIENT_SECRET=
+# Key to sign and encrypt cookies with. Change this to a long, random string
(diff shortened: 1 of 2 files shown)


carols10cents:
Whoops found one more thing-- that you need to have phantomjs installed to run the ember tests :) I'm all up and running now!!

alexcrichton:
Awesome, thanks!

Yeah I'm not sure what the best thing to do is about the `alexcrichton-test` bucket name. Right now those data files are generated from a live recording of an interaction with S3, so I'm not sure what the best way would be to edit them after-the-fact. Basically any way to test S3 requests would also suffice, although I couldn't think of any :(

PR #366, opened 2016-06-20
Upgrade openssl and its dependencies

carols10cents:
When running `cargo test` on OSX, I was getting assertion failures in openssl similar to [this issue](https://github.com/sfackler/rust-openssl/issues/303), and I tried all manner of setting openssl-related env vars, `cargo clean`ing, turning around three times and throwing salt over my shoulder, etc, recommended in that issue and others, but the only thing that solved the problem was upgrading the openssl crate. I have no idea which change in openssl solved my issue, but it's gone now!


alexcrichton:
Thanks!

PR #375, opened 2016-06-23
Upgrade

steveklabnik:
This does various upgrades to various things. Removes a bunch of warnings.


carols10cents:
I bet you're seeing a deprecation warning with ember-route-alias too-- I submitted https://github.com/nathanhammond/ember-route-alias/pull/10 and it got merged the other day but it doesn't look like @nathanhammond has cut a new version yet :)

alexcrichton:
Thanks!

PR #376, opened 2016-06-23
Serve git index in production

### LINKED ISSUE #373: crates.io isn't serving the index; cargo gets it from github.com
I'm trying to make it easier for people to set up mirrors of crates.io-- right now I'm building off of [this bare minimum mirror](https://gmjosack.github.io/posts/dissecting-cratesio-minimum-mirror/) blog post, which goes over how to set up a *local* index mirror of https://github.com/rust-lang/crates.io-index.

But what if I wanted to run a read-only mirror on heroku? I know crates.io runs on heroku right now, but I give up-- I cannot for the life of me figure out how crates.io has a checkout of the index!!!

I see the [`scripts/init-local-index.sh`](https://github.com/rust-lang/crates.io/blob/master/script/init-local-index.sh) script that appears to just be for working on the backend of crates.io locally; it was created way later than crates.io existed.

I see [this code at the start of `src/bin/server.rs`](https://github.com/rust-lang/crates.io/blob/master/src/bin/server.rs#L23-L36) that looks like if the checkout dir isn't a git repository, delete that dir, recreate it, and fetch the `GIT_REPO_URL` into that directory.

But I don't see how either of these two could work on heroku, since it's an [ephemeral file system](https://devcenter.heroku.com/articles/dynos#ephemeral-filesystem) and you can't count on files you've written being there later.

When I try to use, say, `tmp/index-co` as my `GIT_REPO_CHECKOUT` value (with `GIT_REPO_URL=https://github.com/rust-lang/crates.io-index.git`), I get a 404 from my mirror app when I point cargo at my mirror and cargo tries to update the registry.

So how do I get a registry available on heroku??? What am I missing? Thank you so much <3

carols10cents:
This is a way to fix #373 for running mirrors or private crate servers.

### DIFF
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -29,9 +29,12 @@ fn main() {
             cb.credentials(cargo_registry::git::credentials);
             let mut opts = git2::FetchOptions::new();
             opts.remote_callbacks(cb);
-            git2::build::RepoBuilder::new()
+            let repo = git2::build::RepoBuilder::new()
                                      .fetch_options(opts)
-                                     .clone(&url, &checkout).unwrap()
+                                     .clone(&url, &checkout).unwrap();
+            let git_daemon_file = checkout.join(".git").join("git-daemon-export-ok");
+            File::create(git_daemon_file).unwrap();
+            repo
         }
     };
     let mut cfg = repo.config().unwrap();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -120,15 +120,13 @@ pub fn middleware(app: Arc<App>) -> MiddlewareBuilder {
     router.get("/me/updates", C(user::updates));
     router.get("/summary", C(krate::summary));
 
-    let env = app.config.env;
-    if env == Env::Development {
-        let s = conduit_git_http_backend::Serve(app.git_repo_checkout.clone());
-        let s = Arc::new(s);
-        router.get("/git/index/*path", R(s.clone()));
-        router.post("/git/index/*path", R(s));
-    }
+    let s = conduit_git_http_backend::Serve(app.git_repo_checkout.clone());
+    let s = Arc::new(s);
+    router.get("/git/index/*path", R(s.clone()));
+    router.post("/git/index/*path", R(s));
 
     let mut m = MiddlewareBuilder::new(R404(router));
+    let env = app.config.env;
     if env == Env::Development {
         m.add(DebugMiddleware);
     }


steveklabnik:
I am a big :+1: for this in general, as it removes a dependency on github: when github is down, we're still up.

alexcrichton:
I'm somewhat hesitant to do this as it can create inconsistent copies of the index. Github right now is our source of truth, which all instances of the crates.io server point to. There can be a window where each crates.io server (dyno) have disagreeing copies of the index. The best fix for github being down I think is to just have mirrors (and support in Cargo as well).

For running mirrors, though, if they're guaranteed to only have one server then this definitely seems good to me! I'm not really sure how security-proof the git-daemon program is, but if it's just a mirror it isn't so bad.

carols10cents:
> I am a big 👍 for this in general, as it removes a dependency on github: when github is down, we're still up.

Note that today, since cargo defaults to github.com, people would have to change their `.cargo/config` to have `index = "https://crates.io/git/index"` in order to use this. After this is deployed, we could change cargo to have that as the default index URL, though.

> I'm somewhat hesitant to do this as it can create inconsistent copies of the index. Github right now is our source of truth, which all instances of the crates.io server point to. There can be a window where each crates.io server (dyno) have disagreeing copies of the index. The best fix for github being down I think is to just have mirrors (and support in Cargo as well).

What if this route could redirect to an externally-hosted git repo specified by an env var, kind of like the download links redirect to AWS?

As far as the inconsistency goes, aren't there currently inconsistency problems when publishing a crate while waiting for the `git push` to retry since we've added the crate+version to the database, which drives crates.io's UI? I thought I had seen issues related to this but of course can't find them now. I have no idea which inconsistency would happen more often in practice.

> For running mirrors, though, if they're guaranteed to only have one server then this definitely seems good to me! 

Hmmmm I don't know if we can guarantee that! We could at least to start with!

> I'm not really sure how security-proof the git-daemon program is, but if it's just a mirror it isn't so bad.

I don't know either, in actuality, but [its intention is to be a read-only service](https://www.kernel.org/pub/software/scm/git/docs/git-daemon.html).

carols10cents:
> I'm somewhat hesitant to do this as it can create inconsistent copies of the index. Github right now is our source of truth, which all instances of the crates.io server point to. There can be a window where each crates.io server (dyno) have disagreeing copies of the index. The best fix for github being down I think is to just have mirrors (and support in Cargo as well).

Another thing I've been thinking about is keeping up with a remote git index as a mirror-- what if each dyno did a `git pull` of the central repo that they all talk to every x minutes?

alexcrichton:
> What if this route could redirect to an externally-hosted git repo specified by an env var, kind of like the download links redirect to AWS?

Does git support redirects like HTTP? If HTTP didn't we'd just route traffic from S3 through the crates.io server, but that'd lose a theoretical locality property I think.

> As far as the inconsistency goes, aren't there currently inconsistency problems when publishing a crate while waiting for the git push to retry since we've added the crate+version to the database, which drives crates.io's UI?

In theory a crate isn't actually published until it's uploaded to S3 and pushed to github. If anything fails along the way the transaction should get rolled back and nothing should enter the database. I've seen a few bugs though, so this may not behave as I expect.

---

Perhaps a config option for this would suffice? It may not be appropriate for crates.io but it may certainly be appropriate for mirrors!

carols10cents:
> Perhaps a config option for this would suffice? It may not be appropriate for crates.io but it may certainly be appropriate for mirrors!

That sounds like a good plan, I'll work up a slightly different pull request :) Thank you for the feedback!!!

PR #385, opened 2016-07-15
Update license-expr dependency.

withoutboats:
The new version includes additional license recognized as valid by the SPDX spec.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ oauth2 = "0.1"
 log = "0.3"
 env_logger = "0.3"
 rustc-serialize = "0.3"
-license-exprs = "^1.1"
+license-exprs = "^1.3"
 dotenv = "0.8.0"
 
 conduit = "0.7"


withoutboats:
I'd also like to get feedback from the crates.io maintainers on withoutboats/license-exprs#6. This PR removes some licenses which have been deprecated, which is a breaking change. Fortunately, a licensing survey showed that there are no crates on crates.io using any of the deprecated licenses. Would crates.io be willing to upgrade to a list that excluded deprecated licenses?

steveklabnik:
I would be okay with doing that.

alexcrichton:
Sounds good to me as well! Thanks @withoutboats!

withoutboats:
Cool. I'll accept the other PR to license-expr as well, push up a 2.0.0 version of the crate, and open another PR to upgrade crates.io to that version.

alexcrichton:
Sounds good to me, thanks!

PR #387, opened 2016-07-17
Remove obsolete user_id column on crates table

carols10cents:
user_id has been replaced with a join to the crate_owners table since
multiple owners was implemented, but the user_id column on the crates
table still had a non-null foreign key constraint on it.

Discovered during the course of working on ✨secret things✨ ; let me know if there's a use for this that I'm not seeing!


alexcrichton:
I *think* this was once used for a query somewhere, but hey if all tests pass I think everything's covered pretty well so looks good to me, thanks!

PR #391, opened 2016-07-28
Implemented 'sort by relevance/downloads' for search results

lazypenguin:
I have updated the crates API endpoint to be slightly more robust to be able to accept both search terms and sorting parameters. I have also updated the front-end for the search results to mimic the "sort by" functionality found on other pages. 

In my opinion I believe this also addresses issue #281


alexcrichton:
Awesome, thanks! Looks like the JS test failure on travis may be relevant thought? Otherwise looks great to me!

alexcrichton:
Thanks!

PR #398, opened 2016-08-04
Fix sort order in crates api

### LINKED ISSUE #396: Sort by downloads doesn't
Right now, <https://crates.io/search?q=mmap&sort=downloads> looks like this to me:

no-crates-io-thats-not-how-sorting-works

Either this is wrong, or I've taken a serious blow to the head and can't ordering anymore.

lazypenguin:
Fixes ordering error in SQL query in crates API. Closes #396.


alexcrichton:
Thanks! Could you also add a test for this while you're at it?

alexcrichton:
Thanks!

PR #415, opened 2016-08-27
Show the word "rust" on crate detail page

### LINKED ISSUE #369: Rust is not mentioned on the crates pages
That hurts the searchability: if you [google for something like "rust tempfile"](https://www.google.com/search?q=rust+tempfile), it does not show the crates.io page.

wlonk:
This is a crude jamming in of the word, but maybe it'll suffice, or at least start a conversation about what would be better.

Fixes #369.

_Every pull request should come with a cute animal picture:_

seal

### DIFF
--- a/app/routes/crate.js
+++ b/app/routes/crate.js
@@ -7,5 +7,22 @@ export default Ember.Route.extend({
                 return this.controllerFor('application').set('flashError', `Crate '${params.crate_id}' does not exist`);
             }
         });
+    },
+
+    afterModel(model) {
+        this.setHeadTags(model);
+    },
+
+    setHeadTags(model) {
+        var headTags = [{
+            type: 'meta',
+            tagId: 'meta-description-tag',
+            attrs: {
+                name: 'description',
+                content: model.get('description') || 'A package for Rust.'
+            }
+        }];
+
+        this.set('headTags', headTags);
     }
 });
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -9,6 +9,7 @@ export default Ember.Route.extend({
             content: 'cargo is the package manager and crate host for rust'
         }
     }],
+
     model() {
         function addCrates(store, crates) {
             for (var i = 0; i < crates.length; i++) {
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -288,6 +288,13 @@ ul.dropdown {
     h1 { font-size: 1.5em; }
 }
 
+h1 {
+  span.subtitle {
+    display: block;
+    font-size: small;
+  }
+}
+
 #p404 {
     font-size: 120%;
     line-height: 3em;
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -19,6 +19,7 @@
         @include flex-direction(row-reverse);
         @include align-items(center);
     }
+    h3 { font-style: italic; }
     @media only screen and (max-width: 650px) {
         .right { display: none; }
     }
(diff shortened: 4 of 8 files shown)


alexcrichton:
Thanks! Could you show a screenshot of the change as well?

wlonk:
Here it is:

screenshot 2016-08-27 23 51 42

alexcrichton:
cc @rust-lang/docs 

Thoughts on this?

GuillaumeGomez:
Don't find it very useful... Otherwise, don't have a strong opinion on this change.

wlonk:
@GuillaumeGomez: It was intended to address the issue raised in #369, which I believe was for SEO purposes?

steveklabnik:
I feel torn. I agree we can do better on SEO here, but I'm not sure if this is the best way to do it, or if it does it at all. Do we have a citation for this being a good idea, other than "Putting "rust" on the page is likely to make it more relevant"?

wlonk:
@steveklabnik I have _no confidence_ that this actually helps SEO, but I wanted to get that ticket moving, to either a fixed or wontfix state. I am very very open to better suggestions.

GuillaumeGomez:
To add "rust" in the searches, you should better put it `<title>` or keywords add a title under Cargo which says something like "All rust crates in one place" (just an idea).

peschkaj:
For this to help SEO, "Rust" should be in the title, a header element on the page, a description meta tag, and ideally somewhere in the body. Source: [Google's SEO optimization starter guide](https://static.googleusercontent.com/media/www.google.com/en//webmasters/docs/search-engine-optimization-starter-guide.pdf).

To get the most out of SEO, I think a plan of attack should be developed based on existing material (e.g. the Google SEO optimization guide).

wlonk:
@peschkaj Thanks, that's very helpful. I'll read that, and propose something.

wlonk:
OK, per that document, there are three particularly relevant principles:

*Create unique, accurate page titles*

I propose changing the stable part of the page titles from "Cargo" to "Cargo: packages for Rust" or "Cargo: Crates for Rust"

*Make use of the "description" meta tag*

I propose adding a description meta tag (perhaps using [ember-cli-meta-tags](https://github.com/ronco/ember-cli-meta-tags)) to crate detail pages that will include the title ("<Crate name> - Cargo: packages for Rust") and the crate description (or at least the beginning of it?)

*Use heading tags appropriately*

We could change the brand `h1`, which currently says simply "CARGO", to say "CARGO: Packages for Rust". That makes a big visible change, so I'd especially appreciate input on this idea.

I'd love thoughts on this!

steveklabnik:
I like 1 and 2, I just merged https://github.com/rust-lang/crates.io/pull/423 which helps with 1 a bit, but could probably be tightened up for SEO purposes.

3 seems like a big, user visible change, yeah. I wonder if through some CSS magic, it could go below, in smaller type? Hm.

peschkaj:
On 3, I bet we could through a `span` in there and work some typographical magic to do @steveklabnik's idea.

:+1:

rjgoldsborough:
Whoops. Didn't see had diverged to talk about meta descriptions too.

I started a PR for this over at #427.

I added the `ember-cli-meta-tags` package and setup a meta tag for the index route. It looks pretty easy to add more to any page we need.

wlonk:
@rjgoldsborough Ah, duplication, sorry! I just did the same stuff in this branch!

alexcrichton:
Oh oops, sorry @wlonk! I merged #427 but it looks like this has some more bits? Want to rebase this and keep this so we can merge?

wlonk:
Totally.

alexcrichton:
Thanks @wlonk!

Issue #417, opened 2016-08-29
Remove quotes from env file


rjgoldsborough:
References #411 

tl;dr

I was receiving errors about the postgres url format when running the migrate and server/node name missing when running the server.

Removing the quotes in .env file fixed this.
This seems like it *could* be zsh but not 100% sure of the root cause.

Update the comment in the .env.sample file to suggest removing quotes if receiving errors about
malformed postgres url or civet web servername/nodename.

Issue #411, opened 2016-08-25
thread 'main' panicked at 'called `Result::unwrap()` when running migrate command


rjgoldsborough:
Hi all,

I'm trying to compile the rust backend and get down to the migrate command.
```
./target/debug/migrate
```

but then I get the following error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ConnectParams(StringError("url: Invalid character in scheme."))', ../src/libcore/result.rs:789
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

I updated the DATABASE_URL var in .env but that was all I changed.
I followed the naming suggestion and even just named the db 'cargo_registry'.

I'm running rustup using nightly.

Thanks!

rjgoldsborough:
```
stack backtrace:
   1:        0x10f696bda - std::sys::backtrace::tracing::imp::write::h482d45d91246faa2
   2:        0x10f69d07f - std::panicking::default_hook::_{{closure}}::h89158f66286b674e
   3:        0x10f69c485 - std::panicking::default_hook::h9e30d428ee3b0c43
   4:        0x10f69ca96 - std::panicking::rust_panic_with_hook::h2224f33fb7bf2f4c
   5:        0x10f69c934 - std::panicking::begin_panic::hcb11a4dc6d779ae5
   6:        0x10f69c852 - std::panicking::begin_panic_fmt::h310416c62f3935b3
   7:        0x10f69c7b7 - rust_begin_unwind
   8:        0x10f6c7c20 - core::panicking::panic_fmt::hc5789f4e80194729
   9:        0x10ef709d2 - core::result::unwrap_failed::h451ad2af54f32921
  10:        0x10ef661a3 - _<core..result..Result<T, E>>::unwrap::ha9d67f84b238aecd
  11:        0x10ef7db00 - server::main::hdab9661dc694288c
  12:        0x10f69d63a - __rust_maybe_catch_panic
  13:        0x10f69bfc6 - std::rt::lang_start::h352a66f5026f54bd
  14:        0x10ef7ee89 - main
```

alexcrichton:
What's the `DATABASE_URL` you're using? For me it's:

```
export DATABASE_URL="postgres://postgres@localhost/cargo_registry"
```

rjgoldsborough:
Hmm, It looks like there is a slight typo in the comment
```
# Location of the *postgres* database. For example, if you have created a
# blank database locally named `cargo_registry`, this would be
[...]

PR #418, opened 2016-08-29
adding comment to suggest removing double quotes if receiving postgre…

rjgoldsborough:
…s url or server/node name errors

refs #411 #417


alexcrichton:
Thanks! I was actually thinking we could just remove all the quotes from this file? If they're not there they likely won't get added in accidentally?

rjgoldsborough:
Ah, that probably makes more sense. I may just close this PR then and send another.

PR #420, opened 2016-08-29
removing quotes from env sample file

### LINKED ISSUE #417: Remove quotes from env file

rjgoldsborough:
Fixes #417


alexcrichton:
Thanks!

PR #421, opened 2016-08-30
making ember return an error but not redirect if crate not found

rjgoldsborough:
refs #141 

First pass at making this show a better crate not found page.

I took inspiration from github here and show an error but don't change the page.

I would like to show a better title and remove the stack trace (unless people find that valuable there) but I'm new to ember and still feeling it out.

Also, I have only tested this against a local running backend because the heroku staging site was returning a 503.

screen shot 2016-08-29 at 6 34 37 pm


steveklabnik:
Hey @rjgoldsborough ! First of all, thanks for all the work you've been doing on crates.io lately, it's been great.

Code-wise, this looks great, but I'm not 100% sure that a stacktrace in an error page is a great idea. It might be fine? I also like that the stacktracke makes it look like a fullish page, and I'm not sure what I'd replace it with.

@alexcrichton do you have opinions?

alexcrichton:
Ah yeah I think the issue #141 is getting at is that we just shouldn't change the URL bar. I think the UI today is fine (a flash message indicating an error) we just need to not change the URL bar as well so it can be corrected.

rjgoldsborough:
ok, I think I tweaked ember to do that.

screen shot 2016-08-30 at 11 15 21 am

steveklabnik:
:shipit: 

Thanks so much!

PR #426, opened 2016-08-30
Massage the data going into download graph

### LINKED ISSUE #305: Reverse drawing order of download graph
I find the download graphs very noisy, mostly due to the fact that often (if a version has zero downloads) the line color does not match the fill color below.
I think it would be better to draw newest versions first instead of oldest first (in the z-order). This way the line color always matches the fill color below and the color of a version only appears if it was actually downloaded.

TyOverby:
Closes #305

Google graphs does not treat 0-values in a stacked area-graph very well.  The default behavior for something like the crates.io download graph is to draw many lines right over the top of eachother, meaning that a recently released version will have its lines drawn over the top of the the other versions.

This can be fixed by passing google graphs a dataset with nulls instead of zeros for areas where the line shouldn't be drawn.

This is an implementation of an algorithm that detects sequences of zeros and removes all the zeroes in the middle of the sequence, replacing them with null.

For example
```
[1, 2, 0, 0, 0, 0, 5, 10, 0, 0, 0, 3, 0, 0]
```
will become (spaces indicate null)
```
[1, 2, 0,  ,  , 0, 5, 10, 0,  , 0, 3, 0, 0]
```

Leaving the boundary 0s makes the graph look more continuous.

# BEFORE
screen shot 2016-08-30 at 1 06 17 pm

# AFTER
screen shot 2016-08-30 at 1 07 45 pm

### DIFF
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -47,7 +47,7 @@ export default Ember.Component.extend({
                 // the line altogether.
                 if (!on[k] && value === 0) {
                     data[i][k + 1] = null;
-                } 
+                }
                 // If we are off and the value is not zero, we
                 // need to turn back on.  (keep the value the same though)
                 else if (!on[k] && value !== 0) {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice!

Issue #425, opened 2016-08-30
site doesn't have a meta descriptions


rjgoldsborough:
This leads to google simply showing the first content it finds when searching.

We should add these https://support.google.com/webmasters/answer/35624?rd=1

![screen shot 2016-08-30 at 9 21 13 am](https://cloud.githubusercontent.com/assets/868959/18097611/5bfc90dc-6e93-11e6-99cb-3d22093666c1.png)

rjgoldsborough:
The site uses https://www.npmjs.com/package/ember-page-title for page titles.

https://www.npmjs.com/package/ember-cli-meta-tags looks like it does the same for meta tags

PR #427, opened 2016-08-30
adding ember-cli-meta-tags npm package to provide meta descriptions

rjgoldsborough:
refs #425 

first pass at adding meta descriptions

This PR adds the ember-cli-meta-tags module. It also adds a head template. This is needed when using multiple modules using ember-cli-head which we now are.

There is an entry for the index route to have a meta description tag at `app/routes/index.js`.
This can be added to any route by adding an object:
```
headTags: [{
    type: 'meta',
    attrs: {
        name: 'description',
        content: 'cargo is the package manager and crate host for rust'
    }
}],
```

this produces:
`<meta id="ember823" content="cargo is the package manager and crate host for rust" name="description" class="ember-view">`

Feedback and thoughts on other page descriptions most appreciated.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
Nice!

Issue #428, opened 2016-09-01
e.get is not a function


dtolnay:
https://crates.io/crates/bert

![selection_005](https://cloud.githubusercontent.com/assets/1940490/18186267/c005479e-7056-11e6-8c75-fdc2ed23ebb5.png)

The stacktrace is:

```
  e.default<.setHeadTags@https://crates.io/assets/cargo-c12b1b6520f9d4295083e8d167bc0625.js:1:29205
e.default<.afterModel@https://crates.io/assets/cargo-c12b1b6520f9d4295083e8d167bc0625.js:1:29077
p@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:18:15207
n.prototype.runSharedModelHook@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:17:26529
n.prototype.runAfterModelHook@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:17:26238
a/<@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:18:13872
g@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:18:18908
b@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:18:19003
v@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:18:18811
@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:13:16139
r.prototype.invoke@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:5:31351
r.prototype.flush@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:5:31885
n.prototype.flush@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:5:29941
a.prototype.end@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:5:23207
a.prototype.run@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:5:24343
a.prototype.join@https://crates.io/assets/vendor-211b6ccdc7882f9afdd3922ed76fbd35.js:5:24405
[...]

PR #429, opened 2016-09-02
adding check to make sure model.get is function

rjgoldsborough:
a fix for #428 

adds a check to make sure model.get is a function before trying to call it.


alexcrichton:
Thanks!

PR #440, opened 2016-09-10
Add a deploy to heroku button and instructions on running a mirror!

carols10cents:
I am able to install crates from a heroku instance created using this app.json, at https://cratesio-mirror.herokuapp.com/ !

I have an index mirror at https://gitlab.com/integer32llc/crates.io-index that I'm updating every 10 min. I tested out installing a crate through this index and my mirror server, updating that crate on crates.io, verifying that a `cargo update` does not see the update, waiting until my index fetches the changes, then getting the update with a `cargo update` run after that.

It's still probably not doing what people would expect when they hear "mirror", so I tried to add clear caveats about what this does and does not do right now.

Again, this is a baby step in a series of baby steps I hope to help with, but I think it's a step worth taking! 👣

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,9 @@ To deploy the API on Heroku, use this button:
 
 [deploy]: https://heroku.com/deploy
 
-You do not need to fill in any of the optional fields.
+The only config variable you need to set is `GIT_REPO_URL`, which should be the
+git URL of your crates index repository; see the next section for setup
+instructions for that.
 
 ### Index mirror setup
 
--- a/README.md
+++ b/README.md
@@ -196,4 +196,9 @@ registry = "https://[host and path to your git server]/crates.io-index"
 
 [source.crates-io]
 replace-with = "mirror"
+registry = 'https://doesnt-matter-but-must-be-present'
 ```
+
+Once [rust-lang/cargo#3089](https://github.com/rust-lang/cargo/pull/3089) is
+released, it won't be necessary to specify a registry URL for a source being
+replaced.
--- a/README.md
+++ b/README.md
@@ -122,7 +122,16 @@ For more information on using ember-cli, visit
 For more information on using cargo, visit
 [doc.crates.io](http://doc.crates.io/).
 
-## Deploy a read-only, download-API-only mirror
+## Deploying a Mirror
+
+**DISCLAIMER: The process of setting up a mirror is a work-in-progress and is
+likely to change. It is not currently recommended for mission-critical
+production use. It also requires a version of cargo newer than 0.13.0-nightly
+(f09ef68 2016-08-02); the version of cargo currently on rustc's beta channel
+fulfils this requirement and will be shipped with rustc 1.12.0 scheduled to be
+released on 2016-09-29.**
+
+### Current functionality: a read-only, download-API-only mirror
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -121,3 +121,62 @@ For more information on using ember-cli, visit
...
+    ```json
+    {
+      "dl": "https://[your heroku app name].herokuapp.com/api/v1/crates",
+      "api": "https://[your heroku app name].herokuapp.com/"
+    }
+    ```
+
+3. Commit and push to wherever you will be hosting your index (ex: github,
+    gitlab, an internal git server)
+4. In your cargo project, replace the crates.io source with your source by
```
carols10cents: Crap this is still in nightly cargo only, isn't it... I'll look up the versions and specify them.

alexcrichton:
Thanks @carols10cents! Excited to see progress on this! I'll leave a few comments but otherwise I'm fine merging.

README.md
alexcrichton: Ah yeah might be worth mentioning that as well as the dates it'll be released into stable Rust.

app.json
```diff
@@ -0,0 +1,54 @@
...
+      "GH_CLIENT_ID": {
+          "value": "",
+          "required": false
+      },
+      "GH_CLIENT_SECRET": {
+          "value": "",
+          "required": false
+      },
+      "GIT_REPO_CHECKOUT": "./tmp/index-co",
+      "GIT_REPO_URL": "https://github.com/rust-lang/crates.io-index.git",
```
alexcrichton: This in theory needs to be the mirror of the index, right?

app.json
```diff
@@ -0,0 +1,54 @@
...
+  },
+  "addons": [
+      "heroku-postgresql:hobby-dev"
+  ],
+  "buildpacks": [
+      {
+        "url": "https://github.com/rcaught/heroku-buildpack-cmake"
+      },
+      {
+        "url": "https://github.com/alexcrichton/heroku-buildpack-rust"
```
alexcrichton: I wonder, can revisions be specified here as well?

README.md
```diff
@@ -121,3 +121,62 @@ For more information on using ember-cli, visit
 
 For more information on using cargo, visit
 [doc.crates.io](http://doc.crates.io/).
+
+## Deploy a read-only, download-API-only mirror
```
alexcrichton: Could this section perhaps also mention that this is a work-in-progress? Ideally we'd make all deployment a one-click operation, but hopefully coming soon!

app.json
carols10cents: In theory, yes, and eventually it will need to be in practice, but for the moment it doesn't actually matter since client-side cargo is the only thing that uses the index. I'll update the instructions though so that people have it set for when we do start using it.

app.json
carols10cents: They can indeed! I will do so!

carols10cents:
@alexcrichton : I updated the instructions to address all your comments, thank you!! 🎉

alexcrichton:
Thanks @carols10cents!

carols10cents:
YAAAYYYY!!! :) :) <3 <3 <3

PR #453, opened 2016-10-06
breaking long words in the crates list

### LINKED ISSUE #451: Frontpage Columns Messed Up
On the front page of https://crates.io, the really long version number messes up the formatting: 

screenshot from 2016-10-03 20-46-34

This is on Linux (Ubuntu 16.04 x86-64) with both Chrome (Version 50.0.2661.86 (64-bit)) and Firefox (Mozilla Firefox 51.0a2)

rjgoldsborough:
fixes #451 

(I could no longer see that crate so I just made up a very long version number)

screen shot 2016-10-06 at 10 21 53 am

### DIFF
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -111,6 +111,12 @@
         @include justify-content(space-between);
         @include align-items(center);
 
+        span {
+            overflow: hidden;
+            text-overflow: ellipsis;
+            white-space: nowrap;
+        }
+
         img { vertical-align: middle; }
     }
     li a:hover { background-color: darken($main-bg-dark, 5%); }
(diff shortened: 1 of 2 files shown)


steveklabnik:
I like it. @carols10cents / @alexcrichton ?

alexcrichton:
Is there a CSS way to truncate a version? I'd personally prefer  something like `very-long-v...` so everything always fit on one line

rjgoldsborough:
Yeah, there is https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow which I tried first, but that needs to be on the anchor tag I believe and since it's a flex child, it doesn't have a sense of width and thus can't really know what is overflowing and what to clip. That's why I had to set that property up the tree.

I can keep banging on it if that's definitely preferred.

alexcrichton:
Yeah perhaps we can tweak the markup here to take advantage of that?

rjgoldsborough:
Sure, I'll try that

alexcrichton:
Thanks!

PR #459, opened 2016-10-18
adding docs.rs badge to crate sidebar

### LINKED ISSUE #419: docs.rs integration
**See [this comment](https://github.com/rust-lang/crates.io/issues/419#issuecomment-285415774) for detailed info on what we're currently thinking of for the implementation**

Recently, @onur released the awesome https://docs.rs/ , which autogenerates documentation from crates hosted on crates.io.

A lot of people have been asking how this might integrate with crates.io, so I wanted to open this thread to discuss possibilities.

One possibility I'd thought of: we could take any crate which doesn't have a "documentation" link, and replace it with a link to the page on crates.rs. @onur also thought that maybe a badge would work.

/cc @rust-lang/core

ducks:
fixes #419 

Adding a docs.rs badge and link to the sidebar.

The initial issue said to replace the documentation link but I think that badge will look a bit funny down in a text list so I put it in the sidebar. It also sounds like enough people would be in favor of docs.rs by default and we also already have the crates.io badge up there.

screen shot 2016-10-17 at 9 32 06 pm

### DIFF
--- a/app/templates/crate/version.hbs
+++ b/app/templates/crate/version.hbs
@@ -71,6 +71,14 @@
                         alt="{{ crate.name }}’s current version badge"
                         title="{{ crate.name }}’s current version badge" />
                 </p>
+                <p>
+                    <a href="https://docs.rs/{{ crate.name }}">
+                        <img
+                            src="https://docs.rs/{{ crate.name }}/badge.svg"
+                            alt="{{ crate.name }}'s  https://docs.rs badge"
+                            title="{{ crate.name }}'s  https://docs.rs badge" />
+                    </a>
+                </p>
             </div>
 
             <div class="authors">


steveklabnik:
I wonder what @briansmith thinks of this; he's one of the people who's very anti-docs.rs, at least at the moment.

The reason I bring it up is, the idea for putting it in the text wasn't that we'd put the badge in the text, it's that if there was no documentation link, we'd put the current text with the link. This would let people opt out, which isn't really possible with the badge here.

That said, if others are okay with it, I'm okay with it, I just want to make sure those concerns are addressed.

rjgoldsborough:
Ah, yeah, I had an unless in there before to only show it if there wasn't documentation but I thought the badge looked funky below, but having it in two places conditionally would also be jarring.

I didn't think to just use a docs rs link in the current docs link location because the badge looked slick. :)

By opt out do you mean simply not using docs.rs or something more?

onur:
Interesting, I'd like to know why @briansmith is against docs.rs

briansmith:
> I'd like to know why @briansmith is against docs.rs

I never said I was "against docs.rs". Clearly a lot of people like the docs.rs service and I think it's a good service, particularly for people who don't want to bother with hosting their documentation themselves. I'm also giving them the benefit of the doubt that they'll address the issues below.

What I said is I'd prefer to direct people to my own published copy of the documentation. Why does the Rust team publish the libstd documentation at https://doc.rust-lang.org instead of relying on docs.rs? I think the reasons that the Rust teams have for maintaining https://doc.rust-lang.org are likely the same reasons as some people have for wanting to publish their crates' documentation themselves.

Note that docs.rs doesn't show the documentation for my *ring* crate correctly; in particular, it is missing the `RSAKeyPair` documentation (compare https://docs.rs/ring/0.4.3/ring/signature/index.html to https://briansmith.org/rustdoc/ring/signature/index.html). See https://github.com/onur/docs.rs/issues/29. It is disappointing that such basic issues haven't been resolved even though it was reported right away. Also, docs.rs isn't following the terms of my crates' licenses, which you can read in the LICENSE file of each repo.

steveklabnik:
> Interesting, I'd like to know why @briansmith is against docs.rs

I tried to phrase this as something that _didn't_ imply @briansmith was against it in general, only that it isn't perfect and the way in which it isn't perfect is important to some people. I don't mean to put words in his mouth here, sorry if I was misleading in my wording.

onur:
@briansmith, thanks for feedback. I thought you are against docs.rs when Steve referred you as "very anti-docs.rs". My bad sorry.

Actually first thing Rust team asked me to host std docs when I first released docs.rs. Few other people also [asked for it](onur/docs.rs#56).

You are right about onur/docs.rs#29 but I've been extremely busy since I released docs.rs and didn't have much time to work on it. I am back on project and hopefully will fix this issue really soon.

briansmith:
> @briansmith, thanks for feedback. I thought you are against docs.rs when Steve referred you as "very anti-docs.rs". My bad sorry.

I'm definitely not against docs.rs.

> Actually first thing Rust team asked me to host std docs when I first released docs.rs. Few other people also asked for it.

Probably this is a misunderstanding and I think it doesn't matter.

> You are right about onur/docs.rs#29 but I've been extremely busy since I released docs.rs and didn't have much time to work on it. I am back on project and hopefully will fix this issue really soon.

That would be great! I also commented on the license issue in your repo. Anyway, I don't want to distract from the actual review of the PR.

alexcrichton:
Thanks for the PR @rjgoldsborough! In terms of the implementation I think we may also want to link to the right version of the documentation, so perhaps `version` can be included as well in the URL? I also wonder if we may want tighter integration. For example this could be a dead link for a number of crates if documentation failed to build for one reason or another.

From the "if we'd like to have this" point of view I'm personally in favor but like @steveklabnik am also curious of others' opinions. @briansmith I'm curious, if this were added to crates.io would you prefer to opt-out of this link for your crates?

briansmith:
> In terms of the implementation I think we may also want to link to the right version of the documentation, so perhaps version can be included as well in the URL?

I think it's better to link to the latest version like the "Documentation" link does in most cases.

> @briansmith I'm curious, if this were added to crates.io would you prefer to opt-out of this link for your crates?

The license for the documentation for my stuff allows people to republish it. I would prefer that docs.rs republish things in a better way and I'll work with them in their issue tracker regarding those improvements. However, I'm not sure about the next PR adding a different badge, and the one after that, and the ones after those.

BTW, I'm not sure if you fully understand the privacy and business implications of these badges. If you add the badge on every crate then the service hosting the images in the `` will be able to track people's browsing on the crates.io website. This is one reason why companies are so eager to provide badge-based services for GitHub. An obvious privacy improvement would be to inline the SVG into the HTML instead of using ``. Otherwise, I recommend talking with others at Mozilla about the agreements they have with vendors w.r.t. tracking beacons like this.

onur:
BTW I want to mention onur/docs.rs#29 is getting fixed [soon](https://github.com/onur/docs.rs/pull/73).

And TBH I also think it is still early to add a docs.rs link to crates.io. I agree with @steveklabnik docs.rs is ok but far from perfect, there are some flaws, especially cross crate links don't work. Builds with non-default features are getting fixed soon.

@alexcrichton is also right. Docs.rs currently cannot handle dead links properly, [clippy](https://docs.rs/clippy) for example, it's not a useful error message and might be really confusing for newcomers.

Many people started using docs.rs for their documentation links and docs.rs badges in their README's. IMO lets leave it like that, crate owners can decide if they want to use docs.rs or not for their documentation source.

And I will continue developing docs.rs and fix some serious flaws, like cross crate links, improved error handling and more automated way to build/publish docs.

---

For the privacy, I can assure you docs.rs is not tracking anything. Docs.rs doesn't even have any goggle-analytics code unlike crates.io. I also deeply care about privacy and it's not none of my business who is visiting what crate or website.

Anyway, I agree with @briansmith about badges. And depending on a third party service is not a good idea. If we really want to show badges in crates.io we can use my [badge](https://crates.io/crates/badge) crate. It's been working fine on docs.rs for a while and much better than depending on a third party service.

briansmith:
For reference, here are the four issues I think are important, as a crate author and privacy advocate:
* https://github.com/onur/docs.rs/issues/75: Document advertising and tracking policy.
* https://github.com/onur/docs.rs/issues/74: `rel=canonical` support.
* https://github.com/onur/docs.rs/issues/72: Make crate license easier to find.
* https://github.com/onur/docs.rs/issues/29: Ensure documentation is generated as intended by crate authors.

BTW, as far as I can tell, they seem like 100% upstanding people who are just trying to provide a useful service. But you can imagine that somebody else could easily fork their code, wrap all the content in advertising and trackers on a new website, and then ask for equal treatment with their own tracker-ific badge. This is more what I'm trying to discourage.

alexcrichton:
I personally think that at least one major blocker for this will be to *not* generate a link for crates which have documentation that failed to build on docs.rs. That may unfortunately provide a pretty sub-par experience :(

Cobrand:
I'm a little late to the party, but I think that while docs.rs has major flaws that prevent it from being linked directly from crates.io for every crate that don't have a documentation link specified, it is still important to link to docs.rs in one way or another. As of right now I don't think there is a single mention of docs.rs anywhere on the crates.io website, and newcomers that don't know about docs.rs will just give up and try another crate (if there is any equivalent), or try to look at the repository and _maybe_ find a documentation link, but sometimes they might not.

What we could do instead is create a link "Documentation link missing" right where the "Documentation" link usually is, and create a new page on crates.io named "Help! There is no documentation for my crate !", and explain on this page that there is a nice website called docs.rs that hosts documentation for every crate, but that this website is in no way officially supported and some crates might be missing or even incorrect have documentation some rare cases.
For now docs.rs is the only candidate, but in the future there might be another website with similar-but-better features for generating docs, and everyone will be using it. At that point, we could add to the "Help! There is no documentation for my crate!" page a link to the new rust documentation website.

I'm just saying this cause experienced users know about docs.rs, but new users might not know at all about this website. If we could give newcomers a hand to find proper documentation for every crate, I think that would be a huge plus, even if it's as small as saying "we don't officially support this but there is this website hosting docs if you want"

carols10cents:
See #506-- docs.rs now provides a way that crates.io can query for build status, so I'm closing this PR in anticipation that someone will implement a use of that information that only shows a docs.rs link on crates.io if the latest build has succeeded.

PR #472, opened 2016-11-17
Update index creation message

### LINKED ISSUE #471: Instructions provided by the init-local-index.sh should use replace-with
Since `replace-with` has now landed and `registry.index` is being deprecated, [these instructions echoed by the script that sets up a local index for development](https://github.com/rust-lang/crates.io/blob/master/script/init-local-index.sh#L35-L36) should instead instruct you to use `replace-with` (as [the README does](https://github.com/rust-lang/crates.io/blob/master/README.md#local-cargo-setup)) instead of `registry.index`.

kelcecil:
This change rewrites the message at the end of the script to be more consistent with the README file. The README says to run the script but not to modify the config file until after running. The ending message states these instructions, uses the replace-with key, and provides a URL to the README for convenience. Fixes #471.

### DIFF
--- a/script/init-local-index.sh
+++ b/script/init-local-index.sh
@@ -30,10 +30,24 @@ cd ../..
 touch tmp/index-co/.git/git-daemon-export-ok
 
 cat - <<-EOF
-Please add the following to your HOME/.cargo/config:
+Your local git index is ready to go!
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #473, opened 2016-11-17
Add Categorization!

carols10cents:
Hiiii! This is the first part of adding categories to crates.io!

## What this PR does

- On server startup, take the list of categories in `src/categories.txt` and add or remove categories from the database as needed to make the database match the categories. This way, the list of available categories can be changed via pull request to this repo. 
- A `/categories` page not yet linked from anywhere that displays all the categories in alphabetical order
- A `/categories/whatever` page that will list all the crates in a category
- If a crate has categories, they will be listed on that crate's page in the sidebar under Keywords
- If categories are specified in the metadata uploaded with a new crate request from cargo, they will be added to the crate. (Implemented in rust-lang/cargo#3301)

## Testing this PR

```
$ cargo build
$ cargo run --bin migrate
$ cargo run --bin sync-categories
$ cargo run --bin server
```

Start the frontend as well (Note: I had to use `ember server --proxy http://127.0.0.1:8888` instead of `yarn run start:local` with the latest version of yarn in order to actually use my local backend, I'm investigating why this is the case)

- You should be able to go to `/categories` and see 2 categories, "Development Tools" and "Libraries"
- You should be able to click on those categories and see 0 crates. They each have subcategories, that should also have 0 crates.
- Publish a crate to your local crates.io (I had to comment out code uploading to s3 since I didn't want to do that)
- Using rust-lang/cargo#3301, publish to your local index
- You should see your crate on that category's page
- You should see the category on the crate's page (might have to do a shift-reload because of ember caching)

## Deploying this PR 

It should be totally fine to deploy this PR to crates.io. Nothing links to /categories, and cargo does not publish category metadata yet-- and neither cargo nor crates.io will complain if a crate is not in any categories.

## After this PR is merged

- I will start a PR to categories.txt where we can bikeshed what we actually want the initial set of categories to be
- I will open a PR to add a link to /categories somewhere 
- I will send some popular crates PRs to add categories

### DIFF
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -764,6 +764,61 @@ fn migrations() -> Vec<Migration> {
             try!(tx.execute("DROP INDEX users_gh_id", &[]));
             Ok(())
         }),
+        Migration::add_table(20161115110541, "categories", " \
+            id               SERIAL PRIMARY KEY, \
+            category         VARCHAR NOT NULL UNIQUE, \
+            crates_cnt       INTEGER NOT NULL DEFAULT 0, \
+            created_at       TIMESTAMP NOT NULL DEFAULT current_timestamp"),
+        Migration::add_table(20161115111828, "crates_categories", " \
+            crate_id         INTEGER NOT NULL, \
+            category_id      INTEGER NOT NULL"),
+        foreign_key(20161115111836, "crates_categories", "crate_id", "crates (id)"),
+        Migration::run(20161115111846, " \
+            ALTER TABLE crates_categories \
+            ADD CONSTRAINT fk_crates_categories_category_id \
+            FOREIGN KEY (category_id) REFERENCES categories (id) \
+            ON DELETE CASCADE", " \
+            ALTER TABLE crates_categories \
+            DROP CONSTRAINT fk_crates_categories_category_id"),
+        index(20161115111853, "crates_categories", "crate_id"),
+        index(20161115111900, "crates_categories", "category_id"),
+        Migration::new(20161115111957, |tx| {
+            try!(tx.batch_execute(" \
+                CREATE FUNCTION update_categories_crates_cnt() \
+                RETURNS trigger AS $$ \
+                BEGIN \
+                    IF (TG_OP = 'INSERT') THEN \
+                        UPDATE categories \
+                        SET crates_cnt = crates_cnt + 1 \
+                        WHERE id = NEW.category_id; \
+                        return NEW; \
+                    ELSIF (TG_OP = 'DELETE') THEN \
+                        UPDATE categories \
+                        SET crates_cnt = crates_cnt - 1 \
+                        WHERE id = OLD.category_id; \
+                        return OLD; \
+                    END IF; \
+                END \
+                $$ LANGUAGE plpgsql; \
+                CREATE TRIGGER trigger_update_categories_crates_cnt \
+                BEFORE INSERT OR DELETE \
+                ON crates_categories \
+                FOR EACH ROW EXECUTE PROCEDURE update_categories_crates_cnt(); \
+                CREATE TRIGGER touch_crate_on_modify_categories \
+                AFTER INSERT OR DELETE ON crates_categories \
+                FOR EACH ROW \
+                EXECUTE PROCEDURE touch_crate(); \
+            "));
+            Ok(())
+        }, |tx| {
+            try!(tx.batch_execute(" \
+                DROP TRIGGER trigger_update_categories_crates_cnt \
+                ON crates_categories; \
+                DROP FUNCTION update_categories_crates_cnt(); \
+                DROP TRIGGER touch_crate_on_modify_categories \
+                ON crates_categories;"));
+            Ok(())
+        }),
     ];
     // NOTE: Generate a new id via `date +"%Y%m%d%H%M%S"`
 
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -768,6 +768,7 @@ fn migrations() -> Vec<Migration> {
             id               SERIAL PRIMARY KEY, \
             category         VARCHAR NOT NULL UNIQUE, \
             slug             VARCHAR NOT NULL UNIQUE, \
+            description      VARCHAR NOT NULL DEFAULT '', \
             crates_cnt       INTEGER NOT NULL DEFAULT 0, \
             created_at       TIMESTAMP NOT NULL DEFAULT current_timestamp"),
         Migration::add_table(20161115111828, "crates_categories", " \
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -767,6 +767,7 @@ fn migrations() -> Vec<Migration> {
         Migration::add_table(20161115110541, "categories", " \
             id               SERIAL PRIMARY KEY, \
             category         VARCHAR NOT NULL UNIQUE, \
+            slug             VARCHAR NOT NULL UNIQUE, \
             crates_cnt       INTEGER NOT NULL DEFAULT 0, \
             created_at       TIMESTAMP NOT NULL DEFAULT current_timestamp"),
         Migration::add_table(20161115111828, "crates_categories", " \
--- a/src/bin/sync-categories.rs
+++ b/src/bin/sync-categories.rs
@@ -1,0 +1,48 @@
+// Sync available crate categories from `src/categories.txt`.
+// Only needs to be run for new databases or when `src/categories.txt`
+// has been updated.
+//
+// Usage:
+//      cargo run --bin sync-categories
+
+#![deny(warnings)]
+
+extern crate cargo_registry;
+extern crate postgres;
+
+use cargo_registry::env;
+
+fn main() {
+    let conn = postgres::Connection::connect(&env("DATABASE_URL")[..],
+                                             postgres::TlsMode::None).unwrap();
+    let tx = conn.transaction().unwrap();
+    sync(&tx).unwrap();
+    tx.set_commit();
+    tx.finish().unwrap();
+}
+
+fn sync(tx: &postgres::transaction::Transaction) -> postgres::Result<()> {
+    let categories = include_str!("../categories.txt");
+    let categories: Vec<_> = categories.lines().collect();
+    let insert = categories.iter()
+                           .map(|c| format!("('{}')", c))
+                           .collect::<Vec<_>>()
+                           .join(",");
+    let in_clause = categories.iter()
+                              .map(|c| format!("'{}'", c))
+                              .collect::<Vec<_>>()
+                              .join(",");
+
+    try!(tx.batch_execute(
+        &format!(" \
+            INSERT INTO categories (category) \
+            VALUES {} \
+            ON CONFLICT (category) DO NOTHING; \
+            DELETE FROM categories \
+            WHERE category NOT IN ({});",
+            insert,
+            in_clause
+        )[..]
+    ));
+    Ok(())
+}
--- a/src/bin/sync-categories.rs
+++ b/src/bin/sync-categories.rs
@@ -23,23 +23,29 @@ fn main() {
 
 fn sync(tx: &postgres::transaction::Transaction) -> postgres::Result<()> {
     let categories = include_str!("../categories.txt");
-    let categories: Vec<_> = categories.lines().collect();
-    let insert = categories.iter()
-                           .map(|c| format!("('{}')", c))
-                           .collect::<Vec<_>>()
-                           .join(",");
-    let in_clause = categories.iter()
-                              .map(|c| format!("'{}'", c))
-                              .collect::<Vec<_>>()
-                              .join(",");
+
+    let slug_categories: Vec<_> = categories.lines().map(|c| {
+        let mut parts = c.split(' ');
+        let slug = parts.next().expect("No slug found!");
+        let name = parts.collect::<Vec<_>>().join(" ");
+        (slug, name)
+    }).collect();
+
+    let insert = slug_categories.iter().map(|&(ref slug, ref name)| {
+        format!("(LOWER('{}'), '{}')", slug, name)
+    }).collect::<Vec<_>>().join(",");
+
+    let in_clause = slug_categories.iter().map(|&(slug, _)| {
+        format!("LOWER('{}')", slug)
+    }).collect::<Vec<_>>().join(",");
 
     try!(tx.batch_execute(
         &format!(" \
-            INSERT INTO categories (category) \
+            INSERT INTO categories (slug, category) \
             VALUES {} \
-            ON CONFLICT (category) DO NOTHING; \
+            ON CONFLICT (slug) DO UPDATE SET category = EXCLUDED.category; \
             DELETE FROM categories \
-            WHERE category NOT IN ({});",
+            WHERE slug NOT IN ({});",
             insert,
             in_clause
         )[..]
--- a/src/categories.rs
+++ b/src/categories.rs
@@ -7,71 +7,76 @@ use env;
 use util::errors::{CargoResult, ChainError, internal};
 
 struct Category {
-    name: String,
     slug: String,
+    name: String,
     description: String,
 }
 
 impl Category {
-    fn concat(&self, child: &Category) -> Category {
-        Category {
-            name: format!("{}::{}", self.name, child.name),
-            slug: format!("{}::{}", self.slug, child.slug),
-            description: child.description.clone(),
+    fn from_parent(slug: &str, name: &str, description: &str, parent: Option<&Category>)
+                   -> Category {
+        match parent {
+            Some(parent) => {
+                Category {
+                    slug: format!("{}::{}", parent.slug, slug),
+                    name: format!("{}::{}", parent.name, name),
+                    description: description.into(),
+                }
+            }
+            None => {
+                Category {
+                    slug: slug.into(),
+                    name: name.into(),
+                    description: description.into(),
+                }
+            }
         }
     }
 }
 
-fn concat_parent_and_child(parent: Option<&Category>, child: Category)
-                           -> Category {
-    parent.map(|p| p.concat(&child)).unwrap_or(child)
-}
-
-fn required_string_from_toml(toml: &toml::Table, key: &str)
-                             -> CargoResult<String> {
+fn required_string_from_toml<'a>(toml: &'a toml::Table, key: &str) -> CargoResult<&'a str> {
     toml.get(key)
         .and_then(toml::Value::as_str)
-        .map(str::to_string)
         .chain_error(|| {
-            internal("Expected Category toml attribute to be a String")
+            internal(format!("Expected category TOML attribute '{}' to be a String", key))
         })
 }
 
-fn optional_string_from_toml(toml: &toml::Table, key: &str)
-                             -> String {
+fn optional_string_from_toml<'a>(toml: &'a toml::Table, key: &str) -> &'a str {
     toml.get(key)
         .and_then(toml::Value::as_str)
         .unwrap_or("")
-        .to_string()
 }
 
-fn category_from_toml(toml: &toml::Value, parent: Option<&Category>)
-                      -> CargoResult<Vec<Category>> {
-    let toml = toml.as_table().chain_error(|| {
-        internal("Category isn't a toml Table")
-    })?;
+fn categories_from_toml(categories: &toml::Table, parent: Option<&Category>) -> CargoResult<Vec<Category>> {
+    let mut result = vec![];
 
-    let category = Category {
-        slug: required_string_from_toml(&toml, "slug")?,
-        name: required_string_from_toml(&toml, "name")?,
-        description: optional_string_from_toml(&toml, "description"),
-    };
+    for (slug, details) in categories {
+        let details = details.as_table().chain_error(|| {
+            internal(format!("category {} was not a TOML table", slug))
+        })?;
 
-    let category = concat_parent_and_child(parent, category);
+        let category = Category::from_parent(
+            slug,
+            required_string_from_toml(&details, "name")?,
+            optional_string_from_toml(&details, "description"),
+            parent,
+        );
 
-    let mut children: Vec<_> = toml.get("categories")
-        .and_then(toml::Value::as_slice)
-        .map(|children| {
-            children.iter()
-                .flat_map(|ref child| {
-                    category_from_toml(child, Some(&category))
-                        .expect("Could not create child from toml")
-                }).collect()
-        }).unwrap_or(Vec::new());
+        if let Some(categories) = details.get("categories") {
+            let categories = categories.as_table().chain_error(|| {
+                internal(format!("child categories of {} were not a table", slug))
+            })?;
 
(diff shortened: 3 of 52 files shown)


carols10cents:
Ummmmm I have no idea why the tests are failing with "Once instance has previously been poisoned" :(

carols10cents:
Aaaand I forgot to add support for subcategories, just realized that. I'd love any thoughts in the meantime, but I'm going to be adding a few commits :)

src/bin/migrate.rs
```diff
@@ -764,6 +764,61 @@ fn migrations() -> Vec<Migration> {
...
+            created_at       TIMESTAMP NOT NULL DEFAULT current_timestamp"),
+        Migration::add_table(20161115111828, "crates_categories", " \
+            crate_id         INTEGER NOT NULL, \
+            category_id      INTEGER NOT NULL"),
+        foreign_key(20161115111836, "crates_categories", "crate_id", "crates (id)"),
+        Migration::run(20161115111846, " \
+            ALTER TABLE crates_categories \
+            ADD CONSTRAINT fk_crates_categories_category_id \
+            FOREIGN KEY (category_id) REFERENCES categories (id) \
+            ON DELETE CASCADE", " \
```
alexcrichton: This just means if we delete a category it'll auto-delete everything from `crates_categories`, right?

src/bin/sync-categories.rs
```diff
@@ -0,0 +1,48 @@
...
+    let categories = include_str!("../categories.txt");
+    let categories: Vec<_> = categories.lines().collect();
+    let insert = categories.iter()
+                           .map(|c| format!("('{}')", c))
+                           .collect::<Vec<_>>()
+                           .join(",");
+    let in_clause = categories.iter()
+                              .map(|c| format!("'{}'", c))
+                              .collect::<Vec<_>>()
+                              .join(",");
```
alexcrichton: Perhaps just for the sake of hygiene, but could we use rust-postgres's ability for escaping here? (rather than doing so ourselves).

I do realize though that `categories.txt` isn't user input, I figure it's just a good idea to stick with built-in escaping.

src/category.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub category: String,
+    pub created_at: String,
+    pub crates_cnt: i32,
+}
+
+impl Category {
+    pub fn find_by_category(conn: &GenericConnection, name: &str)
+                            -> CargoResult<Option<Category>> {
+        let stmt = try!(conn.prepare("SELECT * FROM categories \
+                                      WHERE category = $1"));
```
alexcrichton: Could you also add an index to the database for this `category` field? (unless I already missed it)

Also, perhaps the index and this could be based on `lower(category)` and `lower($1)` so we don't have to worry about case issues?

src/category.rs
```diff
@@ -0,0 +1,164 @@
...
+        let stmt = try!(conn.prepare("SELECT * FROM categories \
+                                      WHERE category = ANY($1)"));
+        let rows = try!(stmt.query(&[&names]));
+        Ok(rows.iter().map(|r| Model::from_row(&r)).collect())
+    }
+
+    pub fn encodable(self) -> EncodableCategory {
+        let Category { id: _, crates_cnt, category, created_at } = self;
+        EncodableCategory {
+            id: category.clone(),
```
alexcrichton: Could this be the lowercase version to be "pretty"?

Although now that I think more about this, maybe we should have two fields in the categories table. One for a slug (url friendly, this field) and another for the name? That way we can support categories with punctuation and spaces and such, all while keeping a nice url.

src/category.rs
```diff
@@ -0,0 +1,164 @@
...
+pub struct EncodableCategory {
+    pub id: String,
+    pub category: String,
+    pub created_at: String,
+    pub crates_cnt: i32,
+}
+
+impl Category {
+    pub fn find_by_category(conn: &GenericConnection, name: &str)
+                            -> CargoResult<Option<Category>> {
```
alexcrichton: If you want to change this to just return `CargoResult<Category>` an internalize the `chain_error(|| NotFound)` that'd also be fine.

src/category.rs
```diff
@@ -0,0 +1,164 @@
...
+        let old_categories = try!(krate.categories(conn));
+        let old_categories_ids: HashSet<_> = old_categories.iter().map(|cat| {
+            cat.id
+        }).collect();
+        // If a new category specified is not in the database, filter
+        // it out and don't add it.
+        let new_categories = try!(
+            Category::find_all_by_category(conn, categories)
+        );
+        let new_categories_ids: HashSet<_> = new_categories.iter().map(|cat| {
```
alexcrichton: It may be a good idea here to put a limit on the number of new categories that can be added. Something high like 50 or 100 but just don't want to blow out the database or anything like that.

src/tests/all.rs
```diff
@@ -245,32 +246,46 @@ fn mock_keyword(req: &mut Request, name: &str) -> Keyword {
...
     return req;
 }
 
+fn new_req_body_foo_version_2() -> Vec<u8> {
+    new_req_body(::krate("foo"), "2.0.0", Vec::new(), Vec::new(), Vec::new())
+}
+
 fn new_req_body(krate: Crate, version: &str, deps: Vec<u::CrateDependency>,
-                kws: Vec<String>) -> Vec<u8> {
+                kws: Vec<String>, cats: Vec<String>) -> Vec<u8> {
```
alexcrichton: 🐈  🐱

src/tests/category.rs
```diff
@@ -0,0 +1,106 @@
...
+    assert_eq!(cnt(&mut req, "cat2"), 1);
+
+    // Removing all categories
+    Category::update_crate(tx(&req), &krate, &[]).unwrap();
+    assert_eq!(cnt(&mut req, "cat1"), 0);
+    assert_eq!(cnt(&mut req, "cat2"), 0);
+
+    // Attempting to add one valid category and one invalid category
+    Category::update_crate(tx(&req), &krate, &["cat1".to_string(),
+                                               "catnope".to_string()]).unwrap();
```
alexcrichton: I was initially thinking we should error out on bad category names, but this make so much more sense. We shouldn't prevent everyone from publishing just because we deleted a category...

carols10cents:
> I wonder if there's a warning message we could provide back to Cargo for nonexistent categories? Either for a typo'd category or for just picking something that doesn't exist.

Yeah, I've been struggling with this one, because there doesn't seem to be a mechanism for returning *warnings* to cargo from crates.io right now, just *errors*, unless I'm missing something. I was thinking of making a separate request that only fetches the list of categories and then cargo creates the warning about an invalid category name? Or do you think it should get rolled into the /new crate request response?

> Right now we can't update a crate's metadata without publishing a new version, the addition of this feature is unfortunately likely to exacerbate this problem. Not something that needs to be fixed here per se, but in theory it's not too hard to add a cargo subcommand to update crate metadata...

I'm up for working on that if you'd like! We're tracking under my estimations at this point, so we should have hours for it. Definitely a new set of PRs though.

src/bin/migrate.rs
carols10cents: Correct, is that ok?

alexcrichton:
Oh I was thinking that whatever endpoint you use to upload errors would also transmit back warnings. We'd have to update Cargo yeah to process those warnings. I haven't looked at it in awhile, but I'd hope at least that it'd be backcompat to add more fields to the json response without breaking existing Cargos...

src/bin/sync-categories.rs
carols10cents: rust-postgres doesn't support this for the bulk import of multiple values that I wanted to do here, see sfackler/rust-postgres#218. I can change the `WHERE NOT IN` to use rust-postgres' escaping, though.

Another option is to make one insert for every category... this doesn't really NEED to be fast... wdyt?

src/category.rs
carols10cents: I can do that!

src/category.rs
carols10cents: Postgres automatically adds an index for UNIQUE fields (see https://www.postgresql.org/docs/current/static/ddl-constraints.html 5.3.3. Unique Constraints, "Adding a unique constraint will automatically create a unique B-tree index on the column or group of columns listed in the constraint.").

I can make it lower(category) though! I was thinking exact should mean exact, but it would probably be a bit friendlier to not care about case.

src/category.rs
carols10cents: Sure! Should categories.txt still list only the title-cased, punctuated version, and then the slug is lowercased, replace all spaces with hyphens, and remove other punctuation? Or should categories.txt list the slug as well?

If people specify the slug as their crate's category value, should that be valid?

src/category.rs
carols10cents: Hm, is that not taken care of by [Decode in upload.rs](https://github.com/rust-lang/crates.io/pull/473/files#diff-43c11b39ee478e157ef269ebb3922b54R125) before getting here? Originally, we weren't going to add a limit on categories, but then I saw keywords are limited to 5, and that seemed reasonable, so I did that as well. I'm happy to add a limit here too, though, preventing database overload is good! Just want to check that adding logic here isn't been *too* paranoid given upload.rs...

carols10cents:
> Oh I was thinking that whatever endpoint you use to upload errors would also transmit back warnings. We'd have to update Cargo yeah to process those warnings. I haven't looked at it in awhile, but I'd hope at least that it'd be backcompat to add more fields to the json response without breaking existing Cargos...

Yeah, I think that should work, I was just hesitant to add new protocols without checking. But since you're into it, I'll give it a try! :)

src/category.rs
alexcrichton: Oh yup that'd do it, missed that before I got here, sounds good to me!

alexcrichton: (that's the better location for a validation imo)

src/category.rs
alexcrichton: Postgres foils me again! If we add slugs then I think this becomes a non-issue, we can just ensure that all slugs are always lowercase.

src/bin/sync-categories.rs
alexcrichton: Eh I ended up seeing more of these throughout the codebase anyway (which were all reasonable), so it's fine to ignore this. A minor "concern" anyway.

alexcrichton:
Oh and a thought about sync-categories, perhaps that could just get executed whenever the server starts? That way we don't need to maintain a separate binary and don't need to modify deployments.

src/category.rs
carols10cents: Ok, if that's the case, then as a crate author I'd want to be able to see what the valid slugs are. I'm going to make a page [like pypi has](https://pypi.python.org/pypi?%3Aaction=list_classifiers) that's a plain text list of the exact valid category specifiers. We can reference that in the warning message when someone specifies an invalid slug too!

carols10cents:
Made some progress on this today, addressed all your comments I think, and added subcategories.

Buuuuut I broke category display on the crates page :( Something isn't hooked together the way ember wants it to be, and I'm not sure what I changed yet.

And I want to make the plain-text list of slugs still.

And my cargo PR needs to be updated to do something with the warnings this should return on unknown category names now.

Just wanted to push up my progress!

src/categories.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub fn sync() -> CargoResult<()> {
+    let conn = pg::Connection::connect(&env("DATABASE_URL")[..],
+                                             pg::TlsMode::None).unwrap();
+    let tx = conn.transaction().unwrap();
+
+    let categories = include_str!("./categories.txt");
+
+    let slug_categories: Vec<_> = categories.lines().map(|c| {
+        let mut parts = c.split(' ');
```
alexcrichton: This can be `splitn` so only two matches at most are returned perhaps?

src/category.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+
+        Ok(invalid_categories)
+    }
+
+    pub fn count_toplevel(conn: &GenericConnection) -> CargoResult<i64> {
+        let sql = format!("\
+            SELECT COUNT(*) \
+            FROM {} \
+            WHERE category NOT LIKE '%::%'",
```
alexcrichton: Are we thinking two levels of nesting max? I wonder if it's perhaps better to have a `parent_id` field in the table for a checked pointer?

alexcrichton: (also helps things be more explicit elsewhere)

src/categories.rs
carols10cents: Ah, cool!

src/category.rs
alexcrichton: Ah yeah if you have it on hand I wouldn't mind taking a peek.

carols10cents: > Ah yeah if you have it on hand I wouldn't mind taking a peek.

Ugh, it looks like I didn't check it in anywhere, unfortunately. 

The gist of it is for updating the category names, we wouldn't be able to do [one bulk upsert](https://github.com/rust-lang/crates.io/pull/473/files#diff-13d46f677e11061fc2b3c241e80537b9R30), we'd have to select each category's parent category first (in order to be able to set the parent_id), which means making sure parents are in the table before children, etc.

And to get all crates in a particular category and any of its subcategories, instead of [this query](https://github.com/rust-lang/crates.io/pull/473/files#diff-7ab30501521c288e1d8e1e994692b1a2R525) we'd have to do [a recursive CTE](http://stackoverflow.com/questions/3307480/postgresql-recursive-with) like:

```

WITH RECURSIVE recursetree(id, parent_ids) AS (
    SELECT id, NULL::int[] || parent_id
    FROM categories 
    WHERE parent_id IS NULL
  UNION ALL
    SELECT 
    c.id, 
    rt.parent_ids || c.parent_id
    FROM categories c
    JOIN recursetree rt ON rt.id = c.parent_id
  )

SELECT * 
FROM crates
INNER JOIN crates_categories 
ON crates.id = crates_categories.crate_id 
WHERE crates_categories.category_id IN (
  SELECT id
  FROM recursetree
  WHERE parent_ids @> ARRAY[(
      SELECT id 
      FROM categories 
      WHERE slug = 'development-tools'
  )]
  UNION
  SELECT id
  FROM categories
  WHERE slug = 'development-tools'
);
```

There might be ways to make this a little better, but it makes me feel kind of icky :-/

carols10cents:
Hmmm, interesting.... travis can't seem to fetch https://github.com/rust-lang/crates.io-index.... neat.

carols10cents:
Ok, I gave up trying to get around Ember to render plain text like pypi and I just made a page to list all the valid category slugs within the usual crates.io template:

slugs-list-page

I'm going to add this URL to the warning on the cargo side when a crate has an invalid category slug.

Some more screenshots:

The top-level categories list page:

categories-page

Clicking into the Libraries category, showing the subcategories and all crates in that category and its subcategories:

libraries-category-page

Clicking into the Libraries::Date and Time subcategory:

date-and-time-subcategories-page

Categories on a crate's page:

categories-on-crate-page

carols10cents:
Ok! I changed my mind about how to communicate about unrecognized category slugs, but that and the cargo side in rust-lang/cargo#3301 are done now. Both sides still happily tolerate the other side not being deployed yet.

So yeah, I think this is done, unless you think we should change to having parent_ids or have any other changes you'd like me to make.

alexcrichton:
Ok I think it makes sense to me to use `::` instead of `parent_id` from what you're saying, so sounds good to me. Want to update the Travis config to use stable Rust so we can see if this goes green?

carols10cents:
IT'S PASSING IT'S PASSING!!!!!

carols10cents:
I decided to get the PR for categories available started: https://github.com/rust-lang/crates.io/pull/488

est31:
:-1: 

I don't like cateogries. If you publish a crate thats outside of the list of already available categories, you will be forgotten. This leads to a totally pointless bikeshed disucssion which wil never end. Keywords are better, you can chose them yourself. Text search is better, it also finds the crates which don't have it in their keywords but on the description.

How do categories improve upon keywords? Yes, now there are not "HTTP-Server" and "HTTPServer" keywords, so its more unified, but what overall advantage is there?

est31:
Mhh, maybe there is some advantage in browsing. One idea for improvement: why not implement categories as groups of keywords? So e.g. a crate is in the Cryptography cateogry if it has one of the (crypto, cryptography, encryption) keywords. This would remove the need to retrofit each crate to the categories system and make publishing of new crates more simple.

Nemo157:
> why not implement categories as groups of keywords?

That was exactly what I was thinking while reading through this, it would definitely help with getting existing crates with good metadata into this system without forcing a new version to be published.

carols10cents:
> I don't like cateogries. If you publish a crate thats outside of the list of already available categories, you will be forgotten. This leads to a totally pointless bikeshed disucssion which wil never end. Keywords are better, you can chose them yourself. Text search is better, it also finds the crates which don't have it in their keywords but on the description.
> 
> How do categories improve upon keywords? Yes, now there are not "HTTP-Server" and "HTTPServer" keywords, so its more unified, but what overall advantage is there?

Categories are not meant to replace keywords, they are meant to *augment* them. Crate authors and crate users will be free to use whichever they find most useful! :)

carols10cents:
> Mhh, maybe there is some advantage in browsing. One idea for improvement: why not implement categories as groups of keywords? So e.g. a crate is in the Cryptography cateogry if it has one of the (crypto, cryptography, encryption) keywords. This would remove the need to retrofit each crate to the categories system and make publishing of new crates more simple.

> That was exactly what I was thinking while reading through this, it would definitely help with getting existing crates with good metadata into this system without forcing a new version to be published.

One of the problems with this approach is that there are keywords that should be split into two categories instead-- for instance, the [cli keyword](https://crates.io/keywords/cli) currently includes [argonaut](https://crates.io/crates/argonaut), "A simple argument parser" to help you *build* CLIs, and [betsey](https://crates.io/crates/betsy), "An AppVeyor cli written in Rust", which is an *application with a CLI* for use with a particular tool, appveyor. IMO these should end up in Libraries::Command-line interface and Applications::System tools, respectively.

carols10cents:
@alexcrichton We're feeling good about this PR-- categories.txt is now categories.toml, and we've tested out the following locally:

- Migrating to where crates.io currently is, to simulate deploying this PR to production
- Running the migrations added in this pull request
- Starting the server
- Navigating through categories without any crates in them
- Listing /category_slugs
- Sorting top-level categories alphabetically (default) and by number of crates

Using this PR and my cargo PR, I tested:

- Publishing a crate to a top-level category
- Publishing a crate to a mid-level category
- Publishing a crate to a leaf category
- Attempting to publish to a nonexistent category
- Viewing a crate page and seeing/navigating to its categories
- Removing a category from a crate
- Sorting crates within a category by downloads (default) and alphabetically

I also tested making these changes to categories.toml and restarting the server:

- Removing a category that a crate was in; the crate was disassociated with that record
- Adding a new category
- Editing a name and description; it did not affect crate associations

I was also seeing a bunch of postgres deadlocks, so I changed the travis config to use 1 test thread and that seems to be helping.....

PR #488 has been updated to have descriptions, so once everyone feels good about those, I think we can merge these two PRs and rust-lang/cargo#3301!

alexcrichton:
@carols10cents that all sounds great to me! I'm a little worried about the postgres deadlocks though. I can't quite seem to find a reference to that though, was that removed at some point?

carols10cents:
> @carols10cents that all sounds great to me! I'm a little worried about the postgres deadlocks though. I can't quite seem to find a reference to that though, was that removed at some point?

Reference to what, exactly? [This is the change I made to combat it](https://github.com/rust-lang/crates.io/pull/473/commits/1637de0a333b49f6fd609e20768991912bcfa395), here are a few builds that failed because of deadlocks:

* https://travis-ci.org/rust-lang/crates.io/builds/184356820
* https://travis-ci.org/rust-lang/crates.io/builds/184350232
* https://travis-ci.org/rust-lang/crates.io/builds/181026043
* https://travis-ci.org/rust-lang/crates.io/builds/178101962

I haven't seen this happening at all locally. I've seen this happen with tests run on rails apps with postgres and I've never had much luck debugging them :-/ decreasing the amount of queries trying to happen all at the same time seems to help... I'm going to try to get travis to dump some logs or something.

alexcrichton:
Oh sorry I didn't even realize we had a `--test-threads` argument, I was grepping for `RUST_THREADS_*` etc. I'll dig into those logs.

alexcrichton:
@carols10cents ok so the chapter on [postgres deadlocks](https://www.postgresql.org/docs/current/static/explicit-locking.html#LOCKING-DEADLOCKS) is helpful here, but this is definitely something we don't want to paper over. If we do it sounds like we'll *definitely* hit it in deploying.

Looks like there's just two concurrent transactions that want locks in different orders here. My guess is that the locks are table/row level based on `UPDATE` and `INSERT` statements rather than explicit locks (which IIRC we don't use).

It looks like it's always `good_categories` that's failing, which I guess locks the `crates` table for an insertion followed by a lock of the categories table to insert a relation. I guess the deadlock happens when something else locks the categories table and then tries to lock the crates table?

I can't say I've ever encountered an error like this before and am currently not away of conventional fixes (if any), but does that ring any bells? Perhaps we can reorder some statements somewhere to resolve this?

carols10cents:
I haven't had success reproducing this on my mac, and I downloaded a travis image was unable to reproduce the deadlock there, either. I have turned on travis on the integer32llc organization and I HAVE reproduced the deadlock on a different branch that's not a PR, I'm going to experiment over there in order to avoid clogging/waiting on the rust-lang travis queue :) 🤞

carols10cents:
I THINK I HAVE VANQUISHED THE DEADLOCK!!!! 

The categories::update_crate test does a lot of different manipulations of categories and crate categories and it was using a crate named foo. The good_categories test also used a crate named foo, and these two tests were causing a postgres deadlock.

I was able to cause deadlocks more often by duplicating the update_crate test and the good_categories test:

https://travis-ci.org/integer32llc/crates.io/builds/187302718

Making this change and running the duplicated tests resulted in 0 deadlocks:

https://travis-ci.org/integer32llc/crates.io/builds/187306433

This long-running editing of a crate's categories is unlikely to happen in production; requests get a database connection that gets closed when the request finishes, and the publish request only modifies the categories once, not as much as the update_crate test is. It seems unlikely that two people would publish the same crate at exactly the same time.

alexcrichton:
Ok that sounds good to me. Want to make sure crates have unique names and I'll merge?

carols10cents:
@alexcrichton done! all test crates now have a unique name :)

alexcrichton:
🎊

alexcrichton:
@carols10cents hm it looks like `cargo test` locally is failing, maybe due to a recent push to master? Mind taking a peek at that?

carols10cents:
When can we have bors on this repo? ;)

carols10cents:
@alexcrichton Hm, `cargo test` on master isn't failing for me locally, nor is it failing [on travis](https://travis-ci.org/rust-lang/crates.io/builds/187547951). Did you happen to try out a previous version of this branch? I did change some of the migrations along the way, maybe dry dropping and creating your cargo_registry_test database and see if that fixes it?

alexcrichton:
Oh looks like I was missing the S3_BUCKET business, my bad!

PR #475, opened 2016-11-22
Update travis config for yarn changes

steveklabnik:
See https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn


steveklabnik:
Ah, it's unfortunate that until `yarn` is included by default (which is coming, but not done yet), we'd need to renable root, which we don't need right now. Let's hold off until then.

PR #476, opened 2016-11-23
Use crate.id with links to force ember to load all the crate data

### LINKED ISSUE #331: Keywords are not shown via search box
When using the search box to find a crate and the using the link from the search results to navigate to that crates's page, keywords are not shown until the page is reloaded

This is yet another strange dependency on the referrer page, like issue #306.

carols10cents:
Fixes #331. Thanks to @locks, who knew exactly what the problem was!!!!

On crate list pages, we had *some* data about each crate, but not
keywords. The way the links were constructed, ember saw the partial data
that matched and would use that on the crate page instead of reloading
the whole crate from the crate show api JSON, which *does* have keywords
available. Using the id with the links instead of the whole model forces
the request for all the data to happen.

### DIFF
--- a/app/templates/components/crate-downloads-list.hbs
+++ b/app/templates/components/crate-downloads-list.hbs
@@ -1,7 +1,7 @@
 <ul>
     {{#each crates as |crate|}}
         <li>
-            {{#link-to 'crate' crate class='name'}}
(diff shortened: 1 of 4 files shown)


alexcrichton:
Nice!

PR #478, opened 2016-11-27
adding docs, homepage, and repo links to crate-row/search

### LINKED ISSUE #474: Show documentation/homepage/repository links directly in search results
This is quite a common workflow when I want documentation on crate 'foo':

1. go to crates.io
2. search for foo
3. click the correct result
4. click the 'documentation' link

It would be faster if it wasn't necessary to first click the result (step 3) to get the documentation link (also for home and repo, though less important).

ducks:
Fixes #474 

cratesio-search-quick-links-mobile
cratesio-search-quick-links-desktop

### DIFF
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -97,6 +97,7 @@
         padding: 20px 0;
         @include display-flex;
         @include justify-content(space-between);
+        @include flex-wrap(wrap);
 
         &:last-of-type { border: none; }
     }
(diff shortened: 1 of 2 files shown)


carols10cents:
zomg i love it!! 😻 😻 😻 😻 

The only tiny little decision that i can't decide which way would be better is what should happen if some of the links are missing. Right now whatever links are there float left:

screen shot 2016-11-28 at 5 04 24 pm

which I think is probably fine, and probably looks nicer than the alternative, which is having a "spot" for each link and leaving that link's spot empty if it's not there, so that you could scan down the page and know, say, the leftmost link will always be 'homepage', etc... That would probably be a pain to code actually 😂 

So I think I just convinced myself that the way it is here is fine :)

r+ modulo bors not being on this repo and me not being a committer ;) ;) ;)

ducks:
\o/

alexcrichton:
Ok, sounds good to me

dhardy:
Ace!

brson:
Sweet change @ducks!

withoutboats:
This has been such a benefit to me over the last few months! Thanks so much!!

PR #481, opened 2016-11-30
Display the version badge on crate list pages

carols10cents:
Instead of the plain-text version number.

This has the advantage of quickly showing pre-1.0 (orange) vs. post-1.0
(blue) when scanning down a list of crates.

Since people might want to copy the version number of a crate from the
crate list page, I've changed the alt text of the image to just be the
version number that used to appear here, so that you can highlight the
image and copy the version number.

Screenshot:

screen shot 2016-11-30 at 10 32 50 am

### DIFF
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -117,7 +117,12 @@
         text-decoration: none;
         font-size: 120%;
     }
-    .vers { margin-left: 10px; }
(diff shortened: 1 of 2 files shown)


bluss:
Now that this is live; this is not a better UI: the version number text is small and the badge is completely out of place.

PR #488, opened 2016-12-07
What categories should be available on crates.io? [BIKESHED ALERT 🚲🏠🚨]

carols10cents:
# TL;DR

[Categories are coming](https://github.com/rust-lang/crates.io/pull/473) to a crates.io near you! Help decide what categories should be made available! 

[Take a look at the diff](https://github.com/rust-lang/crates.io/pull/488/files) and consider:

As a crate user:

- Are the proposed categories all things that would fill in the blanks in statements like "I want a crate for ______" or "I want a crate that is a _____" or "I want a crate that does ______"?
- Are there any categories missing that you have gone to crates.io in search of?
- Are any of the proposed categories confusing, ambiguous, or duplicates?
- **What changes would you propose to fix the above issues?**

As a crate author:

- Are you struggling to decide what category your crate would fit into, in order for people to be able to find your crate for its intended use case? If so, why? (note: a crate can have up to 5 categories)
- Is the category you would want your crate to be in missing?
- **What changes would you propose to fix the above issues?**

# Details

- I have made an assumption/decision that categories should mostly be about the use case and purpose of a crate, and not for other attributes such as operating system support, development maturity, compatibility with other libraries, license, etc, and that we should address those attributes in another manner like badges. This is a different decision than, for example, [PyPI classifiers](https://pypi.python.org/pypi?%3Aaction=list_classifiers), where "classifiers" are what I'm calling "categories", and classifiers exist for all these axes *including* a "topic" classifier type. I'm proposing that we limit crates.io categories to the equivalent of PyPI topic classifiers. **If you disagree with this assumption/decision, please note that as well, I'm open to changing my mind on this**.
- This initial set of categories was adapted from [Awesome Rust](https://github.com/kud1ing/awesome-rust).
- Format of categories.txt
  - Each line the file is "[slug for URLs] [English display name]". 
  - Slugs are all lowercase and contain hyphens instead of spaces or punctuation [not allowed to appear in the path segment of a URL](https://tools.ietf.org/html/rfc3986#section-3.3).
  - Slugs are what crate authors will specify in their Cargo.toml and what will be used in URLs, [and they exist for these reasons](https://github.com/rust-lang/crates.io/pull/473#discussion_r88936117).
  - Anything after the first space on a line is part of the English display name for the label, and can contain any characters.
  - `::` indicates a sub-category relationship. [See screenshots for how subcategories will work](https://github.com/rust-lang/crates.io/pull/473#issuecomment-263601670).
- Once we decide on an initial list of available categories, the list can be changed by submitting a pull request to this file, which is the source of truth for category info. The crates.io database gets synced with this file [when the server boots](https://github.com/rust-lang/crates.io/pull/473/files#diff-c110e1433afdc14adcd9cda130d53415R65).
- Categories are different from keywords because:
  - Anyone can start using any keyword without having to go through a PR
  - Since available categories are curated, they should be higher quality than keywords (ex: we won't have categories for both [math](https://crates.io/keywords/math) and [mathematics](https://crates.io/keywords/mathematics)
  - Categories' purpose is to aid browsing, while keywords aid searching since they are [included in search rankings](https://github.com/rust-lang/crates.io/blob/be9e617f575743ef3e1d2a9963ebbf4f7a618306/src/bin/migrate.rs#L342) (categories won't be).
- Note that an RFC for the default ordering in which crates within a category are presented is coming; [the current implementation for categories](https://github.com/rust-lang/crates.io/pull/473) lists crates alphabetically.


shepmaster:
My first thought reading through is that there's a hierarchy of structured data. Representing that in a line-oriented manner is sub optimal. There's repetition of aspects like the English text for each category that seems brittle. I also am confused what would go in a category like `libraries::computation`, so I'd love a brief description for that. 

Some hypothetical JSON including this:

```js
[
    {
        slug: 'libraries',
        title: 'Libraries',
        description: "Libraries are the powerhouse of the cell"

        children: [
            {
                slug: 'astronomy',
                title: 'Astronomy',
                description: "Astronomy is the study of the things 'up there'. These libraries help in that pursuit"
            }
        ]
    }
]
```

carols10cents:
> My first thought reading through is that there's a hierarchy of structured data. Representing that in a line-oriented manner is sub optimal. There's repetition of aspects like the English text for each category that seems brittle. I also am confused what would go in a category like libraries::computation, so I'd love a brief description for that.
> 
> Some hypothetical JSON including this:

I like the JSON, especially with the ability to add a description. Should it be TOML tho? :)

```toml
[[categories]]
slug = "libraries"
title = "Libraries"
description = "Libraries are the powerhouse of the cell"

[categories.children] 
  slug = "astronomy"
  title = "Astronomy"
  description = "Astronomy is the study of the things 'up there'. These libraries help in that pursuit"
```

The current text representation came from me starting out with just the display name for each category per line, then @alexcrichton suggested [adding slugs](https://github.com/rust-lang/crates.io/pull/473#discussion_r88936117) in a simple way. Alex, wdyt about switching to JSON or TOML? (I know you're at #mozloha, I don't expect a quick response, but I don't want to put the work into doing this until you have a chance to weigh in ❤️ )

BurntSushi:
Ah this looks cool! I've got two thoughts/questions. :-)

First, can a crate belong to more than one category? If the answer is *yes*, then I feel like this makes it more like a *tag* concept than a categorical one, but that's just my own perspective. I don't know if others feel the same way.

Second...

> I have made an assumption/decision that categories should mostly be about the use case and purpose of a crate, and not for other attributes such as operating system support, development maturity, compatibility with other libraries, license, etc, and that we should address those attributes in another manner like badges. This is a different decision than, for example, PyPI classifiers, where "classifiers" are what I'm calling "categories", and classifiers exist for all these axes including a "topic" classifier type. I'm proposing that we limit crates.io categories to the equivalent of PyPI topic classifiers. If you disagree with this assumption/decision, please note that as well, I'm open to changing my mind on this.

I feel like this is a reasonable starting place. But let's say a year from now we wanted to change our minds and expand our categories into things other than just topic classifiers. Would that be feasible to do?

carols10cents:
> First, can a crate belong to more than one category? 

Yes, up to 5.

> If the answer is yes, then I feel like this makes it more like a tag concept than a categorical one, but that's just my own perspective. I don't know if others feel the same way.

To me, the current keyword functionality is more like tags, for the reasons I have in the description under the bullet point "Categories are different from keywords because:". 

Can you elaborate a bit on what the difference is, to you, between keywords and categories and tags, and how to best communicate that? 

> I feel like this is a reasonable starting place. But let's say a year from now we wanted to change our minds and expand our categories into things other than just topic classifiers. Would that be feasible to do?

Minimally, we could add additional categories with different semantics to this file, perhaps by adding some "top level" categories to distinguish them. I would not recommend adding the top-level "Topic" category like PyPI has at that point, since everyone would need to update their Cargo.tomls. I suppose we *could* just change the display name of existing categories to add "Topic::" or similar, with the understanding that the slugs wouldn't have that (and potentially be confusing).

BurntSushi:
> To me, the current keyword functionality is more like tags, for the reasons I have in the description under the bullet point "Categories are different from keywords because:".

> Can you elaborate a bit on what the difference is, to you, between keywords and categories and tags, and how to best communicate that?

Ah I see. I guess I feel like tags express a *many-to-many* relationship where as categories express a *one-to-many* relationship. I think keywords is the right word for what we have now, because they express a *many-to-many* relationship and are not curated. I guess I see tags as things that are curated, but also express *many-to-many* relationships.

This is all kind of subjective though. I think you might have been coming at it from the curated/uncurated perspective where as I'm coming at from the relationship perspective. No clue which one is right!

> Minimally, we could add additional categories with different semantics to this file, perhaps by adding some "top level" categories to distinguish them. I would not recommend adding the top-level "Topic" category like PyPI has at that point, since everyone would need to update their Cargo.tomls. I suppose we could just change the display name of existing categories to add "Topic::" or similar, with the understanding that the slugs wouldn't have that (and potentially be confusing).

Ah cool that seems reasonable to me!

carols10cents:
>>    I feel like this is a reasonable starting place. But let's say a year from now we wanted to change our minds and expand our categories into things other than just topic classifiers. Would that be feasible to do?

> Minimally, we could ....

Oh, thought of another thing we might want to add if we have categories for multiple dimensions... the way you navigate through classifiers on PyPI is that you can select any number of them to find libraries in the intersection. [Try it out](https://pypi.python.org/pypi?%3Aaction=browse)!

One of the reasons I'm not so much a fan of this approach is that I'd want the ability to filter *searches* for some of these attributes as well (ex: search for "piston" and limit to crates with windows support only), and it doesn't appear that PyPI lets you do that.

lise-henry:
Is this the good place to suggest categories? If yes, I'd like to suggest "(libraries::)internationalization".

(Also, not sure I get why there are separate categories for applications and libraries instead of e.g. having a separate flag?)

carols10cents:
> Is this the good place to suggest categories? If yes, I'd like to suggest "(libraries::)internationalization".

Yes it is!! Adding that one now :)

> (Also, not sure I get why there are separate categories for applications and libraries instead of e.g. having a separate flag?)

I could be argued out of it :) I mostly did that because that's what Awesome Rust did!

joshtriplett:
One potential ambiguity: do the various `development-tools::ffi::(language)` categories exist solely for *tools* that help people interface with those languages (such as binding generators or FFI helper libraries), or do they also include bindings to the core libraries of those languages (such as libpython bindings), or do they include bindings to non-core libraries of those languages as well?

If those categories have one of the broader meanings, I'd suggest widening the names from `development-tools::ffi::(language)` to `development::ffi::(language)` (or `dev::ffi::(language)`).  If those categories have one of the narrower meanings, then I'd suggest adding an additional category for the other meanings.

pornel:
I have mixed feelings about this. 

I do think that better discovery mechanism is definitely needed — currently homepage features very little content (most downloaded crates are boring), and exploring crates by alphabetically sorted names doesn't really work.

I'm a little bit worried about adding predetermined categories. This seems like a top-down decision strongly suggesting what Rust is intended to be used for. It's a change from a bottom-up approach of a soup of tags. I worry it may be discouraging writing something that doesn't fit any category, or that is in a category that's much smaller than other categories. Maybe categories could be built from most popular tags? (after cleanup of synonyms)

Being able to choose multiple categories is weird. To me it makes sense to have one primary category + keywords. Otherwise it overlaps with tags too much, especially when anybody can put their package into a category (i.e. categories are just another form of tag search, rather than [curated collections/showcases](https://github.com/showcases)).

It'd make sense to me if categories *were* tags, i.e. packages tagged with a category-related tag appear in that category.

joshtriplett:
The `applications::` categories seem to lack a few areas.  For instance, they include `applications::audio`, but no video or multimedia.  I'd suggest either adding those or making a single `applications::multimedia` category for all of them.

It *might* make sense to have a category for type of UI: `applications::gui` versus `applications::cli`, and possibly `applications::repl`.

I feel like `libraries` should have a category for libraries that exist specifically to solve Rust-specific problems, as opposed to some specific problem domain.  For instance, libraries like [owning_ref](https://crates.io/crates/owning_ref) and [rental](https://crates.io/crates/rental).

Finally, I think the existing categories need to distinguish between "graphics" (which should include things like image formats and drawing libraries) and "gui" (which should include things like toolkits and user-interface libraries).

pornel:
Now, the bikeshed:

* The file is hard to read. My first thought was — why did you lowercase "python", and why "python Development Tools" is a subcategory of the "FFI" category?

* I hope to cross-compile from Mac to Windows, so "Embedded::Cross compiling" seems too narrow.

* There's ruby and mruby FFI. Does Rust have that many packages for each? Shouldn't it be ffi::ruby::mruby or just one category (with details handled via tags?)

* libraries::command-line-argument-parsing could be subcategory of libraries::command-line-interface

Luthaf:
Let's bikeshed !

> development-tools::debugging Development Tools::Debugging
> development-tools::debugging::gdb Development Tools::Debugging::GDB
> development-tools::debugging::lldb Development Tools::Debugging::LLDB

I see the point of `development-tools::debugging`, but having sub-categories for GDB and LLDB feels weird: do we really expect to have a lot of crates in these categories?

> development-tools::ffi Development Tools::FFI::*

How do we decide which language gets in this list of categories? For example I see both mruby and ruby, but no micro-python. C++ as a well known language is missing too, Objective-C is there but not Swift or Go.

> libraries::astronomy Libraries::Astronomy
> libraries::bioinformatics Libraries::Bioinformatics
> libraries::machine-learning Libraries::Machine learning

I would unify these two categories in a `libraries::science` category, and maybe have a full hierarchy here:

```
libraries::science::math
libraries::science::physics  <= contains astronomy
libraries::science::chemistry
libraries::science::biology  <= contains bioinformatics
libraries::science::machine-learning
 ...
```

Having subcategories (`libraries::science::physics::astronomy`, `libraries::science::biology::bioinformatics`, etc.) could be appealing, but seems a bit premature given the number of existing scientific crates.


> libraries::command-line-argument-parsing Libraries::Command-line argument parsing
> libraries::command-line-interface Libraries::Command-line interface

These two could be unified too.

> libraries::computation Libraries::Computation

As @shepmaster, I do not see what would fit in there. 

> > (Also, not sure I get why there are separate categories for applications and libraries instead of e.g. having a separate flag?)
>
> I could be argued out of it :) I mostly did that because that's what Awesome Rust did!

I feel the same, for example my next project to be released on crates.io will be both an application and a library, plus the extra `libraries/applications` takes a lot of space and create noise.

Luthaf:
Also, we could maybe use the most used tags on crates.io as a starting point for the category list.

pornel:
* "cloud" is vague and probably overlaps with web-programming::deployment

* I don't know in which category I'd put analytics software (like https://piwik.org)

* Is crates.io intended for non-developers? The distinction between applications and libraries seems good for me on assumption that a non-technical user will be able to use apps, but won't have use for libraries. But if the whole site is for developers only, then maybe there's no point using that as top-down category.

pornel:
Maybe categories could be built from tags the way StackOverflow does it?

Every tag search page has "Related tags" in sidebar that narrows down the search:

https://stackoverflow.com/questions/tagged/rust
https://stackoverflow.com/questions/tagged/rust+ffi
https://stackoverflow.com/questions/tagged/rust+ffi+python
https://stackoverflow.com/questions/tagged/rust+ffi+python+numpy

That seems to automatically create a good hierarchy.

carols10cents:
> Also, we could maybe use the most used tags on crates.io as a starting point for the category list.

> Maybe categories could be built from most popular tags? (after cleanup of synonyms)

Here's [all the keywords](https://crates.io/keywords). Top 20, would any of these be good categories that aren't already covered here?

* ffi 495 crates
* windows 452 crates
* winsdk 400 crates
* google 228 crates
* api 175 crates
* web 175 crates
* protocol 164 crates
* cli 143 crates
* parser 90 crates
* graphics 82 crates
* database 82 crates
* crypto 77 crates
* piston 71 crates
* audio 65 crates
* bindings 64 crates
* math 62 crates
* macro 59 crates
* text 58 crates
* json 57 crates
* plugin 56 crates

> I'm a little bit worried about adding predetermined categories. This seems like a top-down decision strongly suggesting what Rust is intended to be used for. It's a change from a bottom-up approach of a soup of tags where anything goes.

I didn't start from the keywords because I personally don't find the current keywords useful at all, and it's mostly because of the "anything goes" approach. For example, [crates with the CLI keyword](https://crates.io/keywords/cli) include [argonaut](https://crates.io/crates/argonaut), "A simple argument parser" to help you *build* CLIs, and [betsey](https://crates.io/crates/betsy), "An AppVeyor cli written in Rust", which is an *application with a CLI* for use with a particular tool, appveyor.

And technically this is an *addition*, not a *change*. Adding categories is not getting rid of keywords, and people can add both to their crates and use either to find crates they're looking for.

>  I worry it may be discouraging writing something that doesn't fit any category, or that is in a category that's much smaller than other categories. 

I get what you're saying here, and I don't want to discourage anyone. Perhaps adding some documentation to the "All Categories" page about how to send a PR to suggest a new category would help?

Many other package websites have categories in ways similar to this, and I haven't experienced or heard of any problems like this. Have you? I'd be really interested, especially if there are any blog/forum posts discussing this from other communities!

I will be addressing the actual category bikeshed comments in a bit, wanted to speak to some of the meta-comments first :)

pornel:
libpng and Unreal Engine would probably both end up in Graphics category, which doesn't feel right (there's no specific category for 3d engines or image encoding).

dpc:
Just my 2c.

I wish we could have some community-driven tag-based system.

Eg. Maybe in `Cargo.toml` one could add tags to dependencies. Eg.


    [dependencies]     
    tokio = "1" #async #aio    
    log = "1" #logging     
    nom = "1" #parsing    
    serde = "1" #serialization  
    serde-json = "1" #json #serialization  


This way crates.io could use it as "votes", and each popular crate would get described by people that actually use it. Plus tags like that serve as a nice documentation of dependencies.

Wild idea, I know.

Diggsey:
Categories are a lot like OO class hierarchies, and I think the problems there will carry across (diamond inheritance and being unable to find a good structure that every crate can fit into). I think it could be OK as long as the hierarchies are kept relatively shallow.

What if within each category, crates.io could show "pseudo categories" based on the most common tags within those categories?

For example, instead of trying to list up-front all the sub-categories for "dev tools", explicit categorisation could stop there, and then within that, crates.io would look at the common tags within that category (eg. "testing") and show those as the next layer of subcategories. At some point later those common tags could be promoted into a full subcategory.

It could choose tags to show based on what fraction of the crates have the tag - ideal tags would be those where approximately 1/n crates have the tag, where "n" is the desired branching level of categories, say 10. The tags which come closest to this ratio would be shown as pseudo-subcategories.

This relies on being able to put crates into categories which are not leaf nodes on the tree of categories - is that possible? I think it should be possible to put a crate directly in the "Libraries" category without putting it in one of the sub-categories.

nagisa:
Something I want to chime in about categories.

As a crate author, I have hard time picking out categories. Mostly because I’m a lazy goon and dislike running through a list of ten hundred strings to pick five best fitting to my crate. This probably signifies that I’m much more comfortable with a list of ten (not tens of hundreds) very generic categories from which I pick one.

Whatever way this feature goes, I’ll probably end up not using it, but having less categories rather than more will certainly make me more prone to the appeals of categories :)

nokaa:
I agree with nagisa; too many categories defeats the purpose. I think that categories should *not* be added, and instead more emphasis should be placed on tags.

Differentiating between libraries and applications is something that would be very useful. Perhaps a more advanced search allowing for filtering of libs/applications and easier searching of tags. We already have tags as a part of the `Cargo.toml`; to me a better solution would be to display tags and allow searching for specific tags.

LeopoldArkham:
I have a [file size humanization library](https://github.com/LeopoldArkham/humansize), I don't think it fits in an existing category, and I'm not sure what to suggest.
I guess it's a utility maybe? That's fairly vague but it might be a good way to catch other such crates.

joshtriplett:
What many sites do is call everything a "tag", but distinguish between a small list of "official" tags that have a defined hierarchy and descriptions, and arbitrary "freeform" tags.  Both would appear in the same tag list within crates, but the site would prominently show official tags with descriptions, while using freeform tags just for search.

Keats:
I think tags (I think of tags and keywords as the same thing) and a better search would be better than categories. 
Searching for markdown and sorting by downloads has those results: http://imgur.com/a/khg7P 

2 issues looking at that screnshot:
 - it's not sorted by downloads (Imentioned it in https://github.com/rust-lang/crates.io/issues/396)
 - why is clippy number 2 when they don't mention markdown in the description or the tags

If we end up having categories, I'd agree with @BurntSushi and have only one possible category. The fact that categories are curated is not a big enough reason to have several imo.

If we have categories, I'd imagine something like (didn't list possible categories):

```yaml
bin:
   - games
   - audio etc
library:
   - ffi
   - science
   - web
   - cli
   - graphics
   - gamedev
   - crypto
   - database
   - markup etc
```

In short:
- 2 top levels, not 3
- just another level after the top one, no sub-subcategory
- fewer categories than what the current diff shows

Category should just be there to give a general idea, we have keywords for filtering down.

Aside: Is the full crate.io data available (ie with description, tags etc)? The [index](https://github.com/rust-lang/crates.io-index) only seems to contain what cargo needs.

jethrogb:
I'm missing categories for several of my crates:
- alloc_buddy_simple. (unless you think this should go under data structures)
- enclave, enclave-interface, sgx-isa, sgxs. These are all support libraries for very specific CPU features. The broadest category I can come up with is libraries::hardware-support or something like that.
 
Will there be descriptions and/or a manual to go along with the categories? For example I think it'd be important to note that libraries::hashing is not libraries::cryptography. And I'd like some guidance for where to put my library-example crates (same as library? applications? development tools?).

carols10cents:
> One potential ambiguity: do the various development-tools::ffi::(language) categories exist solely for tools that help people interface with those languages (such as binding generators or FFI helper libraries), or do they also include bindings to the core libraries of those languages (such as libpython bindings), or do they include bindings to non-core libraries of those languages as well?
> 
> If those categories have one of the broader meanings, I'd suggest widening the names from development-tools::ffi::(language) to development::ffi::(language) (or dev::ffi::(language)). If those categories have one of the narrower meanings, then I'd suggest adding an additional category for the other meanings.

I think If I were looking for a library to help me FFI with, say, Ruby, I probably wouldn't have a preconception necessarily about whether a binding generator/FFI helper library or a binding to libraries of those languages would be what I wanted, so I think the categories should have a broader meaning. Anyone who has more experience searching for FFI related crates, please chime in if this isn't the case for you!

I think I'm just going to move the FFI categories beneath "Libraries" for now, since most of the tools I know of integrate with your code in a library fashion somehow. Are there any purely external applications that do, say, binding generation?

carols10cents:
> The applications:: categories seem to lack a few areas. For instance, they include applications::audio, but no video or multimedia. I'd suggest either adding those or making a single applications::multimedia category for all of them.

Multimedia sounds good :)

> It might make sense to have a category for type of UI: applications::gui versus applications::cli, and possibly applications::repl.

I'm a little unsure about this one. I'm trying to imagine going to crates.io and saying "I really want an application that has a GUI today, to do X". I feel like I'd just want to do X and not particularly care about the interface. Can you give an example of a case where these categories would be useful?

> I feel like libraries should have a category for libraries that exist specifically to solve Rust-specific problems, as opposed to some specific problem domain. For instance, libraries like owning_ref and rental.

Adding Libraries::Ownership, wdyt? Any others? 

> Finally, I think the existing categories need to distinguish between "graphics" (which should include things like image formats and drawing libraries) and "gui" (which should include things like toolkits and user-interface libraries).

Added Libraries::GUI; this is definitely one that a description would help with.

carols10cents:
> I hope to cross-compile from Mac to Windows, so "Embedded::Cross compiling" seems too narrow.

Yeah, I can see that. I'm moving Cross compiling out of embedded.

> There's ruby and mruby FFI. Does Rust have that many packages for each? Shouldn't it be ffi::ruby::mruby or just one category (with details handled via tags?)

I'm fine merging them into one Ruby category :)

> libraries::command-line-argument-parsing could be subcategory of libraries::command-line-interface

I feel weird about this but I'm not sure why. I'm going to think about this one some more.

carols10cents:
> I see the point of development-tools::debugging, but having sub-categories for GDB and LLDB feels weird: do we really expect to have a lot of crates in these categories?

Sounds good-- Removing GDB and LLDB!

> How do we decide which language gets in this list of categories? For example I see both mruby and ruby, but no micro-python. C++ as a well known language is missing too, Objective-C is there but not Swift or Go.

Removed mruby at the suggestion of another comment; adding C++, Swift, and Go :)

> I would unify these two categories in a libraries::science category, and maybe have a full hierarchy here:
> 
> libraries::science::math
> libraries::science::physics  <= contains astronomy
> libraries::science::chemistry
> libraries::science::biology  <= contains bioinformatics
> libraries::science::machine-learning
>  ...
> 
> Having subcategories (libraries::science::physics::astronomy, libraries::science::biology::bioinformatics, etc.) could be appealing, but seems a bit premature given the number of existing scientific crates.

This sounds good!

>>    libraries::command-line-argument-parsing Libraries::Command-line argument parsing
>>    libraries::command-line-interface Libraries::Command-line interface
>
> These two could be unified too.

Yeah, I'm just not sure what to do about these, see [this comment](https://github.com/rust-lang/crates.io/pull/488#issuecomment-265647802). Maybe my hesitation is that I feel like "I need a library to parse arguments" is a really common thing to want, so I'd want that to be within easy reach. Still thinking on this one.

>>    libraries::computation Libraries::Computation
>
> As @shepmaster, I do not see what would fit in there.

Ok, this category is getting deleted then :)

>>>        (Also, not sure I get why there are separate categories for applications and libraries instead of e.g. having a separate flag?)
>>
>>    I could be argued out of it :) I mostly did that because that's what Awesome Rust did!
>
> I feel the same, for example my next project to be released on crates.io will be both an application and a library, plus the extra libraries/applications takes a lot of space and create noise.

Ok, I'm considering dropping the "Applications::" and "Libraries::" categories... going to sleep on it though. More opinions one way or the other would be welcome :)

carols10cents:
> "cloud" is vague and probably overlaps with web-programming::deployment

Gone! :) 🌩 

>    I don't know in which category I'd put analytics software (like https://piwik.org)

Is that the sort of thing that would be on crates.io though? I don't see anything for piwik now...

>    Is crates.io intended for non-developers? The distinction between applications and libraries seems good for me on assumption that a non-technical user will be able to use apps, but won't have use for libraries. But if the whole site is for developers only, then maybe there's no point using that as top-down category.

In my mind, crates.io is only for developers. There's so much developer-relevant and end-user irrelevant info. But developers use applications too! All of the cargo extensions are applications, and clippy, and rustfmt.

carols10cents:
> libpng and Unreal Engine would probably both end up in Graphics category, which doesn't feel right (there's no specific category for 3d engines or image encoding).

Added!

carols10cents:
> This relies on being able to put crates into categories which are not leaf nodes on the tree of categories - is that possible? I think it should be possible to put a crate directly in the "Libraries" category without putting it in one of the sub-categories.

Yes, this is possible. I'm not sure the pseudo-keyword-category and promotion to real category is worth the implementation complexity. Displaying keywords used by crates within a category could be easily added, though! I'd recommend filing a new issue for after we actually get categories merged :)

carols10cents:
> Something I want to chime in about categories.
> 
> As a crate author, I have hard time picking out categories. Mostly because I’m a lazy goon and dislike running through a list of ten hundred strings to pick five best fitting to my crate. This probably signifies that I’m much more comfortable with a list of ten (not tens of hundreds) very generic categories from which I pick one.
> 
> Whatever way this feature goes, I’ll probably end up not using it, but having less categories rather than more will certainly make me more prone to the appeals of categories :)

It's perfectly fine to pick just one category! Or no categories! I can see people feeling the opposite way though-- "none of these categories meet my needs, I need category X, Y, or Z or else I'll end up not using it".

It's mostly about how you want people to be able to find your crate to solve the problem your crate solves. If that's not something that's important to you, then you don't have to use categories :)

carols10cents:
> I have a [file size humanization library](https://github.com/LeopoldArkham/humansize), I don't think it fits in an existing category, and I'm not sure what to suggest.
> I guess it's a utility maybe? That's fairly vague but it might be a good way to catch other such crates.

Hm, yeah, utility does seem like it would be a junk drawer. Formatting? Emberobserver calls these things [Template Helpers](https://emberobserver.com/categories/template-helpers), and while that seems a little too specific for Rust, maybe Formatting helpers? Display helpers?

carols10cents:
> What many sites do is call everything a "tag", but distinguish between a small list of "official" tags that have a defined hierarchy and descriptions, and arbitrary "freeform" tags. Both would appear in the same tag list within crates, but the site would prominently show official tags with descriptions, while using freeform tags just for search.

You've pretty much described the exact behavior of Categories and Keywords :)

carols10cents:
> I'm missing categories for several of my crates:
> 
>     alloc_buddy_simple. (unless you think this should go under data structures)

Hm, I think Allocators would be common enough to warrant a category! Adding it to see what people think :)

>     enclave, enclave-interface, sgx-isa, sgxs. These are all support libraries for very specific CPU features. The broadest category I can come up with is libraries::hardware-support or something like that.

I'm into Hardware Support, adding that!


> Will there be descriptions and/or a manual to go along with the categories? For example I think it'd be important to note that libraries::hashing is not libraries::cryptography. 

I think descriptions will help a lot to keep the quality high. We have to figure out what data format we want for this file though :)

> And I'd like some guidance for where to put my library-example crates (same as library? applications? development tools?).

Is there a reason you've uploaded separate crates for the purpose of demonstrating the use of a library rather than [adding files in the library's `/examples` directory](http://doc.crates.io/manifest.html#examples)? 


And with that, I'm going to bed. Will be back to shed some more bikes with yinz tomorrow!!! ❤️ ❤️ ❤️

joshtriplett:
@carols10cents
> I think I'm just going to move the FFI categories beneath "Libraries" for now, since most of the tools I know of integrate with your code in a library fashion somehow. Are there any purely external applications that do, say, binding generation?

Yes: bindgen works as a standalone command-line tool.

joshtriplett:
@carols10cents 
>>  I feel like libraries should have a category for libraries that exist specifically to solve Rust-specific problems, as opposed to some specific problem domain. For instance, libraries like owning_ref and rental.

> Adding Libraries::Ownership, wdyt? Any others?

This seems much narrower than I meant.

I'm talking about the general libraries that exist to solve meta-problems that arise in Rust programming, rather than libraries that use Rust to solve some problem.  `owned_ref`, `rental`, `lazy_static`, `traitobject`, `rustc_version`, `void`, `unsafe-any`, `unreachable`, `nodrop`, ...

Does that make the category more clear?

joshtriplett:
> Hm, I think Allocators would be common enough to warrant a category! Adding it to see what people think :)

Perhaps generalize to memory-management?  That would include allocation, memory mapping, garbage collectors, reference counters, interfaces to foreign memory managers, and similar.

tomaka:
My remarks:

- "Game development" is very vague. For example should you put a crate that plays sounds in "game development"? Sure it could be used to create a game, but it's not specific to game development. What about `image`? `crossbeam`? `rand`?

- "3D Engine" should be "Game engines" because you probably also want to accept engines that only make 2D games.

- Why is there a "cmake" category? Do we expect dozens of wrappers around cmake?

- I have absolutely no idea what kind of thing you could put in the "Cross compiling" category.

- Shouldn't "Command line interface" and "Command line arguments parsing" be only one category? Sure there's a big difference between clap and ncurses, but ncurses would be more in the "GUI" category.

- I don't see why I would want to take a look at the list of FFI libraries for specifically the language X. If I know a library in X that I want to use in Rust, I'll just write its name in the search box.

- What's the "I/O" category? That's very vague. Should you put an HTTP client/server in it?

carols10cents:
> Perhaps generalize to memory-management? That would include allocation, memory mapping, garbage collectors, reference counters, interfaces to foreign memory managers, and similar.

That sounds good! Changing Allocators to Memory Management 👍

carols10cents:
> "Game development" is very vague. For example should you put a crate that plays sounds in "game development"? Sure it could be used to create a game, but it's not specific to game development. What about image? crossbeam? rand? hyper? Almost every library could be used for game development.
> 
> "3D Engine" should be "Game engines" because you probably also want to accept engines that only make 2D games.

I think "Game engines" could replace both "3D Engine" and "Game development", good call!

> Why is there a "cmake" category? Do we expect dozens of wrappers around cmake?

Literally just took that from Awesome Rust. I think we could get rid of the two subcategories for Build system and just have that.

> I have absolutely no idea what kind of thing you could put in the "Cross compiling" category.

[xargo](https://crates.io/crates/xargo) is one.

> Shouldn't "Command line interface" and "Command line arguments parsing" be only one category? Sure there's a big difference between clap and ncurses, but ncurses would be more in the "GUI" category.

Ok, I think I'm convinced that there should just be "Command line interface" and a description of this category would help keep it from becoming the [cli keyword](https://crates.io/keywords/cli) that also contains applications that are a CLI to something.

> I don't see why I would want to take a look at the list of FFI libraries for specifically the language X. If I know a library in X that I want to use in Rust, I'll just write its name in the search box.

What about [ruru](https://crates.io/crates/ruru) or [ruby-mri](https://crates.io/crates/ruby-mri)?

> What's the "I/O" category? That's very vague. Should you put an HTTP client/server in it?

Ok, taking it out! If anyone has a use case for the I/O category, please chime in!

wesleywiser:
I see a lot of `libraries::ffi::{lang}` categories. Are these categories for FFI bindings to libraries *written* in that language or for libraries that help you interface with that language? 

If it is the first category, I'm not sure how useful these categories will be. It seems like the `libraries::ffi::C` category will have 80% of the crates and the other categories will be extremely sparse by comparison. It also seems to me that if I want to use an external library, I already know the name of that library so what I really want is a way to search by library name and filter to just FFI crates.

If the category is just for libraries that help you interface with a specific language, I'm not sure that having the sub categories is useful either. How many competing libraries do we have that let you bind to Go/Javascript/Ruby/etc? It seems to me that all of those tools could simply go in a more general `libraries::ffi::binding` category. At least for `C`, it seems that some of the libraries that would go in that category are really targeting different frameworks, for example [`gir`](https://github.com/gtk-rs/gir) which can generate bindings for GNOME libraries (and therefore are less a `C` binding library than they are a GNOME binding library).

> Bikeshed 

- `libraries::os::windows` - for `winapi` + friends
- `libraries::os::linux` - for core Linux libraries like `DBus`, `systemd`, or raw Linux apis not exposed via `std::os::linux`
- `libraries::os::mac` - for macos [System Frameworks](https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/OSX_Technology_Overview/SystemFrameworks/SystemFrameworks.html)
- `development-tools::cargo-plugins` - for `cargo-{name}` tools
- Maybe one for libraries that have `no-std` support?

pornel:
`s/linux/unix/` maybe?

Knowing which are `no-std` is useful, but that cuts across all libraries. Similarly I'd like to know which are pure Rust, and which are wrappers of C libraries, but that's probably better implemented as filters or badges displayed in categories.

willi-kappler:
Hi carols10cents,

I would probably put my crate darwin.rs in these two categories:

```
libraries::science::math
libraries::science::biology

```
but I'm not sure if users would look for it in those.

I don't think that a special category `libraries::science::evolutionary` or `libraries::science::genetic` make sense, since it's so special and there would be too few crates.

But what about adding `libraries::science::cs` or `libraries::algorithms`, we also have `libraries::data-structures` ?

carols10cents:
> Maybe categories could be built from tags the way StackOverflow does it?
> 
> Every tag search page has "Related tags" in sidebar that narrows down the search:
> 
> https://stackoverflow.com/questions/tagged/rust
> https://stackoverflow.com/questions/tagged/rust+ffi
> https://stackoverflow.com/questions/tagged/rust+ffi+python
> https://stackoverflow.com/questions/tagged/rust+ffi+python+numpy
> 
> That seems to automatically create a good hierarchy.

I've filed issue #489 for this suggestion :)

carols10cents:
Just noticed I missed this:

> Yes: bindgen works as a standalone command-line tool.

Hmmmm, hmmm. This is further pushing me towards the side of "get rid of the library/application categories"

carols10cents:
Re: FFI: There are 495 [crates with the keyword ffi](https://crates.io/keywords/ffi) currently, which I feel like is a lot and could be broken down somehow to be more useful.

> If it is the first category, I'm not sure how useful these categories will be. It seems like the libraries::ffi::C category will have 80% of the crates and the other categories will be extremely sparse by comparison. It also seems to me that if I want to use an external library, I already know the name of that library so what I really want is a way to search by library name and filter to just FFI crates.
> 

I think this is a lot of what the ffi keyword is used for right now, and I agree with you that searching for the particular library you want to bind with is probably how people will find these crates. It seems like we should have SOME category for these crates to fit into, though, if only to keep them separate from the second kind. Do you have any ideas on what this category could be called? FFI::External libraries?

> If the category is just for libraries that help you interface with a specific language, I'm not sure that having the sub categories is useful either. How many competing libraries do we have that let you bind to Go/Javascript/Ruby/etc? 

I *think* I'd put these 6 crates in the Ruby category, I'm not totally familiar with the subtleties in this area though, so please correct me if I'm categorizing these oddly:

* [ruru](https://crates.io/crates/ruru)
* [ruby-mri](https://crates.io/crates/ruby-mri)
* [ruby-sys](https://crates.io/crates/ruby-sys)
* [rurust](https://crates.io/crates/rurust)
* [mri-sys](https://crates.io/crates/mri-sys)
* [mrusty](https://crates.io/crates/mrusty)

Searching for [python](https://crates.io/search?q=python) and [javascript](https://crates.io/search?q=javascript) looks like there are similar amounts for those as well. I'm mostly thinking that someone going for a Ruby FFI lib isn't going to want to have to sift through Python FFI libs too...

carols10cents:
>  libraries::os::windows - for winapi + friends
> libraries::os::linux - for core Linux libraries like DBus, systemd, or raw Linux apis not exposed via std::os::linux
> libraries::os::mac - for macos System Frameworks

WDYT about calling these "Libraries::OS::Windows APIs" etc, to make it clearer that these categories are not for indicating compatibility (which I think is better done with a badge)?

> development-tools::cargo-plugins - for cargo-{name} tools

That sounds good especially since I got rid of the Cargo/Cmake subcategories, adding!

> Maybe one for libraries that have no-std support?

I agree with @pornel's comment here, which I think was in response to yours:

> Knowing which are no-std is useful, but that cuts across all libraries. Similarly I'd like to know which are pure Rust, and which are wrappers of C libraries, but that's probably better implemented as filters or badges displayed in categories.

carols10cents:
> Hi carols10cents,
> 
> I would probably put my crate darwin.rs in these two categories:
> 
> libraries::science::math
> libraries::science::biology
> 
> but I'm not sure if users would look for it in those.
> 
> I don't think that a special category libraries::science::evolutionary or libraries::science::genetic make sense, since it's so special and there would be too few crates.
> 
> But what about adding libraries::science::cs or libraries::algorithms, we also have libraries::data-structures ?

Are evolutionary algorithms considered "machine learning" or no? That's probably where I'd put darwin-rs. 

What other crates would go in libraries::science::cs or libraries::algorithms that wouldn't go in any other category that we have so far?

willi-kappler:
> Are evolutionary algorithms considered "machine learning" or no? That's probably where I'd put darwin-rs.

In a classical sense no, but probably by some more modern definition ;-)
So I would put it in there also, thanks for the hint!

> What other crates would go in libraries::science::cs or libraries::algorithms that wouldn't go in any other category that we have so far?

itertools from bluss and for example some fancy sort / search algorithms s.o. may invent in the future ?
Other kind of optimization crates (linear optimization, min / max searches, etc.).

Maybe `libraries::science::plotting` or `libraries::science::visualization` is also worth adding.

And there could also be a `libraries::science::geo` (https://github.com/georust) category.

wesleywiser:
> Do you have any ideas on what this category could be called? FFI::External libraries?

That sounds like a good idea to me 👍 

> I think I'd put these 6 crates in the Ruby category, I'm not totally familiar with the subtleties in this area though, so please correct me if I'm categorizing these oddly:

I guess it just seems strange to me to have a separate category for such a small number of crates. `libraries::data-structures`, for example, might have hundreds or even thousands of crates in it. Breaking up the "language ffi crates" into sub categories seems like a case of *"premature categorization"* if you will.

> Searching for python and javascript looks like there are similar amounts for those as well. I'm mostly thinking that someone going for a Ruby FFI lib isn't going to want to have to sift through Python FFI libs too...

Sorry to keep bringing up search filters for categories, but if you could limit search results by category, then searching for "ruby" within `libraries::ffi` seems like it would be sufficient.

> WDYT about calling these "Libraries::OS::Windows APIs" etc, to make it clearer that these categories are not for indicating compatibility (which I think is better done with a badge)?

Sounds good to me 👍 

> re `no-std`

Yes, I agree. Badges are a better idea.

carols10cents:
> Sorry to keep bringing up search filters for categories, but if you could limit search results by category, then searching for "ruby" within libraries::ffi seems like it would be sufficient.

Don't be sorry!!! I think this is a really good point, you've convinced me-- but I've opened up [a new issue](https://github.com/rust-lang/crates.io/issues/491) and I'm going to cut the categories down to just Libraries::FFI, assuming that we'll add the ability to search within a category/keyword. 👍

wesleywiser:
Thanks @carols10cents!

carols10cents:
>> What other crates would go in libraries::science::cs or libraries::algorithms that wouldn't go in any other category that we have so far?
> 
> itertools from bluss and for example some fancy sort / search algorithms s.o. may invent in the future ?
> Other kind of optimization crates (linear optimization, min / max searches, etc.).

Great examples! You've convinced me that this is useful to have; I think I like libraries::algorithms for them. Hmm, should we remove hashing then, and have hashing be under algorithms? I think so.
 
> Maybe libraries::science::plotting or libraries::science::visualization is also worth adding.

What about just Libraries::Visualization? I'm adding that one; I could see people wanting to visualize some data but not really think they're doing it in a scientific setting so not thinking to look under science.

> And there could also be a libraries::science::geo (https://github.com/georust) category.

Is this different than Libraries::Geospatial? Should libraries::science::geo be instead of Libraries::Geospatial?

carols10cents:
> Ah, yes. I'm just worried that if we make a category called "Development Tools::Rust" that people will put every crate in there, since every crate is developed in Rust 😂😂😂 "Development Tools::Rust patterns"? "Development Tools::Rust helpers"?

Thought about this; I'm going to put Libraries::Rust Patterns in unless anyone has a better suggestion :)

joshtriplett:
@carols10cents Rust Patterns sounds perfect.

willi-kappler:
> Great examples! You've convinced me that this is useful to have; I think I like libraries::algorithms for them. Hmm, should we remove hashing then, and have hashing be under algorithms? I think so.

Thanks for adding!
Yes hashing seems too special and fits into algorithms.

> What about just Libraries::Visualization?

Perfect - people may want to visualize page access by web browsers from an apache log, or think about flame graphs that show CPU utilization by function calls, etc.

> Is this different than Libraries::Geospatial? Should libraries::science::geo be instead of Libraries::Geospatial?

There is more to geo-science than just geospatial data: think of topography, morphology, plate tectonics, climate, thermo-chronology, ...

shepmaster:
Just a note that we've reformatted the categories as TOML and added some descriptions. We dropped some categories that we found hard to describe or that overlapped too much with other categories.

carols10cents:
[Support for categories has been merged!!!](https://github.com/rust-lang/crates.io/pull/473) 

So how's everyone feeling about what categories should be available?

tomaka:
"Multimedia" is far too vague. Me alone I have ten crates that would go in this category, and I can name off the top of my head at least 20 other crates that would go in there.

At the very least it should at least be split between "Encoders/decoders", "Rendering" and "Audio".

joshtriplett:
@tomaka The categories are supposed to be reasonably broad. Tiny leaf categories doesn't help. If it turns out that a category has a huge number of crates in it, we can always create more subcategories at that point.

tomaka:
> Tiny leaf categories doesn't help.

That's a weird answer.
`authentication`, `caching`, `email`, `date-and-time`, `value-formatting`, `http-client`, `http-server` and `websockets` for example are all very tiny leaves.

In fact you can guess that the authors of these categories are mostly web developers, because between seven and nine of the categories are directly related to web development.
While game development (which has **far** more crates in the ecosystem than webdev today) has just "multimedia" and "game engine".
And game engine will most likely be empty because I'm not aware of any game engine in Rust yet.

If you have three different categories for http clients, http servers and websockets, I don't think it's unreasonable to ask to split crates that output or record audio, crates that perform rendering, and crates that decode binary data, since these three areas have absolutely nothing to do with each other.

carols10cents:
> At the very least it should at least be split between "Encoders/decoders", "Rendering" and "Audio".

I'm into adding more subcategories to Multimedia! I'm not sure which though...

Why audio but not image or video? 

Also we currently have a top-level [encoding](https://github.com/rust-lang/crates.io/pull/488/files#diff-a12c6cfaaf8154c288341caac09425f2R207) category, that I was thinking would be multimedia formats as well as text-based formats like JSON, etc... should the top-level Encoding category exist as well as a Multimedia::Encoding category? Should a crate that does image encoding be in both? Or should the top-level Encoding category be only for non-multimedia formats, and how do we define and communicate that?

carols10cents:
> And game engine will most likely be empty because I'm not aware of any game engine in Rust yet.

I get 52 results if I [search for game engine on crates.io](https://crates.io/search?q=game%20engine); looks like a lot of them on the front page call themselves game engines.

tomaka:
> Why audio but not image or video?

With "audio" I was thinking of crates that allow you to play sounds on your speakers, or record sound from your microphone. 

> Also we currently have a top-level encoding category, that I was thinking would be multimedia formats as well as text-based formats like JSON, etc... should the top-level Encoding category exist as well as a Multimedia::Encoding category?

I saw this category, but the description talks about serializing/deserializing which isn't an appropriate vocabulary for formats such as H264 or MP3 for example.

If crates can belong to multiple categories, maybe a crate that decodes MP3 could be in both "Audio" and "Encoding". In this situation "Multimedia" could be split into "Audio" and "Video".

However there are still some crates that I don't know where they would belong, like mathematical crates (vectors, matrices, collision detection, ... and geometry in general) or crates for inverse kinematics for example.

> I get 52 results if I search for game engine on crates.io; looks like a lot of them on the front page call themselves game engines.

Right, even though most of these project don't even more than a few hundred lines of code, it's probably worth having a category for this.

carols10cents:
> However there are still some crates that I don't know where they would belong, like mathematical crates (vectors, matrices, collision detection, ... and geometry in general) or crates for inverse kinematics for example.

There's Science, which we folded math and physics into:

```toml
[libraries.categories.science]
name = "Science"
description = """
Crates related to solving problems involving math, physics, chemistry, \
biology, machine learning, geoscience, and other scientific fields.\
"""
```

carols10cents:
@tomaka I've added some subcategories to Multimedia; please suggest concrete changes if this isn't what you meant (I might need you to provide exact wording for the descriptions, for example, since you're more familiar with this area than I am). Thanks!

I'd like to start a **final comment period** of sorts, lasting 1 week from whenever this week's TWiR comes out. Please let me know about any change suggestions before then! 

Note that pull requests can be made to this file after that time- editing display text, descriptions, and adding new categories should be no problem. Removing categories or editing slugs will need to take into account any crates that are using that category.

Marwes:
How about something like a scripting category? Thinking of it in context of both full scripting languages such as https://github.com/gluon-lang/gluon/ (mine) and https://github.com/PistonDevelopers/dyon but also for https://crates.io/crates/cpython, https://crates.io/crates/ruru etc.

```toml
[libraries.categories.scripting]
name = "Scripting"
description = """
Crates used to integrate scripting languages which can be used for extra flexibility in applications.
"""
```

ArtemGr:
"HTTP Server", but no "Server"? "HTTP Client", but no "Client"?

I think that a fixed set of tags (keywords) would be a mistake.
We should expose the existing tag folksonomy more, letting the crates.io visitors to easily navigate the most used and/or manually pinned tags, then the crate authors will adjust their tags accordingly.

carols10cents:
> "HTTP Server", but no "Server"? "HTTP Client", but no "Client"?

@ArtemGr Could you provide a description for the "Server" and "Client" categories that would distinguish them from HTTP Server or HTTP Client? Or are you proposing that we get rid of HTTP Server and HTTP Client and *replace* them with the more general categories? I'm not clear.

> I think that a fixed set of tags (keywords) would be a mistake.
> We should expose the existing tag folksonomy more, letting the crates.io visitors to easily navigate the most used and/or manually pinned tags, then the crate authors will be adjusting their tags accordingly.

Note that keywords are not going away, we're just *adding* categories to them. I agree that keywords could be more visible, which is why I filed #490.

carols10cents:
> How about something like a scripting category? Thinking of it in context of both full scripting languages such as https://github.com/gluon-lang/gluon/ (mine) and https://github.com/PistonDevelopers/dyon but also for https://crates.io/crates/cpython, https://crates.io/crates/ruru etc.
> 
> [libraries.categories.scripting]
> name = "Scripting"
> description = """
> Crates used to integrate scripting languages which can be used for extra flexibility in applications.
> """

@Marwes Thank you for providing a description! That helps me a lot :) ❤️ 

Do you think cpython and ruru belong in here or in [Development Tools::FFI](https://github.com/rust-lang/crates.io/pull/488/files#diff-a12c6cfaaf8154c288341caac09425f2R104)? I guess since crates can have up to 5 categories, they could be in both :) 

I'm mostly trying to decide if there's a useful distinction between crates that interface with existing scripting languages and crates that implement scripting languages in Rust. Also, do you see "Domain Specific Languages" (DSL) as an alternative label for the latter category?

ArtemGr:
> Could you provide a description for the "Server" and "Client" categories

From stackoverflow:
"A client is an application or system that accesses a service made available by a server."
"A server is a running instance of a software application capable of accepting requests from a client and giving responses accordingly."

> that would distinguish them from HTTP Server or HTTP Client?

The question makes no sense to me.

> Or are you proposing that we get rid of HTTP Server and HTTP Client and replace them with the more general categories?

I think the tree categories are inherently flawed. I am proposing to encourage the community in building a folksonomy and then hand-picking selected folksonomy for navigation. In other words, instead of putting packages into categories, we should put tags into categories. That way the categories remain flexible, they can be changed without affecting the packages and the other way around.

CamJN:
Because of the This week in Rust mention, I'm dropping by to beg for `stable` and `safe` categories, where `stable` means "works on stable rust compiler" and `safe` means "no unsafe blocks (or other unsafe stuff I'm not aware of)".

LukasKalbertodt:
(*I haven't had time to read all comments; sorry! I hope I don't repeat something that's already been mentioned*)

My (currently shitty) library `telegram-bot` strives to provide an easy way to interact with a HTTP-based API. This API has one use only: creating bots for the instant-messenger *Telegram*. I'm not quite sure what categories to put that in:

- `network-programming` ("Crates dealing with higher-level network protocols such as FTP, HTTP, or SSH, or lower-level network protocols such as TCP or UDP.") --> I'm not sure about that. The Telegram-API is not something most people would consider a real network protocol, especially because it's limited to this one use-case.
- `external-bindings` --> It's called "ffi" and the description says "other languages". Rather unfitting.
- `http-client`: Well... it's kind of a HTTP client, but I think this would be a confusing label.

So: I wouldn't really know in what categories said library would fit into. Maybe add a new category like `api-wrapper` or something?

carols10cents:
> My (currently shitty) library telegram-bot strives to provide an easy way to interact with a HTTP-based API. This API has one use only: creating bots for the instant-messenger Telegram. I'm not quite sure what categories to put that in:

>So: I wouldn't really know in what categories said library would fit into. Maybe add a new category like api-wrapper or something?

Hi @LukasKalbertodt! I'm conflicted about this. On one hand, these crates should have a place to go. On the other hand, if someone is looking for an API wrapper for a particular API like telegram, I think it's more likely they'd search for telegram and be able to find what they're looking for rather than browsing by category. Back to the first hand, maybe if we implement [searching within a category](https://github.com/rust-lang/crates.io/issues/491), this category would be useful to filter by and then search.

What about "HTTP API wrappers" to distinguish better from the "external bindings" category, that could also be described as API wrappers for library APIs?

LukasKalbertodt:
> I think it's more likely they'd search for telegram and be able to find what they're looking for rather than browsing by category.

Absolutely! And this is probably already sufficient; filtering by category as a first step might not be necessary for most of these API-wrappers. So, I think user won't really have a problem discovering `telegram-bot`, but it would be kind of unfortunate if a crate has not category to fit into...

"HTTP API wrappers" sounds too restrictive to me; there are probably plenty of api-wrappers not using HTTP. I also don't think it's easy to confuse with "external bindings" as the latter is still called `ffi`. I wouldn't call a crate which helps calling Ruby functions an "api wrapper"... Maybe rename "external bindings" to "external ffi bindings" or something like that? "external bindings" is pretty vague, maybe?

jan-hudec:
> I'm proposing that we limit crates.io categories to the equivalent of PyPI topic classifiers. If you disagree with this assumption/decision, please note that as well, I'm open to changing my mind on this.

I agree with it in principle, but think that:

1. Having a `library` (and `application`) main category goes _against_ that principle. Library is not a topic, developing an application for something is. `bindgen` is a good example of this and is certainly not the only one—the `libraries` and `dev-tools` categories have many common subcategories and for most of them the merge would make a lot of sense.

2. There are still several orthogonal axis that would make sense to cover here. If I am looking for something to do DNA sequencing, I can quite conceivably be looking for both tools that I can just use directly and libraries I can glue together in case the tools don't do exactly what I want. So I would be looking at `topic::science::biology` and there would be both applications and libraries and they would be distinguished by libraries also belonging to `interface::library`, command-line applications to `interface::command-line`, server applications to `interface::web`, plugins to `interface::plugin` etc. Or perhaps that could still be split to `role` (`library`, `plugin`, `application`, `service`) and `interface` (`cli`, `gui`, `network` (means program-oriented), `web` (means user-oriented)). Vaguely inspired by [debtags](https://debtags.debian.org/search/bytag).

    In this case, the number of categories a crate could be in should probably be adjusted to something like two per top-level category.

jan-hudec:
The categories should be curated. Shouldn't that means that assignment of crates to categories should be as well?

carols10cents:
> The categories should be curated. Shouldn't that means that assignment of crates to categories should be as well?

I don't think that necessarily follows. I think the guidance of what categories are available will be enough for crate authors to categorize their crates in a useful fashion; once this PR is merged, @shepmaster and I plan to send PRs to popular crates to help seed the categories with crates.

If not the crate authors, who do you propose should do this work?

> That could also handle author failing to find the right category and not having to upload a new version just to change the classification.

[We've considered adding a command to cargo to update metadata only](https://github.com/rust-lang/cargo/issues/971); as we add more metadata, that command becomes more important, definitely, it's just that no one has implemented it yet.

carols10cents:
Ok, I decided I've been convinced to remove the top-level Application and Library categories (but I kept the top level Developer Tools category). I moved all the subcategories up to the top level and merged them together where appropriate. 

The thing that tipped my decision is that now on /categories, you see 37 categories available, instead of only 3 (Applications, Developer Tools, and Libraries). I think this makes it clearer, sooner, the breadth of all the different kinds of crates available and makes a better first impression. Here's a preview:

Preview of what the main category list page will look like

I'd like to get a general sense of what everyone who's contributed to this issue thinks of this late, big change, if you have a minute please add a reaction to this comment:

- If you like the current version WITHOUT the top-level Applications and Libraries categories and think this is good enough to deploy, please add a thumbs-up.
- If you think the previous version WITH the tep-level Applications and Libraries categories was better, please add a thumbs-down.
- If you think there are other modifications that should be made before this is deployed, please leave a comment.

carols10cents:
> "HTTP API wrappers" sounds too restrictive to me; there are probably plenty of api-wrappers not using HTTP. I also don't think it's easy to confuse with "external bindings" as the latter is still called ffi. I wouldn't call a crate which helps calling Ruby functions an "api wrapper"... Maybe rename "external bindings" to "external ffi bindings" or something like that? "external bindings" is pretty vague, maybe?

@LukasKalbertodt In the last commit when I was merging the libraries and applications together, I also added "API bindings" ("Wrappers of specific APIs for convenient access from Rust. Includes 
HTTP API wrappers.") and changed "External bindings" to "External FFI bindings".

LukasKalbertodt:
@carols10cents Having only three categories shown at the top level (`/categories`) may look suboptimal, I agree. But I'd prefer to have a proper category tree instead of something too flat. 

It seems like the main reason for your decision to flatten out the tree is a visual one. Why not fix this by displaying the categories in a different manner instead of removing semantics from the category-tree? For example, we could have a special rendering rule for `/categories` which renders two levels of categories. Then everyone would clearly see the whole breadth of available categories, but one could still search in, for example, "applications" only.

carols10cents:
> It seems like the main reason for your decision to flatten out the tree is a visual one.

That was only the *last* reason for my decision :) The other factors include:

- "Application" and "Libraries" are not really "topic" areas and could better be expressed with an icon or a badge
- Some crates have both an application and a library
- Sometimes you might be fine if there was either a tool or libraries that could be used to build a tool to accomplish some task

Relevant comments: 
- https://github.com/rust-lang/crates.io/pull/488#issuecomment-265813624
- https://github.com/rust-lang/crates.io/pull/488#issuecomment-265596899
- https://github.com/rust-lang/crates.io/pull/488#issuecomment-271679879

carols10cents:
If no one comments on this today, I'm going to say it's done...

carols10cents:
Whew. Ok @alexcrichton, I think we're at least out of objections to the current color of the bikeshed :)

alexcrichton:
Awesome, thank for pushing this through @carols10cents!

tomaka:
Err I just realized that it's impossible to ever remove a category, since crates put the category name in their Cargo.toml.

Now I regret not continuing the discussion about the huge imbalance between gamedev and non-gamedev categories.

carols10cents:
> Err I just realized that it's impossible to ever remove a category, since crates put the category name in their Cargo.toml.

It is not impossible: crates that have invalid categories in their Cargo.toml will still get published, just with a warning in the output of `cargo publish` that directs them to take a look at https://crates.io/category_slugs.

Or was there something else you mean by "impossible"?

tomaka:
> Or was there something else you mean by "impossible"?

No, that's what I was thinking of.
I thought you'd get an error when publishing, which you be quite bad IMO.

If categories can be changed then it's ok to see in practice which categories will be full and need to be split, which categories will be empty and could be removed, and which categories are missing.

PR #503, opened 2017-01-15
Update postgres version used on travis to 9.5

carols10cents:
Since we've upgraded the version of postgres crates.io is using in prod, it should use the same version in tests.

This also means the [dist needs to be trusty](https://docs.travis-ci.com/user/database-setup/#Using-a-different-PostgreSQL-Version).


alexcrichton:
Thanks!

PR #504, opened 2017-01-16
Add support for showing Travis and Appveyor current build status badges

carols10cents:
This goes with corresponding PR rust-lang/cargo#3546. It probably makes the most sense for this one to go in first.

If a maintainer is testing their project on Travis and Appveyor and so chooses, they can specify attributes for badges for the crate's current build status in its Cargo.toml. 

The badges will then get displayed in alphabetical order on both pages that list crates:

preview of badges shown in a list of crates, next to the version badge

and an individual crate's page:

preview of badges shown on an individual crate's page, in the sidebar

If any badges other than appveyor or travis are specified, crates.io will ignore them and send back a warning for cargo to display. If the "repository" attribute for either badge is missing, crates.io will ignore that badge and warn about it. If any attributes other than the ones crates.io knows about for each badge is specified, those will get ignored and not warned about.


alexcrichton:
Looks great to me, thanks @carols10cents!

PR #506, opened 2017-01-19
Link to docs.rs if no documentation is provided

Manishearth:
/r/golang asked if we autolink to docs.rs, and we don't, but I think that's a really good idea!

I'm only touching UI code here so that the "default" docs.rs links don't filter down to the models and then the database.

r? @steveklabnik @wycats @carols10cents

### DIFF
--- a/app/templates/components/crate-row.hbs
+++ b/app/templates/components/crate-row.hbs
@@ -30,6 +30,8 @@
         {{/if}}
         {{#if crate.documentation}}
           <li><a href="{{crate.documentation}}">Documentation</a></li>
+        {{else}}
+          <li><a href="https://docs.rs/crate/{{crate.name}}">Documentation</a></li>
         {{/if}}
         {{#if crate.repository}}
           <li><a href="{{crate.repository}}">Repository</a></li>
--- a/app/templates/crate/version.hbs
+++ b/app/templates/crate/version.hbs
@@ -155,6 +155,8 @@
             {{/if}}
             {{#if crate.documentation}}
                 <li><a href="{{crate.documentation}}">Documentation</a></li>
+            {{else}}
+                <li><a href="https://docs.rs/crate/{{crate.name}}">Documentation</a></li>
             {{/if}}
             {{#if crate.repository}}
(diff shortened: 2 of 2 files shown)


Manishearth:
This does not update https://crates.io/crates/url/docs ; I'm not sure if it should.

carols10cents:
I think this has the same problems as #459, namely that:

- There's not a way to opt-out, other than specifying your own documentation link
- There's not a way to NOT generate this link if docs.rs failed to build the docs :-/

onur:
Hi Carol. I actually [asked you](https://github.com/onur/docs.rs/issues/87#issuecomment-265846799) about this but didn't get a reply. I can actually provide built info with a CORS header so crates.io can check if a doc successfully built or not.

Manishearth:
@onur that would be great, thanks!

onur:
Docs.rs is providing build info as json from following URL now:

`https://docs.rs/crate/<CRATE_NAME>/<CRATE_VERSION>/builds.json`.

It is also sending `Access-Control-Allow-Origin: *` header, it must be
ok to query this URL from crates.io.

A get request to this URL will return a list of build attempts. First
object in this list will always contain latest build attempt. This
object will contain following fields:

Example output of <https://docs.rs/crate/rand/0.3.15/builds.json>:

```json
[
  {
    "build_status":true,
    "build_time":"2016-12-29T10:10:49Z",
    "build_time_relative":"Dec 29, 2016",
    "cratesfyi_version":"cratesfyi 0.2.3 (eea831c 2016-12-29)",
    "id":1,
    "output":null,
    "rustc_version":"rustc 1.15.0-nightly (71c06a56a 2016-12-18)"
  }
]
```

`build_status` is `true` if docs.rs successfully built documentation of
a crate and `false` if docs.rs failed to build or a crate doesn't have
any documentation.

This list will contain zero elements if requested crate doesn't exists
in docs.rs (building a crate is usually taking 1-10 minutes for new
releases).

carols10cents:
@onur great! I'm a little concerned about the number of ajax requests that'll happen when loading a list of crates, but I think someone will have to try it out and see!

I'm going to close this one and #459 then!

PR #515, opened 2017-01-20
Category description tweaks

shepmaster:
We had some inconsistent casing of category names (`Foo Bar` vs `Foo bar`) and people were confused about the difference of `external-ffi-bindings` and `api-bindings`.


carols10cents:
LGTM!

PR #517, opened 2017-01-20
Show most popular keywords and categories on the home page

### LINKED ISSUE #490: Show keywords with the most crates on the front page (and categories too when they're done)
I really like how [Package Control](https://packagecontrol.io/) has "labels" (their version of keywords) on their front page, alongside similar things to what crates.io has on its homepage:

screenshot of packagecontrol.io showing how it lists the most used labels

I don't think /keywords is linked anywhere prominent right now, so it's hard to even know what keywords people are using on their crates unless you know to type in that URL :) This would help make keywords more visible and useful.

carols10cents:
Fixes #490. 

Below the lists of new, most downloaded, and just updated 10 crates will be the 10 most popular keywords and the 10 most popular categories, plus links to /keywords and /categories. It looks like this:

screen shot 2017-01-20 at 4 50 45 pm

### DIFF
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -125,7 +125,7 @@
 
 #home-crates {
     @include flex-wrap(wrap);
-    @include justify-content(center);
+    @include justify-content(left);
 
     > div {
       margin: 0;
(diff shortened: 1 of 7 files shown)


alexcrichton:
Oops sorry I forgot about this, looks great though!

PR #518, opened 2017-01-20
Make development category less Rust-specific

### LINKED ISSUE #511: Development tools category is too specific
I'm trying to determine the category to place the [addr2line](https://github.com/gimli-rs/addr2line) crate in, and I think `development-tools` is probably the best fit. However, the category description currently only allows tools that make **Rust** code easier to write. addr2line is language agnostic, and instead acts directly on binaries, which is not "allowed" by the description. Could it be changed to talk about development tools in general? Alternatively, could we split it into two categories: one for Rust tools, and one for development tools?

jonhoo:
Fixes #511.

### DIFF
--- a/src/categories.toml
+++ b/src/categories.toml
@@ -123,9 +123,8 @@ dimension.\
 [development-tools]
 name = "Development tools"
 description = """
-Make writing Rust code easier with code autocompletion, linting, \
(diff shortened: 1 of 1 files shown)


carols10cents:
Yay! Looks great, thank you!! ❤️

PR #523, opened 2017-01-23
categories: add parser and interpreter

Keruspe:


carols10cents:
What are some examples of crates that would go in this category, please?

carols10cents:
Also, what distinguishes the "Parser" category from the "Parsing" category?

Keruspe:
Not sure interpreter would fit that much crates after thinking back about it.

The description of the parsing category indicate that it's there for libraries that help you write parsers and that concrete parsers should go in other categories. The "parser" category would be for libraries implementing a parser to be used in come software.

carols10cents:
Ok! I've taken out interpreter then, and I tweaked parser to be "parser implementations" instead and expanded the description to hopefully distinguish clearly from parsing. Thanks!! 🌻

PR #525, opened 2017-01-24
Add rendering category and subcategories

tomaka:
I think a new category for this is justified, because of the large number of crates that belong in there.

Example crates:

For "engine": sdl, skia, ggez, kiss3d, piston-graphics and some of its plugins, the 2 or 3 ray tracing crates, etc. and the numerous bindings to existing libraries in other languages
For "graphics-api": gl, glutin, glfw, glium, gfx and all its plugins, glitter, the 4 or 5 vulkan bindings, the 4 metal bindings, directx, libovr, openvr, etc.
For "data-formats": the 4 crates that parse wavefront, the 4 bindings for assimp, gltf, collada, spine, text-related crates like freetype, etc.

For the base "rendering" category, everything that doesn't fit elsewhere: scene management (gfx_scene), maybe maths crates that have quaternions, frustrums and stuff (eg. cgmath).
I don't think it was worth creating more categories. Maybe in practice they will be needed but for now I don't think so.

### DIFF
--- a/src/categories.toml
+++ b/src/categories.toml
@@ -249,7 +249,7 @@ Crates that record, output, or process video.
 [multimedia.categories.images]
 name = "Images"
 description = """
-Crates that process or render images.
+Crates that process or build images.
 """
 
 [multimedia.categories.encoding]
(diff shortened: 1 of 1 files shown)


tomaka:
cc @carols10cents I guess you review categories?

src/categories.toml
```diff
@@ -297,6 +297,33 @@ formats. Format-specific parsers belong in other, more specific \
 categories.\
 """
 
+[rendering]
+name = "Rendering"
+description = """
+Crates related to rendering 2D or 3D graphics, \
+usually on the screen and with the help of the graphics card.\
```
shepmaster: If it's "usually on the screen", what else could it be? A file? If so, why would I pick this category vs `Images`?

tomaka: There are two domains when it comes to rendering: "real-time" rendering (as in video games) and "offline" rendering, for example when you render a movie like Toy Story into a movie file.

The category "images" would be more appropriate for example to resize or change the colors of an image, or for face regonition, etc.

Nobody familiar with offline rendering would have the idea to look in a category named "image", because it has nothing to do with that.

shepmaster: I like the online/offline terminology. What do you think of

> Crates related to online or offline rendering of 2D or 3D graphics, potentially with the help of a graphics card.

tomaka: "Online rendering" doesn't exist though. "Real-time rendering" is a really established terminology.

shepmaster: Oh yes, that's just me being scatter-brained.

carols10cents:
Looks great now!! Thank you!!

PR #526, opened 2017-01-24
Add testing in description of Development tools

malbarbo:
Searching (in the browser) for testing in https://crates.io/categories/ does not find anything. Adding testing in the description of "Development tools" helps find the "Development tools/Testing" subcategory.


carols10cents:
I'm into it. I think someone suggested somewhere (not an issue apparently since I can't find it right now) to have all the categories and subcategories expanded on /categories, but this will help in the meantime.

Issue #119, opened 2015-01-21
Reverse dependencies are listed multiple times


huonw:
e.g. https://crates.io/crates/libz-sys/reverse_dependencies lists libgit2-sys many times (presumably because of target-specific dependencies?)

PR #531, opened 2017-01-29
Show and sort by downloads on reverse dependency page

### LINKED ISSUE #495: Reverse dependencies should be ordered by number of downloads by default
When looking at a crate's reverse dependencies, right now it's ordered alphabetically. This doesn't convey as much information as ordering this list by number of downloads by default, since a crate could be assumed to be "better" if the crates that depend on this crate are popular.

LeopoldArkham:
Fixes #495 

I swapped the redundant arrow link on the right for the download count, I don't think anybody used it (?).

### DIFF
--- a/app/models/dependency.js
+++ b/app/models/dependency.js
@@ -15,6 +15,7 @@ export default DS.Model.extend({
     default_features: DS.attr('boolean'),
     features: DS.attr('string'),
     kind: DS.attr('string'),
+    downloads: DS.attr('number'),
 
     featureList: computed('features', function() {
         return this.get('features').split(',');
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -127,6 +127,8 @@
     }
     .downloads { @include display-flex; @include align-items(center); }
 
+    .rev-dep-downloads {padding-left: 7px}
+
     .quick-links {
         ul {
             @include display-flex;
--- a/app/templates/crate/reverse-dependencies.hbs
+++ b/app/templates/crate/reverse-dependencies.hbs
@@ -15,7 +15,7 @@
 
 <div class='white-rows'>
     {{#each model as |dependency|}}
-        <div class='row'>
+        <div class='crate row'>
             <div>
             {{#link-to 'crate' dependency.crate_id}}{{dependency.crate_id}}{{/link-to}} requires {{dependency.req}}
             </div>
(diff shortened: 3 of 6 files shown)


LeopoldArkham:
@carols10cents Ok, it works now. I had to abuse the CI because the tests won't run on my machine.
Let me know how it looks.

carols10cents:
@LeopoldArkham Hm, why doesn't the ordering in SQL work?

LeopoldArkham:
@carols10cents Whenever a query contains a `DISTINCT ON` statement and an `ORDER BY` clause, the leftmost argument to `ORDER BY` has to be the same as the one to `DISTINCT`.

So since we're DISTINCT ON crate_name, I can only ORDER BY crate_name. I think it would be possible to do in all SQL with a sub-query, but this seemed like a clearer and more maintainable way.

If there's a policy against this (or I'm missing something), I can go back and look into it.

carols10cents:
Oh! There's no policy or anything, it's just that SQL is really good at ordering things really fast :)

So the reason a reverse dependency might be duplicated is if a crate lists another crate as a dependency multiple times, [such as target-specific dependencies](https://github.com/rust-lang/crates.io/issues/119). So for crate A, it might depend on crate B twice, so crate B's reverse dependencies would have duplicate listing of A. However, in both listings of A, the number of crates.downloads for A will be the same. So we should be able to add crate_downloads to the list of columns specified in the DISTINCT clause, and meet the necessary criteria for the ordering, like this:

```
SELECT DISTINCT ON (crate_downloads, crate_name)
    dependencies.*,
    crates.downloads AS crate_downloads,
    crates.name AS crate_name
FROM dependencies
INNER JOIN versions
ON versions.id = dependencies.version_id
INNER JOIN crates
ON crates.id = versions.crate_id
WHERE dependencies.crate_id = 2
AND versions.num = crates.max_version
ORDER BY crate_downloads DESC;
```

What do you think? Is there anything *I'm* missing? (it's entirely possible!)

LeopoldArkham:
No, you're totally right, thank you! I had misunderstood the behavior of `DISTINCT ON` when provided with more than 1 argument. (See? No programming Sunday evenings.)

I also got the test suite to (mostly) work, so we should be OK.

LeopoldArkham:
^ I jinxed myself.
Although it's the yarn install that failed. Gonna close and re-open to trigger a rebuild.

carols10cents:
Looks great, works like a charm!!! 🍀 Sorry I took so long to review this, thank you for your patience!! ❤️

LeopoldArkham:
And you for yours!

PR #534, opened 2017-01-30
Fix #529 - Scroll position is reset when going back in history

### LINKED ISSUE #529: Scroll position is reset when going back in history
Steps to reproduce:

 * go to https://crates.io;
 * scroll to end of page;
 * click on the last crate of the list of new crates;
 * go back.

Expected result: the view is scrolled to the end, like it was before navigation.
Actual result: the view shows the top of the page.

This is very frustrating.

simonvandel:
Fixes #529 
Uses https://github.com/dollarshaveclub/ember-router-scroll.
Followed the usage steps given.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -1,8 +1,9 @@
 import Ember from 'ember';
 import config from './config/environment';
 import googlePageview from './mixins/google-pageview';
+import RouterScroll from 'ember-router-scroll';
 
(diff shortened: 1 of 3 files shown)


carols10cents:
Wooo this works great and will be so much nicer!!!!! Thank you!! ❤️

PR #535, opened 2017-01-30
adding auto redirect to repo when hitting crates /repo url

### LINKED ISSUE #520: Redirect to repo
I have a Chrome search for the crates io searchbar, so typing 'crate regex' gives the the search results directly for the regex crate. 

However, very often i just require the repo or the documentation of a crate. Currently I search for the crate on crates.io and then click (this got much quicker with the new links in the search results, thanks for that!). It would be terrific if crates.io could do a redirect service for me, so I could use a Chrome autosearch to quickly jump to the documentation. For example, `https://crates.io/crates/regex/documentation` should redirect to `https://doc.rust-lang.org/regex/regex/index.html`.

ducks:
fixes #520 by adding a /repo url redirect similar to /docs redirect

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -22,6 +22,7 @@ Router.map(function() {
 
         // Well-known routes
         this.route('docs');
+        this.route('repo');
     });
     this.route('me', function() {
         this.route('crates');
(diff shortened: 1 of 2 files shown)


carols10cents:
Works great!!! Thank you!! ❤️

PR #538, opened 2017-01-31
Added a new category for graphics related cargo packages

mmmbacon:
Added a new category for related packages


carols10cents:
What are some examples of crates that would fit into this category, please?

carols10cents:
I'm not sure what would distinguish this proposed Graphics category from either the Hardware Support category or the Multimedia category, which is primarily why I asked for examples of crates that would fit into this category. I'm closing this for now, please reopen with some more detail when you have time!

PR #545, opened 2017-02-05
Add development-tools.procedural-macro-helpers category

Nemo157:
There are two major tools that I think deserve this category on their own, `syn` and `quote`. I'm also working on a small helper for parsing attributes inside procedural macros, which doesn't fit into any of the existing categories directly.

@dtolnay have you looked into categories for `syn` and `quote` yet? Would you apply this category (or something similar) if it existed?

I'm not 100% sold on the name, if anyone can suggest a better one I'd be happy to change it.

### DIFF
--- a/src/categories.toml
+++ b/src/categories.toml
@@ -153,6 +153,12 @@ Crates to help you better interface with other languages. This \
 includes binding generators and helpful language constructs.\
 """
 
+[development-tools.categories.procedural-macro-helpers]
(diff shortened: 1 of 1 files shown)


dtolnay:
I agree that none of the existing categories are a good fit. I would apply this category (or something similar) if it existed. Thanks!

carols10cents:
Sounds good!!

PR #547, opened 2017-02-07
Fix intermittent badge::update_crate.

### LINKED ISSUE #546: Nightly build broke
Not entirely sure why because the PRs all passed, but i'll look into it tonight.

https://travis-ci.org/rust-lang/crates.io/builds/198708315

Susurrus:
Failure introduced in 4ce099b.

Closes #546

### DIFF
--- a/app/components/badge-gitlab.js
+++ b/app/components/badge-gitlab.js
@@ -8,6 +8,6 @@ export default Ember.Component.extend({
         return this.get('badge.attributes.branch') || 'master';
     }),
     text: Ember.computed('badge', function() {
-        return `Gitlab build status for the ${ this.get('branch') } branch`;
+        return `GitLab build status for the ${ this.get('branch') } branch`;
     })
 });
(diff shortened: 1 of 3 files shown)


Susurrus:
The first commit should fix the intermittent failures bug but I got frustrated when I was extending these badge tests originally because it was all one big test with a lot of continuing state throughout the test. I added the second commit (which I'll be fixing tonight) to break these up into better stop understand tests. There were also some things that weren't properly tested in the old huge test (like a test for adding an appveyor badge without a required field), so I added those tests in this refactoring. If you'd rather I break them out as a separate commit I can do that as well.

carols10cents:
I'm fine having it in the same PR, I'm just not going to merge it til you've got the tests passing, whenever you have a chance is fine! :)

Susurrus:
Will do, now that you documented the pgsql stuff (thanks, btw!), I'll get the tests working locally instead of trying to just use Travis to do everything! Very slow that way.

Susurrus:
@carols10cents got tests working locally which sped things up and I think this PR should be good to go. I snuck in a GitLab capitalization fix that is also necessary.

carols10cents:
Awesome, looks great!

Issue #491, opened 2016-12-08
Be able to search within a keyword or category


carols10cents:
Great idea from @wesleywiser on #488:

> It seems like the libraries::ffi::C category will have 80% of the crates and the other categories will be extremely sparse by comparison. It also seems to me that if I want to use an external library, I already know the name of that library so what I really want is a way to search by library name and filter to just FFI crates.

> I guess it just seems strange to me to have a separate category for such a small number of crates. libraries::data-structures, for example, might have hundreds or even thousands of crates in it. Breaking up the "language ffi crates" into sub categories seems like a case of "premature categorization" if you will.

> Sorry to keep bringing up search filters for categories, but if you could limit search results by category, then searching for "ruby" within libraries::ffi seems like it would be sufficient.

carols10cents:
Changed my mind about this being easy. The problem is that [each query parameter to the crate `index` API route is currently mutually exclusive](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L501), so this code would need to be rearranged to allow for building up the SQL to possibly include BOTH the [full text search](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L507-L510) AND the [category](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L542-L548) or [keyword](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L530-L534) conditions.

Then we'll need to decide whether to add a different search box that's closer to the category/keyword heading, or just change the behavior and appearance of the search box at the top to make it clear that using it on a keyword or category page will search within that keyword or category and not the whole page.

This could be done in pieces-- just getting the API accepting multiple parameters and using all of them could be one PR, and the UI stuff could be another.

carols10cents:
The backend part is done - see #666! Now to figure out the frontend...

neunenak:
I'll take a crack at this.

GregWoods:
Searching for no-std crates is vital for embedded (and others). Is there any progress on this?

carols10cents:
Any progress on this would appear on this issue, so no.

follower:
FWIW, I started looking into this and discovered that the work in #666 does appear to have also been _deployed_ on the server, so if you want, it seems it's possible to at least craft the queries by hand & view the returned JSON with browser dev tools...

e.g. <https://crates.io/api/v1/crates?page=1&per_page=50&q=graphics&category=no-std>

[...]

PR #548, opened 2017-02-07
Add a "No standard library" category

tarcieri:
Allows browsing of crates that are `#![no_std]` compatible.

### DIFF
--- a/src/categories.toml
+++ b/src/categories.toml
@@ -278,6 +278,12 @@ Crates dealing with higher-level network protocols such as FTP, HTTP, \
 or SSH, or lower-level network protocols such as TCP or UDP.\
 """
 
+[no-std]
+name = "No standard library"
+description = """
+Crates that are able to function without the Rust standard library.
+"""
+
 [os]
 name = "Operating systems"
 description = """


carols10cents:
How does this differ from the recently-added [Embedded category](https://crates.io/categories/embedded)?

tarcieri:
Granted there would be a lot of overlap, however...

`#![no_std]` is a pretty specific need: any project using `#![no_std]` can only transitively depend on `#![no_std]`-compatible crates.

Right now there simply isn't a good way to discover this information about crates. You can look through their source code or check their README, but you can't easily browse/search for crates which explicitly meet this requirement.

Shortly before Cargo categories were announced I posted this question on the internals Discourse:

https://internals.rust-lang.org/t/no-std-true-metadata-in-cargo-toml/4684

To make a specific point about how this differs from embedded though... from the "Embedded" description:

> Crates that are primarily useful on embedded devices or without an operating system.

`#![no_std]` is also useful for things like *authoring* operating systems or other OS infrastructure (e.g. a "libc" in Rust) which could run on systems that very much aren't embedded (e.g. your typical Dell PowerEdge-style server).

The actual use case is probably closer to "systems programming" than "embedded development". I would say "embedded development" is a subset of `#![no_std]` use cases.

steveklabnik:
As a practical example, https://crates.io/crates/ref_slice is not about
doing embedded work, but is a `no_std` crate.

tarcieri:
Generally I think of `#![no_std]` as being an entirely separate ecosystem of crates. If crates.io had an advanced search, it's the sort of thing that probably deserves a checkbox.

I think categories could suffice to accomplish the same thing though, if there were e.g. a "search within a category" feature

carols10cents:
Ok. When we were bikeshedding the initial set of categories, we were shying away from kinds of categorization that wasn't topic related (that is, would be useful to someone looking for "a crate that does X"). But I can see this being useful, and since we don't have a `no_std` badge/checkbox mechanism now...... let's give this a try.

Issue #491 is for tracking search within a keyword or category, which I agree would also be useful.

tarcieri:
Awesome, thank you!

PR #550, opened 2017-02-08
Add more documentation about crates.io's database needs

carols10cents:
Fixes #541.

PR #551, opened 2017-02-08
Revise backend documentation

Susurrus:
Following on #550, these changes separate out the initial backend configuration from the separate paths for hosting crates.io locally and running the backend tests. There's an extra commit in here to remove trailing whitespace as git likes to do that whenever you edit a text file, so I figured I'd include it.


carols10cents:
Love it, thank you!!!!

PR #552, opened 2017-02-10
Add 7-day moving averages to download graph. Plus UI tweaks.

hellp:
The non-stacked scattered chart as a basis reflects the download numbers
correctly. The stacked area chart before was a bit midleading. Both
approaches are a bit "noisy", so I added a 7-day moving average line for
each version number. This gives a higher-level overview and also makes it
easy to see when new versions overtake older ones in usage.

The colors were chosen to be colorblindness-friendly as well as having
a hot-cold metaphor.

Example, the futures crate, before and after:

screenshot-20170210-01701a

screenshot-20170210-01430a

### DIFF
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -81,7 +81,7 @@ export default Ember.Component.extend({
 
         // use a DataView to calculate an x-day moving average
         var days = 7;
-        var view = new google.visualization.DataView(myData);
+        var view = new window.google.visualization.DataView(myData);
         var moving_avg_func_for_col = function(col) {
             return function(dt, row) {
                 // For the last rows (the *first* days, remember, the dataset is
@@ -110,7 +110,8 @@ export default Ember.Component.extend({
         var [headers, ] = data;
         // Walk over the headers/colums in reverse order, as the newest version
         // is at the end, but in the UI we want it at the top of the chart legend.
-        _.forEach(_.range(headers.length - 1, 0, -1), (dataCol, i) => {
+
+        window._.range(headers.length - 1, 0, -1).forEach((dataCol, i) => {
             columns.push(dataCol); // add the column itself
             columns.push({ // add a 'calculated' column, the moving average
                 type: 'number',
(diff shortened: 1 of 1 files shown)


carols10cents:
Thanks, this looks awesome!! 🎁 

Since you sent in this PR, we've upgraded ember, which seems to have gotten rid of `window._`, so I pushed a commit to your branch importing lodash, which worked for me locally. I also merged in upstream and I'm going to wait to make sure CI is ok before I merge this in :)

carols10cents:
welp that didn't work, investigating....

carols10cents:
I think this is hitting https://github.com/ember-cli/ember-cli-qunit/issues/162 now...

hellp:
I haven't looked into the newly appeared issue, but if it's only about the lodash usage, I think I can rewrite this without using lodash. It was just convenient to have it.

carols10cents:
Turns out it was my fault :) I just rebased and force pushed to pick up this change: https://github.com/rust-lang/crates.io/pull/562

I think lodash is convenient too, seems fine. Let's see if this goes green.

carols10cents:
Yaaay, tests are green!!! 🍏 📗 💚 🥗 Tested it out locally and it's looking awesome! Thank you for your patience!!

durka:
Just my two cents: I saw the new graphs and I thought something was broken. Now reading this I like the moving average idea, but I think it needs to be indicated in the legend somehow.

durka:
I would even suggest removing the noisy dots. Or have a toggle to switch between averaged and raw data.

carols10cents:
Please open a new issue since this has been merged. Thanks!

On Feb 22, 2017 10:59 AM, "Alex Burka" <notifications@github.com> wrote:

I would even suggest removing the noisy dots. Or have a toggle to switch
between averaged and raw data.

—
You are receiving this because you modified the open/close state.

Reply to this email directly, view it on GitHub
<https://github.com/rust-lang/crates.io/pull/552#issuecomment-281712144>,
or mute the thread

bstrie:
Loving the color selection and the use of a rolling average. Agreed with @durka that the dots initially made me think that something was broken, and I'd second a PR to remove them from the main graph (though I think it's useful to retain the raw data in the hover).

carols10cents:
@bstrie could you make your comment on #567 too please? thanks!

Issue #553, opened 2017-02-11
Dependent Crates Link is broken for some users on Live


danielbodart:
Clicking the dependent crates link shows "Crate not found" for some users:

For example on lazy static:

https://crates.io/crates/lazy_static/reverse_dependencies

Shows:

http://imgur.com/a/szSaC

In the console the following URL is 404ing:

https://crates.io/crates/lazy_static/api/v1/crates/lazy_static/reverse_dependencies?page=1&per_page=10&_=1486799702812

This appears to be a per user problem so maybe it's relying on Ember state or a cookie. I.e some users it works for.

Have tried in Chrome and Firefox, logged in and out. Note was logged out to start so if feature requires to be logged in maybe it's caching the error.

carols10cents:
Looking into this now. I deployed an upgrade to ember last night, so that might be the cause.

carols10cents:
Yup, the ember upgrade seems to be the cause, I've rolled back that deploy til I figure out what the issue is. Thanks!

PR #501, opened 2017-01-06
Upgrade Ember and other dependencies where possible

bjornharrtell:
Thought I'd try an Ember and deps upgrade here as an exercise. Almost made it.

## Known TODOs

* [x] Assertion Failed: You attempted to define a `{{link-to "crate.version"}}` but did not pass the parameters required for generating its dynamic segments. You must provide a param `version_num`."
* [x] Assertion Failed: You modified crate.owners twice on <cargo@model:crate::ember764:log> in a single render
* [x] Two test failures

### Before
 - cargo-78d091696888a1996f55b55e06305ed1.js: 326.6 KB (32.12 KB gzipped)
 - cargo-dd3626a8e865f3873803b828b74aec11.css: 18.27 KB (3.91 KB gzipped)
 - vendor-31836fffce5288795738abce6ce5df0a.js: 804.41 KB (219.53 KB gzipped)
 - vendor-db1aae18a50d64dc7d57cc30890fa4a9.css: 1.82 KB (889 B gzipped)

### Ember 2.10
 - cargo-a2ea731112e31eb4ad816a2673775132.js: 195.24 KB (25.6 KB gzipped)
 - cargo-dd3626a8e865f3873803b828b74aec11.css: 18.27 KB (3.91 KB gzipped)
 - vendor-344087d1962d26f87e103b37bb470b18.js: 898.98 KB (234.69 KB gzipped)
 - vendor-924cc62bda01ca298267563aa0a649b0.css: 2.11 KB (969 B gzipped)

### Ember 2.11
 - cargo-1055d8d7726ed3652b8c5b15698dad5f.js: 203.79 KB (26.89 KB gzipped)
 - cargo-87493fb73363b542afc2eb3fb4cd5e3e.css: 18.24 KB (3.91 KB gzipped)
 - vendor-8c012c271ee0b49d3b8c93c77601c24f.js: 906.82 KB (235.8 KB gzipped)
 - vendor-924cc62bda01ca298267563aa0a649b0.css: 2.11 KB (969 B gzipped)

PR #555, opened 2017-02-12
Fix reverse dependencies page that didn't work after ember upgrade

carols10cents:
Apparently some behavior around `baseURL`/`rootURL` changed with the ember upgrade, and the reverse dependencies adapter constructs a custom URL that started being relative when it should be absolute. 

I also added a bunch more acceptance tests, which meant a bunch more mirage fixtures, to help us feel more confident about ember upgrades not breaking things in the future :)

I merged and deployed the ember upgrade (PR #501) last night, then I reverted the deploy but not the code today when I saw #553. So no git gymnastics needed, just a merge and then a deploy :)

### DIFF
--- a/app/adapters/dependency.js
+++ b/app/adapters/dependency.js
@@ -8,7 +8,7 @@ export default ApplicationAdapter.extend({
         delete query.reverse;
         var { crate } = query;
         delete query.crate;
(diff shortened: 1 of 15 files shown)


alexcrichton:
Nice!

carols10cents:
Nice I broke production but then i fixed it? LOL :)

Issue #524, opened 2017-01-24
Crate listings duplicated on category subpages


wesleywiser:
Some crates are being listed multiple times on the top-level category pages. For example, ["Development tools"](https://crates.io/categories/development-tools):

![image](https://cloud.githubusercontent.com/assets/831192/22256387/8a5f5c80-e228-11e6-8638-63a3903d2a3f.png)

It looks like this happens when the crate declares the top level category as well as one or more sub categories in its `Cargo.toml` file.

carols10cents:
Ah interesting. I guess it wasn't clear that a crate would automatically be in the supercategory if listed in the subcategory.

carols10cents:
The fix to this will involve [changing the query that gets crates in a category](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L549) to do a `SELECT DISTINCT`, and we should probably add a test for this... looks like I didn't add any tests for the crate index API route that use the `category` parameter, oops. 

The test should be [kind of like this keyword parameter test](https://github.com/rust-lang/crates.io/blob/master/src/tests/krate.rs#L127-L132) but will involve creating a category and a subcategory, adding a crate to that category and subcategory, and querying for the category.

Eh2406:
So I tried to write a test, I expect this version to pass: https://is.gd/1E9xIQ
But it fails with `bad response: (404, "Not Found")`.

When it passes the plan is to replace `&["kw1".into()]` with `&["kw1".into(), "kw1::foo".into()]` and have it fail. Then fix the bug and have it pass again.

What am I doing wrong? Am I on the right track?

carols10cents:
@Eh2406 so the example test I cited was with *keywords*, while this bug is for *categories*. The test code you have is still calling the code to add a *keyword* to the crate: `Keyword::update_crate(tx(&req), &krate, &["kw1".into()]).unwrap();` Instead, you'll need to add a *category*.

One difference between keywords and categories is that anyone can add a keyword by just starting to use it on their crate. Categories are set by crates.io, so you can only use a category that exists on crates.io already. So your test will need to create a category before putting a crate in the category. An example of a test that creates a category is [this one](https://github.com/rust-lang/crates.io/blob/master/src/tests/krate.rs#L897). And in this particular case, you'll want to create two categories actually: the parent category and the child category.

[...]

PR #558, opened 2017-02-18
Only show crate once, even if search is like several category #524

Eh2406:
This is a fix for #524.


Eh2406:
I don't know what ` || '::%'` syntax means so I left it.

carols10cents:
Looks great!!! 

The `|| '::%'` syntax means:

- `||` is the concat operator in postgres, confusingly
- `::` is the separator we're using to separate categories and subcategories
- `%` is the wildcard operator in postgres

so this finds all subcategories beneath a category. So leaving it there is the correct thing to do!

Thank you!!! 🎏

Issue #379, opened 2016-06-24
Dependent crates link not shown for crates without reverse dependencies


gnzlbg:
The dependent crates link is not shown for crates without reverse dependencies. 

I found this confusing. 

Maybe instead of removing the link we could add some text stating that the crate has no reverse dependencies: 

> (no dependent crates)

carols10cents:
This will involve adding an `else` to [this `if`](https://github.com/rust-lang/crates.io/blob/master/app/templates/crate/version.hbs#L162-L168) and displaying the text "(No dependent crates)" instead of the link.

Arnavion:
This can be closed right? (It wasn't autoclosed because the PR commit message didn't have a `Fixes ...`)

carols10cents:
Whoops, you are correct! Thank you!!

PR #559, opened 2017-02-19
Add text stating that the crate has no dependent crates #379

jptomanelli:
This is a fix for #379


carols10cents:
Awesome, thank you!!! 🍔

PR #560, opened 2017-02-19
Add explicit size to appveyor badge

killercup:
Appveyor badges seem to sporadically not load (for me). Previously, it looked like this:

bildschirmfoto 2017-02-19 um 16 02 22

But now, thanks to explicit width and height, it looks at least 42% less shitty:

bildschirmfoto 2017-02-19 um 16 02 45

I took the 106px width from a "build passing" badge. Let's hope the other variants are of that size, too :)

(Might also be a good idea to do that for the other badges.)


carols10cents:
> I took the 106px width from a "build passing" badge. Let's hope the other variants are of that size, too :)

It should scale if not :) Definitely looks *less* shitty! 🤣

PR #568, opened 2017-02-22
Tweak layout for elements on crate page to fix some small issues with narrow screens

### LINKED ISSUE #235: Insufficient spacing between Dev-Dependencies and Versions
Looking at: https://crates.io/crates/pnet_macros/

Using Firefox 42.0 on Linux:

2015-12-11-215439_439x229_scrot

You'll notice that the formatting makes it look like the dev-dependency `pnet_macros_support` requires a version `>= 0.10.6.0`, since there is insufficient spacing between the two columns.

evestera:
- Add some margin to header nav/menu and mobile search field to keep them away from the edges.
- Set flex-wrap: wrap on name/version heading to wrap rather than overflow (and potentially break layout) when the screen is too narrow. This approach is not the best-looking, but a lot better than doing nothing.
- Add some margin to crate-links sections (Fixes #235 ).

Before:
screen shot 2017-02-22 at 22 09 58

After:
screen shot 2017-02-22 at 22 10 29

Before:
screen shot 2017-02-22 at 22 12 51

After:
screen shot 2017-02-22 at 22 13 11

For the `#crate-links.section` margin it also applies to the right-most element(s), which is not quite optimal. An alternative solution without this minor issue is shown below, but this unfortunately makes the gray divider line too long, so I've stuck with the simple solution.

```diff
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -313,11 +313,14 @@
     padding-bottom: 20px;
     border-bottom: 5px solid $gray-border;
     margin-bottom: 30px;
+    margin-left: -5px;
+    margin-right: -5px;

     h3 { margin-bottom: 5px; }
     .section {
         @include flex(1);
-        margin-right: 10px;
+        padding-left: 5px;
+        padding-right: 5px;
         li { line-height: 24px; }
         .date { color: $main-color-light; padding-left: 5px; }
     }
```

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -65,7 +65,11 @@ body {
     }
 
     .sep { margin: 0 10px; color: darken($html-bg, 10%); }
-    .nav, .menu { @include flex-grow(2); text-align: right; }
+    .nav, .menu {
+        @include flex-grow(2);
+        text-align: right;
+        margin-right: 5px;
+    }
 
     .menu { display: none; }
     .menu ul.dropdown {
@@ -129,6 +133,8 @@ body {
 #mobile-search {
     display: none;
     margin-bottom: 10px;
+    margin-left: 5px;
+    margin-right: 5px;
     input.search {
         width: 100%;
         margin: 0;
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -9,6 +9,7 @@
 
     .info {
         @include display-flex;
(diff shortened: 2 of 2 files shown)


carols10cents:
Cool!! Looking good, I should have some time to take a closer look at this and try it out tomorrow :)

carols10cents:
This looks great to me at a whole bunch of widths! There's still more to do, as always, but this is better than what it was before :) Thank you!!! 💯

Issue #563, opened 2017-02-21
Add a "Copy to clipboard" button


Insomgla:
When on a page about a crate, we have first to select the content of the text area in order to paste it in the `Cargo.toml` file, which is quite a pain to do when browsing keyboard-only.

It would be a nice little improvement to have a button that copies the content of this textarea directly into the clipboard. Libraries such as https://github.com/zenorocha/clipboard.js seem to be able to handle that pretty smoothly, and is compatible with a broad variety of browsers.

carols10cents:
I'm confused, you'd like a button that you'd click on with your mouse to copy, to make it easier to browse keyboard-only?

steveklabnik:
Even without keyboard-only, this kind of thing is convenient.

Historically, it's required flash, but IIRC the requisite APIs in browsers are starting to land?

Insomgla:
Indeed, such features can now be easily done via Javascript. As for the keyboard-only navigation, that would allow me to use "TAB" to focus on the "copy button" and here we go. Besides, as @steveklabnik said, it's quite convenient anyway.

carols10cents:
Ok, that makes sense, just wanted to make sure I understood! 

So this will involve:

- `bower install`ing [clipboard.js](https://github.com/zenorocha/clipboard.js)
- Adding it to the HTML around [here](https://github.com/rust-lang/crates.io/blob/master/app/templates/crate/version.hbs#L60), setting it up to copy the text in that `code` element
- Testing it out on the major browsers (firefox, chrome, ie, safari?)
- Testing it out with keyboard navigation

If anyone thinking about picking this up has any other questions, please ask! ❤️

Insomgla:
I can try doing it.

Insomgla:
Just a little question, as I'm not really a pro with EmberJS, where do I put my `<script> `thingie so I can initialize `clipboardjs` ?

Insomgla:
Currently trying to implement it. I'll do my best to test on as most browsers as I can.

PR #570, opened 2017-02-23
Issue #563 - Add a "Copy button" for the crates

Insomgla:
A copy button is now displayed to the user if his browser supports ClipboardJS.

Has to be discussed now : how to inform the user that the content has actually been yanked.
    - tooltip: from scratch, using another library ?
    - display a message in an existing element of the page: where, and how to render it nicely ?


carols10cents:
Awesome!! On first glance this looks good, I'm going to give it a try tomorrow :)

Insomgla:
Sure thing. Again, I still need to figure out how tell the user that the content has been copied (or if there was an error). I would normally use a tooltip for that, but no Javascript library that's already imported does handle tooltips. I could try crafting one from scratch in pure CSS.

Insomgla:
I went very simple on this one, didn't like to import a bunch of Javascript libraries. The information is here, now I'm open to any suggestion on the design or else.

Insomgla:
Any input?

elliotekj:
A couple of things on this for me:

- The button doesn't show at all in Safari 9.x (OS X). 
- The button is using the vendor defaults for styling, it should use one of the button styles previously laid out in Crates' CSS. 

Tested on OS X with the latest releases of Firefox nightly and Chrome and Safari 9.x.

Insomgla:
Thanks ! I'll look at it ;)

Insomgla:
By the way, if the button doesn't show, it's probably because the browser is not supported. At least, the `isSupported()` method probably returns false on your browser.

Insomgla:
I did a little edit on the button's design according to your comment @elliotekj . I tried to use the button style that already exist, but it wasn't really nice. The only buttons already present on the websites are the rounded ones with the gradient-like colors, and I couldn't make it fit with the squarish shape of the crate info div.

carols10cents:
Hi @Insomgla, sorry about the delay!!! This looks great, I've tried it out and it works great!!

One question, what's the license on the copy to clipboard SVG? Did it come from somewhere? Just want to make sure we're allowed to use it before merging :) Thanks!!

Insomgla:
Hi @carols10cent, I used the icon they use on their demo website, which seems to be this one : https://commons.m.wikimedia.org/wiki/File:Octicons-clippy.svg

I am on my mobile so I cannot check if the icons are the same, but if they are, here you go :) If not, I'll replace the svg in my PR with the Wikimedia one.

elliotekj:
@carols10cents @Insomgla Just double checked this: both clipboard.js and octicons, where the icon is from, are licensed under MIT so we're all good.

carols10cents:
Perfect! Thank you for checking, both of you! <3

PR #571, opened 2017-02-23
Fix #467 - Author ordering

### LINKED ISSUE #467: Authors seem to be listed in exactly reverse order from Cargo.toml
https://crates.io/crates/error-chain/0.5.0

image

https://github.com/brson/error-chain/blob/6af2cc5be10f2ee848978ea9b168680c5fe48a23/Cargo.toml

```toml
authors = [ "Brian Anderson <banderson@mozilla.com>",
            "Paul Colomiets <paul@colomiets.name>",
            "Colin Kiegel <kiegel@gmx.de>"]
```

Also observed with other multi-author crates like `backtrace` and `hyper`.

Edit: If I'm reading the code right the order is whatever [the SQL query](https://github.com/rust-lang/crates.io/blob/fadf6b94caec4c9fc1afd4cc7767afdb1105668c/src/version.rs#L168-L180) returns. If so, and if you don't want to preserve the Cargo.toml order, perhaps they could be ordered alphabetically? It's not a big deal anyway.

Eh2406:
Rebased and deld the pr by accident.  so I resubmitted it.

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -10,7 +10,7 @@ use time::Duration;
 use time::Timespec;
 use url;
 
-use {Model, Crate, User};
+use {Model, Crate};
 use app::RequestApp;
 use db::RequestTransaction;
 use dependency::{Dependency, EncodableDependency, Kind};
@@ -33,9 +33,8 @@ pub struct Version {
     pub yanked: bool,
 }
 
-pub enum Author {
-    User(User), //TODO: not used. this should be removed.
-    Name(String),
+pub struct Author {
+    pub name: String
 }
 
 #[derive(RustcEncodable, RustcDecodable)]
@@ -170,14 +169,9 @@ impl Version {
                                        WHERE version_id = $1
                                        ORDER BY name ASC")?;
         let rows = stmt.query(&[&self.id])?;
-        rows.into_iter().map(|row| {
-            let user_id: Option<i32> = row.get("user_id");
-            let name: String = row.get("name");
-            Ok(match user_id {
-                Some(id) => Author::User(User::find(conn, id)?),
-                None => Author::Name(name),
-            })
-        }).collect()
+        Ok(rows.into_iter().map(|row| {
+            Author { name: row.get("name") }
+        }).collect())
     }
 
     pub fn add_author(&self,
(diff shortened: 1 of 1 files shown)


carols10cents:
Hi! I'm sorry for the delay in taking a look at this. I think it'd be better if the ordering was done by SQL instead of in Rust, just since ordering things is one of the tasks SQL is really really good at.

Could you change [the query for authors](https://github.com/rust-lang/crates.io/blob/fd7fb0517e3018452bd9b8c0d1ad2c7eb10fe675/src/version.rs#L169-L170) to include `ORDER BY name ASC` instead of the code you have here? Thanks! ❤️

Eh2406:
I want the sort to be done on what is displayed. What is the value of `row.get("name")` where `row.get("user_id")` is some? Is it always the same as `Author::User(try!(User::find(conn, id))).name`?
If so, we can simplify the authors loop a lot. If not, then we cannot just sort in sql on name.

Eh2406:
Looks like the only insert to `version_authors` is at [l189](https://github.com/rust-lang/crates.io/blob/fd7fb0517e3018452bd9b8c0d1ad2c7eb10fe675/src/version.rs#L187) with a `TODO: at least try to link name to a pre-existing user` and it only inserts a name. So a sort by `ORDER BY name ASC` will work the same but will brake when the `TODO` is fixed.

I see 3 fixes:
1. Add a left join to users and a `ORDER BY CASE WHEN version_authors.name IS NOT NULL version_authors.name ELSE users.name END ASC` Pro: Sort in sql. Con: Complex join for a feature we are not using and a complex sort that sql may not optimize.
2. Do the sort in rust. This is the current pr. Pro: No additional join. Con: Complexity in the rust and using sort, which may not be as fast as sql sort.
3. Make an issue for fixing the `TODO`. Add the simple `ORDER BY name ASC` as you suggested, with a `TODO` that links to the issue to fix it when the other `TODO` is fixed.

Do you have other suggestions?

carols10cents:
Ohhhh I understand now. I disagree with that TODO, actually, there's nothing requiring values in `authors` in Cargo.toml to match with anything in crates.io's database; authors is entirely separate from owners and I don't think it's possible to connect them. There are no rows with a value for `user_id` in the `version_authors` table in production, and I don't think anyone has asked for this feature.

So what about a 4th solution:

- Remove the TODO, and the `println!` right before it while you're in there, there's no reason for that either
- Remove the attempt at using `version_authors.user_id` and just map rows to `Author::Name(row.get("name"))`
- Add `ORDER BY name` to the query

Eh2406:
If we do that, presumably, we should remove all the Author as User code. By removing the variant of the [enum](https://github.com/rust-lang/crates.io/blob/fd7fb0517e3018452bd9b8c0d1ad2c7eb10fe675/src/version.rs#L37), and everything that references it. Including the [JSON response](https://github.com/rust-lang/crates.io/blob/fd7fb0517e3018452bd9b8c0d1ad2c7eb10fe675/src/version.rs#L334). And whatever consumes that JSON, would that be a breaking change to the API? Maybe we leave the api with a comment that it has no data, but is just there backwards compatibility.

Eh2406:
Ok this pr is now just a fix for #467 I will do a separate pr for removing `User(User)`.

carols10cents:
Looks great!!! Thank you for your patience!!! I don't think we need to worry too much about anything relying on this API, since it never returned any values anyway :) Perhaps we can make an /api/v2/ someday that gets rid of the `users` key in the json :)

Thank you!!! ❤️

Eh2406:
a Next Step would be to delete the useless columns in the database, I don't
know how to do that. but if you're mucking about with the database tomorrow
it might be a good opportunity.

PR #572, opened 2017-02-23
Replace try with ?

Eh2406:


Insomgla:
Or we can run `rustfmt` when your PR is merged? There's still an issue to solve with `rustfmt`,  so I guess you have time :)

PR #573, opened 2017-02-23
Sort optional dependencies last

### LINKED ISSUE #527: Sort optional dependencies last
e.g. https://crates.io/crates/image has multiple dependencies, some of which are optional. But from the layout it's not easy to see which ones are optional. Sorting optional dependencies separately would IMHO make it clearer.

Eh2406:
fixes #527

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -158,7 +158,8 @@ impl Version {
                                       FROM dependencies
                                       LEFT JOIN crates
                                         ON crates.id = dependencies.crate_id
-                                      WHERE dependencies.version_id = $1")?;
+                                      WHERE dependencies.version_id = $1
+                                      ORDER BY optional, name")?;
         let rows = stmt.query(&[&self.id])?;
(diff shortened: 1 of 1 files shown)


carols10cents:
Just curious, why does the `ORDER BY` include `crate_id` too?

Eh2406:
No really good reason. While I am removing arbitrary ordering, may as well do it hear. Alphabetical is easy and better than arbitrary. But I am not strongly wedded to it.

carols10cents:
Yeah, I think alphabetical by name would make more sense. Mind making that change? Thanks!

Eh2406:
I was assuming order by `crate_id` and order by name would be similar. I will go look at the schema to determine if `crates` or `dependencies` has a name. Then to make the switch.

Eh2406:
switched and rebased.

carols10cents:
Awesome, thank you!!! ☃️

Issue #493, opened 2016-12-13
Exact match not being obvious


Keats:
When doing a search, a crate matching exactly the search will appear first, even if sorted by downloads. I initially thought it was a bug: https://github.com/rust-lang/crates.io/issues/396#issuecomment-265080412

Would it be possible to extract the exact match out of the list and display it differently?

carols10cents:
Changing this bug's level to medium, because it involves a bunch of little pieces scattered all over. I think each of the pieces individually should be easy, but putting them all together might take a little while.

The pieces:

- SQL: Adding to the `SELECT` a column that has the value `name = $1`, which is what indicates that [this sort put the crate to the top](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L511). Example of how this works:

```
select id, name, name = 'sassers' as exact_match from crates;
 id |      name       | exact_match
----+-----------------+-------------
  2 | sassers         | t
  6 | piston          | f
```
- Rust: Adding an `exact_match: bool` field to [the `EncodableCrate` struct](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L58) to hold the value from that select query and send it back in the API results. Note that all the other places that `EncodableCrate` is used should just get `false` for this field... not sure if adding `exact_match: false` or implementing [the `Default` trait](https://doc.rust-lang.org/stable/std/default/trait.Default.html) on `EncodableCrate` will be more annoying :)

- Ember: Adding the `exact_match` property to the [Ember `crate` model](https://github.com/rust-lang/crates.io/blob/master/app/models/crate.js)

- Handlebars: Adding a class to [this element](https://github.com/rust-lang/crates.io/blob/master/app/templates/components/crate-row.hbs#L1) if `crate.exact_match` is true

- CSS: Styling that class to visually differentiate the exact match crate somehow :)

If anyone wants to pick off a piece of this, I'd accept PRs for each little bit separately-- I don't *think* any of these individually should break other things (famous last words ;) )

venkatagiri:
@carols10cents I wanted to work on this and ran `cargo test` before making any changes and few tests are failing trying to upload to S3 bucket. Logs are here http://pastebin.com/Hrei2g7H
README says "No actual requests to s3 will be made". Not sure what's wrong with my setup.

carols10cents:
[...]

PR #578, opened 2017-02-25
fix: incorrect host when running tests

venkatagiri:
As noted in https://github.com/rust-lang/crates.io/issues/493#issuecomment-282514820,
the recorded `http-data` in `src/tests` expects "http://alexcrichton-test.s3.amazonaws.com".
But, as the `S3_REGION` env var is loaded as `Some("")`,
it matches wrong leading to wrong host "http://alexcrichton-test-.s3.amazonaws.com".
This fixes it by handling match when region is empty.


carols10cents:
Looks great!! Thank you!! ❤️

Issue #574, opened 2017-02-24
Run rustfmt on the whole codebase and send in the changes


carols10cents:
I know I've been adding code to crates.io that isn't necessarily idiomatic rust formatting and I'd like to stop thinking about it and just run rustfmt :) There's probably a lot of code that isn't formatted the way rustfmt would do it though!!

This bug involves:

- Installing [rustfmt](https://github.com/rust-lang-nursery/rustfmt) on your machine if you haven't yet, using the "Quick start" instructions in their README
- Run `cargo fmt` in your crates.io checkout
- Send in a PR for whatever it recommends changing
- Fix the tests if any fail (hopefully they shouldn't!)

Insomgla:
1 PR for each change that rustfmt suggests, or a global one ? :)

carols10cents:
I think one PR with one commit that makes all the changes that rustfmt suggests in all the files is probably the easiest thing, although I haven't worked with rustfmt much. Is there an easy way to turn off all suggestions, then turn them on one at a time and make one commit per type of suggestion? That might be slightly easier to review, and would make it easier to see what caused any test failures (by using `git bisect`), should there happen to be any.

Wdyt?

elliotekj:
I happened to be looking at this one at the same time @Insomgla first commented. rustfmt can automatically fix all but 2 mis-formattings in the project. Of those two it can't fix, one involves renaming a function call to get under a 100 character line count (this one if memory serves https://github.com/rust-lang/crates.io/blob/master/src/db.rs#L61) but I'm not sure that can be fixed because the function is from the OpenSSL crate (https://github.com/sfackler/rust-openssl/blob/master/openssl/src/ssl/connector.rs#L112) and the other seemed to be a bug in rustfmt which couldn't correct a trailing whitespace mis-format, yet if it was corrected manually then rustfmt would add the whitespace back in and show an error again. It was late so I didn't look into it too closely and could be wrong though.

All that to say that my suggestion, FWIW, would be 1 commit with all of the auto-fixes then 1 commit per manual fix as they are not of the same kind.

Insomgla:
@elliotekj I fixed the second misformatting by removing the comment. As for the first one, I agree, we can't do much, except if Rust allowed to split function names :)

elliotekj:
[...]

PR #581, opened 2017-02-28
Issue #574 - Initial run of rustfmt on the whole codebase.

Insomgla:
Still 2 errors that rustfmt is not able to solve on its own.

I'll fix one by hand, and skip the other one since it's not possible to fix it automatically (it involves renaming a method of the OpenSSL library).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@
 npm-debug.log
 testem.log
 .env
+# rustfmt backup files
+*.bk
--- a/src/app.rs
+++ b/src/app.rs
@@ -32,23 +32,29 @@ pub struct App {
 
 /// The `AppMiddleware` injects an `App` instance into the `Request` extensions
 pub struct AppMiddleware {
-    app: Arc<App>
+    app: Arc<App>,
 }
 
 impl App {
     pub fn new(config: &Config) -> App {
-        let mut github = oauth2::Config::new(
-            &config.gh_client_id,
-            &config.gh_client_secret,
-            "https://github.com/login/oauth/authorize",
-            "https://github.com/login/oauth/access_token",
-        );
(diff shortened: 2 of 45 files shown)


Insomgla:
Wierd, by running `cargo fmt` again, it seems that some files didn't get formatted correctly at first.

Insomgla:
There's an issue with `rustfmt`, the skip attribute doesn't seem to work properly. It will still yell at me for a method, the name of which is too long. I'll open an issue on rustfmt's repository. Maybe I use it badly, though.

carols10cents:
@Insomgla let me know if you need help figuring out the rustfmt issue! Have you filed an issue with rustfmt yet? 

I think I'm going to merge the `?` change first, since it should be easier to run rustfmt again, hope you don't mind!!

Insomgla:
That would be a smart choice indeed :p and I opened an issue, yes.

carols10cents:
Ah cool, found it: https://github.com/rust-lang-nursery/rustfmt/issues/1345

carols10cents:
While not ideal, wdyt about setting rustfmt's max width to 138 (the length of [this line](https://github.com/rust-lang/crates.io/blob/master/src/db.rs#L61))? That would let us move forward, and it's not much higher than the default of 100.

carols10cents:
Hi @Insomgla, since this has gotten a bit out of date and both crates.io and rust-fmt have changed in the meantime, I'm going to give this a close. I'm still interested in having rust-fmt enabled, please feel free to submit a new PR in the future and I won't let it sit so long next time, I promise! :)

steveklabnik:
Also, using https://github.com/rust-lang-nursery/rustfmt/blob/master/rfc-rustfmt.toml would be a good idea 💯

Issue #502, opened 2017-01-11
Allow navigation from the "this crate has been yanked" page to the normal crate info page


sfackler:
If you're trying to track down e.g. the author of a yanked crate, it's currently pretty hard to do.

shepmaster:
You mean something other than navigation to the list of all versions, yes? For example, when I see a yanked page, it has a link to the other versions:

![image](https://cloud.githubusercontent.com/assets/174509/22159460/00bc5972-df0f-11e6-9e85-9e1cb09c0cbb.png)

Are you saying you'd like a way to see the content that *would* have appeared on `https://crates.io/crates/foo/1.2.3`, were the crate not yanked?

mathstuf:
Yes, that way authors, repositories, documentation, etc. can still be accessed.

dpc:
I'd like to suggest that if latest version have been yanked, the latest not-yanked version should be used. 

Eg. https://crates.io/crates/slog-syslog should just show version 0.8.2, like it's business as usual.

shepmaster:
@dpc you mean #510 ?

dpc:
@shepmaster Yes. :)

carols10cents:
Yes, @dpc, this issue is for the case when *all* versions of a crate have been yanked. Sorry for the confusion!

Issue #510, opened 2017-01-20
Redirecting to the latest crate should only use non-yanked versions


shepmaster:
Right now, if I visit `https://crates.io/crates/quickcheck`, I am redirected to the "This crate has been yanked" page (for version 0.4.2). However, we could just instead redirect to 0.4.1, the previous non-yanked version.

Eh2406:
related to #76

pnkfelix:
also related to (but distinct from) #145

carols10cents:
This has been fixed by the changes in #592 !

carols10cents:
Ok, this is not fixed completely. If I navigate to https://crates.io/crates/quickcheck directly, I correctly end up on the page for quickcheck 0.4.1.

HOWEVER if I *search* for quickcheck and click the search result link, I still end up with a page that says "this crate has been yanked" :( Something's not hooked up the way ember needs it to be :(

hoodie:
@carols10cents I'm still seeing this with `actix` right now
1. last "stable" version is `0.7.10` (yanked)
2. they just released `0.8.0-alpha.1`
3. suddenly the start page show "This crate has been yanked, but it is still available for download for other crates that may be depending on it."

sgrif:
@hoodie That's the expected behavior. crates.io/crates/actix shows the latest non-prerelease version, which has been yanked.

sgrif:
Scratch that, we should be redirecting to 0.7.9

sgrif:
This should be fixed by #1684  which I am deploying now

sgrif:
This appears to be fixed

Issue #470, opened 2016-11-16
Make some env vars optional for easier dev setup


carols10cents:
Right now [`cargo_registry::env` panics if there isn't a value set for an env var](https://github.com/rust-lang/crates.io/blob/e190f2beff4a57ee08b77986b66bc60c6508fe95/src/lib.rs#L198), so you can't run the server locally unless you set values for everything in `.env`. 

If you're just working on part of the backend that you don't need users logged in for, there's no reason to have to set up github tokens and all that.

I tried to make this easier to set up by setting the values in `.env.sample` to `""`, which gets around the panic, but apparently [that caused problems with zsh](https://github.com/rust-lang/crates.io/issues/417).

I think it'd be better to have another function, like `env_with_default` or something, and maybe warn that functionality that needs that env var isn't going to work?

I'm happy to do this, just not right this second, so I wanted to open this to remind me (and maybe entice someone else to do it, i don't think it would be very hard #famouslastwords)

alexcrichton:
Sounds good to me!

kelcecil:
I'd be happy to take this task.

kelcecil:
My thought is that we may want to still panic for certain environment variables not being set where we absolutely can't progress. TEST_DATABASE_URL, S3_* variables, and GH_* variables would all be okay for defaults, but the Postgres URL and GIT_* variables would pretty much be required and should panic if not set. Thoughts?

alexcrichton:
@kelcecil yeah sounds good to me. We may want to assert that in production everything is set, but otherwise sounds good

carols10cents:
I *think* we fixed this. If anyone hits this problem, please reopen!

PR #582, opened 2017-03-01
Update max_version on yank

### LINKED ISSUE #76: Yanking a crate should update its max_version
If a crate has its max version yanked, it should update the max_version field to go back to the otherwise maximum version.

yodaldevoid:
Fixes #76

### DIFF
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -29,7 +29,7 @@ export default Ember.Route.extend({
         return crate.get('versions')
             .then(versions => {
                 const version = versions.find(version => version.get('num') === params.version_num);
-                if (!version) {
+                if (params.version_num && !version) {
                     this.controllerFor('application').set('nextFlashError',
                         `Version '${params.version_num}' of crate '${crate.get('name')}' does not exist`);
                 }
--- a/src/bin/update-max-versions.rs
+++ b/src/bin/update-max-versions.rs
@@ -33,3 +33,73 @@ fn update(tx: &postgres::transaction::Transaction) {
                      &[&new_max.map(|v| v.to_string()), &crate_id]).unwrap();
     }
 }
+
+#[cfg(test)]
+mod test {
+    use std::collections::HashMap;
+
+    use postgres;
+    use semver;
+
+    use cargo_registry::{Version, Crate, User, Model, env};
+
+    fn conn() -> postgres::Connection {
+        postgres::Connection::connect(&env("TEST_DATABASE_URL")[..],
+                                      postgres::TlsMode::None).unwrap()
+    }
+
+    fn user(conn: &postgres::transaction::Transaction) -> User{
+        User::find_or_insert(conn, 2, "login", None, None, None,
+                             "access_token", "api_token").unwrap()
+    }
+
+    #[test]
+    fn max_to_null() {
+        let conn = conn();
+        let tx = conn.transaction().unwrap();
+        let user = user(&tx);
+        let krate = Crate::find_or_insert(&tx, "foo", user.id, &None, &None,
+                                          &None, &None, &None, &None,
+                                          &None, None).unwrap();
+        let v1 = semver::Version::parse("1.0.0").unwrap();
+        let version = Version::insert(&tx, krate.id, &v1, &HashMap::new(), &[]).unwrap();
+        version.yank(&conn, true).unwrap();
+        ::update(&tx);
+        assert_eq!(Crate::find(&tx, krate.id).unwrap().max_version, None);
+    }
+
+    #[test]
+    fn max_to_same() {
+        let conn = conn();
+        let tx = conn.transaction().unwrap();
+        let user = user(&tx);
+        let krate = Crate::find_or_insert(&tx, "foo", user.id, &None, &None,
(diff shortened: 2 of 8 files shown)


yodaldevoid:
One thing to note, these changes have yet to be tested with the front end. The only thing that comes to mind as a possible problem is the fact that max_version can now be expected to be null. It was never restricted to not being null, but it was never a true possibility due to the backend code.

yodaldevoid:
I have looked into testing the frontend response to a null max_value for a crate. The only test I can think of is to start up a local backend and frontend, publish a crate to the local backend, yank the crate from the local backend, and finally try to view it on the frontend. My problem comes from trying to publish to the local backend using `cargo`. I have added the config block spat out by script/init-local-index.sh, but cargo always fails with a 404 error if `--host http://127.0.0.1:8888/` is passed to `cargo` and an SSL error if `--host https://127.0.0.1:8888/` is passed to `cargo`. Is there something I could be forgetting, or am I simply going about this wrong?

carols10cents:
> I have added the config block spat out by script/init-local-index.sh, but cargo always fails with a 404 error if --host http://127.0.0.1:8888/ is passed to cargo and an SSL error if --host https://127.0.0.1:8888/ is passed to cargo. Is there something I could be forgetting, or am I simply going about this wrong?

Ugh yeah, `--host` isn't actually host of the crates server, it wants the location of the index :( When running locally with crates.io setup as recommended in the readme, this value should be something like `file:///path/to/your/crates.io/checkout/tmp/index-co/`. This has bothered me for a while, I'm going to go file a bug with cargo right now.

The next problem that you'll probably run into is that, unless you've set up an s3 bucket, publishing to your local crates.io will fail trying to upload the code to s3. I get around this by temporarily applying [this commit](https://github.com/integer32llc/crates.io/commit/4633bb5d67017a1673eeb8518dd177be41ca9d68) to disable uploads to s3 and just update the database-- this has been bothering me as well and I want to fix it as part of #470.

Basically I'm sorry i'm sorry.

carols10cents:
Alternately, I'm happy to test this out on my setup, since I've already gone through this pain. Let me know!! I'm looking at the code now :)

carols10cents:
The code is looking great!! 

Other than checking out the frontend, the only other thing this change will need is a way to go through existing crates and update those whose max_versions have been yanked to have their next highest non-yanked version used (or set to null if all versions have been yanked) as [retep998 reminded me](https://github.com/rust-lang/crates.io/issues/76#issuecomment-281502943) 

I should have elaborated in the issue, but we have some one-off tasks in src/bin, such as [this task that backfilled users' github IDs](https://github.com/rust-lang/crates.io/blob/master/src/bin/fill-in-user-id.rs) (ignore the docs at the top of that file, they're copy-pasted from another binary).

If you don't want to take care of that for whatever reason, I'm happy to, let me know!

yodaldevoid:
Thanks for pointing me in the right direction! Don't worry about the problems, now we know more about what needs to be improved. I'll be honest, whenever I find a place where documentation is lacking when working with Rust I feel a little better as it reminds me that everyone working on this language/project is human.

It's interesting that `--host` is used for publishing, but `--index` (which honestly makes more sense) is used for yanking. At any rate, here is the front page of a crate with a null max_version.
null_crate
I don't think the red bar at the top is acceptable, but everything else seems fine to me.

I'll look at writing the one off task. It should be fairly easy to make as most of the logic is already in `yank`.

yodaldevoid:
Hmmm, this is a little concerning. I tested publishing a newer version of a crate when the max version was null, and it seems that the max version is not being updated. I'll look at adding a test and code to fix this.

carols10cents:
Everything you've said sounds great! Also I [filed a bug with cargo about the `--host` flag.](https://github.com/rust-lang/cargo/issues/3797)

> I'll be honest, whenever I find a place where documentation is lacking when working with Rust I feel a little better as it reminds me that everyone working on this language/project is human.

We are all very much human :) ❤️

yodaldevoid:
Everything should be wrapped up. I tested it against my local copy and it worked. I don't have tests for the update task, but can add those if you wish.

yodaldevoid:
Going through the issues I found a couple more issues that this addresses. #510 might be fixed, but there might be some more work to be done there.

Fixes #502

sgrif:
Just to note, #592 is an alternate solution to this.

carols10cents:
Oh geez. Before merging either this or #592, I want to hear from alex why max_version was added in the first place. And yes, the tests in here are valuable no matter what ❤️

yodaldevoid:
I have added tests to the one-off.

carols10cents:
@yodaldevoid thank you so much for your work on this, and I hate to do this-- I really should have questioned whether we actually need this column like @sgrif did in #592. I'm definitely going to merge your tests in, which will be valuable in making sure the code in #592 works, but if it does and if production performs ok with the max_version column removed, we might not need this code. 

But I want to reassure you that your code is great and I'm very grateful!! ❤️

yodaldevoid:
No problem, really! I'm all for simplifying the database.

carols10cents:
Thank you for understanding!!!

yodaldevoid:
Closing because #592

Issue #588, opened 2017-03-03
Consider migrating from Rust-Postgres to Diesel


Eh2406:
Diesel has better performance at runtime and moves errors to compile time. This sounds like a win win.

However, they are both good libraries, and diesel is under active development. So, it is an open question whether it is worth migrating. I hope this can be a place for that discussion to happen.

steveklabnik:
Tagged with those tags.

I know @sgrif has said that this is a pre-req for diesel 1.0, but I don't think we as a team have made the decision to actually do this, just that it's something I know he tries from time to time.

Eh2406:
Maybe next time he or someone else tries it, they can leave a link hear. It would be a great demo/blog of how to do a migration, if nothing else.

sgrif:
Last I heard everyone on the team was open to it, it's just missing a PR. I had planned to lay out a "we can do this in small chunks" PR some time soon, but my daughter came earlier than expected and I haven't had time. If nothing else, I think the biggest step would be to confirm that we would be interested in doing this, and lay out the groundwork required to move one module or one method over at a time. I don't think anybody wants to review a "port the whole project over to Diesel" PR in one go.

carols10cents:
I am interested in doing this! And yeah, incremental would be awesome.

sgrif:
Ok. I'll open a PR.

carols10cents:
There are PRs out for the rest of this, I just need to review them!

sgrif:
I think we can safely close this issue though at this point. It's been thoroughly considered. :D

PR #589, opened 2017-03-04
Set up the groundwork for a port to Diesel

sgrif:
https://github.com/rust-lang/crates.io/issues/588#issuecomment-284155257

This PR changes the `users#show` endpoint to use Diesel instead of
rust-postgres, (and adds a test for that endpoint). I chose that
endpoint as it seemed likely to touch as few things as possible, but
enough to require setting up Diesel connection pooling, and testing.

I have also ported over the migrations to use Diesel's migration
infrastructure. The migration files (except the final migration) were
generated programatically. Any non-reversible migrations were not dumped
as these contained procedural data updates which aren't relevant to
setting up a new database.

`cargo run --bin migrate` will move all the entries in the old
migrations table into the one used by Diesel. That function is brittle
since it relies on Diesel internals. However, it really only matters for
existing contributors (and one deploy to Heroku), so when that function
breaks we can just delete it.

I've added an additional migration to make the schema compatible with
`infer_schema!`, which doesn't support tables that have no primary key.
I'm not using `infer_schema!` just yet, as it doesn't work with non-core
types and there are some tsvector columns.

I re-ordered the columns on the `User` struct, as diesel emits an
explicit select clause and then fetches columns by index rather than by
name. This means that the order has to match how it was defined to
Diesel (which in the case of `infer_schema!` will be the database
definition order). If we don't want this restriction, we can replace the
`infer_schema!` call with manual `table!` calls (which can be
automatically generated by `diesel print-schema`), and have the columns
ordered there to match the struct.

Differences to note
-------------------

The rust-postgres connection is set to require TLS when the env var
`HEROKU` is set. In Diesel, this would be to add `?sslmode=require` onto
the database URL. I'm unsure if heroku-postgres URLs already have this
or not, so I have not added any explicit code for that. It's a question
that should be answered before this is deployed.

Additionally, I chose not to automatically wrap each request in a
transaction. It's unneccessary most of the time, and seems like it was
only done to make testing easier. Since Diesel has explicit support for
"run tests in a transaction that never commits", we don't need that
here.


sgrif:
So just glancing through where the codebase is at now, a few notes in terms of feature support.

- Diesel 0.11 has an API for the cross join used when the `q` parameter is present in `crates#index`, but it's going away in 0.12, and I'm not sure I'll have the replacement in that version
- There's a lot of many-to-many associations. We don't *not* support them, it's just a few lines of boilerplate instead of something like `#[belongs_to(Crate, through="crates_categories")]`
- [Whatever is going on here](https://github.com/rust-lang/crates.io/blob/61274681aa0e9c0490290f24299a07c3093753ee/src/category.rs#L161-L166) isn't something we can write one-to-one with the query builder. We may be able to restructure the query to something that can be written in the query builder.
- [OMG CAROL STAHP](https://github.com/rust-lang/crates.io/blob/61274681aa0e9c0490290f24299a07c3093753ee/src/category.rs#L185-L190)

None of this would block finishing the port, but we will have a few places where we'll have to fall back to raw SQL. And I probably have 2 more releases of Diesel worth of work to do to support all this. :P

carols10cents:
I'm fine with having to fall back to raw sql sometimes. THEY'RE GOOD QUERIES, SGROF

carols10cents:
> Diesel 0.11 has an API for the cross join used when the q parameter is present in crates#index, but it's going away in 0.12, and I'm not sure I'll have the replacement in that version

but but https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L497

I'd like to refactor that part, one because it could use some more structure and two to eventually support things like [searching with a query within a keyword](https://github.com/rust-lang/crates.io/issues/491). I'll think about how we could rework this.

sgrif:
Yes, everything except the `q` branch factors out nicely.

carols10cents:
> In Diesel, this would be to add ?sslmode=require onto
> the database URL. I'm unsure if heroku-postgres URLs already have this
> or not, so I have not added any explicit code for that.

The DATABASE_URL env var set by heroku does not have `?sslmode=require`, and from what I can tell, we shouldn't need it when querying the database from within heroku. [Their docs say to add `?sslmode=require` in application code](https://devcenter.heroku.com/articles/connecting-to-heroku-postgres-databases-from-outside-of-heroku#ssl), which sounds like what you're saying. We have a staging heroku instance where we can try this out, although we had some issue with the prod instance that we didn't have with staging..... i forget exactly what though.... 



> Additionally, I chose not to automatically wrap each request in a
> transaction. It's unneccessary most of the time, and seems like it was
> only done to make testing easier. Since Diesel has explicit support for
> "run tests in a transaction that never commits", we don't need that
> here.

There's at least one place I know crates.io depends on this-- [when uploading a new version, if the git updates or s3 uploads fail, we roll back the database additions](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L816-L818). By "it's unnecessary most of the time", do you mean that Diesel has a way to wrap those requests where it is necessary in a transaction?

@alexcrichton, do you know of any other places besides new crate and tests where we're relying on each request being wrapped in a transaction?

sgrif:
Yes. We can wrap in a transaction where it's needed (this PR does that in `NewUser#create_or_update`).

sgrif:
I've updated to require SSL on Heroku.

alexcrichton:
> @alexcrichton, do you know of any other places besides new crate and tests where we're relying on each request being wrapped in a transaction?

Hm. Well so I wrote the whole web app on this assumption, so we may want to leave it for now :). In general there's lots of transactional points such as:

* Crate is inserted into db
* Crate is uploaded to S3
* Crate is uploaded to github

We currently rely on database transactions so if github fails the insertion is rolled back. Same for tons of API calls and such, the database is updated willy-nilly in the assumption that a failure down the line will roll back updates.


---

Also yeah I believe SSL is required on Heroku nowadays (I found it hard to find docs on that...). The ironic part is that although SSL is required they're not using valid certificates, so it's still MitM happy all over the place. I emailed them awhile back about this and they said yeah that's what most database drivers do, ignore certs. ;_;

sgrif:
To be clear, I have no issue with wrapping everything in a transaction, I just would rather do it in the requests that perform updates than around every request implicitly. Since our transactions are lexically scoped, it's difficult to establish lazily if we need to wrap each request.

sgrif:
Returning an error would still rollback the transaction, we just wouldn't wrap it in endpoints like `crates#summary` which perform only reads.

alexcrichton:
Ah yeah so if errors still roll back transactions then that's fine by me, it was the only intention of the setup. (I thought that was standard "write a web app" practice, right?)

Unfortunately crates.io doesn't have a lexical scope for the transaction, you can see that with the bit of unsafe code around managing the postgres transaction (I think it's still there at least...)

Also FWIW one speed bump I hit with Heroku + Diesel in the past was that Diesel needed a db connection at compile-time but w/ Heroku you don't have that (only at runtime). Is that `#[derive]` just not needed though?

sgrif:
> I thought that was standard "write a web app" practice, right?

Errors rolling back definitely is. I haven't seen "wrap all requests in a transaction" too often though.

> Unfortunately crates.io doesn't have a lexical scope for the transaction, you can see that with the bit of unsafe code around managing the postgres transaction (I think it's still there at least...)

We can do it in a middleware, but it would require eagerly acquiring the database connection. And yeah, we'd have to do something similar to what you're doing with the postgres transaction (and rely on Diesel internals) to avoid that.

> Also FWIW one speed bump I hit with Heroku + Diesel in the past was that Diesel needed a db connection at compile-time but w/ Heroku you don't have that (only at runtime)

`infer_schema!` is the only thing that needs a database connection. We have a CLI command that you can use instead (`diesel print-schema`) -- I will update this to not require a connection at compile time.

sgrif:
Also I thought you could get a DB connection at build time with Heroku now? Maybe I'm mistaken

sgrif:
Asked my friend who works at Heroku.

> db connections are available at build time, however if it's the first push for a rails app a database may not be provisioned yet

sgrif:
@alexcrichton So we can use `infer_schema!` if we want, or we can choose not to. I don't have a strong opinion one way or the other. Not using it just means there's one file that has to get updated when the schema changes (even if it's generated with a command)

alexcrichton:
Oh interesting, how do you roll back without a transaction? (I thought transactions were how that was done).

Acquiring a db connection eagerly is probably ok, we can always investigate other routes if it turns out to be a problem.

Ah yeah and that makes sense (first deploy not having a db). I'm fine either way w/ crates.io, but it does mean that redeployment may be difficult if we ever try to do it...

sgrif:
> Oh interesting, how do you roll back without a transaction? (I thought transactions were how that was done).

Sorry to be more clear, you do use a transaction, just one that's explicit (or implicit at a lower level. e.g. Rails wraps every `.save` call in a transaction, so anything that happens in callbacks is already wrapped. If you are doing `foo.save && bar.save` in a controller action, you would want to wrap that in `ActiveRecord::Base.transaction do`)

> Acquiring a db connection eagerly is probably ok, we can always investigate other routes if it turns out to be a problem.

It was a problem in tests, but I can dig in and see if I can figure out why.

> Ah yeah and that makes sense (first deploy not having a db). I'm fine either way w/ crates.io, but it does mean that redeployment may be difficult if we ever try to do it...

I'll go with the manual file for now, since getting that `TsVector` column to generate properly from `infer_schema` requires a feature that's in 0.12.

sgrif:
I've rebased and updated to not connect to the database at compile time.

alexcrichton:
👍

sgrif:
Ah, I've figured out what the issue with tests was. I'll go ahead and update this to retain the auto transaction wrapping for now. I'll continue to explicitly wrap anything that looks like it should be in a transaction though, so we can easily go to a lazy connection if we decide it's necessary.

sgrif:
Actually I think that's best left for a separate PR, since this one is quite large already

carols10cents:
> cargo run --bin migrate will move all the entries in the old
> migrations table into the one used by Diesel. That function is brittle
> since it relies on Diesel internals. However, it really only matters for
> existing contributors (and one deploy to Heroku), so when that function
> breaks we can just delete it.

So I ran `cargo run --bin migrate` locally and it just says `The `migrate` binary is no longer used. Use `diesel migration run` and `diesel migration revert` instead.`... is that what you expected to happen?

Also could you update the README with new setup instructions? Based on the travis changes, it looks like people need to `cargo install diesel_cli` and then run `diesel migration run` instead of migrate?

carols10cents:
Looks like it needs to be `cargo install diesel_cli --debug --no-default-features --features postgres`, `cargo install diesel_cli` failed for me because of something about mysql

sgrif:
Yeah, by default it installs with all the supported backends so if you don't have mysql and sqlite installed you'll want `--no-default-features --features postgres`.

sgrif:
> is that what you expected to happen?

Yes. It also migrated your old migrations table over to the Diesel one.

carols10cents:
oops, found another spot-- could you update the Procfile too?

src/db.rs
```diff
@@ -16,13 +17,17 @@ use r2d2_postgres::PostgresConnectionManager as PCM;
...
+use url::Url;
 
 use app::{App, RequestApp};
 use util::{CargoResult, LazyCell, internal};
 
 pub type Pool = r2d2::Pool<PCM>;
 pub type Config = r2d2::Config<pg::Connection, r2d2_postgres::Error>;
 type PooledConnnection = r2d2::PooledConnection<PCM>;
+pub type DieselPool = r2d2::Pool<ConnectionManager<PgConnection>>;
+type DieselPooledConn = r2d2::PooledConnection<ConnectionManager<PgConnection>>;
```
carols10cents: Nice!

src/schema.rs
```diff
@@ -0,0 +1,171 @@
+// This file can be regenerated with `diesel print-schema`
```
carols10cents: YAYYYY A SCHEMA!!!!!!

src/user/mod.rs
```diff
@@ -3,34 +3,94 @@ use std::collections::HashMap;
...
-    pub gh_login: String,
-    pub gh_id: i32,
-    pub name: Option<String>,
     pub email: Option<String>,
-    pub avatar: Option<String>,
     pub gh_access_token: String,
     pub api_token: String,
+    pub gh_login: String,
+    pub name: Option<String>,
+    pub avatar: Option<String>,
```
carols10cents: Wait, `Queryable` cares about the order of the fields on `User` but it doesn't care that the field here is named `avatar` but the field in the db is named `gh_avatar`?

carols10cents:
Seems to work locally, just have that one question about `gh_avatar` tho... but let's try deploying to staging!!

src/user/mod.rs
sgrif: Correct. We will generate an explicit `select` clause so we know the order that the fields come back in, and then access them by order. This is because we can really only represent rows as a tuple, we can't express "has a field called `name` that is of this type" in the type system. We also don't assume that your `Queryable` structs are one-to-one with a database table. That said, I'm looking at ways to optionally allow by-name instead of by-index.

carols10cents:
wait how does heroku know it needs diesel-cli?

sgrif:
It doesn't. I'm making a buildpack now.

carols10cents:
lolololol

.travis.yml
```diff
@@ -42,6 +43,7 @@ addons:
 
 env:
   global:
+    - DATABASE_URL=postgres://postgres:@localhost/cargo_registry_test
     - TEST_DATABASE_URL=postgres://postgres:@localhost/cargo_registry_test
```
carols10cents: oh so if travis is using DATABASE_URL now, do we need to set TEST_DATABASE_URL anymore?

sgrif: `DATABASE_URL` is just used by diesel for migrations by convention. The tests still run using `TEST_DATABASE_URL`

src/user/mod.rs
carols10cents: Ah, ok, gotcha.

.travis.yml
sgrif: I was also using `DATABASE_URL` for compilation before, but since we don't need that now I can just specify `--database-url` instead

carols10cents: You mean just for `diesel database setup`, not migrations, right? because it looks like  `src/tests/all.rs` takes care of the migrations and uses TEST_DATABASE_URL? I just want to make sure I understand :)

sgrif: Yes. `diesel database setup` runs migrations as well.

carols10cents: ah i see

carols10cents:
omg it's working on staging, hold onto your butts

PR #590, opened 2017-03-05
Implement request extensions on `T: Request` not `Request`

sgrif:
There are a bunch of tests which are explicitly casting `&MockRequest`
into a `&Request` trait object, because the extension methods provided
are only on the trait object, not anything which implements the trait.

By changing from `Request + 'a` to `T: Request + ?Sized` (which is more
ideomatic anyway), we can just call the method directly on `MockRequest`
in tests.

The deletion of the `#[macro_use]` line was to get the tests to compile locally. It's an unused annotation which emits a warning on my machine.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -101,7 +101,7 @@ pub trait RequestApp {
     fn app(&self) -> &Arc<App>;
 }
 
-impl<'a> RequestApp for Request + 'a {
+impl<T: Request + ?Sized> RequestApp for T {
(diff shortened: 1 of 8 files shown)


alexcrichton:
Heh did I mention that crates.io originates from *ancient* Rust? Like... really ancient rust...

alexcrichton:
(Thanks!)

sgrif:
It doesn't show at all :trollface:

PR #591, opened 2017-03-05
canonicalize crate name before checking if it's reserved

sgrif:
Currently reserved crate names come from a text file in the source.
There are only two entries which consider canonicalization (e.g. the
list contains `compiler-rt` and `compiler_rt`). The fact that the name
is not canonicalized allows crates like https://crates.io/crates/std to
be uploaded. While this definitely isn't a vulnerability, since the
registry does *not* canonicalize the name, we should still disallow
crates to be uploaded which appear at the same path as a reserved name,
and would prevent a crate with a reserved name from being uploaded
later.

I've moved the actual data out of the text file, and into the database,
so we can have a constraint which verifies this at the closest possible
location. (We can't use an actual check constraint, as that disallows
subselects).

The crate referenced above should be manually deleted.


Eh2406:
This also removes the code that was scanning the list linearly. +1

alexcrichton:
FWIW this list gets updates from time to time, so having it relatively easy to modify would be nice. This'd just require a migration per modification, right?

sgrif:
Yes. There's no way to provide a database level constraint that doesn't require a migration per modification.

carols10cents:
Seems fine 👍

PR #592, opened 2017-03-05
Remove the `max_version` column from Crates

sgrif:
Updating columns like this tends to be pretty brittle unless it's done
via a trigger. We can't update this one from a trigger, since PG doesn't
know about semver's comparison rules. Either way, we don't really need
to cache this unless it turns out to be a performance bottleneck. We can
instead calculate the value on the fly.

This does introduce some N+1 queries, and those may end up being a
hotspot. We can definitely get rid of them, but it's a pain in the ass
to do manually and Diesel has support for it built in, so I'd rather
just fix them as we move those endpoints over.

### DIFF
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -29,7 +29,7 @@ export default Ember.Route.extend({
         return crate.get('versions')
             .then(versions => {
                 const version = versions.find(version => version.get('num') === params.version_num);
-                if (!version) {
+                if (params.version_num && !version) {
                     this.controllerFor('application').set('nextFlashError',
                         `Version '${params.version_num}' of crate '${crate.get('name')}' does not exist`);
                 }
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -11,7 +11,7 @@ export default Ember.Route.extend({
         const maxVersion = crate.get('max_version');
 
         // Fall back to the crate's `max_version` property
-        if (!requestedVersion) {
+        if (!requestedVersion && maxVersion !== '0.0.0') {
             params.version_num = maxVersion;
         }
 
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -868,19 +868,6 @@ fn migrations() -> Vec<Migration> {
             tx.execute("DROP TABLE reserved_crate_names", &[])?;
             Ok(())
         }),
-        Migration::new(20170305123234, |tx| {
-            tx.execute("ALTER TABLE crates DROP COLUMN max_version", &[])?;
-            Ok(())
-        }, |tx| {
-            tx.execute("ALTER TABLE crates ADD COLUMN max_version VARCHAR NOT NULL DEFAULT '0.0.0'", &[])?;
-            tx.execute("UPDATE crates SET max_version = COALESCE((
-                SELECT num FROM VERSIONS
-                 WHERE num SIMILAR TO '[0-9]+\\.[0-9]+\\.[0-9]+'
-              ORDER BY split_part(num, '.', 1)::int, split_part(num, '.', 2)::int, split_part(num, '.', 3)::int
-              LIMIT 1
-            ), '0.0.0')", &[])?;
-            Ok(())
-        }),
     ];
     // NOTE: Generate a new id via `date +"%Y%m%d%H%M%S"`
 
--- a/src/bin/migrate.rs
+++ b/src/bin/migrate.rs
@@ -868,6 +868,12 @@ fn migrations() -> Vec<Migration> {
             tx.execute("DROP TABLE reserved_crate_names", &[])?;
             Ok(())
         }),
+        Migration::new(20170305123234, |tx| {
+            tx.execute("ALTER TABLE crates DROP COLUMN max_version", &[])?;
+            Ok(())
+        }, |_| {
+            panic!("Unreversible migration")
+        }),
     ];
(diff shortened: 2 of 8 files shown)


carols10cents:
@alexcrichton I found [this commit that adds the max_version column](https://github.com/rust-lang/crates.io/commit/b2315e4e6d608374caf6501ec8178d8e6e95d6eb), and [this commit that adds the tracking of the max_version](https://github.com/rust-lang/crates.io/commit/0ff222039adc26af26681b8a6cfae437db7830d3), but I don't see any explanation in there of why this was needed.... were there performance problems? complex queries?

Eh2406:
cc #510 Redirecting to the latest crate should only use non-yanked versions and 
#468 ... not -alpha nor -beta alternative fix in https://github.com/rust-lang/crates.io/pull/577

alexcrichton:
@carols10cents I think it was a "too clever by half" attempt to make these operations fast:

* Rendering a crate on the home page (doesn't require hitting versions table)
* Easily rendering a crate on the crate page (w/o a version in the url)

crates.io has always operated blazingly fast though so I don't think we really need to be optimizing sql queries :)

Looks great to me!

src/bin/delete-crate.rs
```diff
@@ -7,7 +7,6 @@
 
 #![deny(warnings)]
 
-#[macro_use]
```
carols10cents: sean.

carols10cents:
Ok, this looks good, and I'm all for it. The irreversible migration makes me nervous though, and I know this is basically just a cached column and we're not really losing any data blah blah blah but it would make me feel better if I deployed the query change without the migration at first to just make sure.

And I want to cherry-pick the tests from #582 and run them with this code. And then deploy when I'm awake and alert and can watch the stats and be ready to roll back, so tomorrow morning perhaps. I know I'm probably being overdramatic, I miss the adrenaline of running a Big Production App ok???

src/bin/delete-crate.rs
sgrif: It literally doesn't compile if I don't remove it!

sgrif:
So should I remove the migration? Or do you want me to leave it in?

sgrif:
I've rebased at least so the `#[macro_use]` is gone. ;)

carols10cents:
Oh it's no problem for me to split it up... which I'll have to do again since I did it right before you rebased ;) ;) ;) <3

sgrif:
Lol sorry. I was going to cherry-pick the tests for you, too.

carols10cents:
Don't you have a 👶 to take care of or something? :) <3

sgrif:
She's sleeping. :P

carols10cents:
Actually, ok, so I think the commits that add tests we should keep from #582 are:

365077fec132dc2320843ac6cf389576b47b9c0b
1292f7f1c291f71a13f51face15da052a6c64f21

Then there's this commit: 575829e12fede16d3e71e9c451b3d5704015daea

That PR changes max_version to be an `Option`, while yours sets the default to 0.0.0. I think 575829e12fede16d3e71e9c451b3d5704015daea fixes #502, while I'm not sure what effect your pr has (but i'm about to try). WDYT?

sgrif:
I chose 0.0.0 because it was the default value in the database. We could certainly change it to be `NULL` instead. Not sure what the behavior of #502 would be. I can certainly change it to return an option if that's preferable.

I pulled in the tests from that commit (I can't pull that commit directly, but I credited @yodaldevoid as the author in the commit which pulled the tests)

carols10cents:
Yeah ok so with your change, this is what happens on a crate page when all versions have been yanked:

screen shot 2017-03-07 at 9 59 35 am

#582 doesn't fix #502, but at least it doesn't have an error because of 575829e.

With 575829e and another tweak that I just pushed, your change doesn't show the flash error either. Wdyt?

Both PRs fix #510.

carols10cents:
oops hm not sure where I pushed it but it's not here

sgrif:
Sorry that was probably me. I think I got travis green at least. I went ahead and wrote an *almost correct* down migration as well since that was a concern.

carols10cents:
there we go. no i pushed it to a new branch instead of to your repo.

sgrif:
Yeah, your fix looks good. I think I might have to futz with the line endings of that file one more time but we'll see what travis says.

carols10cents:
Ok i'm going to try deploying this to staging while we're waiting on travis

sgrif:
Ok I have no clue how to get that file in a format that travis likes. Let me see if I can cherry pick it from that branch and not have git "fix" the line endings

carols10cents:
lol oops merge conflict

sgrif:
Ok rebased, squashed the line ending commits, and cherry picked the files in exactly the form they were in from #582 so let's see if travis is happy now

sgrif:
Now I must feed :baby:

carols10cents:
uhhhhh soooooo the deploy to staging went ok, the 95th percentile response time is a bit higher than prod but still way lower than the median of most rails apps, buuuut i tried out the down migration and it is not "pretty darn close".... it picks the first version number it sees out of all crates and sets all crates' max versions to that... look i'm just going to take out the migration and we'll do it in a separate deploy

sgrif:
Oh I forgot to make the order `DESC`

carols10cents:
uhhh and join crates to versions on crate.id

sgrif:
Oh yeah that too

carols10cents:
WE'VE GOT GREEN GO GO GO

PR #594, opened 2017-03-06
Simplify and optimize `Categories::toplevel`

sgrif:
`EXPLAIN ANALYZE` output (tested assuming `ORDER BY crates_cnt DESC LIMIT 10 OFFSET 0`)

Before:

```
(cost=7129.79..7129.82 rows=10 width=127) (actual time=1.184..1.187 rows=10 loops=1)
```

After:

```
(cost=5.88..5.88 rows=1 width=131) (actual time=0.170..0.171 rows=10 loops=1)
```

About half of the performance comes from removing the `COALESCE`. Since
the subselect is including the `crates_cnt` from the toplevel category
(which is why it doesn't have to add `c.crates_cnt`), so it can never
return null.

The second big win is changing from a subselect to a join. PG is usually
quite good at figuring out when these cases are equivalent, but I
suspect that the use of an aggregate function in the subselect means
that it will actually have to subselect in a loop.

Finally, we avoid using `LIKE`, since it's more expensive than we need,
and can't be indexed. I've opted to use `split_part(slug)` in both the
join and outer filter, so that both can be covered by a single index
later. The "cheapest" way to do the outer filter is probably
`strpos(slug, '::') = 0`, but the difference is so small that it doesn't
matter.

I explicitly did not include an index here, since the data set is small
enough that it would never be used. If the number of categories grows
beyond a few hundred, this query can also benefit from an index on
`split_part(slug, '::', 1)`.

The test coverage around this method was pretty light, so I've added
some unit tests to give it a bit more coverage for correctness.

### DIFF
--- a/src/category.rs
+++ b/src/category.rs
@@ -157,16 +157,13 @@ impl Category {
         // Collect all the top-level categories and sum up the crates_cnt of
         // the crates in all subcategories
         let stmt = conn.prepare(&format!(
-            "SELECT c.id, c.category, c.slug, c.description, c.created_at, \
-                COALESCE (( \
-                    SELECT sum(c2.crates_cnt)::int \
-                    FROM categories as c2 \
-                    WHERE c2.slug = c.slug \
-                    OR c2.slug LIKE c.slug || '::%' \
-                ), 0) as crates_cnt \
-             FROM categories as c \
-             WHERE c.category NOT LIKE '%::%' {} \
-             LIMIT $1 OFFSET $2",
+            "SELECT c.id, c.category, c.slug, c.description, c.created_at,
+                sum(c2.crates_cnt)::int as crates_cnt
+             FROM categories as c
+             INNER JOIN categories c2 ON split_part(c2.slug, '::', 1) = c.slug
+             WHERE split_part(c.slug, '::', 1) = c.slug
+             GROUP BY c.id
+             {} LIMIT $1 OFFSET $2",
             sort_sql
         ))?;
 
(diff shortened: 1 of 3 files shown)


src/category.rs
```diff
@@ -157,16 +157,13 @@ impl Category {
...
-                    OR c2.slug LIKE c.slug || '::%' \
-                ), 0) as crates_cnt \
-             FROM categories as c \
-             WHERE c.category NOT LIKE '%::%' {} \
-             LIMIT $1 OFFSET $2",
+            "SELECT c.id, c.category, c.slug, c.description, c.created_at,
+                sum(c2.crates_cnt)::int as crates_cnt
+             FROM categories as c
+             INNER JOIN categories c2 ON split_part(c2.slug, '::', 1) = c.slug
+             WHERE split_part(c.slug, '::', 1) = c.slug
```
carols10cents: Cool!!

src/category.rs
```diff
@@ -278,3 +275,110 @@ pub fn slugs(req: &mut Request) -> CargoResult<Response> {
...
+    use pg::{Connection, TlsMode};
+    use dotenv::dotenv;
+    use std::env;
+
+    fn pg_connection() -> Connection {
+        let _ = dotenv();
+        let database_url = env::var("TEST_DATABASE_URL")
+            .expect("TEST_DATABASE_URL must be set to run tests");
+        let conn = Connection::connect(database_url, TlsMode::None).unwrap();
+        // These tests deadlock if run concurrently
```
carols10cents: 💩

src/category.rs
```diff
@@ -278,3 +275,110 @@ pub fn slugs(req: &mut Request) -> CargoResult<Response> {
     struct R { category_slugs: Vec<Slug> }
     Ok(req.json(&R { category_slugs: slugs }))
 }
+
+#[cfg(test)]
+mod tests {
```
carols10cents: Yay more tests!!! 😻 😻 😻 😻

Issue #163, opened 2015-06-26
404 responce for HEAD request with no Accept: 


nico-izo:
Steps to reproduce:

> $ curl -I https://crates.io/
> HTTP/1.1 404 Not Found
> <...>
> Content-Type: application/json; charset=utf-8

With Accept: text/html everything is okay:

> $ curl -H "Accept: text/html" -I https://crates.io/
> HTTP/1.1 200 OK
> <...>
> Content-Type: text/html

alexcrichton:
Unfortunately had to revert https://github.com/rust-lang/crates.io/pull/464

tamird:
Seems like we need separate treatment of headers by route.

alexcrichton:
Yeah I'd be down for that!

PR #597, opened 2017-03-06
Add security headers

### LINKED ISSUE #586: Implement Security Headers
_Related to #585_ 

As asked by @carols10cents, here's a tasklist on what headers should be implemented.

- [ ] Implement Content Security Policy (CSP) header

- [x] Implement HTTP Strict Transport Security (HSTS) header

- [ ] Implement X-Content-Type-Options header 

- [ ] Implement X-Frame-Options (XFO) header

- [ ] Implement X-XSS-Protection header

sgrif:
These headers are all pretty straightforward except CSP. For CSP I
defined the sources based on what was loaded from visiting the main page
and all crates. Images should be safe, so I've allowed them from all
sources.

This should be checked on staging before deploying.

Fixes #586.

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -107,7 +107,7 @@ impl Middleware for SecurityHeadersMiddleware {
                     "default-src 'self'; \
                       connect-src 'self' https://docs.rs; \
                       script-src 'self' https://www.google-analytics.com https://www.google.com; \
-                      style-src 'self' https://www.google.com; \
+                      style-src 'self' https://www.google.com https://ajax.googleapis.com; \
                       img-src *; \
                       object-src 'none'"
                         .into(),
--- a/src/http.rs
+++ b/src/http.rs
@@ -105,6 +105,7 @@ impl Middleware for SecurityHeadersMiddleware {
                 "Content-Security-Policy".into(),
                 vec![
                     "default-src 'self'; \
+                      connect-src 'self' https://docs.rs; \
                       script-src 'self' https://www.google-analytics.com https://www.google.com; \
                       style-src 'self' https://www.google.com; \
                       img-src *; \
--- a/src/http.rs
+++ b/src/http.rs
@@ -91,6 +91,7 @@ pub fn token(token: String) -> Token {
     }
 }
 
+#[derive(Clone, Copy, Debug)]
 pub struct SecurityHeadersMiddleware;
 
 impl Middleware for SecurityHeadersMiddleware {
--- a/src/http.rs
+++ b/src/http.rs
@@ -1,3 +1,5 @@
+use conduit::{Request, Response};
+use conduit_middleware::Middleware;
 use curl;
 use curl::easy::{Easy, List};
 use oauth2::*;
@@ -6,7 +8,7 @@ use util::{CargoResult, internal, ChainError, human};
 use serde_json;
 use serde::Deserialize;
 use std::str;
-
+use std::error::Error;
 
 /// Does all the nonsense for sending a GET to Github. Doesn't handle parsing
 /// because custom error-code handling may be desirable. Use
@@ -86,5 +88,40 @@ pub fn token(token: String) -> Token {
         access_token: token,
         scopes: Vec::new(),
         token_type: String::new(),
+    }
+}
+
+pub struct SecurityHeadersMiddleware;
+
+impl Middleware for SecurityHeadersMiddleware {
+    fn after(&self, _: &mut Request, mut res: Result<Response, Box<Error+Send>>)
+        -> Result<Response, Box<Error+Send>> {
+        if let Ok(ref mut response) = res {
+            response.headers.insert(
+                "Content-Security-Policy".into(),
+                vec!["default-src 'self'; \
+                      script-src 'self' https://www.google-analytics.com https://www.google.com; \
+                      style-src 'self' https://www.google.com; \
+                      img-src *; \
+                      object-src 'none'".into()],
+            );
+            response.headers.insert(
+                "X-Content-Type-Options".into(),
+                vec!["nosniff".into()],
+            );
+            response.headers.insert(
+                "X-Frame-Options".into(),
+                vec!["SAMEORIGIN".into()],
+            );
+            response.headers.insert(
+                "X-XSS-Protection".into(),
+                vec!["1; mode=block".into()],
+            );
+            response.headers.insert(
+                "Strict-Transport-Security".into(),
+                vec!["max-age=31536000; includeSubDomains".into()],
+            );
+        }
+        res
     }
 }
--- a/src/http.rs
+++ b/src/http.rs
@@ -101,6 +101,11 @@ impl Middleware for SecurityHeadersMiddleware {
         mut res: Result<Response, Box<Error + Send>>,
     ) -> Result<Response, Box<Error + Send>> {
         if let Ok(ref mut response) = res {
+            // It would be better if we didn't have to have 'unsafe-eval' in the `script-src`
+            // policy, but google charts (used for the download graph on crate pages) uses `eval`
+            // to load scripts. Remove 'unsafe-eval' if google fixes the issue:
+            // https://github.com/google/google-visualization-issues/issues/1356
+            // or if we switch to a different graph generation library.
             response.headers.insert(
                 "Content-Security-Policy".into(),
                 vec![
--- a/src/http.rs
+++ b/src/http.rs
@@ -117,10 +117,6 @@ impl Middleware for SecurityHeadersMiddleware {
                 "X-XSS-Protection".into(),
                 vec!["1; mode=block".into()],
             );
-            response.headers.insert(
-                "Strict-Transport-Security".into(),
-                vec!["max-age=31536000; includeSubDomains".into()],
-            );
         }
         res
     }
--- a/src/http.rs
+++ b/src/http.rs
@@ -106,7 +106,8 @@ impl Middleware for SecurityHeadersMiddleware {
                 vec![
                     "default-src 'self'; \
                       connect-src 'self' https://docs.rs; \
-                      script-src 'self' https://www.google-analytics.com https://www.google.com; \
+                      script-src 'self' 'unsafe-eval' \
+                                 https://www.google-analytics.com https://www.google.com; \
                       style-src 'self' https://www.google.com https://ajax.googleapis.com; \
                       img-src *; \
                       object-src 'none'"
--- a/src/http.rs
+++ b/src/http.rs
@@ -95,16 +95,22 @@ pub fn token(token: String) -> Token {
 pub struct SecurityHeadersMiddleware;
 
 impl Middleware for SecurityHeadersMiddleware {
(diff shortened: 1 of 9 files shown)


carols10cents:
I have this up on staging now, the deploy went fine, the [observatory now gives staging an A+](https://observatory.mozilla.org/analyze.html?host=staging-crates-io.herokuapp.com), installing a crate via cargo from staging worked fine. Anything else we should check?

sgrif:
Not that I can think of.

sgrif:
Actually I'm seeing a console error when I run it locally. It looks like google analytics is trying to use unsafe eval.

g-k:
👍 Looks good.

Putting the inline script in its own file with the [async ga analytics.js](https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet) should workaround the error.

```diff
diff --git a/app/index.html b/app/index.html
index 3de7338..a4ebf8c 100644
--- a/app/index.html
+++ b/app/index.html
@@ -7,15 +7,8 @@

     {{content-for 'head'}}

-    <script>
-      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-      })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
-
-      ga('create', 'UA-58390457-3', 'auto');
-      ga('send', 'pageview');
-    </script>
+    <script src="{{rootURL}}assets/ga-pageview.js"></script>
+    <script async src="https://www.google-analytics.com/analytics.js"></script>

     <link rel="stylesheet" href="{{rootURL}}assets/vendor.css">
     <link rel="stylesheet" href="{{rootURL}}assets/cargo.css">
diff --git a/dist/assets/ga-pageview.js b/dist/assets/ga-pageview.js
new file mode 100644
index 0000000..68f8501
--- /dev/null
+++ b/dist/assets/ga-pageview.js
@@ -0,0 +1,3 @@
+window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
+ga('create', 'UA-58390457-3', 'auto');
+ga('send', 'pageview');
```

bluetech:
Just a note that you should be careful with `Strict-Transport-Security`; I don't know if someone runs `crates.io` privately, but if they do, and expose it to users, then setting HSTS as done here is a commitment to only use HTTPS for 1 year for the domain and all subdomains.

You might want to make this one opt-in, unless this is really not a concern.

This should not be taken as being against HSTS - `https://crates.io` should definitely use it (and looks like it already does, though without `includeSubdomains`).

carols10cents:
> Just a note that you should be careful with Strict-Transport-Security; I don't know if someone runs crates.io privately, but if they do, and expose it to users, then setting HSTS as done here is a commitment to only use HTTPS for 1 year for the domain and all subdomains.
> 
> You might want to make this one opt-in, unless this is really not a concern.

I don't feel like I have enough knowledge in this area to make this call :-/ I'll try to do some research around this soon.

g-k:
+1 on rolling out HSTS slowly as mentioned [here](https://github.com/rust-lang/crates.io/pull/597#issuecomment-302346596).

I don't remember the exact value, but services usually start with a low HSTS max-age (minutes) and make sure everything keeps working for awhile before bumping it to a year (like DNS changes).

vignesh-sankaran:
What's the status on this pull request? What needs to be done to unblock this?

carols10cents:
> What's the status on this pull request? What needs to be done to unblock this?

I think the ember folks changed google analytics recently, I'm not sure if it's now async or not or whatever we need google analytics to be-- that needs to be researched and changed and tested if necessary.

I need to read up on HSTS and understand the implications before deploying-- that's unfortunately not something you can unblock for me :)

carols10cents:
Ok reading a bit about HSTS on these sites:

- https://www.troyhunt.com/understanding-http-strict-transport/
- https://hstspreload.org/

it sounds like until we get https://doc.crates.io working (see https://github.com/rust-lang/crates.io/issues/621), we definitely should NOT be sending `includeSubdomains`, which means we're not eligible to be `preload` either.

> This should not be taken as being against HSTS - https://crates.io should definitely use it (and looks like it already does, though without includeSubdomains).

Indeed it is! Running this:

```
curl -v 'https://crates.io/' -H 'Accept: text/html' 
```

shows this in the response headers:

```
< Strict-Transport-Security: max-age=31536000
```

Looks like this is happening because we use the nginx buildpack and we've [specified that in config/nginx.conf.erb](https://github.com/rust-lang/crates.io/blob/master/config/nginx.conf.erb#L39). People deploying crates.io on their own servers may not be using that file at all, and if they do, they can choose to remove or change that line.

So I don't think we need to be setting the STS headers in the spot that this PR does.

HOWEVER, mozilla observatory is not seeing this. In their "raw server headers", I see "Content-Type | application/json; charset=utf-8", which means they're not requesting a content type, and crates.io defaults to JSON which returns a 404 for /:

```
$ curl -v https://crates.io
* Rebuilt URL to: https://crates.io/
*   Trying 50.16.231.196...
* TCP_NODELAY set
* Connected to crates.io (50.16.231.196) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate: crates.io
* Server certificate: Let's Encrypt Authority X3
* Server certificate: DST Root CA X3
> GET / HTTP/1.1
> Host: crates.io
> User-Agent: curl/7.51.0
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Connection: keep-alive
< Server: nginx
< Date: Tue, 11 Jul 2017 20:45:58 GMT
< Content-Type: application/json; charset=utf-8
< Content-Length: 35
< Set-Cookie: cargo_session=--3Mz45cH3itfVkmY41+nVjWRm9+0=; HttpOnly; Secure; Path=/
< Via: 1.1 vegur
<
* Curl_http_done: called premature == 0
* Connection #0 to host crates.io left intact
{"errors":[{"detail":"Not Found"}]}
```

Soo it looks like we need to solve https://github.com/rust-lang/crates.io/issues/163 before we can get the observatory to see that :-/

src/http.rs
```diff
@@ -76,3 +78,38 @@ pub fn token(token: String) -> Token {
...
+                "X-Frame-Options".into(),
+                vec!["SAMEORIGIN".into()],
+            );
+            response.headers.insert(
+                "X-XSS-Protection".into(),
+                vec!["1; mode=block".into()],
+            );
+            response.headers.insert(
+                "Strict-Transport-Security".into(),
+                vec!["max-age=31536000; includeSubDomains".into()],
```
jtgeibel: Assuming that #895 is merged, this should be removed.  This should probably be configured further up the stack anyway since mirrors could be deployed over HTTP.

If we want to continue setting the header here then as mentioned by @carols10cents `includeSubdomains` should be removed.

src/lib.rs
```diff
@@ -150,6 +150,9 @@ pub fn middleware(app: Arc<App>) -> MiddlewareBuilder {
     m.add(conduit_cookie::Middleware::new(app.session_key.as_bytes()));
     m.add(conduit_cookie::SessionMiddleware::new("cargo_session",
                                                  env == Env::Production));
+    if env == Env::Production {
+        m.add(http::SecurityHeadersMiddleware);
+    }
```
jtgeibel: Its unclear to me exactly what the order of middleware layers will be.  Lower in the file `dist::Middleware` is added with `around` so I think that static files will be served without these headers.  In particular, that would include `index.html`.  The security middleware should probably be the outermost middleware if it isn't already.

URLs matching `^/assets/` will be served directly by nginx, but static files in the root (such as `index.html` and `robots.txt`) are served through this static `dist` middleware.  I have no idea if any of these headers should also be applied to static assets.

jtgeibel: Since these headers were picked up when this was deployed on staging it looks like the middleware layers are stacked in the correct order.

Given that in production most static assets are served directly from nginx, I think `X-Content-Type-Options: nosniff` should also be included in `nginx.conf.erb`.   Looking over the other headers, I don't think any of the rest apply to such assets (except HSTS).

src/http.rs
```diff
@@ -76,3 +78,38 @@ pub fn token(token: String) -> Token {
         token_type: String::new(),
     }
 }
+
+pub struct SecurityHeadersMiddleware;
```
vignesh-sankaran: Could we add a documentation comment here?

carols10cents:
Ok. I poked at this a bit more, it's up on https://staging-crates-io.herokuapp.com right now and gets an A+. 

There were some content security policy violations, but not from google analytics:

- We had some SVGs with inline `style` attributes, that's not allowed
- We had an `iframe`, that's not allowed (but we weren't using it anyway)
- Since this PR was made, we added AJAX requests to docs.rs so those needed to be whitelisted

The only problem I still see and haven't figured out yet is that the google chart on crate pages (ex: https://staging-crates-io.herokuapp.com/crates/blah-types) isn't working because google's stuff uses `eval` :( I'd rather not go through all this trouble of adding these headers and then just have to allow `eval` anyway :( (afaik you have to allow eval for scripts/styles from any domain, not a whitelist)

I tried upgrading to the "new" way of [loading google charts](https://developers.google.com/chart/interactive/docs/basic_load_libs#updateloader) because [this comment](https://github.com/google/google-visualization-issues/issues/1356#issuecomment-287750632) claimed the "current" version (in march) didn't use `eval`, but as far as i could tell it does use `eval` today :(

I guess some headers are better than none? Idk. I'm sleepy and out of ideas.

carols10cents:
Adding `unsafe-eval` lets the charts work. There are no more errors reported in Chrome; Firefox still complains about an `overflow: hidden` (on crate pages only) but I can't tell where it's coming from. Nothing looks broken though.

carols10cents:
@kureuil says this works in Edge (thank you!!!), so let's do this.

bors: r+

kureuil:
@carols10cents I couldn't find anything about the FF Nightly warning, but everything seems to be working perfectly though :)

src/http.rs
```diff
@@ -88,3 +90,48 @@ pub fn token(token: String) -> Token {
...
+            // to load scripts. Remove 'unsafe-eval' if google fixes the issue:
+            // https://github.com/google/google-visualization-issues/issues/1356
+            // or if we switch to a different graph generation library.
+            response.headers.insert(
+                "Content-Security-Policy".into(),
+                vec![
+                    "default-src 'self'; \
+                      connect-src 'self' https://docs.rs; \
+                      script-src 'self' 'unsafe-eval' \
+                                 https://www.google-analytics.com https://www.google.com; \
```
howaboutsynergy: Upstream [issue](https://github.com/google/google-visualization-issues/issues/1356#issuecomment-442852787) seems fixed, is this still needed?

PR #599, opened 2017-03-07
Update the Diesel buildpack

sgrif:
The new version caches the installed binary, and only re-installs when
the contents of `.diesel_version` changes.


carols10cents:
Yay! Gonna go try a deploy to staging right now :)

carols10cents:
Confirmed that on the 2nd deploy after this change, diesel-cli was cached :)

PR #600, opened 2017-03-07
WIP: Port `crates#index` to use Diesel

sgrif:
All the tests are failing since they set up their test data using the
rust-postgres connection, and I still need to port them over. I'm done
for the evening though, and I doubt porting the tests will change the
code significantly, so I figured I'd toss this up now to see if there's
any feedback.

### DIFF
--- a/migrations/20170307211844_versions_yanked_is_not_nullalbe/up.sql
+++ b/migrations/20170307211844_versions_yanked_is_not_nullalbe/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE versions ALTER COLUMN yanked SET NOT NULL;
--- a/src/app.rs
+++ b/src/app.rs
@@ -54,7 +54,8 @@ impl App {
             .helper_threads(if config.env == ::Env::Production {3} else {1})
             .build();
         let diesel_db_config = r2d2::Config::builder()
-            .pool_size(if config.env == ::Env::Production {1} else {1})
+            .pool_size(if config.env == ::Env::Production {50} else {1})
+            .min_idle(if config.env == ::Env::Production {5} else {1})
             .helper_threads(if config.env == ::Env::Production {3} else {1})
             .build();
 
--- a/src/category.rs
+++ b/src/category.rs
@@ -8,10 +8,13 @@ use pg::rows::Row;
 
 use {Model, Crate};
 use db::RequestTransaction;
+use schema::*;
 use util::{RequestUtils, CargoResult, ChainError};
 use util::errors::NotFound;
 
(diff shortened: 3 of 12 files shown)


migrations/20170307211844_versions_yanked_is_not_nullalbe/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE versions ALTER COLUMN yanked SET NOT NULL;
```
carols10cents: 👍 👍 👍 👍

src/app.rs
```diff
@@ -54,7 +54,8 @@ impl App {
             .helper_threads(if config.env == ::Env::Production {3} else {1})
             .build();
         let diesel_db_config = r2d2::Config::builder()
-            .pool_size(if config.env == ::Env::Production {1} else {1})
+            .pool_size(if config.env == ::Env::Production {50} else {1})
```
carols10cents: This 50 is going to have to come down as we discovered yesterday

src/category.rs
```diff
@@ -21,6 +24,15 @@ pub struct Category {
     pub crates_cnt: i32,
 }
 
+#[derive(Associations)]
+#[belongs_to(Category)]
+#[table_name="crates_categories"]
+#[allow(dead_code)] // FIXME: Hey @sgrif add better many-to-many support to Diesel, wouldya?
```
carols10cents: ❤️

src/app.rs
sgrif: I thought prod had 120 connections available? I told it to not try to keep more than 5 idles at a time so it should never get bad on staging.

carols10cents: Oh I see. I thought this would mean I couldn't deploy it to staging.

carols10cents: Wait... `min_idle` means "not more than 5 idles"? That sounds like `max_idles` to me...?

sgrif: Yeah, r2d2's naming/behavior/defaults are all really questionable here.

sgrif: You could totally take down staging by sending 20 concurrent requests though, but I figured that's not a huge concern.

migrations/20170307211844_versions_yanked_is_not_nullalbe/up.sql
sgrif: inb4 prod has a ton of rows where this is null

sgrif:
This requires porting more endpoints than I thought. Going after some other endpoints for now. No point leaving this open.

PR #601, opened 2017-03-07
Local upload instead of S3 to enable publishing in development

carols10cents:
This should make working on crates.io LOTS easier, I hope :)

In order to test out stuff with a development instance of crates.io running locally, it's nice to be able to publish to your local instance and see what happens. But currently, crates.io tries to upload crate files to S3 even if you don't have any values in the S3 config vars. I've gotten around this by having a commit that comments out the S3 publishing parts and running that code when I want to publish, but that's pretty annoying.

Instead, with this PR, if you're running locally (the env var HEROKU is not set) and you have not set the S3_BUCKET env var, then when you publish a crate to your crates.io with `cargo publish --host file:///path/to/crates.io/tmp/index-co`, it'll put the `.crate` tarball in `dist/local_uploads`. Then if you try to download that crate, it'll take advantage of the static file serving that's already happening for things in `dist` and it'll all work!

You still have to set up github oauth so that you can log in and get a token.

I have to delete all the S3 env vars from `.env` to get this to work, but I didn't remove them from `.env.sample` because I'd rather [fix this](https://github.com/slapresta/rust-dotenv/issues/54) and be able to leave them there as a template.

Also, I hardcoded the S3 bucket to `alexcrichton-test` when running the tests, since the tests fail if you don't have that as the value anyway. And there's also a `NoOp` uploader, because I saw one of the scripts basically constructed an uploader that wouldn't do anything anyway.

I did test that uploading to actual S3 actually still works if you choose to set up a bucket and all the variables. When the HEROKU env var is set (ie we're running in production), the server will still panic if any of these variables aren't set. I also tested that files would still be deleted from s3 (and would also be deleted locally if that's what you're running) if the git commit to the index failed, which I tested by putting a sleep in the new crate function just before the git commit and then moving `tmp/index-co` out of the way while a new crate request was sleeping.

I like how this also makes s3 more of an implementation detail that could be swapped out for some other storage someday, by gathering all the s3 functionality into one spot. I didn't go so far as to make the uploader functionality a a trait+trait objects yet, though.

### DIFF
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -7,7 +7,7 @@ extern crate git2;
 extern crate env_logger;
 extern crate s3;
 
-use cargo_registry::env;
+use cargo_registry::{env, Env};
 use civet::Server;
 use std::env;
 use std::fs::{self, File};
@@ -44,13 +44,13 @@ fn main() {
 
     let heroku = env::var("HEROKU").is_ok();
     let cargo_env = if heroku {
-        cargo_registry::Env::Production
+        Env::Production
     } else {
-        cargo_registry::Env::Development
+        Env::Development
     };
 
     let uploader = match (cargo_env, mirror) {
-        (cargo_registry::Env::Production, false) => {
+        (Env::Production, false) => {
             // `env` panics if these vars are not set
             cargo_registry::Uploader::S3 {
                 bucket: s3::Bucket::new(env("S3_BUCKET"),
@@ -61,7 +61,7 @@ fn main() {
                 proxy: None,
             }
         },
-        (cargo_registry::Env::Production, true) => {
+        (Env::Production, true) => {
             // Read-only mirrors don't need access key or secret key,
             // but they might have them. Definitely need bucket though.
             cargo_registry::Uploader::S3 {
@@ -113,7 +113,7 @@ fn main() {
     } else {
         env::var("PORT").ok().and_then(|s| s.parse().ok()).unwrap_or(8888)
     };
-    let threads = if cargo_env == cargo_registry::Env::Development {1} else {50};
+    let threads = if cargo_env == Env::Development {1} else {50};
     let mut cfg = civet::Config::new();
     cfg.port(port).threads(threads).keep_alive(true);
     let _a = Server::start(cfg, app);
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -73,7 +73,7 @@ fn main() {
                 proxy: None,
             }
         },
-        (cargo_registry::Env::Development, _) => {
+        _ => {
             if env::var("S3_BUCKET").is_ok() {
                 println!("Using S3 uploader");
                 cargo_registry::Uploader::S3 {
@@ -89,8 +89,6 @@ fn main() {
                 cargo_registry::Uploader::Local
             }
         },
-        // See immediately before this match where we choose either prod or dev
-        (cargo_registry::Env::Test, _) => unreachable!(),
     };
 
     let config = cargo_registry::Config {
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -5,6 +5,7 @@ extern crate conduit_middleware;
 extern crate civet;
 extern crate git2;
 extern crate env_logger;
+extern crate s3;
 
 use cargo_registry::env;
 use civet::Server;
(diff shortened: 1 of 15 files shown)


src/bin/server.rs
```diff
@@ -38,26 +39,71 @@ fn main() {
...
+
     let heroku = env::var("HEROKU").is_ok();
     let cargo_env = if heroku {
         cargo_registry::Env::Production
     } else {
         cargo_registry::Env::Development
     };
+
+    let uploader = match (cargo_env, mirror) {
+        (cargo_registry::Env::Production, false) => {
```
sgrif: Do you think this code would be a bit more self documenting if an enum with descriptively named variants were used instead of a boolean?

src/bin/server.rs
```diff
@@ -38,26 +39,71 @@ fn main() {
...
+                                            &api_protocol),
+                    proxy: None,
+                }
+            } else {
+                println!("Using local uploader, crate files will be in the dist directory");
+                cargo_registry::Uploader::Local
+            }
+        },
+        // See immediately before this match where we choose either prod or dev
+        (cargo_registry::Env::Test, _) => unreachable!(),
```
sgrif: Since we only care about "prod" and "not prod", do you think it makes sense to have those be the match arms?

src/uploaders.rs
```diff
@@ -0,0 +1,147 @@
...
+            Uploader::S3 { ref bucket, .. } => {
+                format!("https://{}/{}",
+                        bucket.host(),
+                        Uploader::crate_path(crate_name, version))
+            },
+            Uploader::Local => {
+                format!("/local_uploads/{}",
+                        Uploader::crate_path(crate_name, version))
+            }
+            Uploader::NoOp => panic!("no-op uploader does not have crate files"),
```
sgrif: Do you think it's worth returning an `Option` here so that callers are a bit more explicit when they require this?

src/util/io_util.rs
```diff
@@ -22,3 +23,25 @@ impl<R: Read> Read for LimitErrorReader<R> {
         }
     }
 }
+
+pub fn read_le_u32<R: Read + ?Sized>(r: &mut R) -> io::Result<u32> {
```
sgrif: Is it worth using the byteorder crate here?

sgrif:
Looks great! Does this axe S3 stuff in tests as well? Can we delete the `src/test/http-data` if so?

src/bin/server.rs
carols10cents: Maybe? Just an enum of, like, `{PrimarySource, ReadOnlyMirror}`? There may be different kinds of mirrors in the future...

src/bin/server.rs
carols10cents: Yeah, I guess this is coding a bit defensively, and no one's going to go trying to use server.rs in test mode...

src/uploaders.rs
carols10cents: The `NoOp` uploader is mostly for one-off scripts, and I don't see anyone even calling this method in that situation. So I'm not sure it's worth the complexity of making this method return an `Option`, but I don't feel super strongly about it either

src/util/io_util.rs
carols10cents: I have no idea, probably! This code probably predates the byteorder crate :) I didn't change these functions at all, just moved them around, and if we change to use the byteorder crate, I'd want to do that in a separate PR.

carols10cents:
> Does this axe S3 stuff in tests as well? Can we delete the src/test/http-data if so?

Nooooo the tests are still using S3, but with a proxy, so that we're testing the S3 requests and I still very much want to do that. We should eventually extract the recording functionality into a VCR crate though :)

alexcrichton:
Sounds like a great idea! 

My initial intention was that because all the s3 bindings are vendored locally I wanted to make sure they were thoroughly tested. It seemed like it was easy to introduce a regression where local uploads worked and then in production everything died immediately. I'd be wary to remove the http-data until we have a good solution for that as it's caught bugs for me in the past where I made a change that accidentally broke the S3 requests.

src/util/io_util.rs
```diff
@@ -22,3 +23,25 @@ impl<R: Read> Read for LimitErrorReader<R> {
...
+    let mut b = [0; 4];
+    read_fill(r, &mut b)?;
+    Ok(((b[0] as u32) <<  0) |
+       ((b[1] as u32) <<  8) |
+       ((b[2] as u32) << 16) |
+       ((b[3] as u32) << 24))
+}
+
+pub fn read_fill<R: Read + ?Sized>(r: &mut R, mut slice: &mut [u8])
```
alexcrichton: FWIW nowadays we've got `Read::read_exact` which I believe serves this purpose.

src/uploaders.rs
```diff
@@ -0,0 +1,148 @@
...
+
+pub struct Bomb {
+    app: Arc<App>,
+    pub path: Option<String>,
+}
+
+impl Drop for Bomb {
+    fn drop(&mut self) {
+        if let Some(ref path) = self.path {
+            drop(self.app.config.uploader.delete(self.app.clone(), &path));
```
alexcrichton: On second thought, could this emit a log message if this fails?

sgrif:
> I'd be wary to remove the http-data until we have a good solution for that as it's caught bugs for me in the past where I made a change that accidentally broke the S3 requests.

FWIW we're not really actually testing the S3 integration unless those files are in git-ignore, and expire every few days.

carols10cents:
> FWIW we're not really actually testing the S3 integration unless those files are in git-ignore, and expire every few days.

The S3 API doesn't change very often. I don't want to force anyone who wants to run the tests to set up an S3 bucket...

src/uploaders.rs
carols10cents: No problem, done!

alexcrichton:
btw @carols10cents feel free to merge when you're comfortable, this lgtm

PR #607, opened 2017-03-09
Use the database to generate API tokens

sgrif:
There's a ton of code which is setting this field in misleading ways,
and it seems really odd to pass it to `find_or_insert` when it's
completely ignored in the update case. We can just use a database
default for this, and simplify everything a bit.


carols10cents:
tastes great! less filling! 🍰

PR #609, opened 2017-03-09
Port `crates#index` to use Diesel

sgrif:
I decided to tackle this one next since we have an N+1 queries bug on
the max versions, it's a relatively high traffic endpoint, and the code
benefitted the most from Diesel.

There was probably more code added to support migrating the tests over
than to support the actual `index` function, but all those pieces will
start to get used as more endpoints are ported over. There were a few
tests that were hitting the followings and owners endpoints which I have
temporarily changed to hit the database directly instead.

Porting those endpoints would mean porting over more tests, which might
mean porting over more endpoints, etc. We didn't lose any coverage
overall, but we should still go back to hitting the endpoints directly
in tests when we can.

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -9,10 +9,13 @@ use std::sync::Arc;
 use conduit::{Request, Response};
 use conduit_router::RequestParams;
 use curl::easy::Easy;
+use diesel::prelude::*;
+use diesel::pg::{Pg, PgConnection};
+use diesel::pg::upsert::*;
+use diesel_full_text_search::*;
 use license_exprs;
 use pg::GenericConnection;
 use pg::rows::Row;
-use pg::types::ToSql;
 use pg;
 use rustc_serialize::hex::ToHex;
 use rustc_serialize::json;
@@ -31,13 +34,14 @@ use category::EncodableCategory;
 use badge::EncodableBadge;
 use upload;
 use user::RequestUser;
-use owner::{EncodableOwner, Owner, Rights, OwnerKind, Team, rights};
+use owner::{EncodableOwner, Owner, Rights, OwnerKind, Team, rights, CrateOwner};
 use util::errors::NotFound;
 use util::{LimitErrorReader, HashingReader};
 use util::{RequestUtils, CargoResult, internal, ChainError, human};
 use version::EncodableVersion;
+use schema::*;
 
-#[derive(Clone)]
+#[derive(Clone, Queryable, Identifiable, AsChangeset)]
 pub struct Crate {
     pub id: i32,
     pub name: String,
@@ -52,6 +56,19 @@ pub struct Crate {
     pub repository: Option<String>,
     pub max_upload_size: Option<i32>,
 }
+
+/// We literally never want to select textsearchable_index_col
+/// so we provide this type and constant to pass to `.select`
+type AllColumns = (crates::id, crates::name, crates::updated_at,
+    crates::created_at, crates::downloads, crates::description,
+    crates::homepage, crates::documentation, crates::readme, crates::license,
+    crates::repository, crates::max_upload_size);
+
+pub const ALL_COLUMNS: AllColumns = (crates::id, crates::name,
+    crates::updated_at, crates::created_at, crates::downloads,
+    crates::description, crates::homepage, crates::documentation,
+    crates::readme, crates::license, crates::repository,
+    crates::max_upload_size);
 
 #[derive(RustcEncodable, RustcDecodable)]
 pub struct EncodableCrate {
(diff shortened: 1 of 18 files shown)


src/krate.rs
```diff
@@ -81,6 +98,133 @@ pub struct CrateLinks {
...
+    pub homepage: Option<&'a str>,
+    pub documentation: Option<&'a str>,
+    pub readme: Option<&'a str>,
+    pub repository: Option<&'a str>,
+    pub license: Option<&'a str>,
+    pub max_upload_size: Option<i32>,
+}
+
+impl<'a> NewCrate<'a> {
+    pub fn create_or_update(
```
sgrif: This is the same functionality as `find_or_insert`, but with a more clear name, and using Diesel. It's only used in tests currently.

src/krate.rs
```diff
@@ -490,136 +634,88 @@ impl Model for Crate {
 /// Handles the `GET /crates` route.
 #[allow(trivial_casts)]
 pub fn index(req: &mut Request) -> CargoResult<Response> {
-    let conn = req.tx()?;
+    let conn = req.db_conn()?;
     let (offset, limit) = req.pagination(10, 100)?;
-    let query = req.query();
```
sgrif: I recommend using split diff here.

src/krate.rs
```diff
@@ -490,136 +634,88 @@ impl Model for Crate {
...
-    // Collect all the crates
-    let stmt = conn.prepare(&q)?;
-    let mut crates = Vec::new();
-    for row in stmt.query(&args)?.iter() {
-        let krate: Crate = Model::from_row(&row);
-        let badges = krate.badges(conn)?;
-        let max_version = krate.max_version(conn)?;
-        crates.push(krate.minimal_encodable(max_version, Some(badges)));
+    let mut query = crates_query(&params, req.user())?;
+    let sort = params.get("sort").map(|s| &**s).unwrap_or("alpha");
```
carols10cents: Love how the sorting is all in one place now 😻

src/tests/krate.rs
```diff
@@ -490,11 +509,15 @@ fn new_crate_owner() {
                              .with_body(body.as_bytes())));
 
     // Make sure this shows up as one of their crates.
-    let query = format!("user_id={}", u2.id);
-    let mut response = ok_resp!(middle.call(req.with_path("/api/v1/crates")
-                                               .with_method(Method::Get)
-                                               .with_query(&query)));
-    assert_eq!(::json::<CrateList>(&mut response).crates.len(), 1);
+    // FIXME: Once owner endpoints use diesel, go back to hitting the real endpoint
+    assert_eq!(1, req.tx().unwrap().query("SELECT COUNT(*) FROM crate_owners
```
carols10cents: So this FIXME test is still doing an assertion...

src/tests/krate.rs
```diff
@@ -730,46 +753,63 @@ fn dependencies() {
...
-    assert!(::json::<O>(&mut response).ok);
-    let mut response = ok_resp!(middle.call(&mut req));
-    assert!(::json::<O>(&mut response).ok);
+    // let mut response = ok_resp!(middle.call(&mut req));
+    // assert!(!::json::<F>(&mut response).following);
 
-    req.with_path("/api/v1/crates/foo_following/following")
-       .with_method(Method::Get);
-    let mut response = ok_resp!(middle.call(&mut req));
-    assert!(::json::<F>(&mut response).following);
```
carols10cents: ... but this assertion looks commented out....

src/tests/krate.rs
```diff
@@ -730,46 +753,63 @@ fn dependencies() {
...
-    assert!(::json::<O>(&mut response).ok);
-
-    req.with_path("/api/v1/crates/foo_following/following")
-       .with_method(Method::Get);
-    let mut response = ok_resp!(middle.call(&mut req));
-    assert!(!::json::<F>(&mut response).following);
+    // FIXME: Go back to hitting the actual endpoint once it's using Diesel
+    req.db_conn().unwrap()
+        .execute("TRUNCATE TABLE follows")
+        .unwrap();
```
carols10cents: This test that's hitting the database instead of the endpoint looks like it's doing a lot more than the endpoint does...

src/tests/krate.rs
```diff
@@ -730,46 +753,63 @@ fn dependencies() {
...
-       .with_method(Method::Delete);
-    let mut response = ok_resp!(middle.call(&mut req));
-    assert!(::json::<O>(&mut response).ok);
-    let mut response = ok_resp!(middle.call(&mut req));
-    assert!(::json::<O>(&mut response).ok);
-
-    req.with_path("/api/v1/crates/foo_following/following")
-       .with_method(Method::Get);
-    let mut response = ok_resp!(middle.call(&mut req));
-    assert!(!::json::<F>(&mut response).following);
```
carols10cents: ...and this assertion looks commented out...

src/krate.rs
```diff
@@ -638,68 +638,71 @@ impl Model for Crate {
...
-        }
-        (Some(q_string), _) => {
-            let q = plainto_tsquery(q_string);
+    if let Some(q_string) = params.get("q") {
+        let q = plainto_tsquery(q_string);
+        query = query.filter(q.matches(crates::textsearchable_index_col));
+
+        let perfect_match = crates::name.eq(q_string).desc();
+        if sort == "downloads" {
+            query = query.order((perfect_match, crates::downloads.desc()));
```
carols10cents: Oh I guess sorting isn't in one place anymore, but I guess that's ok. This is better than it was! 

I'm guessing later calls to `order` override earlier ones, like in activerecord?

carols10cents:
> One thing that I didn't do here, but would be pretty easy to add if we want it is the ability to compose the branches in index. e.g. being able to write ?category=foo&keyword=bar&q=baz and have it do a full text search for baz on crates that have the category foo and the keyword bar.
> 
> And by pretty easy I mean that it's literally just removing the else from all the else ifs

[Yeah I do want to do that eventually](https://github.com/rust-lang/crates.io/issues/491), but I think we also need to put some thought into which ordering takes precedence when, and how we convey what's going on in the UI as well. So it's not quite literally just that, but I'm glad to know the backend is closer at least :)

src/krate.rs
sgrif: Yes, calls to `order` override the earlier ones. It's actually unlike Active Record (`.reorder` replaces earlier ones, `.order` appends). The only method in Diesel which doesn't override earlier calls is `.filter`.

src/tests/krate.rs
sgrif: Yes, I'm assuming that the table was empty when the test started running.

src/tests/krate.rs
sgrif: So basically this endpoint is tested elsewhere, and I'm assuming that `?following=1` is enough to generally verify this behavior, especially since I'm planning on hitting the following endpoints next after this is merged.

carols10cents:
Ok, fair enough :)

sgrif:
Hopefully this will be the largest one. :heart:

PR #610, opened 2017-03-10
Don't parse `Version#num` for reads

sgrif:
On the majority of endpoints involving versions, we're parsing the
version (using `.unwrap` even, which could take down the server) only to
immediately convert it back into a string. This causes needless
allocations, and a larger surface for errors to appear.

We really only care about semver when ensuring a version is valid, and
when determining the max version for a crate. Everywhere else, we can
just use a String.

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -5,6 +5,7 @@ use std::io::prelude::*;
 use std::io;
 use std::mem;
 use std::sync::Arc;
+use std::borrow::Cow;
 
 use conduit::{Request, Response};
 use conduit_router::RequestParams;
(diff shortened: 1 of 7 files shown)


carols10cents:
> (using .unwrap even, which could take down the server)

wellllll only if the database gets something in it that doesn't parse

src/krate.rs
```diff
@@ -64,7 +65,7 @@ pub struct EncodableCrate {
     pub badges: Option<Vec<EncodableBadge>>,
     pub created_at: String,
     pub downloads: i32,
-    pub max_version: String,
+    pub max_version: Cow<'static, str>,
```
carols10cents: 🐮

carols10cents:
lgtm!

Issue #583, opened 2017-03-01
Migrate from rustc-serialize to Serde


Eh2406:
It looks like it went well in [cargo](https://github.com/rust-lang/cargo/pull/3682), at some point down the line we should do migrate as well.

PR #612, opened 2017-03-10
Use serde to deserialize badges from the database

sgrif:
All of the logic we're using to validate the attributes can be entirely
represented with serde.


Eh2406:
cc: https://github.com/rust-lang/crates.io/issues/583

carols10cents:
LGTM!

PR #613, opened 2017-03-10
Port `crates#summary` over to use Diesel

sgrif:
This endpoint was really contained, so very little supporting code had
to change. This endpoint is extremely light on test coverage, so it
should be verified on staging. As a bonus, it no longer has an N+1
queries bug, and will execute 8 queries instead of 35 (we can get that
down to 6 technically but I don't think it's worth it)

### DIFF
--- a/src/category.rs
+++ b/src/category.rs
@@ -8,11 +8,11 @@ use diesel::pg::PgConnection;
 use pg::GenericConnection;
 use pg::rows::Row;
 
-use {Model, Crate};
 use db::RequestTransaction;
 use schema::*;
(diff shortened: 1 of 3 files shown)


carols10cents:
Deployed this to staging and confirmed the front page has the same info on it 👍

sgrif:
Do you want me to rebase it for you?

carols10cents:
nah i got this, just running tests locally real quick

carols10cents:
merged in but because i rebased it didn't match up, but it's [here](https://github.com/rust-lang/crates.io/commit/3c14e8ab1d1c10f31090ecb69d3108a6abf1c578)

PR #618, opened 2017-03-11
Add clippy, make it pass

sgrif:
Adds the ability to run clippy against the project (disabled by default since it requires nightly), and makes all the required changes to make it pass. The failures were mostly to make the code idiomatic, though there were a batch that was due to requiring unnecessary allocations. I plan on enabling some additional lints to make this more pedantic in future PRs, this is just the default set of lints.

### DIFF
--- a/src/dependency.rs
+++ b/src/dependency.rs
@@ -18,6 +18,12 @@ pub struct Dependency {
     pub kind: Kind,
 }
 
+pub struct ReverseDependency {
+    dependency: Dependency,
+    crate_name: String,
+    crate_downloads: i32,
+}
+
 #[derive(RustcEncodable, RustcDecodable)]
 pub struct EncodableDependency {
     pub id: i32,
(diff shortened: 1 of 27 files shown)


sgrif:
Rebased

src/dependency.rs
```diff
@@ -18,6 +18,12 @@ pub struct Dependency {
     pub kind: Kind,
 }
 
+pub struct ReverseDependency {
```
carols10cents: Mostly just curious, what was it about clippy that made this new `ReverseDependency` struct needed?

src/dependency.rs
```diff
@@ -88,6 +96,12 @@ impl Dependency {
     }
 }
 
+impl ReverseDependency {
+    pub fn encodable(self) -> EncodableDependency {
```
carols10cents: Is the `download` argument in `Dependency.encodable` still necessary, since it says it's only needed when generating a reverse dependency, and this looks like it handles generating a reverse dependency then?

src/dependency.rs
sgrif: Return type was too complex.

PR #619, opened 2017-03-11
README.md: simplify notes on Cargo version

nodakai:


README.md
```diff
@@ -195,10 +195,8 @@ categories will be synced from this file.
 
 **DISCLAIMER: The process of setting up a mirror is a work-in-progress and is
 likely to change. It is not currently recommended for mission-critical
-production use. It also requires a version of cargo newer than 0.13.0-nightly
-(f09ef68 2016-08-02); the version of cargo currently on rustc's beta channel
-fulfils this requirement and will be shipped with rustc 1.12.0 scheduled to be
-released on 2016-09-29.**
+production use. It also requires Cargo from Rust distribution 1.12.0 or
```
sgrif: Should we just say "Cargo 0.13.0" here?

nodakai: No. Crates.io users shouldn't have to care which Cargo version they are using. They should just get the latest Rust distribution with rustup and find Cargo in it.

sgrif: Fair enough.

PR #622, opened 2017-03-12
Follow the Diesel

### LINKED ISSUE #438: "Latest updates" never loads
On https://crates.io/dashboard the "Latest updates" section on the right never seems to load.

The spinner keeps being displayed for some time, then is replaced by a "Load More" button.

Same thing happens again when clicking "Load More".

Note that when the page is loading, and then each time I click "Load More", I get this in the console:

<details>
TypeError: e is undefined
a()
vendor-b5eaa22629fc0ae42aba95c5979b7b1f.js:8
n()
vendor-b5eaa22629fc0ae42aba95c5979b7b1f.js:8
C()
vendor-b5eaa22629fc0ae42aba95c5979b7b1f.js:16
o.inline()
vendor-b5eaa22629fc0ae42aba95c5979b7b1f.js:16
u.prototype.populateNodes()
vendor-b5eaa22629fc0ae42aba95c5979b7b1f.js:17
u.prototype.render()
vendor-b5eaa22629fc0ae42aba95c5979b7b1f.js:17
o()
</details>

Not sure how I can provide more information.

sgrif:
Get it? Because this ports the follow endpoints. To Diesel. ...Anyone?

This ports 4 endpoints over to Diesel. The 3 endpoints which manipulate
the `follows` table, as well as the `/me/updates` endpoint since it
is only hit by the tests for the follows endpoints.

I ended up changing the updates endpoint quite a bit. I wanted to
eliminate the N+1 queries on the max version, and was wondering why we
needed the max version at all here. I went to go look at it in the UI,
and it turns out that the dashboard page which displayed it is actually
broken as well. After fixing it, I noticed that it doesn't need the
crates at all, just the name (which we tell Ember is the id).

I couldn't actually find a good way in Ember to reference the ID of an
association without loading the whole thing. If anybody knows a better
way to do it than what I'm doing here, please let me know.

Since we don't need the crates, I've just opted not to include that data
in the response body (note that just not including the max version is a
bad idea, since ember caches stuff and it could result in a page that
does need the max version displaying wrong later).

While I was touching these endpoints, I also went ahead and reduced them
all to a single query.

Fixes #438.

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -4,9 +4,11 @@ use std::collections::HashMap;
 
 use conduit::{Request, Response};
 use conduit_router::RequestParams;
-use diesel::pg::PgConnection;
+use diesel::associations::Identifiable;
 use diesel::pg::upsert::*;
+use diesel::pg::{Pg, PgConnection};
 use diesel::prelude::*;
+use diesel;
 use diesel_full_text_search::*;
 use license_exprs;
 use pg::GenericConnection;
@@ -35,7 +37,8 @@ use util::{RequestUtils, CargoResult, internal, ChainError, human};
 use version::EncodableVersion;
 use {Model, User, Keyword, Version, Category, Badge, Replica};
 
-#[derive(Clone, Queryable, Identifiable, AsChangeset)]
+#[derive(Clone, Queryable, Identifiable, AsChangeset, Associations)]
+#[has_many(versions)]
 pub struct Crate {
     pub id: i32,
     pub name: String,
@@ -63,6 +66,8 @@ pub const ALL_COLUMNS: AllColumns = (crates::id, crates::name,
     crates::description, crates::homepage, crates::documentation,
     crates::readme, crates::license, crates::repository,
     crates::max_upload_size);
+
+type CrateQuery<'a> = crates::BoxedQuery<'a, Pg, <AllColumns as Expression>::SqlType>;
 
 #[derive(RustcEncodable, RustcDecodable)]
 pub struct EncodableCrate {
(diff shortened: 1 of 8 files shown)


src/krate.rs
```diff
@@ -1093,44 +1107,61 @@ fn user_and_crate(req: &mut Request) -> CargoResult<(User, Crate)> {
...
-    let stmt = tx.prepare("SELECT 1 FROM follows
-                                WHERE user_id = $1 AND crate_id = $2")?;
-    let rows = stmt.query(&[&user.id, &krate.id])?;
-    if !rows.iter().next().is_some() {
-        tx.execute("INSERT INTO follows (user_id, crate_id)
-                         VALUES ($1, $2)", &[&user.id, &krate.id])?;
-    }
+    let follow = follow_target(req)?;
+    let conn = req.db_conn()?;
+    diesel::insert(&follow.on_conflict_do_nothing())
```
carols10cents: 👍 👍 👍 👍

src/user/mod.rs
```diff
@@ -356,62 +355,45 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
...
 
-    // Encode everything!
-    let crates = crates.into_iter().map(|c| {
-        let max_version = c.max_version(tx)?;
-        Ok(c.minimal_encodable(max_version, None))
-    }).collect::<CargoResult<_>>()?;
-    let versions = versions.into_iter().map(|v| {
-        let id = v.crate_id;
-        v.encodable(&map[&id])
+    let followed_crates = Follow::belonging_to(user)
```
carols10cents: OMG this is so much better!! And it works now tooo yay!!!!!

carols10cents:
I don't really get the "follow the diesel" thing but this is a good PR!

PR #623, opened 2017-03-12
Optimize the reverse dependencies endpoint

sgrif:
Right now on production the reverse_dependencies endpoint takes around a
second to load. This commit decreases that to ~70ms on my machine.
I've moved the query to its own file to give syntax highlighting and add
comments. The query barely differs from the original though. I've made
two main changes.

The first is using a window function to get the total count instead of a
second query. The second is the addition of `where crate_id =
any(subselect)` when gathering the versions.

Ultimately this query is an intersection between all the max versions,
and the versions which depend on this crate. Prior to this commit, the
query essentially did exactly that. The problem is that even if an index
is used, collecting all the max crates takes ~300ms. This speeds that up
considerably by first scoping the versions to versions of crates which
have ever depended on the target, then collecting the max versions, then
filtering it to versions which depend on the target.

The query looks considerably more complex (and it is), but it's the
difference between the query taking 70ms and the query taking 300ms.

One further alternative would be to have a materialized view for the max
versions, which gets the query time down to ~10ms. However, I don't
think the complexity of adding (and refreshing) the view is worth it to
simplify one query.

Honestly, if we don't want this complexity it's probably fine to just leave it
as is. 1s is slow, but ultimately not the end of the world.

### DIFF
--- a/migrations/20170312165531_index_dependencies/down.sql
+++ b/migrations/20170312165531_index_dependencies/down.sql
@@ -1,0 +1,1 @@
+DROP INDEX dependencies_crate_id_version_id_idx;
--- a/migrations/20170312165531_index_dependencies/up.sql
+++ b/migrations/20170312165531_index_dependencies/up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX ON dependencies (crate_id, version_id);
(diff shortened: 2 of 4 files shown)


carols10cents:
Seems good to me!

PR #628, opened 2017-03-14
Don't tie the lifetime of the db connection to the request

sgrif:
Consider code like the following:

```rust
let conn = req.db_conn()?;
let body = req.body();
```

Since the `body` function requires `&mut self`, this code will fail
since the lifetime of the connection is tied to the lifetime of the
request. These are of course disjoint, but the compiler can't know that.

Really there's no reason to stick the request in `extensions` like we
are. I think the reason the postgres code did it that way was to allow
mocking the connection out in tests, but we just set the pool size to 1.

Additionally, some endpoints (like `crates#summary` are embarassingly
parallelizable, so we should expose the capability for multiple database
connections in a single request.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -205,20 +205,11 @@ impl Middleware for TransactionMiddleware {
     }
 }
 
-pub struct DieselMiddleware;
-
-impl Middleware for DieselMiddleware {
-    fn before(&self, request: &mut Request) -> Result<(), Box<Error+Send>> {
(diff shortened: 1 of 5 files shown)


carols10cents:
This looks good, I think I'm going to deploy it separately from the batch of changes that I just merged, so I can make sure all those work without changing the connection as well, and then be able to roll just this one back if something goes sideways.

sgrif:
👍

PR #629, opened 2017-03-15
Update the dependencies schema to match use

sgrif:
Dependencies kind is assigned a default value when deserialized from the
database, and all newly inserted values are given a non-null value.
Let's just give it a default at the database level and mark it not-null.

Additionally, we were storing it as a comma-delmited string but there
was really no reason to do so. It was converted back to a string before
being sent to the front-end, but we converted to/from an array in
between so there was no gain from storing it as a string. Additionally,
the front-end never used it as a comma delimited string (actually, it
doesn't use `dependencies.features` at all unless there's some crazy
metaprogramming that doesn't show up when grepping for "features" and
"featureList")

I honestly can't find any reason in the history that they were stored
this way. I suspect it might be because they used to be stored as a
string which used pipes as a delimiter, so that heritage just stuck
around when it moved to being a separate value in the database and
index.

### DIFF
--- a/app/models/dependency.js
+++ b/app/models/dependency.js
@@ -3,8 +3,6 @@ import Ember from 'ember';
 
 Ember.Inflector.inflector.irregular('dependency', 'dependencies');
 
-const { computed } = Ember;
-
 export default DS.Model.extend({
     version: DS.belongsTo('version', {
         async: false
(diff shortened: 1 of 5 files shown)


carols10cents:
LGTM. Ran that migration on a dump of prod I have locally just to make sure it wouldn't take very long, it was like a second so this is fine.

Issue #88, opened 2014-12-03
Migrate to using error-chain


alexcrichton:
This module should largely obsolete what we have in `util::error`

Eh2406:
What is the best why to get started on this?

alexcrichton:
Oh dear this is quite an ancient bug! Nowaday's I'd recommend moving to `error-chain`, and that'd just start out by looking in `src/util/errors.rs` and migrating it all to error-chain's errors.

Eh2406:
I have not yet had the pleasure of using `error-chain`, so I am going to go do some reading.
Is there a way to do this cleanup incrementally? If it involves groking all error code in the project than re architecting the hole thing it may be more than I can do. If it can be done one piece at a time, it sounds more doable.

alexcrichton:
Unfortunately this is likely an "all at once transition" rather than a piecemeal one

carols10cents:
I don't think we're going to do this.

PR #631, opened 2017-03-15
Ensure Diesel NotFound results in a 404

sgrif:
Any method in Diesel's API which returns a single record (specifically
`.get_result` and `.first`) treats 0 records as an error case by
default (calling `.optional` on the `Result` turns it into a
Result<Option<T>>`). However, right now if a
`diesel::result::Error::NotFound` is returned, the server will return a
500. We don't test every possible place a 404 can be returned, so this
is never apparent in tests.

The implementation is a bit of a kludge, but I'm not sure that there's a
better way to do this without either requiring nightly for
specialization or significantly reworking the error handling of the
application (I do think that we should eventually do the latter)

### DIFF
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -774,6 +774,21 @@ fn dependencies() {
 }
 
 #[test]
+fn diesel_not_found_results_in_404() {
+    let (_b, app, middle) = ::app();
+    let mut req = ::req(app.clone(), Method::Get, "/api/v1/crates/foo_following/following");
+
+    {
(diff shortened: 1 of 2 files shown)


carols10cents:
> significantly reworking the error handling of the
> application (I do think that we should eventually do the latter)

Agreed: #88

carols10cents:
Confirmed that before this PR, making a request to `http://localhost:4200/api/v1/crates/nonexistent-crate/following` caused a 500 and after this PR it returns 404 :)

PR #632, opened 2017-03-15
Don't show yanked versions on the summary page

sgrif:


carols10cents:
Does what it says on the tin!

PR #636, opened 2017-03-18
Upgrade to Ember 2.12

bjornharrtell:
Also uses latest Node LTS at Travis.


carols10cents:
Looks great! Tested out a bunch of things locally and it all seems to look and work like I'd expect.

One small thing, do you know why [these warnings](https://travis-ci.org/rust-lang/crates.io/builds/212435437#L1152) are showing up when running `ember test`? It'd be nice to clean those up, but I'm not sure what the problem is (I haven't done any research yet either).

```
ok 29 PhantomJS 2.1 - Acceptance | categories: listing categories
    ---
        Log: |
            { type: 'warn',
              text: '\'DEPRECATION: Ember.K is deprecated in favor of defining a function inline. [deprecation id: ember-metal.ember-k] See http://emberjs.com/deprecations/v2.x#toc_code-ember-k-code for more details.\\n    http://localhost:7357/assets/vendor.js:20788:15
```

bjornharrtell:
I suspect they are coming from an add-on. No reason try try to get it fixed, but might make it harder.

bjornharrtell:
Culprit seems to be here:
https://github.com/RSSchermer/ember-rl-dropdown/blob/master/addon/mixins/rl-dropdown-component.js#L11

bjornharrtell:
@carols10cents as ember-rl-downdown seems abandoned I decided to adopt it as source, as it's fairly small, and fix the deprecation. Perhaps it can replaced by a more contemporary component at some point, mabye http://www.ember-power-select.com/ which potentially can replace both rl-dropdown and emberx-select.

carols10cents:
Looks great!!! Thank you!!! 🆙

PR #640, opened 2017-03-18
Port `/crates/new` and supporting endpoints to use Diesel

sgrif:
I regret that this PR had to be so large, but there was a cluster of endpoints that all had to be ported together or we'd lose test coverage. I have kept separate commits that are roughly the size of what I would normally submit as a PR, so I suggest reviewing this that way.

### DIFF
--- a/src/badge.rs
+++ b/src/badge.rs
@@ -3,7 +3,7 @@ use krate::Crate;
 use schema::badges;
 use util::CargoResult;
 
-use diesel::pg::Pg;
+use diesel::pg::{Pg, PgConnection};
 use diesel::prelude::*;
 use pg::GenericConnection;
 use pg::rows::Row;
@@ -67,7 +67,48 @@ impl Badge {
         }
     }
 
-    pub fn update_crate(conn: &GenericConnection,
+    pub fn update_crate<'a>(conn: &PgConnection,
+                            krate: &Crate,
+                            badges: Option<&'a HashMap<String, HashMap<String, String>>>)
+                            -> CargoResult<Vec<&'a str>> {
+        use diesel::{insert, delete};
+
+        #[derive(Insertable)]
+        #[table_name="badges"]
+        struct NewBadge<'a> {
+            crate_id: i32,
+            badge_type: &'a str,
+            attributes: serde_json::Value,
+        }
+
+        let mut invalid_badges = vec![];
+        let mut new_badges = vec![];
+
+        if let Some(badges) = badges {
+            for (k, v) in badges {
+                let attributes_json = serde_json::to_value(v).unwrap();
+
+                let json = json!({"badge_type": k, "attributes": attributes_json});
+                if serde_json::from_value::<Badge>(json).is_ok() {
+                    new_badges.push(NewBadge {
+                        crate_id: krate.id,
+                        badge_type: &**k,
+                        attributes: attributes_json,
+                    });
+                } else {
+                    invalid_badges.push(&**k);
+                }
+            }
+        }
+
+        conn.transaction(|| {
+            delete(badges::table.filter(badges::crate_id.eq(krate.id))).execute(conn)?;
+            insert(&new_badges).into(badges::table).execute(conn)?;
+            Ok(invalid_badges)
+        })
+    }
+
+    pub fn update_crate_old(conn: &GenericConnection,
                         krate: &Crate,
                         badges: HashMap<String, HashMap<String, String>>)
                         -> CargoResult<Vec<String>> {
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -5,6 +5,7 @@ use std::collections::HashMap;
 use conduit::{Request, Response};
 use conduit_router::RequestParams;
 use diesel::associations::Identifiable;
+use diesel::helper_types::Select;
 use diesel::pg::upsert::*;
 use diesel::pg::{Pg, PgConnection};
 use diesel::prelude::*;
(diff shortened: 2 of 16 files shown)


src/tests/krate.rs
```diff
@@ -1115,9 +1173,12 @@ fn good_categories() {
...
     let krate = ::krate("foo_good_cat");
     let cats = vec!["cat1".into()];
-    let mut req = ::new_req_with_categories(app, krate, "1.0.0", cats);
-    ::mock_category(&mut req, "cat1", "cat1");
-    ::mock_user(&mut req, ::user("foo"));
+    let mut req = ::new_req_with_categories(app.clone(), krate, "1.0.0", cats);
+    ::sign_in(&mut req, &app);
+    {
+        let conn = app.diesel_database.get().unwrap();
+        ::new_category("Category 1", "cat1").find_or_create(&conn).unwrap();
```
carols10cents: 👍 👍 👍 👍

src/tests/team.rs
```diff
@@ -142,8 +142,8 @@ fn remove_team_as_named_owner() {
             "{:?}", json.errors);
 }
 
-// Test removing team as team owner
```
carols10cents: 😻

src/version.rs
```diff
@@ -207,6 +215,55 @@ impl Version {
...
+            crate_id: crate_id,
+            num: num.to_string(),
+            features: features,
+        })
+    }
+
+    pub fn save(&self, conn: &PgConnection, authors: &[String]) -> CargoResult<Version> {
+        use diesel::{select, insert};
+        use diesel::expression::dsl::exists;
+        use schema::versions::dsl::*;
```
carols10cents: do i need to check this for underhandedness? :P https://twitter.com/sgrif/status/842850645577154561

src/krate.rs
```diff
@@ -529,37 +534,34 @@ impl Crate {
...
-        }
+        let crate_owner = CrateOwner {
+            crate_id: self.id,
+            owner_id: owner.id(),
+            created_by: req_user.id,
+            owner_kind: owner.kind() as i32,
+        };
+        diesel::insert(&crate_owner.on_conflict(
+                crate_owners::table.primary_key(),
+                do_update().set(crate_owners::deleted.eq(false)),
```
carols10cents: 😍 upsert 😍

src/krate.rs
```diff
@@ -529,37 +534,34 @@ impl Crate {
...
-                              SET deleted = TRUE
-                            WHERE crate_id = $1 AND owner_id = $2
-                              AND owner_kind = $3",
-                     &[&self.id, &owner.id(), &owner.kind()])?;
+        let target = crate_owners::table.find((
+            self.id(),
+            owner.id(),
+            owner.kind() as i32,
+        ));
+        diesel::delete(target).execute(conn)?;
```
carols10cents: I don't see where delete on `CrateOwner` is defined to set `deleted` to true, does this change get rid of the soft deletion then? (I'm not sure why soft delete was implemented for crate owners)

src/tests/all.rs
```diff
@@ -17,11 +17,12 @@ extern crate time;
 extern crate url;
 extern crate s3;
 
+use rustc_serialize::json::{self, Json};
+use std::borrow::Cow;
```
carols10cents: 🐮

src/tests/team.rs
```diff
@@ -54,13 +52,12 @@ fn weird_name() {
             "{:?}", json.errors);
 }
 
-// Test adding team without second `:`
+// // Test adding team without second `:`
```
carols10cents: Nit: Is this line so nice we need to comment it out twice? ;) This happens a few times in this file/commit.

src/tests/krate.rs
```diff
@@ -986,16 +985,14 @@ fn yank() {
 #[test]
 fn yank_not_owner() {
     let (_b, app, middle) = ::app();
-    let mut req = ::req(app, Method::Delete, "/api/v1/crates/foo_not/1.0.0/yank");
-    ::mock_user(&mut req, ::user("foo"));
-    ::mock_crate(&mut req, ::krate("foo_not"));
-    ::mock_user(&mut req, ::user("bar"));
+    let mut req = ::request_with_user_and_mock_crate(
+        &app, ::new_user("bar"), "foo_not");
+    ::sign_in(&mut req, &app);
```
carols10cents: I don't see how this test is actually attempting a yank anymore, since `Method::Delete, "/api/v1/crates/foo_not/1.0.0/yank"` was removed and I don't see anything about delete or yank in what was added.....

src/version.rs
```diff
@@ -348,14 +348,15 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
...
             let id = &req.params()["version_id"];
             let id = id.parse().unwrap_or(0);
-            let conn = req.tx()?;
-            let version = Version::find(&*conn, id)?;
-            let krate = Crate::find(&*conn, version.crate_id)?;
-            (version, krate)
+            let conn = req.db_conn()?;
+            versions::table.find(id)
+                .inner_join(crates::table)
+                .select((versions::all_columns, ::krate::ALL_COLUMNS))
```
carols10cents: whoa cool!

src/badge.rs
```diff
@@ -67,7 +67,48 @@ impl Badge {
...
+                    });
+                } else {
+                    invalid_badges.push(&**k);
+                }
+            }
+        }
+
+        conn.transaction(|| {
+            delete(badges::table.filter(badges::crate_id.eq(krate.id))).execute(conn)?;
+            insert(&new_badges).into(badges::table).execute(conn)?;
```
sgrif: No. A conflict could never happen here, because I simplified the logic a bit. Instead of "delete all badges not in the list, add badges and everything in the list with an on conflict clause" I'm just deleting them all and adding everything in the list.

src/krate.rs
sgrif: Oh whoops, you're right this should be soft deleting. I'm not sure why soft delete was implemented either. It seems odd.

src/tests/krate.rs
sgrif: Ah good point. We should probably have that endpoint return a 403 instead of a 200 with an error message, which would have caught this.

sgrif: (Actually the endpoint it's hitting should return a 403, too...)

sgrif:
Updated. All good?

sgrif:
Rebased

carols10cents:
Thaaaank youuuu sorry I took so long :) 🐢

PR #647, opened 2017-03-24
Sync doc sections

matklad:
See https://github.com/rust-lang/cargo/pull/3865

### DIFF
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -48,7 +48,7 @@
                 <li><a href='http://doc.crates.io/pkgid-spec.html'>Package ID specs</a></li>
                 <li><a href='http://doc.crates.io/environment-variables.html'>Environment Variables</a></li>
                 <li><a href='http://doc.crates.io/source-replacement.html'>Source Replacement</a></li>
-                <li><a href='http://doc.crates.io/machine-readable-output.html'>Machine readable output</a></li>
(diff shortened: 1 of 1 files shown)


carols10cents:
Thank you!!!! I don't like how this is split across repos.... someday we'll fix it...

matklad:
@carols10cents should something else be done to upgrade the crates.io site? It still points to the old machine readable output section, which now gives 404. 

> I don't like how this is split across repos.....

I also wonder why `doc.crates.io` is an entirely separate machine...

steveklabnik:
> I also wonder why doc.crates.io is an entirely separate machine...

It's just github pages, rather than being a dynamic site. That's all it
really needs anyway.

carols10cents:
@matklad I need to deploy. Working on it right now, I had an issue.

@steveklabnik the problem is that the docs menu on the dynamic crates.io pages needs to match the menu on github pages and that is 💩

carols10cents:
@matklad deployed! Sorry for the delay! ❤️

steveklabnik:
Oh yeah, it's certainly less than ideal.

carols10cents:
Ok. Redeployed. 🤞🤞🤞🤞🍀🍀🍀🍀🍀

carols10cents:
Ok for real real real this is actually deployed and i dont think i'll have to roll this one back. Famous last words, I know :)

PR #650, opened 2017-03-26
Search field keyboard shortcut

### LINKED ISSUE #643: Pressing 'S' should focus the search bar
Like on the [docs](https://doc.rust-lang.org/stable/std/)

kureuil:
Hey there, first PR to this project, I hope I did not make too much mistakes.

Most of the code was adapted from rust-doc (wondering if and how I should credit them) to using with Ember, though it doesn't really feel Ember *-ish* it would seem that adding custom events to an entire application isn't. I found some addons that achieved this but I wasn't sure it was worth to add a dependency for one keyboard shortcut.

This functionality required adding an ID to the search bar in order to be able to grab its focus as its name is not unique in the document (the mobile search bar shares the same name 'q').
I also made an acceptance test for this feature.

Fixes #643

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -9,6 +9,53 @@ export default Ember.Controller.extend({
     nextFlashError: null,
     search: Ember.computed.oneWay('searchController.q'),
 
+    init() {
+        this._super(...arguments);
+        Ember.$(document).on('keypress', this.handleKeyPress.bind(this));
+        Ember.$(document).on('keydown', this.handleKeyPress.bind(this));
+    },
+
+    // Gets the human-readable string for the virtual-key code of the
+    // given KeyboardEvent, ev.
+    //
+    // This function is meant as a polyfill for KeyboardEvent#key,
+    // since it is not supported in Trident.  We also test for
+    // KeyboardEvent#keyCode because the handleShortcut handler is
+    // also registered for the keydown event, because Blink doesn't fire
+    // keypress on hitting the Escape key.
+    //
+    // So I guess you could say things are getting pretty interoperable.
+    getVirtualKey(ev) {
+        if ("key" in ev && typeof ev.key !== "undefined") {
+            return ev.key;
+        }
+        const c = ev.charCode || ev.keyCode;
+        if (c === 27) {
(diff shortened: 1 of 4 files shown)


carols10cents:
Welcome!! Thank you so much, I'm very excited about this feature :) 

I especially love that you added tests!! ✅ ✅ ✅ 

My only suggestion is-- could we make the search box a little wider? The placeholder text is cut off for me. I think we've got room, especially at browsers 900px or wider:

screen shot 2017-03-28 at 2 59 53 pm

The tricky part is going to be for widths between 900px and 820px, when the mobile layout kicks in:

screen shot 2017-03-28 at 2 59 28 pm

How are you with CSS? I can play around and help figure stuff out if you need it, but I'm slow :)

kureuil:
I'm totally OK to make the search bar wider, I did not have this problem using my browser + font. If we want to make it wider for "medium" (820px) layout maybe the simplest way is to display the full width search bar from the mobile layout ? By the way, I'm pretty good with CSS :)
I'll make the adjustments for the large layout tomorrow.

Just wondering, but maybe the search bar would be more discoverable if it was always full-width, like on the mobile layout (thought I don't have any stats/numbers to back this).

kureuil:
So I just made the width of the search bar dynamic, which should make the placeholder completely visible on most screens. However, on 820+px layouts, The search bar might not be big enough depending on the font provided by the system. The two solutions I see are:
1. Having the mobile layout take place until we reach a wide enough screen (880-900px)
2. Having the mobile navigation dropdown when the screen width is comprised between 821px and 880-900px

@carols10cents What are your thoughts on this ? I think option 2. might make more sense given the current layout of the page...

carols10cents:
Sorry for the delay in answering this! I agree that 2 sounds like the better option and would make sense.

kureuil:
I just made the changes to the stylesheet. Here are screenshots for the current version, does this look good to you ?

screenshot from 2017-04-04 13-53-09

screenshot from 2017-04-04 13-52-29

screenshot from 2017-04-04 13-51-57

carols10cents:
Love it!!!! Thank you!!

PR #656, opened 2017-03-29
Fix date

carols10cents:
This reverts the revert of #641, adds to the tests so they failed with what I was seeing on staging, and fixes the problem :)

r? @sgrif ?


sgrif:
LGTM

PR #665, opened 2017-04-01
Versions should be sorted in semver order in crate show json

### LINKED ISSUE #664: Crate version order is randomly generated
+ Example: https://crates.io/crates/ndarray/versions
+ Example: https://crates.io/crates/libc/versions

They are not correct in neither publish date nor version order. The output order also changes from time to time when your reload (but sometimes it's cached or something).

The same order is used for the version list on the crate's page.

carols10cents:
Fixes #664.

Also added a test for /crates/:crate_id/versions; that was functioning
correctly before this commit though since it's still calling
krate::versions.

Also added some notes-- I'm not sure why /crates/:crate_id/versions
exists (and I am seeing ember request it) if the version information is
included in the crate show json. I also don't know where or how the
/api/v1/versions endpoint is used, with the `ids=[]` params... I wanted
a reminder to investigate that some other time.

r? @sgrif

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -814,8 +814,11 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
     let name = &req.params()["crate_id"];
     let conn = req.db_conn()?;
     let krate = Crate::by_name(name).first::<Crate>(&*conn)?;
-    let versions = Version::belonging_to(&krate).load::<Version>(&*conn)?;
+
+    let mut versions = Version::belonging_to(&krate).load::<Version>(&*conn)?;
+    versions.sort_by(|a, b| b.num.cmp(&a.num));
     let ids = versions.iter().map(|v| v.id).collect();
+
     let kws = CrateKeyword::belonging_to(&krate)
         .inner_join(keywords::table)
         .select(keywords::all_columns)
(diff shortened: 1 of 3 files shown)


src/krate.rs
```diff
@@ -814,8 +814,11 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
     let name = &req.params()["crate_id"];
     let conn = req.db_conn()?;
     let krate = Crate::by_name(name).first::<Crate>(&*conn)?;
-    let versions = Version::belonging_to(&krate).load::<Version>(&*conn)?;
+
+    let mut versions = Version::belonging_to(&krate).load::<Version>(&*conn)?;
+    versions.sort_by(|a, b| b.num.cmp(&a.num));
```
sfackler: Seems like it'd be nice to handle the sort on the DB side.

carols10cents: Sure would be. Can we write a semver sorting plugin for postgres in rust?

carols10cents:
@sgrif has approved this via mobile :)

src/krate.rs
steveklabnik: yes but you can't run it on heroku

carols10cents: ahh true. maybe someday.

PR #669, opened 2017-04-06
Fix the infinite spinner on the follow button

### LINKED ISSUE #644: Logging in on a crate page displays a spinner on the follow button
## Steps to reproduce:

- Log out of crates.io.
- Navigate to a crate's page, like https://crates.io/crates/libc
- Log in.
- You should still be on https://crates.io/crates/libc, and the follow button will appear but will have a spinner in it instead of "Follow" or "Unfollow": 

screen shot 2017-03-20 at 11 30 50 am

- This happens whether you're following the crate or not.

## Expected behavior:

That the button says "Follow" or "Unfollow" as appropriate. Reloading the page after logging in produces the expected behavior.

## Additional information:

When I log in, I don't see the ajax request to the `https://crates.io/api/v1/crates/libc/following` request happening. That's what determines whether the button should say "Following" or "Unfollowing".

[This is the code that determines what gets displayed on the button](https://github.com/rust-lang/crates.io/blob/9e0aef3dae59c88aca2f384f4cb7a6452910b8c9/app/templates/crate/version.hbs#L12-L22). 

[This is the code that does the ajax request to the following API](https://github.com/rust-lang/crates.io/blob/672eaa9feb66ac6c7f6e4a5d4ac590eb5487ad30/app/routes/crate/version.js#L60-L66).

[This is the code that opens the github popup and then does stuff when logging in is successful](https://github.com/rust-lang/crates.io/blob/0f3adf36108dfac662e3463b4ff97567261f5de9/app/routes/login.js).

I'm not sure exactly how to fix this, but I'm up for digging and trying things with someone! Please let me know if you have questions!!

kureuil:
Currently, when you log in on a page, it would seem only the `session.currentUser` object gets updated. This automatically updates the menu and that's fine for most of the pages. However, on crate/version, `session.currentUser` is used in the template but its value is not watched in the route, which means that the template knows that a user logged in but the route doesn't.
This PR forces the route to reload when the `session.isLoggedIn` value is changed, allowing the route to correctly update the follow button.

Fixes #644

### DIFF
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -2,6 +2,9 @@ import Ember from 'ember';
 import ajax from 'ic-ajax';
 
 export default Ember.Route.extend({
+    refreshAfterLogin: Ember.observer('session.isLoggedIn', function() {
+        this.refresh();
+    }),
 
     model(params) {
         const requestedVersion = params.version_num === 'all' ? '' : params.version_num;


carols10cents:
Looks great!!! I don't know if I would ever have figured out how to fix this in the way that you did. Thank you so much!! ❤️ ❤️ ❤️

Issue #654, opened 2017-03-29
Show latest stable version on search result page


badboy:
Currently it seems to show the latest released version, even if it is a pre-release.

Example: https://crates.io/search?q=bincode
I see: bincode v1.0.0-alpha6

Following the link I get to `bincode v0.6.1`

It should show v0.6.1 on the search page as well.

carols10cents:
Thank you! This is related to, but slightly different from, #510.

briansmith:
Not sure if this is the same thing, but pages like https://crates.io/categories/cryptography also show prerelease (e.g. alpha) versions instead of showing the latest stable version.

Turbo87:
I just had a quick look at this as it was labeled "A-ember". The problem is that from the API we only get `max_version` (highest version; semver order) and `newest_version` (the most recently updated version), but both include prereleases. I guess this would need to be changed on the API side, so that `max_version` only considers regular releases, but not prereleases. Unfortunately, I can't judge whether that might break anything else... 😞

PR #676, opened 2017-04-10
Add docs.rs integration

### LINKED ISSUE #419: docs.rs integration

kureuil:
When a crate does not specify a documentation link, we query docs.rs to know
if the documentation for the requested version was successfully built. If
if is, we set the crate's documentation link to the corresponding docs.rs
page, if not nothing changes.

Fixes #419.

### DIFF
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -35,6 +35,19 @@ export default Ember.Route.extend({
             return result;
         };
 
+        const fetchCrateDocumentation = () => {
+            if (!crate.get('documentation')) {
+                let crateName = crate.get('name');
+                let crateVersion = params.version_num;
+                ajax(`https://docs.rs/crate/${crateName}/${crateVersion}/builds.json`)
+                    .then((r) => {
+                        if (r.length > 0 && r[0].build_status === true) {
+                            crate.set('documentation', `https://docs.rs/${crateName}/${crateVersion}/`);
+                        }
+                    });
+            }
+        };
+
         // Fallback to the crate's last stable version
         // If `max_version` is `0.0.0` then all versions have been yanked
         if (!requestedVersion && maxVersion !== '0.0.0') {
@@ -52,10 +65,13 @@ export default Ember.Route.extend({
                     } else {
                         params.version_num = latestStableVersion.get('num');
                     }
-                });
+                }).then(fetchCrateDocumentation);
             } else {
                 params.version_num = maxVersion;
+                fetchCrateDocumentation();
             }
+        } else {
+            fetchCrateDocumentation();
         }
 
         controller.set('crate', crate);
@@ -66,17 +82,6 @@ export default Ember.Route.extend({
             ajax(`/api/v1/crates/${crate.get('name')}/following`)
                 .then((d) => controller.set('following', d.following))
                 .finally(() => controller.set('fetchingFollowing', false));
-        }
-
-        if (!crate.get('documentation')) {
-            let crateName = crate.get('name');
-            let crateVersion = params.version_num;
-            ajax(`https://docs.rs/crate/${crateName}/${crateVersion}/builds.json`)
-                .then((r) => {
-                    if (r.length > 0 && r[0].build_status === true) {
-                        crate.set('documentation', `https://docs.rs/${crateName}/${crateVersion}/`);
-                    }
-                });
         }
 
         // Find version model
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -8,7 +8,6 @@ export default Ember.Route.extend({
 
     model(params) {
         const requestedVersion = params.version_num === 'all' ? '' : params.version_num;
-
         const crate = this.modelFor('crate');
         const controller = this.controllerFor(this.routeName);
         const maxVersion = crate.get('max_version');
@@ -36,7 +35,8 @@ export default Ember.Route.extend({
         };
 
         const fetchCrateDocumentation = () => {
-            if (!crate.get('documentation')) {
+            if (!crate.get('documentation') ||
+                crate.get('documentation').substr(0, 16) === 'https://docs.rs/') {
                 let crateName = crate.get('name');
                 let crateVersion = params.version_num;
                 ajax(`https://docs.rs/crate/${crateName}/${crateVersion}/builds.json`)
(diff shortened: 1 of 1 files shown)


carols10cents:
I am so excited about this!!! I've been waiting for someone to take care of this ❤️ ❤️ ❤️  I'm sorry for letting this sit for a few days!!!

So I tried this out locally against production's API by running `yarn run start:live`, and the following things work great!! 

- I don't see an ajax request to docs.rs if a crate has a documentation link specified!!!
- I see an ajax request and then don't see a documentation link for crates that docs.rs wasn't able to build docs for, like http://localhost:4200/crates/a !!!
- I see a Documentation link appear for a crate without one specified when I'm on a crate page like http://localhost:4200/crates/abra !!!
- I see a Documentation link appear for a crate without one specified when I'm on a crate version page like http://localhost:4200/crates/sassers/0.13.4-maybe-not-broken !!!!

I think there's still a problem with crates that only have unstable versions on their crate page-- sassers is one of those, and I'm seeing the version number be `undefined` in the docs.rs request. I don't think that's a problem with your code though, I think that's basically a new symptom of #654/#510 :( So don't worry about that, and I'll try to fix that problem soon.

One question I have, though, is whether we should be doing this on crate list pages, like search result pages, `/crates`, category lists, keyword lists, etc since we have the documentation links there now too. I think it would be useful, and I hope doing a bunch of ajax requests to docs.rs on loads of those pages won't take too long or DDoS docs.rs, but we'll find out?

What do you think about trying to add that too?

onur:
@carols10cents when I started adding `builds.json` endpoint to docs.rs we didn't have documentation links in crate list pages. I don't think it's a good idea to send multiple requests to docs.rs in this pages (docs.rs can handle thanks to rust but it will probably take too long for client).

I'll try to add another method to query multiple crates with only one request.

BTW please use: `https://docs.rs/${crateName}/${crateVersion}` link instead of `https://docs.rs/${crateName}/${crateVersion}/${crateName}/`. Last part of the link might be different than `${crateName}`. <https://docs.rs/calculator> for example, using `exp` for `lib.name` and `calculator` for `package.name`.

kureuil:
So, I just pushed some changes to the branch:

* The generated link to docs.rs now omits the `lib.name` (thanks @onur)
* No more `undefined` version numbers on a crate page with only unstable versions (I didn't see the query fetching all of the crates' versions was asynchronous)

@carols10cents I also think it would be very useful to have the auto-link on every page where a crate is referenced but I agree with @onur's argument, it might not be the best experience for end users to have up to *N* requests going to docs.rs for documentation links. Maybe we could fetch the build status in the backend and have the API return the docs.rs link if the crate did not specified a documentation link. This could also allow for more aggressive caching of the documentation link.
BTW, shouldn't the documentation link be stored in the `Version` model ? Why isn't the crate metadata (except for features) versionned ?

carols10cents:
Thanks for the updates! Taking a look at the changes now. It might be a bit confusing why a documentation link appears sometimes when you click into a crate's page, but we can add the docs.rs link to lists of crates when either docs.rs has an endpoint to query multiple crates at once or we add caching to crates.io's backend.

> BTW, shouldn't the documentation link be stored in the Version model ? Why isn't the crate metadata (except for features) versionned ?

I don't know exactly :) I could see a reason for the documentation links to be associated with a version, but I don't think anyone was hosting anything other than their most recent documentation when crates.io was created. Links like the homepage and repo either won't change or if they change, the old ones will be inaccessible, so there isn't really much reason to store links per version. Same with description, authors, README, LICENSE, etc.

carols10cents:
Something's still not quite right navigating around versions and getting doc links for versions, and it might be due to our misuse of ember in general, i'm not sure... here's how to reproduce what I'm seeing:

- Run this code against the production api with `yarn run start:live`
- Visit http://localhost:4200/crates?page=2
- Click on the accelerate-provider crate to end up on http://localhost:4200/crates/accelerate-provider
- The documentation link on the crate page correctly goes to https://docs.rs/accelerate-provider/0.2.1/, 0.2.1 is the latest version
- Click on one of the other versions, like version 0.2.0
- You'll now be on http://localhost:4200/crates/accelerate-provider/0.2.0, but the documentation link on this page still goes to https://docs.rs/accelerate-provider/0.2.1/, using the latest version
- If you reload http://localhost:4200/crates/accelerate-provider/0.2.0, then the documentation link correctly uses 0.2.0

Let me know if you have any questions or if this doesn't make sense or if you'd like help investigating why this is happening :)

kureuil:
Good catch on the version change! I'll investigate this next week once I'm back from vacation.

kureuil:
Finally took time to resolve this last issue. It should now work properly :)

Bonus feature: Any crate that has a docs.rs link as documentation automatically gets versioning on it (i.e: If I navigate to the lazy_static crate page and change version, the documentation is updated for the current version, because the crate specified `https://docs.rs/crate/lazy_static` as its documentation link)

kureuil:
I'll take care of updating the cargo docs in the coming days :)

jplatte:
According to <http://doc.crates.io/manifest.html#the-documentation-field-optional> this should be in effect, right? I published my first crate today, but after about an hour I still don't see a documentation link on the [crates.io page](https://crates.io/crates/libproxy). docs.rs seems to have built the documentation just fine though.

carols10cents:
@jplatte I'm seeing the documentation link on libproxy's page:

screen shot 2017-06-17 at 12 02 32 pm

What browser are you using? Can you open developer tools to the network tab, reload the page, and see if there's a request to docs.rs as shown in this screenshot, and if that request succeeded or not?

Are you running any plugins in your browser that might be blocking the request, such as a privacy tool like EFF's privacy badger, an adblocker, anything like that?

jplatte:
Ohhh, yeah my browser was blocking the request... I forgot that crates.io is rendered client-side.

PR #687, opened 2017-04-26
Organize docs

### LINKED ISSUE #667: Document in the README instructions for backend setup that `cmake` is also required
At least on OSX. (and it can be installed from homebrew there).

I don't think I ever ran into this because I had cmake installed for something else. Travis definitely apt-gets it though.

carols10cents:
At a recent OSS event that I helped out with, [we discovered](https://github.com/rust-lang/crates.io/issues/667) that the setup instructions were missing some things and were organized in a confusing way.

I wanted some room to elaborate, so I moved the dev setup instructions and the mirror setup instructions into their own files.

I also removed the instructions that the init-local-index.sh script printed out, since the instructions for setting up cargo there are confusing (ex: you should not put the `.cargo/config` in your crates.io checkout and then try to build crates.io, which isn't clear) and out of date.

Fixes #667.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,180 +5,41 @@
 Source code for the default [Cargo](http://doc.crates.io) registry. Viewable
 online at [crates.io](https://crates.io).
 
-This project is built on ember-cli and cargo, visit
-[iamstef.net/ember-cli](http://ember-cli.com/) or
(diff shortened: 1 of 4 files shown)


alexcrichton:
👍

PR #689, opened 2017-04-28
Added a base font size

rap2hpoutre:
I just added a base font size, because it seems there is none right now and it results in a weird display on different browser for the inline code. Here is a screenshot of firefox (same with chrome):

capture d ecran 2017-04-28 a 16 37 57

You can see the `rustc` and `cargo new` are really smaller and there is no reason(?). I think a base font size for the site may be better. Here is a screenshot with base font-size:

capture d ecran 2017-04-28 a 16 46 09

It seems more readable. Sorry if this PR is useless (I'm not sure it's my first PR here and I'm not a CSS master)

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -26,6 +26,7 @@ html, body { margin: 0; }
 
 body {
   font-family: "Helvetica Neue","Helvetica",Helvetica,Arial,sans-serif;
+  font-size: 16px;
   @include display-flex;
   @include flex-direction(column);
   @include align-items(center);


carols10cents:
This PR is not useless! I'm not a CSS or typography expert either, so I did a little research... it looks like the recommended practice for body font size has changed over the years and browser versions, but [this Google PageSpeed Insights reference updated in 2015](https://developers.google.com/speed/docs/insights/UseLegibleFontSizes) at least recommends exactly what you've done here ❤️ 

BUT.... there's a catch. I see that your screenshots include content from doc.crates.io. And I regret to inform you...

Our princess is in another castle

The content for doc.crates.io actually lives in [cargo's repo](https://github.com/rust-lang/cargo/tree/master/src/doc) for Reasons™, and this is terrible because we have to keep the HTML, docs menu, and CSS in sync between the two repos :( 

Someday hopefully we'll figure out a solution for this, but for now, could you also submit this same change to [all.css](https://github.com/rust-lang/cargo/blob/master/src/doc/stylesheets/all.css) in cargo's repo as well? Sorry sorry sorry

PR #691, opened 2017-04-28
Make crates.io publishable on crates.io

carols10cents:
These changes should make the cargo-registry crate publishable on crates.io, which I want to do to make the docs about crates.io's code available on docs.rs to help people working on crates.io.

This will also involve publishing the crates in src/s3 and src/migrate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,48 +1,11 @@
 [root]
-name = "cargo-registry"
+name = "cargo-registry-s3"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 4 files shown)


carols10cents:
> Wanna switch to using a workspace as well to avoid the multiple target dirs?

ooooooo good idea!

carols10cents:
All workspaced up! Still good with this @alexcrichton ? ❤️

alexcrichton:
👍

PR #695, opened 2017-04-30
Reserve windows crate names.

withoutboats:
These names will break cargo on windows machines if uploaded to
the registry.

### DIFF
--- a/migrations/20170430202433_reserve_windows_crate_names/down.sql
+++ b/migrations/20170430202433_reserve_windows_crate_names/down.sql
@@ -1,0 +1,4 @@
+DELETE FROM reserved_crate_names WHERE name IN
+    ('nul', 'con', 'prn', 'aux', 'com1', 'com2', 'com3', 'com4',
+    'com5', 'com6', 'com7', 'com8', 'com9', 'lpt1', 'lpt2',
+    'lpt3', 'lpt4', 'lpt5', 'lpt6', 'lpt7', 'lpt8', 'lpt9');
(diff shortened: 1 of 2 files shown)


withoutboats:
NOTE: I do not have a working dev environment for crates.io and I have **NOT** tested these migrations.

withoutboats:
(also I believe crates.io's namespace is case insensitive let me know if that's wrong)

est31:
Also worth testing is whether crates.io has name length restrictions, as too long names could trigger similar situations.

carols10cents:
@est31 while i'm fixing this, could you create an issue to investigate long names?

est31:
@carols10cents sure! #696

carols10cents:
Staging deploy complete, deploying to production now.

carols10cents:
Production deploy done, and crate removed from the index. Please open new issues for any other problems. Thanks!

kurtseifried:
This is covered by CWE-67 https://cwe.mitre.org/data/definitions/67.html and may need a CVE, I assume private rust repos are a possibility?

jpluimers:
Based on https://msdn.microsoft.com/en-us/library/aa578688.aspx you might want to avoid `CLOCK$` and drive letters like `A:` as well. 

Various other sources confirm that for instance https://support.microsoft.com/en-us/help/74496/ms-dos-device-driver-names-cannot-be-used-as-file-names

withoutboats:
@jpluimers crate names cannot contain either $ or :.

jpluimers:
@withoutboats thanks.

PR #697, opened 2017-05-01
Allow users to create multiple API tokens

### LINKED ISSUE #688: Allow generating multiple authentication tokens
I use Travis CI for publishing all my crates to avoid any chance of weird local settings influencing the package that gets published. At the moment that means all my repositories are sharing the one authentication token that crates.io makes available. I have accidentally leaked and had to revoke this token multiple times when trying to improve my CI scripts, this then provokes a round of replacing the token in all the repositories that use it. If I was instead able to generate a new token for each repository then I would be able to just revoke the one token that was leaked and update that one repository.

🌟 Bonus points for being able to restrict the tokens to only have access for publishing a subset of my crates, but I assume that would probably come in a second feature request after multiple full-access tokens are available.

Nemo157:
Fixes #688

Summary of changes:

 * Database:
   * Add new `api_tokens` table that stores the named tokens, their creation date and last used date

 * API:
   * Remove singular `api_token` field from `User`, it is still in the database for easy rollback, but should be removed from there as well soon.
   * Remove existing `/me/reset_token` endpoint.
   * Update user middleware to search incoming tokens via the `api_tokens` table, updating the last used date as it does so, and to record whether the current request was authenticated via an API token or a session cookie
     * This also changes this search to function via `diesel` where it was previously using `pg`
   * Add new set of endpoints at `/me/tokens` allowing creating, listing and revoking API tokens
     * Listing tokens doesn't return the tokens value, that's only available once in the response from creating the token.
   * Tests

 * Frontend:
   * Remove the special support for an `api_token` parameter coming back from the `/me` endpoint
   * Add new `api-token` model
     * Has a special `adapter` as the endpoints are namespaced under `/me/tokens` instead of the default `/api/v1/api_tokens`
     * Has a special `serializer` to set the key used when creating a new API token
   * Update `/me/index` controller and template
     * Render a sorted list of all API tokens in the data store instead of the single token
     * Remove button to reset the token
     * Add button to start creating a new API token
   * Add new `api-token-row` component for rendering the API tokens in
     the list on `/me/index`
     * Has three major states:
       * New token that has yet to be saved to the API, shows a text box to enter a name and button to save the token
       * Newly saved token, shows the normal token data + the actual token value and details on how to save it with cargo
       * Normal token, shows the name, creation date and last used date along with a button to revoke
     * In any of those states can have an error message shown below the token data if creating or revoking the token failed
   * Some small general CSS changes:
     * A small variant for buttons that reduces the padding
     * Support for disabled state on buttons making them a little lighter than the specified color

<details><summary>old description</summary>
WIP for #688, still has a bit of work to go, but it's currently working and stylish enough that feedback would be useful.

I took inspiration from GitHub's "Personal Access Tokens" around showing the token value only once along with the last usage of the token. 

**TODOs**
 - ~Put one of the existing api tokens back in when reverting the SQL migration (not really sure if this is worth doing)~
 - [x] Remove deletion of `user.api_token` column
 - [x] Limit length of body read during POSTs (potential DOS)
 - [x] Limit the number of tokens a single user can have
 - [x] Make the name required ~(need something to set as the name of the original token)~
 - [x] Put spinner in place of 'Revoke' button during revocation (not sure how to store the metadata needed for this with Ember, maybe the rows of the table need to be components so they have their own controller that knows when it's being revoked)
 - [x] Test with non-instantaneous API (need to inject some delays as localhost is too quick for the creating token spinner to even appear)
 - [x] *Maybe* add a small permissions model so that API tokens don't have permission to create a new API token.
 - [x] Re-add error handling on actions
 -  ~Warning before revocation~
 - [x] Resurrect user tests, add new token tests
    - [x] test `last_used_at` changes when using token
</details>

<details><summary><b>Screenshots</b></summary>

The settings page
screen shot 2017-05-01 at 3 58 36 pm

Creating a new token
screen shot 2017-05-01 at 3 59 17 pm

Showing the newly created token
screen shot 2017-05-01 at 3 59 49 pm

</details>


Nemo157:
This is my first time using Ember.js, so if I've completely misunderstood how some parts of this should be structured just give me a pointer to how it should be structured and I'll fix it up. The big thing that seems the most likely to be non-idiomatic to me is manually fiddling around with the `api_tokens` array, but since this is mostly a read-only website I couldn't find any example of how that should be handled.

carols10cents:
The screenshots look awesome!!!

Some thoughts:

> Put one of the existing api tokens back in when reverting the SQL migration (not really sure if this is worth doing)

I don't think this is worth doing, no. Honestly, the only time I use down migrations are in development. To make production a bit less risky, though, could you remove [the line in the up migration that drops the `api_token` column from `user`](https://github.com/rust-lang/crates.io/pull/697/commits/0ab7ddba12054f16cb9c8009f05fea3a31f12857#diff-9678a4bdff78e23191d4252ba1038138R16)? That way, if we decide to roll back this code when deploying, we don't actually need to roll back the database as well. Then in a separate, future deploy, we can remove that column once we're sure it's no longer being used. 

> Make the name required (need something to set as the name of the original token)

I think empty string, that you have here, is fine.

migrations/20170428154714_multiple_api_tokens/up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE api_tokens (
+  id SERIAL PRIMARY KEY,
+  user_id integer NOT NULL REFERENCES users(id),
+  token character varying DEFAULT random_string(32) NOT NULL UNIQUE,
```
Nemo157: The random_string function is using a 62 character dictionary, giving a 32 character string [~190 bits of randomness](https://www.wolframalpha.com/input/?i=log2(62%5E32)). That makes it more collision resistant than a UUID, and UUIDs are already collision resistant enough that you're recommended not to ever bother checking for a collision.

Using an approximation to the [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem) the chance of having two colliding tokens in the first 10,000 users is [~2.2×10<sup>-50</sup>](https://www.wolframalpha.com/input/?i=1+-+e%5E((-(10000%5E2))+%2F+(2+*+62+%5E+32))), if every person on Earth were to create two accounts the chance of any two colliding is [~5×10<sup>-38</sup>](https://www.wolframalpha.com/input/?i=1+-+e%5E((-(15000000000%5E2))+%2F+(2+*+62+%5E+32))).

But yeah, if `random_string` somehow does create a duplicate it would raise a constraint violation, which would presumably be returned as an error in [`token::new`](https://github.com/rust-lang/crates.io/pull/697/files#diff-8eaf0fafad7d2ed9ec9a701c04424c96R75), which I assume `conduit` or some sort of middleware would transform to a 500 error response. Once I re-add the client-side error handling that should then be shown to user somehow, even if it's just a generic please try again message.

carols10cents:
> Limit length of body read during POSTs (potential DOS)

This sounds like a good idea.... and something that I don't think we're doing anywhere right now, oops....

Along these lines, wdyt about putting a limit on the number of tokens one user can have? Something high by reasonable standards (100? 500? 1000?) but not unlimited?

carols10cents:
What's the reason to require a token name? We're going to have tokens with empty string names in the database from the current tokens anyway, and if people don't want to name their tokens that's their choice.

Nemo157:
> This sounds like a good idea.... and something that I don't think we're doing anywhere right now, oops....

I think it is actually checked everywhere right now, `krate` and `uploader` seem to be the only modules that read the request body, and they both obey `app.config.max_upload_size` which appears to be 10MB.

Actually, looking at those `krate` seems to assume there will be a `Content-Length` header sent, which is probably a valid assumption here for a POST from a browser; so that will be super easy to implement.

> Along these lines, wdyt about putting a limit on the number of tokens one user can have? Something high by reasonable standards (100? 500? 1000?) but not unlimited?

Sure, that should be easy to add. 

> What's the reason to require a token name? We're going to have tokens with empty string names in the database from the current tokens anyway, and if people don't want to name their tokens that's their choice.

Since the actual token value is only returned during creation the only distinguishing factor for unnamed tokens is the date they were created. I guess it's not strictly necessary, but it seems better UX to force the user to add _something_ there so they can distinguish them later. GitHub even goes so far as to disallow having duplicate names for its tokens.

Nemo157:
I've been thinking about

> a small permissions model so that API tokens don't have permission to create a new API token

what's your opinion on adding something really simple like

```rust
enum AuthenticationMode {
    SessionCookie,
    ApiToken,
}

impl<'a> RequestUser for Request + 'a {
    fn auth_mode(&self) -> CargoResult<AuthenticationMode> {
        self.extensions().find::<AuthenticationMode>().chain_error(|| Unauthorized)
    }
}
```

and setting it during the `user::middleware` processing?

It probably wouldn't scale for any other permissions, but if API tokens ever get limited permissions those could be reused for denying API tokens to create new tokens and this could be dropped.

Nemo157:
I'm pretty happy with where this is now, I'll squash everything that's there now down to a single commit  tonight (~10 hours from now); I don't think there's anything really useful in the history. Then I'll take another look at the two TODOs on the diesel code for the user lookup by api token, but the rest of it is done other than review feedback.

carols10cents:
Hi @Nemo157, sorry this has taken me so long to look at but I'll be reviewing it in the next few days hopefully!

Nemo157:
I just rebased this onto master, there was only one significant change which was that the new `UserShowResponse` ended up identical to the `MeResponse` in the tests so I merged them. We can just consider `/me` to be an alias of `/users/<my user id>`.

carols10cents:
Thank you and I'm sorry this has taken me so long!!! I'm trying this out right now :)

carols10cents:
Ok 2 small things:

- Could we make it so that hitting enter while in the token name field submits the form?
- Could we add a label for the migrated tokens? It looks a bit empty without one, and since we don't allow creation of tokens without labels, the migrated ones sort of imply that you can create a token without a label... maybe like `Initial token` or something?

Everything seems to be working great with the permissons and revoking the tokens!

carols10cents:
Also we added rustfmt, so I rebased and ran it on your changes, which should get the build passing again :)

Nemo157:
Cool, made those changes and re-ran the migration to check it worked.

carols10cents:
Hmmmmm ok one more thing: I see you have `autofocus` on the input field for a new token's name, but when I click the new token button, I'm not getting the cursor automatically focused in the name field, in either Firefox 54.0 or Chrome 59.0.3071.104 on macOS. Is it working for you?

Nemo157:
That is odd, it autofocuses for me with Safari 9.1.2 (I know, I can't update) but doesn't with Chrome 51.0.2704.10 or 59.0.3071.104 on macOS. By the sounds of https://github.com/turbolinks/turbolinks-classic/issues/365 and it's fix Firefox (and now Chrome) don't autofocus on injected elements. Added manual focussing to support browsers that don't autofocus.

migrations/20170428154714_multiple_api_tokens/up.sql
sgrif: This is super niche, but can we use `text` instead of `varchar` here? They're identical under the hood, but are considered different types and specifically have conversion issues if you were to use `array_agg` or similar.

src/schema.rs
```diff
@@ -132,7 +143,6 @@ table! {
         id -> Int4,
         email -> Nullable<Varchar>,
```
sgrif: This made me realize that we're using `varchar` all over the place. :sweat:

src/token.rs
```diff
@@ -0,0 +1,221 @@
...
+            .filter(api_tokens::user_id.eq(user_id))
+            .count()
+            .get_result::<i64>(conn)
+            .map(|count| count as u64)
+            .map_err(From::from)
+    }
+
+    /// Converts this `ApiToken` model into an `EncodableApiToken` for JSON
+    /// serialization.
+    pub fn encodable(self) -> EncodableApiToken {
```
sgrif: Is this method separate from `encodable_with_token` for security reasons? What do you think about separating them into separate types to make future mistakes less likely, and further clarify intent?

src/token.rs
```diff
@@ -0,0 +1,221 @@
...
+            id: self.id,
+            name: self.name,
+            token: Some(self.token),
+            created_at: ::encode_time(self.created_at),
+            last_used_at: self.last_used_at.map(::encode_time),
+        }
+    }
+}
+
+struct BadRequest<T: CargoError>(T);
```
sgrif: Should this be in a file focused on HTTP responses?

src/token.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+fn bad_request<T: CargoError>(error: T) -> Box<CargoError> {
+    Box::new(BadRequest(error))
+}
+
+/// Handles the `GET /me/tokens` route.
+pub fn list(req: &mut Request) -> CargoResult<Response> {
+    let tokens = ApiToken::find_for_user(&*req.db_conn()?, req.user()?.id)?
```
sgrif: What do you think about pulling these out into local variables? I personally find the condensed punctuation hard to read

src/token.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+    Ok(req.json(&R { api_tokens: tokens }))
+}
+
+/// Handles the `POST /me/tokens` route.
+pub fn new(req: &mut Request) -> CargoResult<Response> {
+    /// The incoming serialization format for the `ApiToken` model.
+    #[derive(RustcDecodable, RustcEncodable)]
+    struct NewApiToken {
+        pub name: String,
```
sgrif: Nit: Does `pub` do anything here?

src/token.rs
sgrif: Also -- reading further down, this is only ever passed `human`. Should it just take a `String` instead until there's evidence that it needs to take something else?

src/user/mod.rs
```diff
@@ -103,16 +102,18 @@ impl User {
...
-            "SELECT * FROM users \
-                                      WHERE api_token = $1 LIMIT 1",
-        )?;
-        let rows = stmt.query(&[&token])?;
-        rows.iter()
-            .next()
-            .map(|r| Model::from_row(&r))
-            .chain_error(|| NotFound)
+    pub fn find_by_api_token(conn: &PgConnection, token_: &str) -> CargoResult<User> {
+        // TODO: This should be doable as a single query, but I can't figure
```
sgrif: What's the SQL that you would write to do this as a single query? I don't see any way that you could do this in one query.

src/user/mod.rs
```diff
@@ -419,54 +391,3 @@ pub fn updates(req: &mut Request) -> CargoResult<Response> {
         meta: Meta { more: more },
     }))
 }
-
-#[cfg(test)]
```
sgrif: Does this PR make these tests invalid? It doesn't seem like the behavior they're testing has changed, and I'm not sure where these tests moved to

src/token.rs
Nemo157: Sort of for security reasons. The HTTP API is following the common practice of only returning the token value once when creating the token. I'm pretty ambivalent about separating this into two types or not; if there were a way to specify one as a superset of the other + an extra field then it would definitely be the better way to go, but since it currently requires manual synchronisation if a field is added I see downsides to both approaches.

src/token.rs
Nemo157: It's a bit of a hack to support the Ember frontend. My preference would be to drop this and just [have `human` return errors with the status code `BadRequest`](https://github.com/rust-lang/crates.io/issues/712), but [`cargo` does not handle that well](https://github.com/rust-lang/cargo/issues/3995). I could move this to `util::errors` and add a comment about what it's intended use is.

src/token.rs
Nemo157: Sure.

src/token.rs
Nemo157: Nope, not sure where that came from...

src/user/mod.rs
Nemo157:
```SQL
UPDATE api_tokens
SET last_used_at = now()
FROM users
WHERE api_tokens.token = $1
  AND users.id = api_tokens.user_id
RETURNING users.*
```

src/token.rs
sgrif: Do you think fields are likely to be added to this type frequently enough that we need to be concerned about synchronization?

src/user/mod.rs
sgrif: Hm intersting. I didn't realize you could set `FROM` on an update like that. I'll look into providing an API. I think we can ✂️ this comment though, the answer is "you can't"

src/user/mod.rs
Nemo157: Nope, I can readd and update them. They just seemed very low value tests with the tokens pulled out to be a separate resource.

src/token.rs
Nemo157: No, I'll separate them out now so we can at least see what that looks like.

src/user/mod.rs
sgrif: Looking at the [commit which added them](https://github.com/rust-lang/crates.io/commit/9ed51dd2) they seem to be testing something very specific which is still relevant here

Nemo157: Actually, I believe the intent of the first test is covered under `create_token_multiple_have_different_values` in `src/test/token`. The second one I can try and come up with an equivalent with the new data structure.

carols10cents:
I'm really really really sorry this has taken me so long. I'm feeling ready to get this merged in-- just testing it out one last time ❤️

carols10cents:
Merged and deployed to prod, it seems to be doing well!! I'm so excited about this change, thank you so much for your patience!!!

PR #698, opened 2017-05-01
Delete crate script should delete categories and badges too

carols10cents:
When deleting the `nul` crate, I couldn't use the delete crate script for a number of reasons, one being that there were categories and when I added categories I forgot to add deleting them to this script :(


sgrif:
What do you think about cascading the deletion at the database level?

carols10cents:
I'd be into that. Would make this script a lot simpler!

alexcrichton:
Oh er a db removal sounds fine, I'll also merge that sort of PR!

carols10cents:
Meh, i'll make an issue to remind me :)

PR #699, opened 2017-05-01
Disable name trigger temporarily

carols10cents:
This builds on top of #698 but I wasn't sure how you'd feel about this, so I wanted to submit a separate PR.

What happened with the `nul` crate was that I deployed #695 to make the Windows files reserved crate names first, and then I tried to delete the crate.

But the keywords and categories tables have triggers that touch the crates table to set `updated_at`, and crates has a trigger to check for reserved crate names on update, which fails once a crate has a reserved name.

So I added a wrapper to disable the reserved crate name trigger around the two statements that need it, and re-enable the trigger right after, within the transaction that the whole delete crate script uses.

This does lock the table until the transaction is rolled back or committed.

I'm also doing this unconditionally, and this situation will hopefully be pretty rare. I'd be up for adding code that checks if the name of the crate being deleted is in the reserved crate name table and only adding these statements if needed, if that sounds better.

### DIFF
--- a/src/bin/delete-crate.rs
+++ b/src/bin/delete-crate.rs
@@ -73,6 +73,10 @@ fn delete(tx: &postgres::transaction::Transaction) {
                        &[&krate.id]).unwrap();
     println!("  {} deleted", n);
 
+    println!("disabling reserved crate name trigger");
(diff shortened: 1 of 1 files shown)


alexcrichton:
👍

PR #707, opened 2017-05-02
Remove migration crate that has been replaced with diesel

carols10cents:
I was going to publish the migration crate, but then I remembered that since we added diesel, the `migrate` binary has just ensured that crates.io is migrated to use diesel and then it tells you to run `diesel migration run` instead.

Staging and production have long since been migrated, and anyone else that we don't know about who is running crates.io has had [about 2 months](https://github.com/rust-lang/crates.io/pull/589) to get the message, which I think is plenty.


alexcrichton:
🎊

PR #710, opened 2017-05-03
Lock down some crate versions; update others semver compatibly

carols10cents:
In order to be able to build successfully without the lockfile, which
happens when verifying with `cargo package` since [package never includes
`Cargo.lock`](https://github.com/rust-lang/cargo/issues/2263).

If we don't constrain these crates' versions, they update to
incompatible versions of, most problematically, serde_json.

openssl has also deprecated the `finish` method and changed some types
in a minor version bump. I don't feel like figuring it out right now so I'm locking it down too :P

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,6 +99,7 @@ dependencies = [
  "oauth2 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "openssl 0.9.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "postgres 0.14.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "postgres-shared 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "r2d2 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "r2d2-diesel 0.12.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


alexcrichton:
Hm I'm generally quite anti-`=`-dependencies, I'll look into fixing these.

alexcrichton:
I think [this is the problem](https://github.com/sfackler/rust-postgres/blob/413d1db5cde427ed9b2023ee800dd5823005bddf/postgres-shared/Cargo.toml#L30), Cargo doesn't know that it needs to use serde_json 0.9 there, so postgres-shared is getting serde_json 1.0 while we're getting serde_json 0.9, causing compile issues.

alexcrichton:
cc @sfackler (on previous comment), just to make sure you're aware, not sure you want to take any action though

alexcrichton:
Ok after https://github.com/rust-lang/crates.io/pull/711 let's just solve it here. @carols10cents my theory makes me think that we can get by locking *just* postgres-shared, or basically finding a version of postgres that doesn't have the wider range of deps on serde. Did you find that locking openssl/serde/etc were required as well?

If not, could you back out the other locked versions here and just lock postgres-shared, and then place a comment above it as to why it's locked and when the lock can be removed? https://github.com/rust-lang/crates.io/pull/711 should have an example of dealing with the OpenSSL change as well. You can also just `#[allow(deprecation)]`

carols10cents:
@alexcrichton you're totally right, i just tried locking down the others first and that didn't work so I kept adding and didn't go back to remove.

PR #711, opened 2017-05-03
Update dependencies

alexcrichton:
Also fix crates.io deprecation warning


alexcrichton:
Hm no this isn't actually solving the problem.

PR #715, opened 2017-05-05
Add category for built utils

dathinab:
Gives utilities for build scripts there own category. 
For example a utility which helps you pack resources into your binary can fall into this category.

### Open Questions
- is the name ok?
- maybe restrict it to build script utilities only?


carols10cents:
Seems good to me! Thank you!! ❤️

Issue #625, opened 2017-03-14
Unique constraint teams_github_id_key violated


alexcrichton:
Saw some log messages last night:

```
 Mar 13 14:37:31 crates-io app/web.2: ERROR:conduit_log_requests: 127.0.0.1:36656 [2017-03-13T21:37:31Z] Put /api/v1/crates/unbase/owners - 982ms 500: Error reported by Postgres Error reported by Postgres: ERROR: duplicate key value violates unique constraint "teams_github_id_key"

Mar 13 14:37:40 crates-io app/web.2: ERROR:conduit_log_requests: 127.0.0.1:36799 [2017-03-13T21:37:40Z] Put /api/v1/crates/unbase/owners - 547ms 500: Error reported by Postgres Error reported by Postgres: ERROR: duplicate key value violates unique constraint "teams_github_id_key"

Mar 13 14:37:58 crates-io app/web.2: ERROR:conduit_log_requests: 127.0.0.1:37307 [2017-03-13T21:37:58Z] Put /api/v1/crates/unbase/owners - 1347ms 500: Error reported by Postgres Error reported by Postgres: ERROR: duplicate key value violates unique constraint "teams_github_id_key"

Mar 13 14:38:15 crates-io app/web.1: ERROR:conduit_log_requests: 127.0.0.1:49664 [2017-03-13T21:38:14Z] Put /api/v1/crates/unbase/owners - 935ms 500: Error reported by Postgres Error reported by Postgres: ERROR: duplicate key value violates unique constraint "teams_github_id_key"

Mar 13 14:38:18 crates-io app/web.1: ERROR:conduit_log_requests: 127.0.0.1:49741 [2017-03-13T21:38:18Z] Put /api/v1/crates/unbase/owners - 508ms 500: Error reported by Postgres Error reported by Postgres: ERROR: duplicate key value violates unique constraint "teams_github_id_key"

Mar 13 14:42:28 crates-io app/web.1: ERROR:conduit_log_requests: 127.0.0.1:54280 [2017-03-13T21:42:27Z] Put /api/v1/crates/unbase/owners - 215ms 500: Error reported by Postgres Error reported by Postgres: ERROR: duplicate key value violates unique constraint "teams_github_id_key"

[...]

PR #720, opened 2017-05-12
Upsert based on GitHub team ID in case of org/team renames

carols10cents:
The GitHub team ID is unique and stable, so we look up teams based on
that, but the `github:org:team` login value, the team name, and the team
avatar might have changed so we should update that.

Fixes rust-lang/crates.io#446.
Fixes rust-lang/cargo#4019.
Fixes rust-lang/crates.io#625.

r? @sgrif because diesel?


Gankro:
👍

Issue #722, opened 2017-05-13
Keywords aren't being lowercased when inserted with diesel :(


carols10cents:
Problem is here where `NewKeyword`s that get inserted aren't using the lowercased keyword: https://github.com/rust-lang/crates.io/blob/b41061fb80ab8e77a0f83ef9d01da1db01de8ae2/src/keyword.rs#L66

Discovered while trying to convert more stuff to use diesel.

Tests didn't catch it because the tests are using the [non-diesel way of inserting keywords](https://github.com/rust-lang/crates.io/blob/b41061fb80ab8e77a0f83ef9d01da1db01de8ae2/src/keyword.rs#L78), which [does lowercase them](https://github.com/rust-lang/crates.io/blob/b41061fb80ab8e77a0f83ef9d01da1db01de8ae2/src/keyword.rs#L87) :(

* https://github.com/rust-lang/crates.io/blob/b41061fb80ab8e77a0f83ef9d01da1db01de8ae2/src/tests/keyword.rs#L48
* https://github.com/rust-lang/crates.io/blob/b41061fb80ab8e77a0f83ef9d01da1db01de8ae2/src/tests/all.rs#L300

This is a problem in production, for example the keywords 2d and 2D: https://crates.io/keywords?sort=alpha

* 2d 37 crates
* 2D 14 crates

We'll need to collapse these back down to the lowercased version once this is fixed.

PR #723, opened 2017-05-13
Switch most of badges and keywords to use diesel

carols10cents:
Also a fix for #722.

### DIFF
--- a/src/keyword.rs
+++ b/src/keyword.rs
@@ -1,5 +1,4 @@
 use std::ascii::AsciiExt;
-use std::collections::HashMap;
 use time::Timespec;
 
 use conduit::{Request, Response};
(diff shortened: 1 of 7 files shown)


src/keyword.rs
```diff
@@ -60,58 +56,26 @@ impl Keyword {
         struct NewKeyword<'a> {
             keyword: &'a str,
         }
-        sql_function!(lower, lower_t, (x: ::diesel::types::Text) -> ::diesel::types::Text);
```
sgrif: I tend to keep these definitions close to where they're used for functions which can have varying signatures, but I don't feel strongly about it.

carols10cents: Yeah I don't think `lower` is going to have a varying signature.

sgrif: It varies due to nullability (which is why I don't expose it in Diesel proper)

PR #725, opened 2017-05-13
Adjust db connections

carols10cents:
Staging was using 20/20, prod was using 122/120, due to the upgrade in r2d2 with this change? https://github.com/sfackler/r2d2/commit/df5eb7422f2b41d5c55d8d93f86c17c073e2d727

This change makes staging use 10/20 on boot and 16/20 after the first request, seems better.


sgrif:
Lgtm. We should change this to use env vars at some point.

Issue #724, opened 2017-05-13
GitHub profile link sometimes works


frewsxcv:
1. go to https://crates.io/users/frewsxcv
1. github link doesn't work
1. click on a crate
1. click browser back
1. github link works

![untitled](https://cloud.githubusercontent.com/assets/416575/26026842/34435852-37d1-11e7-8f1b-723aa9cb77f0.gif)

nisargthakkar:
@carols10cents I tried looking into what is going on here. It seems that the github url is set in the `pub fn encodable(self) -> EncodableOwner` method in the `src/owners.rs` file. This function is not called on the `users/` route. In the users route, `pub fn encodable(self) -> EncodableUser` method in the `src/user/mod.rs` file is responsible to provide the model for ember js to pick up and display the data. When the user goes to the `/crates` route, the `src/owners.rs` file is hit and it adds the github url in the store. So, the next time the user hits the `/users` route, the url is already present.

I am new to Rust and not very familiar with EmberJS. So, some terminology I've used or my understanding of the code might be totally wrong. Anyways, I'd like to pick this task up

carols10cents:
This was fixed in #727!

PR #727, opened 2017-05-14
Github link in /users route works now

nisargthakkar:
Addresses #724


carols10cents:
Thanks! Sorry this has taken me so long to look at, I will take a look in the next few days hopefully!

carols10cents:
Ugghhhh I tried to add an ember test for this but hitting the user page in a test makes ember say it can't fetch a single user from the store, when it clearly does when I actually use this route???? Aaaanyway, this works when I try it out manually, so I'm merging. Thank you so much!! ❤️

PR #729, opened 2017-05-17
Made the header for exact matches be a link to the crate homepage. Th…

### LINKED ISSUE #683: Make Search Results' "Exact Match" heading a link to the matching result
I just searched for "ndarray" and on the [results page](https://crates.io/search?q=ndarray) I tried clicking the "Exact Match" heading a few times before realizing the link was really below.

It would be nice if the heading was also a link to the matching result.

baileyn:
…is fixes #683.

When displaying search results, the `Exact Match` header will be a link to the crate's homepage. Exactly as if the user clicked the crate's name.

### DIFF
--- a/app/templates/search.hbs
+++ b/app/templates/search.hbs
@@ -49,7 +49,9 @@
     {{#each model as |crate|}}
         {{#if crate.exact_match}}
             <div class='exact-match'>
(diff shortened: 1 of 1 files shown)


carols10cents:
Thank you for your patience @baileyn!!! This is great!!!

PR #730, opened 2017-05-18
Update to serde 1.0, diesel 0.13

sgrif:
There was one test failure due to a change in how serde handles null
values going into `HashMap<String, String>`, but other than that this
appears to be a simple migration.


carols10cents:
Looks good, works on staging-- I'll probably deploy this tuesday :) Thanks!

PR #732, opened 2017-05-19
Enable setting database pool limits from environment variables

carols10cents:
@sgrif I'm trying to test out #730 on staging and I'm overwhelming the db limit of 20 connections there again :( This'll let me set prod and staging differently. I did unify the settings for rust-postgres and diesel though, with the thought that hopefully we'll be getting rid of the rust-postgres connections soon anyway.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::error::Error;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -47,15 +48,33 @@ impl App {
 
         github.scopes.push(String::from("read:org"));
 
+        let db_pool_size = match (env::var("DB_POOL_SIZE"), config.env) {
```
sgrif: Do you think it's worth just removing the `env` defaults and assume that if the var is unset it's in development mode?

carols10cents: I'd like reasonable production defaults as well.

Issue #704, opened 2017-05-02
Add a badge for crate authors to self-report their maintenance intentions


carols10cents:
Part of [RFC 1824](https://github.com/rust-lang/rfcs/blob/master/text/1824-crates.io-default-ranking.md), rust-lang/rust#41616.

Add to the `badges` table in `Cargo.toml` a `maintenance` key that crate authors could use to
self-report their maintenance intentions. The valid values would be along the
lines of the following, and would influence the ranking in the order they're
presented:

<dl>
  <dt>Actively developed</dt>
  <dd>
    New features are being added and bugs are being fixed.
  </dd>

  <dt>Passively maintained</dt>
  <dd>
    There are no plans for new features, but the maintainer intends to respond
    to issues that get filed.
  </dd>

  <dt>As-is</dt>
  <dd>
    The crate is feature complete, the maintainer does not intend to continue
    working on it or providing support, but it works for the purposes it was
    designed for.
  </dd>

  <dt><i>none</i></dt>
  <dd>
    We display nothing. Since the maintainer has not chosen to specify their
    intentions, potential crate users will need to investigate on their own.
  </dd>

  <dt>Experimental</dt>
  <dd>
    The author wants to share it with the community but is not intending to meet
    anyone's particular use case.
  </dd>

  <dt>Looking for maintainer</dt>
  <dd>
    The current maintainer would like to transfer the crate to someone else.
  </dd>

  <dt>Deprecated</dt>
  <dd>
    The maintainer does not recommend using this crate (the description of the crate can describe why, there could be a better solution available or there could be problems with the crate that the author does not want to fix) Added from a suggestion in #549
  </dd>

</dl>

These would be displayed as badges on lists of crates.

These levels would not have any time commitments attached to them-- maintainers
who would like to batch changes into releases every 6 months could report
"actively developed" just as much as mantainers who like to release every 6
weeks. This would need to be clearly communicated to set crate user
expectations properly.

This is also inherently a crate author's statement of current intentions, which
may get out of sync with the reality of the crate's maintenance over time.

Also need to update [the documentation that currently lives in cargo's repo](https://github.com/rust-lang/cargo/blob/8cc22d2c9d9985122f16789ea24cfe6411ae408a/src/doc/manifest.md#package-metadata).

Please let me know if you have any questions, potential implementers!

marti1125:
Hi @carols10cents!! I would like to work on this

marti1125:
[...]

PR #742, opened 2017-05-31
add new badges #704

marti1125:
I added new Badges @carols10cents please review the enums. Thanks!!
And also I had this error =/
https://gist.github.com/marti1125/d2ddda017b9e1c8a4224cf48cc2a42b1

### DIFF
--- a/src/badge.rs
+++ b/src/badge.rs
@@ -44,26 +44,14 @@ pub enum Badge {
 }
 
 #[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
 pub enum MaintenanceValue {
-    #[serde(rename = "actively-developed")]
     ActivelyDeveloped,
-
-    #[serde(rename = "passively-maintained")]
     PassivelyMaintained,
-
-    #[serde(rename = "as-is")]
     AsIs,
-
-    #[serde(rename = "none")]
     None,
-
-    #[serde(rename = "experimental")]
     Experimental,
-
-    #[serde(rename = "looking-for-maintainer")]
     LookingForMaintainer,
-
-    #[serde(rename = "deprecated")]
     Deprecated,
 }
 
--- a/src/badge.rs
+++ b/src/badge.rs
@@ -40,34 +40,33 @@ pub enum Badge {
         branch: Option<String>,
         service: Option<String>,
     },
+    #[serde(rename = "maintenance")]
+    Maintenance { value: MaintenanceValue },
+}
+
+#[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
+#[serde(tag = "badge_type", content = "attributes")]
+pub enum MaintenanceValue {
     #[serde(rename = "actively-developed")]
-    ActivelyDeveloped {
-        repository: String, branch: Option<String>,
-    },
+    ActivelyDeveloped,
+
     #[serde(rename = "passively-maintained")]
-    PassivelyMaintained {
-        repository: String, branch: Option<String>,
-    },
+    PassivelyMaintained,
+
     #[serde(rename = "as-is")]
-    AsIs {
-        repository: String, branch: Option<String>,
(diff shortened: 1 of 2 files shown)


carols10cents:
Hi @marti1125! Thank you for the PR and I'm sorry it's taken me so long to look at this!

Regarding the errors you got, was this when you ran `cargo test` or `cargo build` or `cargo run --bin server` or something else?

Could you run `cargo clean` and then whatever command you ran with `-v` to get verbose output and put everything that outputs into a gist please?

Regarding the code, I think I was unclear when we chatted last. I was imagining one new enum variant for maintenance that would have ActivelyDeveloped, PassivelyMaintained, and so on as the possible values for that variant. Also, this variant would not need the `repository` key; those are used to configure the other badges since crates.io needs to tell services like TravisCI where the project that goes with the crate is, which might not match the crate name. This badge doesn't involve any outside services, and it will just display the value without caring where the project's repository is.

In other words, what I think crate authors should be able to specify in their `Cargo.toml` would look like this (the TOML parser is expecting a `HashMap` of `String` keys mapping to `String` values, so we do need to have some sort of key/value in here even though it won't be `repository`):

```toml
[badges]
maintenance = { value = "passively-maintained" }
```

That will have cargo send this JSON to crates.io when a crate gets published:

```json
{
    "name": "your-crate-name",
    // all the other metadata in here
    "badges": {"maintenance": {"value":"passively-maintained"}}
}
```

Then, when crates.io uses serde to deserialize this JSON, it should match a `Maintenance` variant in the `Badge` enum. A crate will have the `Badge::Maintenance` badge if the maintainer has specified any value, and the crate won't have this badge if the maintainer has not specified any value.

When we go to display the badge, the text of the badge will come from looking up the `value` within that badge.

Does that make sense? Are you up for making that change to this PR? Please ask questions if I'm still not being clear! Thank you!!

carols10cents:
Hi @marti1125, just checking in on this PR-- will you have time soon to make changes to this PR? Thanks!

marti1125:
@carols10cents sorry for the big delay I will restart with this issue tomorrow. I will see you at irc channel 👍

marti1125:
@carols10cents please review my commit thanks!

carols10cents:
Hi! The code needed a few more changes, and I added some tests to make sure everything was working. I also updated [the checklist](https://github.com/rust-lang/crates.io/issues/704#issuecomment-306618745)! Thank you!! 🌞 🌝 🌷

marti1125:
@carols10cents great! 👍

marti1125:
could you explain me more about the others tasks? please 😄

carols10cents:
what would you like to know?

marti1125:
- Add an ember component like the other badge components in app/components
- Add a template in app/templates/components to display the badge

carols10cents:
What questions do you have about these tasks?

marti1125:
- Add an ember component like the other badge components in app/components 
could you tell me what components I have to modify? please

carols10cents:
> could you tell me what components I have to modify?

You do not need to modify any components, you need to create a new component in a file named app/components/badge-maintenance.js. In the app/components/ directory, there are other badge components that will be good examples for what needs to go in that new file.

Does that make sense?

Do you have the frontend of the app running on your local machine? Do you have the backend running on your local machine? If so, have you tried publishing a crate to your local backend? If you are able to do that, you should be able to publish a crate with a maintenance badge in that crate's `Cargo.toml` in order to get the right data in your database in order to test with. If you don't have a backend set up or you have problems publishing a crate to it, another option is to modify the frontend test fixture data in mirage/fixtures/crates.js to add to the fake badges data in, for example, the fake "rust_mixin" crate here: https://github.com/rust-lang/crates.io/blob/master/mirage/fixtures/crates.js#L60 Then you can run the front end with `npm run start` to use the fixture data.

That way, you can better see the results of what you're doing. Does that make sense? 

If not, please ask more questions!!

PR #744, opened 2017-05-31
Update app.scss

inta:
Fore- and background color should both be defined. Currently the search input is somehow broken on systems with a dark theme because black on dark grey is hardly readable.

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -132,6 +132,7 @@ body {
     outline: 0;
     margin-left: 30px;
     padding: 5px 5px 5px 25px;
+    background-color: white;
     background-image: url('/assets/search.png');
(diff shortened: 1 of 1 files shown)


carols10cents:
Thanks! Sorry this has taken me so long-- the code change looks great, I just want to try this out before I merge it and I hope to have time to do that in the next few days.

inta:
No problem, take your time. I just fixed it a few times with the developer tools and thought it would be small but – at least for me – valuable improvement.

carols10cents:
Thank you for your patience! While this is indeed a very small change, I've been bitten in the past by shipping changes that look small without trying them out first, so now I'm overly cautious :) This looks great though, I should be able to deploy this soon!

PR #745, opened 2017-06-03
Introduce a CrateBuilder for tests

shepmaster:


sgrif:
I like this idea, but I don't think we need to call the build method `expect_build`. "panic if this fails" is generally assumed the default for tests.

carols10cents:
I could see a test wanting to just call `build` and do its own custom error handling for some reason. I'd be open to a pull request changing this though, if you feel super strongly about it @sgrif.

I really like this change though, I think it makes the tests lots clearer, and I'm going to merge it so we can start using the builder in more places.

PR #748, opened 2017-06-05
Port version/authors over to Diesel

sgrif:
Since `authors` was only used in one place for this query, I didn't
think it was worth keeping an instance method or even a Diesel
association around.


sgrif:
Rebased. Anything blocking this?

carols10cents:
Looks good!

PR #751, opened 2017-06-05
Refactor paginating queries to a trait

sgrif:
This pattern gets duplicated quite a bit (and it should be duplicated
even more since we still have places where we're doing more than one
query to paginate). Additionally, using the `sql` function will cause a
query to be removed from the prepared statement cache (since Diesel
doesn't know if the string was generated dynamically or not).

This moves it out into a trait, and represents a paginated query in a
bit more of an abstract form from the SQL point of view. This will
slightly change the SQL in the places that it's being used, so that it
looks more like what we have in `src/krate_reverse_dependencies.sql`,
applying the limit/offset as late as possible. This will not affect the
query plan unless they are using `distinct`, or if `limit`/`offset` are
called prior to `paginate`. Even if those cases do happen, this query is
probably what we meant anyway.

### DIFF
--- a/src/keyword.rs
+++ b/src/keyword.rs
@@ -10,6 +10,7 @@ use pg::rows::Row;
 
 use {Model, Crate};
 use db::RequestTransaction;
+use pagination::Paginate;
 use schema::*;
 use util::{RequestUtils, CargoResult};
 
@@ -138,8 +139,6 @@ impl Model for Keyword {
 
 /// Handles the `GET /keywords` route.
 pub fn index(req: &mut Request) -> CargoResult<Response> {
-    use diesel::expression::dsl::sql;
-    use diesel::types::BigInt;
     use schema::keywords;
 
     let conn = req.db_conn()?;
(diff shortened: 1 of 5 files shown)


src/keyword.rs
```diff
@@ -138,28 +139,22 @@ impl Model for Keyword {
...
+    let mut query = keywords::table.into_boxed();
 
     if sort == "crates" {
         query = query.order(keywords::crates_cnt.desc());
     } else {
         query = query.order(keywords::keyword.asc());
     }
 
-    let data = query.load::<(Keyword, i64)>(&*conn)?;
+    let data = query.paginate(limit, offset).load::<(Keyword, i64)>(&*conn)?;
```
carols10cents: It's a little unfortunate that selecting the total number of results is now gone from this location, but we have to still remember to include the `i64` type in the `load` type here. Not sure what to do differently though, maybe if we made a `paginated_load` that only took the types this code knows about and inserts the `i64`? 

Not a huge deal, this PR is still a huge improvement for factoring out duplication and raw SQL everywhere, just a thought for a possible next step and/or if pagination becomes a more general diesel thing.

sgrif: I doubt I'd add it to Diesel, but this could definitely be extracted to a crate. I have considered adding a method that lets you specify the type you're going to deserialize to outside of `load`. (The primary motivation is that `Version::belonging_to(&crate).load::<Version>` bugs me) If that were present, we could definitely have this be `.paginate::<Keyword>` and have the paginate method call that with `(T, i64)`. I haven't explored how much work it'd be to add that method though

PR #755, opened 2017-06-06
Port `crate/versions` over to use Diesel

sgrif:


src/krate.rs
```diff
@@ -1189,11 +1189,13 @@ pub fn following(req: &mut Request) -> CargoResult<Response> {
...
     let crate_name = &req.params()["crate_id"];
-    let tx = req.tx()?;
-    let krate = Crate::find_by_name(tx, crate_name)?;
-    let versions = krate.versions(tx)?;
-    let versions = versions.into_iter().map(|v| v.encodable(crate_name))
-                           .collect();
+    let conn = req.db_conn()?;
+    let krate = Crate::by_name(crate_name).first::<Crate>(&*conn)?;
+    let mut versions = Version::belonging_to(&krate).load::<Version>(&*conn)?;
+    versions.sort_by(|a, b| b.num.cmp(&a.num));
```
carols10cents: This sorting is getting duplicated in multiple places-- do you think we could make a method that uses diesel that behaves like [`krate.versions()`](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L568-L577) does now, that would always sort the versions? Or is there some reason why we wouldn't want that?

sgrif: Normally we'd just do the sort in SQL, it sucks that we can't here. There's no specific reason why we can't have `krate.versions`, I mostly just wanted to move more behavior off that struct since it's already doing so much. Code that relies on the fact that versions are sorted should make it clear that they care about that fact IMO. I liked [the solution in the build info PR](https://github.com/rust-lang/crates.io/pull/764/files#diff-7ab30501521c288e1d8e1e994692b1a2R596)

sgrif: @carols10cents Did you want me to copy the method from that PR? Or can refactoring that wait until both have been merged?

carols10cents: I'm cool with waiting

PR #756, opened 2017-06-06
Port crates/reverse_dependencies over to Diesel

sgrif:
This took a bit more code than I'd like, as `#[derive(Queryable)]`
doesn't like `ReverseDependency` at the moment. The explicit select
clause is because Diesel grabs columns by order not by name.


src/krate.rs
```diff
@@ -736,20 +736,22 @@ impl Crate {
...
+
+        type SqlType = ((dependencies::SqlType, Integer, Text), BigInt);
+        let rows = sql::<SqlType>(include_str!("krate_reverse_dependencies.sql"))
+            .bind::<Integer, _>(self.id)
+            .bind::<BigInt, _>(offset)
+            .bind::<BigInt, _>(limit)
+            .load::<(ReverseDependency, i64)>(conn)?;
+
+        let (vec, counts): (_, Vec<_>) = rows.into_iter().unzip();
+        let cnt = counts.into_iter().nth(0).unwrap_or(0i64);
```
carols10cents: oooh fancy!

src/tests/krate.rs
```diff
@@ -1682,23 +1716,33 @@ fn prerelease_versions_not_included_in_reverse_dependencies() {
...
+    }
 
     let mut response = ok_resp!(middle.call(&mut req));
     let deps = ::json::<RevDeps>(&mut response);
     assert_eq!(deps.dependencies.len(), 1);
     assert_eq!(deps.meta.total, 1);
     assert_eq!(deps.dependencies[0].crate_id, "c2");
 
-    t!(c2v2.yank(req.tx().unwrap(), true));
+    diesel::update(versions::table.filter(versions::num.eq("2.0.0")))
```
carols10cents: AFAIK this part could be changed back to calling `version.yank()` again once yank is converted to use diesel; I'm going to add in a note here because I'd prefer to have it call the method in case we add extra behavior to the yank method someday.

PR #764, opened 2017-06-08
Display reported build information in a badge and on crate version pages

shepmaster:
(This is a replacement of #540)

This goes with https://github.com/rust-lang/cargo/pull/5099. @carols10cents worked on these too :)

This PR adds a table that stores recorded crate version, rust version, target, and pass/fail as reported by crate owners using the `cargo publish-build-info` command.

The purpose is to enable crate authors to automatically report to potential crate users on the Rust version and platform compatibility of each version of the crate.

[There's documentation in the cargo PR](https://github.com/rust-lang/cargo/pull/5099) for doc.crates.io about how this feature is intended to work.

The badges, as described in that documentation, will display the latest version of the most stable channel that reported any `pass` results for the max version of that crate. Here's what that will look like, including the hover text:

Stable:

screen shot 2017-01-30 at 4 39 04 pm

Beta:

screen shot 2017-01-30 at 4 35 44 pm

Nightly:

screen shot 2017-01-30 at 4 40 53 pm

Each crate version page will display more detail, but not all of the possible information that collection supports. It's in a new section above the download stats graph as a table showing the most recent stable, beta, and nightly with any reported results for the Tier 1 platforms using 64 bit architectures:

screen shot 2017-06-08 at 11 53 40 am

There's also a new page with full information for the 64-bit tier 1 platforms, showing more detailed information:

screen shot 2017-06-08 at 1 06 56 pm

Some enhancements to the display that we've thought about but haven't implemented yet:

- Have dropdowns for all the available versions for each channel
- Have a dropdown to select from additional targets
- Show icons or badges or colors instead of or in addition to the table

We also thought about, but didn't implement:

- Ember tests for the UI
- Adding a different API token that would only be valid for publishing build info
- Showing the max build version badge on crate page in the sidebar

We're excited to hear what anyone thinks, and of course happy to answer questions or make modifications!


src/version.rs
```diff
@@ -185,6 +270,32 @@ impl Version {
...
+            rust_version: info.rust_version,
+            target: info.target,
+            passed: info.passed,
+        };
+
+        diesel::insert(&bi.on_conflict(
+            build_info::table.primary_key(),
+            do_update()
+                .set((build_info::passed.eq(excluded(build_info::passed)),
+                      build_info::updated_at.eq(now_utc().to_timespec())))
```
sgrif: `updated_at` is handled by database triggers everywhere else. Should we do the same here? (adding `SELECT diesel_manage_updated_at('build_info');` to your migration will have Diesel set up the trigger automatically.

src/version.rs
```diff
@@ -404,6 +515,143 @@ pub fn downloads(req: &mut Request) -> CargoResult<Response> {
...
+        for row in build_infos {
+            let BuildInfo { rust_version, .. } = row;
+            let rust_version: ChannelVersion = rust_version.parse()?;
+
+            match rust_version {
+                ChannelVersion::Stable(semver) => stable.push(semver),
+                ChannelVersion::Beta(date) => beta.push(date),
+                ChannelVersion::Nightly(date) => nightly.push(date),
+            }
+        }
```
sgrif: Do you think this amount of code is worth it just to avoid the one allocation of an intermediate vec? What do you think about this instead?

```rust
let rust_versions = build_infos.into_iter().map(|row| row.rust_version.parse::<ChannelVersion>()).collect::<Result<Vec<_>, _>>()?;
let nightly = rust_versions.iter().filter_map(|version| match *version {
    ChannelVersion::Nightly(date) => Some(date),
    _ => None,
}).max();
let beta = rust_versions.iter().filter_map(|version| match *version {
    ChannelVersion::Beta(date) => Some(date),
    _ => None,
}).max();
let stable = rust_versions.into_iter().filter_map(|version| match version {
    ChannelVersion::Stable(semver) => Some(semver),
    _ => None,
}).max();
```

src/version.rs
```diff
@@ -68,9 +71,90 @@ pub struct VersionLinks {
...
+#[table_name="build_info"]
+#[primary_key(version_id, rust_version, target)]
+pub struct BuildInfo {
+    version_id: i32,
+    rust_version: String,
+    target: String,
+    passed: bool,
+}
+
+pub const BUILD_INFO_FIELDS: (build_info::version_id, build_info::rust_version, build_info::target, build_info::passed) = (build_info::version_id, build_info::rust_version, build_info::target, build_info::passed);
```
sgrif: Do we even need the `created_at` and `updated_at` fields if we aren't selecting them?

src/version.rs
```diff
@@ -404,6 +515,143 @@ pub fn downloads(req: &mut Request) -> CargoResult<Response> {
...
+                let key = ::encode_time(date);
+                nightlies.insert(date);
+                encodable_build_info.nightly.entry(key)
+                                  .or_insert_with(HashMap::new)
+                                  .insert(target, passed);
+            }
+        }
+    }
+
+    encodable_build_info.ordering.insert(
```
sgrif: Do you think using the `ordermap` crate instead of a hashmap could simplify this code?

src/version.rs
```diff
@@ -452,3 +700,26 @@ fn modify_yank(req: &mut Request, yanked: bool) -> CargoResult<Response> {
     struct R { ok: bool }
     Ok(req.json(&R{ ok: true }))
 }
+
+/// Handles the `POST /crates/:crate_id/:version/build_info` route.
+pub fn publish_build_info(req: &mut Request) -> CargoResult<Response> {
+    let mut body = String::new();
+    try!(req.body().read_to_string(&mut body));
```
sgrif: `?`?

src/krate.rs
```diff
@@ -399,19 +402,33 @@ impl Crate {
...
+            None,
+            None,
+            None,
+            badges,
+            exact_match,
+            max_build_info,
+        )
     }
 
+    #[cfg_attr(feature = "lint", allow(too_many_arguments))]
```
sgrif: Do you think we should listen to this lint here, and start to refactor this out a bit? `encodable` feels like it's becoming a bit of a dumping ground.

src/version.rs
```diff
@@ -68,9 +71,90 @@ pub struct VersionLinks {
     pub dependencies: String,
     pub version_downloads: String,
     pub authors: String,
+    pub build_info: String,
+}
+
+#[derive(Insertable, AsChangeset)]
+#[table_name="build_info"]
+#[primary_key(version_id, rust_version, target)]
+struct NewBuildInfo {
```
sgrif: If we expect this table to be relatively stable, it might be worth just dumping this struct and adding `#[derive(Insertable, AsChangeset)]` onto `BuildInfo`.

src/version.rs
shepmaster: I don't see which part of the proposed code only keeps the most recent value?

sgrif: What do you mean by "most recent"? The code simply grabs the maximum of each one, does it not?

shepmaster: Ah, there's a `max` call at the end of each of those. I'm not the biggest fan of that syntax for that reason. I see what you mean now though.

shepmaster: >  just to avoid the one allocation of an intermediate vec

To be honest, I've never really worked on mission-critical infrastructure supporting the entire ecosystem behind a language, so I don't really know when it's appropriate. From the comment, I assume you think it's not, so I'll defer to your greater experience.

src/version.rs
carols10cents: Ah yes, good point. We should be using the db triggers here.

src/version.rs
carols10cents: I'd like to have them, I could see wanting to display them somewhere, but they're useful to me for auditing purposes.

src/version.rs
carols10cents: I think we'd still want the "stable", "beta", "nightly" keys for the JSON... or am I misunderstanding? How would ordermap instead of a hashmap simplify this code?

src/krate.rs
carols10cents: I do think we should refactor this, I am agnostic about whether it should be done in this PR or in a separate PR

src/version.rs
carols10cents: That sounds good.

src/version.rs
shepmaster: Will the triggers also handle `created_at`?

src/version.rs
shepmaster: `?`!

shepmaster:
I've rebased to address some conflicts and applied rustfmt to each commit. @sgrif I've added [a commit](https://github.com/rust-lang/crates.io/pull/764/commits/4e74433eea46dd44467c1a190ad6011d77e8e111) that does your proposed max finding. What are your thoughts on it?

I'm still going to address these comments:

- https://github.com/rust-lang/crates.io/pull/764#discussion_r120949596
- https://github.com/rust-lang/crates.io/pull/764#discussion_r120953510
- https://github.com/rust-lang/crates.io/pull/764#discussion_r120954231
- https://github.com/rust-lang/crates.io/pull/764#discussion_r120955247

I just wanted to save them from being hidden

src/version.rs
sgrif: We wouldn't need to maintain this `ordering` field, since `OrderMap` preserves insertion order. Even if we still need `ordering` for the front-end, it'd just be `.keys().collect()` to get the corresponding vec.

Though on second reading, that's not even what this code is doing. Since it's in a `BTreeSet`, it's just going to be sorting the strings in ascending order. Is that the desired behavior? It seems like we could just sort them when we're ready to display them in that case.

shepmaster: I've updated the function with an example of the output as well as the rationale behind it. In short: no, I don't thing the ordermap crate would help here.

As far as I understand it, JSON and JavaScript do not guarantee the order of objects. The `ordering` field exists to avoid requiring the frontend to understand how to sort by date and/or semver.

While `OrderMap` preserves insertion order, our iteration order over the returned rows is not guaranteed to be anything useful, therefore the insertion order is not valuable.

We aren't relying on the `BTree` part, just the `Set` part; we could just have easily used a `HashSet`. I only use `BTree*` stuff as a favor to @Gankro, who told me that B-Trees are just all-around-cooler. 

Additionally, the sets don't contain `String`s — the stable one contains version numbers and the others two contain dates. We only convert to a string to serialize.

shepmaster:
Ah, diesel-rs/diesel#991

shepmaster:
All talking points addressed. Ready for next round of review.

src/upload.rs
```diff
@@ -55,6 +57,19 @@ pub struct CrateDependency {
     pub kind: Option<DependencyKind>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct VersionBuildInfo {
```
vignesh-sankaran: Could we add a doc comment here?

src/version.rs
```diff
@@ -70,9 +72,115 @@ pub struct VersionLinks {
...
     pub version_downloads: String,
     pub authors: String,
+    pub build_info: String,
+}
+
+#[derive(Insertable, AsChangeset, Identifiable, Queryable, Associations)]
+#[belongs_to(Version)]
+#[table_name = "build_info"]
+#[primary_key(version_id, rust_version, target)]
+pub struct BuildInfo {
```
vignesh-sankaran: Could we have a doc comment here as well as for all structs, enums, and functions in this file?

carols10cents:
Ok, I've rebased and revived this old thing, in the start of an effort to actually get through all the open PRs.

I'd love reviews from:

- @sgrif to make sure I updated the diesel parts properly
- @jtgeibel to see if I put things in the right spots in the new MVC layout, there are a few things i'm not sure about
- @shepmaster to make sure the original functionality we intended is still there

whenever yinz have time <3

jtgeibel:
@carols10cents I've finally gotten around to looking at this.  In terms of MVC style, I have no issues.

I'm testing this PR in a staging area with the cargo PR as well.  Here are a few things I've run into:

* The crate.png image has been replaced with SVG, but the build-info page still references the PNG.
* I've pushed build info for stable, beta, and nightly.  I'm only seeing the information for stable.
  * At first I pushed nightly build info.  This showed up as a badge on the browse page, but didn't show up on the crate's page.
  * Once I pushed stable build info, the summary showed up on the crate's page, but only for stable.  The summary page and build-info details page are both missing information on beta and nightly.  All the data is present in the API response, so I think this is a frontend issue.

For reference, the staging area is [here](https://jtgeibel-staging-crates-io.herokuapp.com/crates/build-info-test/0.1.0/build_info).

jtgeibel:
@carols10cents I've pushed a commit that should pass your new test.  I'm explicitly formatting as "%Y-%m-%d" and tweaked the test a bit to match.  I'm still reviewing but will probably r+ this weekend.

src/views/mod.rs
```diff
@@ -119,6 +129,145 @@ pub mod krate_publish;
...
+    pub id: i32,
+    pub ordering: HashMap<String, Vec<String>>,
+    pub stable: HashMap<String, HashMap<String, bool>>,
+    pub beta: HashMap<NaiveDate, HashMap<String, bool>>,
+    pub nightly: HashMap<NaiveDate, HashMap<String, bool>>,
+}
+
+/// `MaxBuildInfo` in JSON form.
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct EncodableMaxVersionBuildInfo {
```
sgrif: Every field on `MaxBuildVersionInfo` implements `Serialize`, doesn't it? Is there a reason we need this as a separate struct?

src/version/build_info.rs
```diff
@@ -0,0 +1,129 @@
...
+    version_id: i32,
+    pub rust_version: String,
+    pub target: String,
+    pub passed: bool,
+}
+
+/// The columns to select from the `build_info` table. The table also stores `created_at` and
+/// `updated_at` metadata for each row, but we're not displaying those anywhere so we're not
+/// bothering to select them.
+pub const BUILD_INFO_FIELDS: (
```
sgrif: Are the `created_at` and `updated_at` fields ever used at all? Is it worth even having them in the first place?

src/version/build_info.rs
```diff
@@ -0,0 +1,129 @@
...
+use schema::*;
+
+#[derive(Clone, Identifiable, Associations, Debug, Queryable)]
+#[belongs_to(Version)]
+#[table_name = "build_info"]
+#[primary_key(version_id, rust_version, target)]
+/// Stores information about whether this version built on the specified Rust version and target.
+pub struct BuildInfo {
+    version_id: i32,
+    pub rust_version: String,
```
sgrif: Is this always a semver version? Is it worth pointing at https://github.com/steveklabnik/semver/pull/169 to make things easier here?

vignesh-sankaran:
@jtgeibel @shepmaster Did we still want to merge this in? This probably needs a rework given that the codebase has been refactored since the PR was opened.

carols10cents:
Ugh, yeah, I'm going to close.

PR #769, opened 2017-06-11
Ensure exact match crates are always returned by search

### LINKED ISSUE #321: Searching for "which" does not show the "which" crate
https://crates.io/search?q=which

Crates that should have shown:
- https://crates.io/crates/which
- https://crates.io/crates/quale

sgrif:
The query assumed that if the name was equal to the query string, that
the full text search would always match that crate. This is not true in
the case where the query string is a stop word (words like a, the, just,
which), which are excluded from full text search.

Explicitly adding an or condition to include a crate whose name matches
is always returned solves the issue. We purposely continue to exclude
crates which would only match because their readme contains a stopword,
as this behavior is important for things like stemming to do their job.

Fixes #321.

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -801,7 +801,11 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 
     if let Some(q_string) = params.get("q") {
         let q = plainto_tsquery(q_string);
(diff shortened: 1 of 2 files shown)


alexcrichton:
Needs a rebase but otherwise looks great!

PR #770, opened 2017-06-11
Fix crate/downloads

sgrif:
I appear to have fucked up in #746. The API was sending the detailed
stats for the oldest 5 versions instead of the newest 5 versions,
causing the whole graph to just show "other". This fixes that error.

(as an aside, I kinda prefer just having the whole picture as one stat
like that)

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -1074,7 +1074,7 @@ pub fn downloads(req: &mut Request) -> CargoResult<Response> {
     let krate = Crate::by_name(crate_name).first::<Crate>(&*conn)?;
 
     let mut versions = Version::belonging_to(&krate).load::<Version>(&*conn)?;
-    versions.sort_by(|a, b| a.num.cmp(&b.num));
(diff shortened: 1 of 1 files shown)


sgrif:
@carols10cents This is higher priority than my other PRs, it fixes a bug that is currently deployed

carols10cents:
Looking at #746 more closely, I think there's another difference too: [this](https://github.com/rust-lang/crates.io/pull/746/files#diff-7ab30501521c288e1d8e1e994692b1a2L1097) used to sort by date asc, but [this](https://github.com/rust-lang/crates.io/pull/746/files#diff-7ab30501521c288e1d8e1e994692b1a2R1082) sorts by date desc. I'm going to merge this and make that change then deploy.

Sorry I didn't catch these, and I guess we should have tests for these too!!

sgrif:
Ah crap I should have caught the date sorting too. Thanks!

sgrif:
(And sorry for breaking prod... again...)

carols10cents:
All better now!!

Issue #575, opened 2017-02-24
Add rustfmt to run on travis and fail the build


carols10cents:
After #574 is done, so that we have one commit with all the old code reformatted, I'd like to keep the code formatted nicely going forward by adding rustfmt to Travis like we have jslint that fails the build if the javascript isn't formatted correctly.

So this bug entails:

- Waiting until #574 is done
- Follow [the instructions in rustfmt's README on adding rustfmt to .travis.yml](https://github.com/rust-lang-nursery/rustfmt#checking-style-on-a-ci-server)
- Get the build to pass, if there's code that snuck in between #574 and this one that isn't formatted right
- Add documentation to crates.io's README that we use rustfmt, and point to rustfmt's README for instructions for people working on crates.io to run rustfmt while they're developing if they want

ishitatsuyuki:
rustfmt has awful side effects. Clippy deprecated the approach: https://github.com/Manishearth/rust-clippy/issues/1487

carols10cents:
@ishitatsuyuki Are you talking about the contributor-scaring-away side effects? We already have jslint that fails the build if the javascript isn't formatted right, and I have seen PRs that contributors have had the jslint check fail and then they fix it and push another commit, which hasn't seemed to be a deterrent to contributing. I'd like to try, and I'll be here to help people understand... 

If those aren't the awful side effects you're talking about, could you elaborate please? Thanks!

ishitatsuyuki:
rustfmt could generate wrong, or unintended ugly output. See the clippy issue for details.

carols10cents:
Wrong output sounds like a bug in rustfmt, and more users of rustfmt would help find and fix these bugs. 

Ugly is in the eye of the beholder; what I'm trying to solve here is consistent code formatting.

Unintended sounds like something we could add a rustfmt ignore annotation to. 

I don't see anything egregious in the issues linked in that clippy issue, I remain unconvinced that we shouldn't even try.

ishitatsuyuki:
At least, we would get a failure even when the code is not wrong (only formatting problems), and that looks bad. Clippy's PR list was filled with failed statuses before, which was caused by probably broken formatting in master.

vignesh-sankaran:
Hey, is this still an issue that is still blocked? If so, what's the block?

carols10cents:
I forget why I added the blocked label, tbh!

carols10cents:
Closed by #771!

PR #771, opened 2017-06-12
cargo fmt

### LINKED ISSUE #574: Run rustfmt on the whole codebase and send in the changes

vignesh-sankaran:
Fixes #574. Ran rustfmt over the codebase, went smoothly except for this line here (https://github.com/rust-lang/crates.io/blob/master/src/db.rs#L66), but I think that there isn't much we're going to be able to do about that

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -6,4 +6,4 @@ generics_indent = "Block"
 fn_call_style = "Block"
 combine_control_expr = true
 fn_args_paren_newline = false
-max_width = 140
+max_width = 160
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,9 @@
+fn_args_layout = "Block"
+array_layout = "Block"
+control_style = "Rfc"
+where_style = "Rfc"
+generics_indent = "Block"
+fn_call_style = "Block"
+combine_control_expr = true
+fn_args_paren_newline = false
+max_width = 140
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ install:
   - yarn
   - yarn run bower install
   - cargo install diesel_cli --debug --no-default-features --features postgres && export PATH=$HOME/.cargo/bin:$PATH
-  - which rustfmt || cargo install rustfmt
+  - cargo install rustfmt
+  - rustfmt --version
 
 before_script:
   - diesel database setup
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,8 @@ matrix:
   - rust: stable
     script:
     - cargo fmt -- --write-mode=diff
+  - rust: stable
+    script:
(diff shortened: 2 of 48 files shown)


elliotekj:
Hey @vignesh-sankaran :)

I have a few suggestions for this:

1. A rustfmt config file should probably be added to the codebase as part of this PR for consistency going forward. The RFC rustfmt config [was suggested](https://github.com/rust-lang/crates.io/pull/581#issuecomment-306519550) by @steveklabnik — was that the one you used?

2. Rustfmt has [a skip attribute](https://github.com/rust-lang-nursery/rustfmt/blob/master/tests/target/skip.rs) which should let us work around the long function name from the OpenSSL crate.

3. Again for consistency going forward, [the Travis config](https://github.com/rust-lang/crates.io/blob/master/.travis.yml) should probably be updated as part of this PR to fail if code isn't formatted correctly: https://github.com/rust-lang-nursery/rustfmt#checking-style-on-a-ci-server

Just my suggestions — best to wait and see what @carols10cents has to say before acting on anything.

Cheers!

carols10cents:
Yup, I agree with @elliotekj's suggestions-- do you mind making those changes @vignesh-sankaran? Please let me know if you have any questions!

vignesh-sankaran:
In response @elliotekj, I just used rustfmt with overwrite, I'll redo the rustfmt pass with the config file. Thanks for the tip regarding the skip attribute, I'll add that in too. 

No problems @carols10cents, looks like I'll fix up #575 in the process as well.

carols10cents:
Thank you so much!! I've merged this in, and I decided to move the documentation to CONTRIBUTING.md and I added some notes about jslint and clippy too.

I also was able to get the max_width down to 118, but then I discovered the [error_on_line_overflow](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#error_on_line_overflow) setting, which I think does about what we want here, since the line with the super long SSL function can't be fixed [or ignored](https://github.com/rust-lang-nursery/rustfmt/issues/1298).

I also changed the default write mode to Overwrite, since that's what I usually want when I run locally.

Thank you!! I'm so excited to be able to stop agonizing over the formatting!!!

vignesh-sankaran:
No worries @carols10cents, good to hear you'll be stressing less over code formatting :)

Issue #776, opened 2017-06-14
Travis is failing during setup


elliotekj:
Travis seems to failing whilst setting up diesel_cli:

![screen shot 2017-06-14 at 12 53 30](https://user-images.githubusercontent.com/1547124/27128864-89f7a684-5100-11e7-8d2c-248544859e48.png)

I'm seeing this in https://github.com/rust-lang/crates.io/pull/775 but @Turbo87 is seeing it too in https://github.com/rust-lang/crates.io/pull/772.

At the time of writing this, https://github.com/rust-lang/crates.io/pull/772 is 13 hours old. The last passing test looks to be [this commit](https://github.com/rust-lang/crates.io/pull/771/commits/91898fa138f3348813449dede519cf4d7b800b24) from https://github.com/rust-lang/crates.io/pull/771 20 hours ago.

cc'ing @sgrif because Diesel :)

carols10cents:
Ah, we [added cache: cargo: true](https://github.com/rust-lang/crates.io/commit/209dab0f40bbcaa7826188dd3d3ff85b72005ecd#diff-354f30a63fb0907d4ad57269548329e3). We should probably add `--force` to always get the latest version of diesel, as the error message suggests. Can you send me a PR? Otherwise I can get to this in a few hours :)

elliotekj:
Good call @carols10cents — adding `--force` fixed the diesel_cli issue. Unfortunately Travis failed with a similar error when installing rustfmt.

![screen shot 2017-06-14 at 16 09 43](https://user-images.githubusercontent.com/1547124/27136821-83720112-511c-11e7-8356-182e9cc84b51.png)

If `--force` is going to be needed on both of the `cargo install`s then reverting https://github.com/rust-lang/crates.io/commit/209dab0f40bbcaa7826188dd3d3ff85b72005ecd#diff-354f30a63fb0907d4ad57269548329e3 seems like the best course of action because we won't be getting any benefit from it anyway (unless I'm missing something?).

carols10cents:
I think it'll cache all of our other dependencies, which should help overall time of the travis builds. I think adding `--force` to rustfmt too is fine!

elliotekj:
Right'o, I'll make the change :)

PR #777, opened 2017-06-14
Force Travis to get the latest versions of diesel_cli and rustfmt

elliotekj:
This should hopefully fix https://github.com/rust-lang/crates.io/issues/776.

As noted by @carols10cents, `--force` is needed to get the latest version of diesel_cli because [we added cargo caching](https://github.com/rust-lang/crates.io/commit/209dab0f40bbcaa7826188dd3d3ff85b72005ecd#diff-354f30a63fb0907d4ad57269548329e3).


elliotekj:
@carols10cents Looks like we're back in action 👍

carols10cents:
Thank you!!!

PR #778, opened 2017-06-15
Update to Babel 6

Turbo87:
/cc @carols10cents @locks


carols10cents:
Looks like it works to me! I updated the targets to match rust-lang/rfcs#1985, deploying shortly!! Thank you!! ❤️

PR #780, opened 2017-06-16
Make analytics send later

pornel:
I've noticed "waiting for google-analytics" in my status bar while the page was still loading.

Since the site doesn't have to support old IE, it can use `<script async>`. I've left creation of the `ga` function where it was in case other scripts use it.


locks:
I'm very confused by what happened here. Did you merge into your own master? I don't see it in this repository.

pornel:
It looks like the analytics code has been refactored in master, so this change no longer makes sense.

PR #781, opened 2017-06-16
Replace normalize-css bower component with npm one

pornel:
Bower is dead, so I thought I'd start replacing bower components bit by bit.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -20,7 +20,7 @@ module.exports = function(defaults) {
   // please specify an object with the list of modules as keys
   // along with the exports of each module as its value.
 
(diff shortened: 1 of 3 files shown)


ember-cli-build.js
```diff
@@ -20,7 +20,7 @@ module.exports = function(defaults) {
   // please specify an object with the list of modules as keys
   // along with the exports of each module as its value.
 
-  app.import('bower_components/normalize-css/normalize.css');
+  app.import('node_modules/normalize.css/normalize.css');
```
Turbo87: this doesn't work yet. `this.import()` for files in `node_modules` will only work starting with Ember CLI v2.15

Issue #736, opened 2017-05-22
Show licenses of crates on a per version basis


est31:
Follow up of #227.

Currently, crates.io maintains license information on a per crate basis. I'd propose to base it on the (crate, version) tuple instead.

The issue is that it might mislead people to believe that the license of a crate is something different from what it actually is.

Like for example, take someone creating a crate with portions of code from someone else, licensed under the MPL. Now assume this was the reason to license the remainder of the crate under the MPL as well, and version 0.1 is published as MPL. Now the author finds out that the functionality of that external code is already implemented by someone else on crates.io and removes it, adding a dependency instead. This allows the author to relicense under the more custom Apache2/MIT duo, which what the author then does, releasing version 0.2 of the crate under Apache2/MIT.

Now, take some user who for some reason thinks that version 0.1 is better than version 0.2 and wants to use it instead. The user looks at the crates.io page of the version of the crate, and sees its licensed under Apache2/MIT which is okay, while the MPL would be not okay for them. They then mistakenly use the 0.1 version without knowing its in fact MPL licensed!

Or take another case (inspired by the winehq story): Some person happily releases version after version of their crate under the Apache2/MIT license, until they discover that someone else took their crate, improved it and now sells the library in exchange for money. The author doesn't like that and decides to relicense the crate under the LGPL license, in order to be able to compete with the proprietary library (so that not every improvement can automatically be imported to the proprietary library). Now the future versions of the crate are LGPL licensed. Does that change the license of the old versions though? No it doesn't! Once you license some code under an open source license, it can be continued to be used under that license, even if you reconsider. Only changes you make to that code can be put under a new license. So crates.io saying that the older versions are LGPL as well might scare away users who would use that crate if it were Apache2/MIT licensed and can live with older versions.

---

## Tasks

Ok so I (@carols10cents) think what needs to be done here is:

- [x] Make a diesel migration to add a `license` column to the `versions` table
- [x] Write a script in `src/bin` that, for all current `crates`. will copy the value of `crate.license` to each of that crate's `version.license` values to backfill that column for existing crates
[...]

PR #787, opened 2017-06-19
Started working on having the ability to have a different license for each version of a crate.

baileyn:
This is some work towards fixing #736

### DIFF
--- a/src/bin/update-licenses.rs
+++ b/src/bin/update-licenses.rs
@@ -21,8 +21,7 @@ fn main() {
 }
 
 fn transfer(tx: &postgres::transaction::Transaction) {
-    let stmt = tx.prepare("SELECT id, name, license FROM crates")
-        .unwrap();
+    let stmt = tx.prepare("SELECT id, name, license FROM crates").unwrap();
     let rows = stmt.query(&[]).unwrap();
 
     for row in rows.iter() {
@@ -37,10 +36,9 @@ fn transfer(tx: &postgres::transaction::Transaction) {
         );
 
         let num_updated = tx.execute(
-                "UPDATE versions SET license = $1 WHERE crate_id = $2",
-                &[&license, &id],
-            )
-            .unwrap();
+            "UPDATE versions SET license = $1 WHERE crate_id = $2",
+            &[&license, &id],
+        ).unwrap();
         assert!(num_updated > 0);
     }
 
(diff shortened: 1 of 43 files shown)


carols10cents:
Looks like rustfmt is still sad :( I'm not sure why it's different... I'm going to try running it on here.

carols10cents:
This is going well!! I love the `VersionBuilder` in the tests ❤️ 

I tried out the code locally, and publishing versions locally was doing what I would expect as far as continuing to update the crate license, continuing to reject invalid licenses, and starting to keep track of the licenses in the versions table 💯 

One small issue I found when running `update-licenses` however: the license column on the crate table can be null (that is, we haven't put a `NOT NULL` restriction on it in the database), so of course that means there are some crates with a null license in the production database (36 of them. and somehow one in my local database??) 😱😱😱

I'm going to file a separate issue for that, because I think we should be able to make it not null, but I'm not sure how it's happening, if it's something that's been fixed or not, so I think in this PR we should support a null license.

Currently, though, when `update-licenses` sees a null value, I get this error:

```
thread 'main' panicked at 'error retrieving column "license": Conversion(WasNull)', /Users/carolnichols/.cargo/registry/src/github.com-1ecc6299db9ec823/postgres-0.14.0/src/rows.rs:206
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: postgres::rows::Row::get
   7: update_licenses::transfer
   8: update_licenses::main
   9: __rust_maybe_catch_panic
  10: std::rt::lang_start
  11: main
```

I *think* on line 30 of update-licenses.rs we can just change the type annotation from `String` to `Option<String>` and it should work? Do you mind trying that out and letting me know?

Since I'm not sure how crates.io is creating crates with null licenses, you can change a crate in your database to have a null license in `psql` by running:

```
UPDATE crates SET license = NULL where crates.name='whatever';
```

Thank you!!! Let me know if you have questions!!!

src/bin/update-licenses.rs
```diff
@@ -27,19 +27,26 @@ fn transfer(tx: &postgres::transaction::Transaction) {
...
-        );
-
-        let num_updated = tx.execute(
-            "UPDATE versions SET license = $1 WHERE crate_id = $2",
-            &[&license, &id],
-        ).unwrap();
-        assert!(num_updated > 0);
+        let license: Option<String> = row.get("license");
+
+        if let Some(license) = license {
```
carols10cents: nice use of `if let`!

Issue #494, opened 2016-12-18
Have "favorite authors" and highlight crates by your favorite authors in crate lists


carols10cents:
In the user research study we did, many people mentioned the reputation of an author as a factor they considered when deciding which crates to try. 

We could do something like show how many other crates an author has, with the assumption that more crates = more experience = better quality, but that might not be true, and publishing lots of empty crates could gain you more reputation defined in this way.

We could distinguish members of any of the Rust teams, but there are lots of experienced people providing high quality crates that would not get this distinction.

Instead, I think it would be useful if each person who uses crates.io could add a crate author to their "favorite" authors, and any search that user does would show an indicator in the list of crates if the crate was written by a favorite author. The list of someone's favorite authors would be private so that each person could decide who they think is experienced, and this feature wouldn't get gamed or become a popularity contest.

PritiKumr:
@carols10cents This is a very cool feature and I'd love to work on this.

carols10cents:
Awesome! Do you have any questions about getting started?

PritiKumr:
Hey @carols10cents 

> I think this should be on crate owners, that is, the users or groups that are in crates.io's database and associated with a crate, not crate author information, that can be anything, read from Cargo.toml

So in terms of the tables and associations, shouldn't we be creating a table, probably called `favorite_users` which will have references to the favorited user and the currentUser?

The other alternative is to associate the user with `crate_owners` records, which is what I was initially thinking. But then a crate owner can be removed from a crate. So this would be not be a good idea IMO.

What do you think?.

carols10cents:
Creating a favorite_users join table sounds right to me!

PritiKumr:
> potentially with what they've been up to lately on crates.io

@carols10cents could you elaborate more on what sort of things would be useful to show here? I was thinking we can show things like new crates, version updates, users / crates they follow. But then this would probably require us to have an activity feed system to implement this effectively, which could be a separate task on its own.

Wondering if we can just list the favorite users in the dashboard for now...

carols10cents:
Yeah, I think just having a list of favorite users on the dashboard (and the ability to unfavorite them there) is a great start!

steverob:
@carols10cents  here are two options regarding the UI indication part of this issue - 

**Highlighted background: **
![2017-07-22_14-39-56_scrot](https://user-images.githubusercontent.com/1220480/28489951-a53a0982-6eec-11e7-96ad-7e3a82f65df4.png)

[...]

PR #789, opened 2017-06-20
Favorite crate owner #494

PritiKumr:
This is a WIP. Here are some things that are still pending - 

1. Mark crates by the user's favorite owners with a special highlight in various crate listings
2. Tests

This is where one can Favorite / Unfavorite a user - 
favorite-button

This is where favorite owners are listed in the dashboard (on hover, we show a `x` icon which unfavorite the owner).
favorites


PritiKumr:
All, need some thoughts on which sort of crate listings needs the favorite owner highlight.

Can we only do this for the `crate-row` component? That's used in places like - https://crates.io/categories/authentication and https://crates.io/crates

We do have the mini `crates-list` listing used here ->
screenshot from 2017-06-20 19-03-12

Thoughts?

src/user/mod.rs
```diff
@@ -336,6 +339,86 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
...
+#[table_name="favorite_users"]
+pub struct FavoriteUser {
+    user_id: i32,
+    target_id: i32,
+}
+
+
+fn favorite_target(req: &mut Request) -> CargoResult<FavoriteUser> {
+    let user = req.user()?;
+    let target_user_id: i32 = req.params()["user_id"].parse().expect("User ID not found");
```
sgrif: Do you think this should return an error rather than using `expect`?

src/user/mod.rs
```diff
@@ -336,6 +339,86 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
...
+}
+
+
+
+
+
+/// Handles the `GET /users/:user_id/favorite_users` route.
+pub fn favorite_users(req: &mut Request) -> CargoResult<Response> {
+    let user_id: i32 = req.params()["user_id"].parse()
+        .expect("User ID not found");
```
sgrif: Do you think this should return an error rather than using `expect`?

src/user/mod.rs
```diff
@@ -336,6 +339,86 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
...
+    let conn = req.db_conn()?;
+
+    let users = users::table.filter(users::id.eq_any(
```
sgrif: Can this  be `users.inner_join(favorite_users).filter(favorite_users::user_id.eq(user_id))``?

app/controllers/dashboard.js
```diff
@@ -32,8 +37,12 @@ export default Ember.Controller.extend({
         return this.get('myFollowing.length') > TO_SHOW;
     }),
 
+    hasMoreFavorites: computed('favoriteUsers', function() {
+        return this.get('favoriteUsers.length') > TO_SHOW;
+    }),
+
     actions: {
-        loadMore() {
+        loadMore: function() {
```
acorncom: Generally the ES6 style of method would be preferred here

app/routes/user.js
```diff
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import ajax from 'ic-ajax';
```
acorncom: In other cleanup PRs that have been going on, ic-ajax use has been removed in favor of ember-ajax

app/routes/user.js
```diff
@@ -12,6 +13,17 @@ export default Ember.Route.extend({
 
         controller.set('fetchingFeed', true);
         controller.set('crates', this.get('data.crates'));
+        controller.set('user', model.user);
+        controller.set(
+            'allowFavorting',
```
acorncom: `allowFavoriting`?

carols10cents:
> All, need some thoughts on which sort of crate listings needs the favorite owner highlight.
> 
> Can we only do this for the crate-row component? That's used in places like - https://crates.io/categories/authentication and https://crates.io/crates
> 
> We do have the mini crates-list listing used here ->

I think we should highlight crates by favorite owners wherever crates are listed, so in both crate-row and crates-list.

vignesh-sankaran:
Just wanted to follow up on this @PritiKumr, was there anything you needed help with?

src/user/mod.rs
```diff
@@ -336,6 +339,86 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
     Ok(req.json(&R{ user: user.encodable() }))
 }
 
+#[derive(Insertable, Queryable, Identifiable, Associations)]
+#[belongs_to(User)]
+#[primary_key(user_id, target_id)]
+#[table_name="favorite_users"]
+pub struct FavoriteUser {
```
vignesh-sankaran: Could you add a documentation comment here? :)

src/user/mod.rs
```diff
@@ -336,6 +339,86 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
...
+    user_id: i32,
+    target_id: i32,
+}
+
+
+fn favorite_target(req: &mut Request) -> CargoResult<FavoriteUser> {
```
vignesh-sankaran: Doc comment here too please :)

migrations/20170614104258_create_index_on_favorite_users/up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX index_favorites_user_id ON favorite_users (user_id);
```
sgrif: This index is redundant with the primary key.

src/lib.rs
```diff
@@ -153,6 +153,10 @@ pub fn middleware(app: Arc<App>) -> MiddlewareBuilder {
     api_router.get("/users/:user_id", C(user::show));
     api_router.get("/users/:user_id/stats", C(user::stats));
     api_router.get("/teams/:team_id", C(user::show_team));
+    api_router.get("/users/:user_id/favorited", C(user::favorited));
```
sgrif: Is this route redundant with `favorite_users`?

src/lib.rs
```diff
@@ -153,6 +153,10 @@ pub fn middleware(app: Arc<App>) -> MiddlewareBuilder {
...
+    api_router.put("/users/:user_id/favorite", C(user::favorite));
+    api_router.delete("/users/:user_id/favorite", C(user::unfavorite));
+    api_router.get("/users/:user_id/favorite_users", C(user::favorite_users));
```
sgrif: Are we intending to show the users that another user has favorited? Do these routes actually need the `user_id` in the URL?

app/routes/user.js
PritiKumr: Thanks!

vignesh-sankaran:
Could you also run a `cargo fmt` and re upload the code?

vignesh-sankaran:
@PritiKumr Were you still working on this? Perhaps we could close the PR if there's no work being done on this.

PR #790, opened 2017-06-20
Team page

### LINKED ISSUE #409: Show all crates owned by a user or group
Much like I have a page of "my crates", I would like to be able to see a page of all crates that, say, the pistondevelopers github organization is listed as an owner on.

This would let me discover crates that are officially supported to work well together, as well as find other crates by authors who I trust.

natboehm:
Creates a team page listing all crates associated with given team. 

screen shot 2017-06-20 at 2 28 58 pm

Fixes #409

### DIFF
--- a/app/routes/team.js
+++ b/app/routes/team.js
@@ -3,7 +3,7 @@ import Ember from 'ember';
 export default Ember.Route.extend({
     queryParams: {
         page: { refreshedModel: true },
-        sort: { refreshedModel: true }, 
+        sort: { refreshedModel: true },
     },
     data: {},
 
--- a/app/routes/team.js
+++ b/app/routes/team.js
@@ -1,0 +1,38 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+    queryParams: {
+        page: { refreshedModel: true },
+        sort: { refreshedModel: true }, 
+    },
+    data: {},
+
+    setupController(controller, model) {
+        this._super(controller, model);
+
+        controller.set('fetchingFeed', true);
+        controller.set('crates', this.get('data.crates'));
+    },
+
+    model(params) {
+        const { team_id } = params;
+        return this.store.find('team', team_id).then(
+            (team) => {
+                params.team_id = team.get('id');
+                return Ember.RSVP.hash({
+                    crates: this.store.query('crate', params),
+                    team
+                });
+            },
+            (e) => {
+                if (e.errors.any(e => e.detail === 'Not Found')) {
+                    console.log(params);
+                    this
+                        .controllerFor('application')
+                        .set('nextFlashError', `User '${params.team_id}' does not exist`);
+                    return this.replaceWith('index');
+                }
+            }
+        );
+    },
+});
--- a/app/routes/team.js
+++ b/app/routes/team.js
@@ -26,7 +26,6 @@ export default Ember.Route.extend({
             },
             (e) => {
                 if (e.errors.any(e => e.detail === 'Not Found')) {
-                    console.log(params);
                     this
                         .controllerFor('application')
                         .set('nextFlashError', `User '${params.team_id}' does not exist`);
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -603,9 +603,9 @@ impl Crate {
     }
 
     pub fn owner_team(&self, conn: &PgConnection) -> CargoResult<Vec<Owner>> {
-        let base_query = CrateOwner::belonging_to(self)
-            .filter(crate_owners::deleted.eq(false));
-        let teams = base_query.inner_join(teams::table)
+        let base_query = CrateOwner::belonging_to(self).filter(crate_owners::deleted.eq(false));
+        let teams = base_query
+            .inner_join(teams::table)
             .select(teams::all_columns)
             .filter(crate_owners::owner_kind.eq(OwnerKind::Team as i32))
             .load(conn)?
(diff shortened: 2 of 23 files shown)


src/krate.rs
```diff
@@ -598,6 +602,32 @@ impl Crate {
         Ok(users.chain(teams).collect())
     }
 
+    pub fn owner_team(&self, conn: &PgConnection) -> CargoResult<Vec<Owner>> {
```
sgrif: `Crate` has a lot of behavior on it as is. What do you think about putting this on `Team` instead and calling it `Team::owning(crate: &Crate, conn: &PgConnection)`

src/krate.rs
```diff
@@ -598,6 +602,32 @@ impl Crate {
...
+            .select(teams::all_columns)
+            .filter(crate_owners::owner_kind.eq(OwnerKind::Team as i32))
+            .load(conn)?
+            .into_iter()
+            .map(Owner::Team);
+
+        Ok(teams.collect())
+    }
+
+    pub fn owner_user(&self, conn: &PgConnection) -> CargoResult<Vec<Owner>> {
```
sgrif: `Crate` has a lot of behavior on it as is. What do you think about putting this on `User` instead and calling it `User::owning(crate: &Crate, conn: &PgConnection)`

src/krate.rs
```diff
@@ -864,6 +894,15 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
                     .filter(crate_owners::owner_kind.eq(OwnerKind::User as i32)),
             ),
         );
+    } else if let Some(team_id) = params.get("team_id").and_then(|s| s.parse::<i32>().ok()) {
```
sgrif: Is there a reason to put this on `index` instead of adding a new route?

src/krate.rs
```diff
@@ -1423,6 +1462,42 @@ pub fn owners(req: &mut Request) -> CargoResult<Response> {
     Ok(req.json(&R { users: owners }))
 }
 
+/// Handles the `GET /crates/:crate_id/owner_team` route.
+pub fn owner_team(req: &mut Request) -> CargoResult<Response> {
```
sgrif: What do you think about making this a single route and adding a `?kind` parameter if needed?

src/tests/all.rs
```diff
@@ -212,6 +215,33 @@ fn user(login: &str) -> User {
...
+    NewTeam {
+        github_id: NEXT_ID.fetch_add(1, Ordering::SeqCst) as i32,
+        login: login,
+        name: None,
+        avatar: None,
+    }
+}
+
+fn add_team_to_crate(t: &Team, krate: &Crate, u: &User, conn: &PgConnection) -> CargoResult<()> {
+
```
sgrif: ✂️

src/tests/all.rs
```diff
@@ -212,6 +215,33 @@ fn user(login: &str) -> User {
...
+        crate_id: krate.id,
+        owner_id: t.id,
+        created_by: u.id,
+        owner_kind: 1, // Team owner kind is 1 according to owner.rs
+    };
+
+    diesel::insert(&crate_owner.on_conflict(
+        crate_owners::table.primary_key(),
+        do_update().set(crate_owners::deleted.eq(false)),
+    )).into(crate_owners::table)
```
sgrif: Is this the indentation `rustfmt` came up with?

app/routes/team.js
```diff
@@ -0,0 +1,38 @@
...
+            (team) => {
+                params.team_id = team.get('id');
+                return Ember.RSVP.hash({
+                    crates: this.store.query('crate', params),
+                    team
+                });
+            },
+            (e) => {
+                if (e.errors.any(e => e.detail === 'Not Found')) {
+                    console.log(params);
```
carols10cents: Looks like a debugging `console.log` left in here :)

src/owner.rs
```diff
@@ -172,47 +215,48 @@ impl Team {
...
+        }
+    }
 
-        diesel::insert(&new_team.on_conflict(
-            teams::github_id,
-            do_update().set(&new_team),
-        )).into(teams::table)
-            .get_result(conn)
-            .map_err(Into::into)
+    fn get_org(login: &str) -> String {
```
carols10cents: I think we can refactor this a bit into one function that returns two string slices in a tuple-- I'll explain what I mean in person :)

src/tests/krate.rs
```diff
@@ -1740,3 +1745,95 @@ fn author_license_and_description_required() {
         json.errors
     );
 }
+
+/*  Testing the crate ownership between two crates and one team.
+    Given two crates, one crate owned by both a team and a user,
+    one only owned by a user, check that the CrateList returned
+    for the user_id contains only the crates owned by that user,
+    and that the CrateList returned for the team_id contains
+    only crates owned by that team.
```
carols10cents: 😻 love these comments!!

src/krate.rs
carols10cents: For consistency with how getting a user's crates works, mostly. If we're going to pull this out into its own route, then I think users should be pulled out too, which imo should be a separate PR. WDYT?

src/tests/all.rs
carols10cents: You're just talking about cutting the newline here, right? #ambiguousemoji

src/tests/all.rs
carols10cents: yeppers

src/krate.rs
carols10cents: We tried, but the structure of the response has `user:` vs `team:` change that ember cares about, and I suck at ember and I'm not sure why it cares about that. If we duplicate the response structure too, then it's most of the function that's still duplicated :-/

I wouldn't be opposed to refactoring this in the future, it would just take me more time in ember than we'd like to spend right now :P

src/krate.rs
sgrif: I think that users should be pulled out as well, but I don't think that should block this PR being a separate route, if it's not significantly more work for this PR.

PR #793, opened 2017-06-21
Use async/await for tests

Turbo87:
as suggested by e.g. https://emberway.io/maintainable-user-interface-testing-with-ember-40f1b58040b6


Turbo87:
this is blocked on https://github.com/rust-lang/crates.io/pull/794

PR #797, opened 2017-06-22
Cleanup

locks:
I'll elaborate as I go.


app/components/google-jsapi.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+    didInsertElement() {
```
Turbo87: `tagName: ''`?

locks:
@Turbo87 please re-review.

Issue #645, opened 2017-03-20
The transfer_crates script should use gh_id instead of gh_avatar


carols10cents:
This script was made before we had gh_id on all users, which should be more reliable than avatar URL.

vignesh-sankaran:
I'd be happy to take this up @carols10cents, I'm not sure how to test the script though once the changes are made.

carols10cents:
@vignesh-sankaran yeah, this script is only run occasionally, so I think manual testing is fine. You'll need to create two users, publish a crate with one user, then run this script to transfer the crate to the other user. Then check that the UI shows the crate as being owned by the second user, and that the second user can publish new versions of that crate and the first user cannot.

Publishing locally can be tricky, I think I've documented the setup in docs/CONTRIBUTING.md but please let me know if you have any trouble!

PR #798, opened 2017-06-23
Switch to GitHub IDs #645

vignesh-sankaran:
Testing procedure I used was the following:

- Log in as a user
- Upload a crate under that user
- Log out, and log in as another user
- Attempt publishing the crate as the prior user
- Log in as the other user, modify the crate and publish the crate

@carols10cents A couple of things I noticed were that the script exited if an uppercase "Y" was entered, and that there is an assertion that fails if more than 1 crate is being transferred. 

As a matter of interest, when is this script utilised?

### DIFF
--- a/src/category.rs
+++ b/src/category.rs
@@ -1,4 +1,3 @@
-use std::collections::HashSet;
 use time::Timespec;
 
 use conduit::{Request, Response};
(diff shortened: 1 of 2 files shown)


vignesh-sankaran:
Argh I should learn not to use git push --force, lost my changes

vignesh-sankaran:
Argh sorry for the mess, forgot to branch things out :(

Turbo87:
@vignesh-sankaran FWIW you should probably not use your `master` branch for PRs 😉

vignesh-sankaran:
@Turbo87 I've learned that the hard way now :)

PR #803, opened 2017-06-25
Licenses

baileyn:
Fixed the front-end to display licenses based on versions rather than crates. I also added a test to make sure the licenses were displayed correct and changed based on version. This should finish up #736. The only thing that's left to do now is remove the old code there for crate-based versions.


carols10cents:
Hi @baileyn sorry I've taken so long to review this again!! Taking a look now :) ❤️

carols10cents:
This code is looking great!! I just have a few things:

- I rebased this branch on master and force pushed, so please fetch and rebase to get my changes-- there are some folks helping out making the ember code and tests better, and they're starting to move the mirage fixtures to mirage factories, so since we changed that out from under you, I went ahead and updated your code here :) (plus this was an opportunity to get to play with the fixtures :)) Please let me know if you have any questions about factories vs. fixtures!

- Since we're removing the license key from crate JSON in the fixtures/factories, and since the fixtures/factories represent the JSON that the backend returns, I think this PR would be a great time to remove the license key from [`EncodableCrate`](https://github.com/rust-lang/crates.io/blob/65780b2/src/krate.rs#L107). This will change everywhere we create JSON for crates, but the license will still be in the database (since that's controlled by the `Crate` struct), so if we need to roll this back for some reason, we won't have lost any data. You might get compiler errors or Rust test errors when removing the license field from `EncodableCrate`-- do you mind removing the field and fixing whatever errors come up? Does this make sense?

- Since you've done a few PRs now, I think you're ready for one of my very favorite git commands that I use all the time-- it's called "interactive rebase" and it lets you change your commits around after you've made them :) It looks like your commit message for your first commit, "Maybe now I fixed the formatting." was kind of a development note, which is totally fine and I do that all the time too (you should see how many commits I have locally that just say "poop" haha). However, if someone else was looking back on the history of commits and wanted to know why the changes in the commit were being made, the message here doesn't really describe the change-- that the license displayed on crate and version pages is now coming from the version JSON rather than the crate JSON, and that this is part of fixing #736. One of the things interactive rebasing does is lets you change your commit message easily! 

The way I usually do interactive rebasing is `git rebase -i upstream/master` (where `upstream` is your remote that points to github.com/rust-lang/crates.io), which will rebase your branch on the latest master and will put you in vim with all of your commits that are not yet on master listed at the top. There are instructions in comments at the bottom of what you can do here-- the one you'll want here is `r` for `reword`, so change the `pick` next to your first commit to say `r` then `:wq` out of vim. Next you'll be put into vim editing that commit's message! Make it a bit more descriptive, then `:wq`, then the rebase will finish. You'll have to force push to github again, and then you should be all set.

There's some more documentation about interactive rebasing here: https://help.github.com/articles/about-git-rebase/ Please let me know either here or on IRC if you have any questions!! ❤️

baileyn:
Sorry it's taken me so long to reply, @carols10cents , I've been helping my parents move over the past couple of weeks so I haven't been near my development box. I'm heading back home today so I'll be sure to take a look at resolving all of the issues you've mentioned.

carols10cents:
No apologies necessary :) ❤️

carols10cents:
And this is deployed! Check it out-- I've got a crate that has v0.1.2 licensed MIT https://crates.io/crates/carol-test/0.1.2 and v0.1.3 licensed MIT/Apache-2.0 https://crates.io/crates/carol-test !! Wooooo!!!

baileyn:
Wooooo! It's so exciting to see something I made actually work on a production server 😁

PR #806, opened 2017-06-25
Cleanup before we can make crates.io run on FastBoot 

sivakumar-kailasam:
- Replace bower dependencies with ember addon equivalents & remove bower dependency.
  - [x] clipboard-copy
  - [x] normalize-css
  - [x] mirage
- [x] Replace ic-ajax & other ajax code with ember-fetch.


app/controllers/crate/version.js
```diff
@@ -30,12 +30,14 @@ export default Ember.Controller.extend({
 
     hasMoreVersions: computed.gt('sortedVersions.length', NUM_VERSIONS),
 
-    anyLinks: computed.or('crate.homepage',
-                          'crate.wiki',
-                          'crate.mailing_list',
-                          'crate.documentation',
-                          'crate.repository',
-                          'crate.reverse_dependencies'),
+    anyLinks: computed.or(
```
locks: should we use brace expansion here?

sivakumar-kailasam: Done!

PR #807, opened 2017-06-25
Support CircleCI badges

### LINKED ISSUE #800: Support badges from CircleCI
https://circleci.com

sfackler:
Closes #800 

ember-cli seems unhappy with Node 8 that Arch Linux has, so I haven't run frontend tests unfortunately :(

Badge docs: https://circleci.com/docs/1.0/status-badges/

### DIFF
--- a/app/components/badge-circle-ci.js
+++ b/app/components/badge-circle-ci.js
@@ -5,7 +5,7 @@ export default Ember.Component.extend({
     classNames: ['badge'],
     repository: Ember.computed.alias('badge.attributes.repository'),
     branch: Ember.computed('badge.attributes.branch', function() {
-        return this.get('badge.attributes.branch') || 'master';
+        return encodeURIComponent(this.get('badge.attributes.branch') || 'master');
     }),
     text: Ember.computed('branch', function() {
         return `Circle CI build status for the ${ this.get('branch') } branch`;
--- a/app/components/badge-circle-ci.js
+++ b/app/components/badge-circle-ci.js
@@ -1,0 +1,13 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+    tagName: 'span',
+    classNames: ['badge'],
+    repository: Ember.computed.alias('badge.attributes.repository'),
+    branch: Ember.computed('badge.attributes.branch', function() {
+        return this.get('badge.attributes.branch') || 'master';
+    }),
+    text: Ember.computed('branch', function() {
+        return `Circle CI build status for the ${ this.get('branch') } branch`;
+    })
+});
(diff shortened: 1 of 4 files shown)


sfackler:
Ping @carols10cents

locks:
cc @Turbo87 on node 8 + CLI

Turbo87:
https://github.com/creationix/nvm

sfackler:
Rebased

sfackler:
The test failure looks unrelated to this PR.

carols10cents:
Thank you for your patience on this! Rebased and force pushed to resolve a merge conflict and get tests to run again.

The code looks good, but I noticed that the CircleCi docs say this:

> (You can also link to the image without providing a circle-token, in which case the image will only appear for people who are currently logged-in to circleci.com!)

That seems unfortunate. Do you think we should optionally accept a circle-token, so at least crate authors have the ability to make their image show up for everyone?

carols10cents:
Oooooh, one other thing:

> One small wrinkle: if your branch name includes characters that are “special” in URLs (most commonly ‘/’) they’ll need to be URL encoded when constructing status badge URLs. For example, the badge for a branch called ‘feature/nonexistent’ would be at ‘…/tree/feature%2Fnonexistent.png’.

I just tested this out by making a branch named "cratesio/test" and specifying that:

```
[badges]
circle-ci = { repository = "carols10cents/sassers", branch = "cratesio/test" }
```

and indeed, we're not URL encoding the slash; the HTML getting generated for the image is:

```
<img src="https://circleci.com/gh/carols10cents/sassers/tree/cratesio/test.svg?style=shield" alt="Circle CI build status for the cratesio/test branch" title="Circle CI build status for the cratesio/test branch">
```

and indeed the image doesn't load :-/

sfackler:
I think the circle token comment only applies to private repos - I was worried about this as well initially. For example, [this badge](https://circleci.com/gh/sfackler/cargo-sls-distribution.svg?style=shield) page is visible to me when logged out from circle - is it visible for you as well?

Fixed up the URL encoding!

Turbo87:
@carols10cents I'm wondering why we don't just use http://shields.io/ instead for all the badges?

carols10cents:
> @carols10cents I'm wondering why we don't just use http://shields.io instead for all the badges?

@Turbo87 does it support everything we want? i don't see gitlab CI on there, for example

Turbo87:
yeah, seems like GitLab is missing 😞

sfackler:
I've also noticed shields's Cargo version badge being really slow to load.

carols10cents:
Thanks! Sorry for taking so long on this, but I'll be deploying it once travis passes! 🍏

sfackler:
Sweet, thanks! There's no list of badges in Cargo I need to update as well, is there?

carols10cents:
Yeah there is: https://github.com/rust-lang/cargo/blame/1ee4edba1fa1a1d323be97bfec280764407d89e5/src/doc/manifest.md#L157

I got distracted yesterday and didn't finish deploying, but the deploy is going now!

Issue #306, opened 2016-04-08
Going to dependent crates then back makes the crate look like it depends on itself


troplin:
1. Go to the page of a crate that has some dependencies.
2. Click on the link of one of the dependencies
3. On the page of the dependecy, click on "dependent crates".
4. Click on the link to the original crate again
5. Look at the dependency links:
   - The dependency that we were following is now gone
   - The crate now seems to depend on itself

tbu-:
This also happens with the back button:

1) Go to a crate (e.g. [itertools](https://crates.io/crates/itertools)).
2) Click a dependency (e.g. permutohedron).
3) Click the back button in the browser.

Expected result: Crate page for itertools, listing permutohedron 0.1 as dependency.
Actual result: Crate page for itertools, listing itertools 0.1 as the only dependency.

Tested in Firefox 45.0.2 and Chromium 50.0.2661.75.

carols10cents:
Just tried this and this still happens. Not sure how to fix it though.

locks:
Back button works properly for me, but I still see the problem with OP's reproduction.

Turbo87:
the issue here is as follows:

when we visit `/crates/itertools` it will download the dependencies from `/api/v1/crates/itertools/0.6.0/dependencies` which includes:

```json
{
  "crate_id":"either",
  "id":186109,
  "kind":"normal",
  "req":"^1.0",
  "target":null,
  "version_id":49566
}
```

when we click on the `either` dependency on the page it will request data from `/api/v1/crates/either/reverse_dependencies` which includes:

```json
{
  "crate_id":"itertools",
  "id":186109,
  "kind":"normal",
  "optional":false,
  "req":"^1.0",
  "target":null,
  "version_id":49566
}
```

when we click the back button Ember Data will reuse the dependencies that it already downloaded for the `itertools` crate, and since the `reverse_dependencies` call overwrote the `dependency:186109` record it will show `itertools` instead.

tl;dr it seems wrong that the API is returning a different `crate_id` content for the record depending on how it was queried (deps vs. reverse deps).

PR #810, opened 2017-06-26
WIP: Fix reverse dependencies query

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/306

@carols10cents could you check if this is indeed working as intended? I didn't have any crates in my local env to check and only used the unit tests for verification. Is there some sort of sandbox env that we could deploy this too?

This PR only includes the backend changes for now until we have verified that those work as intended.

### DIFF
--- a/src/dependency.rs
+++ b/src/dependency.rs
@@ -30,7 +30,6 @@ pub struct Dependency {
 
 pub struct ReverseDependency {
     dependency: Dependency,
-    crate_name: String,
     crate_downloads: i32,
 }
 
@@ -125,9 +124,9 @@ impl Dependency {
 }
 
 impl ReverseDependency {
-    pub fn encodable(self) -> EncodableDependency {
+    pub fn encodable(self, crate_name: &str) -> EncodableDependency {
         self.dependency.encodable(
-            &self.crate_name,
+            crate_name,
             Some(self.crate_downloads),
         )
     }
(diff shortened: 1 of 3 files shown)


carols10cents:
I have a backend setup locally, and we also have staging-crates-io.herokuapp.com that I can deploy to, but there's not too much data there. Will be testing this out in a min!

Turbo87:
@carols10cents would be awesome if you could deploy this to the staging server. there seem to be enough crates on there to verify if everything works as I imagined.

carols10cents:
I just created merge conflicts; I'm going to resolve these, force push, then review/try out/deploy to staging :)

sgrif:
Can you better explain why #306 occurred and how this fixes it in your commit message?

Turbo87:
@sgrif is the cross reference to https://github.com/rust-lang/crates.io/issues/306#issuecomment-310693592 not enough?

carols10cents:
@Turbo87 Bad news, good news, then bad news!

I've got this deployed to https://staging-crates-io.herokuapp.com now, and I'm not sure if the backend changes aren't quite working right or if there's frontend stuff that needs to change?

Here's what I'm seeing:

- Visit https://staging-crates-io.herokuapp.com/crates/rand
- Click on the link to libc under dependencies
- Click the dependent crates link
- The page lists libc twice, where instead it should list rand and num_cpus

The good news is that the original issue seems to be fixed! That is, if I:

- Visit https://staging-crates-io.herokuapp.com/crates/rand
- Click on the link to libc under dependencies
- Click the back button
- The dependency is still listed as libc, rather than the wrong case of changing the dependency name to rand! 🎉

Back to the bad news, this problem happens but it happens on live crates.io too so I'm going to file a new issue 😂 When I:

- Visit https://staging-crates-io.herokuapp.com/crates/libc/reverse_dependencies
- The "back to" link at the top just says "back to", but should say "back to libc"
- Clicking on the "back to" link has no effect
- It says "Displaying 1-2 of 2 reverse dependencies of ", without libc displayed at the end

carols10cents:
I'm going to revert the merge on the master branch, but I'm leaving it deployed to staging for you to play with it :)

Turbo87:
@carols10cents sorry about the confusion, there was indeed still something missing from this PR which is why it was still marked as WIP. I'll reopen as another PR with the necessary frontend changes.

PR #811, opened 2017-06-26
CI: Use recommended yarn installation method

Turbo87:
see https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/.travis.yml#L12-L13


locks:
Merging as we're already using yarn for development.

PR #812, opened 2017-06-26
Update "ember-cli-mirage" to v0.3.3

Turbo87:
We don't seem to use relationships in mirage yet, so their breaking change doesn't require us to change anything.


locks:
Wait, no breaking changes? 😱

Turbo87:
none that are relevant to us here

Issue #626, opened 2017-03-14
Blacklist rust-ci.org


akubera:
`rust-ci DOT org` is no longer a legitimate website for hosting rust-related content, and is now an entrypoint for a dubious ad-network. Looks like there are [~61 tomls on github](https://github.com/search?l=TOML&q=rust-ci.org&type=Code&utf8=%E2%9C%93) using that site for hosting documentation. I don't know if there is a blacklisting procedure in place, but adding that site may be a good idea.

carols10cents:
How should we implement blacklisting? 

- Should we not display the documentation link at all? 
- Should we display the link but with a warning that it's to a known spam site? 
- Should we display "documentation link has been blacklisted" text instead of the link?

That takes care of existing crates that link to a blacklisted domain; should we also disallow blacklisted domains on `cargo publish` to prevent more versions referencing this and alert crate authors of the problem? Should that be a warning or an error?

carols10cents:
I'm seeing 35 crates that have rust-ci.org in their documentation link in crates.io's database:

free_macros
 epsilonz
 crc24
 alfred
 tau
 fmod
 pcre
 noisy
 epsilonz_algebra
 tailrec
 free
 cpuid
 capture
 number_words
 rust-iteratorcomprehensions
 leveldb-rs
 monad_macros
 rope
 rust-GSL
 tweetust_macros
 superchan
 apply_pub
 fingertree
 rust-assimp
 tinycdb
 underscore
 GSL
 cc
 duktape
 jlens
 touptek
 rclist
 pipes
 cld2-sys
 fastcgi

carols10cents:
Excerpts from [discussion on Reddit](https://www.reddit.com/r/rust/comments/601es3/psa_rustciorg_appears_to_be_defunct_the_domain/): 

> Perhaps this isn't the best forum for this, but I've found at least one link on crates.io that led to ads/malware. The crate (rustless) seemed to have a custom domain, which I'm guessing expired and then picked up by the kind of people that pick up old domain names to monetize any traffic.
>
> Has there ever been discussion over using a whitelist of allowed domains to host rust documentation so that users can feel safe that clicking on the Documentation link will lead to a known-safe site?

---

> I feel like this could be better accomplished by a blacklist than a whitelist. I definitely don't want to have malware linked on crates.io, but I also don't think we should be forcing developers to use one of a curated list of 'safe' community-maintained sites which may or may not be just as likely to be dropped as the developer's personal site.
>
[...]

PR #816, opened 2017-06-27
Initial implementation of documentation blacklist

vignesh-sankaran:
Initial work on #626.

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -970,6 +970,7 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
         &*conn,
     )?;
     let max_version = krate.max_version(&conn)?;
+    let documentation_link = check_bad_documentation_link(krate.documentation);
 
     #[derive(Serialize)]
     struct R {
@@ -997,6 +998,16 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
             categories: cats.into_iter().map(|k| k.encodable()).collect(),
         }),
     )
+}
+
+fn check_bad_documentation_link(url: Option<String>) -> Option<String> {
+    let blocked_doc_link = "rust-ci.org";
+
+    match url {
+        Nome => return None,
+        Some(url) => return url.contains(blocked_doc_link),
+    }
+
 }
 
 /// Handles the `PUT /crates/new` route.
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -39,6 +39,8 @@ use util::{RequestUtils, CargoResult, internal, ChainError, human};
 use version::{EncodableVersion, NewVersion};
 use {Model, User, Keyword, Version, Category, Badge, Replica};
 
+const DOCUMENTATION_BLACKLIST: [&'static str; 1] = ["rust-ci.org"];
+
 #[derive(Debug, Clone, Queryable, Identifiable, AsChangeset)]
 pub struct Crate {
     pub id: i32,
@@ -544,13 +546,28 @@ impl Crate {
         }
     }
 
+    // Return None if the documentation URL host matches a blacklisted host
     fn remove_blacklisted_documentation_urls(url: Option<String>) -> Option<String> {
-        let blacklisted_url = "rust-ci.org";
+        let url = match url {
+            Some(url) => url,
+            None => return None,
+        };
 
-        match url {
-            Some(ref url) if url.contains(blacklisted_url) => None,
-            Some(_) => url,
-            None => None,
+        // Handles the case if the documentation URL does not parse successfully
+        let parsed_url = match Url::parse(&url) {
+            Ok(parsed_url) => parsed_url,
+            Err(_) => return None,
+        };
+
+        let url_host = match parsed_url.host_str() {
+            Some(url_host) => url_host,
+            None => return None,
+        };
+
+        if DOCUMENTATION_BLACKLIST.contains(&url_host) {
+            None
+        } else {
+            Some(url)
         }
     }
 
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -548,25 +548,27 @@ impl Crate {
         }
     }
 
-    // Return None if the documentation URL host matches a blacklisted host
+    /// Return `None` if the documentation URL host matches a blacklisted host
     fn remove_blacklisted_documentation_urls(url: Option<String>) -> Option<String> {
+        // Handles if documentation URL is None
         let url = match url {
             Some(url) => url,
             None => return None,
         };
 
-        // Handles the case if the documentation URL does not parse successfully
+        // Handles unsuccessful parsing of documentation URL
         let parsed_url = match Url::parse(&url) {
             Ok(parsed_url) => parsed_url,
             Err(_) => return None,
         };
 
-        // Unwrap is fine here since a None type is handled in the first match statement
+        // Extract host string from documentation URL
         let url_host = match parsed_url.host_str() {
             Some(url_host) => url_host,
             None => return None,
         };
 
+        // Match documentation URL host against blacklisted host array elements
         if DOCUMENTATION_BLACKLIST.contains(&url_host) {
             None
         } else {
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -1,7 +1,6 @@
 use std::ascii::AsciiExt;
 use std::cmp;
 use std::collections::HashMap;
-use std::mem;
 
 use conduit::{Request, Response};
 use conduit_router::RequestParams;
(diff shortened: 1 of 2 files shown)


src/krate.rs
```diff
@@ -1032,6 +1033,10 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
     )?;
     let max_version = krate.max_version(&conn)?;
 
+    // Cheaper than cloning the value to the function
+    let doc_url = mem::replace(&mut krate.documentation, None);
```
sgrif: `let doc_url = krate.documentation.take()`?

vignesh-sankaran:
@sgrif I talked to @carols10cents about this on IRC, I'll let them provide their opinion here but a problem that was raised with an upload-time approach is if the blacklisted URL isn't in the crate database entry, then a crate author asks why their documentation URL isn't showing up, it won't be immediately obvious.

carols10cents:
Yeah I want the URL in the database for my own debugging purposes. What if, instead of calling `documentation_blacklist` from `fn show`, we call it in the `Crate` `encodable` method so that anywhere we encode a crate for returning via JSON, we've blacklisted documentation URLs?

vignesh-sankaran:
Implemented the changes as suggested @carols10cents, was there anything else that needed to be discussed?

src/krate.rs
```diff
@@ -545,6 +548,16 @@ impl Crate {
         }
     }
 
+    fn documentation_blacklist(url: Option<String>) -> Option<String> {
```
sgrif: Can we give this a more descriptive name? How about `remove_blacklisted_documentation_urls`

src/krate.rs
```diff
@@ -1052,6 +1065,8 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
     )
 }
 
+
```
sgrif: Can you delete these extra blank lines?

src/krate.rs
```diff
@@ -505,7 +505,7 @@ impl Crate {
             downloads,
             description,
             homepage,
-            documentation,
+            mut documentation,
```
sgrif: You can remove this `mut` and just do `let documentation =` below. We're never actually mutating anything.

vignesh-sankaran:
Cleaned up the history here as well @sgrif, learned a lot about git :)

Regarding the line `let documentation = Crate::remove_blacklisted_documentation_urls(documentation);`,  is the `documentation` value being passed into the function a copy of the existing variable, or a reference?

carols10cents:
> Regarding the line let documentation = Crate::remove_blacklisted_documentation_urls(documentation);, is the documentation value being passed into the function a copy of the existing variable, or a reference?

It's transferring ownership of the original documentation value to the `Crate::remove_blacklisted_documentation_urls` function; technically it does a copy but that might get optimized away. If it was a reference, the signature of `Crate::remove_blacklisted_documentation_urls` would have `&` in front of its parameters' types, and you'd call it with `&documentation`. The [ownership chapter of the book](https://doc.rust-lang.org/nightly/book/second-edition/ch04-00-understanding-ownership.html) goes into this in more detail :)

It looks like you might be done with this, so I'm going to take a look :)

src/krate.rs
```diff
@@ -545,6 +547,16 @@ impl Crate {
         }
     }
 
+    fn remove_blacklisted_documentation_urls(url: Option<String>) -> Option<String> {
+        let blacklisted_url = "rust-ci.org";
```
carols10cents: This is a great start!! Thank you!! Two improvements I'd like to see before we merge this in, however:

- I think a `contains` is a bit overzealous. While I'm definitely making up a silly case, this implementation would block someone who had their docs published at http://rust-ci.organists.com, and that would be confusing and unexpected both for crate users and for me trying to figure out what was going on. I think we should use [the url crate's `parse` function and `host_str` method](https://docs.rs/url/1.5.1/url/) to parse the documentation URL (if parsing fails, return `None`), extract the host as a string slice, and then compare against the blacklist for exact match.
- It'd be super nice if instead of one hardcoded URL, we had a hardcoded list here (that would contain one item) so that adding another URL to this blacklist would be adding that to the list, not having to change the code. The blacklist could be a constant at the top of the file so that we're not allocating the list every time we encode a crate, something like:

```
const DOCUMENTATION_BLACKLIST: [&'static str; 1] = [
    "rust-ci.org",
];
```

Then in the code, since with the first change we'll have just the host string, we actually *could* use `contains`, since then we'd be asking if the list contains the host exactly :)

Do these changes make sense? Please let me know if you have any questions!

vignesh-sankaran:
Thank you to SneakySnake on IRC for helping me with the latest commit, had some difficulties with handling this many cases before running the blacklisting check :)

src/krate.rs
```diff
@@ -547,13 +549,28 @@ impl Crate {
         }
     }
 
+    // Return None if the documentation URL host matches a blacklisted host
     fn remove_blacklisted_documentation_urls(url: Option<String>) -> Option<String> {
-        let blacklisted_url = "rust-ci.org";
+        let url = match url {
+            Some(url) => url,
+            None => return None,
+        };
```
carols10cents: This is such a perfect case for when `?` is supported on `Option`! (nothing to change here, I'm just noting that I wish there was a prettier way to do this but there isn't so this is fine :))

src/krate.rs
```diff
@@ -547,13 +549,28 @@ impl Crate {
...
+
+        // Handles the case if the documentation URL does not parse successfully
+        let parsed_url = match Url::parse(&url) {
+            Ok(parsed_url) => parsed_url,
+            Err(_) => return None,
+        };
```
carols10cents: I've been trying to see if we can use methods like `map` on the `Result`s and `Option`s in here, but it's just complex enough that I don't think the end result is any clearer than what you have here. I see what you mean about handling lots of cases!! (again nothing to do here I'm just talking)

src/krate.rs
```diff
@@ -39,6 +39,8 @@ use util::{RequestUtils, CargoResult, internal, ChainError, human};
 use version::{EncodableVersion, NewVersion};
 use {Model, User, Keyword, Version, Category, Badge, Replica};
 
+const DOCUMENTATION_BLACKLIST: [&'static str; 1] = ["rust-ci.org"];
```
carols10cents: What do you think about adding a doc comment for this constant explaining why a host might be in this list?

src/krate.rs
```diff
@@ -547,13 +549,28 @@ impl Crate {
         }
     }
 
+    // Return None if the documentation URL host matches a blacklisted host
```
carols10cents: Should this be a doc comment with 3 slashes instead?

src/krate.rs
vignesh-sankaran: Hmm is there an RFC for this somewhere?

src/krate.rs
vignesh-sankaran: Will do :)

src/krate.rs
vignesh-sankaran: I can do that, but will documentation comments appear on private functions?

src/krate.rs
vignesh-sankaran: I did try something fancier involving `if let`, but it was heading to an arrow anti-pattern and you can't have an `if let` statement with the second variable depending on the result of the first. This solution is a lot simpler :).

vignesh-sankaran:
@carols10cents I agree with covering the edge cases, but I felt unit tests shouldn't be in the `tests/` directory since they are meant for integration tests. That, and I'd have to make `remove_blacklisted_documentation_urls()` a public function, which I don't see a need for outside the realm of testing. So I created a new tests module in `src/krate.rs` and put the unit tests there instead. Should we have a long term view of putting unit tests e.g. tests specifically for `Crate::valid_feature_name` into associated modules instead, rather than have them in the `tests/` directory?

src/krate.rs
carols10cents: Yup! https://github.com/rust-lang/rfcs/blob/9c14315070678e71c01f82a41d50b2489747b62c/text/1859-try-trait.md

carols10cents:
> Should we have a long term view of putting unit tests e.g. tests specifically for Crate::valid_feature_name into associated modules instead, rather than have them in the tests/ directory?

Yup, that sounds good to me! There are a few unit tests in a few files, like [category.rs](https://github.com/rust-lang/crates.io/blob/master/src/category.rs#L286). I would like if there were more :)

carols10cents:
Looks good!!! Thank you!!!!! 🍂 📗 💯

carols10cents:
@vignesh-sankaran are you on twitter?

vignesh-sankaran:
Yay, thanks @carols10cents :). Yes I am on Twitter, my handle is @rustbackend.

Issue #702, opened 2017-05-02
Sort crates by number of downloads in the last 90 days


carols10cents:
Part of [RFC 1824](https://github.com/rust-lang/rfcs/blob/master/text/1824-crates.io-default-ranking.md), rust-lang/rust#41616.

The decision made in RFC 1824 was to change the default ordering for crates within keywords and categories to be the number of downloads in the last 90 days rather than the number of downloads for all time.

An implementation question unresolved in that RFC: should we *change* the current sort option "Downloads" to always be "Downloads in the last 90 days" and remove the ability to sort by all-time downloads at all (Option 1), or should we *add* a new option to all sorting menus for downloads in the last 90 days and *keep* the option to sort by all-time downloads (Option 2)?

What needs to be done for either option:

For pages like [crates in the category "Data structures"](https://crates.io/categories/data-structures) or [crates with the keyword "ffi"](https://crates.io/keywords/ffi) (and notably NOT for a [search result page](https://crates.io/search?q=test) which should remain sorted by relevance by default, or the [Browse All Crates page](https://crates.io/crates), which should remain sorted alphabetically by default):

- Change the default ordering from all-time downloads to downloads in the last 90 days only. Actual implementation will depend on which option we go with, but [the diesel query being built here](https://github.com/rust-lang/crates.io/blob/102a91b61517bc0c1907b436fd3a70c9f27f517b/src/krate.rs#L670-L671) will need to be modified at some point.

## What needs to be done if we go with Option 1 (**replacing** sorting by all-time downloads with sorting by downloads in the last 90 days):

- Change the label of the sort, everywhere, from "Downloads" to "Recent Downloads" or similar (fix CSS if necessary to continue to look nice wider)
- Change the number of downloads displayed with each crate, everywhere, to be the number of downloads in the last 90 days so that the numbers match the sort
- Keep the `sort=downloads` query parameter, and just change what it does to the diesel query [here](https://github.com/rust-lang/crates.io/blob/102a91b61517bc0c1907b436fd3a70c9f27f517b/src/krate.rs#L670-L671) and [here](https://github.com/rust-lang/crates.io/blob/102a91b61517bc0c1907b436fd3a70c9f27f517b/src/krate.rs#L682-L683)

## What needs to be done if we go with Option 2 (**adding** sorting by downloads in the last 90 days to the sorting options and **keeping** sorting by all-time downloads as well): 

[...]

PR #817, opened 2017-06-27
overflow evaluating requirement error on inner_join

natboehm:
I'm currently working on issue #702, sorting crates by downloads in the past 90 days. I can't seem to get past the following error when doing an inner_join between crates::table and crate_downloads::table:

```
error[E0275]: overflow evaluating the requirement 
`<diesel::query_builder::BoxedSelectStatement<'_, ((diesel::types::Integer, diesel::types::Text, 
diesel::types::Timestamp, diesel::types::Timestamp, diesel::types::Integer, 
diesel::types::Nullable<diesel::types::Text>, diesel::types::Nullable<diesel::types::Text>, 
diesel::types::Nullable<diesel::types::Text>, diesel::types::Nullable<diesel::types::Text>, 
diesel::types::Nullable<diesel::types::Text>, diesel::types::Nullable<diesel::types::Text>, 
diesel::types::Nullable<diesel::types::Integer>), diesel::types::BigInt, diesel::types::Bool), 
schema::crates::table, _> as diesel::InternalJoinDsl<_, diesel::query_source::joins::Inner, _>>::Output`
      --> src/krate.rs:828:23
       |
   828 |         query = query.inner_join(crate_downloads::table)                                                             
       |                       ^^^^^^^^^^                                                                                    
       |
```

I am at a loss to understand what this error message actually means or what is provoking it. I thought that there might be something going wrong with the imports and the way crates is associated with crate_downloads, however after comparing to and replicating other places in which inner_join is called successfully I can't figure out what is being done differently in this case. 

I read the [Associations documentation](https://nemo157.com/rustdoc-features-example/diesel-impl2/diesel/associations/index.html) as well as the [BoxedDsl documentation](https://nemo157.com/rustdoc-features-example/diesel-impl2/diesel/prelude/trait.BoxedDsl.html#method.into_boxed) and haven't found anything obviously awry.

I also went through the getting started tutorial listed on the diesel website in hopes of finding something fundamental that I'm missing, but most everything done there seems to be done in this case as well. 

The inner_join seems to not error on a single isolated query: 

`let new_query = crates::table.inner_join(crate_downloads::table).select((ALL_COLUMNS));`

This is still commented out in the code with a label of 'works'. Since this standalone query works (compiles), I didn't think that the problem could be due to the crates and crate_downloads tables being set up correctly to do the join. I thought that the problem might instead be coming from the `into_boxed()` function called when initializing `query`, as the error message mentions `BoxedSelectStatement`, however I cannot find any information that supports this.

I have some code for the rest of the query below the inner_join line, however the problem persists with or without the addition of this code. For simplicity I currently have it commented out.

Any help would be very much appreciated! @sgrif @carols10cents

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -19,6 +19,7 @@ use rustc_serialize::json;
 use semver;
 use time::Timespec;
 use url::Url;
+use chrono::NaiveDate;
 
 use app::{App, RequestApp};
 use badge::EncodableBadge;
@@ -38,7 +39,19 @@ use util::{RequestUtils, CargoResult, internal, ChainError, human};
 use version::{EncodableVersion, NewVersion};
 use {Model, User, Keyword, Version, Category, Badge, Replica};
 
-#[derive(Debug, Clone, Queryable, Identifiable, AsChangeset)]
+#[derive(Clone, Queryable, Identifiable, Associations, AsChangeset)]
+#[belongs_to(Crate)]
+#[primary_key(crate_id, date)]
+#[table_name="crate_downloads"]
+pub struct CrateDownload {
+    pub crate_id: i32,
+    pub downloads: i32,
+    pub date: NaiveDate,
+}
+
+#[derive(Debug, Clone, Queryable, Identifiable, AsChangeset, Associations)]
+#[has_many(crate_downloads)]
+#[table_name="crates"]
 pub struct Crate {
     pub id: i32,
     pub name: String,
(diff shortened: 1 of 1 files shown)


carols10cents:
Yep, @sgrif, halp. I am out of ideas on this one :-/

natboehm:
I found this short description of the error message: https://manishearth.github.io/rust-internals-docs/error-index.html#E0275

sgrif:
You can't join a boxed query. The error should just say that the method doesn't exist, but it looks like there's an impl that rustc can't figure out doesn't apply hence the error vomit.

You need to call `inner_join` before you box the query

natboehm:
Okay thanks! I'll give that a go.

PR #819, opened 2017-06-27
[WIP] FastBoot compatibility 

sivakumar-kailasam:
**Pages:**
- [x] Index
- [x] Install
- [x] Logout
- [x] keywords 
- [x] individual keyword
- [ ]  Individual Crate 
- [ ] search
- [ ] Login
- [ ] User


app/controllers/application.js
```diff
@@ -52,17 +57,23 @@ export default Ember.Controller.extend({
...
+        }
     },
 
     stepFlash() {
-        this.set('flashError', this.get('nextFlashError'));
-        this.set('nextFlashError', null);
+        this.setProperties({
+            'flashError': this.get('nextFlashError'),
+            'nextFlashError': null
+        });
```
Turbo87: how is that related to fastboot?

app/mixins/fastboot-utils.js
```diff
@@ -0,0 +1,20 @@
...
+
+    appURL: computed(function() {
+        let url = '';
+        if (this.get('fastboot.isFastBoot')) {
+            let protocol = this.get('fastboot.request.protocol');
+            let host = this.get('fastboot.request.host');
+            url = `${protocol}://${host}`;
+        }
+        return url;
+    }),
```
Turbo87: comment here might be valuable to understand why we need this

app/routes/logout.js
```diff
@@ -1,8 +1,11 @@
 import Ember from 'ember';
+import fetch from 'fetch';
+import FastbootUtils from '../mixins/fastboot-utils';
+
+export default Ember.Route.extend(FastbootUtils, {
 
-export default Ember.Route.extend({
     activate() {
-        Ember.$.getJSON('/logout', () => {
+        fetch(`${this.get('appURL')}/logout`, () => {
```
Turbo87: should probably use `ajax()` for consistency for now

config/environment.js
```diff
@@ -23,6 +23,10 @@ module.exports = function(environment) {
         APP: {
             // Here you can pass flags/options to your application instance
             // when it is created
+        },
+
+        fastboot: {
+            hostWhitelist: ['crates.io', /^localhost:\d+$/]
```
Turbo87: can we add localhost only in dev mode?

app/controllers/application.js
sivakumar-kailasam: Not related. I just grouped couple of individual set calls. Should've probably kept in a different commit.

app/mixins/fastboot-utils.js
```diff
@@ -8,6 +8,12 @@ export default Ember.Mixin.create({
 
     isNotFastBoot: computed.not('fastboot.isFastBoot'),
 
+    /**
+     * When there's a need to raise network requests from the server
+     * during Server Side Rendering(SSR) via FastBoot,
+     * the JS Fetch API doesn't work with absolute urls.
+     * This property gives the URL prefix for those network calls.
+     * */
```
Turbo87: isn't it the other way around? it doesn't work with relative URLs, right?

sivakumar-kailasam: danngit, you're right. I don't think we need it anymore if I use ajax service.

app/controllers/me/index.js
```diff
@@ -1,20 +1,20 @@
 import Ember from 'ember';
-import ajax from 'ic-ajax';
+
+const { inject: { service } } = Ember;
```
Turbo87: this will make it harder for the new modules codemod to transform it into an import from `@ember/service`... I'd rather just keep using the globals for now.

sivakumar-kailasam: Oh so that's why you've used full form everywhere. Got it!

Turbo87: yep, and I don't multi-level destructuring that much 😉

sivakumar-kailasam:
Closing this PR in favor of smaller PRs to get this repo to be fastboot compatible and then land one PR with the fastboot specific changes.

PR #826, opened 2017-06-28
Convert images to SVGs

Turbo87:
This PR converts most of the images in this repository to SVGs based on the contents of the Sketch file and the original PNG files. It also introduces [ember-svg-jar](https://github.com/ivanvotti/ember-svg-jar), which allows us to use inline SVGs and automatically optimizing them using `svgo`.


hannni:
`svgo` seems to still miss at least one optimisation opportunity: 

The image next to the download button is: `<g fill="#C88E1C"><path d="M1 13h12v2H1zM5 2h4v5h3l-5 4.996L2 7h3z"></path></g>`. The `fill` attribute could be in the path tag, making the group tag useless.

It's a bit hard to check all svgs with ember's template magic.

Turbo87:
> It's a bit hard to check all svgs with ember's template magic.

@hannni if you run the dev server you should be able to visit http://localhost:4200/ember-svg-jar/index.html and see all the SVG files in a list

PR #835, opened 2017-06-29
Adds ember-service-worker and a couple plugins

locks:


Turbo87:
@locks can you describe what the service worked does in this configuration?

Turbo87:
It's caching the API calls?

locks:
@Turbo87 oui

Turbo87:
how? what is the behavior for getting up-to-date data?

locks:
@Turbo87 API calls are like any other ajax call, no? It registers the response for a specific URL in cache. Per the documentation, we are now using a fallback strategy, meaning the cache is only served if no connection is detected. If it's not updating while online, we should look into tweaking the strategy.

Turbo87:
hmm... I think we probably shouldn't do this until we have some sort of indication that you're currently offline and viewing cached data or something like that

Turbo87:
what's the caching strategy for the app assets?

tchak:
Maybe we can start by adding service worker for assets and index page? It is pretty safe and will give us a nice boost on second visits. Plus some points in Lighthouse score :)

locks:
@tchak yeah, that's my idea for when I return to contributing here!

Turbo87:
@tchak while I agree that serviceworker support would be good, we should make sure that the error handling in this app is working properly. without serviceworker the page won't load at all if offline, but with serviceworker it will load but won't show any data and timeout at some point.

Turbo87:
I guess we should wait until https://github.com/ember-cli/rfcs/pull/117 has landed before moving forward with this

PR #837, opened 2017-06-29
Fix reverse dependencies query

### LINKED ISSUE #306: Going to dependent crates then back makes the crate look like it depends on itself

Turbo87:
This PR is a rebase of https://github.com/rust-lang/crates.io/pull/810, but this time including the necessary changes to the frontend. I've tested this locally using the staging API and everything seems to work as intended now.

Resolves #306

### DIFF
--- a/app/serializers/dependency.js
+++ b/app/serializers/dependency.js
@@ -1,0 +1,7 @@
+import DS from 'ember-data';
+
+export default DS.RESTSerializer.extend({
+    attrs: {
+        version: 'version_id',
+    },
+});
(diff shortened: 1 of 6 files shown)


Turbo87:
@carols10cents can we merge this?

carols10cents:
testing it out now!

PR #839, opened 2017-06-29
Use ember-metrics to handle Google Analytics

sivakumar-kailasam:
Takes care of async loading of ga.js & disables on ember 'dev' & 'test' environments.


app/mixins/google-pageview.js
```diff
@@ -1,13 +1,14 @@
...
-        if (!window.ga) {
-            return;
-        }
-        return window.ga('send', 'pageview', {
-            page: this.get('url'),
-            title: this.get('url')
+        Ember.run.scheduleOnce('afterRender', this, () => {
+            const page = this.get('url');
+            const title = this.getWithDefault('currentRouteName', 'unknown');
+            Ember.get(this, 'metrics').trackPage({ page, title });
```
Turbo87: Is the afterRender scheduling necessary here?

app/mixins/google-pageview.js
```diff
@@ -1,13 +1,14 @@
...
     notifyGoogleAnalytics: Ember.on('didTransition', function() {
-        if (!window.ga) {
-            return;
-        }
-        return window.ga('send', 'pageview', {
-            page: this.get('url'),
-            title: this.get('url')
+        Ember.run.scheduleOnce('afterRender', this, () => {
+            const page = this.get('url');
+            const title = this.getWithDefault('currentRouteName', 'unknown');
```
Turbo87: `title` had been set to `this.get('url')` too until now. maybe we should keep it like that for now and change that in a separate PR?

app/mixins/google-pageview.js
sivakumar-kailasam: Is it an overkill? Thought I'd keep it there for prioritising the route load over ga stuff

app/mixins/google-pageview.js
sivakumar-kailasam: Yeah thought I'd get opinions on the PR on this. Will switch to current values

app/mixins/google-pageview.js
Turbo87: just noticed that it's mentioned in the ember-metrics README like this, so it's probably fine. still wondering if it's really still needed though.

PR #840, opened 2017-06-29
Expanded loading state for a smoother transition

acorncom:
By expanding the screen shown as part of the loading state, we can make page transitions between routes feel less jerky.

Per @locks he's got more elaborate aspirations in mind, but this might be a nice intermediate step ...

loading-state


carols10cents:
Looks great!! Thank you!! 💨 💨 💨 💨

PR #842, opened 2017-06-30
Fix punctuation in version.hbs

Phrohdoh:


locks:
Nicely spotted, thanks for the contribution!

PR #845, opened 2017-07-01
Derive debug on all the things!

carols10cents:
I was trying to track down a problem and kept being sad by missing Debug derivations. No longer!

### DIFF
--- a/src/categories.rs
+++ b/src/categories.rs
@@ -6,6 +6,7 @@ use toml;
 use db;
 use util::errors::{CargoResult, ChainError, internal};
 
+#[derive(Debug)]
 struct Category {
     slug: String,
     name: String,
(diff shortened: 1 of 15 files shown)


sgrif:
Should we add `#![deny(missing_debug_implementations)]` to the crate root?

carols10cents:
> Should we add #![deny(missing_debug_implementations)] to the crate root?

TIL of that warning! I would like that in theory, but we'd have to add some exceptions or custom implementations too, for example [the `App` struct](https://github.com/rust-lang/crates.io/blob/master/src/app.rs#L17) can't have `Debug` because `r2d2_diesel::ConnectionManager<diesel::pg::PgConnection>`, `oauth2::Config`, and `git2::Repository` don't implement `Debug`. That also means [`AppMiddleware`](https://github.com/rust-lang/crates.io/blob/master/src/app.rs#L36) is the same since it holds an `App`.... I'd accept a PR that does those things, but I probably won't get around to it myself.

sgrif:
The lint only applies to things which could have `#[derive(Debug)]` on them.

PR #846, opened 2017-07-02
Always count version downloads on the UTC date

carols10cents:
So recently I noticed that the `krate::download` test would fail for me if I was running them while I was in a different day than UTC was (that is, after 8pm EDT). This test never fails on travis because travis' machines have their system clocks in UTC.

Being in PDT last week gave me 3 extra hours where I could easily test this :) I discovered that what was happening was the version download made by [the request to the `/download` route in the test](https://github.com/rust-lang/crates.io/blob/master/src/tests/krate.rs#L977-L986) was being recorded as happening in the date in my current timezone, but the queries were asking for dates [based on yesterday in UTC](https://github.com/rust-lang/crates.io/blob/master/src/tests/krate.rs#L1011) (which was today in my current timezone!)

This changes the default value of the date column in the version_downloads table so that it's always using the current date in UTC.

This might cause some weirdness in the download graphs for a crate by day when we deploy it since it'll abruptly change how we're recording things, but oh well. I would argue the old way is incorrect and now we're fixing it :)

Also obligatory:

khan timezones meme

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 use conduit::{Request, Response};
 use conduit_middleware::Middleware;
 use diesel::pg::PgConnection;
+use diesel::Connection;
 use openssl::ssl::{SslConnector, SslConnectorBuilder, SslMethod, SSL_VERIFY_NONE};
 use pg::GenericConnection;
 use pg::tls::{TlsHandshake, Stream, TlsStream};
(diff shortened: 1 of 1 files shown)


sgrif:
What do you think about setting the connection's time zone to UTC upon connection instead?

carols10cents:
How do I do that?

sgrif:
`SET TIME ZONE 'UTC'`

carols10cents:
Made you a bug! https://github.com/diesel-rs/diesel/issues/1024 ❤️ ⏲

carols10cents:
Is this what you meant @sgrif ?

sgrif:
Yeah, this works. We could also set [the connection customizer in R2D2](https://docs.rs/r2d2/0.7.3/r2d2/config/struct.Builder.html#method.connection_customizer)

carols10cents:
Since https://github.com/diesel-rs/diesel/pull/1027 made it in diesel 0.15.0, I'm going to close this. Turned my system clock ahead and the test passes!!!

PR #850, opened 2017-07-02
Start using ember-cli-mirage factories

Turbo87:
Instead of using fixtures it is recommended to use factories to setup acceptance tests. This PR introduces a first set of factories and uses them in a few of the existing tests to simplify them. This PR is not yet converting everything to factories, but can be merged at any point.

@samselikoff would be awesome if you could give this a quick review :)

### DIFF
--- a/mirage/factories/category.js
+++ b/mirage/factories/category.js
@@ -1,0 +1,20 @@
+import { Factory, faker } from 'ember-cli-mirage';
+import Ember from 'ember';
+
+export default Factory.extend({
+    category(i) {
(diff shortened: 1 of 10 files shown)


samselikoff:
Looks good so far! I think for stuff like this

```
features: () => {},
```

You could just write 

```
features: {}
```

Turbo87:
@samselikoff the difference is that with `() => {}` you get a new object for every instance, while with just `{}` the same object instance is used for every factory object.

PR #854, opened 2017-07-06
Port `bin/update-crates` over to Diesel

sgrif:
I thought we were done, but we're not... I forgot the binaries. Sorry.
:(

This one was pretty straightforward. There's a bit of refactoring in
here as well, and I moved some of the integrity checks that this code
was doing to the database. It's worth noting that the new body of
`set_updated_at` is identical to `diesel_set_updated_at`, but I can't
actually call that function from the other since it's `RETURNS trigger`.
Not sure if it's worth resolving, since the function is unlikely to
change.

### DIFF
--- a/src/bin/update-downloads.rs
+++ b/src/bin/update-downloads.rs
@@ -2,25 +2,40 @@
 
 extern crate cargo_registry;
 extern crate chrono;
+#[macro_use]
+extern crate diesel;
+#[macro_use]
+extern crate diesel_codegen;
 extern crate openssl;
-extern crate postgres;
 extern crate semver;
 extern crate time;
 
-use std::collections::HashMap;
+use chrono::NaiveDate;
+use diesel::prelude::*;
+use diesel::pg::PgConnection;
+use diesel::pg::upsert::*;
 use std::env;
 use std::time::Duration;
 
-use cargo_registry::{VersionDownload, Version, Model};
+use cargo_registry::VersionDownload;
+use cargo_registry::schema::*;
 
 static LIMIT: i64 = 1000;
+
+#[derive(Insertable)]
+#[table_name = "crate_downloads"]
+struct CrateDownload {
+    crate_id: i32,
+    downloads: i32,
+    date: NaiveDate,
+}
 
 #[allow(dead_code)] // dead in tests
 fn main() {
     let daemon = env::args().nth(1).as_ref().map(|s| &s[..]) == Some("daemon");
     let sleep = env::args().nth(2).map(|s| s.parse().unwrap());
     loop {
-        let conn = cargo_registry::db::connect_now();
+        let conn = cargo_registry::db::connect_now().unwrap();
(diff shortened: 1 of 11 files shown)


sgrif:
Sorry, there's going to be at least 5 more of these. I forgot the `bin/` folder :(

carols10cents:
rustfmt please :) https://travis-ci.org/rust-lang/crates.io/jobs/250921522

sgrif:
Can we either update to the latest version of rustfmt or dump the config file so that the new version won't change the formatting from what we're currently using?

sgrif:
Hm actually even dumping the config doesn't make the newest version have the same output. :\

carols10cents:
I thought we had dumped the config... https://github.com/rust-lang/crates.io/blob/master/.rustfmt.toml

I'm all for updating to the latest version tho

sgrif:
That config is pretty different from what I get by doing `rustfmt --dump-default-config` with the latest version. Doing that doesn't seem to make much of a difference though. I'm guessing that the newest version's differences are either bug fixes or things that aren't specified in the config file

sgrif:
@carols10cents FYI this is marked as waiting on author but I don't think there's anything else waiting on me

carols10cents:
oh sorry

src/bin/update-downloads.rs
```diff
@@ -31,401 +46,335 @@ fn main() {
...
     // Anything older than 24 hours ago will be frozen and will not be queried
     // against again.
     let now = chrono::UTC::now();
     let cutoff = now.naive_utc().date() - chrono::Duration::days(1);
 
-    let mut map = HashMap::new();
-    for row in rows.iter() {
-        let download: VersionDownload = Model::from_row(&row);
-        assert!(map.insert(download.id, download).is_none());
-    }
```
carols10cents: I am so happy this is getting cleaned up. There was some weird stuff going on in here.

src/bin/update-downloads.rs
```diff
@@ -31,401 +46,335 @@ fn main() {
...
-                             VALUES ($1, $2, $3)",
-                &[&crate_id, &amt, &download.date],
-            )?;
-        }
+        let crate_download = CrateDownload {
+            crate_id: crate_id,
+            downloads: amt,
+            date: download.date,
+        };
+        insert(&crate_download.on_conflict(
```
carols10cents: i ❤️ upsert

src/bin/update-downloads.rs
```diff
@@ -31,401 +46,335 @@ fn main() {
...
-            None,
-        ).unwrap();
-        let version = Version::insert(
-            &tx,
-            krate.id,
-            &semver::Version::parse("1.0.0").unwrap(),
-            &HashMap::new(),
-            &[],
-        ).unwrap();
-        tx.execute(
```
carols10cents: \o/

PR #858, opened 2017-07-08
transfer_crates script: switch to GitHub IDs

vignesh-sankaran:
Redo of #798, learned not to use master for PR's :)

### DIFF
--- a/src/bin/transfer-crates.rs
+++ b/src/bin/transfer-crates.rs
@@ -48,7 +48,7 @@ fn transfer(tx: &postgres::transaction::Transaction) {
     let from = User::find_by_login(tx, &from).unwrap();
     let to = User::find_by_login(tx, &to).unwrap();
 
-    if from.gh_avatar != to.gh_avatar {
+    if from.gh_id != to.gh_id {
         println!("====================================================");
         println!("WARNING");
         println!("");
--- a/src/bin/transfer-crates.rs
+++ b/src/bin/transfer-crates.rs
@@ -67,28 +67,27 @@ fn transfer(tx: &postgres::transaction::Transaction) {
     );
     get_confirm("continue");
 
-
     let stmt = tx.prepare(
         "SELECT * FROM crate_owners
                                    WHERE owner_id = $1
                                      AND owner_kind = $2",
     ).unwrap();
     let rows = stmt.query(&[&from.id, &(OwnerKind::User as i32)]).unwrap();
+
     for row in rows.iter() {
-        let owner_id: i32 = row.get("owner_id");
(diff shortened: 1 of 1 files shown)


sgrif:
@vignesh-sankaran Can you rebase this please? Something funky happened with your branch setup. Shouldn't need 10 commits for a 7 line change

carols10cents:
Hi @vignesh-sankaran, sorry it's taken me so long to review this!

Yes, as sgrif mentioned, please rebase on rust-lang/crates.io master and force push to this branch to get rid of all the merge commits-- please let me know if you have questions about how to do that!

Regarding some of the comments you had on the last PR:

> A couple of things I noticed were that the script exited if an uppercase "Y" was entered

That's fine

>  and that there is an assertion that fails if more than 1 crate is being transferred.

I think that's because when the crate_owners table used to have an id column, and this script used the crate_owners id instead of the owner_id, then this script would update one record each time through the `for` loop. I think now that this table doesn't have an id, and we're using the owner_id and updating all the crates at once, the assertion should be removed. In fact, the UPDATE query can now be moved out of the `for` loop and executed once after the loop. Could you make those changes please?

> As a matter of interest, when is this script utilised?

I believe this is for when someone has created a new github account and wants to transfer all their crates from the old account to the new account. I don't think this happens very often anymore (I don't think I've ever needed to run this script), but there's still a possibility of needing to do this. And actually, now that I think about it, because we're now using github ids and we have a uniqueness constraint on github id, the warning in this script will *always* be printed.... I think it's fine though :)

Thank you! 🌷 🌻

vignesh-sankaran:
@carols10cents I've implemented the changes as requested :)

src/bin/transfer-crates.rs
```diff
@@ -67,29 +67,28 @@ fn transfer(tx: &postgres::transaction::Transaction) {
...
         }
-        let n = tx.execute(
-            "UPDATE crate_owners SET owner_id = $1
-                             WHERE id $2",
-            &[&to.id, &id],
-        ).unwrap();
-        assert_eq!(n, 1);
     }
 
+    let _ = tx.execute(
```
carols10cents: I don't think we need the `let _ =` at all here anymore; the `.unwrap()` should make sure the result is getting used :) Could you remove `let _ =` here please? Thank you!!

carols10cents: I don't think we need the `let _ =` at all here anymore; the `.unwrap()` should make sure the result is getting used :) Could you remove `let _ =` here please? Thank you!!

carols10cents:
Thank you!!! 🍰 💃 🌊

PR #859, opened 2017-07-08
Port `bin/delete-crate` over to Diesel

sgrif:
I moved nearly all of the behavior from this binary into the database.  We could instead just do exactly what this binary was doing before, but it's a lot of behavior that has 0 tests. I wanted to figure out how best to actually test this as well, since any test which actually attempted to delete the crate would have to insert rows in all of these tables to be useable, meaning it would have to know about every foreign key that exists. That test would be brittle, and likely to get out of date.

By moving this into the database, we can also test that any new columns which point to the crates table have a foreign key, and won't break in the future.

Additionally, I've tested the updated binary locally against an old dump of the database.

If we're concerned about someone "accidentally" deleting a crate (and we really shouldn't be. Nobody is running "delete from crates" against the production database, and we have backups), I'd recommend that we just have a trigger that checks some `yes_id_really_like_to_delete_this_crate` entry, or leave exactly 1 constraint uncascading (probably the one on `versions.crate_id`)

### DIFF
--- a/migrations/20170708133715_ensure_crate_id_foreign_keys_cascade/up.sql
+++ b/migrations/20170708133715_ensure_crate_id_foreign_keys_cascade/up.sql
@@ -1,0 +1,21 @@
+ALTER TABLE "crate_downloads"
+    DROP CONSTRAINT "fk_crate_downloads_crate_id",
+    ADD CONSTRAINT "fk_crate_downloads_crate_id" FOREIGN KEY (crate_id) REFERENCES crates(id) ON DELETE CASCADE;
+ALTER TABLE "crate_owners"
+    DROP CONSTRAINT "fk_crate_owners_crate_id",
+    ADD CONSTRAINT "fk_crate_owners_crate_id" FOREIGN KEY (crate_id) REFERENCES crates(id) ON DELETE CASCADE;
+ALTER TABLE "crates_categories"
+    DROP CONSTRAINT "fk_crates_categories_crate_id",
+    ADD CONSTRAINT "fk_crates_categories_crate_id" FOREIGN KEY (crate_id) REFERENCES crates(id) ON DELETE CASCADE;
+ALTER TABLE "crates_keywords"
+    DROP CONSTRAINT "fk_crates_keywords_crate_id",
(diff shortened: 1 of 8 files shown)


migrations/20170708133715_ensure_crate_id_foreign_keys_cascade/up.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE "crate_downloads"
+    DROP CONSTRAINT "fk_crate_downloads_crate_id",
+    ADD CONSTRAINT "fk_crate_downloads_crate_id" FOREIGN KEY (crate_id) REFERENCES crates(id) ON DELETE CASCADE;
+ALTER TABLE "crate_owners"
+    DROP CONSTRAINT "fk_crate_owners_crate_id",
+    ADD CONSTRAINT "fk_crate_owners_crate_id" FOREIGN KEY (crate_id) REFERENCES crates(id) ON DELETE CASCADE;
+ALTER TABLE "crates_categories"
+    DROP CONSTRAINT "fk_crates_categories_crate_id",
+    ADD CONSTRAINT "fk_crates_categories_crate_id" FOREIGN KEY (crate_id) REFERENCES crates(id) ON DELETE CASCADE;
```
carols10cents: Is crates_categories in here just for completeness? Since this constraint already has ON DELETE CASCADE and the down migration is the same?

sgrif: It looks like it doesn't already have it. The down migration is a mistake on my part.

sgrif: (https://github.com/rust-lang/crates.io/blob/b0a7a8d2f4356f38145cbbb05715ed0eedb136ce/migrations/20161115111836_dumped_migration_115/up.sql)

carols10cents: Ahhh I was seeing the `ON DELETE CASCADE` for the category on `crates_categories`, not the crate. 

Would love to see the down migration fixed :)

sgrif: Yup, just wasn't at a computer when I replied. Fixing now. :)

Issue #791, opened 2017-06-21
Need to update text about what `cargo login` does


carols10cents:
API tokens are in the process of being moved to `~/.cargo/credentials` instead of `~/.cargo/config`: https://github.com/rust-lang/cargo/pull/3978

This is currently in the version of cargo shipping with nightly rust.

Before this change makes it to stable, we should update [this text](https://github.com/rust-lang/crates.io/blob/0f3adf36108dfac662e3463b4ff97567261f5de9/app/templates/me/index.hbs#L31) that says you need a `~/.cargo/config` so that people aren't confused.

Since people may still be on an old cargo though, we should either mention both files (like the [cargo docs do now](http://doc.crates.io/crates-io.html#before-your-first-publish) or just not talk about a file at all....

nodakai:
+1 to not mentioning the file path at all. Knowing the path won't help users very much as they aren't supposed to tweak the credentials manually.

Also,
```
cargo login AbcDefgHijkLmnOpqrStuVwXyz
```
we should not encourage users to type their credentials onto their shell prompt.  Other users on the same server may read it with `ps -a` (don't think it's a momentary event; slow NFS might keep the process for tens of seconds.) Users should type their credentials **onto the `cargo login` prompt** as opposed to their shell prompt.

carols10cents:
care to send in a PR @nodakai ?

carols10cents:
Closed by #980!! 🎉

PR #862, opened 2017-07-08
Change ~/.cargo/config to ~/.cargo/credentials in another spot

carols10cents:
PR #847 got one spot, and I think we might have introduced this other
spot while that PR was outstanding.

### DIFF
--- a/app/templates/components/api-token-row.hbs
+++ b/app/templates/components/api-token-row.hbs
@@ -65,7 +65,7 @@
     <div class='row new-token'>
         <div>
             Please record this token somewhere, you cannot retrieve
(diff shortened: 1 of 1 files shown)


vignesh-sankaran:
I've had a look at this @carols10cents, the text appears where it should in the account settings page :). Just a quick question, I believe the switch to ~/.cargo/credentials is happening in the next Cargo version, and is currently not the case with the current version?

carols10cents:
It is not yet in the cargo that is shipping with stable Rust, no, I forget if it's in beta or nightly though. A better solution might actually be not mentioning the filename at all: https://github.com/rust-lang/crates.io/issues/791#issuecomment-315789238 So I think I'm going to close this actually :)

PR #864, opened 2017-07-08
Fix users/stats endpoint erroring when the user doesn't own any crate

kureuil:
Spot this while setting up a local instance today, but it seems production is also affected. Because diesel's `sum` returns an error when the sum result is null, the endpoint returned a 500 error and the Dashboard page would not load up.

The fix I applied is making a separate query to count the number of crates owned. Ideally, we should use [COALESCE](https://www.postgresql.org/docs/9.2/static/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL "PostgresQL COALESCE documentation entry") to make it a single query, but I couldn't find any information in Diesel's documentation.

### DIFF
--- a/src/tests/user.rs
+++ b/src/tests/user.rs
@@ -252,6 +252,27 @@ fn user_total_downloads() {
 }
 
 #[test]
+fn user_total_downloads_no_crates() {
+    let (_b, app, middle) = ::app();
+    let u;
+    {
+        let conn = app.diesel_database.get().unwrap();
+
+        u = ::new_user("foo").create_or_update(&conn).unwrap();
+    }
+
+    let mut req = ::req(app, Method::Get, &format!("/api/v1/users/{}/stats", u.id));
+    let mut response = ok_resp!(middle.call(&mut req));
+
+    #[derive(Deserialize)]
+    struct Response {
+        total_downloads: i64,
+    }
+    let response: Response = ::json(&mut response);
+    assert_eq!(response.total_downloads, 0);
+}
+
+#[test]
 fn updating_existing_user_doesnt_change_api_token() {
     let (_b, app, _middle) = ::app();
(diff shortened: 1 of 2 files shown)


sgrif:
> Because diesel's sum returns an error when the sum result is null, the endpoint returned a 500 error and the Dashboard page would not load up.

That's a bug in Diesel if so. Can you open an issue there with more details about the specific error that was returned? Are you sure it's about the result being NULL and not `diesel::result::Error::NotFound`?

sgrif:
Actually glancing at the source code for `sum`, this is 100% a bug in Diesel. I would strongly prefer that we fix it there, I can release a minor version with the fix. Can you open an issue on Diesel with the exact error that you're seeing?

I'm going to close this since we shouldn't introduce workaround code here for something that should be quick to fix upstream. @carols10cents Feel free to re-open if you disagree.

sgrif:
https://github.com/diesel-rs/diesel/commit/cf32f6f7

I guess I forgot...

carols10cents:
I'm cool with the fix being in diesel. Great catch @kureuil! I definitely want your test though-- @sgrif I'm assuming you'll make a PR when there's a new diesel version? Could you also cherry pick https://github.com/rust-lang/crates.io/pull/864/commits/ad85196feb7b4bf0c90346d5af1edfc5aadbed47 into that PR when you do so? I'd love for @kureuil to get credit for at least some of their work here ❤️

sgrif:
Yes. I've identified the bug in Diesel and will be releasing 0.13.1 and 0.14.1 shortly. I will cherry-pick that commit. Thank you @kureuil! <3

kureuil:
No problem :)
@sgrif do you still want an issue on diesel's repo ?

sgrif:
@kureuil Nope, I've found it. Thank you! ❤️

PR #869, opened 2017-07-09
Show Contents of Readme on Crate Pages

### LINKED ISSUE #81: Show Contents of Readme on Crate Pages
I consider the inline Readme file to be one of the most useful things on npm.

The Readme data is already available (and searchable, cf. #72), but I couldn't find any mention of a plan to show it on crate pages. As this can add a lot of content to a crate page (with possibly complicated markup), a new layout for the page might also be necessary.

kureuil:
Hey everyone, I started working on README rendering earlier and I thought that I should get some feedback now that it is working correctly on my local instance.

I edited the `/crates/new` endpoint so that when a crate is uploaded to the registry, its README (markdown only) is rendered to HTML and a `readmes/{crate}/{crate}-{version}.html` file is created on S3 when the crate file is uploaded there.
Then, when visiting a crate page, the Ember application makes a request to `{s3}/readmes/{crate}/{crate}-{version}.html`: if the file is found, its content is rendered in place, if not nothing happens.
This way, the renderd README is cached and versioned, following @ashleygwilliams [recommendations](https://github.com/rust-lang/crates.io/issues/81#issuecomment-64968314).

On the `crate/version` route, the README contents is displayed in place of the package description.

The generated HTML is sanitized using [Ammonia](https://crates.io/crates/ammonia). This crate removes all tags and attributes that are not whitelisted, making the resulting markup as safe as possible for the end user. Bonus point: is uses the html5ever crate (created by the Servo project) to parse HTML.

~~~Codeblocks syntax highlighting is done server-side using [syntect](https://crates.io/crates/syntect). This crate uses Sublime syntax files to highlight code samples, which means that we automatically have access to a lot of pre-existing grammars and we can easily add/create missing ones.~~~ Syntax highlighting is now done client-side using PrismJS due to syntect's dependency onigumura conflicting with git2.

Finally, I added a new script in `src/bin` named `render-readmes`. Its goal is to render the README of every uploaded crate version and to upload it on S3. This means that if modify the renderer behavior and we want to retroactively apply it, we're just one command (and a lil' bit of time) away ! (Though I might have left too much `unwrap()`/`expect()` in it)

TODO:

* [x] Security checks in `render::markdown_to_html` (script tag, javascript callbacks, more ?)
* [x] Tests
* [x] Documentation
* [x] ~Edit delete-crate & delete-version scripts to delete README files~ It would seem that the scripts only delete database entries, not stored files
* [x] Script to render published crates README (?)

Fixes #81

Preview:

![screenshot from 2017-07-10 00-22-45](https://user-images.githubusercontent.com/1275463/27998080-f83bad7a-6505-11e7-884e-95ee4f7c2a47.png)


kureuil:
@carols10cents This PR should be ready for review now :)

Tell me if you need anything concerning this.

Cargo.toml
```diff
@@ -49,6 +50,9 @@ serde_derive = "1.0.0"
 serde = "1.0.0"
 clippy = { version = "=0.0.142", optional = true }
 chrono = "0.4.0"
+pulldown-cmark = { version = "0.0.15", default-features = false }
+ammonia = "0.5.0"
+syntect = "1.7.1"
```
vignesh-sankaran: For dependency version specification, did we want to leave out the specific patch version e.g. `0.5` compared to `0.5.0`?

kureuil: I don't know, I just copied and pasted the requirement line from the crates' pages ([example](https://crates.io/crates/ammonia "Ammonia's crate page")). If I understand correctly the [cargo documentation](http://doc.crates.io/specifying-dependencies.html#specifying-dependencies-from-cratesio "Specifying dependencies from crates.io"), this is just a stylistic issue. I'd be happy to change it if that's what we want, but I'm not sure this is needed.

vignesh-sankaran: If we leave out the patch version, we get the latest patch version when the Cargo.lock file gets updated via `cargo update`.

kureuil: I'm not sure I understand... The current [lock file](https://github.com/rust-lang/crates.io/blob/master/Cargo.lock#L408) references `diesel 0.14.1` despite having `diesel = 0.14.0` in the [Cargo.toml](https://github.com/rust-lang/crates.io/blob/master/Cargo.toml#L43). Where is the behavior you describe documented ?

vignesh-sankaran: Ah yes you're right, I remember reading about this behaviour in the Rust Book, even if one specifies a specific patch version it'll get updated anyway to the latest one.

src/uploaders.rs
```diff
@@ -54,35 +54,54 @@ impl Uploader {
         }
     }
 
+    pub fn readme_location(&self, crate_name: &str, version: &str) -> Option<String> {
```
vignesh-sankaran: Could you add a documentation comment to this?

src/uploaders.rs
```diff
@@ -102,57 +121,76 @@ impl Uploader {
...
+            checksum,
+            crate_bomb,
+            Bomb {
+                app: app.clone(),
+                path: readme_path,
+            },
+        ))
+    }
+
     pub fn delete(&self, app: Arc<App>, path: &str) -> CargoResult<()> {
```
vignesh-sankaran: Could you put a documentation comment here too?

src/uploaders.rs
vignesh-sankaran: Could you add a documentation comment here?

src/bin/render-readmes.rs
```diff
@@ -0,0 +1,330 @@
...
+                return filepath == path;
+            }
+        })
+        .expect(&format!(
+            "[{}-{}] file is not present: {}",
+            version.krate,
+            version.num,
+            path.display()
+        ));
+    match file {
```
carols10cents: LOL this code was really confusing to me until I looked at the `tar` docs and saw that `tar::Entries` iterates over `Result<Entry>` 😂😂😂

Since everywhere else in `find_file_by_path` is calling `expect` and crashing, wdyt about calling `expect` instead of this `match` and making this method return just `String` rather than `Option<String>`, and removing the `unwrap` on line 200? 

I'm not entirely sure what the error would be here, so I don't know what to *say* in the `expect`. I was going to suggest making the `unwrap` on line 200 be an `expect` that says "file not found" but I see that there's already an `expect` that says just that in this method!!

src/bin/render-readmes.rs
```diff
@@ -0,0 +1,330 @@
...
+                version.krate,
+                version.num
+            ));
+            return Some(contents);
+        }
+    }
+}
+
+/// Creates and Arc over an App instance.
+fn make_app() -> Arc<App> {
```
carols10cents: This duplication with `src/bin/server.rs` is a bit 😿😿😿

Do you think it would be possible to factor out the duplication?

src/render.rs
```diff
@@ -1,18 +1,174 @@
...
+        };
+        MarkdownRenderer {
+            html_sanitizer: html_sanitizer,
+            syntax_set: SyntaxSet::load_defaults_newlines(),
+            theme_set: ThemeSet::load_defaults(),
+        }
+    }
+
+    /// Renders the given markdown to HTML using the current settings.
+    pub fn to_html(&self, text: &str) -> CargoResult<String> {
```
carols10cents: This is awesome!!!!!!

carols10cents:
Again, this is looking awesome!! Since this is such a big change though (which again, is awesome!!!), I want to do a bit more review and testing before getting this out in production.

- [ ] I'd love to get a skim on this from @alexcrichton and @ashleygwilliams. 
  - [ ] Alex, this would render a README per crate version as HTML and upload that HTML to s3 and serve it from there. I know s3 space is cheap, but do you have any concerns around that?
  - [ ] Ashley, are there any things that npm does with its READMEs now that we've forgotten here? I'm about to reread your blog post again to see if I find anything. Things like the list of tags that are whitelisted-- could you help me find those in npm's code and crossreference them? I'd also be happy to jump on a call to discuss if you'd rather.

- [ ] I want to do some performance testing:
  - [ ] Individual publish before and after this change, using the local uploader
  - [ ] Individual publish running locally but uploading to s3
  - [ ] How long it takes the render-readmes script to run for how many versions with realistic READMEs (I want to know approximately how long this will take before running this on production!!)
- Depending on the results of those performance tests, that might sway my thoughts about whether we should do these possible enhancements now or whether we can put them off until we see more of a performance hit:
  - Should we move the readme rendering/uploading into a background job and out of the publish request entirely?
  - We have the README markdown for the most recent published version stored in the database since it's used for search. Would it be worth it to have a render-latest-readmes script that could rerender just those? The use case I'm thinking of is a bug of some sort, and that not having to download the tar files would be faster and let us fix a bug on at least the majority of crate landing pages sooner.
  - If the render-readmes task is going to take a long time on production (like over, idk, 10 min?), I wouldn't want to have to keep rerunning it on everything-- it'd be nice to have a way for it to pick up where it left off if it hits a problem and stops. So perhaps store a `readme_last_rendered` datetime and have a query that gets only the versions whose readmes haven't been rendered since a given time?

@vignesh-sankaran I'm definitely interested in how this affects development-- let's try to get your development environment fixed. Please ping me if you're still having trouble!

src/uploaders.rs
kureuil: I'm on it ! :)

src/uploaders.rs
```diff
@@ -127,134 +121,76 @@ impl Uploader {
...
-                                internal(&format_args!(
-                                    "failed to upload readme to S3: `{}`",
-                                    readme_path
-                                ))
-                            })?;
-                        }
-                        response
-                    };
-                    if handle.response_code().unwrap() != 200 {
-                        if let Err(e) = self.delete(req.app().clone(), &crate_path) {
```
kureuil: This `delete` call is here to remove the uploaded crate if the readme couldn't be uploaded. It technically wasn't removed. In the new code, I'm creating the `Bomb` for the uploaded crate before uploading the readme so that if the readme upload fails a `CargoResult::Error()` is returned, triggering the Bomb and deleting the uploaded crate file. I hope I got the `Bomb` behavior correct.

carols10cents:
Oh ummmm one more thing I need to debug before merging: when I run the tests locally on osx with 1.19.0 (and 1.18.0), I get an invalid memory access 😱😱😱:

```

     Running `/Users/carolnichols/rust/crates.io/target/debug/deps/all-d0ac81dc8b97766c`

running 132 tests
error: process didn't exit successfully: `/Users/carolnichols/rust/crates.io/target/debug/deps/all-d0ac81dc8b97766c` (signal: 11, SIGSEGV: invalid memory reference)
```


This doesn't make much sense to me because it worked on travis! What are you running, @kureuil? Have you ever seen this? @vignesh-sankaran once you get your dev env up, please try running the tests and report back!

src/bin/render-readmes.rs
kureuil: Yeah, I might have done a little bit too much `expect`/`unwrap` for a script supposed to be long-running though it would seem that `panic`'ing in a thread does not affect the main one.

While I'm at it, do we want to refactor the threading model on this script to use a real threadpool and not create a thread for each crate ? It would certainly improve performance but we couldn't `expect`/`unwrap` as much.

src/bin/render-readmes.rs
kureuil: Completely agree with you ! Do you know where we would want this, as a new function in the `app` module (`app::make`) ? As a new method on the `App` structure (`App::from_env`) ? Somewhere else ?

kureuil:
@carols10cents I've never seen this during development of this PR. I wrote this code using Rust 1.18 and nightly-2017-07-04 on Fedora 25 and 26.

Do you have any more logs, or a coredump file ?

src/bin/render-readmes.rs
sgrif: I doubt that we're at the scale where we need to be concerned about performance at that level

vignesh-sankaran:
@carols10cents Yes I have run into that problem in macOS, and I don't know how to figure out what dependency is causing that problem :(. The Cargo.lock file has been updated, I'll try it again. All tests run fine on Linux, specifically Debian Sid.

The problem I'm specifically having is not knowing which part of the fn new() function handling the /crates/new endpoint requires superuser permissions. This could take me a while to figure out since I'll have to log everything happening in that function until I see where the 500 error is being thrown.

kureuil:
@vignesh-sankaran I did not run into this problem when I was developing, and I even tested it with the local uploader. However, I did not test the local upload after the refactor of `Uploader` [63c3a5f](https://github.com/rust-lang/crates.io/pull/869/commits/63c3a5f9f553eb30069dd43d7ae96da883a575d0) as all tests were passing. The source of the problem is the path of the crate that get prefixed at [line 155](https://github.com/kureuil/crates.io/blob/63c3a5f9f553eb30069dd43d7ae96da883a575d0/src/uploaders.rs#L155).
This was only done for S3 before, and is now done for every uploaders, so it is now trying to write to `/crates/{name}` instead of `./dist/local_uploads/crates/{name}`.
Looking at the source code of the S3 crate, it is removing the leading slash of every path given to it. I really don't get why we're doing it in the uploader, I think I'll just remove the leading slash in `src/uploaders`.
@carols10cents any thoughts on this ? Did I miss something ?

vignesh-sankaran:
@kureuil Ah yes that'd also explain the 404 I ran into when trying to run this in the local dev environment, since the directory to upload with READMEs doesn't exist.

The permissions problem actually exists even on master, so I'll still have to look into this. This gives me a starting point to work with though :).

kureuil:
Little update on this, I think I found the source of the segfault. The culprits are git2 *and* syntect and their regex functions. git2 wraps libgit2 which has a regfree function (declared as a weak alias) and syntect uses the onigumura regex library which also defines a regfree function which because it is not declared as a weak alias is overriding libgit2's function. So when libgit2 think it is calling its function it is in reality calling onig's one. I don't even know how it is reliably working on linux... 

I'll remove it for now and will make the syntax highlighting client-side for now. I'll try to submit another PR later when this issue will be fixed.

src/krate.rs
```diff
@@ -1265,6 +1279,28 @@ pub fn download(req: &mut Request) -> CargoResult<Response> {
...
+        struct R {
+            url: String,
+        }
+        Ok(req.json(&R { url: redirect_url }))
+    } else {
+        Ok(req.redirect(redirect_url))
+    }
+}
+
 fn increment_download_counts(req: &Request, crate_name: &str, version: &str) -> CargoResult<()> {
```
vignesh-sankaran: Could we add a documentation comment here too? There is a way to make documentation comments visible for private types via `cargo doc --no-defaults --passes "collapse-docs" --passes "unindent-comments"` I believe, though this may have changed.

src/uploaders.rs
```diff
@@ -54,35 +54,54 @@ impl Uploader {
...
+                Some(format!(
+                    "/local_uploads/{}",
+                    Uploader::readme_path(crate_name, version)
+                ))
+            }
+            Uploader::NoOp => None,
+        }
+    }
+
     fn crate_path(name: &str, version: &str) -> String {
```
vignesh-sankaran: Doc comment please :)

src/uploaders.rs
```diff
@@ -54,35 +54,54 @@ impl Uploader {
...
         // No slash in front so we can use join
         format!("crates/{}/{}-{}.crate", name, name, version)
     }
 
+    fn readme_path(name: &str, version: &str) -> String {
```
vignesh-sankaran: And a doc comment here too

kureuil:
The syntax highlighting is now done on the client-side using prismjs, which should remove the segmentation fault on macOS. Also, I removed the leading slash when uploading a file, so the files are now stored in the correct folder.

@carols10cents Performance-wise, uploading a crate locally with [serde/json's README](https://github.com/serde-rs/json/blob/master/README.md) takes 0.10s using S3 uploader. Running the render-readmes on ~1000 crate versions uploaded to S3 takes 40 seconds on my machine (i7-4600U CPU @ 2.10GHz & 8GB RAM), and I probably was mostly I/O bound.

vignesh-sankaran:
A quick update on this, front end and backend unit tests all pass on Linux (Debian Sid) and macOS, and README rendering also works on Linux. I'm having trouble getting it to work on macOS though, I'm looking into this.

vignesh-sankaran:
Still trying to fix up my development environment up before I get back to reviewing this, I'll put up a plan on how I'm going to test this up soon. This is a pretty big PR to review :)

kureuil:
@carols10cents Do you have any clue about why the build is failing on Rust beta ? I'd like to understand what's happening there...

carols10cents:
> Do you have any clue about why the build is failing on Rust beta ? I'd like to understand what's happening there...

I have no idea. I was hoping it was a fluke :( I'm going to go digging in the beta regression issues I guess...

carols10cents:
(shrug) apparently it's a memory regression?

carols10cents:
I did some `time cargo publish`es locally using a crate with clap's readme:


| uploader | before | after |
|----------|--------|------|
| Local uploader | 1.745s | 1.731s |
| S3 uploader | 2.036s | 34.079s , did not actually upload |

The first time I tried publishing on this branch with S3, the publish timed out. The second time, it said everything succeeded, but neither the crate file nor the readme appear in the s3 bucket. I'm pretty sure I have the s3 bucket configured correctly, because I do see the crate files in the bucket for when i was testing publishing to s3 on master to get the 2.036s time.

I'm going to dig into it a little and see why the upload might not be working for me.

The other thought I had, testing with clap's readme that's really really long, is that in order to get to the links to documentation, repo, dependencies, other versions, etc you now have to scroll a long time 😱 wdyt about:

- Reorganizing the page so that the readme is "last", or
- If the readme is over some length, put the rest of it behind a "details" tag that you can choose to toggle open if you want
- something else?

carols10cents:
Nevermind, it totally worked, I just needed to refresh the bucket in amazon's web ui :P Also the publish took 2.293s this time, totally reasonable. I'm going to blame amazon and guess that they were still setting up my bucket or something.

kureuil:
> Nevermind, it totally worked

Phew 😬

> * Reorganizing the page so that the readme is "last", or
> * If the readme is over some length, put the rest of it behind a "details" tag that you can choose to toggle open if you want
> * something else?

Not sure about which way to go. I'll try to tweak the page and keep you posted :)

carols10cents:
@kureuil I mergred this into master, deployed this to staging, and discovered that the CSP headers now blocked it 🤣 so I added a commit to add the S3 bucket host as a whitelisted connect-src as well.

Now, though, I'm getting a CORS error:

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://alexcrichton-test.s3.amazonaws.com/readmes/blah-types/blah-types-0.1.8.html. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).
```

Did you get readme rendering working from your S3 bucket? If so, was this all you added in your bucket CORS config? The bucket used for staging is alex's so I'll bug him about that and production tomorrow, just want to know what to tell him.

```
<CORSConfiguration>
	<CORSRule>
		<AllowedOrigin>*</AllowedOrigin>
		<AllowedMethod>GET</AllowedMethod>
		<MaxAgeSeconds>3000</MaxAgeSeconds>
		<AllowedHeader>Authorization</AllowedHeader>
	</CORSRule>
</CORSConfiguration>
```

kureuil:
Yeah, I have the CORS config and a custom bucket policy to make all my uploaded objects public, but I think this one is already in place.

```
{
    "Version": "2008-10-17",
    "Id": "Policy1397632521960",
    "Statement": [
        {
            "Sid": "Stmt1397633323327",
            "Effect": "Allow",
            "Principal": {
                "AWS": "*"
            },
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::kureuil-crates/*"
        }
    ]
}
```

kureuil:
So I tried some changes, and the one I prefer is moving the content of the "Links" section in the quick-links list and putting the other sections in the sidebar. What do you think of this layout ?

screenshot from 2017-08-14 22-59-40

carols10cents:
let's queue this up (i'm about to get on a plane so i'm not going to deploy it til later)

bors: r+

carols10cents:
IT'S ALIVE!!!!!! (render-readmes is still running :) but everything is looking great!!!!!

alfiedotwtf:
Is this documented anywhere on crates.io's website? Sorry, I couldn't find any just now

vignesh-sankaran:
@alfiedotwtf Looks like it's rendering [here](https://crates.io/crates/gallop) now :). I'm not across how exactly readme rendering works, but I think it takes a little while to render server side?

alfiedotwtf:
Yeah, I think that's the confusion...

I read about this pull request a while back in "This Week in Rust", and so I added a README.md to my crate this morning, but when I `cargo publish`d, nothing showed up.

Looking at the cargo.io documentation, I couldn't find anything on how to trigger the rendering on the crate page. So I ended up looking at other crates that did have it working, and found that to trigger this feature I needed to add the "readme" config item in Config.toml :)

vignesh-sankaran:
Hmm sounds like something that could be made clearer, could you perhaps open an issue up [in the cargo repository](https://github.com/rust-lang/cargo/issues)? doc.crates.io's source code is currently [in the cargo repo](https://github.com/rust-lang/cargo/tree/master/src/doc) for some reason :).

alfiedotwtf:
Ah, thanks for the pointer... done: https://github.com/rust-lang/cargo/issues/4620

Issue #176, opened 2015-08-06
Cannot prevent the original publisher from publishing


Gankro:
There's a hard check for `user_id == krate.user_id` that should just be removed. I'll do this in the GH-teams feature.

Gankro:
(I successfully published blist 0.0.4 while steveklabnik was the only owner as a proof of concept)

PR #871, opened 2017-07-09
Use "New Module Imports"

Turbo87:
see Ember.js [RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md) and https://github.com/ember-cli/ember-rfc176-data

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -1,9 +1,10 @@
-import Ember from 'ember';
-import Resolver from './resolver';
+import Application from '@ember/application';
 import loadInitializers from 'ember-load-initializers';
+
+import Resolver from './resolver';
 import config from './config/environment';
 
-let App = Ember.Application.extend({
+let App = Application.extend({
     modulePrefix: config.modulePrefix,
     podModulePrefix: config.podModulePrefix,
     Resolver
--- a/app/components/api-token-row.js
+++ b/app/components/api-token-row.js
@@ -1,8 +1,9 @@
-import Ember from 'ember';
+import Component from '@ember/component';
+import { empty, or } from '@ember/object/computed';
 
-export default Ember.Component.extend({
-    emptyName: Ember.computed.empty('api_token.name'),
-    disableCreate: Ember.computed.or('api_token.isSaving', 'emptyName'),
+export default Component.extend({
+    emptyName: empty('api_token.name'),
+    disableCreate: or('api_token.isSaving', 'emptyName'),
     serverError: null,
 
     didInsertElement() {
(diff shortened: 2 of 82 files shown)


app/app.js
```diff
@@ -1,9 +1,10 @@
-import Ember from 'ember';
-import Resolver from './resolver';
+import Application from '@ember/application';
 import loadInitializers from 'ember-load-initializers';
+
+import Resolver from './resolver';
```
locks: can you put resolver before loadInitializers? or did the order change in blueprints?

app/components/api-token-row.js
```diff
@@ -1,8 +1,9 @@
-import Ember from 'ember';
+import { empty, or } from '@ember/object/computed';
+import Component from '@ember/component';
```
locks: Component first

app/components/badge-appveyor.js
```diff
@@ -1,16 +1,18 @@
...
+import { computed } from '@ember/object';
+import { alias } from '@ember/object/computed';
 
-export default Ember.Component.extend({
+export default Component.extend({
     tagName: 'span',
     classNames: ['badge'],
-    repository: Ember.computed.alias('badge.attributes.repository'),
-    branch: Ember.computed('badge.attributes.branch', function() {
+    repository: alias('badge.attributes.repository'),
```
locks: we might want to introduce some newlines here, for readability

app/components/google-jsapi.js
```diff
@@ -1,15 +1,16 @@
...
+import Component from '@ember/component';
+import $ from 'jquery';
 
-export default Ember.Component.extend({
+export default Component.extend({
     tagName: '',
 
     didInsertElement() {
-        Ember.$.getScript('https://www.google.com/jsapi', function() {
+        $.getScript('https://www.google.com/jsapi', function() {
```
locks: we won't have a problem with these?

app/app.js
Turbo87: my convention is usually to import external packages first before importing local files. since `ember-load-initializers` is external and `./resolver` is a local file the resolver is imported afterwards

app/components/google-jsapi.js
Turbo87: I don't think so, why?

locks: hm, I guess these will be short lived anyway, as we'll want to use something fastboot-compatible.

Turbo87: sure, but replacing `$.getScript()` is unrelated to the purpose of this PR

locks: I know, I'm just saying it won't matter soon enough :P

PR #878, opened 2017-07-11
Fixup some fallout from the revert of #865

carols10cents:
Sooo I have already deployed #865 to production, but not the revert. This means the migration got run, which means we need to undo it, so I've added it back here with the up/down swapped. It shouldn't affect anyone who hasn't run the reverted migration.

I also cherry-picked the commit from #865 that removes `badge_type`-- that method is indeed still not being used since we switched to serde, so that commit is fine. In fact, not having this is causing small problems because #807 was made between #865 and the revert, and it adds a new badge, so if I merge in #807 that variant isn't in `badge_type`.


carols10cents:
sgrif approved via dm!

PR #882, opened 2017-07-12
Use QueryResult in place of CargoResult where possible and remove some unused code

jtgeibel:
There are several functions where a QueryResult can be returned instead of the wider range of errors supported by CargoResult.  Basically I just did a single pass through the code and for every function with a `use diesel` I changed the return type and did a `cargo check`.

I'm wondering if there is general support for moving all diesel queries and associated structs into source files under a `model` module.  This is my first pass through the code to start to identify functions that would be candidates for such a module.  On master there are 11 other functions already returning QueryResult.  Then the remaining functions using diesel could be reviewed to factor out the query logic.

/cc @carols10cents @sgrif

### DIFF
--- a/src/badge.rs
+++ b/src/badge.rs
@@ -1,6 +1,5 @@
 use krate::Crate;
 use schema::badges;
-use util::CargoResult;
 
 use diesel::pg::{Pg, PgConnection};
 use diesel::prelude::*;
@@ -66,7 +65,7 @@ impl Badge {
         conn: &PgConnection,
         krate: &Crate,
         badges: Option<&'a HashMap<String, HashMap<String, String>>>,
-    ) -> CargoResult<Vec<&'a str>> {
+    ) -> QueryResult<Vec<&'a str>> {
         use diesel::{insert, delete};
 
         #[derive(Insertable)]
--- a/src/category.rs
+++ b/src/category.rs
@@ -9,8 +9,7 @@ use pg::rows::Row;
 
 use db::RequestTransaction;
 use schema::*;
-use util::errors::NotFound;
-use util::{RequestUtils, CargoResult, ChainError};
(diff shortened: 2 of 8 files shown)


sgrif:
Cool, this seems like a good refactoring to me. I like returning the more narrow type when possible. I don't personally like the idea of moving these out to a `model` module, but I don't feel strongly about it. I'll let Carol give her opinion on it. :)

jtgeibel:
Thanks for the feedback @sgrif.  I'm still getting familiar with the codebase so I'll definitely hold off on moving code around in modules until I can make a more solid case.

I've added another commit with further cleanups.

Mostly for my own future reference, here are some of the main places I've run across that require the more fully featured CargoResult:

**Request parsing, validation, and database connection pool**
* Expected at api_router level
* parse and verify pagination parameters
* json parsing of requests
* req.db_conn()

**Application level**
* add_dependencies() and validate_license() during crate upload
* semver parsing in krate::max_versions()
* git operations, such as in commit_and_push()
* Authorization checks
* Interactions with Github
* TOML parsing
* curl and std::io in uploaders.rs

carols10cents:
Love this!!! Deleting code is the best!!! 

> I don't personally like the idea of moving these out to a model module, but I don't feel strongly about it. I'll let Carol give her opinion on it. :)

Files like src/krate.rs are getting rather unwieldy at this point, so I think it's time to do *something*. I don't feel incredibly strongly about what we do though, I'm open to suggestions!

sgrif:
:(

I was hoping to get to delete all the dead code in one giant commit once I finished the port.

jtgeibel:
Oh, sorry to steal some of glory there.  Congrats on the wave of PRs finishing the port, way to go!

PR #887, opened 2017-07-15
Add default link for documentation to https://docs.rs/{{crate.name}}

dns2utf8:
I made the default for crates that did not specify a path documentation.

### DIFF
--- a/app/templates/components/crate-row.hbs
+++ b/app/templates/components/crate-row.hbs
@@ -30,6 +30,8 @@
         {{/if}}
         {{#if crate.documentation}}
           <li><a href="{{crate.documentation}}">Documentation</a></li>
+        {{else}}
+          <li><a href="https://docs.rs/{{crate.name}}">Documentation</a></li>
         {{/if}}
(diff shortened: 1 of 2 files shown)


carols10cents:
Sorry, [we already have this functionality](https://github.com/rust-lang/crates.io/blob/88a505c4071b799f280a94927ffb0dea340ad11c/app/routes/crate/version.js#L43), it just only works on a crate's page rather than on search result pages. We do an AJAX request to docs.rs to make sure that the docs have built successfully and are available-- it'd be sad if we linked people to a broken page!

We don't want to make 10 ajax requests per search results page, though ([see this discussion](https://github.com/rust-lang/crates.io/blob/88a505c4071b799f280a94927ffb0dea340ad11c/app/routes/crate/version.js#L43)). Docs.rs might provide a way to query for multiple crates' build status at once, I'm not sure if they've added that yet or not. Another option is to make this request from the backend of crates.io and cache the result.

So thank you, but this isn't the way we'd like to implement this :)

PR #892, opened 2017-07-18
Add recent downloads

### LINKED ISSUE #702: Sort crates by number of downloads in the last 90 days

natboehm:
This addresses issue #702, sorting crates by downloads in the past 90 days. The implementation follows Option 2 detailed in the issue, adding a new option to all sorting menus for downloads in the last 90 days and keeping the option to sort by all-time downloads. The number of recent downloads has been added to display below all-time downloads next to each crate, labeled accordingly.

This PR is currently waiting on updates to Diesel, without which it will not function correctly.
( edit: updates to Diesel complete :) )

Fixes #702


vignesh-sankaran:
@natboehm: Could you run a `cargo fmt` against the codebase please? :)

natboehm:
@vignesh-sankaran forgot about that one, will do!

vignesh-sankaran:
@natboehm Thanks for that :). Which version of Diesel were you waiting on in order for this PR to work?

natboehm:
@vignesh-sankaran not sure about the version but I think we're waiting on [this PR](https://github.com/diesel-rs/diesel/pull/1026) to be merged.

sgrif:
I'm going to release tomorrow

vignesh-sankaran:
@sgrif Were you also going to update the Diesel version that crates.io specifies?

sgrif:
@natboehm If you rebase you should have everything you need now

src/krate.rs
```diff
@@ -39,7 +40,17 @@ use util::{RequestUtils, CargoResult, internal, ChainError, human};
 use version::{EncodableVersion, NewVersion};
 use {Model, User, Keyword, Version, Category, Badge, Replica};
 
-#[derive(Debug, Clone, Queryable, Identifiable, AsChangeset)]
+#[derive(Insertable, Queryable, Identifiable, Associations, AsChangeset)]
```
sgrif: `#![deny(missing_debug_implementations, missing_copy_implementations)]` :)

PR #898, opened 2017-07-23
Port `bin/delete-version` to use Diesel

sgrif:
The changes here were broadly similar to those in `delete-crate`. The
child row deletions are no longer needed here, as they cascade in the
database. (Arguably neither of these binaries are really needed anymore
since it can just be done in the database)

### DIFF
--- a/src/bin/delete-version.rs
+++ b/src/bin/delete-version.rs
@@ -8,28 +8,29 @@
 #![deny(warnings)]
 
 extern crate cargo_registry;
-extern crate postgres;
+extern crate diesel;
 extern crate time;
 extern crate semver;
 
+use diesel::prelude::*;
+use diesel::pg::PgConnection;
 use std::env;
 use std::io;
 use std::io::prelude::*;
 
 use cargo_registry::{Crate, Version};
+use cargo_registry::schema::versions;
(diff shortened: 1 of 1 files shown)


src/bin/delete-version.rs
```diff
@@ -8,28 +8,29 @@
...
-    }
+    let conn = cargo_registry::db::connect_now().unwrap();
+    conn.transaction::<_, diesel::result::Error, _>(|| {
+        delete(&conn);
+        Ok(())
+    }).unwrap()
 }
 
-fn delete(tx: &postgres::transaction::Transaction) {
+fn delete(conn: &PgConnection) {
```
vignesh-sankaran: Would it make sense to add doc comments to functions in the `bin/` directory?

sgrif:
@vignesh-sankaran Have you run the migrations? I cannot replicate, and it should be impossible to violate that FK

src/bin/delete-version.rs
carols10cents: I think for the scripts in src/bin, adding documentation more like a man page using docopt or clap would be more appropriate and useful than doc comments. WDYT?

PR #904, opened 2017-07-23
Use diesel from inside the users middleware

sgrif:
I'm a bit worried about the fact that there were not tests which needed
to be updated. It seems like we never actually test the authentication
directly, we always just stub the user into the request.


carols10cents:
bors: r+

PR #909, opened 2017-07-23
Update to latest conduit-cookie

carols10cents:
Will invalidate everyone's cookies, but we're only setting session
cookies right now anyway

Want to get this change:
https://github.com/conduit-rust/conduit-cookie/pull/4


carols10cents:
bors: r+

PR #917, opened 2017-07-26
Add lints to require Debug and Copy on all types that can be

carols10cents:
@sgrif I thought you had said that the lint would know if it wasn't possible to derive Debug.... that doesn't appear to be the case 😢 Soo I'm allowing it in places!


sgrif:
I think it's only for `Copy` that it doesn't warn (since `Copy` cannot be implemented without `derive`)

sgrif:
bors: r+

sgrif:
I don't know how to push a commit to the branch. :\ I guess I'll do this in the web client

sgrif:
bors: r+

PR #919, opened 2017-07-28
Fix keywords that weren't lowercased before inserting

carols10cents:
Background for anyone unfamiliar with crates.io's database schema:

- We've got crates
- We've got keywords
- Crates can have many keywords and keywords can have many crates
- There's a join table `crates_keywords` that contains records associating crate.ids and keyword.ids

This migration fixes invalid data that currently exists in the database. Between
commits d5c563560 and 4c2bd4225, it was possible for keywords to be
created without lowercasing them first. Then, due to this code:

```
pub fn find_or_create_all(conn: &PgConnection, names: &[&str]) ->
QueryResult<Vec<Keyword>> {
    // ...elided...

    keywords::table
        .filter(::lower(keywords::keyword).eq(any(&lowercase_names)))
        .load(conn)
}
```

which selected any keywords that matched the given keywords *regardless*
of case and associated these keywords with crates, many crates now have
BOTH the uppercase and lowercase variants of a keyword associated with
them.

That is, the keywords table contains:

| id | keyword | |
|----|---------|-|
| 1  | foo     | |
| 2  | Foo     | this record should not exist |

And a crate with id 5 that was published with any of:

`keywords = ["foo"]`
`keywords = ["Foo"]`
`keywords = ["FOO"]` etc

in their Cargo.toml metadata would get the following records created in
the crates_keywords table:

| crate_id | keyword_id | |
|----------|------------|-|
| 5        | 1          | this record is correct, for all cases |
| 5        | 2          | this record is incorrect, for all cases |

So this migration removes all the crates_keywords rows associated with
keywords that aren't lowercased and then removes the keywords
themselves.

BUT!

There are some crates that only have these associations currently with
the uppercased keyword! That is, crate id 5 has this in crates_keywords:

| crate_id | keyword_id |
|----------|------------|
| 5        | 2          |

So BEFORE making the fix above, this migration inserts the corresponding
lowercased keyword association for these crates (so 5, 1 in this case)
so that these crates can be treated like all the rest that are
associated with both.

ALSO!

There are some keywords that have only ever been used in their uppercase
variant, so the corresponding lowercase keyword doesn't exist at all.
Such as:

| id | keyword |
|----|---------|
| 3  | BAR     |

In this case, the migration updates the keyword value, leaving the
associated crates_keywords alone.

WHEW!

### DIFF
--- a/migrations/20170728002039_fix_keywords/down.sql
+++ b/migrations/20170728002039_fix_keywords/down.sql
@@ -1,0 +1,1 @@
+-- Not reversible
--- a/migrations/20170728002039_fix_keywords/up.sql
+++ b/migrations/20170728002039_fix_keywords/up.sql
@@ -1,0 +1,68 @@
+-- For all keywords that have a corresponding lowercased keyword,
+-- (these keywords should not have been created but there was a bug
+-- that created them; that has been fixed and no more are being created)
+WITH messed_up_keywords AS (
+    select keywords.id as upper_id, k.id as lower_id
+    from keywords
+    inner join keywords as k on LOWER(keywords.keyword) = k.keyword
+    where LOWER(keywords.keyword) != keywords.keyword
+)
+-- Find all the crates that use the uppercased keyword BUT NOT the lowercased keyword
+-- (many crates are associated with both lower and upper cased because of a bug)
+, messed_up_crates AS (
+    select crate_id, upper_id, lower_id
+    from crates_keywords
+    inner join messed_up_keywords on crates_keywords.keyword_id = messed_up_keywords.upper_id
+    where messed_up_keywords.lower_id not in (
+        select keyword_id
+        from crates_keywords as ck
+        where ck.crate_id = crates_keywords.crate_id
+    )
+)
+-- Associate these crates with the lowercased keyword AS WELL AS the uppercased keyword
+INSERT INTO crates_keywords (crate_id, keyword_id)
+SELECT crate_id, lower_id as keyword_id
+FROM messed_up_crates
+;
+
+-- For all keywords that have a corresponding lowercased keyword,
+-- (this is repeated exactly from above)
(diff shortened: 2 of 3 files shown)


carols10cents:
What I'd love to know from reviewers:

- Is this SQL doing what I want it to do?
- Are there any unforeseen consequences of this SQL?
- Am I forgetting any cases that might exist based on my description of the bug?

srobbibaro:
@carols10cents - I'll give it a look over. Since I don't know the schema (at all), I'll look more toward the general SQL side of it.

srobbibaro:
@carols10cents - Absolutely wonderful write up in the PR description and in the query comments! It made it so easy for somebody like me, who has never seen this database, to understand exactly what the problem is and how you intend to fix it.

With that said, I believe the SQL will do what you want it to. There might have been a small opportunity for an optimization using a `NOT EXISTS` (see: https://stackoverflow.com/questions/173041/not-in-vs-not-exists), but if you know the data set to be "reasonable," it's probably not worth even making the change--especially if you've already tested it out.

Hopefully that review helps. I'm going to follow this ticket. Let me know when you run it - I'm curious to hear how it goes!

LGTM 💯

carols10cents:
Thank you so much @srobbibaro!!! I miss having you and the other TTMers around to review my SQL all the time ❤️ ❤️ ❤️ ❤️

srobbibaro:
It was fun to review - just like old times. Happy to help out! :)

carols10cents:
I'm feeling good about this and I'd like to get the database cleaned up, so I'm going to merge this, test on staging, and deploy :)

bors r+

carols10cents:
Deployed to staging where i had set up keywords and crates in each of these conditions, the migration did everything i expected it to and nothing I didn't expect it to so we're going to prod!

carols10cents:
prod is looking much better now!!!

PR #899, opened 2017-07-23
Start to add lots of documentation

carols10cents:
I started this a while ago, but an email from @vignesh-sankaran prompted me to get what I have out there rather than bitrotting on my machine.

I'd love a review of what I have so far, @vignesh-sankaran! What do you think of the outline of the architecture document? What kinds of things would make that document a more useful high-level overview? 

The doc comments should hopefully make the docs at https://docs.rs/cargo-registry/ better, I don't think the little bit I've added here warrants a version bump and publish yet, though.

PR #920, opened 2017-07-28
Doc comments on new pull requests

vignesh-sankaran:
Along the lines of #899, we should start mandating doc comments on all new types and functions, and encourage doc comments on modified types. 

Is there a way to check for this in TravisCI as well?


carols10cents:
Once we have everything documented, we can add `#![deny(missing_doc)]` to src/lib.rs to make sure docs get added to new items :)

bors: r+

carols10cents:
bors retry

Issue #808, opened 2017-06-26
Allow editing your own email address


carols10cents:
We save the email address that we get through github oauth, but you can have multiple email addresses or no email address publicly viewable from your github account.

We don't actually use user email anywhere right now, but before we add anything that might use it (ex: alerts/notifications), we should make it editable (and confirm email address before sending email to it)

carols10cents:
@natboehm is working on this!

carols10cents:
Here's how npm is rolling out email verification: http://blog.npmjs.org/post/163187838935/upcoming-change-verified-email-required?utm_campaign=newsletter20170720&utm_source=newsletter_mailer&utm_medium=email

carols10cents:
@natboehm there were some questions about what the issues are around email and how we're planning to implement emails over in #924. Could you please write up what we've talked about? Here are some questions from over there too: 

- > Also to address the email thing, when I was working with the passport-github package, I found out that they actually do a [separate API call](https://github.com/jaredhanson/passport-github/blob/451db704e3b7e6b72266da3e1fa8f36496cd8881/lib/strategy.js#L140) to fetch the email after authentication. Maybe that could work as a longer term solution for getting emails? It's not ideal, but it does seem to work. 
- > By mailer service do you mean something that sends emails? What about using something existing like mailgun? The pricing is really great even at scale.
- > Maybe we can even figure out a way to fetch emails lazily (and then store them) as we need to.
- > To prevent being spammy, it might be prudent to also ask people if we can email them at all.

natboehm:
The plan for user emails we've come up with is:
1. Add a field to edit a user's email address
2. Ability to send a confirmation email to the user for verification
3. Figure out a way to make adding an email mandatory

(1) is almost completed in PR #921. 

For (2), the plan was to send the user a verification email on the addition or edit of their email address. We would generate a token in the database and send in the email as part of a link. Once the user clicks the link, we would look up by the token, see if the user matches, and if the account is found, the email is confirmed. I've been looking into doing this yesterday and today, and found the crate [lettre](https://github.com/lettre/lettre) that might be helpful used in combination with the Heroku app [Mailgun](https://elements.heroku.com/addons/mailgun) to be able to send an email from the app to the user.

[...]

PR #921, opened 2017-07-28
Ability to add/update user email

natboehm:
This PR addresses issue #808, allow editing your own email address. This is the first step in a three part implementation, adding a field to edit an email address. Yet to be implemented before the issue is resolved is (1) the ability to verify emails with users and (2) figuring out how to get everyone to add their email to their account. This PR only implements the addition of a button to the email field such that a user's email can be added or updated, as well as hiding the email field from public view by separating `EncodableUser` into `EncodablePublicUser` and `EncodablePrivateUser`.

### DIFF
--- a/app/components/email-input.js
+++ b/app/components/email-input.js
@@ -17,12 +17,12 @@ export default Component.extend({
         },
 
         saveEmail() {
-            var userEmail = this.get('value');
-            var user = this.get('user');
+            let userEmail = this.get('value');
+            let user = this.get('user');
 
-            var emailIsProperFormat = function(userEmail) {
-                var regExp = /\S+@\S+\.\S+/;
-                return egExp.test(userEmail);
+            let emailIsProperFormat = function(userEmail) {
+                let regExp = /\S+@\S+\.\S+/;
+                return regExp.test(userEmail);
             };
 
             if (!emailIsProperFormat(userEmail)) {
@@ -31,7 +31,17 @@ export default Component.extend({
             }
 
             user.set('email', userEmail);
-            user.save();
+            user.save()
+                .then(() => this.set('serverError', null))
+                .catch(err => {
+                    let msg;
+                    if (err.errors && err.errors[0] && err.errors[0].detail) {
+                        msg = `An error occurred while saving this email, ${err.errors[0].detail}`;
+                    } else {
+                        msg = 'An unknown error occurred while saving this token.';
+                    }
+                    this.set('serverError', msg);
+                });
 
             this.set('isEditing', false);
             this.set('notValidEmail', false);
--- a/app/components/email-input.js
+++ b/app/components/email-input.js
@@ -13,16 +13,16 @@ export default Component.extend({
 
     actions: {
         editEmail() {
-            let user = this.get('user');
-            let isEmailNull = function(user) {
-                if (user.email == null) {
+            let email = this.get('value');
+            let isEmailNull = function(email) {
+                if (email == null) {
                     return true;
                 } else {
                     return false;
                 }
             };
 
-            this.set('emailIsNull', isEmailNull(user));
+            this.set('emailIsNull', isEmailNull(email));
             this.set('isEditing', true);
             this.set('prevEmail', this.get('value'));
         },
--- a/app/components/email-input.js
+++ b/app/components/email-input.js
@@ -1,0 +1,30 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+    type: '',
+    value: '',
+    isEditing: false,
+    user: null,
+
+    actions: {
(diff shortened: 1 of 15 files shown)


src/tests/all.rs
```diff
@@ -189,6 +189,19 @@ fn new_user(login: &str) -> NewUser {
     }
 }
 
+// I want to be able to seed the github id for use in test
```
sgrif: Since this is only used in one place, how about just having that place just do:

```rust
NewUser { gh_id: whatever, ..new_user(login) }
```

src/tests/user.rs
```diff
@@ -66,12 +66,18 @@ fn user_insert() {
...
     let (_b, app, middle) = ::app();
-    let mut req = ::req(app, Method::Get, "/me");
+    let mut req = ::req(app.clone(), Method::Get, "/me");
     let response = t_resp!(middle.call(&mut req));
     assert_eq!(response.status.0, 403);
 
-    let user = ::mock_user(&mut req, ::user("foo"));
-
+    // with GET /me update gives 404 response
+    // let user = ::mock_user(&mut req, ::user("foo"));
```
sgrif: How about `::sign_in(&mut req, &app)`?

natboehm: I'm not sure I understand the difference, could you clarify? I forgot to delete these commented out lines, what I was getting at is that `mock_user()` seems to do the same thing as calling `new_user()` and `sign_in_as(user)` used below except below doesn't cause the 404.

src/tests/all.rs
natboehm: Cool, that makes more sense, thanks!

app/templates/components/email-input.hbs
```diff
@@ -1,27 +1,32 @@
-<div class='row'>
-    <div class='label'>
-        <dt>Email</dt>
-    </div>
-    {{#if isEditing }}
+{{#if isEditing }}
```
carols10cents: Good call on moving the check for editing out to be able to separate the markup between editing and not editing :) 💯

src/tests/user.rs
```diff
@@ -291,3 +297,76 @@ fn updating_existing_user_doesnt_change_api_token() {
     assert_eq!("bar", user.gh_login);
     assert_eq!("bar_token", user.gh_access_token);
 }
+
+/*  Email GitHub private overwrite bug
+    Please find a better description, that is not english
```
carols10cents: 😂 I know that feel :)

src/tests/user.rs
```diff
@@ -291,3 +297,76 @@ fn updating_existing_user_doesnt_change_api_token() {
...
+        ::sign_in_as(&mut req, &user);
+        user
+    };
+
+    let mut response = ok_resp!(middle.call(req.with_path("/me").with_method(Method::Get)));
+    let r = ::json::<R>(&mut response);
+    assert_eq!(r.user.email, None);
+    assert_eq!(r.user.login, "mango");
+
+    let body = r#"{"user":{"email":"mango@mangos.mango","name":"Mango McMangoface","login":"mango","avatar":"https://avatars0.githubusercontent.com","url":"https://github.com/mango","kind":null}}"#;
```
carols10cents: lolololol @ "Mango McMangoface" :) :) :)

src/user/mod.rs
```diff
@@ -381,11 +381,27 @@ pub fn logout(req: &mut Request) -> CargoResult<Response> {
...
+    // Changed to getting User information from database because in
+    // src/tests/user.rs, when testing put and get on updating email,
+    // request seems to be somehow 'cached'. When we try to get a
+    // request from the /me route with the just updated user (call
+    // this function) the user is the same as the initial GET request
+    // and does not seem to get the updated user information from the
+    // database
+    // This change is not preferable, we'd rather fix the request,
+    // perhaps adding `req.mut_extensions().insert(user)` to the
+    // update_user route, however this somehow does not seem to work
```
carols10cents: This is a great comment!! I'm so excited to find this later when I'm trying to remember why we had to do this this way and this will explain everything ❤️

app/components/email-input.js
```diff
@@ -0,0 +1,66 @@
...
+    disableSave: empty('user.email'),
+    notValidEmail: false,
+    prevEmail: '',
+    emailIsNull: true,
+
+    actions: {
+        editEmail() {
+            let email = this.get('value');
+            let isEmailNull = function(email) {
+                if (email == null) {
```
natboehm: I think this might be special for models? I'm not sure exactly how it works on line 9 but when I print out the result of empty it isn't just a boolean type.

carols10cents:
hmmm

bors: r+

PR #925, opened 2017-08-01
Always bump the `updated_at` on uploads

### LINKED ISSUE #908: "Last updated" on a crate page isn't updating on new version publish
ex: https://crates.io/crates/conduit-cookie, it says last updated 4 months ago but 0.8.4 was published yesterday 

screen_shot_2017-07-23_at_6_20_37_pm

notriddle:
Fixes #908

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -135,7 +135,7 @@ pub struct CrateLinks {
     pub reverse_dependencies: String,
 }
 
-#[derive(Insertable, AsChangeset, Default, Debug)]
+#[derive(Insertable, AsChangeset, Debug)]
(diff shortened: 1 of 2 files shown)


vignesh-sankaran:
Could you run `cargo fmt` and push the code up again please? :)

sgrif:
This is the wrong solution. We should never be managing `updated_at` in the application. It looks like the issue here is that we are missing a trigger on `versions` to set `updated_at` on the crate, which is the reason it should be getting bumped.

notriddle:
This pull request was obsoleted by #931

PR #927, opened 2017-08-01
Add steps to avoid sudo for npm and node on Linux

### LINKED ISSUE #809: Possible anti-pattern for using sudo to install node/npm/yarn in CONTRIBUTING.md
CONTRIBUTING.md has the following note, which I believe is an anti-practice:

    > Note that you may need to install these as root using sudo in some cases.

As far as I know, installing node/npm/yarn with sudo is not recommended and a sign that something went wrong with the installation process.
Can someone else confirm that we should remove it? cc @ashleygwilliams

vignesh-sankaran:
Resolves #809 by recommending `nvm` to install `node` and `npm` on Linux

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -73,7 +73,7 @@ as well.
 
 #### Frontend requirements
 
-In order to run the frontend, you will need to have installed:
+In order to run the frontend on Windows and macOS, you will need to have installed:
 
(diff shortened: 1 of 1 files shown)


carols10cents:
bors: r+

vignesh-sankaran:
Think bors is stuck again?

carols10cents:
bors: r+

PR #931, opened 2017-08-01
Touch a crate whenever its versions are changed

### LINKED ISSUE #908: "Last updated" on a crate page isn't updating on new version publish

sgrif:
Prior to this commit, if uploading a new version of a crate didn't
otherwise change the metadata (description, etc), the crate's updated_at
record wouldn't be changed.

The migration specifically checks if `updated_at` changed on the
versions to make sure the touch doesn't happen if `downloads` was
bumped. I wanted to avoid duplicating that here, so instead this will
only happen if the row is new, or as a result of the trigger bumping
updated_at on the version.

I have specifically not included a query to correct existing data, as
it's an irreversable operation. We should manually run a query like
`UPDATE crates SET updated_at = MAX(crates.updated_at,
SELECT(MAX(versions.updated_at) FROM versions WHERE versions.crate_id =
crates.id))` if we want to correct existing data.

Fixes #908.
Close #925.

### DIFF
--- a/migrations/20170801133919_touch_crates_when_versions_modified/down.sql
+++ b/migrations/20170801133919_touch_crates_when_versions_modified/down.sql
@@ -1,0 +1,2 @@
+DROP TRIGGER touch_crate ON versions;
+DROP FUNCTION touch_crate_on_version_modified();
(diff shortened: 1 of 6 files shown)


carols10cents:
I wanted to deploy the readme change by itself, so since that's out now:

bors: r+

PR #936, opened 2017-08-04
Add more documentation

vignesh-sankaran:
As part of the effort to better document crates.io, I added more documentation comments to the `krate::index` and `krate::new` functions as well as to some associated functions. Also added some comments in some functions and to document. Thanks for providing the background on these functions @carols10cents :).

I'd like to doc comment the git::yank function as well before this gets merged.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -122,7 +122,7 @@ pub fn yank(app: &App, krate: &str, version: &semver::Version, yanked: bool) -> 
 /// on Heroku, and they race against each other e.g. if 2 pushes occur,
 /// then one will succeed while the other will need to be rebased before
 /// being pushed.
-/// 
+///
 /// A maximum of 20 attempts to commit and push to the index currently
 /// accounts for the amount of traffic publishing crates, though this will
 /// eventually need to be changed past a certain point.
--- a/src/git.rs
+++ b/src/git.rs
@@ -124,7 +124,7 @@ pub fn yank(app: &App, krate: &str, version: &semver::Version, yanked: bool) -> 
 /// being pushed.
 ///
 /// A maximum of 20 attempts to commit and push to the index currently
-/// accounts for the amount of traffic publishing crates, though this may 
+/// accounts for the amount of traffic publishing crates, though this may
 /// have to be changed in the future.
 ///
 /// Notes:
--- a/src/git.rs
+++ b/src/git.rs
@@ -77,8 +77,6 @@ pub fn add_crate(app: &App, krate: &Crate) -> CargoResult<()> {
 /// file, deserlialise the crate from JSON, change the yank boolean to
 /// `true` or `false`, write all the lines back out, and commit and
 /// push the changes.
-// TODO: factor out the tasks this function does into smaller, separate
-// functions.
 pub fn yank(app: &App, krate: &str, version: &semver::Version, yanked: bool) -> CargoResult<()> {
     let repo = app.git_repo.lock().unwrap();
     let repo_path = repo.workdir().unwrap();
@@ -118,6 +116,21 @@ pub fn yank(app: &App, krate: &str, version: &semver::Version, yanked: bool) -> 
     })
 }
 
+/// Commits and pushes to the crates.io index.
+///
+/// There are currently 2 instances of the crates.io backend running
+/// on Heroku, and they race against each other e.g. if 2 pushes occur,
+/// then one will succeed while the other will need to be rebased before
+/// being pushed.
+/// 
+/// A maximum of 20 attempts to commit and push to the index currently
+/// accounts for the amount of traffic publishing crates, though this will
+/// eventually need to be changed past a certain point.
+///
+/// Notes:
+/// Currently, this function is called from the HTTP thread and is blocking.
+/// This could be changed to run from a different thread and use a callback
+/// upon completion to the HTTP thread.
 fn commit_and_push<F>(repo: &git2::Repository, mut f: F) -> CargoResult<()>
 where
     F: FnMut() -> CargoResult<(String, PathBuf)>,
--- a/src/git.rs
+++ b/src/git.rs
@@ -73,6 +73,12 @@ pub fn add_crate(app: &App, krate: &Crate) -> CargoResult<()> {
     })
 }
 
+/// Yanks or unyanks a crate version. This requires finding the index
+/// file, deserlialise the crate from JSON, change the yank boolean to
+/// `true` or `false`, write all the lines back out, and commit and
+/// push the changes.
+// TODO: factor out the tasks this function does into smaller, separate
+// functions.
 pub fn yank(app: &App, krate: &str, version: &semver::Version, yanked: bool) -> CargoResult<()> {
     let repo = app.git_repo.lock().unwrap();
     let repo_path = repo.workdir().unwrap();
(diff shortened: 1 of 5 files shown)


src/krate.rs
```diff
@@ -785,6 +785,27 @@ impl Model for Crate {
...
+/// front end, including:
+/// - Alphabetical listing of crates
+/// - List of crates under a specific owner
+/// - Listing a user's followed crates
+///
+/// Notes:
+/// The different use cases this function covers is handled through passing
+/// in parameters in the GET request.
+///
+/// This function was has had functionality bolted on over the years, this
```
carols10cents: "This function was has had" => I think the "was" can come out :)

"this needs to stop" => I would say "we would like to stop adding more functionality in here" :)

src/krate.rs
```diff
@@ -785,6 +785,27 @@ impl Model for Crate {
...
+/// needs to stop :). It was built like this to keep the number of database
+/// trasnactions low, though given Rust's low performance overhead, this is
```
carols10cents: I would say database queries rather than database transactions here and on line 801. Everything in this function should just be requesting data from the database rather than modifying it, so there aren't any transactions that start and can be rolled back like in `krate::new`.

src/krate.rs
```diff
@@ -913,6 +934,8 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
         ));
     }
 
+    // The database query returns a struct within a struct, with the root
+    // struct containing 3 items.
```
carols10cents: These are tuples actually, not structs.

src/krate.rs
```diff
@@ -1100,6 +1123,7 @@ pub fn show(req: &mut Request) -> CargoResult<Response> {
 }
 
 /// Handles the `PUT /crates/new` route.
+/// Used by `cargo publish` to add a new crate or publish an existing crate.
```
carols10cents: I would say "to publish a new crate or to publish a new version of an existing crate".

src/krate.rs
```diff
@@ -1140,6 +1164,8 @@ pub fn new(req: &mut Request) -> CargoResult<Response> {
         };
 
         let license_file = new_crate.license_file.as_ref().map(|s| &**s);
+        // Create a transaction on the database, if there are no errors,
+        // commit the transactions to record a new or updated crate.
```
carols10cents: The transaction actually starts with the `conn.transaction` call [up here](https://github.com/rust-lang/crates.io/pull/936/files#diff-7ab30501521c288e1d8e1e994692b1a2L1129), I think this comment should be moved to go with that.

src/krate.rs
```diff
@@ -1236,6 +1262,13 @@ pub fn new(req: &mut Request) -> CargoResult<Response> {
     })
 }
 
+/// Used by `krate::new` function.
+///
+/// This function parses the JSON headers for validity before returning the
```
carols10cents: I wouldn't say it parses the JSON *for* validity, it parses the JSON so that we can interpret the data, and during and after parsing it is also checking the validity of the data :)

src/krate.rs
```diff
@@ -1236,6 +1262,13 @@ pub fn new(req: &mut Request) -> CargoResult<Response> {
...
+/// crate metadata and user information.
+///
+/// Notes:
+/// This
```
carols10cents: Oops, looks like a sentence trailed off here!

src/lib.rs
```diff
@@ -164,6 +166,7 @@ pub fn middleware(app: Arc<App>) -> MiddlewareBuilder {
         "/crates/:crate_id/reverse_dependencies",
         C(krate::reverse_dependencies),
     );
+    // Used to generate download graphs
```
carols10cents: I might have led you astray, and I just checked to be sure-- I think this comment should go with the `/crates/:crate_id/downloads` route defined [up here](https://github.com/rust-lang/crates.io/pull/936/files#diff-b4aea3e418ccdb71239b96952d9cddb6L151).

src/version.rs
```diff
@@ -461,6 +461,8 @@ pub fn authors(req: &mut Request) -> CargoResult<Response> {
 }
 
 /// Handles the `DELETE /crates/:crate_id/:version/yank` route.
+/// This does not delete a crate; it makes a crate accessible
+/// only to Cargo.lock files with the yanked version.
```
carols10cents: I would say "it makes a crate accessible only to crates that already have a Cargo.lock containing this version." We could also elaborate a bit more on why deleting isn't implemented (deleting a crate could break people's builds) and that yanking is meant to prevent crates from *starting* to depend on a crate.

src/upload.rs
```diff
@@ -1,3 +1,6 @@
+//! This module handles the expected information a crate should have
+//! and manages the serialising and deserialising of this information
```
carols10cents: Oh I was actually surprised to see *serialize* stuff in here, so I investigated a bit, and I'm pretty sure the serialization is only used in tests. I think that'd be good to note in this comment :)

vignesh-sankaran: Maybe we could look into moving the serialize stuff into the `tests/` module at some point?

carols10cents: Yeah, except I don't think we'll be able to move the `#[derive(Serialize)]`s.

vignesh-sankaran: I think so too, because the only way I see it being separated is to duplicate these structs in `src/tests/`, which isn't ideal.

src/git.rs
```diff
@@ -112,6 +116,21 @@ pub fn yank(app: &App, krate: &str, version: &semver::Version, yanked: bool) ->
...
+/// being pushed.
+///
+/// A maximum of 20 attempts to commit and push to the index currently
+/// accounts for the amount of traffic publishing crates, though this will
+/// eventually need to be changed past a certain point.
+///
+/// Notes:
+/// Currently, this function is called from the HTTP thread and is blocking.
+/// This could be changed to run from a different thread and use a callback
+/// upon completion to the HTTP thread.
```
carols10cents: I wouldn't say that it would callback to the HTTP thread, since that means the HTTP thread would still be waiting. The point of spawning a thread would be that the HTTP request could return a response and be finished without having to wait for the git push to complete (or uploading to S3, or rendering the README, etc). That would mean we wouldn't return success or failure to `cargo publish` at all, we'd say something like "publish in progress, check [something] for status" and we'd have to figure out how we want the user to be able to check for status (run `cargo publish --status`, check in crates.io's UI, we send them an email, etc).

src/git.rs
```diff
@@ -112,6 +116,21 @@ pub fn yank(app: &App, krate: &str, version: &semver::Version, yanked: bool) ->
...
+/// Commits and pushes to the crates.io index.
+///
+/// There are currently 2 instances of the crates.io backend running
+/// on Heroku, and they race against each other e.g. if 2 pushes occur,
+/// then one will succeed while the other will need to be rebased before
+/// being pushed.
+///
+/// A maximum of 20 attempts to commit and push to the index currently
+/// accounts for the amount of traffic publishing crates, though this will
+/// eventually need to be changed past a certain point.
```
carols10cents: I'm not *sure* that the 20 attempts *will* need to be changed, it *might* need to be changed.

carols10cents:
I wanted to deploy the readme change by itself, so since that's out now:

bors: r+

carols10cents:
oops

bors: r+

Issue #937, opened 2017-08-04
Broken dev environment with conduit cookie crate update


vignesh-sankaran:
After #909, I now get ``thread 'main' panicked at 'bad master key length: expected at least 32 bytes, found 6', /home/vignesh/.cargo/registry/src/github.com-1ecc6299db9ec823/cookie-0.9.1/src/secure/key.rs:53
note: Run with `RUST_BACKTRACE=1` for a backtrace.`` with `cargo run --bin server`, and all the integration tests fail :(. This is a problem on macOS and Linux.

kureuil:
Requirements for the secret key have changed in #909, it now has to be at least 32 bytes long: https://docs.rs/cookie/0.9.1/cookie/struct.Key.html#method.from_master

Comments in the `.env.sample` file were not updated to reflect this change though.

vignesh-sankaran:
How do I change the key? And which key is this precisely?

carols10cents:
> EDIT: It appears I can't just put in any arbitrary string, going to need some help with this :(

What does it say?

vignesh-sankaran:
Ok so using crates.io locally works, but all the integration tests are failing :/. Is that the case for you too @carols10cents and @kureuil?

kureuil:
@vignesh-sankaran Both `cargo test` and `npm test` are running just fine on my end.

carols10cents:
Fixed in #941!

PR #940, opened 2017-08-07
Better error messages for empty string ENV variables

vignesh-sankaran:
Fixes up #937, which was caused by me not having the `TEST_DATABASE_URL` set i.e. it was `TEST_DATABASE_URL=`. Thanks to Arnavion on the #rust channel for pointing out I could use unwrap_or() to cover both the `None` and `""` cases at once :).

### DIFF
--- a/src/tests/all.rs
+++ b/src/tests/all.rs
@@ -138,11 +138,17 @@ fn app() -> (record::Bomb, Arc<App>, conduit_middleware::MiddlewareBuilder) {
     return (bomb, app, middleware);
 }
 
+// Return the environment variable only if it has been defined
 fn env(s: &str) -> String {
-    match env::var(s).ok() {
-        Some(s) => s,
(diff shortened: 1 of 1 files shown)


src/tests/all.rs
```diff
@@ -138,11 +138,17 @@ fn app() -> (record::Bomb, Arc<App>, conduit_middleware::MiddlewareBuilder) {
     return (bomb, app, middleware);
 }
 
+/// Returnx the environment variable only if it has been defined
```
vignesh-sankaran: Are doc comments visible in integration tests? I don't think they are.

carols10cents: AFAIK tests don't show up in documentation at all. While you're changing this to be a regular comment, could you remove the stray "x" at the end of "Returnx" too please? :)

carols10cents:
bors: r+

PR #941, opened 2017-08-07
Update .env-sample

vignesh-sankaran:
Also addresses #937 and add notes for #909 in the .env-sample. Also decided to fix up an issue which we sometimes ran into with the `S3_GLOBAL` variable, which was noted in the CONTRIBUTING.md. I thusly removed those notes from there too :).

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -6,9 +6,9 @@ export DATABASE_URL=
 # If you are running a mirror of crates.io, uncomment this line.
 # export MIRROR=1
 
-# Key to sign and encrypt cookies with. Change this to a long, random string
-# for production.
-export SESSION_KEY=badkey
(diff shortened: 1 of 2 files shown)


.env.sample
```diff
@@ -7,8 +7,8 @@ export DATABASE_URL=
 # export MIRROR=1
 
 # Key to sign and encrypt cookies with. Change this to a long, random string
-# for production.
-export SESSION_KEY=badkey
+# that is at least 32 bytes for production
```
carols10cents: Weren't you seeing dev not able to start if this wasn't at least 32 bytes all the time? So I think this should be split up differently, so that "32 bytes" isn't connected to "for production". Something like:

```
# Key to sign and encrypt cookies with. Must be at least 32 bytes. Change this 
# to a long, random string for production.
```

?

carols10cents:
bors: r+

PR #944, opened 2017-08-08
Add a comment explaining why gh_id needs a `> 0` condition here

carols10cents:
This is weird and not a great example and we should have a comment here
explaining why :)


carols10cents:
bors: r+

PR #946, opened 2017-08-10
[DO NOT MERGE] trying to find cause of beta failure on #869

carols10cents:
my current hypothesis is it's one of the deps that was added so i'm trying adding them one at a time


carols10cents:
lol nope

Issue #894, opened 2017-07-21
Documentation link can go to latest published version, not latest non-pre-release version


carols10cents:
Which can be confusing. 

Example: 
- https://crates.io/crates/data-encoding currently recommends installing v1.2.0, since that's the latest non-pre-release version
- Version 2.0.0-rc.1 has been published.
- Documentation link does go to https://ia0.github.com/data-encoding/v2.0.0-rc.1/data_encoding, which is what was specified in 2.0.0-rc.1's Cargo.toml.

Now, we could switch to storing documentation URL as a property of the version rather than the crate, but I don't think what this crate is doing is what people normally do! If you're using docs.rs, for example, the documentation for a version won't exist until after you publish and docs.rs has a chance to build it, so the documentation link would 404 for a bit after publishing if you specified the version. Updating the version in another spot in your Cargo.toml would also be annoying.

For people hosting their own docs, I don't even know if it's common to publish beneath a version path and keep old docs around, or if people are always overwriting to the same spot. This seems like a tooling issue that perhaps cargo could help with.

Or there may be other solutions to this, or other problems associated with these solutions that I haven't thought of yet. Mostly I just want to track this for discussion.

Turbo87:
I believe this has been fixed. We are now loading the docs.rs metadata for the visible crate version and build the documentation link from that if necessary.

PR #948, opened 2017-08-11
Stop showing documentation link when documentation is removed from Cargo.toml

### LINKED ISSUE #945: crates.io shows old documentation link when documentation field removed from Cargo.toml
crates.io shows old documentation link when documentation field removed in the latest version of a crate.

[take_mut](https://crates.io/crates/take_mut) crate has `https://crates.fyi/crates/take_mut/0.1.3/` documentation link [in the version 0.1.3](https://docs.rs/crate/take_mut/0.1.3/source/Cargo.toml). Documentation link is removed in [the latest version](https://docs.rs/crate/take_mut/0.2.0/source/Cargo.toml.orig). But crates.io [shows](https://crates.io/crates/take_mut) old documentation link, used in 0.1.3.

sunjay:
Fixes #945 

This is exactly the fix suggested in https://github.com/rust-lang/crates.io/issues/945#issuecomment-321115760 but applied to the `NewCrate` struct instead of the `Crate` struct since `NewCrate` is actually what gets stored in the DB.

Problems with this fix:

* The webpage for the crates that are affected by this will only get updated when they next publish
* The version specific webpages for the crate do not show a documentation link even if that crate once had a documentation key
    * This is because we do not track this metadata per version and should probably be a separate issue


sunjay:
This is the first time I'm working with this codebase (and with diesel, etc.) so please forgive me (and correct me) if I make any errors. :smile: 

I'm looked into this in more detail now and I agree that the fix proposed in @sgrif's comment should fix this issue. After some debugging and understanding the code a bit more, I moved the suggested line to the actual spot it needed to go to and it worked.

So now if you publish a crate and remove the documentation link from Cargo.toml that link will disappear off of the crate page. **The problem** is that because we don't track that information per version, if you click on any of the other versions to inspect them, you aren't able to see any documentation link.

I'm not sure if you want to fix this as part of this PR, but it should probably get an issue of its own. The migration to fix that is going to be substantial... :fearful: 

**Note** This fix only updates the metadata when you update the crate. Any existing crates that are wrong will not be fixed. So the person from #945 will need to publish a new version before the crates.io page gets updated.

Here's an example:

* `testcrate` has no `documentation` key in its Cargo.toml in version 0.1.0 - no documentation link shows up on crates.io
*  `testcrate` adds a `documentation` key in version `0.1.1` - documentation link shows up on crates.io
* `testcrate` removes the `documentation` key in version `0.1.2` - documentation link stops showing up on crates.io **but also stops showing when we view the page for version `0.1.1`**

I'll summarize all of this in the PR description above.

sunjay:
Also we might want to add a test for this. I don't know how to do that in this codebase, so if you think that's necessary please give me an idea of what to do and I'll go and do it. :smile:

carols10cents:
> This is because we do not track this metadata per version and should probably be a separate issue

See #894 for this issue.

> Also we might want to add a test for this. I don't know how to do that in this codebase, so if you think that's necessary please give me an idea of what to do and I'll go and do it. 😄

Yup, I think we need a test to make sure we don't break this again! What have you tried so far with the tests? 

I would expect a new test to be in src/tests/krate.rs, and consist of:

- Using `CrateBuilder`, create a crate in the database that has a documentation URL (see other tests in that file for examples of using `CrateBuilder`
- Publish a new version of the crate through the API, without sending a documentation URL
- Request the crate information and assert that the documentation URL isn't present

This test is a pretty good example of adding a crate to the dabatase and then publishing another version of that crate through the API, it uses some helper functions to do that: https://github.com/rust-lang/crates.io/blob/master/src/tests/krate.rs#L657-L673

The show test demonstrates how to request and assert on the crate information: https://github.com/rust-lang/crates.io/blob/master/src/tests/krate.rs#L390-L416

Please let me know if you have any other questions!

sunjay:
@carols10cents I'm having trouble working with the `RECORD` environment variable. After many errors about headers not matching up with the cached http-data file, I figured out that I need to set `RECORD=1` and delete the current cached file to get the tests to try to cache again. The problem is that whenever I try to record, the test just hangs and nothing ever gets recorded. Have you run into this before? Do you know what usually causes this?

The test hangs on a line like this:
```rust
        let mut response = ok_resp!(middle.call(&mut req));
```

More specifically, [this line](https://github.com/rust-lang/crates.io/blob/master/src/tests/record.rs?utf8=%E2%9C%93#L135) in the recorder tries to read the socket to its end and doesn't ever seem to get there. This happens in the second request of my test. For some reason the first request works, but then the second fails. Are tests limited to a single request? 

I'm not doing anything drastically different from any of the tests you showed me as examples. I'm not sure what's going wrong.

carols10cents:
Oh sorry, I forgot the publish tests make a request to S3 that needs to be recorded! Rather than setting up an S3 bucket, I usually copy another recorded http-data file and change the crate name as necessary.

src/tests/krate.rs
```diff
@@ -1447,6 +1447,226 @@ fn yank_max_version() {
...
+    {
+        let mut req = ::new_req_with_documentation(app.clone(), "docscrate", "0.2.1", Some("http://foo.rs".to_owned()));
+        ::sign_in_as(&mut req, &user);
+        let mut response = ok_resp!(middle.call(&mut req));
+        let json: GoodCrate = ::json(&mut response);
+        assert_eq!(json.krate.documentation, Some("http://foo.rs".to_owned()));
+    }
+
+    // Ensure latest version also has the same documentation
+    {
```
sunjay: Notice that I've done the same assertion twice. Once to make sure the response of the PUT request is correct, and another to make sure that the latest information returned for the crate is correct. Both are necessary to make sure this code works correctly no matter what.

sunjay:
@carols10cents Thanks! I was able to figure it out based on what you said.

I added two tests for this fix. One of them is a basic test and ensures the functionality required to fix #945. The other is a more advanced version of the same test which should be verified in #894. The behaviour being tested there is that the information for the latest crate version (the max version) doesn't get changed by publishing a lower version. This doesn't work yet, so the test is ignored. When that other issue is fixed, that test should be enabled as part of it because once we're properly tracking this information per version, we should be able to keep everything up to date properly.

sunjay:
@carols10cents I don't know what the solution to the build error is. It's some kind of linking problem or something. Could you look into it? Maybe the build just needs to be rerun?

carols10cents:
@sunjay sorry about that, can you try rebasing on master? Travis is running out of memory, we were seeing this over on #869 too and the commit I just cherry picked onto master, to change `sudo: false` to `sudo: required` in the travis config, gives us more memory in the travis env for some reason. I've reported this over at https://github.com/rust-lang/rust/issues/43789 too.

sunjay:
@carols10cents I've rebased. The build passes now! :smile:

src/tests/krate.rs
```diff
@@ -1447,6 +1447,251 @@ fn yank_max_version() {
...
+
+    // Ensure latest version no longer has documentation
+    {
+        let mut req = ::req(app.clone(), Method::Get, "/api/v1/crates/docscrate");
+        let mut response = ok_resp!(middle.call(&mut req));
+        let json: CrateResponse = ::json(&mut response);
+        assert_eq!(json.krate.documentation, None);
+    }
+
+    // 4. Add documentation again to make sure this change isn't permanent
```
carols10cents: I think this test is a bit overzealous :) I can't think of a scenario where we would accidentally change the code to allow adding and removing documentation once but not allow you to add the documentation link again after removing... so I think from line 1512 through 1549 could be removed without loss in test coverage. Could you make that change please?

src/tests/krate.rs
```diff
@@ -1447,6 +1447,251 @@ fn yank_max_version() {
...
+    {
+        let conn = app.diesel_database.get().unwrap();
+        user = ::new_user("foo").create_or_update(&conn).unwrap();
+        ::CrateBuilder::new("docscrate", user.id)
+            .version("0.2.0")
+            .expect_build(&conn);
+    }
+
+    // Verify that crates start without any documentation so the next assertion can *prove*
+    // that it was the one that added the documentation
```
carols10cents: Similarly, this block (L1568-1575) tests the same thing that the previous test does in L1464-1471. I can't think of a scenario where one of them would pass and the other would fail. Could you remove this block please?

src/tests/krate.rs
```diff
@@ -1447,6 +1447,251 @@ fn yank_max_version() {
...
+
+    // Ensure latest version still has the same documentation
+    {
+        let mut req = ::req(app.clone(), Method::Get, "/api/v1/crates/docscrate");
+        let mut response = ok_resp!(middle.call(&mut req));
+        let json: CrateResponse = ::json(&mut response);
+        assert_eq!(json.krate.documentation, Some("http://foo.rs".to_owned()));
+    }
+
+    // 3. Remove the documentation
```
carols10cents: And again, imo lines 1637-1691 are not providing any additional coverage. Please remove, thank you!

carols10cents:
I tested this out and it works great-- I just think the tests are a bit too much :) Please let me know if you'd rather that I just take care of this, or if you have any questions! Thanks!

sunjay:
@carols10cents Sorry if the intention in my test isn't clear! Thanks for your feedback!

The test is designed to give us confidence that adding or removing the documentation will work across versions, even if it is added or removed multiple times over a project. This has more to do with testing a realistic scenario than just making sure the lines are covered. You're right in that the second part of both of those tests won't change the test coverage. It's reasonable to say that over the long life of a project the documentation can be added or removed a few times. Now, it doesn't make sense to do that over and over again a crazy number of times in a test, so I've just done it twice. The test adds documentation, removes it, and then adds it and removes it again. The lines you highlighted are the second part of that scenario.

I think it's a good idea to make sure this works multiple times in case something breaks if we ever track documentation by version or make any other change that updates the documentation conditionally based on version. In particular, the test that does not contain `_basic` in its name tests some pretty important behaviour which I'll talk about more in a bit. (#894 suggests we won't do that for now, but I imagine we may need something similar to address the issues this causes.)

If you still think I should remove that second part of each of the tests, I will, but I think it's a good idea to keep it so we know this scenario works well. I'll leave that to your judgement.

> I can't think of a scenario where one of them would pass and the other would fail. 

I agree! These tests are actually one test that I had to simplify and split in order to get everything to pass for now. The test with `_basic` in its name could actually be removed entirely at some point. The other test is ignored for now with `#[ignore]`. The difference between the tests is that the non-basic version tests if publishing a lower version can accidentally overwrite the documentation of the max version. This is an important property which doesn't actually work right now, but should eventually. (maybe after #894?)

---

It might also be that I don't understand your feedback, so please let me know if that is what comes across after my explanation. :smile:

carols10cents:
> It's reasonable to say that over the long life of a project the documentation can be added or removed a few times. Now, it doesn't make sense to do that over and over again a crazy number of times in a test, so I've just done it twice.

To me, it doesn't make sense to even do it twice :) If we did every action in the test suite twice, we'd be doubling the time that every run of the test suite takes for negligible benefit! 

> I think it's a good idea to make sure this works multiple times in case something breaks if we ever track documentation by version or make any other change that updates the documentation conditionally based on version.

I would expect that the tests would be modified at the point that we change documentation stored per-version, but not before. Who knows, we may never get to making that change, and now we've added time to every run of the test suite from now until forever for no reason. Please remove the duplicated parts.



> The other test is ignored for now with #[ignore]. The difference between the tests is that the non-basic version tests if publishing a lower version can accidentally overwrite the documentation of the max version. This is an important property which doesn't actually work right now, but should eventually. (maybe after #894?)

I missed that it was ignored. If the intention is to remove the `#[ignore]` someday, could you please add a comment detailing the conditions under which the ignore can and should be removed? And remove the duplication between this test and the basic test, and the duplication within the second test. 

If you make these changes, I'd be happy to merge this PR.

sunjay:
@carols10cents Sounds good! I will update all of that hopefully in the next few days when I get the chance. :smile: 

Thanks for all your feedback on this Carol!

sunjay:
Hi @carols10cents! Sorry for the delay! This is done now. There is just a single, minimal test now. My branch was pretty out of date so I copied the changes over to a new one. That's why the history and review comments are a bit messed up. 

Should be good to go now. Let me know if you need any more changes. :smile:

sunjay:
@carols10cents Hi Carol! Could you take a look at this PR? It should be much smaller and easier to review now. I made the changes you requested a while back, so unless you see more adjustments I should make, this can probably be merged. :smile:

carols10cents:
Sorry for the delay on this! I made one little tweak but otherwise I think this is ready to go!

bors: r+

sunjay:
Thank you for taking the time! Glad it will be merged! :tada:

sgrif:
Failure is just from needing to run rustfmt

carols10cents:
bors: r+

PR #951, opened 2017-08-12
Serve locally uploaded crates from a different folder

### LINKED ISSUE #934: Frontend recreates dist/ directory even when it exists
Uploading a crate locally results in the `dist/` directory being created with the crate available. However, when I run the frontend, it deletes the directory and recreates it, thus removing the uploaded crate. This is a problem specific to macOS so far.

vignesh-sankaran:
Fix #934, serve locally uploaded crates from a different location since `dist/` gets recreated each time the front end restarts.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -402,10 +402,10 @@ sense as a name for this flag](https://github.com/rust-lang/cargo/issues/3797).
 
 Note that when you're running crates.io in development mode without the S3
 variables set (which is what we've done in these setup steps), the crate files
-will be stored in `dist/local_uploads/crates` and served from there when a
-crate is downloaded. This directory gets cleared out if you stop and restart
-the frontend. If you try to install a crate from your local crates.io and
-`cargo` can't find the crate files, that's probably why.
+will be stored in `local_uploads/crates` and served from there when a
+crate is downloaded.  If you try to install a crate from your local crates.io and
+`cargo` can't find the crate files, it is probably because this directory does not
+exist.
 
 ##### Downloading a crate from your local crates.io
 
--- a/src/dist.rs
+++ b/src/dist.rs
@@ -1,3 +1,5 @@
+//! This module implements middleware to serve the compiled emberjs
+//! frontend
 use std::error::Error;
 
 use conduit::{Request, Response, Handler};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,6 +96,7 @@ pub mod user;
 pub mod util;
 pub mod version;
 
+mod local_upload;
 mod pagination;
 
 /// Used for setting different values depending on whether the app is being run in production,
@@ -217,6 +218,7 @@ pub fn middleware(app: Arc<App>) -> MiddlewareBuilder {
     if env == Env::Development {
         // DebugMiddleware is defined below to print logs for each request.
         m.add(DebugMiddleware);
+        m.around(local_upload::Middleware::default());
     }
 
     if env != Env::Test {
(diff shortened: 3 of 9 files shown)


src/lib.rs
```diff
@@ -238,10 +238,10 @@ pub fn middleware(app: Arc<App>) -> MiddlewareBuilder {
     // Sets the current user on each request.
     m.add(user::Middleware);
 
-    // Serve the static files in the *dist* directory, which are the frontend assets.
-    // Not needed for the backend tests.
+    // Serve the static files in the *local_uploads* directory, which contains locally uploaded
+    // crates. Not needed for the backend tests.
     if env != Env::Test {
-        m.around(dist::Middleware::default());
+        m.around(local_uploads::Middleware::default());
```
kureuil: Wouldn't this break prod ?
It seems that the `dist::Middleware` is used to serve the Ember application, styles & other assets. Does it still work after this change ?

vignesh-sankaran:
Argh yes you're right it doesn't work :(. I'll find another way to get this to work, maybe by adding another bit of middleware specifically for the uploader.

vignesh-sankaran:
Bear with me for a few mins, I'm updating a previous PR and I would like to add notes about `local_uploads/` in `docs/` :)

vignesh-sankaran:
Ok, I've updated the docs too :)

docs/CONTRIBUTING.md
```diff
@@ -402,10 +402,10 @@ sense as a name for this flag](https://github.com/rust-lang/cargo/issues/3797).
 
 Note that when you're running crates.io in development mode without the S3
 variables set (which is what we've done in these setup steps), the crate files
-will be stored in `dist/local_uploads/crates` and served from there when a
-crate is downloaded. This directory gets cleared out if you stop and restart
-the frontend. If you try to install a crate from your local crates.io and
-`cargo` can't find the crate files, that's probably why.
+will be stored in `local_uploads/` and served from there when a crate is 
+downloaded. This directory gets cleared out if you stop and restart the frontend.
```
carols10cents: I thought the point of this change was to make it so that the local uploads *wouldn't* get cleared out if you stop and restart the frontend? So that should mean this sentence should go away, right?

src/dist.rs
```diff
@@ -1,3 +1,5 @@
+//! This module implements middleware to host the compiled emberjs
```
carols10cents: I would say "serve" rather than "host".

src/local_uploads.rs
```diff
@@ -0,0 +1,41 @@
+//! This module implements middleware to host crates from the
```
carols10cents: Same here-- please replace "host" with "serve" :)

carols10cents:
Nevermind what I said about it not working, it's totally working now.. I don't know what my issue was.

However, since the README rendering has been merged in, and since I merged upstream into this branch for you, could you get README rendering to work as well? The HTML for readmes is getting saved in `/local_uploads/readmes/crate/crate-version.html, but the frontend is requesting http://localhost:4200/local_uploads/readmes/crate/crate-version.html which 404s.

docs/CONTRIBUTING.md
vignesh-sankaran: Yes it should, I'll make the change :)

vignesh-sankaran:
I've made  the requested changes @carols10cents, this should now work locally with readme files too :)

carols10cents:
bors: r+

carols10cents:
bors: retry

carols10cents:
bors: r+

carols10cents:
bors: r+

PR #952, opened 2017-08-12
Maintenence badge front end

vignesh-sankaran:
Start front end work for #704 by creating a maintenance badge component. I'm not quite sure what the alias() function does in ember.js, think it takes out the attributes from the badges for a crate?

I'm going to work on all the frontend changes here, since the tests failed when I attempted to add the fixture without the badge component.

NOTE: If #996 is approved, I'll have to change this PR to work with it.

### DIFF
--- a/app/components/badge-maintenance.js
+++ b/app/components/badge-maintenance.js
@@ -1,0 +1,33 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { alias } from '@ember/object/computed';
+
+export default Component.extend({
+    tagName: 'span',
+    classNames: ['badge'],
+    escapedStatus: computed('badge', function() {
+        return this.get('badge.attributes.status').replace(/-/g, '--');
+    }),
+    none: computed('badge', function() {
+        return this.get('badge.attributes.status') === 'none'
+        || !this.get('badge.attributes.status');
+    }),
+    status: alias('badge.attributes.status'),
+    color: computed('badge', function() {
+        switch (this.get('badge.attributes.status')) {
+        case 'actively-developed':
+            return 'brightgreen';
+        case 'passively-maintained':
+            return 'yellowgreen';
+        case 'as-is':
+            return 'yellow';
+        case 'experimental':
+            return 'blue';
+        case 'looking-for-maintainer':
+            return 'orange';
+        case 'deprecated':
+            return 'red';
+        }
+    }),
+    text: 'Maintenance intention for this crate'
+});
--- a/app/templates/components/badge-maintenance.hbs
+++ b/app/templates/components/badge-maintenance.hbs
@@ -1,0 +1,7 @@
+{{#unless none}}
+    <img
+        src="https://img.shields.io/badge/maintenance-{{escapedStatus}}-{{color}}.svg"
+        alt="{{text}}"
+        title="{{text}}"
+    />
+{{/unless}}
--- a/mirage/fixtures/crates.js
+++ b/mirage/fixtures/crates.js
@@ -1,6 +1,11 @@
 /* eslint-disable quotes */
 export default [{
-    "badges": [],
+    "badges": [{
+        "badge_type": "maintenance",
+        "attributes": {
+            "value": "actively-developed"
+        }
+    }],
     "categories": [],
     "created_at": "2014-12-08T02:08:06Z",
     "description": "A high-level, Rust idiomatic wrapper around nanomsg.",
@@ -191,6 +196,12 @@ export default [{
     "updated_at": "2015-11-11T00:10:43Z",
     "versions": null
 }, {
+    "badges": [{
+        "badge_type": "maintenance",
+        "attributes": {
+            "status": "actively-developed"
+        }
+    }],
     "created_at": "2014-11-23T09:01:21Z",
     "description": "A Kinetic protocol library written in Rust",
     "documentation": "https://icorderi.github.io/kinetic-rust/doc/kinetic/",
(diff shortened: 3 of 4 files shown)


vignesh-sankaran:
This is now ready to be reviewed @carols10cents :)

carols10cents:
#996 looks good to me and bors should be merging it in soon, so please update this when you get a chance!

vignesh-sankaran:
@carols10cents I've updated the PR :)

app/components/badge-maintenance.js
```diff
@@ -0,0 +1,46 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { alias } from '@ember/object/computed';
+
+export default Component.extend({
+    tagName: 'span',
+    classNames: ['badge'],
+    escapedStatus: computed('badge', function() {
+        return this.get('badge.attributes.status').replace('-', '--');
```
carols10cents: So I tried out all of the values for the badges, and `looking-for-maintainer` didn't load. It looks like it's because this `replace` only handles the *first* hyphen:

screen shot 2017-08-26 at 1 05 53 pm

Take a look at [the replace docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) and look for "global" for how to fix this :)

app/components/badge-maintenance.js
```diff
@@ -0,0 +1,46 @@
...
+    }),
+    none: computed('badge', function() {
+        if (this.get('badge.attributes.status') === 'none'
+            || this.get('badge.attributes.status') == undefined) {
+            return true;
+        }
+        return false;
```
carols10cents: I think this function could be simplified by returning the value from the comparisons directly, wdyt? 

```
return this.get('badge.attributes.status') === 'none' || !this.get('badge.attributes.status');
```

app/components/badge-maintenance.js
```diff
@@ -0,0 +1,46 @@
...
+    }),
+    status: alias('badge.attributes.status'),
+    color: computed('badge', function() {
+        let color = 'undefined';
```
carols10cents: I think this function could be simplified by removing this `color` variable and just returning from each case, something like:

```
        switch (this.get('badge.attributes.status')) {
        case 'actively-developed':
            return 'brightgreen';
```

WDYT?

I also don't think we need to explicitly handle anything not listed here; whether we set the color to the string 'undefined' or whether we don't set a color, the badge is going to be broken anyway. Thoughts?

app/components/badge-maintenance.js
```diff
@@ -0,0 +1,46 @@
...
+            color = 'orange';
+            break;
+        case 'deprecated':
+            color = 'red';
+            break;
+        }
+        return color;
+    }),
+    text: computed('badge', function() {
+        return 'Maintenance intention for this crate';
```
carols10cents: Hm, this `text` function isn't really computing anything, right? It's not using `badge` for anything... I think this could just be `text: 'Maintenance intention for this crate',` like `tagName` and `classNames` are defined.

app/components/badge-maintenance.js
vignesh-sankaran: Yep I agree, I've gone ahead and removed handling it.

app/components/badge-maintenance.js
vignesh-sankaran: Hmm I'm not quite sure what `!this.get('badge.attributes.status')` does, I mean if it gets the status attribute then it is a not against it? Does the statement return true if the value is `undefined`, and false if there is a string returned?

vignesh-sankaran: I decided to change it since it works the same anyway :). Should I add a comment explaining this intruiging JavaScript behaviour?

carols10cents: Most people experienced with javascript have had experience with how it treats true/false/undefined behaviors, so I don't think this deserves a comment, no.

vignesh-sankaran:
Argh this has been a nightmare :(. I'm not sure why there are other changes in this PR, I rebased with the latest changes from master and now I have  this

carols10cents:
Hm, i'm not sure why github is still showing 9 commits... i'm going to try rebasing again

carols10cents:
weird, github, wat r u doin

app/components/badge-maintenance.js
```diff
@@ -0,0 +1,34 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { alias } from '@ember/object/computed';
+
+export default Component.extend({
+    tagName: 'span',
+    classNames: ['badge'],
+    escapedStatus: computed('badge', function() {
+        let regex = /-/g;
+        return this.get('badge.attributes.status').replace(regex, '--');
```
carols10cents: Hm, in my opinion, this would actually be clearer if the regex is inline rather than in a variable, so `.replace(/-/g, '--')`. That way, what we're replacing and what we're replacing it with is closer together and easier to read. Can you elaborate on why you have it in a variable here?

vignesh-sankaran: Because it didn't occur to me I could do it this way too :). The inline approach does look neater, I've changed it to be this way.

vignesh-sankaran:
Alright the commit history cleaned itself up :)

vignesh-sankaran:
Once this is merged, I believe #704 will be finished since the cargo manifest docs have been merged :)

carols10cents:
Looks great now!! Thanks!!

bors: r+

vignesh-sankaran:
Thanks for your help on this @carols10cents :)

PR #957, opened 2017-08-12
Add dependencies for top-level build script

alexcrichton:
Helps avoid unnecessarily compiling the main library!

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,6 +8,8 @@ use dotenv::dotenv;
 use std::env;
 
 fn main() {
+    println!("cargo:rerun-if-env-changed=TEST_DATABASE_URL");
+    println!("cargo:rerun-if-changed=build.rs");
     if env::var("PROFILE") == Ok("debug".into()) {
         let _ = dotenv();
         if let Ok(database_url) = env::var("TEST_DATABASE_URL") {


build.rs
```diff
@@ -8,6 +8,8 @@ use dotenv::dotenv;
 use std::env;
 
 fn main() {
+    println!("cargo:rerun-if-env-changed=TET_DATABASE_URL");
```
kureuil: Isn't this supposed to be `TEST_DATABASE_URL` ? :)

alexcrichton: Oops indeed!

carols10cents:
awesome!!! thanks!! 

bors: r+

jtgeibel:
I've just tested this locally, by: running `cargo test`, adding a migration, and running `cargo test` again.  My new migration was not run.

I've added the following to the build script, which works for my simple test, but I'm not sure it is a complete solution:

```rust
println!("cargo:rerun-if-changed=migrations/");
```

Here are the reasons I think this may not be complete:

* The build script documentation notes that the behavior of timestamps on directories can be platform specific.  However, I'm not aware of what those differences may be.
* The documentation notes that the directory is not traversed to check for additional timestamp changes.  We can manually walk the tree and add a line for each directory and file, but I'm not certain that is necessary for this use case.  If someone changes a migration during development (after running the initial migration), then they will need to run `diesel migration redo` before running `cargo test` anyway.

Thoughts?

carols10cents:
@jtgeibel I would like a pull request for that please, I think it'll be good enough for most of the cases!

PR #961, opened 2017-08-15
Fix windows build

kureuil:
Windows MSVC being a Tier1 platform it would seems logical that crates.io should be able to build on it.

TODO:

* [x] Update civet to `0.10.0` (blocked on https://github.com/wycats/rust-civet/pull/10)
* [x] Check & update install/contributing documentation
* [x] Add an appveyor configuration to make sure it doesn't break in the future (https://ci.appveyor.com/project/kureuil/crates-io)

PS: ~~I'm sorry for messing with the package-lock.json but it kept complaining about package integrity issues so I had to delete it and regenerate it in order to silence it, please forgive me. If anyone has clues about it I'll gladly try (`npm cache clean` & `npm cache verify` didn't helped me as much as I hoped)~~ Seems that updating npm to the latest version (v5.4.2) fixed the problem.


kureuil:
@carols10cents This PR is now ready for review.

* I wanted to add AppVeyor badges to the `README.md` and `Cargo.toml` files, but we might have to wait for #1050 to land due to AppVeyor's non deterministic naming convention.
* Even though it's labeled as a Windows-related PR, the civet web server was updated from 1.6 to 1.9.1. Everything seems good right now but more testing cannot harm.

If anyone has more information on what this code is supposed to do, can you please explain ?

https://github.com/rust-lang/crates.io/blob/master/src/tests/git.rs#L26-L32

From what I understand it removes the checkout and bare folders each time the function is called but creates them only once, when the program is launched.

carols10cents:
> If anyone has more information on what this code is supposed to do, can you please explain ?
> 
> https://github.com/rust-lang/crates.io/blob/master/src/tests/git.rs#L26-L32
> 
> From what I understand it removes the checkout and bare folders each time the function is called but creates them only once, when the program is launched.

Yes, that's what this function does. It's only called in the tests, [for each `app` in each test](https://github.com/rust-lang/crates.io/blob/master/src/tests/all.rs#L125). What else would you like to know about it?

kureuil:
> Yes, that's what this function does. It's only called in the tests, for each app in each test. What else would you like to know about it?

I don't understand why it would delete the folders each time the function is called (at least once for each test, so ~140 calls) but only create them once. Reading `git2-rs`'s documentation, the [`Repository::init`](https://docs.rs/git2/0.6.8/git2/struct.Repository.html#method.init) documentation states that will create any needed folder while [`Repository::init_bare`](https://docs.rs/git2/0.6.8/git2/struct.Repository.html#method.init_bare) documentation states that the folder in which the repository will be created must exist prior to invoking the function. But the old code was only deleting the folders and never creating the bare repo one ? I guess that git2-rs documentation is wrong and that it will create any needed folder for both `init` and `init_bare` given that the tests are running without any problems ?

Following this logic I don't understand why the folders were created in the first place ? I guess its safe to delete...

carols10cents:
Sorry this has taken me so long to get to :( I'm having issues building civet-sys 0.2.0 on macos, going to file an issue over there....

kureuil:
So, I've been thinking a lot about this and I'm not sure I want this PR to be merged. However, I'd like to use this opportunity to define the platform support of the crates.io project: should we support running the application on Windows/macOS/Linux or should we only support Linux as a target and tell people to use our newly introduced docker support (#1106) to develop the application locally ?

I think we need to have an answer to this question because of its consequences. If we tell people that we only officially support running the application on Linux, then maybe we shouldn't have to care about its dependency not building on macOS. However, if we tell people that we will support all three major OS, then we need to step up our CI infrastructure to make sure that we don't accidentally break the project on either Windows or macOS and that our future development will have to be compatible with all three major OS which might complicate things given that in the end, the application will probably only run on Linux when used in production.

Finally, it would seem that even though the project isn't so young anymore, there hasn't been any issue about the backend not building on Windows MSVC, although it is not documented anywhere in the setup docs. Worse, it is even assumed that the backend can currently be build on Windows MSVC, although it never was possible due to its, now outdated, civetweb dependency. Which means that it might not be that much of a big deal, and it would seem that we haven't drove off any potential contributor. Moreover, with the imminent release of the Windows Subsystem for Linux, any person with Windows 10 FCU would be able to get this project running without us doing anything except requiring contributors to use the WSL.

tl;dr I'm mostly worried that due to its current & past inactivity, the Windows support for the backend might not get enough contribution to justify its integration.

Currently blocked on https://github.com/wycats/rust-civet/pull/13.

jtgeibel:
I'm +1 for pushing people toward docker and/or WSL.  Both seem like reasonable paths towards development on Windows.

carols10cents:
For entirely selfish reasons, I don't want to break the ability to develop natively on macos; I prefer not to use docker :)

I agree that until we get more Windows contributors, the docker support for windows is probably sufficient.

I'm going to close this. Thanks for your work!

PR #969, opened 2017-08-19
crate.scss: When rendering code in README, don't show scrollbar unless necessary

joshtriplett:
The current `overflow-x: scroll` shows a scrollbar all the time, even
for code blocks that don't need them. Switch to `overflow-x: auto`, to
only show the scrollbar when needed.


carols10cents:
Thanks!!! 🌄 

bors: r+

PR #973, opened 2017-08-19
Fallback to description if the crate has no readme

### LINKED ISSUE #967: Crate info page: Show description if there's no readme
Crates now show the readme on their information page on crates.io. That's nice, allows us a bit of markup and stuff to show more information.

Consider falling back to the old behaviour -- showing the description, if there is no readme. The main reason you should do this is to make the transition smooth. As it is now, a lot of crates have lost their description in one stroke, and crates.io is a worse experience for this reason: I can see crates, but not their one line summary that begins to explain what the crate does.

I think that preserving the old behaviour as a fallback is what users and crate authors expect.

fbernier:
Fixes #967.

This is still untested as I will need to setup my environment, but I am opening this to get early feedback.

(Hello from rustconf)

### DIFF
--- a/app/templates/crate/version.hbs
+++ b/app/templates/crate/version.hbs
@@ -95,6 +95,13 @@
             <div class="crate-readme">
                 {{crate-readme rendered=crate.readme}}
             </div>
+        {{else}}
+            {{#if crate.description}}
+                <div class='about'>
+                    <h3>About This Package</h3>
+                    <p>{{ crate.description }}</p>
+                </div>
+            {{/if}}
         {{/if}}
     </div>
     <div class='authorship'>


carols10cents:
Hi!! Thank you for this!!

When I run `npm run start:live` on this branch, the description shows up for crates that don't have a readme (ex: http://localhost:4200/crates/cbindgen) and doesn't show up for crates that do have a readme (ex: http://localhost:4200/crates/libc). The descriptions are even rendered for crates that *should* have a readme but don't because something is broken somewhere that I need to figure out! (ex: http://localhost:4200/crates/diesel)

There is a bit of flicker if the readme ajax request takes a second to load, but I'm ok with that. This is still an improvement overall, thank you!!!

bors: r+

jansol:
Seems to not show anything at all if the crate has a readme but the ajax request is shot down by something like noscript.

PR #979, opened 2017-08-20
CSS: Restrict readme image width

### LINKED ISSUE #971: README images overflow into area's they're not supposed to.
See https://crates.io/crates/goblin.

Adding "max-width: 100%" solves the problem.

anthonynguyen:
Fixes #971


carols10cents:
Thank you!!! ✨ 👾 Looks much better!!

bors: r+

PR #980, opened 2017-08-20
Issue #791 - Rephrasing cargo login comments to remove reference to a specific file

autodidaddict:
Per issue, modified reference to cargo package commands to refer to `cargo login` rather than the specific file that the user shouldn't be modifying directly.


app/templates/me/index.hbs
```diff
@@ -30,9 +30,9 @@
     </div>
 
     <p>
-        If you want to use package commands from the command line, you'll need a
-        <code>~/.cargo/credentials</code>, the first step to creating this
-        is to generate a new token using the button above.
+        If you want to use package commands from the command line, you will need to
+        login with <code>cargo login</code> using one of the tokens listed below. 
+        For security reasons, you should supply a token when prompted rather than passing it on the command line.
```
vignesh-sankaran: Was there a reason why it's more secure to login when prompted rather than on the command line? The latter is what I do :)

vignesh-sankaran: Is there a reason why it's more secure to pass the token in when prompted rather than passing it in via the command line? That's what I do :)

autodidaddict: If someone were to access the history from a terminal prompt, or snoop via `ps`, they'll see the token you entered. It's not an issue if you login from your own workstation, but if you're doing it in a shared environment like a jump box, you could compromise your token.

autodidaddict:
Sorry about the package-lock.json mixup. I'm not sure why that got removed from my checkin.

app/templates/me/index.hbs
vignesh-sankaran: Could you modify the paragraph to mention that this is an issue in shared computing environments?

vignesh-sankaran:
Hmm npm install doeesn't seem to be working. Can you do an interactive git rebase to remove the `package-lock.json` file from the commit history? I'm on the #rust-infra channel on IRC if you need help with this, otherwise feel free to email me

app/templates/me/index.hbs
```diff
@@ -31,8 +31,8 @@
 
     <p>
         If you want to use package commands from the command line, you will need to
-        login with <code>cargo login</code> using one of the tokens listed below. 
-        For security reasons, you should supply a token when prompted rather than passing it on the command line. 
+        login with <code>cargo login</code> using one of the tokens listed below. <br/>
+        When working in shared environments, it is more secure to supply a token when prompted rather than passing it on the command line.
```
vignesh-sankaran: How about mentioning that it's more secure because of being able to access the key via shell history? I'm admittedly not sure how accessing the key via `ps` would work.

autodidaddict: Someone could see the running process or the remnants of it for a few seconds...and if your token was on the command line they'd see the argument and your token.

Working on the package-lock file

vignesh-sankaran: Yep, could you add that to the paragraph too?

app/templates/me/index.hbs
```diff
@@ -30,9 +30,9 @@
     </div>
 
     <p>
-        If you want to use package commands from the command line, you'll need a
-        <code>~/.cargo/credentials</code>, the first step to creating this
-        is to generate a new token using the button above.
+        If you want to use package commands from the command line, you will need to
+        login with <code>cargo login</code> using one of the tokens listed below. <br/>
+        When working in shared environments, supplying the token on the command line could expose it to other users, so it is more secure to supply a token when prompted.
```
vignesh-sankaran: Hmmm.... I'm thinking we could rephrase this to state the difference in using the `cargo login` command on workstations and in shared environments, since I feel entering the token on the 1 line is more convenient if safe to do so :)

So on line 34, you could state `<code> cargo login "token" </code>`, and on line 35, you could say "When working in shared environments, supplying the token via the aforementioned method could expose it to other users via the command line history. To avoid this, enter `cargo login` without the token, and enter the token when prompted to do so". Also, could we please keep the lines at a consistent length? Makes it easier to read in code editors and IDEs :).

How does this sound?

autodidaddict: I'll do that. I was just trying to use the wording suggested by the original issue.

app/templates/me/index.hbs
```diff
@@ -31,8 +31,10 @@
 
     <p>
         If you want to use package commands from the command line, you will need to
-        login with <code>cargo login</code> using one of the tokens listed below. <br/>
-        When working in shared environments, supplying the token on the command line could expose it to other users, so it is more secure to supply a token when prompted. 
+        login with <code>cargo login (token)</code> using one of the tokens listed below. <br/>
+        When working in shared environments, supplying the token on the command line could 
+        expose it to prying eyes via command line history. To avoid this, enter <code>cargo login</code>
```
vignesh-sankaran: Hmmm... maybe we could also add `...via command line history or other users on the network running<code>ps -a</code>`.

autodidaddict: This is just a simple hint. I'm not sure this is the place to give people a complete list of how their command line parameters can be hijacked by people who have permissions for `ps -a`. I'm concerned that too much wording/documentation here is just going to look like noise and people won't read it.

autodidaddict: I'll add some more words, though.

vignesh-sankaran: Ok fair enough, yeah noise is an issue too, think we'll leave `ps -a` out since it'd take an extra sentence to explain it. Sorry to negate your latest change :)

vignesh-sankaran:
Think we'll remove the `ps -a` notes if that's ok?

app/templates/me/index.hbs
autodidaddict: I got the ps -a in there without a new sentence. Let me know if you think it's too noisy and I'll take it out.

vignesh-sankaran: I'm worried that if we put the `ps -a` notes in there, we'll have to explain the full background context, and that starts adding noise to what is an instruction for shared environments. We're avoiding the problem anyway if we tell people to use `cargo login` without the token insertion. Thoughts?

autodidaddict:
I like the phrasing that is there now. It's enough to explain that there are potential security concerns in shared environments, but doesn't require a doctoral dissertation on linux security either :)

vignesh-sankaran:
I think it's good too. I'll test this locally then I think we're good :)

app/templates/me/index.hbs
```diff
@@ -33,8 +33,8 @@
         If you want to use package commands from the command line, you will need to
         login with <code>cargo login (token)</code> using one of the tokens listed below. <br/>
```
vignesh-sankaran: Could you make this 2 separate paragraphs instead? Would make it easier to read :)

vignesh-sankaran:
Yep think the wording will do, thanks @autodidaddict. Could you try removing package-lock.json from the history again? The diff is showing up with it. It's getting late for me here, I'll take another look at it later today :).

vignesh-sankaran:
Sorry, one other thing @autodidaddict, could you remove the changes to the .gitignore and put them into a separate PR? That way we have 1 change per PR :).

autodidaddict:
I'm in an airport about to get on a plane... not sure I'll be able to figure out how to remove a file from history before I get on the plane.

vignesh-sankaran:
No worries, there isn't a rush on things. Hope you have a safe flight home :).

autodidaddict:
So I attempted to rebase without the changes to package-lock.json and it did some horrendous things to the branch. If there's a git command you want me to run then let me know.

carols10cents:
No problem, I got this! Thank you!!

carols10cents:
I squashed everything down to one commit, removed the changes to .gitignore and package-lock.json, and pushed to master by hand (so this has been merged in even though github can't tell that). If you ever want to talk git commands, catch me on IRC! I have a bunch of tips that I use all the time to make my commits look wayyyyy better than they did when I originally made them :) If you want to look them up on your own, my favorites are `git rebase -i` and `git add -p`.

Thank you for taking care of this!!! ❤️ ❤️ ❤️ ❤️

Issue #966, opened 2017-08-19
Rendered README fails to parse links on images


daboross:
Hi! Loving the READMEs showing in crates.io!

I have one bug I'm experiencing for one of my crates, where links surrounding images fail to parse. Specifically, links like `[![Linux Build Status][travis-image]][travis-builds]` - showing the image `[travis-image]` and linking to `[travis-builds]` (both specified at the end of the README).

[github](https://github.com/daboross/fern#fern) vs [crates.io](https://crates.io/crates/fern).

![github screenshot](https://user-images.githubusercontent.com/1152146/29483729-bc424686-8462-11e7-9c0b-4afff66ae55b.png)
![crates.io screenshot](https://user-images.githubusercontent.com/1152146/29483735-db031fe6-8462-11e7-95e5-edbd10abd896.png)

For some reason the first badge shows up as an image surrounded by `[]` and then plain text, and the second one shows up as an image surrounded by `[]` followed by a link.

Source is:
```
fern
====
[![Linux Build Status][travis-image]][travis-builds]
[![Windows Build Status][appveyor-image]][appveyor-builds]

Simple, efficient logging for [Rust].
```

carols10cents:
This has been fixed by https://github.com/rust-lang/crates.io/pull/981!

PR #981, opened 2017-08-20
Switch from `pulldown-cmark` to `comrak`

### LINKED ISSUE #978: Markdown README rendering doesn't handle checkbox lists
Rendering markdown like

```md
- [ ] foo
- [x] bar
```

Should be generate a series of checkboxes, like so:

- [ ] foo
- [x] bar

However, this is what we see on crates.io:

image

Here's a crate/version exhibiting the issue: https://crates.io/crates/postgres-cursor/0.2.2

anthonynguyen:
Fixes #978 

2017-08-19-235915_693x247_scrot

This only enables the task list feature. There are [other GFM features] that aren't enabled.

[other GFM features]: https://docs.rs/comrak/0.1.7/comrak/struct.ComrakOptions.html


anthonynguyen:
Seems to fix #966 as well

2017-08-20-001118_680x672_scrot

vignesh-sankaran:
@anthonynguyen Could you run `cargo fmt` against your code and reupload please? :)

kureuil:
I'm not sure about this... I initially chose `pulldown-cmark` because it was chosen by the `rustdoc` team to render crates documentation and I wanted to re-use as much work as possible from the `rustdoc` migration. Also, https://github.com/rust-lang/rfcs/pull/1990 seems pretty important to me ?

@kivikakk as the maintainer of comrak, do you have any thoughts on this ?
This might not be as big a problem as I think.

kivikakk:
I haven't got any specific thoughts; `comrak` was created (a) to keep something as close to the reference C implementation as possible, so that changes in the spec could be adapted to quickly without effort, and (b) to have full support for GFM, since I maintain that for GitHub in our fork of the C version. (`pulldown-cmark`'s parsing strategy is very different so the changes can't be taken 1:1.) 

There was a period of inactivity, but `pulldown-cmark` seems to be getting regularly updated again recently. It's still failing 54 out of 624 examples in the latest CommonMark spec. I'll be keeping `comrak` up to date for the forseeable future as other projects come to depend on it (e.g. [docs.rs](http://docs.rs), among others), but I see a lot of value in keeping crates.io at parity with rustdoc.

carols10cents:
Since people write their readmes with GFM tables, checkmarks, etc, I think we should try to have parity with GFM for now. Maybe someday there will actually be one markdown standard.................. lol.

Using the same renderer as docs.rs sounds great to me! It's still pretty early for the new rustdoc, so when that becomes production-ready, we can re-evaluate at that point. Right now, though, READMEs aren't included in rustdoc by default. [RFC 1990](https://github.com/rust-lang/rfcs/pull/1990) should make it easier to include README text in docs, so we may have some re-evaluating to do once that's implemented as well.

Testing this out now!

carols10cents:
Yeah I think we want tables enabled: https://github.com/rust-lang/crates.io/issues/983

anthonynguyen:
Should strikethrough be enabled as well @carols10cents?

kivikakk:
Not presuming, but _if_ we want parity with how GitHub would render a README.md, the extensions to enable are _all of them_, per `github-markup`:

https://github.com/github/markup/blob/793148bb6a577cbd9fd25650d01593911ff77a6d/lib/github/markup/markdown.rb#L8

anthonynguyen:
Seems reasonable

anthonynguyen:
The tagfilter extension is at odds with the Ammonia cleaning that's done, so I'm not sure which one we'd rather have

carols10cents:
What's at odds about it?

anthonynguyen:
I guess what I mean is there's some overlap intheir functionality. `tagfilter` will mangle a [select few tags], so it acts as a blacklist, whereas Ammonia acts as a whitelist. 

[select few tags]: https://github.github.com/gfm/#disallowed-raw-html-extension-

carols10cents:
Mmmm I think I'd prefer the list of allowed tags that Ammonia uses then.

steveklabnik:
To chime in slightly, `pulldown-cmark` was chosen because comrak didn't exist at the time we made the decision; `pulldown-cmark` was the only commonmark equivalent implementation that existed.

I've spoken to `pulldown-cmark`'s maintainer about GFM equivalence, it's someting we'd all like, but it's not clear who is going to have the time and when to update it.

kivikakk:
@anthonynguyen @carols10cents Just to give some background, `tagfilter` isn't intended as a comprehensive sanitizer and you should definitely use something on top of it even if it was used. The list of tags is specifically those that cause a HTML parser to switch parsing modes ([the GFM spec touches on this](https://github.github.com/gfm/#disallowed-raw-html-extension-)).

carols10cents:
I just force pushed, sorry about that-- I took out the commit turning the tag filtering off, it sounds like to try and be as similar to github as possible, we should have the tag filtering on even though we're ALSO using ammonia to further sanitize.

Tested this out and the issues mentioned look all fixed! 🎉 

bors: r+

carols10cents:
oops looks like i broke something, fixing

carols10cents:
bors: r+

anthonynguyen:
Cool, thanks @carols10cents!

carols10cents:
Thank YOU! ❤️

Issue #958, opened 2017-08-13
Yanked crates shouldn't be displayed on user/team pages


carols10cents:
Related to #145.

Since we currently only keep track of yanked-ness on version records rather than on crate records, implementing this involves:

- [ ] Changing the [query for the crates a user owns](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/src/krate.rs#L653-L661) and the [query for the crates a team owns](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/src/krate.rs#L662-L670) to have an additional condition that there must be at least one version associated with that crate that has `yanked` equal to `false`. 
  - I expect figuring out the diesel incantation to express this condition to be the most difficult part of this issue; give it a try and post what you've tried and I can help figure it out if you get stuck!
- [ ] Make a new test similar to [this one](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/src/tests/krate.rs#L1873-L1914) that if the user or team in the test owns a crate where all versions have been yanked, that crate does NOT show up in the JSON response.

sodiumjoe:
I'd like to give this a shot.

sgrif:
Is there *ever* a case where we want the `/crates` route to include yanked crates? If not something like this should do it:

```patch
diff --git a/src/krate/search.rs b/src/krate/search.rs
index 3a8030a..dfa59d0 100644
--- a/src/krate/search.rs
+++ b/src/krate/search.rs
@@ -50,6 +50,11 @@ pub fn search(req: &mut Request) -> CargoResult<Response> {
     let recent_downloads = sql::<Nullable<BigInt>>("SUM(crate_downloads.downloads)");
 
     let mut query = crates::table
+        .filter(exists(
+            versions::table
+                .filter(versions::crate_id.eq(crates::id))
+                .filter(versions::yanked.eq(false)),
+        ))
         .left_join(
             crate_downloads::table.on(
                 crates::id
```

carols10cents:
> Is there ever a case where we want the /crates route to include yanked crates?

Yes. When you're searching to see if a name is taken, we should show yanked crates because that name can't be used.

https://github.com/rust-lang/crates.io/issues/1058
https://github.com/rust-lang/crates.io/issues/145

sgrif:
[...]

Issue #145, opened 2015-04-17
Yanked crates shouldn't show up in searches


daboross:
If all versions of a crate have been yanked, it probably shouldn't show up in searches, at least I can't see why it would want to. If a user clicks on a completely-yanked crate that shows up in search, they will be brought to a page which simply states "This crate has been yanked, but it is still available for download for other crates that may be depending on it.".

It seems like it would be better to not display those crates at all, or maybe add a checkbox with `Show deprecated crates` to toggle showing them.

For example, https://crates.io/search?q=fern brings up the now-useless `fern_macros` crate (which has been completely yanked, and doesn't compile on the beta), instead of just the `fern` crate.

carols10cents:
Hm, what if you're searching to see if a name is taken yet? You're still not allowed to use the name of a crate that has had all its versions yanked. 

The `Show deprecated crates` checkbox is a good idea, I'd want it to be checked by default though and saved in a cookie (as opposed to saving it with your user account, so that you don't have to be logged in to save your preference) and I don't think we have anything stored in cookies right now so this wouldn't be super trivial.

Other options: 

- show a badge, or possibly show 'yanked' in the badge where the latest version number is now
- fade a crate in a list when all its versions have been yanked
- sort yanked crates to the bottom, unless it's an exact name match for your query (those get ranked at the top right now)

daboross:
A "yanked" badge would definitely be better than what we have now, and could work well alongside a 'Show deprecated crates' checkbox, if that was implemented later.

I don't know or have any preference on what should be done with crates where some versions are yanked - I'm just concerned that crates that aren't at all for new projects to use are showing up as if they were available crates.

fwrs:
I think they should disappear from user pages

carols10cents:
@fwrs thanks for the suggestion! I've filed #958 for that.

PR #985, opened 2017-08-20
Hides crates on index calls unless they have at least one available v…

mauricio:
…ersion

This makes the index call only show crates that have at least one available version. Crates that have all versions yanked will not show up anymore.

Main problem for the implementation here is that `diesel` doesn't allow you to build a subquery that references a field from the parent query, so the only way to get this done is through pure SQL.

Fixes https://github.com/rust-lang/crates.io/issues/958

### DIFF
--- a/src/krate.rs
+++ b/src/krate.rs
@@ -696,6 +696,17 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
         ));
     }
 
+    if vec![params.get("user_id"), params.get("team_id")]
+        .iter()
+        .any(|s| s.is_some())
+    {
+        let not_yanked_versions = sql::<Bool>(
+            "crates.id = ANY (SELECT vs.crate_id FROM versions vs WHERE vs.crate_id = crates.id AND vs.yanked IS FALSE)",
+        );
+
+        query = query.filter(not_yanked_versions.clone())
+    }
+
     // The database query returns a tuple within a tuple , with the root
     // tuple containing 3 items.
     let data = query
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -1,4 +1,5 @@
 extern crate diesel;
+extern crate conduit;
 
 use std::collections::HashMap;
 use std::io::prelude::*;
@@ -1853,6 +1854,7 @@ fn yanked_versions_not_included_in_reverse_dependencies() {
     assert_eq!(deps.dependencies[0].crate_id, "c1");
 
     // TODO: have this test call `version.yank()` once the yank method is converted to diesel
+
     diesel::update(versions::table.filter(versions::num.eq("2.0.0")))
         .set(versions::yanked.eq(true))
         .execute(&*app.diesel_database.get().unwrap())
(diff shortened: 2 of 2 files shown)


carols10cents:
This excludes yanked crates from ALL functionality that is powered by the index api route, which unfortunately is a lot:

- search results
- crates in a category
- crates with a keyword
- the "browse all crates" list
- user pages
- team pages
- crates you're following

Issue #985 was intended to be just excluding yanked crates from user pages and team pages. [This comment of mine on #145](https://github.com/rust-lang/crates.io/issues/145#issuecomment-281492212) explains why I don't think we should just hide yanked crates everywhere all the time without a way to opt-out of doing that.

If you'd like to just fix #958, could you change this PR to only include this `filter` if there's a user or team in the query params please?

Thinking about the query, could this be done with a LEFT JOIN that has a condition that the version is not yanked, and then the version must not be null? I haven't tried this yet, so it might not work, let me know if what I'm saying makes sense at all and I can elaborate.

mauricio:
Definitely!

Gonna fix and send it again later today!

src/krate.rs
```diff
@@ -564,6 +564,10 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 
     let recent_downloads = sql::<Nullable<BigInt>>("SUM(crate_downloads.downloads)");
 
+    let not_yanked_versions = sql::<Bool>(
+        "crates.id = ANY (SELECT vs.crate_id FROM versions vs WHERE vs.crate_id = crates.id AND vs.yanked IS FALSE)",
```
sgrif: This could also be `left_join(crates_versions.on(crates::id.eq(crates_versions::crate_id).and(crates_versions::yanked.eq(false)))).filter(crates_versions::id.is_null())`

mauricio:
@carols10cents updated and included a new test for the no `user_id` or `team_id` case.

@sgrif couldn't get that to compile, couldn't find a table/class `crates_versions` anywhere. There's no link table between crates and versions as well, it's a belongs to relationship.

mauricio:
@carols10cents so I tried to add a new join [here](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L588-L594) but changing this changed the whole return type and then it doesn't compile anymore.

Couldn't find any other examples on diesel's documentation on how to do it. Do you have any tips?

carols10cents:
@mauricio could you show me what you tried, explain what you're trying to do, and the error you were getting please?

sgrif:
> Main problem for the implementation here is that diesel doesn't allow you to build a subquery that references a field from the parent query, so the only way to get this done is through pure SQL.

Perhaps we should defer this until we're pointing at Diesel master (soon)? Subselects can reference the outer table on master.

src/krate.rs
```diff
@@ -696,6 +696,17 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
         ));
     }
 
+    if vec![params.get("user_id"), params.get("team_id")]
+        .iter()
+        .any(|s| s.is_some())
+    {
+        let not_yanked_versions = sql::<Bool>(
+            "crates.id = ANY (SELECT vs.crate_id FROM versions vs WHERE vs.crate_id = crates.id AND vs.yanked IS FALSE)",
```
sgrif: Should this just be `EXISTS (SELECT 1 FROM versions WHERE versions.crate_id = crates.id AND versions.yanked IS FALSE)`?

mauricio:
@sgrif that would be much better! Much less magic to handle here.

sgrif:
This has gotten quite stale so I'm going to go ahead and close it. Feel free to open a new PR if you're interested in reviving this. Implementing this feature should no longer require dropping down to raw SQL

PR #987, opened 2017-08-20
Ensure the build script is run when migrations are added

jtgeibel:
When the timestamp on the migrations/ directory changes the build script will be run so that any pending migrations are run.

I've tested this on Linux, but management of timestamps on directories may be platform specific.


carols10cents:
This worked for me when I did the following:

- Ran `cargo test` to compile the tests
- Deleted the test database
- Ran `cargo test` without changing anything to confirm #988 (this doesn't fix that problem, unfortunately)
- Ran `diesel migration generate foo` and put a silly query in
- Ran `cargo test`, the migrations did get run and the tests passed!

bors: r+

PR #990, opened 2017-08-20
Adds a title to elaborate on how we count recent downloads.

tcr:
On http://crates.io/categories/no-std, we list "All time downloads" and "Recent downloads" without explaining how recent the recent downloads are. This diff just adds an `<abbr>` element with a title to explain:

image

The default `<abbr>` styles on my browser made this illegible (Brave on OS X) so I replaced the default abbr style with a border-bottom, but it could potentially be changed to use a newer CSS property (text-decoration-position).

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -350,3 +350,8 @@ h1 {
 a.arrow svg {
     background: #EEECDD;
 }
+
+abbr[title] {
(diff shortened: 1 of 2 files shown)


carols10cents:
Looks great!! Thank you!! 🔠 

bors: r+

PR #991, opened 2017-08-20
Adding VSCode and IntelliJ local directories to gitignore

autodidaddict:
Per a suggestion from another pull request, I'm suggesting these changes as a separate PR. Rumor has it there are people who actually use something other than `vim` to edit code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ npm-debug.log
 yarn-error.log
 testem.log
 .env
+.idea/
+.vscode/
+
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,3 @@ testem.log
 .env
 .idea/
 .vscode/
-


.gitignore
```diff
@@ -18,3 +18,6 @@ npm-debug.log
 yarn-error.log
 testem.log
 .env
+.idea/
+.vscode/
+
```
vignesh-sankaran: Could you remove the blank line at the bottom of this file?

vignesh-sankaran:
Think we're good to merge this @carols10cents, thanks @autodidaddict :)

autodidaddict:
I'm going to blame that trailing blank line on the elbows of the person who was next to me on the plane :)

carols10cents:
I think this is more appropriate for each developer to set in their global gitignore file for the editors they use: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

PR #996, opened 2017-08-21
Rename maintenance struct key from value to status

vignesh-sankaran:
Feel status is a better name for the `Maintenance` struct key than value. I also renamed the `MaintenanceValue` enum to `MaintenanceStatus`.


carols10cents:
Fine with me!

bors: r+

Issue #680, opened 2017-04-16
Impossible to remove yourself as crate owner, even though there are other owners


aochagavia:
I recently wanted to transfer ownership of a crate to someone else, so I tried the following:

```
cargo owner --add someone
cargo owner --remove aochagavia
```

However, the latter command failed with the message: `failed to remove owners from crate sudoku: api errors: cannot remove yourself as an owner`.

I think it would make sense to allow this when a crate has more than one owner. The current workaround is to ask the new owner to remove you. It is not a big deal, but it does feel unnecessary.

carols10cents:
Removing yourself, as long as there is at least one other owner, sounds reasonable.

autodidaddict:
I have this change in a local branch. I want to do more testing then I'll submit a PR.

carols10cents:
Fixed and deployed to production!! Thank you @autodidaddict!

PR #999, opened 2017-08-21
Issue #680 : Crate owners can now remove themselves as owner 

autodidaddict:
If there is at least one other owner remaining, then an owner can remove themselves from a crate.

### DIFF
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -1294,10 +1294,6 @@ fn owners_can_remove_self() {
     let r: R = ::json(&mut response);
     assert_eq!(r.users.len(), 1);
 
-    let mut response = ok_resp!(middle.call(req.with_method(Method::Get)));
-    let r: R = ::json(&mut response);
-    assert_eq!(r.users.len(), 1);
-
     let body = r#"{"users":["secondowner"]}"#;
     let mut response = ok_resp!(middle.call(req.with_method(Method::Put).with_body(
         body.as_bytes(),
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -1294,24 +1294,38 @@ fn owners_can_remove_self() {
     let r: R = ::json(&mut response);
     assert_eq!(r.users.len(), 1);
 
+    // Deleting yourself when you're the only owner isn't allowed.
+    let body = r#"{"users":["firstowner"]}"#;
+    let mut response = ok_resp!(middle.call(req.with_method(Method::Delete).with_body(
+        body.as_bytes(),
+    )));
+    let json = ::json::<::Bad>(&mut response);
+    assert!(json.errors[0].detail.contains(
+        "cannot remove the sole owner of a crate",
+    ));
+
     let body = r#"{"users":["secondowner"]}"#;
     let mut response = ok_resp!(middle.call(req.with_method(Method::Put).with_body(
         body.as_bytes(),
     )));
     assert!(::json::<O>(&mut response).ok);
 
-    // This is a self delete from owners
+    // Deleting yourself when there are other owners is allowed.
     let body = r#"{"users":["firstowner"]}"#;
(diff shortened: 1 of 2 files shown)


autodidaddict:
The build has failed twice because `cargo fmt --diff` returned differences. So I ran the two files I modified through rustfmt and the build is still showing failing diffs. I'm not sure what the right approach is here.

autodidaddict:
@carols10cents is there something you need from me to facilitate the review?

carols10cents:
Oh @vignesh-sankaran mentioned he wanted to try reviewing this one :)

src/tests/krate.rs
```diff
@@ -1225,6 +1225,61 @@ fn following() {
     assert_eq!(::json::<CrateList>(&mut response).crates.len(), 0);
 }
 
+// Ensures that so long as at least one owner remains associated with the crate,
+// a user can still remove their own login as an owner
+#[test]
+fn owners_can_remove_self() {
```
vignesh-sankaran: Thanks for adding a test here, it's appreciated :).

vignesh-sankaran:
Ok so from a functionality standpoint, this works fine, I tried adding another user, deleting the mock crate owner and then attempting to delete the sole crate owner, and it worked fine. 

From a commit perspective, could you try squashing the commits down into 1 commit i.e. the first one? There are some instructions as to how to do this [on this site](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html), let us know if you run into any problems with doing this.

Aside from that, I think we're good to merge this code in :).

carols10cents:
Hm there are still a lot of changes in the full diff, so the one squashed-down commit would still be pretty noisy. I think we can just take out the 2nd commit entirely and then squash the 3rd into the 1st. 

I'm guessing the old version of rustfmt you had reordered the `use` statements, and then the newer version didn't reorder the `use` statements, so it left them reordered :) 

I might just go ahead and take care of this actually...

autodidaddict:
I'm on the road at the moment without access to a laptop. I can take a look
in a day or two

carols10cents:
Ah yeah then, no problem, i got it :)

src/tests/krate.rs
```diff
@@ -1225,6 +1225,57 @@ fn following() {
...
+        ::CrateBuilder::new("owners_selfremove", user.id).expect_build(&conn);
+    }    
+
+    let mut response = ok_resp!(middle.call(&mut req));
+    let r: R = ::json(&mut response);
+    assert_eq!(r.users.len(), 1);
+
+    let mut response = ok_resp!(middle.call(req.with_method(Method::Get)));
+    let r: R = ::json(&mut response);
+    assert_eq!(r.users.len(), 1);
```
carols10cents: Hm, it looks like L1255-1257 are doing the exact same request and exact same assertion as L1251-1253, without any action in between? I'm going to remove the duplication while i'm in here, please let me know if there was a particular reason it was there that i'm missing :)

src/tests/krate.rs
```diff
@@ -1225,6 +1225,57 @@ fn following() {
...
+    let mut response = ok_resp!(middle.call(req.with_method(Method::Delete).with_body(
+        body.as_bytes(),
+    )));
+    assert!(::json::<O>(&mut response).ok);
+    
+    let body = r#"{"users":["secondowner"]}"#;  
+    let mut response = ok_resp!(middle.call(req.with_method(Method::Delete).with_body(
+        body.as_bytes(),
+    )));
+    ::json::<::Bad>(&mut response);
```
carols10cents: It looks like this test is checking that after you remove yourself as an owner, you should not be allowed to remove other people as owners (because you no longer have permission). I think we're missing coverage for the case where the only owner attempts to remove themselves-- I'm going to repeat the request from L1266-1269 before the second owner is added, and assert that it's not allowed.

carols10cents:
bors: r+

PR #1002, opened 2017-08-22
Improve code display

GuillaumeGomez:
A little screenshot with the new style:

screen shot 2017-08-22 at 23 13 55

As you can see, the `code` elements inside the texts are now easier to spot and I find it more attractive when reading.

### DIFF
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -313,6 +313,13 @@
         pre {
             overflow-x: auto;
         }
+
+        p {
+            code {
+                background-color: #fff;
+                padding: 0 2px;
+            }
+        }
     }
     .last-update {
         color: $main-color-light;


app/styles/crate.scss
```diff
@@ -313,6 +313,13 @@
         pre {
             overflow-x: auto;
         }
+
+        p {
+            code {
+                background-color: #fff;
```
sgrif: What's the purpose of this? It's not being rendered, and if it were the text would be unreadable.

GuillaumeGomez: What? It's rendered perfectly for me and the text is readable as my screenshot shows it. The point is to make text code blocks stand out a bit more.

sgrif: The background color of the code block is not white...

sgrif: Oh I see, this is for inline code blocks. I misunderstood.

sgrif: Could you edit your commit message to be more clear on what this change is affecting?

GuillaumeGomez: Sure!

GuillaumeGomez:
@sgrif: Updated.

PR #1004, opened 2017-08-23
Browserstack

carols10cents:
We can get access to browserstack for free as an open source project if we put their logo and URL in the readme!


carols10cents:
bors: r+

PR #1009, opened 2017-08-24
Update to Diesel 0.16.0

sgrif:
The output of `diesel print-schema` has changed, so I've re-run it. I've
also passed the new `--with-docs` flag, since we like documenting all of
the things. :)


carols10cents:
bors: r+

PR #1010, opened 2017-08-25
Readme rendering - only rerender older than a date, only rerender for a crate

carols10cents:
Only rerender in a `Box<T>`, only rerender in Firefox :)

These changes add some arguments to the render-readmes script to make it easier to restart a failed run only with those readmes that have not been rendered since a particular date, and to rerender one particular crate's readmes.

By default, the render-readmes script still rerenders all readmes of all versions. But if the script is killed partway through, it doesn't need to be completely restarted-- you can grab the time that the script printed out that it started, then rerun with `--older-than thattime`, and it'll only try and render the ones it hasn't gotten to yet!

Another thing I've wanted is the ability to render one particular crate's readmes, so that I don't have to scroll through all the logs to see why a particular crate failed. So I added the `--crate` argument too.

I'd like to get this into the same deploy as https://github.com/rust-lang/crates.io/pull/981 since that requires rerendering everything again.

@kureuil, could you look over this and see if it looks like I might be breaking anything?


src/bin/render-readmes.rs
```diff
@@ -2,87 +2,146 @@
...
     } else {
-        versions_count / page_size + 1
+        total_pages + 1
     };
-    for current_page in 0..pages {
-        let versions: Vec<EncodableVersion> = versions::table
+
+    let mut page_num = 0;
+
+    for version_ids_chunk in &version_ids.into_iter().chunks(page_size) {
```
kureuil: I didn't know about itertools, awesome crate!

Couldn't we use [enumerate](https://docs.rs/itertools/0.6.1/itertools/fn.enumerate.html) here to remove the `mut page_num` ?

src/bin/render-readmes.rs
```diff
@@ -2,87 +2,146 @@
...
                 }
                 let readme = readme.unwrap();
-                let readme_path = format!(
-                    "readmes/{}/{}-{}.html",
-                    version.krate,
-                    version.krate,
-                    version.num
-                );
+                let readme_path =
+                    format!("readmes/{}/{}-{}.html", krate_name, krate_name, version.num);
```
kureuil: Should we change the formatting string to `"readmes/{0}/{0}-{}.html"` to avoid the `krate_name` duplication ?

src/bin/render-readmes.rs
```diff
@@ -96,31 +155,34 @@ fn main() {
...
+                        krate_name,
                         version.num
                     ));
             });
             tasks.push(handle);
         }
         for handle in tasks {
             if let Err(err) = handle.join() {
-                println!("Thead panicked: {:?}", err);
+                println!("Thread panicked: {:?}", err);
```
kureuil: Good catch!

src/bin/render-readmes.rs
carols10cents: oooh yep, you're absolutely right!

src/bin/render-readmes.rs
carols10cents: sure!

carols10cents:
bors: r+

Issue #986, opened 2017-08-20
CSRF: logout


ishitatsuyuki:
```
<img src="https://crates.io/logout">
```

This won't work on GitHub of course due to it's secure image proxy, but anyway we don't want a malicious page constantly logging you out of the site!

ishitatsuyuki:
Hmm, seems I'm being constantly logged out on refresh for other reasons. Feel free to close if this isn't a CSRF.

debris:
I experienced the same issue

jtgeibel:
@ishitatsuyuki and @debris the issue with being logged out on a refresh should now be resolved by #910.  If you're still being logged out when refreshing a page please try clearing your browser cache and see if this is still an issue.  You're browser could still caching a version from before we added the necessary header.  Also keep in mind that we are using session cookies only right now, so you currently have to log back in any time you restart your browser (see #918).

However, I agree that CSRF protections should also be added to this route.

PR #1013, opened 2017-08-26
Switch to `DELETE` for /logout route to prevent CSRF

jtgeibel:
Other sites are allowed to make `GET` requests so switching to `DELETE`
to match the destructive nature of the route.

If the user navigates to `GET /logout` we now fall back to the default
behavior which is to serve up the ember app.  Ember then sends a
`DELETE` request.  Fortunately we already enforce sameorigin for
iframes via the `X-FRAME-OPTIONS` header.

Fixes: #986

### DIFF
--- a/app/routes/logout.js
+++ b/app/routes/logout.js
@@ -7,7 +7,7 @@ export default Route.extend({
     ajax: service(),
 
     activate() {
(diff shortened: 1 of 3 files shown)


carols10cents:
Sounds good to me!

bors: r+

PR #1015, opened 2017-08-26
Need to rerender readmes that haven't been recorded

carols10cents:
To pick up the readmes that have been rendered before the change to
record the time a readme is rendered went out.


carols10cents:
bors: r+

PR #1016, opened 2017-08-26
Give tables in readmes a bit of padding and a border

carols10cents:
Before:

before

After:

after


carols10cents:
I got one thumbs up so I'm merging this. It's just CSS, what could go wrong? 👼 

bors: r+

Issue #1012, opened 2017-08-26
Provide an endpoint to view the owner of a token


mike-engel:
Howdy Crates.io team!

As a dev/publisher on crates.io, I usually want to know who I'm logged in as via `cargo` but have no way to know. npm has something similar with `npm whoami` which helps me make sure that I'm both logged in and logged in as who I think I am before I publish a crate.

I just switched to a new computer, so I couldn't remember if I had logged into cargo yet, which is where this came about. Normally I would make this an issue on the cargo repo (and maybe I will depending on how this goes!) but right now I don't see a way to view who owns a token.

Cargo only stores the token and no user info, so it would be nice to have an endpoint that either returns the name or username of the owner of a token. This probably comes with security concerns, but I think everything on crates.io is public (no private accounts) so maybe it wouldn't be too bad? Also if you already have the token, you can already do a lot of stuff on behalf of the owner. The biggest problem would be guessing/brute forcing the token itself, which again you can already try.

Hopefully this all makes sense and is thorough enough. I'd be happy to provide more use cases or explain things further. If it's an idea y'all like, I'd also potentially be willing to take a stab at a PR (still a rust beginner!). Thanks!

carols10cents:
So I *think* this is already technically possible, but I'd like to make one change before adding it to a cargo command. I just tried this:

```
curl 'https://crates.io/me' -H 'Accept: application/json' -H 'Authorization: mytoken'
```

and got:

```
{"user":{"id":396,"login":"carols10cents","email":"carol.nichols@gmail.com","name":"Carol (Nichols || Goulding)","avatar":"https://avatars2.githubusercontent.com/u/193874?v=4","url":"https://github.com/carols10cents"}}
```

so cargo could parse out the user login value and display that!

The change I'd like to make is to move the `/me` backend routes under `/api/v1` like the other backend routes ([see some discussion here](https://github.com/rust-lang/crates.io/pull/910#issuecomment-318250109), so that they're consistent before cargo is using them and we won't be able to change them forever. 

I *think* it should be straightforward but it has the potential to be hard to track down places that might need to be updated-- so normally I'd offer this to you to fix, but I think it'd be faster for me to do it :) I'm kind of excited this is already mostly done, so I'm going to give this a try real quick :)

carols10cents:
That wasn't too bad-- I've got #1017 open :)

mike-engel:
[...]

PR #1017, opened 2017-08-27
Move backend routes under api/v1

carols10cents:
For consistency, to avoid the clash with the frontend /me route that
caused problems with staying logged in, to make it clear that these
routes are part of the public HTTP API provided by crates.io, and to
let us evolve these routes in the future.

This would make me feel comfortable about adding a cargo command that uses the info from the backend me route (see #1012).

We could do this more conservatively by leaving the old routes here and only adding the new routes... but we haven't made any guarantees about the stability of these routes, so...

See discussion: https://github.com/rust-lang/crates.io/pull/910#issuecomment-318250109

@jtgeibel wdyt?

### DIFF
--- a/app/adapters/api-token.js
+++ b/app/adapters/api-token.js
@@ -1,7 +1,7 @@
 import DS from 'ember-data';
 
 export default DS.RESTAdapter.extend({
-    namespace: 'me',
+    namespace: 'api/v1/me',
     pathForType() {
         return 'tokens';
     }
--- a/app/controllers/dashboard.js
+++ b/app/controllers/dashboard.js
@@ -45,7 +45,7 @@ export default Controller.extend({
             this.set('loadingMore', true);
             let page = (this.get('myFeed').length / 10) + 1;
 
-            this.get('ajax').request(`/me/updates?page=${page}`).then((data) => {
+            this.get('ajax').request(`/api/v1/me/updates?page=${page}`).then((data) => {
                 let versions = data.versions.map(version =>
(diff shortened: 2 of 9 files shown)


jtgeibel:
Looks great to me @carols10cents!  Works fine for me in a staging environment.  This should also fix things for people who have HTML cached under the old route.

carols10cents:
thank you @jtgeibel!

bors: r+

carols10cents:
Hi! I had no idea about cargobot. In what form would you like to be notified, and for what kinds of API changes?

jtgeibel:
Fortunately, the only routes that remain at the top level are `/authorize_url`, `/authorize`, and `/logout`.  I don't see any need to move these and I don't expect that anything else would depend on these routes.

However, it would be really nice if we had a list of known projects using the API so that we could reach out when we have such changes and even to just get a better understanding of how the broader ecosystem is using crates.io.

lfairy:
Cargobot only depends on `/api/v1/summary` at the moment, so I just need to be notified about that.

I think a GitHub mention would be okay, since I check that often. If/when we set up a general way to notify consumers about API changes (as suggested by @jtgeibel) I'll be happy to follow that instead.

carols10cents:
Ok! I [started a wiki page](https://github.com/rust-lang/crates.io/wiki/Projects-that-use-crates.io's-API) for things I know use crates.io's api, and I'll try and remember to cc you if we change things about summary!

PR #1022, opened 2017-08-27
Remove explicit use of PgConnection as this is now in the prelude

jtgeibel:
As of diesel 0.16.0 `diesel::pg::PgConnection` is now also exported
from `diesel::prelude`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,7 +2,6 @@ extern crate diesel;
 extern crate dotenv;
 
 use diesel::migrations::run_pending_migrations;
-use diesel::pg::PgConnection;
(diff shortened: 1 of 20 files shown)


jtgeibel:
@carols10cents is it possible to restart https://travis-ci.org/rust-lang/crates.io/jobs/268946820 or do I need to do a force-push and rebuild everything?  It looks like an intermittent issue with npm.

carols10cents:
ugh i hit restart. i've seen that a few times now, hmmm.....

carols10cents:
LGTM! Thank you!!

bors: r+

PR #1023, opened 2017-08-28
Add query to display recent_downloads on krates::show

sgrif:
Additionally, we've updated the join for this data on krate::index to
use the new public API Diesel added for this in the most recent version,
and made sure that any other endpoints which *don't* include this data
are returning NULL instead of 0

/cc @natboehm


carols10cents:
LGTM! Thank you @sgrif and @natboehm!!

bors: r+

PR #1032, opened 2017-09-01
Don't list crates that teams/users used to own

carols10cents:
I noticed that the cargo-canoe crate was still listed under my crates on /dashboard, even though alex changed all the ownership records except one to `deleted=true`. Turns out we didn't have that condition on here!

The new tests with the ownership records changed to deleted failed before these changes and pass after.


carols10cents:
people are starting to notice this, so i'm merging it.

bors: r+

sgrif:
Sorry I just saw this. I was on the road the entire time this was open

Issue #1030, opened 2017-08-31
Move Crates.io package policies to this repo?


behnam:
While we're moving cargo docs to <https://doc.rust-lang.org/cargo>, there's one page that's very crates.io-specific, and not build-system-specific: the [Crates.io package policies](http://doc.crates.io/policies.html).

Should we move that content to this repo, like <http://crates.io/policies>?

carols10cents:
sounds good!

PR #1033, opened 2017-09-01
Add /policies page

behnam:
Moving the "Crates.io Package Policies" page from
<http://doc.crates.io/policies.html>, which is hosted in `cargo` repo,
here, as this is a website-related content and not related to the
build-system.

First step for <https://github.com/rust-lang/crates.io/issues/1030>

Next step will be to redirect the old URL to this new page.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -44,6 +44,7 @@ Router.map(function() {
     this.route('category_slugs');
     this.route('catchAll', { path: '*path' });
     this.route('team', { path: '/teams/:team_id' });
+    this.route('policies');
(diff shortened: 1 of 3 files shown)


behnam:
Side-by-side comparison of the old and new pages:

screenshot from 2017-09-01 11-43-26

NOTE: We're keeping fragment slugs intact, making it easier to update hard-coded URLs in code-bases.

carols10cents:
LGTM! Thank you! 📦 📦 📦 

bors: r+

Issue #1031, opened 2017-08-31
Branding updates for crates.io


behnam:
Since crates.io is getting some new design soon, and we're moving cargo docs to doc.rust-lang.org, I think it's a good time to update some branding matters of crates.io as well. Two things come to my mind right now:

* [x] On the header of <https://crates.io/>, it says "CARGO: packages for Rust". I think it can be more like "Crates.io: Rust Package Registry".

* [x] On the main page, it says "The Rust community’s crate host". I think we better change the wording to "registry" instead of "host".

SeanPrashad:
Hi there! I'm a student at Seneca College whose enrolled in an Open Source course taught by my professor and Mozillian, David Humphrey. I would love to tackle this bug if it's still available and unassigned.

behnam:
Thanks for offering to help, @SeanPrashad! Let's see if we have more to do here, and if so, would be great if you can submit a PR.

IMHO, I feel like we're done here already. I mean, "The Rust community's crate host" title is in align with the "Rust Package Registry" that's in the header, and I think we better not to repeat the word "registry" too much.

What do you think, @carols10cents? Do we want more updates here, or should we just close the task?

carols10cents:
@behnam Oh, I thought the idea was to switch to "registry" everywhere? I think to be consistent we should change "The Rust community’s crate host" to "The Rust community’s crate registry" :) It'd be awesome if you could take care of that @SeanPrashad!

SeanPrashad:
@carols10cents @behnam I would be ecstatic to tackle it in the next few days when I get a break from classwork! I will notify you two once completed!

SeanPrashad:
@carols10cents @behnam I just submitted PR #1098 which includes this small modification! Please let me know if there is anything else that I can help with!!

carols10cents:
This is all done now!

PR #1034, opened 2017-09-01
[header] Replace CARGO with crates.io

behnam:
As Cargo documentation is being moved to <doc.rust-lang.org>, and to
emphasis the difference between `crates.io` and `cargo`, we better not
title the website as *CARGO* anymore.

This is a quick fix for this, until we redirect `docs.crates.io`
completely and fix the menus and links here.

See <https://github.com/rust-lang/crates.io/issues/1031>

### DIFF
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -12,8 +12,8 @@
     {{/link-to}}
     {{#link-to "index"}}
       <h1>
-        CARGO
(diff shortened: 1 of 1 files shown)


behnam:
Preview:
screenshot from 2017-09-01 12-01-07

carols10cents:
Thanks! 📦 📦 📦 

bors: r+

Issue #1029, opened 2017-08-31
Merge https://crates.io/install into http://doc.crates.io/#installing


behnam:
Since we are moving all cargo docs into the cargo repo, I think we can merge <https://crates.io/install> page into that repo as well, and redirect that URL to <http://doc.crates.io/#installing>.

What do you think?

PR #1035, opened 2017-09-01
Redirect /install to doc.crates.io

behnam:
The content of the `install` page here is duplicate of `cargo`'s
*Installation* page/section. Since it's not *registry*-specific feature,
we better just redirect to the original source.

The content of the page is already merged into <http://doc.crates.io/>
and is live.

See <https://github.com/rust-lang/crates.io/issues/1029>

### DIFF
--- a/app/routes/install.js
+++ b/app/routes/install.js
@@ -1,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+    redirect() {
+        window.location = 'http://doc.crates.io/';
+    },
+});
--- a/app/templates/install.hbs
+++ b/app/templates/install.hbs
@@ -4,36 +4,5 @@
 </div>
 
 <p>
-Rust comes with Cargo by default, so the easiest way to get Cargo is to
-<a href='http://www.rust-lang.org/install.html'>install Rust</a> via
-the official binary installers.
+    Redirecting you to <code>http://doc.crates.io/</code>&#8230;
 </p>
-
-<h2>Only Cargo</h2>
-
-<p>
-    To install just Cargo, the current recommended installation method is
-    through the official nightly builds. Note that Cargo will also require that
-    <a href='http://www.rust-lang.org/'>Rust</a> is already installed on the
-    system.
-</p>
-
-<div id='install-downloads'>
-    <table>
-        <tr>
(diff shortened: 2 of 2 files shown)


steveklabnik:
We're planning on deprecating doc.crates.io in the near-ish future; I think it's probably a better idea to just move the content over here, and redirect the other way.

behnam:
@steveklabnik I was thinking of updating the redirect later to the "Installation" chapter of the book, whenever the URL is stable. AFAIU, we want to keep that chapter in the Cargo Manual mdbook anyway, and it's easier to keep them in sync if both are in the `cargo` repo.

What do you think?

behnam:
Uh, and the reason it's redirecting to the homepage of `http://doc.crates.io/` is that we don't have a page for "Installation" in the old docs. When we get that specific page, the URL will be refined.

steveklabnik:
Ugh you know what, I also got confused with the *policies* vs *install* bit here, my bad. Sigh.

> What do you think?

Yeah this seems good. Sorry for the noise!

app/templates/install.hbs
```diff
@@ -4,36 +4,5 @@
 </div>
 
 <p>
-Rust comes with Cargo by default, so the easiest way to get Cargo is to
-<a href='http://www.rust-lang.org/install.html'>install Rust</a> via
-the official binary installers.
+    Redirecting you to <code>https://crates.io/</code>&#8230;
```
carols10cents: Shouldn't this say "Redirecting you to `http://doc.crates/io`"? Since that's where the redirect is actually going...?

app/routes/install.js
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+    redirect() {
+        window.location = 'http://doc.crates.io/';
```
carols10cents: Should this be `https`? Or protocol-relative, so just `//doc.crates.io`?

behnam: Unfortunately, `https` doesn't work on CNAME'd gh-pages, yet. Making this HTTPS or relative would make a false alarm like this:

image

app/templates/install.hbs
behnam: Right. Updated.

carols10cents:
Much better, thanks!!

bors: r+

PR #1044, opened 2017-09-06
[styles] Make header subtitle fit in one line

behnam:
After shipping the update to header title/subtitle and testing it with
logged-in user, the subtitle was broken into two lines, breaking the
layout. Making the subtitle a bit smaller fixes the problem.

See <https://github.com/rust-lang/crates.io/issues/1031>

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -330,7 +330,7 @@ ul.dropdown {
 h1 {
   span.subtitle {
     display: block;
-    font-size: small;
+    font-size: x-small;
   }
 }
 


vignesh-sankaran:
@behnam This repo hasn't got automatic review assigning set up, so the `r?` command won't work here :)

behnam:
> This repo hasn't got automatic review assigning set up, so the r? command won't work here :)

Thanks for the reminder, @vignesh-sankaran. :) My comment is intended to ping the person for a review, but not the auto-assign bot.

carols10cents:
So this doesn't fix it for all names, and it's kind of unreadable at this point:

screen shot 2017-09-09 at 1 19 23 pm

In fact, long names probably had this problem before changing the header text.

Do you have any ideas on a way to make the wrapping make more sense instead, perhaps?

behnam:
This PR fixes the specific problem of "Rust Package Registry" breaking into two lines (in some cases, like when logged in and name is long enough), because of it being longer that the "crates.io" text.

To fix the top-right corner text to be fixed, I think we better just wait for the overall site redesign and consider these cases there. IMHO, the current design doesn't have much room for a nice fix, and spending more time would be duplicate work of the site redesign.

With that, I suggest to just land this to fix the first problem, and take the second to the tracker task for redesign. What do you think?

carols10cents:
Sure, fine. There's no timeline for the redesign though.

bors: r+

PR #1045, opened 2017-09-06
Sending an email to a user to confirm their email address

natboehm:
This PR addresses issue #808, allow editing your own email address. Following the discussion on the issue, this implements the second of three parts, the ability to send a confirmation email to a user for verification. For a full explanation of the issue, see [this comment](https://github.com/rust-lang/crates.io/issues/808#issuecomment-319463368) I left which should fully explain the changes made, as well as parts 1 and 3. 

In short, I added two tables to the database: one for the user's email, the other for a confirmation token associated with the email. Once the user clicks the link sent, the associated token is deleted from the token table and `email_verified` is set to `true` in the email table. It is indicated to the user if they have not verified their email on the account settings page, a message and email resend button being displayed. Sending emails requires Mailgun to be added to the crates.io Heroku account. 

This has applications for issue #924 in that we should now be able to send an email to a user, if they have added their email. I know there was a lot of discussion on that issue regarding the ability to send users an email to be added as an owner.


src/user/mod.rs
```diff
@@ -473,9 +589,209 @@ pub fn update_user(req: &mut Request) -> CargoResult<Response> {
...
+            format!(
+                "Hello {}! Welcome to Crates.io. Please click the
+link below to verify your email address. Thank you!\n
+https://crates.io/confirm/{}",
+                user_name,
+                token
+            ).as_str(),
+        )
+        .build()
+        .expect("Failed to build confirm email message");
```
carols10cents: So if this `expect` gets an `Err`, this would crash the server, which normally would make me go 😱 🙀 😱 🙀 😱 🙀  but looking at [lettre's code](https://docs.rs/lettre/0.6.2/src/lettre/email/mod.rs.html#555), the only way building an email would fail is if:

- no `from` address was set, and we're definitely calling `from` with something we control
- the `to` address is empty string, and we reject empty emails before calling `send_user_confirm_email` in `update_user` 

I'm trying to figure out if it's possible that we get `Some("")` from github and therefore would insert an empty string into the email table and then get an empty email here? I'm going to try some testing and see what happens.

This is also starting to feel a bit spooky-action-at-a-distance since the code that is ensuring the `to` email isn't empty is far away from here. This could be an issue for any email we send.... Hmmmm.... I wonder if, when the generic email functionality is extracted into a function, if we should pass the generic function strings for `to`, `from`, `subject`, and `body`, and then the generic function first checks to make sure `to` isn't an empty string and returns an `Err` if so, and then we can call `expect` and point to the line checking that should be nearby as proof that the `expect` won't fail?

src/user/mod.rs
```diff
@@ -473,9 +589,209 @@ pub fn update_user(req: &mut Request) -> CargoResult<Response> {
...
+
+    #[derive(Serialize)]
+    struct R {
+        ok: bool,
+    }
+    Ok(req.json(&R { ok: true }))
+}
+
+fn generate_token() -> String {
+    let token: String = thread_rng().gen_ascii_chars().take(26).collect();
```
carols10cents: cool!!

carols10cents:
bors: r+

carols10cents:
bors: r+

carols10cents:
bors: r+

PR #1047, opened 2017-09-09
Fix docs.rs metadata for docs of private types

onur:
Fixes: #911

Tested with [this crate](https://docs.rs/docsrs-test/0.0.4/docsrs_test/).


carols10cents:
Thanks!! 🌱 

bors: r+

Issue #693, opened 2017-04-29
Wrong AppVeyor badge


dtolnay:
https://crates.io/crates/serde is showing a badge from:

    https://ci.appveyor.com/api/projects/status/github/serde-rs/serde?svg=true&branch=master

The link is correct and shows that the build is passing, but the badge says "unknown".

AppVeyor tells me to use the following badge:

    https://ci.appveyor.com/api/projects/status/avsuce1irt8h63u2/branch/master?svg=true

Unclear how this is generated but it would be nice if I could have crates.io show the correct badge.

Implementation steps added by @carols10cents:

- [ ] Add an `id` field with type `Option<String>` to [the Appveyor badge enum variant](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/src/badge.rs#L16-L20)
- [ ] Add `id: alias('badge.attributes.id')` to the [appveyor badge component](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/app/components/badge-appveyor.js)
- [ ] Add an `image-url` attribute, computed from the id to the appveyor badge component that checks to see if `id` is defined; if it is, return `https://ci.appveyor.com/api/projects/status/{{ id }}/branch/{{ branch }}?svg=true`. If not, return `https://ci.appveyor.com/api/projects/status/{{ service }}/{{ repository }}?svg=true&branch={{ branch }}` (that's what [the appveyor badge template currently uses for the image src](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/app/templates/components/badge-appveyor.hbs#L3))
  - I've used handlebars interpolation here but you should use js interpolation, I'm being lazy sorry sorry!
- [ ] Change [the appveyor badge template](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/app/templates/components/badge-appveyor.hbs#L3) to use the `image-url` attribute for the img src instead
- [ ] Publish a crate to your local instance that has the URLs dtolnay has listed for serde and manually verify the badge is displayed correctly (we used to have frontend tests for badge tests but they got deleted?)
- [ ] Send a pull request to cargo to [update the badge documentation](https://github.com/rust-lang/cargo/blame/master/src/doc/manifest.md#L187-L190) to indicate you can specify the appveyor project id if you want to use that instead

carols10cents:
[...]

PR #1050, opened 2017-09-13
Fix Wrong AppVeyor badge

mattgathu:
this is part of fixing issue #693

- [x] Add an id field with type Option<String> to the Appveyor badge enum variant

- [x]  Add id: alias('badge.attributes.id') to the appveyor badge component

- [x]  Add an image-url attribute, computed from the id to the appveyor badge component that checks to see if id is defined. (I've used `imageUrl` as attribute)

- [x]  Change the appveyor badge template to use the image-url attribute for the img src instead

- [x]  Publish a crate to your local instance that has the URLs dtolnay has listed for serde and manually verify the badge is displayed correctly.

- [x]  Send a pull request to cargo to update the badge documentation to indicate you can specify the appveyor project id if you want to use that instead. (_https://github.com/rust-lang/cargo/pull/4489_)

### DIFF
--- a/app/components/badge-appveyor.js
+++ b/app/components/badge-appveyor.js
@@ -12,7 +12,7 @@ export default Component.extend({
     imageUrl: computed('badge.attributes.id', function() {
         let id = this.get('badge.attributes.id');
         let branch = this.get('branch');
-        if (id === undefined || id === null) {
+        if (id !== undefined || id !== null) {
             return `https://ci.appveyor.com/api/projects/status/${id}/branch/${branch}?svg=true`;
         } else {
             let service = this.get('service');
--- a/app/components/badge-appveyor.js
+++ b/app/components/badge-appveyor.js
@@ -6,7 +6,21 @@ export default Component.extend({
     tagName: 'span',
     classNames: ['badge'],
 
+    id: alias('badge.attributes.id'),
     repository: alias('badge.attributes.repository'),
+
+    imageUrl: computed('badge.attributes.id', function() {
+        let id = this.get('badge.attributes.id');
+        let branch = this.get('branch');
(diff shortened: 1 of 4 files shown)


mattgathu:
Hi @carols10cents @dtolnay. Kindly review this PR for issue #693

carols10cents:
Hi @mattgathu! Sorry for the delay on reviewing this-- I'm going to be pretty busy today and this weekend prepping for the start of the [impl period](https://internals.rust-lang.org/t/announcing-the-impl-period-sep-18-dec-17/5676), but I will try to get to this next week :) Thank you for your PR!!

app/components/badge-appveyor.js
```diff
@@ -6,8 +6,22 @@ export default Component.extend({
     tagName: 'span',
     classNames: ['badge'],
 
+    id: alias('badge.attributes.id'),
     repository: alias('badge.attributes.repository'),
 
+    imageUrl: computed('badge.attributes.id', function() {
+        let id = this.get('badge.attributes.id');
+        let branch = this.get('branch');
+        if (id === undefined || id === null) {
```
carols10cents: I think this condition is backwards? This says if id IS undefined or id IS null, then try to use id? This should be `(id !== undefined && id !== null)`, yeah? I tried making a badge using only `repository` and `id` and the `id` isn't appearing in the image url :(

jtgeibel:
@mattgathu I think the operator also needs to be switched from `||` to `&&`.

mattgathu:
@carols10cents  @jtgeibel thank you for the bug catch! I've fixed it.

carols10cents:
Thank you!!! Works great now!!! 

bors: r+

PR #1053, opened 2017-09-14
Higher resolution cargo icon

dtolnay:
This icon is cropped from [`public/assets/Cargo-Logo-Small.png`](https://github.com/rust-lang/crates.io/blob/master/public/assets/Cargo-Logo-Small.png).


carols10cents:
bors: r+

PR #1054, opened 2017-09-14
Verify tarballs uploaded to crates.io contain at most one directory

alexcrichton:
Otherwise they could accidentally overwrite other packages!


src/uploaders.rs
```diff
@@ -224,3 +228,22 @@ impl Drop for Bomb {
...
+    let prefix = format!("{}-{}", krate.name, vers);
+    for entry in archive.entries()? {
+        let entry = entry?;
+
+        // Verify that all entries actually start with `$name-$vers/`.
+        // Historically Cargo didn't verify this on extraction so you could
+        // upload a tarball that contains both `foo-0.1.0/` source code as well
+        // as `bar-0.1.0/` source code, and this could overwrite other crates in
+        // the registry!
+        if !entry.path()?.starts_with(&prefix) {
```
erickt: Do you need a trailing '/', or is this looking at a whole directory entry?

erickt: Can you add checks if any of the entries are: '..' or non-file/directories?

alexcrichton: The trailing `/` isn't necessary due to the way `Path` works, entries like `..` are tested in `tar` itself, and non-file/directory entries are a feature, not a bug, but have extra handling in `tar`.

PR #1066, opened 2017-09-20
Accept owner invites - help with Ember

natboehm:
Currently working on deployable chunk 2 of issue #924, looking for some Ember help. When submitting the `PUT` request to the backend we're using ember-ajax to send the `crate_owner_invitation` struct back to be used in the function `accept_invite`, written as follows in `app/controllers/me/pending-invites.js`:

```
this.get('ajax').put('/api/v1/me/accept_owner_invite', {
    contentType: 'application/json; charset=utf-8',
    data: JSON.stringify({
        crate_owner_invitation: {
            invited_by_username: invite.get('invited_by_username'),
            crate_name: invite.get('crate_name'),
            crate_id: invite.get('crate_id'),
            created_at: invite.get('created_at')
        }
    })
})
```
From the frontend route `/me/pending_invites` we're submitting a `PUT` request to the backend route `/api/v1/me/accept_owner_invite`. As written it feels like fighting Ember - not using the `model.save()` functions or interacting with the store as seem to usually be done. We only really need to have the `crate_id` to do what we need to do on the backend side of things, is there a better way to send just the `crate_id`? Is there a more idiomatic way to submit requests while specifying the route? I'm under the impression that `model.save()` doesn't work here as we have to specify the route to submit the request to, but I'm likely wrong about that. As written this currently works, it just feels like there should be a better way to do it.

Another pain point is refreshing the page after the invitation has been accepted to reveal the updated list of pending invites - I've tried using `this.refresh()` and `this.reset()` after calling the ajax `PUT` function but neither seem to do anything - is there a 'typical' way to get the updated model from the store or backend following an action?

Feedback from anyone with some experience in Ember would be greatly appreciated!


carols10cents:
Ping @locks @Turbo87 @sivakumar-kailasam for some ember guidance! ❤️

Turbo87:
I'm not that deeply familiar with the data model that is used here but this sounds like a one-time action to me, and not something that is persisted in any data models. If that is the case then the approach above seems alright to me, though you could also look into addons like https://github.com/mike-north/ember-api-actions that try to put a little more convention into these one-time actions.

vignesh-sankaran:
@natboehm Could you run `cargo fmt` and push the code back up please? :)

natboehm:
@Turbo87 thanks! I'll have to look at that addon for the future. I eventually was able to get the `model.save()` function to work, I think it looks better than what I had previously.

@vignesh-sankaran thanks, I always forget that! I'll probably close this one in favor of a new PR for the issue, though.

PR #1067, opened 2017-09-20
Don't require http-data files for tests that don't make requests

carols10cents:
When alex [rewrote the test recording code to use hyper](https://github.com/rust-lang/crates.io/commit/827671f3c7003badc8a682e88d584568b42b05a9), all tests were then required to have an http-data file whether they make requests to github/S3 or not. I don't want to have to make a file every time I make a new test, so this just uses `[]` if we're not recording and there isn't a file.


carols10cents:
this is pretty straightforward so i'm going to count @jtgeibel's 👍 as approval :)

bors: r+

Issue #712, opened 2017-05-04
HTTP error codes are barely used


Nemo157:
The only HTTP status codes other than 200 that are currently used are 404 and 403 in the places you'd expect. Any other errors (e.g. validation failures while publishing a crate) are returned as a 200 response with a json object describing the error.

I took a look at all the current uses of `util::errors::human` and they all seem to pretty nicely correspond to errors that should be a 400 Bad Request. I [attempted this change](https://github.com/rust-lang/crates.io/compare/master...Nemo157:error-status-code) but unfortunately `cargo` does not handle getting responses with status code 400 nicely (https://github.com/rust-lang/cargo/issues/3995).

Obviously this is something that may take a little finessing to not break existing cargo. Or the work required for that may make this not worth changing at all.

---

As a reason to do this other than just for consistency with HTTP APIs; I noticed this while attempting to get errors back to the UI in #697. When saving a data model Ember.js expects to either get back a response with status code 200 containing the saved model, or a response with an error status code containing the error messages. I can special case the validation errors, but this would introduce some inconsistency between different request handlers intended for the UI and request handlers intended for cargo.

carols10cents:
Hm. Sounds tricky if crates.io needs to support older versions of cargo. We could use /api/v2/ routes I suppose! I wonder if there's a more clever way to do this though...

Nemo157:
One thought I had was to do user-agent detection, looks like cargo 0.17 at least sends a useful user-agent (`User-Agent=["cargo 0.17.0-nightly (67e4ef1 2017-01-25)"]`) but since we would want the default to be sending back error codes and overriding that for old cargo versions  we'd need to check that cargo has always sent a useful user-agent.

Turbo87:
this situation is unlikely to change since we need to support old cargo versions in a backwards compatible way. for new endpoints we are trying to use the correct HTTP response codes, but for the existing ones we unfortunately can't change anything without risking to break older cargo versions.

if we ever get to a API v2 we could add a compatibility layer that emulated API v1 on top of v2, just for cargo, but this does not seem likely to happen in the near future, so I'll close this issue for now until we actually have the capacity to revisit this.

Issue #912, opened 2017-07-25
Restructuring backend as MVC


vignesh-sankaran:
Relaying the thoughts of @carols10cents from an email sent to me. We could look into setting up modules with the following structure:

- Model: All DB interaction structs
- View: Encodable implementations e.g. Encodable Crate
- Controller: All functions associated with API endpoints

This would also require a restructuring of unit and integration tests, so this would definitely not be a small task. 

Thoughts @carols10cents?

carols10cents:
I am into this! I would prefer to do this incrementally, if possible, so that there isn't one big PR that would need to be reviewed and kept up to date. Even though the intermediate state of being half reorganized might be more confusing, it would be MUCH easier for me to have smaller independent-ish PRs to review and merge :)

jtgeibel:
I think a good place to start would be to move the API endpoints under `cargo` and `frontend` modules.  In particular, it would be nice to see the following in `src/lib.rs`:

``` rust
// used by both `cargo search` and the frontend
api_router.get("/crates", C(krate::index));

// Cargo only routes - comments below reference src/crates-io/lib.rs in cargo's repository
api_router.put("/crates/new", C(krate::cargo::new));                              // publish()
api_router.get("/crates/:crate_id/:version/download", C(krate::cargo::download)); // src/cargo/sources/registry/remote.rs:download()
api_router.get("/crates/:crate_id/owners", C(krate::cargo::owners));              // list_owners()
api_router.put("/crates/:crate_id/owners", C(krate::cargo::add_owners));          // add_owners()
api_router.delete("/crates/:crate_id/owners", C(krate::cargo::remove_owners));    // remove_owners()
api_router.delete("/crates/:crate_id/:version/yank", C(version::cargo::yank));    // yank()
api_router.put("/crates/:crate_id/:version/unyank", C(version::cargo::unyank));   // unyank()

// Frontend routes
api_router.get("/crates/:crate_id", C(krate::frontend::show));
api_router.get("/crates/:crate_id/:version", C(version::frontend::show));
api_router.get("/crates/:crate_id/:version/dependencies", C(version::frontend::dependencies));
api_router.get("/crates/:crate_id/:version/downloads", C(version::frontend::downloads));
```

The search API is the only route I've found that is currently used by both.

I've attached my list of all backend routes and where they are used.  I hope to get this cleaned up and submitted as documentation soon.

[routes.xlsx](https://github.com/rust-lang/crates.io/files/1178772/routes.xlsx)

[...]

PR #1068, opened 2017-09-21
Begin refactoring module layout

jtgeibel:
This PR begins some work towards refactoring the module layout of our rust code.  This PR includes several initial changes:

* Group route definitions in `lib.rs` and document where those api endpoints are currently used.
* Move the `krate` and `version` api modules into subdirectories.  This will be explained further below.
* Move `categories.rs` and `categories.toml` under a new `boot` module as these are only used during boot of the server process.
* Pull github related functionality out of `http` into a separate `github` module, leaving the http security middleware behind.

The second bullet point sets the groundwork for changes similar to what I prototyped in [this branch](https://github.com/rust-lang/crates.io/compare/3f67a0ebca7d8e5b0ac67e4ea33b4b6c6d26a63f...jtgeibel:refactor-api-into-submodules).  I'm hoping that by moving and renaming these to `mod.rs` now, without actually changing the files yet, will cut down on merge conflicts with any future PRs.

# Proposed further changes

As proposed in the previous branch, I believe the next step would be to split the `krate` and `version` modules into submodules which support `cargo` and `frontend` functionality.  At the moment, the `api/v1/crates` route is the only one used by both.  My main motivation for this is that for `cargo` routes we need to strictly maintain backwards compatibility.  For instance, as noted in #712, we always return a `200` response code with an `errors` array because this is expected by `cargo` (with the exception of  `krate::download` where it does not).  If we ever decide to change this then we will need to take care that we don't impact routes used by cargo.  (Of course we need to be careful in general, as we've broken unknown external users in the past, but hopefully this will provide an extra speed bump when authoring and reviewing changes.)

While experimenting on that branch I also got a better feel for how we could split logic between the api/controller level and the model/view level.  My notes on this may be a bit dated, but in some modules we mix api functionality with diesel and serde structs.  For instance, the following files contain logic for all these concerns:

* `keyword.rs`
* `krate/mod.rs`
* `token.rs`
* `user/mod.rs`
* `version/mod.rs`

In contrast, the following files contain only diesel and serde functionality, in support of api calls handled by the files above:

* `category.rs`
* `dependency.rs`
* `download.rs`
* `owner.rs`

It would be nice if we could pull out the database and json related functionality in the first list, and then group all of those files in a new submodule.  At this time, I do not think it is necessary to try to split apart the model and view functionality as this logic is very straightforward.

For now, I hope this is a small enough PR to review independently and provide a place to discuss these suggestions for further refactoring.

/cc @vignesh-sankaran


vignesh-sankaran:
Thanks @jtgeibel, I’ll try taking a look at this today or tomorrow, I would like to try reviewing this @carols10cents :)

vignesh-sankaran:
Ok so from an initial look, I like how the endpoints are now organised by what external user utilisees them. 

I'm a bit worried that the `/crates` endpoint is a singular API for both cargo and the frontend. I think we should look into how `cargo search` and the front end use the `/crates` endpoint  for search since if they don't use all of the returned data or use the data differently, we could split up the functionality accordingly so that we don't accidentally break things if we change this endpoint's return data.

I'll have more feedback tomorrow regarding the rest of the changes and suggestions for future refactoring. Once we've decided on an approach, I think it'd be good to update #912 with a list of tasks so that we can make it E-mentor as well.

src/github.rs
```diff
@@ -0,0 +1,93 @@
+use curl;
```
vignesh-sankaran: Could you add a doc comment here to give a high level overview of what this module does and where it's used?

src/owner.rs
```diff
@@ -186,9 +186,9 @@ impl Team {
         // FIXME: we just set per_page=100 and don't bother chasing pagination
```
vignesh-sankaran: So it turns out that the only thing using this function is on line 138 in `pub fn create()`, and that's within this file itself. While you're modifying this function could you remove the `pub` modifier?

jtgeibel:
I've addressed review comments and have rebased on master with the latest rustfmt updates.

src/github.rs
```diff
@@ -1,3 +1,5 @@
+//! This module implements functionality for interacting with github.
```
vignesh-sankaran: *GitHub, not github :)

jtgeibel: I'd hate to burn CI time on this.  @carols10cents do you know if the issue with [no-ci] on bors-voyager has been fixed?  I know last time I used that it froze something in the pipeline and you had to override it.

jtgeibel:
I amended the last commit and used [skip ci].  Other than the change to the doc comment this group of commits passed on travis at https://travis-ci.org/rust-lang/crates.io/builds/280238869.

carols10cents:
LGTM! let's get this to stop having merge conflicts quick!

bors: r+

vignesh-sankaran:
Yep, I reviewed the functionality and the github + category stuff works fine :). We'll continue the discussion regarding the refactoring in the issue itself.

PR #1070, opened 2017-09-21
Re-enable syntax highlighting.

### LINKED ISSUE #1008: When ammonia supports allowing only certain values for attributes, re-enable syntax highlighting
We [had to disable classes on `code` tags](https://github.com/rust-lang/crates.io/commit/2bf0ce294a90e60665782058b4fbcfb41395394b) because [you could use arbitrary classes to do arbitrary things](https://github.com/ChALkeR/notes/blob/master/Improper-markup-sanitization.md). So for now, syntax highlighting in readmes is disabled.

[I've filed an issue with Ammonia to request being able to specify allowed values for attributes](https://github.com/notriddle/ammonia/issues/24) (or at least classes); [these are the classes that marky-markdown (used by npm) allows](https://github.com/npm/marky-markdown/blob/master/lib/sanitize.js#L27) for their README rendering.

Instructions for implementing:

- We've already updated to ammonia 0.7.0, which now has support for this!
- [ ] Add the [`allowed_classes` field](https://docs.rs/ammonia/0.7.0/ammonia/struct.Ammonia.html) to [the ammonia configuration struct](https://github.com/rust-lang/crates.io/blob/9128a22badd71fbece698ac2b768292dfb437017/src/render.rs#L71-L77) with a hash map that contains the tag name `"code"` and a value of a `HashSet` containing all [the languages we want to support highlighting](https://github.com/rust-lang/crates.io/blob/566dc51aff42986a669ee0d7eba0d157ee569ae7/ember-cli-build.js#L9-L22) prefixed by `language-`, so `language-bash`, `language-clike`, etc.
- [ ] Add a test [in here like these tests](https://github.com/rust-lang/crates.io/blob/566dc51aff42986a669ee0d7eba0d157ee569ae7/src/render.rs#L120-L176) that calls `markdown_to_html` with a code block that should get syntax highlighting, and assert that the `code` tag has `class="language-whatever"`
- [ ] Add another test in there that tries to add classes that aren't allowed to a `code` element, or other classes to other elements, and assert that they aren't in the rendered HTML 
- BONUS: I'm happy to test this for real, but if you're feeling ambitious, publish a crate that has a README with some code in it to your local crates.io instance with the code change ([instructions on publishing to your own instance are here](https://github.com/rust-lang/crates.io/blob/master/docs/CONTRIBUTING.md#publishing-a-crate-to-your-local-cratesio)) and make sure it is highlighted.

treiff:
Syntax highlighting seems to be working as intended now.  I tested by publishing a create locally with a `README.md` including a couple of code samples, see below for a screenshot. Thanks for the help, and please let me know if you have any questions.

Closes #1008 

screen shot 2017-09-21 at 1 07 28 pm

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -54,6 +54,7 @@ impl<'a> MarkdownRenderer<'a> {
             .collect();
         let tag_attributes = [
             ("a", ["href", "target"].iter().cloned().collect()),
+            ("code", ["class"].iter().cloned().collect()),
             (
                 "img",
                 ["width", "height", "src", "alt", "align"]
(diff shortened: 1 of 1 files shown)


carols10cents:
Looks great!! Thank you!! 🔴 🔵 ⚫️ ⚪️ 

bors: r+

PR #1071, opened 2017-09-22
Update Ember CLI and Babel

Turbo87:


Turbo87:
let's see if this works...

bors: r+

PR #1072, opened 2017-09-22
Disable prototype extensions

Turbo87:
Prototype extensions are unfortunately not supported by fastboot so we will need to disable them to prepare for server-side rendering.

(see https://github.com/ember-fastboot/ember-cli-fastboot#prototype-extensions)

### DIFF
--- a/app/helpers/format-email.js
+++ b/app/helpers/format-email.js
@@ -1,4 +1,5 @@
 import { helper } from '@ember/component/helper';
+import { htmlSafe } from '@ember/string';
 import Ember from 'ember';
 
 const escape = Ember.Handlebars.Utils.escapeExpression;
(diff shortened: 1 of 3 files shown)


app/helpers/format-email.js
```diff
@@ -13,7 +14,7 @@ export function formatEmail(email) {
         ret = `<a href='mailto:${escape(formatted[2])}'>${ret}</a>`;
     }
 
-    return ret.htmlSafe();
```
locks: this is going to be deprecated anyway ;)

locks:
bors: r+

app/helpers/format-email.js
Turbo87: correct

PR #1073, opened 2017-09-22
Accept owner invites

natboehm:
This implements deployable chunk 2, accepting invitations of issue #924. I created a new route, `/me/crate_invitations/:crate_id` where a `PUT` request is submitted to. The request body contains the invitation information along with a boolean field `accepted`, indicating whether or not the invitation was accepted. This same route can then also be used for the decline route, a matter of toggling the `accepted` field. When an invitation is accepted, the user should receive a message where the invitation used to be acknowledging the successful addition. If an error occurs, a message should also appear indicating so.


carols10cents:
bors: r+

PR #1076, opened 2017-09-22
Fixes errors on "not found" pages

carols10cents:
Turns out when we were checking for crate/category/etc not being found, we were using `any` from the ember prototype extensions. This PR changes those spots to use `some` instead, and cleans up a few other inconsistencies I noticed.

I tried to write tests for "not found" pages, but that's proving to be a bit more.... interesting... so I decided to open this up to fix production first.

r? @Turbo87


Turbo87:
bors: r+

Thanks @carols10cents!

PR #1078, opened 2017-09-23
Various minor cleanup of the `token` module

sgrif:
- Inlined functions which were only used once
  - In the case of `find_for_user`, I think that the order should be
    close to where it is relevant.
  - In the case of `delete`, the only test that was calling it was
    literally testing "a deleted row is deleted", which is not needed.
    There are tests for the `revoke` endpoint
  - The direct usage of Diesel is not meaningfully more verbose than
    these functions were
- Removed `EncodableApiToken`, moved the `#[derive(Deserialize)]` struct
  to tests


carols10cents:
bors: r+

PR #1079, opened 2017-09-23
Include the generated Diesel setup migration

sgrif:
This used to be done implicitly by Diesel, but it now generates a
migration for this explicitly (since 0.15.0 I believe)


carols10cents:
Works on an existing database and on a new database. 👍 

bors: r+

PR #1080, opened 2017-09-23
Move the `readme_rendering` table onto `versions`

sgrif:
Every row in versions had at most a single row in the `readme_rendering`
table, and the table had only a single column. There is no reason for
this table to live separately.


carols10cents:
bors: r+

PR #1082, opened 2017-09-23
Bump clippy & rustfmt to latest versions

sgrif:
The only new lints that clippy wanted were `Arc::clone(&x)` instead of `x.clone()`, and `u64::from(x)` instead of `x as u64`. The rest of these changes are just from the new rustfmt version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,11 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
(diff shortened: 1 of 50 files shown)


sgrif:
Going to bump these as separate PRs, I don't want to deal with merge conflicts for this, and just bumping rustfmt is easier to rebase on its own

PR #1085, opened 2017-09-25
Decline owner invites

natboehm:
This implements deployable chunk 3: declining invitations of issue #924. The api route for declining an invite is the same as for accepting an invite. To know that an invite is being declined instead of accepted, the `accepted` field passed to the backend is set to `false`. If declined, the invite should be replaced with a message to the user confirming that they declined the invitation for the crate.


src/crate_owner_invitation.rs
```diff
@@ -112,11 +112,7 @@ pub fn handle_invite(req: &mut Request) -> CargoResult<Response> {
     if crate_invite.accepted {
         accept_invite(req, conn, crate_invite)
     } else {
-        #[derive(Serialize)]
-        struct R {
-            crate_owner_invitation: InvitationResponse,
-        }
-        Ok(req.json(&R { crate_owner_invitation: crate_invite }))
+        decline_invite(req, conn, crate_invite)
```
carols10cents: I really like how this ended up, the code's easier to read and well organized! 💯

carols10cents:
bors: r+

PR #1086, opened 2017-09-26
Update contribution guidelines to use --index

Deedasmi:
Introduced in https://github.com/rust-lang/cargo/pull/4209

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -428,13 +428,12 @@ live crates.io, you won't be able to publish that crate locally.
 In your crate directory, run:
 
 ```
-cargo publish --host file:///path/to/your/crates.io/checkout/tmp/index-co
+cargo publish --index file:///path/to/your/crates.io/checkout/tmp/index-co
 ```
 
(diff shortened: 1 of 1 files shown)


kureuil:
Thanks for the pull request @Deedasmi :)

If I remember correctly cargo isn't automatically updated, so wouldn't it be safer to keep the old flag in the documentation, and specify the version ranges compatible with each flag (of course, we'd make `--index` the default) ?

What happens when you run `cargo publish --host file:///path/to/your/crates.io/checkout/tmp/index-co` on the latest cargo version ?

Deedasmi:
A friendly deprecation warning saying --host is no longer valid, and that it will 'soon' become a hard error. 

I do believe `rustup update` automatically updates cargo, so most should be getting updated sooner rather than later. I'm cool with adding a footnote or just closing this and revisiting it later, after a couple more updates.

kureuil:
I think leaving a footnote for old versions of cargo is enough for now :+1:

carols10cents:
Love it!!! Thank you!! 

bors: r+

PR #1088, opened 2017-09-26
Make users::show() case insensitive

### LINKED ISSUE #989: Usernames in /users/ urls should be case-insensitive
What happened:

- I typed in crates.io/users/burntsushi
- I got a 404 :(

What I expected to happen:

- I type in crates.io/users/burntsushi
- I am shown the content I see if I go to https://www.crates.io/users/BurntSushi

Github treats usernamse in a case-insensitive manner; I can visit either https://github.com/burntsushi or https://github.com/BurntSushi. 

We should probably lowercase usernames like we do for keywords, and add a unique index. I will write up more instructions later :)

Deedasmi:
Closes #989.

### DIFF
--- a/migrations/2017-09-26-200549_users_lowercase_index/up.sql
+++ b/migrations/2017-09-26-200549_users_lowercase_index/up.sql
@@ -1,0 +1,1 @@
+CREATE UNIQUE INDEX lower_gh_login ON users (LOWER(gh_login));
--- a/migrations/2017-09-26-200549_users_lowercase_index/up.sql
+++ b/migrations/2017-09-26-200549_users_lowercase_index/up.sql
@@ -1,1 +1,3 @@
-CREATE UNIQUE INDEX lower_gh_login ON users (LOWER(gh_login));
+DROP INDEX IF EXISTS index_users_gh_login;
+ALTER TABLE users ADD CONSTRAINT unique_gh_login UNIQUE(gh_login);
+CREATE UNIQUE INDEX lower_gh_login ON users (lower(gh_login));
--- a/migrations/2017-09-26-200549_users_lowercase_index/down.sql
+++ b/migrations/2017-09-26-200549_users_lowercase_index/down.sql
@@ -1,0 +1,1 @@
+DROP INDEX IF EXISTS lower_gh_login;
--- a/migrations/2017-09-26-200549_users_lowercase_index/down.sql
+++ b/migrations/2017-09-26-200549_users_lowercase_index/down.sql
@@ -1,1 +1,3 @@
-DROP INDEX IF EXISTS lower_gh_login;
+DROP INDEX lower_gh_login;
+ALTER TABLE users DROP CONSTRAINT unique_gh_login;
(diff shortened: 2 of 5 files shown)


sgrif:
Should we use the `CITEXT` extension for this instead? https://www.postgresql.org/docs/9.6/static/citext.html Would just need to change the column type and add the index, but no code changes would be needed.

migrations/2017-09-26-200549_users_lowercase_index/up.sql
```diff
@@ -0,0 +1 @@
+CREATE UNIQUE INDEX lower_gh_login ON users (LOWER(gh_login));
```
sgrif: Should we get rid of `index_users_gh_login` since it's redundant?

Deedasmi:
I saw that, but was trying to follow @carols10cents suggestion from the issue.

Additionally, those tests seem to be failing due to multiple calls to ::sign_in. Simply removing all but the first ::sign_in makes the tests pass. Is multiple sign ins something we should be worried about? If I'm understanding correctly, these will still fail if we used citext.

migrations/2017-09-26-200549_users_lowercase_index/up.sql
Deedasmi: Probably, or just change that index to lower(gh_login) instead of using this index. Will need to address the UNIQUE issue though. Don't know enough about this app to tell if it's the way to unit tests are laid out or if it could be a production issue.

sgrif: If it's a production issue we'll deal with it

sgrif:
You shouldn't remove the calls to `sign_in`, but you should create a single user and replace those calls with `sign_in_as`

Deedasmi:
So `CITEXT` doesn't enforce uniqueness, which we should probably do, and doesn't work when something is marked unique. I made gh_login unique (first as an index, then as a constraint) and it caused a couple tests to fail. I did have some issues getting those fixed, so please review and make sure they are still valid.

sgrif:
bors: r+

carols10cents:
Sorry for the delay-- just deployed this!!! Are you on twitter @Deedasmi ?

Deedasmi:
No need to apologize; I noticed #1109 and the bad assumption I made. Thanks for fixing that! I am not actually. Why do you ask?

carols10cents:
I was going to tweet about the features you've added recently from https://twitter.com/cratesiostatus and mention you if you were :) Mind if I say "Deedasmi" or "Richard Petrie"?

Deedasmi:
I found and fixed up an old account, @deedasmi should work now! Sent from a mobile device. On Oct 10, 2017 9:59 AM, "Carol (Nichols || Goulding)" <notifications@github.com> wrote:I was going to tweet about the features you've added recently from https://twitter.com/cratesiostatus and mention you if you were :) Mind if I say "Deedasmi" or "Richard Petrie"?

PR #1090, opened 2017-09-26
Remove the `tokens` table

sgrif:
This table was always one-to-one with the emails table, and we should
have always been updating it when the email was updated. This moves all
of that logic into the database, and merges the two tables.

The only reason that these tables were ever separate as far as I can
tell was to represent when the confirmation email hadn't been sent.
However, we can just do that with a nullable column.

However, having the `token` column leads to some awkward `unwrap`s in
the code where we return it, since we know that updating the `email`
always generates a new token. For that reason I have made the `token`
column `NOT NULL`, but left the timestamp nullable. Rows in the `emails`
table that did not previously have an associated `token` row will have a
`NULL` value for this column.


sgrif:
@carols10cents FYI I've got a branch updating to Diesel master (likely to represent 1.0 API-wise) ready to go, just blocked on this. Stuff like #1108 (and this, which is why I branched off this PR) will conflict heavily with it since we deprecated our existing APIs both for insert and upsert.

I'm fine with keeping it up to date as you review other stuff, but just wanted to let you know where I'm at

carols10cents:
Sorry sorry sorry. We need #1108 out before the release on thursday, so I'm going to merge and deploy that, then look at this.

sgrif:
Lol it's completely fine. No rush. Like I said, just wanted to let you know what's up

carols10cents:
Phew, had a chance to look at this, looks good! Diesel upgrade away!

bors: r+

sgrif:
<3

PR #1092, opened 2017-09-28
Fixed inadequate unit test

Deedasmi:
Previous unit test only actually tested .to_lowercase(), not the lower(gh_login)


Deedasmi:
Was handled in #1109

PR #1096, opened 2017-09-30
Made the search feature -/_ insensitive.

### LINKED ISSUE #1065: Crates.io search for crate-name doesn't find crate_name
E.g. try searching for "serde-json" (serde_json), "noisy-float" (noisy_float) etc.
It doesn't find them but it should.

Deedasmi:
The changes to the test crate descriptions are because the test enforces sort order.

Closes #1065

### DIFF
--- a/src/krate/mod.rs
+++ b/src/krate/mod.rs
@@ -6,6 +6,7 @@ use chrono::{NaiveDate, NaiveDateTime};
 use conduit::{Request, Response};
 use conduit_router::RequestParams;
 use diesel::associations::Identifiable;
+use diesel::expression::helper_types::Eq;
 use diesel::helper_types::Select;
 use diesel::pg::upsert::*;
 use diesel::pg::Pg;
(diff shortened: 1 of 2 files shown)


src/krate/mod.rs
```diff
@@ -602,18 +602,19 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 
     if let Some(q_string) = params.get("q") {
         let sort = params.get("sort").map(|s| &**s).unwrap_or("relevance");
+        let canon_q_string = canon_crate_name(q_string);
         let q = plainto_tsquery(q_string);
         query = query.filter(
             q.matches(crates::textsearchable_index_col)
-                .or(crates::name.eq(q_string)),
+                .or(canon_crate_name(crates::name).eq(canon_q_string)),
```
sgrif: This is getting repeated quite a bit. Can we pull it out into a function? (Worth extracting from `Crate::by_name` as well).

Something like this should work:

```rust
use diesel::expression::helper_types::Eq;

fn name_canonically_equals(s: &str) -> Eq<canon_crate_name<crates::name>, canon_crate_name<&str>> {
    canon_crate_name(crates::name).eq(canon_crate_name(s))
}
```

Deedasmi: I honestly didn't know that was possible. Thanks! I'll work on it soon.

sgrif:
bors: r+

Issue #950, opened 2017-08-11
Define what "crates.io" is to "rust". "Registry"?


behnam:
I was updating cargo's docs to use the term accurately, and I realized that there's clear term for what "crates.io" *is*.

<https://crates.io/> has these lines in the homepage:
* "Cargo / pacakges for Rust",
* "The Rust community’s crate host",

There's no "About crates.io" page.

In <http://doc.crates.io/manifest.html#the-publish--field-optional>, we have "a package repository", to which I added "(like crates.io)" recently to prevent confusions with a git repository of a package.

In <http://doc.crates.io/manifest.html#the-patch-section>, we also have "the https://crates.io registry".

In <http://doc.crates.io/manifest.html#package-metadata> we again have "The crates.io registry".

In <https://doc.rust-lang.org/book/second-edition/>, we have "the Rust package registry site, crates.io".

IIRC, in some of the error messages and issues/pulls, it's call "package index".

# My Recommendation

IMHO, "package registry" sounds very good to me, and clearly denotes the *registry* role of the website.

When agreed on the term, we should update some of these docs to use the same wording.

steveklabnik:
> I was updating cargo's docs to use the term accurately,

❤️ 💯 

> IMHO, "package registry" sounds very good to me, and clearly denotes the registry role of the website.

👍

carols10cents:
Yep, I agree "package registry" is the right term. Seems like the fixes need to go in the docs in cargo's repo, so I'm going to give this a close!

PR #1098, opened 2017-10-01
Resolves #1031 - Branding updates for crates.io

### LINKED ISSUE #1031: Branding updates for crates.io

SeanPrashad:
Modified the `<h1>` header inside `app/templates/index.hbs` from "The Rust community’s crate **host**" to "The Rust community’s crate **registry**".

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -1,5 +1,5 @@
 <div id='title-header'>
-    <h1>The Rust community&rsquo;s crate host</h1>
+    <h1>The Rust community&rsquo;s crate registry</h1>
 
     <div class='links'>
         {{#link-to "install" class='yellow-button'}}


behnam:
Thanks, @SeanPrashad. The change looks good, but I think we better check something before landing, based on my experience from changing the tagline in the header.

The current text, "The Rust community’s crate host", fits one line in most views/browsers. Since "registry" is a wider word than "host" (regularly), I think we want to check if the new sentence also fits in one line.

I just made the change using Developer Tools in my browser and it still fits one line. Would you double-check that on your side, so we don't have to patch it again if it doesn't fit.

(If it doesn't fit in one line, we need to reduce the size of the text a bit.)

SeanPrashad:
Thanks for that heads up @behnam - I'm not sure how to go about using Developer Tools in my browser to change the font size, or to actually run crates.io locally - is there any documentation that I could read to figure out both of the above?

behnam:
This is Chrome's DevTools docs: https://developer.chrome.com/devtools

This is how to run a local server for the crates.io app: https://guides.emberjs.com/v2.15.0/getting-started/quick-start/



Since you've already tested in a couple of browsers, (and I've checked it in Chrome/Linux, and it looks good) I think we're good to go. @carols10cents?

jtgeibel:
We should also keep in mind rust-lang/cargo#4208.  As far as I can tell, the intent (after a depreciation cycle) is to allow people to pass something like `--host https://crates.io` so we may want to keep the existing nomenclature or change it back after that lands.

behnam:
Thanks, @jtgeibel, for bringing that up. In <https://github.com/rust-lang/crates.io/issues/950>, we had a little discussion about it, but the issue of `--host` didn't come up then.

In [RFC 2141](https://github.com/rust-lang/rfcs/blob/master/text/2141-alternative-registries.md), such a service is clearly named "registry". The keyword used in the CLI is `--index` in that RFC.

And recently there's been PRs to replace `--host` with `--index`, like <https://github.com/rust-lang/cargo/issues/4208> and <https://github.com/rust-lang/cargo/pull/4209>.

So, I think it's safe to say that we want to call crates.io the "package registry" and not "package host" anymore.

carols10cents:
This looks great to me! Thank you!!

bors: r+

SeanPrashad:
@behnam @carols10cents Although the commit has been merged successfully, I still see "**host**" being used on [crates.io](https://crates.io/) instead of "**registry**". Did we miss something?

carols10cents:
Oh we don't have auto-deploy or anything. This'll go out when I next deploy, probably later today. Thanks!

carols10cents:
I lied, it went out today because of Reasons! It's out now-- you might have to do a shift-reload to clear browser cache, but I see it :) Thank you @SeanPrashad!!!

PR #1100, opened 2017-10-02
Fix "pushObjects is not a function" error

### LINKED ISSUE #1099: Latest updates in dashboard don't show up
The HTTP request succeeds but the items are never displayed (the "My crates" and "Following" sections work as expected, but not the "Latest updates").

The console logs this error: "e.get(...).pushObjects is not a function"

locks:
This is a bit of a quick patch to address the problem, we should dig some more into why it's happening.

Fixes #1099

[edit] I only checked `config/environment.js` for something disabling prototype extensions, and forgot to check `package.json`, which lists [ember-disable-prototype-extensions](https://github.com/rust-lang/crates.io/blob/master/package.json#L49).

### DIFF
--- a/app/controllers/dashboard.js
+++ b/app/controllers/dashboard.js
@@ -1,4 +1,5 @@
 import Controller from '@ember/controller';
+import { A } from '@ember/array';
 import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 
@@ -14,9 +15,9 @@ export default Controller.extend({
         this.fetchingFeed = true;
         this.loadingMore = false;
         this.hasMore = false;
-        this.myCrates = [];
-        this.myFollowing = [];
-        this.myFeed = [];
+        this.myCrates = A();
+        this.myFollowing = A();
+        this.myFeed = A();
         this.myStats = 0;
     },
 
--- a/app/routes/dashboard.js
+++ b/app/routes/dashboard.js
@@ -1,4 +1,5 @@
 import Route from '@ember/routing/route';
+import { A } from '@ember/array';
 import RSVP from 'rsvp';
 
 import AuthenticatedRoute from '../mixins/authenticated-route';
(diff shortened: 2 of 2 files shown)


Turbo87:
this seems necessary only for `myFeed` due to https://github.com/rust-lang/crates.io/blob/dc195a352cd7e336233c0694cbd6729f2b2de053/app/controllers/dashboard.js#L52

locks:
@Turbo87 my thought was let's keep it consistent. We could also do `this.get('myFeed').push(...versions)` if `versions` is a native array.

Turbo87:
we could, but then the array observers won't get notified for new items in the array

carols10cents:
Sorry this has taken me so long to look at-- I'm still getting `_this.get(...).pushObjects is not a function` on my dashboard page :-/

I'm going to see if I can get a test to fail for this...

Turbo87:
> I'm still getting `_this.get(...).pushObjects` is not a function on my dashboard page

even after applying this patch?

carols10cents:
yep, on this branch

carols10cents:
does this branch fix it for you?

carols10cents:
Welp i just tried to make a test with a logged in user for the past few hours and i fail. Pushed my attempt to this branch, feel free to toss it if it's crap.

Turbo87:
@carols10cents @locks I finally got around to looking into this and after a while I managed to figure out how to setup everything so that auth works locally too.

tl;dr we missed `controller.set('myFeed', []);` in the `dashboard` route file

I've fixed the instruction and pushed the fix to this PR. I've also moved your test to a separate `me-test` branch to make CI pass again here, but we should definitely continue to explore how to add tests for these pages.

locks:
@carols10cents can you 👍 or merge?

carols10cents:
Yay it works!!! Thank you!!!

bors: r+

PR #1101, opened 2017-10-03
Fixes #962 Backend: add "most_recently_downloaded" to 'summary' endpoint

### LINKED ISSUE #962: Add another column to the homepage for "Most Recently Downloaded"
Instructions for fixing this, can be done in 2 separate PRs but they will need to be done in order :)

## PR 1: backend

- [x] In the [`summary` function that returns JSON for the /api/v1/summary route](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/krate.rs#L759), add a query similar to [these queries](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/krate.rs#L782-L797) that selects 10 crates sorted by [recent downloads descending nulls last](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/krate.rs#L603), which is defined [by this query that's part of the `index` function](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/krate.rs#L584-L592) (I'm fine with copy-pasting that code, if you feel ambitious and want to refactor to remove the duplication, feel free to give that a try!)
- [x] Add a field for those crates to the [returned data struct definition](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/krate.rs#L816-L818) and the [returned struct instance](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/krate.rs#L825-L827) in the same manner as the other fields that return crates
- [x] Add a test that 2 or 3 crates are returned in the expected order for this field, [the `CrateBuilder` provides a way to set a crate's recent downloads for testing](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/tests/krate.rs#L283-L287). The summary route [isn't tested very much right now](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/tests/krate.rs#L980-L985); a bonus would be adding more tests for the other parts of summary as well!

## PR #2: Frontend

- [ ] Once /api/v1/summary returns recent downloads, add a section [like these](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/app/templates/index.hbs#L39-L50) to display recent downloads. Please think of a good header for this, "Most Recently Downloaded" sounds like this is what crates people have downloaded in the last minute? "Most Downloaded in the Last 90 Days" is accurate but long. "Most Recent Downloads"? "Most Downloads Recently"?
- [ ] Check how it looks and at different browser widths; I think the CSS should Just Work but there's a chance it might need some tweaks :)

erewok:
## Description

- Adds "most_recently_downloaded" property to JSON response returned from `api/v1/summary`

- Adds tests for `api/v1/summary` endpoint

### DIFF
--- a/src/krate/mod.rs
+++ b/src/krate/mod.rs
@@ -756,9 +756,9 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 /// Handles the `GET /summary` route.
 pub fn summary(req: &mut Request) -> CargoResult<Response> {
     use schema::crates::dsl::*;
-    use diesel::expression::{DayAndMonthIntervalDsl};
+    use diesel::expression::DayAndMonthIntervalDsl;
     use diesel::types::{BigInt, Nullable};
-    use diesel::expression::functions::date_and_time::{now, date};
+    use diesel::expression::functions::date_and_time::{date, now};
     use diesel::expression::sql_literal::sql;
 
     let conn = req.db_conn()?;
@@ -800,11 +800,12 @@ pub fn summary(req: &mut Request) -> CargoResult<Response> {
 
     let recent_downloads = sql::<Nullable<BigInt>>("SUM(crate_downloads.downloads)");
     let most_recently_downloaded = crates
-        .left_join(crate_downloads::table.on(
-            id.eq(crate_downloads::crate_id).and(
-                crate_downloads::date.gt(date(now - 90.days())),
+        .left_join(
+            crate_downloads::table.on(
+                id.eq(crate_downloads::crate_id)
+                    .and(crate_downloads::date.gt(date(now - 90.days()))),
             ),
-        ))
+        )
         .group_by(id)
         .order(recent_downloads.clone().desc().nulls_last())
         .limit(10)
--- a/src/krate/mod.rs
+++ b/src/krate/mod.rs
@@ -755,11 +755,9 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 
 /// Handles the `GET /summary` route.
 pub fn summary(req: &mut Request) -> CargoResult<Response> {
-    use schema::crates::dsl::*;
-    use diesel::expression::DayAndMonthIntervalDsl;
+    use diesel::expression::{date, now, sql, DayAndMonthIntervalDsl};
     use diesel::types::{BigInt, Nullable};
-    use diesel::expression::functions::date_and_time::{date, now};
-    use diesel::expression::sql_literal::sql;
+    use schema::crates::dsl::*;
 
     let conn = req.db_conn()?;
     let num_crates = crates.count().get_result(&*conn)?;
@@ -807,7 +805,7 @@ pub fn summary(req: &mut Request) -> CargoResult<Response> {
             ),
         )
         .group_by(id)
-        .order(recent_downloads.clone().desc().nulls_last())
+        .order(recent_downloads.desc().nulls_last())
         .limit(10)
         .select(ALL_COLUMNS)
         .load::<Crate>(&*conn)?;
--- a/src/krate/mod.rs
+++ b/src/krate/mod.rs
@@ -756,6 +756,10 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 /// Handles the `GET /summary` route.
 pub fn summary(req: &mut Request) -> CargoResult<Response> {
     use schema::crates::dsl::*;
+    use diesel::expression::{DayAndMonthIntervalDsl};
+    use diesel::types::{BigInt, Nullable};
+    use diesel::expression::functions::date_and_time::{now, date};
+    use diesel::expression::sql_literal::sql;
 
     let conn = req.db_conn()?;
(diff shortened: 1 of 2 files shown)


src/krate/mod.rs
```diff
@@ -756,6 +756,10 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 /// Handles the `GET /summary` route.
 pub fn summary(req: &mut Request) -> CargoResult<Response> {
     use schema::crates::dsl::*;
+    use diesel::expression::DayAndMonthIntervalDsl;
+    use diesel::types::{BigInt, Nullable};
+    use diesel::expression::functions::date_and_time::{date, now};
```
sgrif: `use diesel::expression::{date, now};`

src/krate/mod.rs
```diff
@@ -756,6 +756,10 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
...
+    use diesel::expression::sql_literal::sql;
```
sgrif: `use diesel::expression::sql;`

src/krate/mod.rs
```diff
@@ -794,6 +798,20 @@ pub fn summary(req: &mut Request) -> CargoResult<Response> {
...
+    let recent_downloads = sql::<Nullable<BigInt>>("SUM(crate_downloads.downloads)");
+    let most_recently_downloaded = crates
+        .left_join(
+            crate_downloads::table.on(
+                id.eq(crate_downloads::crate_id)
+                    .and(crate_downloads::date.gt(date(now - 90.days()))),
+            ),
+        )
+        .group_by(id)
+        .order(recent_downloads.clone().desc().nulls_last())
```
sgrif: Why are we cloning this?

src/tests/krate.rs
```diff
@@ -71,6 +82,7 @@ fn new_crate(name: &str) -> u::NewCrate {
     }
 }
 
+
```
sgrif: ✂️

src/tests/krate.rs
```diff
@@ -983,6 +995,64 @@ fn summary_doesnt_die() {
...
+        Category::update_crate(&conn, &krate, &["cat1"]).unwrap();
+        Category::update_crate(&conn, &krate2, &["cat1"]).unwrap();
+    }
+
+    let mut req = ::req(app.clone(), Method::Get, "/api/v1/summary");
+    let mut response = ok_resp!(middle.call(&mut req));
+    let json: SummaryResponse = ::json(&mut response);
+
+    assert_eq!(json.num_crates, 4);
+    assert_eq!(json.num_downloads, 0); // need to add a record to metadata
```
sgrif: Sounds like you need to fix this?

src/tests/krate.rs
```diff
@@ -983,6 +995,64 @@ fn summary_doesnt_die() {
...
+    assert_eq!(json.most_downloaded[0].name, "most_recent_downloads");
+    assert_eq!(
+        json.most_recently_downloaded[0].name,
+        "most_recent_downloads"
+    );
+    assert_eq!(json.popular_keywords[0].keyword, "popular");
+    assert_eq!(json.popular_categories[0].category, "Category 1");
+    assert_eq!(json.just_updated.len(), 0); // update a couple before running this request...
```
sgrif: Sounds like you need to fix this?

src/krate/mod.rs
erewok: ah, good point. Bad copy pasta.

src/tests/krate.rs
```diff
@@ -5,11 +5,12 @@ use std::io::prelude::*;
 use std::fs::{self, File};
 
 use conduit::{Handler, Method};
-
 use git2;
+use diesel::update;
 use self::diesel::prelude::*;
 use serde_json;
 use semver;
+use chrono::Utc;
```
sgrif: Can you sort these

src/krate/mod.rs
```diff
@@ -756,6 +756,10 @@ pub fn index(req: &mut Request) -> CargoResult<Response> {
 /// Handles the `GET /summary` route.
 pub fn summary(req: &mut Request) -> CargoResult<Response> {
     use schema::crates::dsl::*;
+    use diesel::expression::DayAndMonthIntervalDsl;
+    use diesel::types::{BigInt, Nullable};
+    use diesel::expression::{date, now};
+    use diesel::expression::sql;
```
sgrif: Can you sort these and squash all the `expression` imports into one

src/tests/krate.rs
erewok: Sure.

sgrif:
bors: r+

PR #1107, opened 2017-10-04
Always update teams' github IDs

### LINKED ISSUE #1104: Github team logins should be the primary key of teams, not github ids - we should update those
To reproduce:

- Create a github org and a team in that org.
- Add that team as an owner of a crate with `cargo owner --add github:org:team`.
- Delete the team in github.
- Recreate the team *with the same name*.
- Add that team as an owner of another crate with `cargo owner --add github:org:team`.
- You'll now get an error that says only members of that team can add it as an owner

The problem is we're never updating the team github_id-- UNLIKE users, where we need to keep the github id the same across renames of the user account, the `github:org:team` name is the "stable" part, but the ID is what gets used when we query github for whether someone is a member of a team or not, but we might have an out of date ID if a team was deleted and then recreated with the same name, which gives it a different github id.

carols10cents:
Fixes #1104.

Every time we check with GitHub for team membership, also check with GitHub for the team ID, since that may have changed even if the team name hasn't changed since you can delete a team and recreate it with the same team.

I don't really have tests for this specifically since it would require modifying stuff in GitHub while the tests are running...

I also noticed that the recorded http data for the team tests only contained 403 forbidden responses because GitHub changed their org OAuth settings to need explicit opt-in at the org level and we hadn't configured that on the test accounts :( So I fixed that and rerecorded those tests.

r? @sgrif


sgrif:
bors: r+

PR #1108, opened 2017-10-05
Require owner approval

natboehm:
This PR addresses issue #924, adding owners with their consent, deployable chunk 4. It implements the Crates functionality necessary for Cargo to invite a user to be an owner instead of automatically adding any user to be an owner of any crate. These changes correlate with the Cargo changes made in [this PR](https://github.com/rust-lang/cargo/pull/4551). Function `owner_add` now creates an owner invitation instead of adding a user to be an owner, and returns a `boolean` okay value with a `String` message indicating that the user was invited to be an owner of the crate. This `boolean` okay value is not explicitly used for anything but is necessary to support older versions of Cargo, without which older versions will fail to decode only a `String` response.


sgrif:
I think you need to update your rustfmt version

carols10cents:
Yeah sorry about that @natboehm -- we updated rustfmt the other day. I wish there was a better way to notify each other when we do that... I have it set up locally so I just pushed a commit with rustfmt run with the newer version, let's see how that goes :)

carols10cents:
I also just resolved the merge conflict since it was due to that PR of mine that I mentioned wednesday, oops!

natboehm:
Okay, it's fine, thanks for resolving all of that!

carols10cents:
Tested this out locally and it works great!! Thank you!! ❤️ ❤️ ❤️ ❤️ ❤️ 

bors: r+

PR #1109, opened 2017-10-07
Gh login cant be unique

carols10cents:
So I can't deploy to production right now because of the unique index added to users.gh_login.

The values on production for gh_login aren't unique, and here's why:

- Someone can sign up for github with username whatever.
- They can sign in to crates.io. We now get their gh_id, say, 3 (which *is* unique) and their gh_login, whatever. Their user account gets crates.io id 10.
- They rename or delete their github account, which leaves the username 'whatever' available again. (If they renamed their github account, their gh_id stays stable and we'd migrate their account to their new username next time they log in)
- Someone else signs up for github with username wHaTeVeR.
- They can sign in to crates.io. We now get their gh_id, 9000, and their gh_login, wHaTeVeR. Their user account gets crates.io id 10000.

We can't get rid of the first whatever account, because they might sign in again with a renamed account but the same id, and they should be able to still access the crates they own and all that. 

All we know is that the account with the latest crates.io id for that lower(gh_login) is the last valid account we know of for that username (and should match the account that you click through to github, as long as they haven't renamed or deleted their github account since).

r? @sgrif

### DIFF
--- a/migrations/2017-10-06-234455_fix_local_dbs_unique_gh_login/up.sql
+++ b/migrations/2017-10-06-234455_fix_local_dbs_unique_gh_login/up.sql
@@ -1,0 +1,7 @@
+-- This migration should have no effect on production.
+-- Its sole purpose is to fix the local database of anyone who ran 2017-09-26-200549 locally--
+-- that migration can't be run on production because lower(gh_login) isn't unique on production.
+
+CREATE INDEX IF NOT EXISTS index_users_gh_login ON users (gh_login);
+ALTER TABLE users DROP CONSTRAINT IF EXISTS unique_gh_login;
+DROP INDEX IF EXISTS lower_gh_login;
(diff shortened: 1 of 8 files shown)


migrations/2017-10-06-234455_fix_local_dbs_unique_gh_login/up.sql
```diff
@@ -0,0 +1,7 @@
+-- This migration should have no effect on production.
+-- Its sole purpose is to fix the local database of anyone who ran 2017-09-26-200549 locally--
+-- that migration can't be run on production because lower(gh_login) isn't unique on production.
+
+CREATE INDEX IF NOT EXISTS index_users_gh_login ON users (gh_login);
```
sgrif: We still want this index to be on the lowercase name though, don't we? Should we have another migration that is along the lines of:

```sql
DROP INDEX index_users_gh_login;
CREATE INDEX lower_gh_login ON users (lower(gh_login));
```

carols10cents: ah, good call!

carols10cents:
hows about now @sgrif ?

sgrif:
bors: r+

PR #1110, opened 2017-10-07
Fix heroku link in architecture.md

LeopoldArkham:


carols10cents:
Oops, good catch!! Thanks!!

bors: r+

PR #1111, opened 2017-10-08
Fix appveyor badge display and link.

### LINKED ISSUE #587: bad appveyor badges for projects with underscore
The appveyor badge on https://crates.io/crates/erlang_nif-sys does not load, however the link does work when clicked.

The problem is that the link to the appveyor project requires underscores converted to dashes, but the link to the badge requires an unmangled name.  The current crates.io code appears to use the same name for both.

I currently use a premangled repository name in the badges section:
```
[badges]
travis-ci = { repository = "goertzenator/erlang_nif-sys" }
appveyor  = { repository = "goertzenator/erlang-nif-sys" }
```

The resulting broken badge URL
https://ci.appveyor.com/api/projects/status/github/goertzenator/erlang-nif-sys?svg=true&branch=master

The corrected URL is:
https://ci.appveyor.com/api/projects/status/github/goertzenator/erlang_nif-sys?svg=true&branch=master

I expect I could make the badge image to work by changing the badge setting to `appveyor  = { repository = "goertzenator/erlang_nif-sys" }`, but then the link would break.  I think both should work.

treiff:
This fixes an issue with appveyor badges where if your repository name included an underscore, the link generated would not work because appveyor urls use dashes.

You can now specify an optional `project_name` for the appveyor badge in your `Cargo.toml`
```
[badges]
appveyor = { repository = "example/test_crate", service = "github", project_name = "example/test-crate" }
```

If the optional `project_name` attribute is present, we will use that for the link URL, otherwise we default to the required `repository` attribute and replace the underscores with dashes.

The reason to provide the `project_name` option is that if you change the repository name appveyor doesn't track the change.

Let me know if I missed anything!  I did create a test crate and set it up with appveyor to confirm.

Closes #587

### DIFF
--- a/app/components/badge-appveyor.js
+++ b/app/components/badge-appveyor.js
@@ -26,6 +26,11 @@ export default Component.extend({
         return this.get('badge.attributes.branch') || 'master';
     }),
 
+    projectName: computed('badge.attributes.project_name', function() {
+        return this.get('badge.attributes.project_name') ||
+            this.get('badge.attributes.repository').replace(/_/g, '-');
+    }),
+
     service: computed('badge.attributes.service', function() {
         return this.get('badge.attributes.service') || 'github';
     }),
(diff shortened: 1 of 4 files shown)


carols10cents:
Works great!!! Thank you!!!!

bors: r+

PR #1112, opened 2017-10-09
Show download graph as a stacked area chart

### LINKED ISSUE #393: Option to display downloads chart as percentage
The current chart is good for determining how absolute number of downloads has changed over time. It might also be helpful to display version downloads as a percent to see version adoption over time. [An example of a similar chart for Android version adoption:](https://mixpanel.com/trends/#report/android_os_adoption)

screen shot 2016-08-01 at 10 40 42 pm

Munksgaard:
The purpose of this PR is to change the download chart such that it shows how the absolute number of downloads compares from version to version. Instead of having lots of different lines that might be hard to compare to each other, we show the 7-day average as an area. Additionally, the areas are stacked, so you can see the total and compare between versions.

Here's an example of how it might look (that's the libc crate):

graph

This PR intends to close #393, although I realize it doesn't do exactly what was proposed. I think this is a simple fix that improves the readability of the graph, but I'm prepared to change it to more closely mimic the proposal in #393 if need be.

### DIFF
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -137,7 +137,7 @@ export default Component.extend({
                 pointShape: 'square'
             };
             seriesOption[i * 2 + 1] = {
-                type: 'line',
+                type: 'area',
                 color: COLORS[i % COLORS.length],
                 lineWidth: 2,
                 curveType: 'function',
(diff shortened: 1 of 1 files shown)


carols10cents:
Awesome! I'm up for getting this out there and seeing what people think! Thank you!!!

bors: r+

PR #1115, opened 2017-10-09
Add endpoint to display deployed git sha.

### LINKED ISSUE #879: Make the git SHA of the currently deployed code visible through ember-cli-app-version
I had this at a previous job and it was *incredibly* useful-- anyone could help debug a problem with the additional information of knowing whether the current master branch matches what's on production.

As discussed below, the frontend part is blocked waiting on [this PR](https://github.com/EmberSherpa/ember-cli-app-version/pull/42) to ember-cli-app-version.

The backend way to get this info isn't blocked though, and implementing that will entail:

- [ ] Add a new `get` route to [the api_router around here in src/lib.rs](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/src/lib.rs#L177) that's something like `/deployed_sha` (I wouldn't call it anything like "version" since we have crate versions all over the place and this isn't a crate version)
- [ ] Make that route call a new function, which could probably just live in `src/lib.rs` since it'll be small and doesn't really have anything to do with anything else, or mayyybe in `src/util/` somewhere, up to you.
- [ ] Make that function return JSON that looks something like `{"sha": "259d028bd03a07b2560f873add4b45f35b1d565b"}`. Assume that I've enabled [this heroku labs feature](https://devcenter.heroku.com/articles/dyno-metadata) and that the sha is available in the env var `HEROKU_SLUG_COMMIT`. [This code](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/src/krate.rs#L1081-L1085) is a decent example of constructing a JSON response that has a `String` value
- [ ] If `HEROKU_SLUG_COMMIT` isn't set, this function shouldn't panic, have it return `{"sha": "unknown"}` or something. [This code](https://github.com/rust-lang/crates.io/blob/4c4381169e82f15a3244d70797092c09bcc225b2/src/bin/server.rs#L95) is a decent example of getting an env var and using a default value if the env var isn't set.
- [ ] Locally, run `HEROKU_SLUG_COMMIT=something cargo run --bin server` and then visit http://localhost:4200/api/v1/deployed_sha (or whatever you decided to call the route), and verify that you get the JSON you expect with the value you set the env var to.
- [ ] Also run without that env var, visit the same url, and verify you get the default value.

I'm not even sure how you'd write automated tests for this backend route, so I don't think figuring that out is worth it. Manual verification should be fine.

treiff:
This PR adds an api endpoint `/deployed_sha`, which returns a JSON response containing that sha `{ "sha": "abc" }` or if it's unavailable returns `{ "sha": "unknown" }`.

Closes #879

### DIFF
--- a/src/site_metadata.rs
+++ b/src/site_metadata.rs
@@ -1,0 +1,20 @@
+use conduit::{Request, Response};
+use util::RequestUtils;
+use util::errors::CargoResult;
+
+/// Returns the JSON representation of the current deployed commit sha.
+///
+/// The sha is contained within the `HEROKU_SLUG_COMMIT` environment variable.
+/// If `HEROKU_SLUG_COMMIT` is not set, returns `"unknown"`.
+pub fn show_deployed_sha(req: &mut Request) -> CargoResult<Response> {
+    let deployed_sha =
+        ::std::env::var("HEROKU_SLUG_COMMIT").unwrap_or_else(|_| String::from("unknown"));
+
+    #[derive(Serialize)]
+    struct R {
+        deployed_sha: String,
+    }
+    Ok(req.json(&R {
+        deployed_sha: deployed_sha,
+    }))
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,6 +97,7 @@ pub mod user;
 pub mod util;
 pub mod version;
 pub mod email;
+pub mod site_metadata;
 
 mod local_upload;
 mod pagination;
@@ -202,6 +203,7 @@ pub fn middleware(app: Arc<App>) -> MiddlewareBuilder {
     api_router.get("/summary", C(krate::summary));
     api_router.put("/confirm/:email_token", C(user::confirm_user_email));
     api_router.put("/users/:user_id/resend", C(user::regenerate_token_and_send));
+    api_router.get("/site_metadata", C(site_metadata::show_deployed_sha));
     let api_router = Arc::new(R404(api_router));
 
     let mut router = RouteBuilder::new();


treiff:
Hey @carols10cents, I manually tested and it seems to be working nicely. Let me know what you think.  

One thing that I didn't do is to add a comment for the `show_deployed_sha` function, I wasn't really sure what to do here.

Also, something felt a little strange about having to pull the few other modules into scope. This could also just be the rubyist in me, but I'm curious if there is a better/alternate way? Thanks.

jtgeibel:
Thanks a lot @treiff!  I have a few comments on the PR, and then a few general thoughts on the API.

I think we should pull the show_deployed_sha() function (and the associated `use`d types) out into a separate file.  Maybe something like `src/site_metadata.rs`.  There are also a few errors on CI from rustfmt and clippy.

@carols10cents I have a few general thoughts regarding the API and route.  I'm not sure if this should under `/api/v1` or not.  I can kind of see an argument for either.  Also, I'm wondering if we should make this more generic.  For instance, I expect in the future we may need to provide an API to advertise the location of the index (so that we can specify `https://crates.io` in cargo and have it obtain a URL for the index via our API).  So this leads me to the following suggestions for this PR, assuming you agree:

* Move to `/api/v1/site_metadata`
* Return `{"deployed_sha": "abc"}`
* In the future, we can add an "index_url" key to this endpoint to advertise the associated git repository for the index.

Also, its worth noting that obtaining this at runtime requires us to enable it on heroku:

```
$ heroku labs:enable runtime-dyno-metadata -a <app name>
```

treiff:
@jtgeibel Thanks for the feedback! What you propose sounds reasonable to me.  I'll wait till we hear back before moving forward.

Also, any suggestions on `rustfmt` I seem to be struggling to get it working.  I'm using the [rust-lang/rust.vim](https://github.com/rust-lang/rust.vim) vim plugin. I'm thinking it's something to do with my installation of `rustfmt`. Anyway, no worries I'll get it figured out.

jtgeibel:
@treiff I'm stealing this from the `.travis.yml` file.  As I understand it, we're using nightly rustfmt and it is a bit difficult at the moment.

```
rustup install nightly
cargo install --force rustfmt-nightly --vers 0.2.7
export LD_LIBRARY_PATH=$(rustc --print sysroot)/lib:$LD_LIBRARY_PATH
```

Then running `cargo fmt` should modify the files in place, or you can run `cargo fmt -- --write-mode=diff` which will provide a diff showing the changes.

Getting clippy up and running is a bit more complicated (there is an open PR to upgrade us to the latest nightly).  For that, it is probably easiest to check the results on travis-ci.  I see that it recommends the following:

```
error: use of `unwrap_or` followed by a function call

   --> src/lib.rs:318:15

    |

318 |     let sha = ::std::env::var("HEROKU_SLUG_COMMIT").unwrap_or(String::from("unknown"));

    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this `::std::env::var("HEROKU_SLUG_COMMIT").unwrap_or_else(|_| String::from("unknown"))`
```

jtgeibel:
@treiff you'll also want to make sure you are using nightly when compiling and running rustfmt.  By either setting it as the default (`rustup default nightly`) or prepending the commands with `rustup run nightly`.

```
rustup run nightly cargo install --force rustfmt-nightly --vers 0.2.7
export LD_LIBRARY_PATH=$(rustup run nightly rustc --print sysroot)/lib:$LD_LIBRARY_PATH
rustup run nightly cargo fmt
```

jtgeibel:
@treiff, one last update, hopefully.  You'll want to drop the `--vers 0.2.7` argument or manually specify `--vers 0.2.8` when installing rustfmt.  On CI we're tied to a slightly older nightly version of the compiler and rustfmt, but 0.2.8 works fine with the latest nightly compiler.

treiff:
Thanks for those great suggestions @jtgeibel! I'll give it a try tomorrow. 😀

carols10cents:
Looks like you've gotten rustfmt working!!! Awesome!!

I think I like @jtgeibel's suggestion of /api/v1/site_metadata for the route name and returning `{"deployed_sha": "abc"}` sounds like a nice way to make room for anything else we might need along these lines. Do you mind making that change? Thanks!

treiff:
@carols10cents yup, I can take care of it.  Might be a couple days as I've got a bunch going on but I'll get to it.  Thanks.

treiff:
Disregard that previous comment @carols10cents, should be good to go.  Also, thanks @jtgeibel.

src/site_metadata.rs
```diff
@@ -0,0 +1,16 @@
+use conduit::{Request, Response};
+use util::RequestUtils;
+use util::errors::CargoResult;
+
+pub fn show_deployed_sha(req: &mut Request) -> CargoResult<Response> {
```
vignesh-sankaran: Could you add a doc comment here outlining what this function does?

carols10cents:
Works great!!! Thank you!!

bors: r+

PR #1118, opened 2017-10-10
Update to Diesel master

sgrif:
Diesel's current master branch is likely to be reasonably representative
of what 1.0 will look like from an API perspective. If there's any other
parts of the code that we dislike and want to improve the API for, now
is the time to speak up!

The test changes are due to some deadlocks that I'm seeing now (which
should have always been happening, I suspect some perf characteristic
has changed somewhere)


src/bin/update-downloads.rs
```diff
@@ -93,23 +80,23 @@ fn collect(conn: &PgConnection, rows: &[VersionDownload]) -> QueryResult<()> {
...
-            (crate_downloads::crate_id, crate_downloads::date),
-            do_update().set(crate_downloads::downloads.eq(crate_downloads::downloads + amt)),
-        )).into(crate_downloads::table)
+        insert_into(crate_downloads::table)
+            .values((
+                crate_downloads::crate_id.eq(crate_id),
+                crate_downloads::downloads.eq(amt),
+                crate_downloads::date.eq(download.date),
+            ))
+            .on_conflict(crate_downloads::table.primary_key())
```
sgrif: Indentation is so much better now...

src/schema.rs
```diff
@@ -783,3 +783,26 @@ joinable!(version_authors -> users (user_id));
 joinable!(version_authors -> versions (version_id));
 joinable!(version_downloads -> versions (version_id));
 joinable!(versions -> crates (crate_id));
+
+allow_tables_to_appear_in_same_query!(
```
sgrif: Hopefully this is less confusing now

carols10cents: If we generate it for all tables, why do we have to explicitly list this at all...?

src/krate/mod.rs
```diff
@@ -280,20 +279,16 @@ impl<'a> NewCrate<'a> {
 }
 
 impl Crate {
```
vignesh-sankaran: Could you add doc comments to the `by_name`, `with_name`, and `all` functions in this struct?

src/schema.rs
sgrif: I don't understand the question

carols10cents: > we just speculatively generate it for all tables

By "we", do you mean "crates.io" or "diesel"? If Diesel is doing this speculative generation, why does it have to be listed explicitly in src/schema.rs at all? Why isn't this built-in to diesel?

sgrif: This file was generated by `diesel print-schema`

sgrif:
I've rebased and killed the remaining raw inserts in `update-downloads.rs`. I'm going to leave the ones in `category.rs` for now, as they're a bit verbose and this PR is already really big

carols10cents:
Working ok for me! Let's do this!

bors: r+

carols10cents:
oops

carols10cents:
bors: r+

carols10cents:
i'm getting real tired of that intermittent faker error...

bors: r+

sgrif:
bors: retry

Looks like it's some NPM issue

sgrif:
I guess retry doesn't work?

bors: r+

carols10cents:
bors: r+

sgrif:
JavaScript is the worst.

Issue #1093, opened 2017-09-30
Provide link to parent category


aturon:
Subcategories don't provide a way to get back to their parents. E.g., at https://crates.io/categories/development-tools::debugging, I expected "Development tools" to be a link.

erewok:
I was looking at this and another issue ([#721](https://github.com/rust-lang/crates.io/issues/721)) and I had a suggestion for organizing the `categories` table that I think would make these things easier. 

Because the categories are hierarchical, it might be more ergonomic if the endpoints were able to rely more on the database to organize the data hierarchically.

I noticed there's a few places where [the code building SQL queries like this](https://github.com/rust-lang/crates.io/blob/726906f6e2196a5c90f0d7dda645f2ff719cf231/src/category.rs#L137) to look up parent categories:

```
split_part(c.slug, '::', 1) = c.slug
```

We could implement more proper trees in the database by utilizing either a nullable `parent_id` column on the `categories` table or adding a `path` column that is of the type `ltree` which gets updated based on the slug (automatically via triggers or in application code).

The latter appeals to me because it allows for some easier queries and because it seems more "harmless" from a migration path (the syntax does look a bit weird, perhaps, if you're not used to it). There are trade-offs, though. For instance, I think it would end up being completely invisible to application code to use an `ltree` and you'd need to hand write SQL (which the app is doing anyway to work with these categories.

I started a POC branch to work on implementing this and I believe it would work and make working with the hierarchy of categories easier, but wanted to see what the community thought. I'm new here, so happy to defer to other opinions.

jtescher:
@carols10cents this seems to be fixed by https://github.com/rust-lang/crates.io/pull/1122, can probably be closed.

carols10cents:
Good catch, thanks!

Issue #721, opened 2017-05-12
category_slugs is very hard to find


lilianmoraru:
I wanted to add some categories to a project and I searched ~40-60 minutes until I found them(I found the page randomly before, when trying to publish one of my own projects - so I knew how it should look like and that these exist).

I went through all the options in the `Docs` on `crates.io` and could not find it.
Searched on Google and found `crates.io/categories`, which is something(you can't find the categories very easily from the main page unless you detect the `see all` at the bottom of the page, between the 5 lists), but still not helpful to a developer that wants to add a category to `Cargo.toml`.

In the end, through Google, I stumbled on this issue: https://github.com/rust-lang/crates.io/issues/544 and observed in the mentioned link: http://doc.crates.io/manifest.html#package-metadata through the comments in the code, [crates.io/category_slugs](https://crates.io/category_slugs) was mentioned - I finally found it.

Could we please have the `category_slugs` link in the `Docs` drop-down(crates.io) or maybe somewhere on the `crates.io/categories` page?

carols10cents:
Yup, great suggestion!

carols10cents:
Let's go with the crates.io/categories page. Doing this involves adding a link to `/category_slugs` somewhere in [the categories template](https://github.com/rust-lang/crates.io/blob/405b2ad74715f22d756de749be4acb9626c1019b/app/templates/categories.hbs) -- where and what the link and the text around it says exactly is up to whoever implements this :)

autodidaddict:
![screenshot from 2017-08-20 19-17-30](https://user-images.githubusercontent.com/13748641/29499319-4999983e-85dc-11e7-9200-b397afc41201.png)

Here's a quick straw man I whipped up. The `here` link goes to the category slugs page while the `Cargo.toml` link goes to the manifest documentation page.

Thoughts? I figured showing this here first would be easier than starting a PR. I tried a bunch of variations with this text at the top and I couldn't find any that didn't distract from the category list.

lilianmoraru:
To be sincere, the visibility of that is not very good.
Personally, I probably would never end-up seeing that link.

May be a mention/"chapter" on the ["Publishing on crates.io" page](http://doc.crates.io/crates-io.html)(both options or just on the Publishing page?)?

autodidaddict:
Yeah it's too subtle but this issue called for it being on this page.

autodidaddict:
[...]

PR #1122, opened 2017-10-11
Proposal: Add `path` column to categories table as `ltree` type to make tree traversing easier

erewok:
## Description

This is a proposal inspired by tickets ([1](https://github.com/rust-lang/crates.io/issues/1093), and [2](https://github.com/rust-lang/crates.io/issues/721)) with the goal of eventually revealing more information about categories on pages where category information is available. It is also motivated by the current application code surrounding categories, which contains lots of SQL with operations like this: `split_part(c2.slug, '::', 1)`.

This PR is intended as a suggestion for what to do with the `categories` table.

Following are changes included here:

- Adds a `path` column as a Postgresql `ltree` type to the `categories` table in order to more easily query trees of categories and subcategories. This allows navigating trees of arbitrary depth and potentially more ergonomic path-finding queries.

- Adds a `parent_categories` method to `Category`, which returns parent categories _in order of traversal_ from root to this node.

### Pros

- Easily handle trees of arbitrary depth, ex: "category::sub-category::sub-catgeory-A::sub-category-A-1".
- Makes queries to find a whole tree of nodes simpler (and possibly quicker?). Ex:
  `select slug, path from categories where path <@ 'root.web_programming';`
- Makes finding all root-level nodes simpler (assuming one is cool with the syntax). Ex:
  `select slug, path from categories where path ~ 'root.*{1}';`
- Makes finding the direct path from the root node to a particular node easier (see `parent_categories` method). Ex:
  `select slug, path from categories where path @> 'root.web_programming.http_client' order by path;`
- Invisible to current application code: nothing needs to be updated to handle the `path` column because it happens via database triggers. (This could also be a _con_ depending on one's perspective.)

### Cons

- Postgresql only (harder to switch databases in the future)
- Makes certain characters in slugs disallowed: `[',",-]`. Path labels must consist of "A-Za-z0-9_" separated by "." We swap `-` for `_` and `::` for `.` in the included Postgresql procedure.
- This `path` column is nowhere in application code, so this is all hand-written SQL (but the current solution is doing that as well).
- Error messages are opaque if you happen to have a bad character in a slug:
```
cargo_registry=# update categories set slug = 'test-new::bla$bla-::parent::child' where slug = 'test-new::parent::child';
ERROR:  syntax error at position 17
CONTEXT:  PL/pgSQL function set_category_path_to_slug() line 3 at assignment
```

## Notes

- [x] Needs a test for the `parent_categories` method.


src/category.rs
```diff
@@ -161,6 +162,24 @@ impl Category {
         ).bind::<Text, _>(&self.category)
             .load(conn)
     }
+
+    pub fn parent_categories(&self, conn: &PgConnection) -> QueryResult<Vec<Category>> {
```
vignesh-sankaran: Could you add a doc comment here outlining what this function does??

erewok: You know I thought I had done that but I forgot about the three-slashes thing. Sorry about that.

kivikakk:
I've now adapted #1130 into this PR! Parent category traversal is now easier (and accounts for arbitrary category tree depth).

carols10cents:
Sorry this has taken me so long to look at-- my first impression is WHERE HAS LTREE BEEN ALL MY LIFE?!!?!? Like seriously, is this new? If not, I am irrationally angry at everyone i've interacted with who has known about this feature and not told me about it. I'm now going to start telling people as soon as I meet them-- "Hi, I'm Carol, have you heard about Postgres' ltree?"

Digging into the code and trying this out now :)

erewok:
@carols10cents `ltree` has been part of Postgresql since at least 2004, I think. I only found out about in the last couple of years and I have been surprised how rarely people recommend it or even know about it because it seems like a surprisingly useful data type. My theory is that Postgresql offers so many different options that it's difficult to actually know about all of them. For instance, I'm only aware of a handful of the myriad index types and I feel like I don't hear mention of those either.

The syntax, perhaps, takes a bit to get used to, but I find it pretty ergonomic myself: 

`@> some_path`: means a path that's greater than or equal to *some_path*.
`<@ some_path`: means a path that's less than or equal to *some_path*.
`~ some_path`: looks sort of like a regex comparison, usually?

In this PR, I added `root.` to all the constructed paths, and this makes path-finding a whole lot easier as long as no one ever adds a new category with a slug called `root`!

migrations/2017-10-08-193512_category_trees/up.sql
```diff
@@ -0,0 +1,40 @@
...
+UPDATE categories
+    SET path =  text2ltree('root.' || trim(replace(replace(slug, '-', '_'), '::', '.')))
+    WHERE path is NULL;
+
+ALTER TABLE CATEGORIES
+    ALTER COLUMN path SET NOT NULL;
+
+-- Create some indices that allow us to use GIST operators: '@>', etc
+CREATE INDEX path_gist_categories_idx ON categories USING GIST(path);
+CREATE INDEX path_categories_idx ON categories USING btree(path);
```
carols10cents: Is there a reason to have both a gist and a btree index? The [postgres docs](https://www.postgresql.org/docs/9.6/static/ltree.html) say:

> F.21.3. Indexes
> ltree supports several types of indexes that can speed up the indicated operators:
>
> * B-tree index over ltree: <, <=, =, >=, >
> * GiST index over ltree: <, <=, =, >=, >, @>, <@, @, ~, ?

which looks like gist indexes take care of the same operators that btree indexes take care of? 

All of this is probably moot anyway since the categories table is so smalll :)

app/templates/category/index.hbs
```diff
@@ -1,8 +1,11 @@
 {{ title category.category ' - Categories' }}
 
 <div id='crates-heading'>
-    {{svg-jar "crate"}}
-    <h1>{{ category.category }}</h1>
+    {{#link-to "categories"}}{{svg-jar "crate"}}{{/link-to}}
```
carols10cents: Neat easter-egg :)

migrations/2017-10-08-193512_category_trees/up.sql
erewok: Honestly, I've just seen everyone do it that way, so I figured it was a best practice, if not necessary.

erewok:
thanks @carols10cents and thanks to @kivikakk for working on the frontend.

One thing I considered was rewriting the other queries that do `split("::", 1)` in order for them to be all be consistent. I didn't do that because I wasn't sure if there would be buy-in for these changes. I knew there would be an issue with diesel, and I didn't want to rock the boat too much, even though I felt like this would be a useful direction to go in.

sgrif:
You can just delete the `path` column from `schema.rs`, or create `struct Ltree;` somewhere and `use` it at the top of the table declaration.

kivikakk:
It looks like `print-schema` won't automatically do what we want with extensions (going by this and #1133).

I've created a super basic `diesel_ltree` crate with an `Ltree` struct that does what we want (i.e. almost nothing right now), and manually added it to `schema.rs` in b7c49f8e9f.

kivikakk:
@erewok 

> One thing I considered was rewriting the other queries that do `split_part(some_value, "::", 1)` in order for them to be all be consistent. I didn't do that because I wasn't sure if there would be buy-in for these changes. I knew there would be an issue with diesel, and I didn't want to rock the boat too much, even though I felt like this would be a useful direction to go in.

[`diesel_ltree`](https://crates.io/crates/diesel_ltree) now supports the common operators and functions on ltrees/lqueries, so we could indeed redo those idiomatically from here on out.

sgrif:
@kivikakk `diesel_ltree` needs to use http://docs.diesel.rs/diesel/pg/struct.PgMetadataLookup.html#method.lookup_type for its `HasSqlType` impls. Types from extensions *do not* have stable OIDs. I suspect that the only reason this isn't breaking things is because you aren't currently dealing with bind params of those types

kivikakk:
Using latest ltree which fixes this. :+1: Thanks sgrif.

kivikakk:
@carols10cents This is brought up to speed with `master`; would you mind taking another look? Just got an accessibility violation to clear up.

src/category.rs
```diff
@@ -156,6 +157,26 @@ impl Category {
...
+    /// Gathers the parent categories from the top-level Category to the direct parent of this Category.
+    /// Returns categories as a Vector in order of traversal, not including this Category.
+    /// The intention is to be able to have slugs or parent categories arrayed in order, to
+    /// offer the frontend, for examples, slugs to create links to each parent category in turn.
+    pub fn parent_categories(&self, conn: &PgConnection) -> QueryResult<Vec<Category>> {
+        use diesel::sql_types::Text;
+
+        sql_query(
+            "SELECT c.id, c.category, c.slug, c.description, \
+            COALESCE(( \
```
sgrif: Is this `COALESCE` actually doing anything?

src/category.rs
```diff
@@ -156,6 +157,26 @@ impl Category {
...
+
+    /// Gathers the parent categories from the top-level Category to the direct parent of this Category.
+    /// Returns categories as a Vector in order of traversal, not including this Category.
+    /// The intention is to be able to have slugs or parent categories arrayed in order, to
+    /// offer the frontend, for examples, slugs to create links to each parent category in turn.
+    pub fn parent_categories(&self, conn: &PgConnection) -> QueryResult<Vec<Category>> {
+        use diesel::sql_types::Text;
+
+        sql_query(
+            "SELECT c.id, c.category, c.slug, c.description, \
```
sgrif: Can we move this whole thing into a `.sql` file and use `include_str!`, add some more newlines and indentation? The whole thing is very hard to parse. It's not obvious where the call to `COALESCE` ends for example

src/category.rs
```diff
@@ -398,4 +424,30 @@ mod tests {
         let expected = vec![("Cat 3".to_string(), 6), ("Cat 1".to_string(), 3)];
         assert_eq!(expected, categories);
     }
+
+    #[test]
+    fn category_parent_categories_includes_path_to_node_with_count() {
+        let conn = pg_connection();
+        conn.batch_execute(
```
sgrif: Can we use Diesel's public API for this (I need to update the other tests). You can write this as

```
insert_into(categories)
    .values(vec![
        (category.eq("Cat 1"), slug.eq("cat1"), crates_cnt.eq(1)),
        ...
    ])
```

src/lib.rs
```diff
@@ -5,7 +5,7 @@
 //! [krate](krate/index.html), [user](user/index.html) and [version](version/index.html) modules.
 #![deny(warnings)]
 #![deny(missing_debug_implementations, missing_copy_implementations)]
-#![recursion_limit = "128"]
+#![recursion_limit = "256"]
```
sgrif: Why was this necessary?

kivikakk: Without:

```
error: recursion limit reached while expanding the macro `table_body`
   --> src/schema.rs:223:1
    |
223 | / table! {
224 | |     use diesel_full_text_search::{TsVector as Tsvector};
225 | |     use diesel_ltree::Ltree;
226 | |     use diesel::sql_types::*;
...   |
322 | |     }
323 | | }
    | |_^
    |
    = help: consider adding a `#![recursion_limit="256"]` attribute to your crate
    = note: this error originates in a macro outside of the current crate

error: Could not compile `cargo-registry`.
```

src/category.rs
kivikakk: Done this (and fixed up the other tests too).

src/category.rs
kivikakk: :+1: Done.

kivikakk:
```
error[E0063]: missing field `parent_categories` in initializer of `category::EncodableCategoryWithSubcategories`
   --> src/category.rs:526:19
    |
526 |         let cat = EncodableCategoryWithSubcategories {
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `parent_categories`
```

There is no line 526 of `category.rs`? The same line where it _actually_ exists does have a `parent_categories`? Travis what are you doing

erewok:
nice work @kivikakk. I was actually wondering about this PR the other day. Thank you for continuing to push it forward.

src/models/category.rs
```diff
@@ -93,43 +93,42 @@ impl Category {
...
         // the crates in all subcategories
-        select(sql::<categories::SqlType>(&format!(
-            "c.id, c.category, c.slug, c.description,
-                sum(c2.crates_cnt)::int as crates_cnt, c.created_at
-             FROM categories as c
-             INNER JOIN categories c2 ON split_part(c2.slug, '::', 1) = c.slug
-             WHERE split_part(c.slug, '::', 1) = c.slug
-             GROUP BY c.id
+        sql_query(format!(
+            "SELECT c.id, c.category, c.slug, c.description, \
```
sgrif: How would you feel about moving this into a separate file and `include_str!`ing it. We can also change this to call `.bind` rather than using string interpolation (other than the order by)

kivikakk: :+1:

kivikakk:
Thank you for the ✅, sgrif!

Let me know if there's anything else to be done to get this merged, if we're still happy with it.

carols10cents:
Thank you @erewok and @kivikakk for keeping this up to date, and sorry for the delay!

I tested this out and it looks like it's working great in the UI!

Buuuuut I noticed in the schema.rs that path was added as a column on the crates table, but if anything, it should be added to the categories table... but if I add it to the categories table by running `diesel print-schema --with-docs > src/schema.rs` and then manually adding `use diesel_ltree::Ltree;` within the `table!` call for categories (i.e. [this commit](https://github.com/rust-lang/crates.io/pull/1122/commits/fb335939b98f9e0646176bfe32b49b4553161bdd)) I then get errors, a bunch like this:

```
error[E0412]: cannot find type `Int4` in this scope
  --> src/schema.rs:83:15
   |
83 |         id -> Int4,
   |               ^^^^ did you mean `Into`?
```

So then I added `use diesel::sql_types::*;` [in this commit](https://github.com/rust-lang/crates.io/pull/1122/commits/60ef7ee2025471c16cdf0677d092c9323ff36f1c) to be more like the crates `table!`, which has custom types and works.

That gets rid of ~15 errors, but leaves me with 4 that look like this:

```
error[E0277]: the trait bound `(i32, std::string::String, std::string::String, std::string::String, i32, chrono::NaiveDateTime): diesel::Queryable<(diesel::sql_types::Integer, diesel::sql_types::Text, diesel::sql_types::Text, diesel::sql_types::Text, diesel::sql_types::Integer, diesel::sql_types::Timestamp, diesel_ltree::Ltree), _>` is not satisfied
  --> src/controllers/category.rs:42:10
   |
42 |         .first::<Category>(&*conn)?;
   |          ^^^^^ the trait `diesel::Queryable<(diesel::sql_types::Integer, diesel::sql_types::Text, diesel::sql_types::Text, diesel::sql_types::Text, diesel::sql_types::Integer, diesel::sql_types::Timestamp, diesel_ltree::Ltree), _>` is not implemented for `(i32, std::string::String, std::string::String, std::string::String, i32, chrono::NaiveDateTime)`
```

which, ok, the `Category` struct that's `Queryable` doesn't have the `path` column specified.

I tried adding `path` as type `LTree` in [this commit](https://github.com/rust-lang/crates.io/pull/1122/commits/be7fba190da706774e5640b19260a3435d3a952c), but then I get these errors:

```
error[E0277]: the trait bound `diesel_ltree::Ltree: diesel::Queryable<diesel_ltree::Ltree, diesel::pg::Pg>` is not satisfied
  --> src/controllers/category.rs:42:10
   |
42 |         .first::<Category>(&*conn)?;
   |          ^^^^^ the trait `diesel::Queryable<diesel_ltree::Ltree, diesel::pg::Pg>` is not implemented for `diesel_ltree::Ltree`
   |
```

The [tests in diesel_ltree use `Queryable` with the `path` column as type `String`](https://github.com/kivikakk/diesel_ltree/blob/master/src/tests.rs#L25), so I tried that too in [this commit](https://github.com/rust-lang/crates.io/pull/1122/commits/5986ee0c7d6f0953b1bf6bba18ce8c6d8e3fd63f), but then I get these errors:

```
error[E0277]: the trait bound `*const str: diesel::deserialize::FromSql<diesel_ltree::Ltree, diesel::pg::Pg>` is not satisfied
  --> src/controllers/category.rs:42:10
   |
42 |         .first::<Category>(&*conn)?;
   |          ^^^^^ the trait `diesel::deserialize::FromSql<diesel_ltree::Ltree, diesel::pg::Pg>` is not implemented for `*const str`
   |
   = help: the following implementations were found:
             <*const [u8] as diesel::deserialize::FromSql<diesel::sql_types::Binary, DB>>
             <*const str as diesel::deserialize::FromSql<diesel::sql_types::Text, DB>>
```

soooo @sgrif @kivikakk am I using diesel or diesel_ltree wrong? or is there something incompatible/missing here?

carols10cents:
Ok, I think I've successfully figured this one out. It worked when the path column was accidentally on the crates table because the crates code was already explicitly selecting fewer than all columns; when the path column was added to the category table in the schema instead, the same don't-select-all-columns code needed to happen for category too.

@sgrif if you have a minute, could you review the last commit of mine and make sure i did this right?

kivikakk:
It'd be nice if we could land this!

carols10cents:
Yes, yes it would. YOLO

bors: r+

carols10cents:
hMMMM why didn't that work....

bors: r+

carols10cents:
there we go. good bot!

carols10cents:
hmmmm looks like rustfmt had differences and clippy failed to install? i get different results when running rustfmt locally, fun. we'll see if this fixes it

carols10cents:
ok let's try this one more time!

bors: r+

carols10cents:
You're welcome! I'll try and get this deployed tomorrow :)

erewok:
Wow, nice work all. I completely forgot about this PR (sorry about that) and found it just now in my github unread.

PR #1128, opened 2017-10-16
Remove some vestigial yarn mentions

carols10cents:
neunenak found some yarn stuff still hanging around!! oops!!


kivikakk:
This momentarily confused me while trying to get setup again shortly ago; lovely to see I'm now not in fact doing it wrong by using npm!

PR #1129, opened 2017-10-16
Update comrak to 0.2.3, use ext_header_ids

### LINKED ISSUE #1037: Anchor links on READMEs don't work
GitHub automatically adds anchor links to header tags, allowing links to sections inside of a README. It would be nice if this also worked on crates.io.

Compare clap on [crates.io](https://crates.io/crates/clap) to [GitHub](https://github.com/kbknapp/clap-rs).

kivikakk:
Fixes #1037. Use latest comrak, and use the header IDs extension to generate GitHub-compatible anchors.

/cc @kejadlen @treiff

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -1,4 +1,4 @@
-use ammonia::Ammonia;
+use ammonia::Builder;
 use comrak;
 
 use util::CargoResult;
@@ -6,7 +6,7 @@ use util::CargoResult;
 /// Context for markdown to HTML rendering.
 #[allow(missing_debug_implementations)]
 pub struct MarkdownRenderer<'a> {
-    html_sanitizer: Ammonia<'a>,
+    html_sanitizer: Builder<'a>,
 }
 
 impl<'a> MarkdownRenderer<'a> {
@@ -53,8 +53,7 @@ impl<'a> MarkdownRenderer<'a> {
             .cloned()
             .collect();
         let tag_attributes = [
-            ("a", ["href", "target"].iter().cloned().collect()),
-            ("code", ["class"].iter().cloned().collect()),
+            ("a", ["href", "id", "target"].iter().cloned().collect()),
             (
                 "img",
                 ["width", "height", "src", "alt", "align"]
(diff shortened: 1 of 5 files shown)


kivikakk:
It occurs to me this won't _quite_ fix #1023; the anchors still won't work because we prefix `"user-content-"` to IDs to prevent nasty name collisions and subsequent security issues, but we need to still support `href`s without this prefix (because users don't expect to see `#user-content-…` in their address bar).

I'll come up with an appropriate solution shortly; on GitHub we use JavaScript which fires on hash change to search for and scroll to the item with the `"user-content-"` prefix.

kivikakk:
So, to prevent ID collisions, we'll also need to sanitize the IDs not coming from comrak-outputted header tags; I've proposed https://github.com/notriddle/ammonia/pull/66 to give us the machinery to do that.

Finally, we'll need either the JavaScript portion to scroll the appropriate tag into view (I could throw this together easily based on what we're using here), or we'd need to change all relative `href`s. The former is a bit prettier since it means we don't end up with `"user-content-"` or similar in the address bar.

kivikakk:
The JavaScript side of the solution is now in this PR; ready for your thoughts/comments!

src/render.rs
```diff
@@ -53,8 +53,7 @@ impl<'a> MarkdownRenderer<'a> {
             .cloned()
             .collect();
         let tag_attributes = [
-            ("a", ["href", "target"].iter().cloned().collect()),
-            ("code", ["class"].iter().cloned().collect()),
```
carols10cents: hm, does `allowed_classes` make this line unnecessary...? I guess so since the tests pass!

carols10cents:
Looks great, thank you!!! I'm going to review your other readme-related PRs then deploy and regenerate the readmes all at once, that'll probably take me a few days :) Thank you so much for taking care of all these things!!!

bors: r+

src/render.rs
kivikakk: Indeed, it even became required in the newer `ammonia`: https://github.com/notriddle/ammonia/blob/2581eec04ef98419fd202e0dd4d8997c9dfdb06e/src/lib.rs#L1030-L1037 (you're allowed to use `allowed_classes` or set `class` as an allowed tag, but not both)

kivikakk:
Thanks so much!

PR #1130, opened 2017-10-16
Link to parent category from category page

kivikakk:
Also link to "All Categories" on the cute SVG crate. Fixes https://github.com/rust-lang/crates.io/issues/1093.

/cc @aturon @erewok as people who've commented over there!

### DIFF
--- a/app/models/category.js
+++ b/app/models/category.js
@@ -8,6 +8,7 @@ export default DS.Model.extend({
     crates_cnt: DS.attr('number'),
 
     subcategories: DS.attr(),
+    parent: DS.attr(),
 
     crates: DS.hasMany('crate', { async: true })
 });
--- a/app/templates/category/index.hbs
+++ b/app/templates/category/index.hbs
@@ -1,8 +1,11 @@
 {{ title category.category ' - Categories' }}
 
 <div id='crates-heading'>
-    {{svg-jar "crate"}}
-    <h1>{{ category.category }}</h1>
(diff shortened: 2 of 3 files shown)


erewok:
One of the things I was unsure about is whether we're going to have arbitrary levels in the hierarchy of categories. In other words, do we wish to be able to easily store and represent relationships like this:

    - Parent
         - subcategory
              - subcategory-child
                  - etc 

As a result, I looked at [changing the database representation of categories](https://github.com/rust-lang/crates.io/pull/1122) and then for any particular category I included a method to return all ancestor categories in order of traversal, so that we could send that ordered array of categories to any particular category page. (I didn't include any changes for the frontend, though).

I'm not committed to my solution, however,  In particular, if the categories are always going to be `parent -> child`, then these changes should be fine. We could also try to use both.

kivikakk:
I'm a big fan of your PR! I'll try to work my changes on top of yours. 👍

erewok:
@kivikakk That sounds fine to me. Do you want commit access to my fork so you can push to that branch? I added you as a collaborator in case you want to do it that way.

kivikakk:
@erewok Sure thing, thank you! I'll push it there once I feel like I've gotten it right & will @mention you when I do. :)

erewok:
@kivikakk Okay, feel free to take it in whatever direction you think is appropriate. If you need any help, let me know and if it doesn't work out, no big deal!

kivikakk:
This is rolled into #1122, and that's (in my opinion) the better lead to take here.

Issue #974, opened 2017-08-19
Rendered intra-project README links are broken


alilleybrinker:
This issue can be seen here, with the "`examples/`" link: https://crates.io/crates/finchers

Basically, GitHub allows links to folders and files within the repo, and those links don't work on crates.io. This seems pointy, although I am not sure how best to resolve it.

joshtriplett:
As a first pass, transforming relative URLs to URLs relative to the repository might make sense.

kivikakk:
This is tricky; it won't work well in 100% of cases because a repository URL might be something like <<https://github.com/kivikakk/blah>>, but the URL to a file in that repository is like <<https://github.com/kivikakk/blah/blob/master/file.md>>.

We can assume that the main branch is `master` and so just always prepend `{repository}/blob/master/` and we'll get the 95% case fixed, which is way better than the current situation (a bunch of broken links).

kivikakk:
A proposal implementing the above is at https://github.com/rust-lang/crates.io/pull/1131. There's an ammonia-related caveat I'll see if I can get support from upstream for a fix.

kivikakk:
This should now be fixed; the example at <<https://crates.io/crates/finchers>> will be once we rerender READMEs.

benoittgt:
I have a weird behavior on https://crates.io/crates/git-changelog
Links in the [original readme](https://github.com/aldrin/git-changelog) are written like this:
```markdown
A commit [like this](src/assets/sample-commit.message)
```
github converts them as 
```html
<p>A commit <a href="/aldrin/git-changelog/blob/master/src/assets/sample-commit.message">like this</a>
this</a>.</p>
```

creates.io generates
```
<a href="https://github.com/aldrin/git-changelog/blob/master/resources/sample-commit.message" rel="nofollow noopener noreferrer">like this</a>
```

carols10cents:
Hi @benoittgt, could you open a new issue please?

carols10cents:
Actually nevermind, I saw the comment saying "this should be fixed" and thought this issue was closed. This is fine!

kivikakk:
I'll have a look into this shortly!

kivikakk:
I've tested locally that the following test passes:

```rust
    #[test]
    fn relative_link_test() {
        assert_eq!(
[...]

PR #1131, opened 2017-10-16
Relative links in README

kivikakk:
This is a proposed fix for #974.

Currently implemented is GitHub/GitLab/Bitbucket-style (insert `/blob/master/` between the repository and the link destination), with a caveat.


src/render.rs
```diff
@@ -11,7 +11,7 @@ pub struct MarkdownRenderer<'a> {
 
 impl<'a> MarkdownRenderer<'a> {
     /// Creates a new renderer instance.
-    pub fn new() -> MarkdownRenderer<'a> {
+    pub fn new(repo: Option<&'a str>) -> MarkdownRenderer<'a> {
```
kureuil: Should this parameter be renamed to something like `base_url` to better reflect its usage ?
Also, its usage should be documented.

src/render.rs
```diff
@@ -139,10 +139,13 @@ impl<'a> Default for MarkdownRenderer<'a> {
 /// use render::markdown_to_html;
 ///
 /// let text = "[Rust](https://rust-lang.org/) is an awesome *systems programming* language!";
-/// let rendered = markdown_to_html(text)?;
+/// let rendered = markdown_to_html(text, None)?;
 /// ```
-pub fn markdown_to_html(text: &str) -> CargoResult<String> {
-    let renderer = MarkdownRenderer::new();
+pub fn markdown_to_html(text: &str, repo: Option<&str>) -> CargoResult<String> {
```
kureuil: Can you add a note in the documentation about the intended usage & the expected format of the new parameter?

src/render.rs
```diff
@@ -139,10 +139,13 @@ impl<'a> Default for MarkdownRenderer<'a> {
...
+    let repo = repo.map(|r|
+        format!("{}{}blob/master/", r, if r.ends_with("/") { "" } else { "/" }));
```
kureuil: We already depend on the `url` crate, should we use it to check that the given `repo` url's scheme is either `http` or `https`? I think it is possible for someone to submit a crate with a repository link using the `git://...` scheme.

Also, would it be worth it to check that the hosts are either `github.com`, `gitlab.com` or `bitbucket.org` ?

kivikakk: > should we use it to check that the given `repo` url's scheme is either `http` or `https`?

I've just looked into it; I think we're safe — the API already checks to ensure a `repository` has either `http` or `https`, and by the documentation [on crates.io](http://doc.crates.io/manifest.html#package-metadata) this is expected:

> These are intended to be webviews of the relevant data, not necessarily compatible with VCS tools and the like.

Here's the relevant code:

https://github.com/rust-lang/crates.io/blob/ef1ac581f9c4ea6a906347d43776418acfdbb8ca/src/krate/mod.rs#L183-L219

I think it's a good idea to check the host; in case it doesn't match any, I feel like we should just drop relative links.

src/render.rs
kivikakk: Will do!

kivikakk:
Finally got a rustfmt nightly matching that of crates.io's Travis config going, so this last push should be all greens.

Cargo.toml
```diff
@@ -55,7 +55,7 @@ serde = "1.0.0"
 clippy = { version = "=0.0.162", optional = true }
 chrono = { version = "0.4.0", features = ["serde"] }
 comrak = { version = "0.1.9", default-features = false }
-ammonia = "0.7.0"
+ammonia = { git = "https://github.com/notriddle/ammonia" }
```
kureuil: Shouldn't this dependency be specified as `{ git = "https://github.com/notriddle/ammonia", rev = "46054ff" }` to ensure reproducible builds ?

src/render.rs
```diff
@@ -224,19 +262,30 @@ mod tests {
...
+        let absolute = "[hi](/hi)";
         let relative = "[there](there)";
 
-        for url in &["https://github.com/rust-lang/test",
-                     "https://github.com/rust-lang/test/"] {
-            // let result = markdown_to_html(absolute, Some(url)).unwrap();
-            // assert_eq!(result, "<p><a href=\"https://github.com/rust-lang/test/blob/master/hi\" rel=\"nofollow noopener noreferrer\">hi</a>");
+        for url in &[
+            "https://github.com/rust-lang/test",
+            "https://github.com/rust-lang/test/",
```
kureuil: Should we have gitlab.com & bitbucket.org tests in here to avoid regressions ?

Cargo.toml
kivikakk: Good idea! I was thinking `Cargo.lock` storing the rev would be sufficient, but this is better.

src/render.rs
kivikakk: Yes, let's.

kivikakk:
Done! Thank you so much for your feedback. :heart:

carols10cents:
@kivikakk I think the resolving is being a bit *too* aggressive now, it's resolving absolute URLs too-- I added a test case that's currently failing to illustrate:

```
---- render::tests::more_relative_links stdout ----
	thread 'render::tests::more_relative_links' panicked at 'assertion failed: `(left == right)`
  left: `"<p><a href=\"https://github.com/kbknapp/clap-rs/blob/master/https://crates.io/crates/clap\" rel=\"nofollow noopener noreferrer\"><img src=\"https://github.com/kbknapp/clap-rs/blob/master/https://img.shields.io/crates/v/clap.svg\" alt=\"Crates.io\"></a></p>\n"`,
 right: `"<a href=\"https://crates.io/crates/clap\" rel=\"nofollow noopener noreferrer\"><img src=\"https://img.shields.io/crates/v/clap.svg\" /></a>\n"`', src/render.rs:309:8
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

And I was also wondering how this interacts with the anchor support in your other PR-- anchors shouldn't be resolved, correct? Having a test case for that would be good too :)

Other than those two concerns, this is looking good!!

kivikakk:
@carols10cents Thank you so much for the test case! What a dork I am. I'll add the extra test case as you suggested and fix. :+1:

kivikakk:
Have resolved conflicts; looks like the overzealous/absolute URL application is possibly an upstream bug, as the custom `UrlRelativeEvalutate` shouldn't be called on absolute URLs. Looking into it!

kivikakk:
-> https://github.com/notriddle/ammonia/pull/77

kivikakk:
Ammonia fix merged upstream and this branch now cleaned up. :) An anchor test was already in `master` (from the `user-content-` prefixing test) so we're all good there.

carols10cents:
Thanks!!! Working great now!!!

bors: r+

PR #1133, opened 2017-10-16
Update database schema to match the code

sgrif:
There are a handful of differences between what we say our schema is in
the code, and what it actually is. These differences make schema.rs
harder to regenerate, and can potentially lead to real issues down the
line. This removes all of the differences between the real schema and
what we say it is in Rust.


carols10cents:
So this is better, but when I'm on this branch and run `diesel setup` and `diesel print-schema --with-docs > src/schema.rs`, the `use` lines added here are just removed again :-/ So this seems *better* as far as ease of regeneration, but not quite ideal...

sgrif:
Yeah, there's not really any way for `diesel print-schema` to know exactly what we've changed. As I mentioned on another PR, I think it might be worth us keeping a `.patch` file around for any changes that we do make to this file to make it easier to regenerate, but I don't think that should block this PR.

carols10cents:
Ok, I'll think about how to handle the other issues and merge this in the meantime.

bors: r+

Issue #1025, opened 2017-08-30
Prerender only max version of crates' readmes, lazily render/cache older versions' as requested


carols10cents:
It takes a long time to rerender all the readmes for all the versions on crates.io :( 

I think a way to cut down on this time by a lot is just to render fewer versions upfront, and render the rest on-demand. What I'm envisioning is:

- Change the readme rendering to be proxied through crates.io's backend rather than served from s3 directly. That way, we can check to see if it exists on s3 first, and try and render and serve it if it doesn't (hmmmm we don't want to keep trying to render readmes that are missing over and over, hmmmm)

Once the proxying/request rendering if it isn't on s3 yet, the render-readmes script should:

- Remove all rendered readmes saved on s3
- For each crate, render and save to s3 the readme for the latest version semvertically (that is, whatever version shows up when you go to crates.io/crates/cratename, not necessarily the most recently published version)

As new versions are published, the publish process will render their readme to s3. The next time we need to rerender readmes, old versions will get cleared out.

Turbo87:
while the faster rerendering would be nice, I'm not sure whether that alone is worth the additional complexity and potential performance impact. the rendered readmes are currently served through a globally distributed CDN, but if we routed all those requests through the API we'd increase the latency unnecessarily. the rerendering might be slow but since we don't need it that often I think this is probably fine?

PR #1134, opened 2017-10-17
Record readme filename; use to determine whether to render as Markdown or not

### LINKED ISSUE #995: Only interpret README files whose filename ends in `.md` as markdown, render plain text otherwise
Hello.
If the readme is in the asciidoc(tor) format, it is not rendered correctly on crates.io.
[Here's an example.](https://crates.io/crates/relm)
Thanks to fix this.

kivikakk:
Fixes #995. Depends on https://github.com/rust-lang/cargo/pull/4633 to have the desired effect, but won't break in the absence of that PR; we'll just assume the readme is called `"README.md"` if the client doesn't report one.

Small CSS fix included to wrap extra long lines.

Comparison of render of https://crates.io/crates/relm's README:

| Before | After |
| :-: | :-: |
| screen shot 2017-10-17 at 5 51 50 pm | screen shot 2017-10-17 at 5 51 51 pm |


src/render.rs
```diff
@@ -141,11 +142,31 @@ impl<'a> Default for MarkdownRenderer<'a> {
 /// let text = "[Rust](https://rust-lang.org/) is an awesome *systems programming* language!";
 /// let rendered = markdown_to_html(text)?;
 /// ```
-pub fn markdown_to_html(text: &str) -> CargoResult<String> {
+fn markdown_to_html(text: &str) -> CargoResult<String> {
```
kureuil: If we drop the `pub` modifier on `markdown_to_html`, we probably want to drop it for `struct MarkdownRenderer` too ?

src/render.rs
```diff
@@ -141,11 +142,31 @@ impl<'a> Default for MarkdownRenderer<'a> {
...
     let renderer = MarkdownRenderer::new();
     renderer.to_html(text)
 }
 
+static MARKDOWN_EXTENSIONS: [&'static str; 7] = [
```
kureuil: Even though this is not public & pretty self-explanatory, we might want to add a small doc comment to this constant.

src/render.rs
```diff
@@ -141,11 +142,31 @@ impl<'a> Default for MarkdownRenderer<'a> {
...
+    ".md",
+    ".markdown",
+    ".mdown",
+    ".mdwn",
+    ".mkd",
+    ".mkdn",
+    ".mkdown",
+];
+
+pub fn readme_to_html(text: &str, filename: &str) -> CargoResult<String> {
```
kureuil: Could you add a doc comment on this function ?

src/render.rs
```diff
@@ -141,11 +142,31 @@ impl<'a> Default for MarkdownRenderer<'a> {
...
+    for e in &MARKDOWN_EXTENSIONS {
+        if filename.to_lowercase().ends_with(e) {
+            return markdown_to_html(text);
+        }
+    }
+
+    Ok(encode_minimal(text).replace("\n", "<br>\n"))
```
kureuil: Why are we pulling `htmlescape` as a new dependency ? Isn't `ammonia` already doing a similar job ?

src/krate/mod.rs
```diff
@@ -65,6 +65,7 @@ pub struct Crate {
     pub homepage: Option<String>,
     pub documentation: Option<String>,
     pub readme: Option<String>,
+    pub readme_file: Option<String>,
```
kureuil: I see that we're now storing the README filename alongside its content, though it doesn't seem to be used anywhere ? I'm not opposed to storing it but I'm not sure to see how useful it is to store it in the database.

src/render.rs
kivikakk: `htmlescape` is replacing characters with their entities in order to basically turn text into HTML that represents/renders as the same text; `ammonia` is parsing HTML itself and then sanitizing that DOM. They're pretty different (and unfortunately, I looked and neither `ammonia` nor any of its dependencies, like, `html5ever`, can do the job of `htmlescape` simply.)

src/render.rs
kivikakk: Whoops, forgot!

src/render.rs
kivikakk: +1

src/render.rs
kivikakk: Yes!

src/krate/mod.rs
kivikakk: You're right, we don't strictly need to store it. I was thinking we might need it in `render-readmes`, but that pulls the README filename from the manifest directly. However, I feel like storing this as well might be important for things like #1025/#1019?

A more general argument is that, if there's no need to store the README filename in the database, then there's no need to store the README _content_ in the database, because we can't properly make sense of the latter without the former.

That's moreorless the motivation of this PR in a nutshell: we don't know how to interpret the README contents without knowing its filename. So if we store the content, we should store the filename too.

kureuil: Yeah, you're totally right about needing to store the filename alongside the README content, my bad.

However, maybe we can improve a little here and instead of storing the README filename as transmitted, how about storing the normalized README type ? i.e only `markdown` for now. This would also allow us to use an `ENUM` instead of a `VARCHAR` field in the database and simplify its usage.

src/render.rs
kureuil: Fine :)

src/krate/mod.rs
kivikakk: I like the idea of doing this, but I'm worried about the loss of fidelity. e.g. say we do so, and currently store `'markdown'` or `NULL` for everything. Later we add AsciiDoc or org-mode support — it'd be great to go back and rerender the newly supported readmes, but to do so we'd have to reach back inside each crate and retrieve the readme filename from the manifest again.

I think storing the filename directly allows us more flexibility later on in how we want to treat certain things, especially in the face of changing/new standards and formats; this is how we approach file classification at GitHub, for example, and it works well for us.

kureuil: > but to do so we'd have to reach back inside each crate and retrieve the readme filename from the manifest again.

In any case, we probably will have to do it. The `crates` table only stores the README of the last published version of a given crate, so if we add org-mode or AsciiDoc support to crates.io we'd still have to reach back inside each crate in order to render it. Anyway, you're right about your approach being more flexible so let's keep it this way :)

sgrif:
I think we should still assume that `README` with no extension is markdown here. Other than that this looks fine.

kivikakk:
@sgrif That works for me! Change made.

carols10cents:
Looks great and works great with the cargo change, and the current behavior still works with stable cargo! Just merged in upstream to resolve the merge conflicts :)

bors: r+

Issue #1058, opened 2017-09-15
Crates who have yanked all their versions should only return name, owners, and that it's yanked


carols10cents:
In search results and on the crate's page, we should be displaying crates with all of their versions yanked (henceforth referred to as "yanked crates") differently (https://github.com/rust-lang/crates.io/issues/145), but the backend needs to tell the frontend that a crate is in this state.

Instructions for fixing this:

- [ ] Add a field named `all_versions_yanked` to [`EncodableCrate`](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/krate.rs#L106) and everywhere we create `EncodableCrate` instances, set that field to false if there exists any non-yanked versions for this crate. I'd recommend doing a `select(exists())` query using diesel similar to [this query](https://github.com/rust-lang/crates.io/blob/76d15c1a909ddcbf23413a5a3ad477874baca859/src/krate.rs#L244)
- [ ] If we're constructing an `EncodableCrate` instance for a yanked crate, these fields should be changed too:
  - `badges`, `description`, `homepage`, `documentation`, and `repository` should be `None`
  - `max_version` should be `"0.0.0"`

Some other fields, like keywords and categories, are used for search/crate lists, and this issue isn't going to change search results, so I think we should still display why a crate showed up in a particular list of crates.

seanlinsley:
I'd like to work on this. It'll be my first Rust contribution! Thanks for making it so accessible to new people :octocat: ❤️ I actually got here from the RustConf keynote on YouTube.

carols10cents:
Hi @seanlinsley!!! It's awesome to have you!! Please let me know if you have any questions ❤️

seanlinsley:
When running `diesel migration run` I noticed this error:

```
NOTICE:  table "schema_migrations" does not exist, skipping
```

The backend server appears to boot, though some tables are missing when running the tests:
```rust
---- category::tests::category_toplevel_includes_subcategories_in_crate_cnt stdout ----
  thread 'category::tests::category_toplevel_includes_subcategories_in_crate_cnt' panicked at 'called `Result::unwrap()` on an `Err` value: DatabaseError(__Unknown, "relation \"categories\" does not exist")', src/libcore/result.rs:860:4

---- category::tests::category_toplevel_applies_limit_and_offset stdout ----
[...]

PR #1141, opened 2017-10-21
WIP: add all_versions_yanked

seanlinsley:
For #1058 

I chose this approach in order to reduce duplication and prevent unnecessary database requests. For example, `max_version` was calling the database even though the versions had already been eagerly loaded, and it duplicated the logic already in `Version::max`.

But there are plenty of issues with this approach. Some of them are:

- only `show` is applying an order to `versions`
  - I would move ordering down into `encodable`, but `show` ends up using the variable again
- `minimal_encodable` now produces an array of version numbers, when it didn't before
  - unless we care about keeping the JSON response as small as possible, this doesn't seem to be a real issue

Is this sort of refactor welcome, and if so how do you think it could be improved?

One possible bug I noticed in the previous code was that the show action doesn't filter out yanked versions, while the index action does. Was that intentional, or an oversight?

### DIFF
--- a/src/krate/mod.rs
+++ b/src/krate/mod.rs
@@ -17,7 +17,6 @@ use diesel_full_text_search::*;
 use license_exprs;
 use hex::ToHex;
 use serde_json;
-use semver;
 use url::Url;
 
 use app::{App, RequestApp};
@@ -125,6 +124,7 @@ pub struct EncodableCrate {
     pub repository: Option<String>,
     pub links: CrateLinks,
     pub exact_match: bool,
+    pub all_versions_yanked: bool,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -328,14 +328,13 @@ impl Crate {
 
     pub fn minimal_encodable(
         self,
-        max_version: semver::Version,
+        versions: Vec<Version>,
         badges: Option<Vec<Badge>>,
         exact_match: bool,
         recent_downloads: Option<i64>,
     ) -> EncodableCrate {
         self.encodable(
-            max_version,
-            None,
+            versions,
             None,
             None,
             badges,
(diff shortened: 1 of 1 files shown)


src/krate/mod.rs
```diff
@@ -366,10 +364,16 @@ impl Crate {
             repository,
             ..
         } = self;
-        let versions_link = match versions {
-            Some(..) => None,
-            None => Some(format!("/api/v1/crates/{}/versions", name)),
-        };
```
seanlinsley: Was this condition really necessary? (It removes the link if any versions are passed to `encodable`.

seanlinsley:
No worries; I figured you were busy with Rust Belt Rust.

>show uses the versions variable again to return encoded version JSON, right? It looks like your current iteration of the code is cloning versions before passing it to Crate encodable, so that should allow for re-use, right? Or is there some other problem with that approach?

The code works as it's currently written; I was saying that I would've wanted to make it so that `encodable` applied ordering instead of `show` so it wouldn't have to be duplicated in other controllers calling `encodable`. But AFAIK that's not possible because of Rust variable ownership (giving ownership back to `show`, of a modified vector).

>Was there some behavior that seemed wrong due to this filtering, or was it just an inconsistency that caught your eye?

I just didn't understand the application's reasoning for the difference. It seems to make sense now.

>While testing this branch out, I'm seeing more ajax requests happen than in production.

The association is `DS.hasMany('versions', { async: true })` so theoretically [that shouldn't be happening](https://discuss.emberjs.com/t/what-is-an-async-relationship-async-true-vs-async-false/4107). Despite that, the version link addition certainly could have caused it.

How do you usually seed your development database?

Honestly I'm surprised that y'all are using Ember (or any heavy JS framework). At work we rewrote a Rails site w/ Ember, only to rewrite again with minimal JS because the site had grown too slow and the code duplication & obtuse bugs significantly slowed down development.

carols10cents:

Thank you!

PR #1147, opened 2017-10-25
Bump clippy and rustfmt

jtgeibel:
Bump clippy and rustfmt to versions compatible with the latest nightly compiler.


carols10cents:
hm i'm getting some clippy warnings locally on rust 2017-11-07 and clippy v0.0.169, but in the interest of not letting this sit too long and since travis is happy:

bors: r+

PR #1152, opened 2017-10-29
Add support for a CDN in front of crates/readmes

alexcrichton:
This should hopefully allow us to configure a CDN on Heroku (namely CloudFront)
to distribute crates/readmes through.

### DIFF
--- a/src/uploaders.rs
+++ b/src/uploaders.rs
@@ -46,7 +46,11 @@ impl Uploader {
     /// It returns `None` if the current `Uploader` is `NoOp`.
     pub fn crate_location(&self, crate_name: &str, version: &str) -> Option<String> {
         match *self {
-            Uploader::S3 { ref bucket, ref cdn, .. } => {
+            Uploader::S3 {
+                ref bucket,
+                ref cdn,
+                ..
+            } => {
                 let host = match *cdn {
                     Some(ref s) => s.clone(),
                     None => bucket.host(),
@@ -65,7 +69,11 @@ impl Uploader {
     /// It returns `None` if the current `Uploader` is `NoOp`.
     pub fn readme_location(&self, crate_name: &str, version: &str) -> Option<String> {
         match *self {
-            Uploader::S3 { ref bucket, ref cdn, .. } => {
+            Uploader::S3 {
+                ref bucket,
+                ref cdn,
+                ..
+            } => {
                 let host = match *cdn {
                     Some(ref s) => s.clone(),
                     None => bucket.host(),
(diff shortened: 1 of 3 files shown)


src/uploaders.rs
```diff
@@ -61,11 +65,14 @@ impl Uploader {
     /// It returns `None` if the current `Uploader` is `NoOp`.
     pub fn readme_location(&self, crate_name: &str, version: &str) -> Option<String> {
         match *self {
-            Uploader::S3 { ref bucket, .. } => Some(format!(
-                "https://{}/{}",
-                bucket.host(),
-                Uploader::readme_path(crate_name, version)
-            )),
+            Uploader::S3 { ref bucket, .. } => {
```
carols10cents: Looks like this line needs to be changed to `Uploader::S3 { ref bucket, ref cdn, .. } => {`

carols10cents:
@withoutboats is CloudFront reachable when using Tor?

alexcrichton:
Oops forgot to verify last minute change, should be fixed now. It's also worth mentioning that the motivation for this is primarily https://github.com/rust-lang/cargo/issues/4670 where I was reminded that S3 is unreachable in China. The reporter there though mentioned that rust-lang.org is reachable (which uses CloudFront).

I've also historically wanted to use CloudFront instead of S3 directly for download speed and reliability, but nothing has been pressing except for the China-blocks-us-west-1 problem

withoutboats:
I don't know anything about cloudfront but I know other CDNs (e.g. cloudflare) will frequently throw up captchas for tor users as a ddos mitigation. We ought to make sure the CDN will never do it in our use case because I expect cargo would just barf.

alexcrichton:
Does www.rust-lang.org work? If so it's the same CDN. I don't know if CloudFront does captchas...

withoutboats:
Based on this white paper I think cloudfront on its own will never intercept requests as long as we don't use a service like AWS WAF https://d0.awsstatic.com/whitepapers/DDoS_White_Paper_June2015.pdf

carols10cents:
bors: r+

messense:
seems unable to load readme now?

```
Refused to connect to 'https://d19xqa3lc3clo8.cloudfront.net/readmes/cc/cc-1.0.3.html' because it violates the following Content Security Policy directive: "connect-src 'self' https://docs.rs https://crates-io.s3-us-west-1.amazonaws.com".
```

carols10cents:
@messense on it! Thank you for letting me know!

carols10cents:
@messense readmes are fixed now!

PR #1157, opened 2017-11-02
Bump version of most dependencies

jtgeibel:
This series of commits bumps many dependencies, with the exception of diesel (as I couldn't figure out how to update the reverse dependencies query).  In particular, this switches from `time` to `chrono` in the `cargo_registry_s3` crate.

I've also updated the contributing instructions for Ubuntu and Fedora to explicitly install `pkg-config`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,10 +25,10 @@ dependencies = [
 
 [[package]]
 name = "ammonia"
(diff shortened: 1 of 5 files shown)


carols10cents:
Looks great!! Thanks!!

bors: r+

PR #1158, opened 2017-11-05
Exclude yanked from the latest version checked

### LINKED ISSUE #1144: Latest Version of Crate Does Not Show For Alpha Versions
Example: https://crates.io/crates/turtle (example will work until I publish a non-alpha version)

Test case:
- Have some 0.x.x versions (not sure if they need to be yanked or not to reproduce the problem, but in my case they were)
- Release a 1.0.0-alpha.0 version (doesn't matter what number you put after the alpha)

Problem:
crates.io does not display the 1.0.0-alpha.0 version. If your 0.x.x versions are yanked, it will actually display: 

> CrateName 0.x.x
> 
> This crate has been yanked, but it is still available for download for other crates that may be depending on it.
>
> You may wish to view all versions to find one that has not been yanked.

Viewing the other versions shows that there is a 1.0.0-alpha.0.

That means that it isn't finding that the latest version of the crate is 1.0.0-alpha.0 (possibly due to the -alpha.0 at the end). It still believes the 0.x.x versions are the latest.

Both cargo itself and docs.rs seem to work. If you use a "*" as your version, cargo will download the alpha correctly. If you just go to the [crate URL on docs.rs](https://docs.rs/turtle), it'll redirect you correctly. 

-----

Instructions to fix added by carol:

- [ ] [This check](https://github.com/rust-lang/crates.io/blob/601ed3ca93bd92c66d5b4194b4bb0c01c878a5cb/app/routes/crate/version.js#L62-L66) that tries to find the latest stable version should find the latest stable NON-YANKED version instead. The `version` has a `yanked` boolean property accessible via `version.get('yanked')`.
- Create more tests similar to [this test](https://github.com/rust-lang/crates.io/blob/601ed3ca93bd92c66d5b4194b4bb0c01c878a5cb/tests/acceptance/crate-test.js#L17-L30) that set up a bunch of different situations:
  - [ ] All versions yanked should get the "all versions yanked" page
  - [ ] Has larger prerelease versions and non prerelease versions, all NOT yanked, the version recommended on the page should be the largest non prerelease version
  - [ ] Has only prerelease versions, the largest prerelease version should be recommended
  - [ ] (This issue) Has larger prerelease versions and non prerelease versions, all non prerelease versions yanked, the version recommended on the page should be the largest *prerelease* version
  - [ ] Has larger prerelease versions (say 1.0.0-alpha), a smaller non-yanked non-prerelease version (say 0.8.0), and a *yanked* non-prerelease version in the middle (say 0.9.0), the recommended version should be the non-yanked non-prerelease version
  - [ ] Any other test cases you think of that I haven't!

behos:
Don't include yanked versions when looking for a latest version to show
to the users. This allows showing pre-release versions which are not
yanked when all stable versions were.

Fixes #1144

### DIFF
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -60,7 +60,7 @@ export default Route.extend({
             if (isUnstableVersion(maxVersion)) {
                 crate.get('versions').then(versions => {
                     const latestStableVersion = versions.find(version => {
-                        if (!isUnstableVersion(version.get('num'))) {
+                        if (!(isUnstableVersion(version.get('num')) || version.get('yanked'))) {
                             return version;
                         }
                     });
--- a/tests/acceptance/crate-test.js
+++ b/tests/acceptance/crate-test.js
@@ -59,6 +59,53 @@ test('visiting /crates/nanomsg/0.6.0', async function(assert) {
     assert.dom('[data-test-heading] [data-test-crate-version]').hasText('0.6.0');
 });
 
+test('crate page with all versions yanked has yanked message', async function(assert) {
+    server.create('crate', { id: 'nanomsg', max_version: '0.6.1' });
+    server.create('version', { crate: 'nanomsg', num: '0.6.0', yanked: true });
+    server.create('version', { crate: 'nanomsg', num: '0.6.1', yanked: true });
+
+    await visit('/crates/nanomsg');
+    assert.dom('[data-test-heading] [data-test-crate-version]').hasText('0.6.1');
+    assert.dom('.crate-info').hasText(/This crate has been yanked/);
+});
+
+test('crate page with max version in prerelease version shows max stable', async function(assert) {
+    server.create('crate', { id: 'nanomsg', max_version: '0.6.1-pre' });
+    server.create('version', { crate: 'nanomsg', num: '0.6.0' });
+    server.create('version', { crate: 'nanomsg', num: '0.6.1-pre' });
+
+    await visit('/crates/nanomsg');
+    assert.dom('[data-test-heading] [data-test-crate-version]').hasText('0.6.0');
+});
+
+test('crate page with only prerelease versions shows max prerelease', async function(assert) {
+    server.create('crate', { id: 'nanomsg', max_version: '0.6.0-pre.1' });
+    server.create('version', { crate: 'nanomsg', num: '0.6.0-pre.0' });
+    server.create('version', { crate: 'nanomsg', num: '0.6.0-pre.1' });
+
+    await visit('/crates/nanomsg');
+    assert.dom('[data-test-heading] [data-test-crate-version]').hasText('0.6.0-pre.1');
+});
+
+test('crate page with all stable versions yanked shows max prerelease', async function(assert) {
+    server.create('crate', { id: 'nanomsg', max_version: '0.6.1-pre' });
+    server.create('version', { crate: 'nanomsg', num: '0.6.0', yanked: true });
+    server.create('version', { crate: 'nanomsg', num: '0.6.1-pre' });
+
+    await visit('/crates/nanomsg');
+    assert.dom('[data-test-heading] [data-test-crate-version]').hasText('0.6.1-pre');
+});
+
+test('crate page with yanked max stable version shows latest non-yanked stable', async function(assert) {
+    server.create('crate', { id: 'nanomsg', max_version: '0.6.2-pre' });
+    server.create('version', { crate: 'nanomsg', num: '0.6.0' });
+    server.create('version', { crate: 'nanomsg', num: '0.6.1', yanked: true });
+    server.create('version', { crate: 'nanomsg', num: '0.6.2-pre' });
+
+    await visit('/crates/nanomsg');
+    assert.dom('[data-test-heading] [data-test-crate-version]').hasText('0.6.0');
+});
+
 test('navigating to the all versions page', async function(assert) {
     server.loadFixtures();
 


carols10cents:
Thank you for your patience and sorry for the delay on getting you feedback for this PR!

Good news: it looks like this fixes #1144 when visiting a crate page directly!

Bad news: it looks like there's still some weirdness when ember has some partial data? Or is this the frontend/backend logic inconsistency [you pointed out](https://github.com/rust-lang/crates.io/issues/1144#issuecomment-341963236)?

Here's what I'm seeing:

- I was testing this out locally but with the production backend by running `npm run start:live` and visiting http://localhost:4200/
- I searched for "apint", one of the crates mentioned in #1144 
- I clicked on the link to apint, which goes to http://localhost:4200/crates/apint
- I got the page that says "This crate has been yanked" with the current version shown as 0.0.0
- I clicked on the "view all versions" link in the notice that says "You may wish to view all versions to find one that has not been yanked. "
- There I can see the non-yanked alpha versions
- I went back in my browser, landed again on http://localhost:4200/crates/apint
- This time I'm shown apint 0.0.0-alpha.2, without the "this crate has been yanked" notice!
- If I visit http://localhost:4200/crates/apint directly, I see apint 0.0.0-alpha.2

Writing this out, I think this is definitely related to the backend/frontend logic... and ohhh no the backend is inconsistent with itself :(

The max version for apint returned from https://crates.io/api/v1/crates/apint is 0.0.0-alpha.2 
The max version for apint returned from https://crates.io/api/v1/crates?page=1&per_page=10&q=apint&sort= is 0.0.0 :(

Also I realized we can't remove max_version form the api JSON because [shields.io uses it](https://github.com/badges/shields/blob/f403c2b5d3895b2501654a5d2cd97d1e223fe138/server.js#L625).

Aaaaand I realize this has strayed pretty far from this PR, it's just that this PR makes the discrepancy more obvious 😅 Are you interested in fixing the backend too? If not, I can take care of it!

behos:
Sure, I'm happy to pick up the back-end work as it will be a good opportunity for me to work on the (more interesting) rust side of things.

There seems to be a lot of different parts doing the same things for calculating the versions. Sometimes it's in the front-end, sometimes it's in the back-end.

This would be a good opportunity to clean this up and have a consistent interface on the back-end. 

I think that not removing max_version from the returned item is fair enough, but we should be very explicit of what max_version means. 

Is it:

* The latest version regardless of yanked/not-yanked and release/prerelease?
* The latest non-yanked version regardless of release/prerelease?
* The latest non-yanked release version followed by the latest non-yanked pre-release version?

The last one seems to be the one that the front-end tries to calculate, should this be the meaning of max_version in all cases?

behos:
So while reproducing this issue, I noticed yet another inconsistency. It seems that on the main page when the crate has been recently updated, you get the correct version in the front-end (non-yanked prerelease).

So it seems like the summary call on the backend also reports the latest non-yanked pre-release as the max version.

Given these results, I think it's safe to say that max_version means:

> The latest non-yanked release version followed by the latest non-yanked pre-release version?

carols10cents:
I think what we *really* want to display just about everywhere on crates.io that we show a single version number for a crate is the version you'd get if, in your `Cargo.toml`, you put `cratename = '*'`. I'm going to go poking around in cargo's code to make sure I've got the rules correct, I *think* it's largest non-pre-release non-yanked.

The only exception I think should be the "just updated" column on the homepage, since "updated" for us means "new version published", so we should show whichever version was just published, maybe-- but I have a separate issue i need to file about that data, so i'd say ignore it for now.

Going to investigate in cargo now!

carols10cents:
Ok i lied, we can't do what cargo does for "*" because cargo WON'T give you prerelase versions, even if there are only prerelease versions, unless you specificially ask for prerelase versions.

> Other than that, I'm not sure how to proceed here. Even when the search becomes consistent with the rest of these, then we still have the inconsistency that visiting a crate directly gives you a "recommended" version (giving non-prerelease versions precedence) but when doing the same via search will give you the latest non-yanked version.
>
> Maybe the best approach would be to:
>
> * Make the max_versions consistent (latest non-yanked version)
> * Remove the "recommended" version functionality from the front-end since it might lead to confusion

I'm not clear on why this solution wouldn't work, for search, individual crate pages, crate list pages' information coming from the backend:

- If there are non-yanked, non-prerelease versions, the max_version is the largest of those.
- If there are no non-yanked, non-prerelease versions but there are non-yanked prerelase versions, the max_version is the largest of those.
- If all versions are yanked, the max_version is 0.0.0.

So apint's max_version should be 0.0.0-alpha.2. Is there a reason why this wouldn't work?

behos:
I think that's perfectly reasonable as long as we are consistent with this calculation :smile:. This has gotten big, so, to summarize:

* Max version will be calculated as the latest non-yanked, non-prerelease version, followed by the latest non-yanked, prerelease version followed by 0.0.0. This should be true for every crate object returned by the API
* "Just updated" will need a separate solution to be able to not follow this convention. (Maybe we can introduce another field in the API which corresponds to _any_ latest version)
* There's no longer any JavaScript magic done on the individual crate page front-end to get the latest version since the one provided from the backend is the correct one. (Which means discarding this PR)
* Shields.io will work out of the box for lists since it uses max_version as well. (Ideally we would change the front-end to always request a specific version from shields.io to avoid this kind of issues in the future)

Does this sound right? If so, I will move this summary to the original issue's page and discard this PR to continue from there.

carols10cents:
> -    Max version will be calculated as the latest non-yanked, non-prerelease version, followed by the latest non-yanked, prerelease version followed by 0.0.0. This should be true for every crate object returned by the API

Yep!

> -    "Just updated" will need a separate solution to be able to not follow this convention. (Maybe we can introduce another field in the API which corresponds to any latest version)

Yep! and this doesn't necessarily have to be fixed in the same PR as the other parts of this max_version problem.

> -   There's no longer any JavaScript magic done on the individual crate page front-end to get the latest version since the one provided from the backend is the correct one. (Which means discarding this PR)

Sounds right :( Sorry for the rework!

> -  Shields.io will work out of the box for lists since it uses max_version as well. (Ideally we would change the front-end to always request a specific version from shields.io to avoid this kind of issues in the future)

Yep, although other pages can always use shields.io without a specific version, so we should keep that working and consistent as well if possible.

Thank you!

behos:
Sorry, I've been away for a while. Picking this up now and will update if I run into any issues

PR #1165, opened 2017-11-09
Switch to `cargo clippy` which also lints the binaries and tests

jtgeibel:
This switches our CI to install `clippy` and use the `cargo clippy` command.  Additionally, the tests are now run on CI in the same build as `rustfmt`.  If `cargo fmt` fails, travis will still run `cargo clippy` and report any errors there.

In this configuration, `clippy` is no longer included in the `Cargo.toml` file.

/cc @carols10cents

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,21 +40,16 @@ matrix:
     - rust: nightly
   include:
   - rust: nightly-2017-11-07
-    env: RUSTFMT=YESPLEASE
     script:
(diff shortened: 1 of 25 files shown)


jtgeibel:
I've confirmed that the log files [show both errors](https://travis-ci.org/rust-lang/crates.io/jobs/299456770) if there is an issue with formatting and a lint.  This is ready for review.

carols10cents:
woooo got my local env configured this way too and everything seems to be matching!!!!

bors: r+

PR #1170, opened 2017-11-14
repository URLs sometimes end in .git in real-world usage

kivikakk:
And this is okay, because the major sites (BitBucket, GitHub, GitLab) all either directly support visiting the `.git`-suffixed URL of any repo, or redirect from it to the non-suffixed version. These all work fine as-is.

What they don't universally work for is letting you append any path onto that. (Which is to say, from my testing BitBucket doesn't mind, but GitHub and GitLab both 404.)  So we trim the suffix when manipulating relative paths.

See https://github.com/kbknapp/clap-rs/issues/1106 for where this came up with `clap`.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -310,11 +310,12 @@ mod tests {
         let relative = "[there](there)";
 
         for host in &["github.com", "gitlab.com", "bitbucket.org"] {
-            for &extra_slash in &[true, false] {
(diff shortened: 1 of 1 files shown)


carols10cents:
Looks great, thank you! Sorry for the delay!!

bors: r+

kivikakk:
Not at all, real life happens!

PR #1180, opened 2017-11-30
Update Ember

tchak:
Controller injections were broken in recent version of Ember. Not sure if it is normal but I don't like them anyway, so I removed them :)

cc @locks @Turbo87

### DIFF
--- a/app/controllers/search.js
+++ b/app/controllers/search.js
@@ -1,16 +1,19 @@
 import Controller from '@ember/controller';
 import { computed } from '@ember/object';
+import { alias, bool, readOnly } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
 
 import PaginationMixin from '../mixins/pagination';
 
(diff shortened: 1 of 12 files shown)


app/controllers/search.js
```diff
@@ -1,16 +1,19 @@
...
+import { alias, bool, readOnly } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
 
 import PaginationMixin from '../mixins/pagination';
 
 export default Controller.extend(PaginationMixin, {
+    search: service(),
     queryParams: ['q', 'page', 'per_page', 'sort'],
-    q: null,
+    q: alias('search.q'),
```
Turbo87: am I understanding correctly that we're now redirecting the `q` query param to the `q` property of the `search` service?

tchak: yes. It allows to sync input value with query string in the url.

PR #1181, opened 2017-11-30
Use ember-keyboard

tchak:
ember-keyboard is FastBoot compatible (noop)


app/controllers/application.js
```diff
@@ -1,57 +1,20 @@
...
-            $('#cargo-desktop-search').focus();
-        }
-    },
-
-    willDestroy() {
-        $(document).off('keypress');
-        $(document).off('keydown');
-    },
+        event.preventDefault();
+        document.querySelector('#cargo-desktop-search').focus();
```
Turbo87: can you check if that also works on small screens? I roughly remember that the search field is not always shown depending on screen size.

tchak: You right, it doesn't work on small (mobile) screen. This is not new - it was already not working before my change. But on a small screen (phone, tablet) you usually do not use keyboard shortcuts. I could test if "desktop" input is visible and focus accordingly, but I am not sure if it worth it.

Turbo87: I think `$('#cargo-desktop-search').focus()` would work since it will just noop if the selector has no matches.

```js
let searchField = document.querySelector('#cargo-desktop-search');
if (searchField) {
  searchField.focus();
}
```

How about this?

Turbo87:
@tchak looks like #1180 caused conflicts. can you rebase?

app/controllers/application.js
Turbo87: nevermind, I just noticed that this is not necessary

tchak: Yes, I added your change but I could not reproduce the error...

tchak: I guess better safe then sorry :)

PR #1183, opened 2017-11-30
Don’t use global events for login action

tchak:
Test coverage for this stuff is not great, but I tested it manually. (I have the whole stack running locally now \o/)


carols10cents:
I can log in! I can log out! I can log in and shake it all about! Do the hokey pokey, and merge it all around, that's what it's all about!

Thank you!!!

bors: r+

Turbo87:
@carols10cents sweet, thanks for the review :)

PR #1184, opened 2017-11-30
Update to use Diesel 0.99.0

sgrif:
Most of the code changes are for us to switch to `sql_query` where
appropriate. `sql` deprecated its API for bind parameters, as it's meant
to be used with fragments of SQL as part of the query builder, not full
queries. `sql_query` does not require the SQL type to be specified, and
it gets the values by name instead of by index. Unfortunately, none of
the places we are using raw SQL were places that
`#[derive(QueryableByName)]` would work, so I had to manually implement
all of them

This also updates to r2d2 0.8, because I couldn't remember the command
to make cargo force r2d2-diesel to resolve to r2d2 0.7

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,6 +84,11 @@ version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "bitflags"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bufstream"
 version = "0.1.3"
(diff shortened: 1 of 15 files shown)


sgrif:
1.0 will let `#[derive(QueryableByName)]` work with all of these manual implementations, except the one for `Dependency`

carols10cents:
looks good!

bors: r+

sgrif:
bors: retry

sgrif:
I guess that isn't a command.

bors: r+

sgrif:
The failure is due to some NPM nonsense that is clearly unrelated to this branch. `stable` and `beta` both passed, so I'm going to YOLO merge

PR #1185, opened 2017-11-30
Update Ember-CLI to v2.17

tchak:
I updated `ember-data` with the rest of packages, but it is a pretty drastic update. Tests seems to be ok. But maybe with it test more manually. I can also remove `ember-data` bump from this PR.

edit: the `ember-data` updates was removed from this PR


package.json
```diff
@@ -28,24 +28,25 @@
...
-    "ember-cli-qunit": "^4.0.0",
+    "ember-cli-qunit": "^4.1.1",
     "ember-cli-sass": "^6.2.0",
-    "ember-cli-shims": "^1.1.0",
+    "ember-cli-shims": "^1.2.0",
     "ember-cli-sri": "2.1.1",
-    "ember-cli-uglify": "1.2.0",
-    "ember-data": "~2.12.1",
+    "ember-cli-uglify": "^2.0.0",
+    "ember-cli-update": "^0.10.1",
```
Turbo87: IMHO this should be a globally installed tool

package.json
```diff
@@ -28,24 +28,25 @@
...
+    "ember-data": "~2.17.0",
```
Turbo87: I would prefer if we could split this update into a separate PR

testem.js
```diff
@@ -3,20 +3,23 @@
...
+    Chrome: {
+      mode: 'ci',
+      args: [
+        '--disable-gpu',
+        '--headless',
+        '--remote-debugging-port=9222',
+        '--window-size=1440,900'
+      ]
     }
+  }
```
Turbo87: tabsize for this file changed. we should try to stay consistent with the other files and use 4 spaces here too.

Turbo87:
thanks! 👍

PR #1186, opened 2017-12-01
Load summary data in the background

Turbo87:
This PR improves the perceived frontpage performance experience by loading the summary data in the background while already displaying the "data" that is available.

crates

We can probably use the same pattern for some other routes too to get rid of the "Loading..." screen.

/cc @carols10cents @tchak

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -24,12 +24,12 @@
     <div id='stats'>
         <div class='downloads'>
             {{svg-jar "download"}}
-            <span class='num' data-test-total-downloads>{{format-num model.num_downloads}}</span>
+            <span class='num' data-test-total-downloads>{{if hasData (format-num model.num_downloads) "---,---,---"}}</span>
             <span class='desc small'>Downloads</span>
         </div>
         <div class='crates'>
             {{svg-jar "crate"}}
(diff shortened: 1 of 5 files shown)


app/templates/index.hbs
```diff
@@ -24,40 +24,40 @@
...
+            <span class='num' data-test-total-crates>{{if hasData (format-num model.num_crates) "---,---"}}</span>
             <span class='desc small'>Crates in stock</span>
         </div>
     </div>
 </div>
 
 <div id='home-crates' class='crate-lists'>
     <div id='new-crates' data-test-new-crates>
-        <h2>New Crates</h2>
+        <h2>New Crates {{#if dataTask.isRunning}}<img src="/assets/ajax-loader.gif">{{/if}}</h2>
```
tchak: Could we have a component for the spinner? Like `{{ui-spinner task=dataTask}}`.

Turbo87: sure, though I'd keep the `if` condition outside of the component and just wrap the `` element. IMHO this is an independent refactoring though and should not block merging this PR here, right?

tchak: :+1:

PR #1188, opened 2017-12-01
Do not listen to ‘hashchange’ events in FastBoot

tchak:


Turbo87:
@tchak are there any tests for this behavior? do you think it would be possible to write robust tests for this given that scrolling is involved?

tchak:
I will investigate

app/controllers/crate/version.js
```diff
@@ -179,7 +178,14 @@ export default Controller.extend({
     },
 
     report: observer('crate.readme', function() {
-        setTimeout(() => $(window).trigger('hashchange'));
-    }),
-
+        if (typeof FastBoot === 'undefined') {
```
sgrif: This is probably a silly question, but I don't do much JS and I'm curious... Why is this better than `FastBoot === undefined`)? (I figure we can reasonably assume that nothing has redefined `undefined` in the crates.io code)

Turbo87: @sgrif see https://stackoverflow.com/a/4725697/1478093

notriddle:
Any news on the testing things? Or can it just be landed without tests right now?

PR #1194, opened 2017-12-03
Start search while typing

Turbo87:
Similar to e.g. https://yarnpkg.com/ this PR changes the code to start the search while typing instead of having to click the <kbd>Enter</kbd> button. It also debounces the search task (250ms) to limit the amount of requests that are hitting the server.

crates-live-search

/cc @carols10cents

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -21,7 +21,11 @@ export default Controller.extend(EKMixin, {
     }),
 
     actions: {
-        search() {
+        search(q) {
+            if (q !== undefined) {
+                this.set('searchQuery', q);
+            }
+
             this.transitionToRoute('search', {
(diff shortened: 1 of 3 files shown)


carols10cents:
cool!! 

bors: r+

durka:
N.B. this caused #1209

Turbo87:
Ouch, thanks for reporting this. I'll have a look tomorrow.

nox:
This breaks submitting to a new tab without changing the current view, which I do often when looking up multiple crates at once.

durka:
Yes, there are a number of other disadvantages, like it's slow :) but I
doubt it'll be reverted so I wanted to point out the most obvious bug.

Issue #820, opened 2017-06-27
Switching to ember-fetch causes github oauth to fail with error "invalid state parameter"


carols10cents:
This happened on production when I deployed e921c6eb; the previous deploy of 9172aa4f worked. A bisect pointed to this commit: https://github.com/rust-lang/crates.io/commit/dbc1c9a08b3d9f79fab8e4bef27692aeb38f5a42

I'm reverting that commit, which fixes the problem....

sivakumar-kailasam:
@carols10cents we're good on this now that we've moved to ember-ajax to be FastBoot compatible via #838

PR #1195, opened 2017-12-03
Replace `ember-ajax` with `ember-fetch`

tchak:
This will path the way to eventually remove jQuery :)

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -1,5 +1,6 @@
 import DS from 'ember-data';
+import AdapterFetch from 'ember-fetch/mixins/adapter-fetch';
 
-export default DS.RESTAdapter.extend({
+export default DS.RESTAdapter.extend(AdapterFetch, {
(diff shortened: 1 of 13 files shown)


Turbo87:
@tchak do the error messages change in any way after this?

carols10cents:
We tried switching to ember-fetch a while ago and had [a problem](https://github.com/rust-lang/crates.io/issues/820)-- i don't know enough about what we were doing then, what might be different now, and what the cause of that problem was to know whether this will still be a problem :-/

tchak:
@carols10cents thanks for pointing it out. I will investigate and report!

tchak:
@Turbo87 it should not :) I will do a serie of manual tests and check.

Turbo87:
@tchak @tylerturdenpants I've rebased this on `master` now

Turbo87:
Awesome, thanks for the verification!

PR #1212, opened 2017-12-23
Account for stable/prerelease when cheking for max version

### LINKED ISSUE #1144: Latest Version of Crate Does Not Show For Alpha Versions

behos:
Max version will be calculated as the latest non-yanked,
non-prerelease version, followed by the latest non-yanked, prerelease
version followed by 0.0.0. This should be true for every crate object
returned by the API

This removes the custom front-end logic to fetch the max-version since this
is pre-calculated in the backend.

It also updates the search filter to exclude yanked versions

Fixes #1144


behos:
As discussed, this does not actually deal with the latest updates where the actual latest version needs to be shown

carols10cents:

Thank you!

PR #1213, opened 2017-12-23
Minor cleanup on dependency insertion and loading

sgrif:
I'm going through some random code as I'm looking to point to this repo
more in Diesel's guides for some actual examples. My goal was actually
to do the following:

- `#[derive(Insertable)]` on `upload::CrateDependency` and use that
  directly
- `#[derive(Queryable)]` on `git::Dependency` and construct it from a
  returning clause

However, doing the first piece would require skipping the `crate_name`
field or pulling it out of that struct, and the second piece requires
some way to tell Diesel that a subselect returns a single row (which is
not hard to do, but is not code that should live in this repo)

Either way, while I was experimenting I did a bit of cleanup here which
I think is worthwhile on its own.

- Removed an `Insertable` struct in favor of tuple inserts, since
  `Insertable` is really only mean to be used when your input comes from
  another source (like `Deserialize`).
- Made some defaults explicit in the tests. It felt sketchy having the
  default impl that was only used in tests. I thought about pushing
  these defaults into the database, but that felt sketchy for the same
  reason
- Removed a potential panic
- Removed a few unneccessary iterations/allocations over the features
  vec.


jtgeibel:
Looks good to me!  I'm not sure why resolving through the web interface didn't work the first time.

bors: r+

PR #1215, opened 2017-12-29
Revert "Merge #1194"

### LINKED ISSUE #1209: Crates.io creates a new browser history entry for every letter typed in the search field
Crates.io creates a new browser history entry for every letter typed in the search field. 
This is very annoying when navigating back because one has to go back through all these artificially created history entries.

Turbo87:
While I still like the feature it self, the implementation currently has some issues (see #1209) that I wasn't aware of initially and are not quite as easy to fix. Since I'll be on vacation for the next weeks I won't have time to work on a proper solution for now, so IMHO the best solution is to revert the problematic behavior until this can be fixed correctly.

Resolves #1209

/cc @durka @Boscop

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -21,11 +21,7 @@ export default Controller.extend(EKMixin, {
     }),
 
     actions: {
-        search(q) {
-            if (q !== undefined) {
-                this.set('searchQuery', q);
-            }
-
+        search() {
             this.transitionToRoute('search', {
                 queryParams: {
                     q: this.get('searchQuery'),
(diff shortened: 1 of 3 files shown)


carols10cents:
😿 it happens. 

bors: r+

Boscop:
Thanks :)
Btw, isn't there a way to do an ajax request to refresh the results without changing the history with pushState()?

Turbo87:
There certainly is, but having your searches in the history is actually a good thing, just not for every key stroke...

Boscop:
Yes, I mean only do pushState on Enter.
But then it won't pushState if someone clicks on a result from the instant suggestions.
So the best way would be to pushState before any further navigation, maybe with a pre-history-change hook, by listening to that event? Then doing pushState before letting the navigation happen..

PR #1216, opened 2017-12-29
Improve build times by caching `cargo install` artifacts for diesel, clippy, and rustfmt

jtgeibel:
The primary change is to set `CARGO_TARGET_DIR=target` which instructs `cargo install` to use the target/ directory which is cached between builds on travis.  This is only available on recent nightlies so I've bumped clippy and rustfmt versions.

Unfortunately, `ring v0.11.0` does not compile on recent nightlies and upstream provides no support for older versions.  Therefore I've added a patch section to use a git repo that incorporates the fix and a few other cleanups.  This issue would have broken the build soon anyway when rust 1.24 goes into beta.

A few other changes:

* only `npm install` on the stable job
* removed travis-cargo which appears to be unused
* removed rustfmt configuration options that were removed upstream

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,11 +1,4 @@
-fn_args_layout = "Block"
-array_layout = "Block"
-control_style = "Rfc"
-where_style = "Rfc"
-generics_indent = "Block"
-fn_call_style = "Block"
 combine_control_expr = true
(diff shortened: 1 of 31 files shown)


carols10cents:
i like faster! 

bors: r+

PR #1226, opened 2018-01-09
Explicitly state our license in the README, use `OR` in Cargo.toml

### LINKED ISSUE #1225: License for this repository?
I'd like to transition from `/` to `AND` or `OR` in crates.io's `Cargo.toml` (compare rust-lang/cargo#4920). My guess is that the intended license is `MIT OR Apache-2.0` because that's the same semantics [Rust uses for most of its code][1].

This repository only recently gained a similar human-readable explanation.  The first mention of “Apache” in this repo landed in fbdf1575a (2014-06-29) with a file header that included a similar human-readable license grant (great, but just for that file).  Then 7a2b8894d (2014-10-20) landed `LICENSE-APACHE` (and `LICENSE-MIT`) but no human-readable grant (e.g. in the README or a COPYRIGHT file).  4b07668170 (2015-01-12) dropped the one file that had an `Apache-2.0 OR MIT` grant header.  4c4381169 (2017-04-26) landed [the current human-readable explanation][2].  663ffc7538 (2017-04-28) added `license` information to `Cargo.toml` with the [underspecified][3] `/` form.

I expect we can express the expectations of most contributors by changing our current:

    license = "MIT/Apache-2.0"

to:

    license = "MIT OR Apache-2.0"

However, without local docs unpacking the meaning (like [Rust][1] has), it's not clear to me whether all 130+ crates.io contributors had the same impression of the intended license.  Are the crates.io maintainers comfortable making the update without contacting all of those contributors?  Or would they rather assume that the contributors were familiar enough with the Rust ecosystem to interpret `/` as `OR`?  What were the intentions of folks who first contributed before 4c4381169 (~⅔ of contributors) when this repository had just `LICENSE-APACHE` and `LICENSE-MIT`?  What about before 7a2b8894d (only @alexcrichton, @stefanpenner, @sfackler, and @wycats) where this repository had no project-wide licensing information at all?

[1]: https://github.com/rust-lang/rust/blob/1.23.0/COPYRIGHT#L16-L19
[2]: https://github.com/rust-lang/crates.io/blame/6802c0500a3fd6aa532af187d9dc1b5160354b33/docs/ARCHITECTURE.md#L13
[3]: https://github.com/rust-lang/cargo/issues/2174

sgrif:
Fixes #1225

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "cargo-registry"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 version = "0.2.2"
-license = "MIT/Apache-2.0"
+license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-lang/crates.io"
 description = "Backend of crates.io"
 
--- a/README.md
+++ b/README.md
@@ -57,3 +57,12 @@ categories will be synced from this file.
 ## Running a mirror
 
 Please see [docs/MIRROR.md](https://github.com/rust-lang/crates.io/blob/master/docs/MIRROR.md) for instructions on setting up a mirror of crates.io.
+
+## License
+
+Licensed under either of these:
+
+ * Apache License, Version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+   https://www.apache.org/licenses/LICENSE-2.0)
+ * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+   https://opensource.org/licenses/MIT)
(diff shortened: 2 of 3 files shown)


jtgeibel:
I checked and package.json already contains `(MIT OR Apache-2.0)`.  The `license_exprs` crate we use doesn't seem to handle parenthesis, but I confirmed that it accepts the expression as modified in this PR.

Would you update src/s3/Cargo.toml as well?

jtgeibel:
bors: r+

PR #1227, opened 2018-01-09
Accessibility fixes (and tests)

sorin-davidoi:
Most commits have a message explaining why I think the change is necessary. Let me know if anything is unclear or if I should follow up with other changes.

Shoutout to @chriskrycho for his talk [Becoming a Contributor](https://www.youtube.com/watch?v=Abu2BNixXak) which encouraged me to do this!

### DIFF
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -113,6 +113,14 @@ button.small {
 
     > section { margin: 0 15px; }
 
+    > section[aria-busy="true"] > h2::after {
+        content: '';
+        background-image: url('/assets/ajax-loader.gif');
+        display: inline-block;
+        height: 16px;
+        width: 16px;
+    }
+
     ul, ol { list-style: none; padding: 0; }
     li { margin: 8px 0; }
 
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -151,7 +151,7 @@ button.small {
     @include flex-wrap(wrap);
     @include justify-content(left);
 
-    > div {
+    > section {
       margin: 0;
       padding: 0 15px;
       width: 33.33%;
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -20,7 +20,7 @@
         margin-right: 10px;
     }
 
-    &:hover { @include vertical-gradient($start_dark, $end_dark); outline: 0; }
+    &:hover, &:focus { @include vertical-gradient($start_dark, $end_dark); outline: 0; }
     &.active { @include vertical-gradient($start_dark, $end_dark); outline: 0; }
     &[disabled] {
         @include vertical-gradient($start_light, $end_light);
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -111,7 +111,7 @@ button.small {
         line-height: 20px;
     }
 
-    > div { margin: 0 15px; }
+    > section { margin: 0 15px; }
 
     ul { list-style: none; padding: 0; }
     li { margin: 8px 0; }
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -113,7 +113,7 @@ button.small {
 
     > section { margin: 0 15px; }
 
-    ul { list-style: none; padding: 0; }
+    ul, ol { list-style: none; padding: 0; }
     li { margin: 8px 0; }
 
     li a {
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -2,7 +2,7 @@
 {{google-jsapi}}
 
 <a href='https://github.com/rust-lang/crates.io' class='fork-me'>
-  <img src='/assets/forkme.png'/>
+  <img src='/assets/forkme.png'/ alt="Fork me on GitHub">
 </a>
 
 <nav id="header">
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -26,6 +26,7 @@
                 tabindex="1"
                 required
                 data-test-search-input>
+        <label for="cargo-desktop-search">Search</label>
     </form>
 
     <div class='nav'>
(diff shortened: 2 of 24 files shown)


tests/axe-config.js
```diff
@@ -0,0 +1,7 @@
+export default {
+    rules: {
+        'color-contrast': {
+            enabled: false,
```
sorin-davidoi: There are multiple elements that currently fail this test. I have disabled it since I think it needs to be addressed by someone more inclined towards design.

sorin-davidoi:
I have integrated [ember-a11y-testing](https://github.com/ember-a11y/ember-a11y-testing) in order to avoid these problems from appearing again.

app/templates/application.hbs
```diff
@@ -100,25 +96,25 @@
...
             value={{searchQuery}}
             oninput={{action (mut searchQuery) value="target.value"}}
             autocorrect="off"
-            tabindex="1"
             required>
+    <label for="cargo-mobile-search">Search</label>
 </form>
 
-<div id="main" class='inner-content'>
+<main id="main" class='inner-content'>
```
sorin-davidoi: This is not ideal since each page should define it's own `main` - but I'm not familiar enough with the pages and don't have time to go through them all, so I think this is good enough for now.

carols10cents:
Thank you so much for this!!!!!! I'm taking a look now!!

app/templates/components/crate-list.hbs
```diff
@@ -1,4 +1,4 @@
-<ul>
+<ol>
```
carols10cents: Should [keyword-list](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/templates/components/keyword-list.hbs) and [category-list](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/templates/components/category-list.hbs) be changed to `ol`s too? They're both ordered by popularity, and those components are only used on the home page.

What about [crate-downloads-list](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/templates/components/crate-downloads-list.hbs)? This is used on the currently-logged-in user's [dashboard](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/templates/dashboard.hbs) and are ordered by recent downloads.

Not sure how far we should go with this in this PR-- there are lists of crates that are just `div`s rather than `ul`s right now that should also perhaps be `ol`s, like [every list of crates](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/templates/crates.hbs#L68) (which should always have SOME order) or [the list of Latest Updates of crates you're following on your dashboard](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/templates/dashboard.hbs#L58)

app/templates/application.hbs
```diff
@@ -2,7 +2,7 @@
 {{google-jsapi}}
 
 <a href='https://github.com/rust-lang/crates.io' class='fork-me'>
-  <img src='/assets/forkme.png'/>
+  <img src='/assets/forkme.png'/ alt="Fork me on GitHub">
```
carols10cents: I think the slash after the `src` attribute is to close the tag, so should go after the `alt` attribute.

app/templates/crate/version.hbs
```diff
@@ -74,7 +74,7 @@
             <code id="crate-toml">{{ crate.name }} = "{{ currentVersion.num }}"</code>
 
             {{#if (is-clipboard-supported)}}
-              {{#copy-button clipboardTarget="#crate-toml" success=(action 'copySuccess') error=(action 'copyError')}}
+              {{#copy-button clipboardTarget="#crate-toml" success=(action 'copySuccess') error=(action 'copyError') title="Copy to clipboard"}}
                     {{svg-jar "copy" alt="Copy to clipboard"}}
```
carols10cents: This isn't repetitive with the alt text on the svg inside the button? I really don't know how screenreaders will read this, just wanted to point it out :)

app/templates/crate/version.hbs
```diff
@@ -104,7 +104,7 @@
             {{/if}}
         {{/if}}
     </div>
-    <div class='authorship'>
+    <section class='authorship' aria-label="Authorship">
```
carols10cents: This class being named "authorship" is kind of... out of date. There's a lot more metadata in this crate than just the author. Not sure what a better label would be- "Crate info"? "Crate metadata"?

app/styles/home.scss
```diff
@@ -111,7 +111,7 @@ button.small {
         line-height: 20px;
     }
 
-    > div { margin: 0 15px; }
+    > section { margin: 0 15px; }
```
carols10cents: I think the [`div` on line 146 of this file](https://github.com/rust-lang/crates.io/pull/1227/commits/bd4c191c8a00b647f0683ed5ebe3ad4e496c66a7#diff-92c0575c9bf034586a7ce42ccddc91c4R146) needs to be updated to a `section` as well, otherwise the columns on the home page take up a variable amount of width when they should be taking up ~1/3 of the width, which starts happening at this commit.

app/templates/crate/version.hbs
sorin-davidoi: [aXe](https://www.axe-core.org/) flagged this as an error, and Orca (the screen reader I test with) didn't read the text from the svg.

app/templates/components/crate-list.hbs
sorin-davidoi: Yes, probably those need to be updated as well. Would rather do it in a follow-up PR if that's OK.

carols10cents: Yup, that's totally fine!

app/templates/application.hbs
```diff
@@ -128,4 +124,8 @@
     <a href='mailto:help@crates.io'>Send us an email</a>
     <span class="sep">|</span>
     {{#link-to 'policies'}}Policies{{/link-to}}
-</div>
+</footer>
+
+<a href='https://github.com/rust-lang/crates.io' class='fork-me'>
+  <img src='/assets/forkme.png'/ alt="Fork me on GitHub">
```
carols10cents: Just one more thing, github hid my comment on this because I made it on a commit and not the overall diff-- could you move the slash that's now between the `src` and the `alt` attributes to the end of the `img` tag so that it closes the tag? Then this PR should be good to go!

sorin-davidoi: @carols10cents Done!

carols10cents:
bors: r+

PR #1228, opened 2018-01-10
Avoid content jumps on front page

sorin-davidoi:
Currently, the crate lists are collapsed while the data is fetched (increasing in size when the data arrives). This makes the headers jump on the screen ([it's a common problem](https://css-tricks.com/content-jumping-avoid/)). This attempts to fix this by setting `min-height` on the lists (and adding a quick fade-in animation for when the crates arrive).

### DIFF
--- a/app/styles/home.scss
+++ b/app/styles/home.scss
@@ -28,6 +28,11 @@
     }
 }
 
+@keyframes fadein {
+    from { opacity: 0; }
+    to   { opacity: 1; }
+}
+
 .yellow-button {
     @include button(#fede9e, #fdc452);
     vertical-align: middle;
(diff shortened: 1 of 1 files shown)


app/styles/home.scss
```diff
@@ -147,11 +152,16 @@ button.small {
       margin: 0;
       padding: 0 15px;
       width: 33.33%;
+      min-height: 788.283px;
```
sgrif: This is an oddly specific number. Where did it come from?

sorin-davidoi: Ah, forgot to add a comment. It is the height of the list when it contains 10 items (which is always the case from what I can tell).

sgrif: This number seems wrong. Since it's dependent on the user's zoom setting and installed fonts, it should probably be measured in ems. For example, on my Windows machine, using Chrome, it's 774.261. I default to 110% zoom though. If I bump it down to 100%, it's 767.938.

sorin-davidoi: That's weird - Firefox and Chrome on Linux always show the same height (in px) regardless of zoom. Will look into it.

sgrif: They definitely shouldn't be. Zoom by definition changes the height in pixels.

sorin-davidoi:
So apparently browsers will scale the layout as well, not only the text. But there is still the problem caused by different fonts, so `min-height` should probably be computed with `calc` I guess. Don't have time to get to the root of this, closing.

Issue #9, opened 2014-10-21
Sort drop-downs are strewn around


alexcrichton:
Similar to #8, with the same examples. There's too much duplication for these methods.

sgrif:
We're cleaning out some of the older issues that haven't seen activity lately, so I'm going to close this.

We'd be happy to accept a PR that refactors this code (note: I'd also love to see our dropdowns move more towards being implemented with CSS than JS). I'm going to try to open an issue in the next few days with more concrete steps as well.

PR #1232, opened 2018-01-13
Add UI for sorting my crates by recent downloads

### LINKED ISSUE #1231: Sort "my crates" by recent downloads
#892 added the ability to sort search results by "recent downloads" in addition to "all-time downloads". This same sorting option should also be available on the "My crates" listing at https://crates.io/me/crates. For consistency, we should probably also rename "Downloads" to "All-time downloads".

jonhoo:
This fixes #1231, and renames "Downloads" to "All-Time Downloads" to be consistent with the sort dropdowns on other listings.

### DIFF
--- a/app/controllers/me/crates.js
+++ b/app/controllers/me/crates.js
@@ -15,6 +15,12 @@ export default Controller.extend(PaginationMixin, {
     totalItems: readOnly('model.meta.total'),
 
     currentSortBy: computed('sort', function() {
-        return (this.get('sort') === 'downloads') ? 'Downloads' : 'Alphabetical';
+        if (this.get('sort') === 'downloads') {
+            return 'All-Time Downloads';
+        } else if (this.get('sort') === 'recent-downloads') {
+            return 'Recent Downloads';
+        } else {
+            return 'Alphabetical';
+        }
     }),
 });
--- a/app/templates/me/crates.hbs
+++ b/app/templates/me/crates.hbs
@@ -33,7 +33,12 @@
                 </li>
                 <li>
                     {{#link-to (query-params sort="downloads")}}
-                        Downloads
+                        All-Time Downloads
+                    {{/link-to}}
+                </li>
+                <li>
+                    {{#link-to (query-params sort="recent-downloads")}}
+                        Recent Downloads
                     {{/link-to}}
                 </li>
             {{/rl-dropdown}}


carols10cents:
Thanks! I think this is *almost* there. I know this is hard to test because there isn't a way to log in when you're running `npm run start:live`, so you have to set up a backend and all that.

But when I was running this PR locally, when I go to /me/crates and then switch the sort to "All-Time Downloads", the closed dropdown says "Downloads", and if I switch to "Recent Downloads" the closed dropdown says "Alphabetical".

That's because [the closed dropdown displays the value in `currentSortBy`](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/templates/me/crates.hbs#L24), which comes from [this function in controllers/me/crates.js](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/controllers/me/crates.js#L17-L19), and that function only knows about "Downloads" and "Alphabetical". That function needs to change to look more like the other controllers, such as [controllers/crates.js](https://github.com/rust-lang/crates.io/blob/e4f8898db18c6d5995812a92d9ef275a0fc525fb/app/controllers/crates.js#L17-L25).

We're likely doing something wrong to end up with all this duplication (and [it's been this way for a while](https://github.com/rust-lang/crates.io/issues/9)), but I'm not sure the best way to fix the problem especially because while *most* of the sort dropdowns are the same, they aren't *all* the same (search result lists have "relevance" as an option).

So would you be willing to make the easy fix of updating the currentSortBy function in app/controllers/me/crates.js? Or would you like me to take care of it? It's no problem, just let me know!

jonhoo:
Fixed in a1eacf4db52840396713822c185ea4e42c107b7a! Thanks for the detailed instructions :)

Wouldn't it be possible to have the same sort be used everywhere, and then override it specifically on the search results page (which should presumably be the only one that differs)?

carols10cents:
Thank you, this works great now!!

bors: r+

> Wouldn't it be possible to have the same sort be used everywhere, and then override it specifically on the search results page (which should presumably be the only one that differs)?

Yep, it's certainly possible, and doing so would fix #9. There might be other places that are different, I'm not sure!

jonhoo:
The `waiting-on-author` label should probably be removed. Also, how/when do merged PRs show up on crates.io?

carols10cents:
Just deployed, this is live now :)

I don't usually bother removing status labels from merged PRs, I only use those when looking at open PRs, so the label hanging around is fine :)

PR #1234, opened 2018-01-15
Update to Diesel 1.1

sgrif:
This is the minimum changes required to fix any deprecation warnings.
There's a lot of additional stuff we could do that I wanted to keep
separate. After updating to Diesel 1.1, we can remove virtually anywhere
that we're doing an explicit `&*`. Additionally, there are probably a
lot of places that we should be using custom types but aren't. We can
remove a lot of code once
https://github.com/steveklabnik/semver/pull/169 is merged/released as well.


carols10cents:
\o/

bors: r+

PR #1236, opened 2018-01-18
Use semver's built-in Diesel support

sgrif:
This is still a PR and hasn't been released yet, but Steve's said he's
fine with the feature and it'll likely be released soon. I also had to
patch conduit so we don't get a mismatch between semver versions there.


carols10cents:
Looks like your patches have been merged into conduit and semver but neither have released a new version yet :( It'd be nice to get on released versions...

steveklabnik:
I'm working on it...

carols10cents:
@sgrif looks like you might have perms on conduit to tag on github but we might need @alexcrichton to either give you perms on crates.io or make a release... and still waiting on a release from @steveklabnik :)

PR #1237, opened 2018-01-18
Actually use JSON storage for JSON data

sgrif:
We also never use `features` as a `HashMap<String, Vec<String>>`, so we
can just pass it through from the database in a lower level
representation.


carols10cents:
Seems fine. The only thing I can think of this affecting is [this as-yet-unimplemented feature request to display a crate's features on its page](How can I teach that, other than giving people time and opportunity to become familiar w/the whole codebase & its history...? idk.), and that would be from JSON anyway.

Clarifying for anyone looking at this that this only affects getting a crate's features (per-version) into and out of the crates.io database, and does not touch the serialization to the crates.io index (which happens in [git.rs](https://github.com/rust-lang/crates.io/blob/6cb501f10824c427f89805ad73817c2bab1fd3d0/src/git.rs).

bors: r+

sgrif:
Failure is legit. I will fix when the baby stops screaming bloody murder

carols10cents:
bors: r+

PR #1238, opened 2018-01-18
Move `readme_rendered_at` off of the versions table

sgrif:
This column is never used in code other than for filtering/recording
purposes. We do not put it on the `Version` struct, and don't need to be
sending it over the wire every time we query that table.


carols10cents:
Uuuhhhh but you did the opposite to this? https://github.com/rust-lang/crates.io/commit/9c6412f07997ef70e879b8d8987aa2443ba4547e#diff-2d162e07be0cf65cc1054b2128c60258

sgrif:
Am I not allowed to change my mind? :(

The main thing I think I was opposed to was the amount of code that existed for no apparent reason, more than the schema itself.

carols10cents:
I just want to be sure you're not going to be changing your mind back and forth forever :(

sgrif:
Well I can at least say that moving this back again will require bringing back the manual `Queryable` impl (which I want to kill), so that's a decent motivation for leaving it in this state.

src/version/mod.rs
```diff
@@ -128,11 +128,10 @@ impl Version {
     }
 
     pub fn record_readme_rendering(&self, conn: &PgConnection) -> QueryResult<usize> {
-        use schema::versions::dsl::readme_rendered_at;
-        use diesel::dsl::now;
+        use schema::readme_renderings::dsl::*;
 
-        diesel::update(self)
-            .set(readme_rendered_at.eq(now.nullable()))
+        diesel::insert_into(readme_renderings)
```
carols10cents: Doesn't this need to be an upsert? Or does `insert_into` handle upserts automatically? Occasionally, when we change the readme rendering, I rerender all the readmes, which should update this timestamp.

sgrif: Good catch

carols10cents:
I added a test to encode the historical/contextual knowledge I had, it is indeed failing on this branch :)

sgrif:
bors: r+

(I'm assuming you're good with this after I fixed that issue)

PR #1240, opened 2018-01-19
Running ember tests in parallel

JoaoGFarias:
These changes:

1. Install a frontend dependency called [ember-exam](https://github.com/trentmwillis/ember-exam)
2. Alter the command to run frontend tests in order to use ember-exam parallelism feature. It will split randomly the frontend tests into four partitions and run them all in parallel. 

A simple further step to improve performance would be to detect how many core are available on the running machine and split by exactly this number - so it can be fully used.

### DIFF
--- a/testem.js
+++ b/testem.js
@@ -5,6 +5,7 @@
 module.exports = {
     test_page: 'tests/index.html?hidepassed',
     disable_watching: true,
+    parallel: -1,
     launch_in_ci: [
         'Chrome'
     ],
--- a/package-lock.json
+++ b/package-lock.json
@@ -5155,6 +5155,109 @@
       "integrity": "sha1-GWkTUhdlS14nj5/i2dTkm1cgMp4=",
       "dev": true
     },
+    "ember-exam": {
+      "version": "1.0.0",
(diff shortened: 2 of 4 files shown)


JoaoGFarias:
@Turbo87 Requested _testem.js_ update on 4dcc981ea0ab99b5be58001a6fe3a4e77b288054

testem.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
                 '--remote-debugging-port=9222',
                 '--window-size=1440,900'
             ]
-        }
+        },
+        "parallel": -1
```
Turbo87: why `-1`? shouldn't this also be `4` like in the `ember exam` call?

JoaoGFarias: This will default to the max number of browsers needed.
If we change ember exam to 8, 16, 32, it will run the same number of browser.

Turbo87: hmm, if you look at the TravisCI output it still looks like the test chunks are running sequentially though and not in parallel 🤔

Turbo87:
@JoaoGFarias I've rebased the branch and fixed the issue with https://github.com/rust-lang/crates.io/pull/1240#pullrequestreview-90428877. It seems we were both right, `parallel: -1` does work, but it needed to be in the main object, instead of the nested `browser_args` one.

PR #1248, opened 2018-01-28
Try to add links to the index

Eh2406:
This is a companion to https://github.com/rust-lang/cargo/pull/4978
It is adding the links attribute to the index git. I am working on windows so cannot test locally, I.E. this may involve some sparing with the CI to get this green. Advice and help are welcome!

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -30,6 +30,7 @@ pub fn publish(req: &mut Request) -> CargoResult<Response> {
 
     let name = &*new_crate.name;
     let vers = &*new_crate.vers;
+    let links = new_crate.links.clone();
     let repo = new_crate.repository.as_ref().map(|s| &**s);
     let features = new_crate
         .features
@@ -147,6 +148,7 @@ pub fn publish(req: &mut Request) -> CargoResult<Response> {
             features: features,
             deps: git_deps,
             yanked: Some(false),
+            links,
         };
         git::add_crate(&**req.app(), &git_crate).chain_error(|| {
             internal(&format_args!(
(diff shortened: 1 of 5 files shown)


carols10cents:
Ok, seems like we should wait to see how https://github.com/rust-lang/cargo/pull/4978 shakes out before merging this in.

Eh2406:
rust-lang/cargo#4978 was just `r+`ed. so I think this is the next step, well except testing that adding links to the index will not break old versions of cargo.

Eh2406:
Ci is green, what's the next step? Dose this need a test, if so where?

carols10cents:
Tests would be good.

@alexcrichton what are the implications of merging and deploying this change?

Eh2406:
@carols10cents Thanks for your help on IRC. What is the best way to test this? I assumed I would `publish` a crate with a links and check that the index has the links. But, I am not seeing anything that tests the contents of the index to work off of. (Testing is not my strongest suit.)

A little bit of background, https://github.com/rust-lang/cargo/pull/4978 added 2 things to cargo:
1. If the record in the index has a links attribute then the resolver will make sure that each link only appears one time. This will reduce the occurrence of lock files that don't work do to duplicated links attributes.
2. When publishing a crate provide the links attribute. So that crates.io can put it in the index.

This pr puts them together. If cargo provides a links attribute then it adds it to the index.

I have tested that adding a links attribute to the index does not brake old versions of cargo.

Anything else I can answer?

alexcrichton:
@carols10cents we should be all good for deploying this! I don't believe there should be any gotchas in terms of Cargo support to result from this.

carols10cents:
Deployed!

Eh2406:
Thanks for all the help!

PR #1250, opened 2018-02-03
Better mobile styling for the /crates page

someguynamedmatt:
It's a little more space per crate in the vertical, but I think it's a bit easier to read when you're on a smaller screen (phone).

> Before
before_mobile

> After
after_mobile

### DIFF
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -127,21 +127,23 @@
 }
 
 @media only screen and (max-width: 639px) {
-    .white-rows div {
-        margin: auto auto 0.2rem auto;
-        text-align: center;
-    }
-    .white-rows > div .row {
-        flex-flow: column wrap;
-    }
-    .row > div.stats {
-        width: 100%;
-        display: flex;
-        margin: auto;
-    }
-    .crate-link {
-        text-align: center;
-        display: block;
+    #crates {
+        .white-rows div {
+            margin: auto auto 0.2rem auto;
+            text-align: center;
+        }
+        .white-rows > div .row {
+            flex-flow: column wrap;
+        }
+        .row > div.stats {
+            width: 100%;
+            display: flex;
+            margin: auto;
+        }
+        .crate-link {
+            text-align: center;
+            display: block;
+        }
     }
 }
 
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -25,6 +25,10 @@
         .team-info {
             @include flex-direction(column);
         }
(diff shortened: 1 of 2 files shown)


app/styles/crate.scss
```diff
@@ -118,12 +118,40 @@
...
 
+.crate-link {
+    display: inline-block;
+}
+
+@media only screen and (max-width: 639px) {
+    .white-rows div {
+        margin: auto auto 0.2rem auto;
+        text-align: center;
+        .info > [data-test-crate-link] {
```
carols10cents: Hmm, my understanding is that these `data-test` attributes are only for the javascript tests to grab onto, and if we want to style things, we should add more classes for that purpose. Would you be willing to make that change?

someguynamedmatt: Absolutely. I’ll get back to you.

someguynamedmatt:
@carols10cents Should be good now. Thanks for checkin' it out! I know you're busy :)

app/styles/crate.scss
Turbo87: bit late, but wanted to confirm: the `data-test-` attributes are stripped out from production builds so we must not rely on them in the CSS files or any other production code

someguynamedmatt: Yes. It has been removed. Please see the updated code. Thanks!

Turbo87:
@someguynamedmatt could you rebase this PR on the current `master`?

someguynamedmatt:
@Turbo87 Done.

Turbo87:
most of the changes look good to me, but https://percy.io/crates-io/crates.io/builds/621108/view/31510569/375?mode=diff&snapshot=31510569 looks a little strange after this PR 🤔

carols10cents:
Ok, after forgetting about this again for another embarrassing length of time, I rebased on master and let Percy test it again and there's still something strange going on that I don't understand :(

At a width of 375px, *most* of the crate names are centered, but some aren't, and I don't see a good reason why or how to fix it.

Behind the "show details" is the screenshot from Percy; the crates whose names aren't centered in this screenshot are postgres and quickcheck_macros.

<details>

e848de6015a18192840fa3567e11cb6bb26f8444f9b5889ebab4885345422f77

</details>

Given the length of time I've let this languish, and the apparent complexity of this seemingly simple change, I'm going to close this PR. Thank you for your work on this, though. If you would like to work on this again in the future, I would be honored, but I also totally understand if you'd rather not.

PR #1259, opened 2018-02-16
Added a vagrant environment for development

moore3071:
I made a Vagrant environment for this repo after Docker failed to work for me, and I ran into errors in my local environment. I figured that I might as well make a PR for it in case it helps others. I'd also like to point out a few problems with it. If this PR is merged, I'll move the list below into an issue.

- **security** All local connections to Postgres are trusted currently. This should be restricted to only what is necessary. This means either trusting only root, postgres, and vagrant users and/or using some other method such as a password. As such this should only be used in development currently.
- Currently all files are synchronized, and `.env` has to match the values that Vagrant expects to find. This includes database_url, git_repo_url, and git_repo_checkout. `.env` should probably not be synchronized, but we want to keep track of other variables from the host such as gh_client info
- The path to pg_hba.conf in the provisioning file could change when Ubuntu updates their repos for postgres. This potential technical debt should be avoided.
- The provisioning file should be idempotent. It isn't. This is mainly due to the installation of packages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ yarn-error.log
 testem.log
 .env
 docker-compose.override.yml
+.vagrant
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ testem.log
 .env
 docker-compose.override.yml
 .vagrant
+*.log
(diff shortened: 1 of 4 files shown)


carols10cents:

I appreciate your work on this, but I don't generally use Vagrant to work on crates.io. We're having enough trouble maintaining the features we have, and I'm hesitant to add any that I'd have difficulty supporting if folks had issues with it. So for that reason, I'm going to close this PR. Thank you, though!

PR #1262, opened 2018-02-20
Use the new rustfmt-preview component when on CI

jtgeibel:
Use the `rustfmt-preview` component on the `stable` channel when checking formatting on CI.   When developing, proper formatting is now just a `cargo +stable fmt` away!

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,5 +1,1 @@
-combine_control_expr = true
 max_width = 100
-error_on_line_overflow = false
-write_mode = "Overwrite"
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,12 +46,12 @@ matrix:
   include:
     - rust: nightly-2018-02-13
       script:
-        - cargo install --force rustfmt-nightly --vers 0.3.6
-        - cargo fmt -- --write-mode=diff
         - cargo install --force clippy --vers 0.0.186
         - cargo clippy
     - rust: stable
       script:
+        - rustup component add rustfmt-preview
+        - cargo fmt -- --write-mode=diff
         - cargo build
         - cargo test
         - npm install
--- a/config/manifest.js
+++ b/config/manifest.js
@@ -3,22 +3,22 @@
 
 module.exports = function(/* environment, appConfig */) {
     return {
-        name: "Cargo: packages for Rust",
-        short_name: "Cargo",
-        description: "Cargo is the package manager and crate host for Rust.",
-        start_url: "/",
-        display: "standalone",
-        background_color: "#3b6837",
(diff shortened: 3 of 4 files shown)


carols10cents:
I'd love to be on stable rustfmt-preview so that I can use that locally too :)

I'm cool with removing the unstable options, here's why:

> `combine_control_expr = true`

Looks like [true is the default anyway](https://github.com/rust-lang-nursery/rustfmt/blob/42af13245f972d7e9efbeb090dbde17e3ecfaf20/Configurations.md#combine_control_expr) so we should be fine taking this out.

> `error_on_line_overflow = false`

I've definitely wanted this in [other projects](https://github.com/rust-lang-nursery/rustfmt/issues/2268) but it doesn't look like we have any long strings rustfmt can't fix right now? I guess we can try taking this out and seeing how it goes?

> `write_mode = "Overwrite"`

[Overwrite is the default value](https://github.com/rust-lang-nursery/rustfmt/blob/42af13245f972d7e9efbeb090dbde17e3ecfaf20/Configurations.md#write_mode), and this option only applies when none is supplied, and in CI we supply `--write-mode=diff`, so I think this one can be removed.

jtgeibel:
Thanks for researching the remaining options @carols10cents.  I've pushed a commit to remove them.  That just leaves `max_width = 100` as our only configuration!

> switch to stable?

For me `cargo +stable fmt` and `cargo +beta fmt` apply the following reversions to a few places.  I'm fine with making that change, but then we'll have to update again in about 12 weeks.

```
-    #[serde(with = "::util::rfc3339")]
-    pub created_at: NaiveDateTime,
-    #[serde(with = "::util::rfc3339::option")]
-    pub last_used_at: Option<NaiveDateTime>,
+    #[serde(with = "::util::rfc3339")] pub created_at: NaiveDateTime,
+    #[serde(with = "::util::rfc3339::option")] pub last_used_at: Option<NaiveDateTime>,
```

I think it is a good sign that in the past 12 weeks of `rustfmt` development this is the only impact we see.  So things appear to be stabilizing.  So I'm split between:

* Get `cargo fmt` working right now for everyone using a stable compiler for development.  Queue a style PR that we can land when 1.25 hits stable, and see how things go from there.
* Wait for 1.25 to hit stable, see how that compares to nightly at the time, and then decide if `rustfmt-preview` on the stable channel looks stable enough to gate CI on.

jtgeibel:
I've rebased and switched to the `rustfmt-preview` component on `stable`.  We should be able to r+ this in a few days once 1.25.0 is released.  After that, proper formatting is just a `cargo +stable fmt` away!

jtgeibel:
1.25 is releasesed, this should work now.

bors: r+

PR #1263, opened 2018-02-21
Update dependencies

jtgeibel:
This essentially does a `cargo update` with the exception of `hyper` which is held back at 0.11.10.  A deprecation was introduced in 0.11.11.  It looks like `bind_connection` should be replaced
with `serve_connection`, however the API is different.

The two other commits account for:
* A minor change to the GNU header format in `tar`.  It appears that this test is the only one we have that successfully uploads a tar file containing file contents, thus triggering the inclusion of the header.
* A deprecation in `openssl`.

Here is the deprecation error when updating to a newer `hyper`:
```
error: use of deprecated item 'record::hyper::server::server_proto::<impl record::hyper::server::Http<B>>::bind_connection': All usage of the tokio-proto crate is going away.
   --> src/tests/record.rs:126:25
    |
126 |             Http::new().bind_connection(
    |                         ^^^^^^^^^^^^^^^
    |
note: lint level defined here
   --> src/tests/all.rs:1:9
    |
1   | #![deny(warnings)]
    |         ^^^^^^^^
    = note: #[deny(deprecated)] implied by #[deny(warnings)]
```


carols10cents:
The only changes that made me a little nervous were the lettre crate changes, but i deployed to staging and emails worked!

bors: r+

sgrif:
We've been seeing regular segfaults since the deploy which included this commit. (That deploy was https://github.com/rust-lang/crates.io/compare/44d01f04...a2eb1f90). This is the only change in that deploy which could be the culprit. For that reason, I've reverted it.

jtgeibel:
Hey @sgrif, I just wanted to check in and see if you were able to resolve the issues seen in production.  Actual segfaults, ouch!  Did you see any clues in the logs as to which routes may have been related to this?

jtgeibel:
> Either way, in the future, I think it will be best if we do upgrades like this in smaller batches, so it's easier to track down if something goes wrong.

Yeah, I was thinking that as well.  Fortunately it should be easy to land these commits as individual PRs, and then the remaining bulk `cargo update` (in 5b65fd0) can be further split as well.  I've also been meaning to look into https://dependabot.com/ which would help keep us from falling behind again.

In this commit series, I started with the `cargo update` so that I could figure out where work was needed and then it was too easy to just keep adding commits on top.

Issue #1210, opened 2017-12-21
Add an option to order crates by last updated time


carols10cents:
Add to at least the API (and may as well add to the UI as well) the option to sort by last updated.

Crates' last_updated time is changed when a new version is published. This sort should put crates that have had a new version published most recently first. 

This should be as straightforward as adding an ORDER BY; this is *not* a stream of new versions. That is, if this set of events happens:

- 3 days ago: crate A published version 1.0.0
- 2 days ago: crate B published version 0.5.0
- 1 day ago: crate A published version 2.0.0

The list of crates ordered by last updated time should be:

- 1 day ago: crate A
- 2 days ago: crate B

As opposed to:

- 1 day ago: crate A
- 2 days ago: crate B
- 3 days ago: crate A

moore3071:
I'd like to go ahead and tackle this, but I have two clarifying questions first.

1. Has there been any progress on #630. If not would it be possible to get a one-time sanitized snapshot for working on this?
2. Should we discuss breaking down the `Krate::search` function as mentioned [here](https://github.com/rust-lang/crates.io/blob/master/src/krate/search.rs#L29)?

carols10cents:
There has not been any progress on #630, and unfortunately I don't have time in the near future to make a recent sanitized snapshot. 

I think it should be possible to publish just a few crates and versions to your local crates.io to get enough data for testing this out-- instructions are [here](https://github.com/rust-lang/crates.io/blob/master/docs/CONTRIBUTING.md#publishing-a-crate-to-your-local-cratesio). Let me know if there are any problems with the instructions!

If you'd like to try refactoring `krate::search`, please feel free!! I don't want to make that a precondition of this change, though, so it's up to you :)

moore3071:
I unfortunately can't publish any crates to my local crates.io until #1254 is resolved, and therefore won't be able to properly deal with this until that issue is resolved.

Since I'd be hesistant to work on an issue that somebody else has called dibs on, I'd like to say that anyone can go ahead and take on this issue in the meantime

moore3071:
Now that my issue with my local crates.io is resolved, I was able to address this issue. I didn't refactor `krate::search`, but the new functionality is in place along with access to it from the frontend.

jtescher:
@carols10cents can this issue now be closed as well? Seems like the work has been done.

carols10cents:
Sounds good!

PR #1264, opened 2018-02-21
Recently updated sort

moore3071:
Allow sorting by which crates were most recently updated as mentioned in #1210. The option is called 'recent-updates' in order to match the naming convention of 'recent-downloads'

### DIFF
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -362,7 +362,7 @@ fn index_sorting() {
     assert_eq!(json.crates[2].name, "bar_sort");
     assert_eq!(json.crates[3].name, "other_sort");
 
-    //Sort by recent-updates
+    // Sort by recent-updates
     let mut response = ok_resp!(middle.call(req.with_query("sort=recent-updates"),));
     let json: CrateList = ::json(&mut response);
     assert_eq!(json.meta.total, 4);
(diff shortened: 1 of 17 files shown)


carols10cents:
Oh geez I am so sorry to have let this sit for so long. This is actually important for performance reasons for an API user of ours.

@jtgeibel @sgrif could yinz prioritize code review of this please?

sgrif:
Will take care of it when I'm back from RailsConf

sgrif:
Is there a reason we needed to introduce the `time` crate just for this? Could this be done with `std::time::Duration` or `chrono::Duration`?

src/tests/krate.rs
```diff
@@ -282,36 +283,155 @@ fn exact_match_first_on_queries() {
...
-        ::CrateBuilder::new("other_sort", user.id)
+        krate4 = ::CrateBuilder::new("other_sort", user.id)
             .description("other_sort const")
             .downloads(999_999)
             .expect_build(&conn);
+
+        // Set the updated at column for each crate
+        let updated = Utc::now().naive_utc();
+        update(&krate1)
+            .set(crates::updated_at.eq(updated - Duration::weeks(3)))
```
sgrif: We don't need to do this in Rust. This can just be `now - 3.weeks()` (`use diesel::dsl::*`)

carols10cents:
I did a rebase and a force-push to avoid having to merge and then revert the Cargo.toml changes 😅 Sorry for letting this sit so long!

PR #1269, opened 2018-02-27
Add WebAssembly category

fitzgen:
I did not add this under `web-programming` because there is nothing inherently "Web-y" about wasm: it can be used as a general arch-independent binary format, as parity exemplifies.

Examples crates that would use this category (some of which are not actually published on crates.io yet, but the point stands):

* `parity-wasm`
* `wasmi`
* `wee_alloc`
* `wasm-bindgen`
* `wasm-gc`
* `wasm-nm`
* `wasm-snip`
* `svelte`
* `binaryen`
* `wabt`

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -429,6 +429,12 @@ description = """
 Ways to view data, such as plotting or graphing.\
 """
 
+[wasm]
+name = "WebAssembly"
+description = """
+Crates for use when targeting WebAssembly, or for manipulating WebAssembly.\
+"""
+
 [web-programming]
 name = "Web programming"
 description = """


fitzgen:
r? @carols10cents 

If you're not a good person to ask for review, please redirect me to someone who is. Thanks!

jtgeibel:
Great idea @fitzgen!  Looking at your list, here are some ideas I had for possible subcategories.

| Crate | Description | Subcategory |
| --- | --- | --- |
| parity-wasm | WebAssembly binary format serialization/deserialization/interpreter | dev or top-level wasm |
| wasmi |  WebAssembly interpreter | dev or top-level wasm |
| wee_alloc | The Wasm-Enabled, Elfin Allocator | no-std |
| wasm-bindgen | Interoperating JS and Rust code | js |
| wasm-gc | A CLI tool to garbage collect webassembly modules | dev |
| wasm-nm | Print the symbols that are imported in and exported from a wasm file | dev |
| wasm-snip | Replace a wasm function with an `unreachable` | dev |
| svelte | A code size profiler | dev |
| binaryen | Compiler infrastructure and toolchain library for WebAssembly, in C++ | dev (not on crates.io) |
| wabt | The WebAssembly Binary Toolkit (also C/C++) | dev (not on crates.io) |

* web - specific to the web platform
* js - assumes a JS host environment, but not necessarily web
* no-std - needs a better name since some parts of std do work in wasm, but basically crates that make no assumptions about the host
* dev - just like the root development-tools, or maybe put a wasm subcategory under there?

Maybe the js subcategory doesn't make sense yet.  It may be better to wait and see how things evolve beyond the `wasm32-unknonwn-unknown` target before trying to place points on the no-std to web continuum.

fitzgen:
I figured that

* we can just do the initial category for now, and as you say, wait and see how things evolve

* for sub categories like "dev" we can label crates with both the "wasm" category and the "development-tools" category

Happy to defer to crates.io folks' judgement in either case :)

jtgeibel:
Sounds good to me @fitzgen.  I'm also interested in the wasm working group you're leading and I've followed the rust-wasm repo.  I think that would be a good place to discuss potential sub-categories further.

bors: r+

fitzgen:
\o/

carols10cents:
And deployed! https://crates.io/categories/wasm

PR #1277, opened 2018-03-24
Upgrade frontend to Ember 3.0.0

CvX:
Upgrades Ember to version 3.0.0.


Turbo87:
As mentioned on the Ember.js Slack I would prefer it if the PR could be broken apart into multiple PRs to make review a little easier. My suggestion would be to start with the new testing API changes, then the native-dom-helpers and finally the upgrade itself.

package.json
```diff
@@ -46,6 +46,7 @@
     "ember-cli-shims": "^1.2.0",
     "ember-cli-sri": "2.1.1",
     "ember-cli-uglify": "^2.0.0",
+    "ember-cli-update": "^0.15.0",
```
Turbo87: I'd prefer to use the global installation of this, since we don't need to install it for CI systems etc.

CvX: Alright, commit removed.

Turbo87:
@CvX could you extract the ESLint changes to a separate PR too? Ideally all that remains in this PR will be the `ember-source` version bump :)

PR #1282, opened 2018-03-25
Remove deprecated ember-normalize.

CvX:
See: https://github.com/HeroicEric/ember-normalize/pull/19
In its current form it breaks Ember 3.0 builds.

/cc @Turbo87


ember-cli-build.js
```diff
@@ -29,7 +29,12 @@ module.exports = function(defaults) {
         'ember-prism': {
             theme: 'twilight',
             components: highlightedLanguages,
-        }
+        },
+        sassOptions: {
+            includePaths: [
+                'node_modules/normalize.css',
```
Turbo87: where is the `normalize.css` package coming from? don't we have to add a dependency?

Turbo87: nevermind, I'm blind... just found it 🙈

PR #1283, opened 2018-03-25
Cleanup `session` service code

Turbo87:
This PR cleans up the `session` service. It is generally recommended to inject services only into those modules that actually need them. Injecting them unconditionally in an initializer makes them much harder to mock at the same time and causes unnecessary injections.

Partly extracted from https://github.com/rust-lang/crates.io/pull/1279, thanks @CvX!

### DIFF
--- a/tests/acceptance/crate-test.js
+++ b/tests/acceptance/crate-test.js
@@ -182,13 +182,12 @@ test('navigating to the owners page when not logged in', async function(assert) 
 test('navigating to the owners page when not an owner', async function(assert) {
     server.loadFixtures();
 
-    this.application.register('service:session-b', Service.extend({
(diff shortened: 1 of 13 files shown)


tests/acceptance/crate-test.js
```diff
@@ -182,14 +182,13 @@ test('navigating to the owners page when not logged in', async function(assert)
 test('navigating to the owners page when not an owner', async function(assert) {
     server.loadFixtures();
 
-    this.application.register('service:session-b', Service.extend({
+    this.application.__deprecatedInstance__.register('service:session', Service.extend({
```
Turbo87: with the new testing APIs, we will soon be able to replace this with `this.owner.register()` instead. using the private `__deprecatedInstance__` is the official migration path. thanks to @rwjblue for the quick help.

PR #1285, opened 2018-03-25
Remove a bit of unused frontend code

CvX:
* Remove fetchingFeed property - it was only being set, but not read anymore
* Remove `data` and `crates` properties - those two controllers (team, user) have been using only their models for some time now


Turbo87:
thanks!

PR #1287, opened 2018-03-27
Finish MVC Split and Organize Middleware

jtgeibel:
This PR largely finishes the reorganization of code into `models`, `views`, and `controllers` modules.  The following table summarizes the breakdown:

| File in src/ | Model | View | Controller |
| --- | --- | --- | --- |
| dependency.rs | X | X | |
| download.rs | X | X | |
| site_metadata.rs | | | X |
| token.rs | X | X | X |
| version/* | X | X | X|
| user/* | X | X | X |

While splitting `user/*` I created a new module for the CurrentUser middleware.  The rest of the commit series organizes and simplifies the remaining middleware logic and pulls the route building logic into `router.rs`.

The individual commits are straightforward, but the PR has grown rather large.  I can split it in half if that helps with review, but I hope its okay as-is.


carols10cents:
This looks great! I'm excited about all of the detangling you've done here, this organization is so much better!!

bors: r+

PR #1288, opened 2018-03-27
Travis config improvements

CvX:
* Unified indentation
* Merged duplicate keys
* Extracted Ember tests to a separate job (just trying it out, we'll see if it works at all)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,12 @@ branches:
 cache:
   cargo: true
 
-addons:
-  chrome: stable
-
 env:
   global:
-    # See https://git.io/vdao3 for details.
-    - JOBS=1
+    - JOBS=1 # See https://git.io/vdao3 for details.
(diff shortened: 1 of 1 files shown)


jtgeibel:
The cleanups in the first 2 commits look really good, thanks!  I'm also unsure about the final commit, as the clippy run usually takes the most time and there is a lot of overhead (like `cargo install diesel_cli ...`) to running another build.

Turbo87:
I tend to agree. Since we still run `cargo install` even for the Ember job this doesn't look like a win to me 🤔 

The other two commits look good though, if you can remove the third one from the PR I guess we can merge.

CvX:
I've removed the failed speedup experiment and added node_modules caching.

It seems that travis didn't pick up on force-pushed changes since there are still 5 build jobs instead of 4 so I'm closing this PR and creating a new one.

PR #1289, opened 2018-03-27
Update to new hyper API and run `cargo update`

jtgeibel:
Replaced an API that was deprecated in hyper, then updated dependencies.


carols10cents:
bors: r+

PR #1290, opened 2018-03-27
Travis config improvements

CvX:
* Unified indentation
* Merged duplicate keys
* Enabled node_modules caching

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,12 @@ branches:
 cache:
   cargo: true
 
-addons:
-  chrome: stable
-
 env:
   global:
-    # See https://git.io/vdao3 for details.
-    - JOBS=1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,18 +4,19 @@ dist: trusty
...
-  - staging.tmp
+    - staging.tmp
 
 cache:
   cargo: true
-
-addons:
-  chrome: stable
+  directories:
+    - node_modules
```
Turbo87: we should cache `$HOME/.npm` instead (see https://github.com/ember-cli/ember-new-output/blob/v3.0.2/.travis.yml#L14 and https://stackoverflow.com/questions/42521884/should-i-have-travis-cache-node-modules-or-home-npm)

CvX: TIL… 😁
Changed!

Issue #1293, opened 2018-03-28
Using random fixture data makes percy screenshots fail because the text on the page is different :(


carols10cents:
Percy seems awesome! I'm excited!

Looking at the first comparison, there are a bunch of screenshots that differ only in the text that's on the page: https://percy.io/crates-io/crates.io/builds/613691

Looking at why the text is different, it looks like in some of the mirage factories, we're generating data that differs every test run, like the category description that gets set to an arbitrary lorem ipsum sentence: https://github.com/rust-lang/crates.io/blob/c83b3cb1ae56b09eec5ffa1d491bd8fcfb35cd25/mirage/factories/category.js#L17

I think we should get rid of the test data that changes every run so that the screenshots are displaying the same data and we get fewer false failures in percy to manually review.

jtgeibel:
A few other observations from the images:

* Relative timestamps ("x days ago") will change over time if the test data is fixed.  We could probably generate all timestamps relative to the current time so that the display is fixed.
* The download charts are empty, but the grid seems to draw slightly differently between runs.  Maybe adding some data points will help stabilize it.

PR #1295, opened 2018-03-29
Set a constant faker seed for deterministic tests.

CvX:
A crucial fix for visual testing.

Connects to #1293.


carols10cents:
Neat!! Thank you!!

bors: r+

PR #1297, opened 2018-03-30
Update civet

jtgeibel:
This bumps civet, the base HTTP server written in C.  For that reason, it may be best to merge and deploy this separately from other changes so that it can easily be rolled back if there are any issues in production.  I have seen no issues locally or in a staging area on heroku.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,7 +133,7 @@ version = "0.2.2"
 dependencies = [
  "ammonia 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "cargo-registry-s3 0.2.0",
- "chrono 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "chrono 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "civet 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


greysteil:
I think it would be better to run `cargo update -p civet` here - you're updating a lot of packages in this PR, not just `civet`.

(I built [Dependabot](https://dependabot.com) to do exactly this, just added Rust support, and the equivalent PR on my fork is [here](https://github.com/greysteil/crates.io/pull/6/files).)

jtgeibel:
Yes, I know. It is split into 2 commits that do that, but I've combined
them in the same PR. I started looking into some of the questions you asked
in the issue you opened, and hope to follow up later today or tomorrow.

greysteil:
❤️

sgrif:
@jtgeibel For reasons I'm sure you understand, let's limit this to *just* civet for now. Worth noting that the latest version of civet does not build on my machine, and I suspect this is true for all mac users.

jtgeibel:
Unfortunately I don't have a way to test on mac.  I have a few minor cleanups that I may backport to civet 0.8.x.

I've been meaning to investigate a port to hyper as it would be nice to move away from this dependency on C code.  Closing this PR.

sgrif:
Big +1 on moving to hyper

Issue #1296, opened 2018-03-29
error: failed to yank


Deedasmi:
Trying to yank a package today I got a 502 bad gateway. I only published the package a few moments before.

https://pastebin.com/4FqwzaL5

Deedasmi:
Just ran it again the the yank was successful. Going to leave the issue open though, since it was an actual error.

jtgeibel:
To me it looks like the error message is from nginx which is for some reason having an issue connecting to the backend.  What crate were you yanking?  That should help to track this down in the logs.

jtgeibel:
@carols10cents are you able to search the logs to see if there are any references to this?  I think you can search for `crates/rustup/0.1.0/yank`.

carols10cents:
Here's what I'm seeing in the logs for a search for `crates/rustup/0.1.0/yank`:

```
Mar 29 16:00:22 crates-io app/web.1: ERROR 2018-03-29T20:00:21Z: conduit_log_requests: 127.0.0.1:39606 [2018-03-29T20:00:21Z] Delete /api/v1/crates/rustup/0.1.0/yank - 30ms 500: entity not found No such file or directory (os error 2) 
Mar 29 16:00:22 crates-io heroku/router: at=info method=DELETE path="//api/v1/crates/rustup/0.1.0/yank" host=crates.io request_id=b5d46c83-c717-402d-8560-a719436616cb fwd="70.179.168.190" dyno=web.1 connect=0ms service=35ms status=502 bytes=356 protocol=https 
Mar 29 16:00:22 crates-io app/web.1: 2018/03/29 20:00:21 [error] 105#0: *228891 upstream prematurely closed connection while reading response header from upstream, client: 10.153.179.73, server: _, request: "DELETE //api/v1/crates/rustup/0.1.0/yank HTTP/1.1", upstream: "http://127.0.0.1:8888//api/v1/crates/rustup/0.1.0/yank", host: "crates.io" 
Mar 29 16:07:27 crates-io app/web.1:  INFO 2018-03-29T20:07:26Z: conduit_log_requests: 127.0.0.1:38764 [2018-03-29T20:07:26Z] Delete /api/v1/crates/rustup/0.1.0/yank - 2039ms 200 
[...]

PR #1301, opened 2018-04-02
Refactor code that generates JSON error messages for standard HTTP error responses

jtgeibel:
This is a light refactoring of the logic that generates JSON responses for typical HTTP status codes: 400, 403, and 404.  An unused function is also removed.


jtgeibel:
This addresses some odd behavior noted by @sgrif in #1296.

jtgeibel:
I've rebased this on top of the latest master.  The original functionality has been addressed elsewhere, so just a light refactoring remains.

PR #1313, opened 2018-04-04
Update `ember-data` to v2.18.2

Turbo87:
I've tested this using the test suite and manually, and it seems to work fine

/cc @tchak

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1027,22 +1027,16 @@
       "dev": true
     },
     "babel-plugin-feature-flags": {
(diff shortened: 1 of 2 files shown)


Turbo87:
@jtgeibel do you have time to check if this branch works okay on your staging env? it looks okay to me, but I'd like to have some verification before merging this :)

Turbo87:
awesome, let's get this merged then. if things fall apart in prod we can always revert.

PR #1316, opened 2018-04-04
ESLint: fix "ember/no-on-calls-in-components" error.

CvX:
“Don't use .on() in components”

cc: @Turbo87

### DIFF
--- a/app/components/rl-dropdown.js
+++ b/app/components/rl-dropdown.js
@@ -1,13 +1,13 @@
 import Component from '@ember/component';
-import { on } from '@ember/object/evented';
 import { alias } from '@ember/object/computed';
-import { computed, observer } from '@ember/object';
+import { computed } from '@ember/object';
 import $ from 'jquery';
 
 import RlDropdownContainer from './rl-dropdown-container';
 
 export default Component.extend({
     classNames: ['rl-dropdown'],
+    classNameBindings: ['isExpanded:open'],
 
     dropdownContainer: computed(function() {
         return this.nearestOfType(RlDropdownContainer);
@@ -18,14 +18,6 @@ export default Component.extend({
     closeOnChildClick: false,
 
     propagateClicks: true,
-
-    manageVisibility: on('didInsertElement', observer('isExpanded', function() {
-        if (this.get('isExpanded')) {
-            this.$().css('display', 'block');
-        } else {
-            this.$().css('display', 'none');
-        }
-    })),
 
     click(event) {
         let closeOnChildClick = this.get('closeOnChildClick');
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -29,7 +29,6 @@ module.exports = {
             'allowTemplateLiterals': true,
             'avoidEscape': true,
         }],
-        'ember/no-on-calls-in-components': 'off',
     },
     overrides: [
         // node files
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -279,6 +279,7 @@ button.dropdown, a.dropdown {
 }
 
 ul.dropdown {
+    display: none;
     position: absolute;
     top: 100%;
     left: 0;
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -98,19 +98,13 @@ body {
     }
 
     .current-user-links {
-        display: none;
         left: auto;
         right: 0;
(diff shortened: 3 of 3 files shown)


app/components/rl-dropdown.js
```diff
@@ -19,13 +18,18 @@ export default Component.extend({
...
+    },
+
+    manageVisibility: observer('isExpanded', function() {
         if (this.get('isExpanded')) {
             this.$().css('display', 'block');
         } else {
             this.$().css('display', 'none');
         }
-    })),
+    }),
```
Turbo87: I think we should replace this entire observer. We could add `style` to `attributeBindings` and bind `display:block/none` to the `isExpanded` property directly.

app/components/rl-dropdown.js
```diff
@@ -19,13 +19,13 @@ export default Component.extend({
...
+    style: computed('isExpanded', function() {
         if (this.get('isExpanded')) {
-            this.$().css('display', 'block');
+            return 'display: block';
         } else {
-            this.$().css('display', 'none');
+            return 'display: none';
         }
-    })),
+    }),
```
Turbo87: I think we have to use `htmlSafe()` here to avoid warnings 🤔 

```js
return htmlSafe(`display: ${this.get('isExpanded') ? 'block' : 'none'}`);
```

CvX: I wonder if it wouldn't be better to use `.hidden` css class (https://github.com/rust-lang/crates.io/blob/master/app/styles/app.scss#L321) instead of inlining styles.
Maybe in the future we'll want to have a strict Content Security Policy? 🙃

Turbo87: hmm, maybe. but I would guess that things like e.g. `ember-animated` will make that hard anyways.

Turbo87:
@jtgeibel does that conflict with the change that we're doing here? I would expect the current jQuery logic to do pretty much the same thing which should then also conflict with that CSP 🤔

jtgeibel:
Maybe the CSP only blocks injection in the source HTML file. I can test in my staging area later tonight, but on second thought it's probably fine. Feel free to merge before then if you want.

app/components/rl-dropdown.js
CvX: I'm leaving it inline (w/ `htmlSafe`) since it already works. We can come back to it in another issue/PR.

jtgeibel:
With this PR in my staging environment, the Docs drop-down no longer works and I'm seeing the following in the console:

```
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src”). Source: display: none.
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src”).
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src”). Source: display: none.
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src”).
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src”). Source: display: none.
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src”).
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src”). Source: display: block.
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src”). Source: display: none.
```

Looks like we probably need to add 'unsafe-inline' for this approach to work.

CvX:
@jtgeibel dropdowns in the production build also operate on inline styles, just via jquery. Should it be broken there too?

jtgeibel:
@CvX I have no idea.  From http://api.jquery.com/css/#css2

> For example, `$( "#mydiv" ).css( "color", "green" )` is equivalent to `document.getElementById( "mydiv" ).style.color = "green"`.

My guess is that there is a difference between setting individual style properties via JS and setting the style to a string which is then evaluated.

CvX:
Okay, I've updated the PR. Less code is better anyway. 😉

Turbo87:
looks like Percy disagrees now :thinking_face:

jtgeibel:
Looks like all the drop down are open. Wrong initial conditions?

CvX:
That's what happens when you check your changes only on a single dropdown. 🙃 Good thing we have Percy. 😉

Fixed!

Turbo87:
> Good thing we have Percy

💯 👍

PR #1318, opened 2018-04-04
ESLint: fix “ember/no-capital-letters-in-routes” error.

CvX:
“Unexpected capital letter in route's name”

cc: @Turbo87


app/router.js
```diff
@@ -45,7 +45,7 @@ Router.map(function() {
         this.route('index', { path: '/' });
     });
     this.route('category_slugs');
-    this.route('catchAll', { path: '*path' });
+    this.route('catch_all', { path: '*path' });
```
Turbo87: should it be `catch_all` or `catch-all`?

Turbo87: Yeah, I guess that would be good :)

PR #1320, opened 2018-04-04
Update ESLint configuration to match ember-cli blueprint

CvX:
cc: @Turbo87

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -35,4 +35,23 @@ module.exports = {
         'ember/no-capital-letters-in-routes': 'off',
         'ember/new-module-imports': 'off',
     },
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -35,4 +35,23 @@ module.exports = {
...
+                'config/**/*.js',
+                'lib/*/index.js',
+            ],
+            parserOptions: {
+                sourceType: 'script',
+                ecmaVersion: 2015,
+            },
+            env: {
+                browser: false,
+                node: true,
```
Turbo87: I think this means that we can remove all the `/* eslint-env node */` comments in those files

PR #1324, opened 2018-04-04
Update the documentation links to point directly to the new domain

jtgeibel:
This updates the links to point directly to the new documentation on rlo, rather than going through the redirect at doc.crates.io.

I know there was a previous discussion of adding `includeSubDomains` to our `Strict-Transport-Security` header but this was blocked by the old documentation (served via github pages I believe).  I'm assuming that is no longer blocked as the redirect works for me over `https://`.  However, this PR does not change the STS header at this time.

### DIFF
--- a/app/routes/install.js
+++ b/app/routes/install.js
@@ -2,6 +2,6 @@ import Route from '@ember/routing/route';
 
 export default Route.extend({
     redirect() {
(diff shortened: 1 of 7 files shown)


sgrif:
bors: r+

PR #1312, opened 2018-04-04
Switch most recently downloaded summary query to be faster

carols10cents:
Related to #1304.

Because we only want to show the top 10 for /api/v1/summary, move the
LIMIT within the query for the top downloads rather than joining on
everything and then limiting.

This should help alleviate at least some of the slowdowns we were seeing today with the /api/v1/summary call; there are other queries that are taking up some load that will need other mitigation strategies though :(

Here's an explain analyze from the old query on a production dump from today:

```
explain analyze SELECT "crates"."id", "crates"."name", "crates"."updated_at", "crates"."created_at", "crates"."downloads", "crates"."description", "crates"."homepage", "crates"."documentation", "crates"."readme", "crates"."readme_file", "crates"."license", "crates"."repository", "crates"."max_upload_size" FROM (
 "crates" LEFT OUTER JOIN "crate_downloads" ON "crates"."id" = "crate_downloads"."crate_id" AND "crate_downloads"."date" > date(CURRENT_TIMESTAMP - INTERVAL '90 DAYS')
 ) GROUP BY "crates"."id" ORDER BY SUM(crate_downloads.downloads) DESC NULLS LAST LIMIT 10;
                                                                                           QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=81546.14..81546.17 rows=10 width=858) (actual time=546.007..546.009 rows=10 loops=1)
   ->  Sort  (cost=81546.14..81583.65 rows=15004 width=858) (actual time=546.006..546.007 rows=10 loops=1)
         Sort Key: (sum(crate_downloads.downloads)) DESC NULLS LAST
         Sort Method: top-N heapsort  Memory: 37kB
         ->  GroupAggregate  (cost=65223.31..81221.91 rows=15004 width=858) (actual time=273.398..539.672 rows=15004 loops=1)
               Group Key: crates.id
               ->  Merge Left Join  (cost=65223.31..78880.53 rows=438269 width=854) (actual time=273.331..485.878 rows=435230 loops=1)
                     Merge Cond: (crates.id = crate_downloads.crate_id)
                     ->  Index Scan using packages_pkey on crates  (cost=0.29..5950.28 rows=15004 width=850) (actual time=0.009..9.989 rows=15004 loops=1)
                     ->  Materialize  (cost=65223.03..67414.37 rows=438269 width=8) (actual time=273.317..386.191 rows=435229 loops=1)
[...]

PR #1328, opened 2018-04-05
Create a smaller index to be used on the summary page

sgrif:
While #1312 improved that query a lot, we can shave off even more time
by giving it an index containing only the data it's going to need. This
index would need to be refreshed daily by a heroku scheduler task at
00:01 UTC.

I don't have a large enough dataset locally to get a great perf
comparison, so it'd be great to have a test comparison on prod before we
merge this.

The funkiness in how we go about creating/dropping the index is due to
the fact that we can't use bind parameters to create indices, nor can we
use `CURRENT_DATE` directly. If we want to create the index concurrently
(I don't think we need to), then we will have to do string concatenation
in Rust.

### DIFF
--- a/migrations/2018-04-05-192744_create_refresh_todays_recent_crate_downloads/up.sql
+++ b/migrations/2018-04-05-192744_create_refresh_todays_recent_crate_downloads/up.sql
@@ -1,0 +1,9 @@
+CREATE FUNCTION refresh_todays_recent_crate_downloads() RETURNS void AS $$
+BEGIN
+  DROP INDEX IF EXISTS todays_recent_crate_downloads;
+  EXECUTE 'CREATE INDEX todays_recent_crate_downloads
+   ON crate_downloads (date)
+   WHERE date > date(''' || CURRENT_DATE::text || '''::date - INTERVAL ''90 days'')';
+  ANALYZE crate_downloads;
+END;
+$$ LANGUAGE plpgsql;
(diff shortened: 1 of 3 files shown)


migrations/2018-04-05-192744_create_refresh_todays_recent_crate_downloads/up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE FUNCTION refresh_todays_recent_crate_downloads() RETURNS void AS $$
+BEGIN
+  DROP INDEX IF EXISTS todays_recent_crate_downloads;
+  EXECUTE 'CREATE INDEX todays_recent_crate_downloads
+   ON crate_downloads (date)
+   WHERE date > date(''' || CURRENT_DATE::text || '''::date - INTERVAL ''90 days'')';
```
jtgeibel: Does the WHERE clause in the query need to match to ensure the index is used?  Especially regarding `CURRENT_TIMESTAMP` vs `CURRENT_DATE`.  The query is currently:

`WHERE crate_downloads.date > date(CURRENT_TIMESTAMP - INTERVAL '90 days')`

sgrif: I confirmed that the query as it is written today will use this index before opening the PR

sgrif:
#1363 is a better approach

PR #1330, opened 2018-04-07
Use latest Rust in backend Docker image.

CvX:
The `pest` crate (a dependency of `comrak`) requires at least Rust 1.23 (https://github.com/pest-parser/pest#usage).

Currently we test crates.io on Travis using latest stable Rust (and newer).
I don't think there's any reason to pin the Docker image to a specific (and outdated) Rust version.


jtgeibel:
This should pass after merging with master (see #1333).

bors: r+

PR #1331, opened 2018-04-07
Allow feature names to begin with numbers

### LINKED ISSUE #1329: crates.io is incorrectly disallowing feature names that start with numbers
While releasing Diesel 1.2 crates.io blocked me from uploading one of the crates because it has an optional feature called `128-column-tables`, which cargo accepts, but crates.io doesn't. We seem to be assuming that feature names always map to crates, which is not what happens in practice.

sgrif:
Diesel 1.2 had planned on renaming our `large-tables` and `huge-tables`
features to `32-column-tables` and `64-column-tables` respectively,
while also introducing the `128-column-tables` feature. This change was
made several months ago in Diesel. Cargo will happily accept those
as feature names, and resolve them properly from other crates.

However while publishing Diesel 1.2, I ran into a snag mid-release when
I realized that Cargo is incorrectly assuming that a feature name must
be the same as a crate name. I suspect this is an artifact of the fact
that feature names often are crate names (and perhaps in the past that
was the only form of feature?).

However, now they are an entirely separate thing, and we should allow
the same set of feature names that Cargo does. (As an aside, do we want
to apply the same 64 character limit that we apply to crate names to
feature names?)

Fixes #1329.

### DIFF
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -250,24 +250,29 @@ impl Crate {
     }
 
     fn valid_ident(name: &str) -> bool {
-        if name.is_empty() {
-            return false;
-        }
-        name.chars().next().unwrap().is_alphabetic()
+        Self::valid_feature_name(name)
+            && name.chars()
+                .nth(0)
+                .map(char::is_alphabetic)
+                .unwrap_or(false)
+    }
+
+    pub fn valid_feature_name(name: &str) -> bool {
+        !name.is_empty()
             && name.chars()
                 .all(|c| c.is_alphanumeric() || c == '_' || c == '-')
             && name.chars().all(|c| c.is_ascii())
     }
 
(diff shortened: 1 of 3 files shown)


sgrif:
The workaround we had to do (putting an `x` in front of all the feature names) is super ugly, so if we are able to land this somewhat quickly I'd really appreciate it.

alexcrichton:
Crate names needed to not start with a number in the sense that they're Rust identifiers, and IIRC it was just easy to use the same validation for feature names. AFAIK feature names are always specified as strings or not interpreted in an identifier context so it should be fine to allow things like numbers out in front.

carols10cents:
Thanks @alexcrichton! I'm good with this then. Will comment again when I've got this deployed to prod.

bors: r+

carols10cents:
Deployed, and I was able to publish a crate with a feature that starts with a number!

PR #1333, opened 2018-04-08
Don't use Diesel CLI 1.0.0

sgrif:
Diesel 1.0.0 was released with overly broad constraints on internal
dependencies. `diesel_cli` relies on internal implementation details of
`diesel` (and other internal crates) which are not bound to semver.
Unfortunately, we didn't anticipate the problems that came from having a
"semver compatible" bound on these dependencies when relying on semver
excempt APIs.

This was fixed in Diesel 1.0.1. We could probably just switch to 1.2.0,
but this is the more conservative change.

Fixes our build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_install:
   - nvm install 8
 
 install:
-  - cargo install --force diesel_cli --vers 1.0.0 --no-default-features --features postgres && export PATH=$HOME/.cargo/bin:$PATH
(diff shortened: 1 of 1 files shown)


sgrif:
@carols10cents Just FYI, all PRs are going to fail CI until this is merged

Turbo87:
as this looks like a very minimal and reasonable change and I would like to get the PR CI working again I'm gonna go ahead and merge this :)

PR #1335, opened 2018-04-10
Use the correct validator for feature

### LINKED ISSUE #1334: Cannot publish packages that enable features in dependencies (regression)
Hi, I'm trying to publish a crate that contains the following in its `Cargo.toml`

```toml
[features]
nightly = ["radix_51", "subtle/nightly", "clear_on_drop/nightly"]
```

to enable `nightly` features in dependencies.  However, it's no longer possible to publish this crate to `crates.io`, since it fails with:
```
error: api errors: invalid upload request: invalid value: string "subtle/nightly", expected a valid feature name at line 1 column 2315
```

This appears to be the error generated by [this code](https://github.com/rust-lang/crates.io/blob/master/src/views/krate_publish.rs#L126), and the same `Cargo.toml` was accepted up until a few days ago, so my guess is that this is a recent regression?

mark-i-m:
cc @sgrif @CvX @hdevalence

fix #1334

### DIFF
--- a/src/views/krate_publish.rs
+++ b/src/views/krate_publish.rs
@@ -121,7 +121,7 @@ impl<'de> Deserialize<'de> for FeatureName {
 impl<'de> Deserialize<'de> for Feature {
     fn deserialize<D: Deserializer<'de>>(d: D) -> Result<Feature, D::Error> {
         let s = String::deserialize(d)?;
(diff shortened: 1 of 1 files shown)


sgrif:
Thanks. This needs a regression test. Since this is fairly urgent, I've just gone ahead and written one. #1336 includes your commit. Thank you.

PR #1336, opened 2018-04-10
Correct uploading features with `/`s

### LINKED ISSUE #1334: Cannot publish packages that enable features in dependencies (regression)

sgrif:
There was a use of `valid_feature_name` that didn't get changed to `valid_feature`. I've added a regression test that exercises more of the stack to ensure this doesn't happen again in the future.

Fixes #1334

### DIFF
--- a/src/views/krate_publish.rs
+++ b/src/views/krate_publish.rs
@@ -121,7 +121,7 @@ impl<'de> Deserialize<'de> for FeatureName {
 impl<'de> Deserialize<'de> for Feature {
     fn deserialize<D: Deserializer<'de>>(d: D) -> Result<Feature, D::Error> {
         let s = String::deserialize(d)?;
-        if !Crate::valid_feature_name(&s) {
+        if !Crate::valid_feature(&s) {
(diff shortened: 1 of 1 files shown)


sgrif:
bors: r=

sgrif:
bors: r+

sgrif:
@carols10cents @alexcrichton Can one of you deploy as soon as this gets merged?

PR #1338, opened 2018-04-11
Test error handling logic in the router

jtgeibel:
There are 3 main types of errors encountered at the router level:

* Specific HTTP error responses such as `NotFound` and `Unauthorized`
* Human errors, returned as a json response with status 200
* Unexpected errors from libraries and `internal()`

These tests help document the current error handling behavior at the router level of the stack.

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -171,3 +171,61 @@ impl Handler for R404 {
         }
     }
 }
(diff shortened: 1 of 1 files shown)


sgrif:
bors: r+

PR #1344, opened 2018-04-13
Allow database timeouts to be configured

sgrif:
We're consistently seeing the endpoint for crate downloads H12. The logs
are showing that we are *eventually* finishing without error, it's just
taking >30s, so Heroku is killing the connection.

Unfortunately, this means that we don't have visibility into *why* it's
timing out. It's either waiting for a long time for a DB connection,
updating the `version_downloads` table (my guess), or talking to S3. If
it's database related, this will turn those requests into actual errors.

This setting will increase our error rates, as we're seeing the 99th
percentile push up to 5s at some points, and we do have requests that
take 15s to complete (but are completing). Eventually I think we should
be able to consider 10s response times unacceptable, but for now I'm
only going to leave this at 10s for long enough for me to see *what* is
failing, at which point I will set it back to 30s until we no longer see
requests pushing 10s regularly.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3,6 +3,7 @@
 use std::env;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
 
 use diesel::r2d2;
 use git2;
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -71,11 +72,19 @@ impl App {
             _ => 1,
         };
 
+        let db_connection_timeout = match (env::var("DB_TIMEOUT"), config.env) {
+            (Ok(num), _) => num.parse().expect("couldn't parse DB_TIMEOUT"),
```
ashleygwilliams: in the long run we might want to make this error message a little better but it's fine for now

ashleygwilliams:
bors: r+

Issue #1304, opened 2018-04-03
crates.io takes ~6 seconds to load


jonathandturner:
Crates.io used to load almost immediately, but now seems to consistently take around 6 seconds to fully load.  This includes spinners like this:

![image](https://user-images.githubusercontent.com/547158/38257800-1717c8c8-37b5-11e8-8e64-deabb6ae7c78.png)

Not sure if it's just for some people (maybe it's a NZ thing?)

carols10cents:
Does this happen in all browsers? Can you grab a dump of the network tab from the developer tools to see what's hanging exactly?

OvermindDL1:
Loading the css and JS is also taking over 700ms here.  The summary API that @aturon linked is taking 5.52 *seconds* here.  `crates.io` itself took ~500ms.  The images are taking 200-500ms.  In comparison, other sites are loading fine at 20-200ms such as google.com, my personal website hosted 1000 miles away, etc...

sgrif:
I was able to reproduce just a minute ago, but that is no longer the case.

OvermindDL1:
It seems to bounce I've noticed, sometimes speedy fast, sometimes dreadfully slow, it's quite random.

carols10cents:
I have not deployed lately. I don't see any incidents on https://status.heroku.com/. There may be database queries that could use some tweaking, i'm going to investigate that next.

OvermindDL1:
Even the css/js is slowed down at times as well though.

Ahh, it's heroku, I never see servers working well on that service...

carols10cents:
> Ahh, it's heroku, I never see servers working well on that service...

@OvermindDL1 I assure you there are quite a few websites you use that serve at large scales that work perfectly well on heroku; generalizations like that aren't welcome here.

OvermindDL1:
Oh sorry, didn't mean it quite that way!  Just bad experiences with a number of other sites on the same service...  ^.^;

sivakumar-kailasam:
Took me more than a few seconds earlier today and when i checked as i respond here it takes about 800ms. I'm on a 30Mbps connection in India,

<img width="966" alt="screenshot 2018-04-04 23 58 30" src="https://user-images.githubusercontent.com/604117/38326774-482cbd10-3864-11e8-9bc3-f8642d2455a0.png">

carols10cents:
[...]

PR #1345, opened 2018-04-13
Don't lock more than 1 row at a time in `update-downloads`

sgrif:
Prior to this commit, `bin/update-downloads` will hold a lock on every
row it updates for the entire batch of 1000. Each row gets locked when
it is `updated`, but the lock isn't released until the whole transaction
is comitted. This means that the first row in the batch is locked for
the whole transaction, but the last row in the batch isn't locked longer
than it should be.

Processing 1000 rows is routinely taking more than 30 seconds to
complete. Whenever it does, this causes requests to time out. Even when
it doesn't, this locking behavior is the reason that our 99th percentile
response time is getting to be so bad.

The endpoint affected by this is `/crates/{name}/{version}/download`. In
other words, this bug is causing `cargo install` to fail. Even when it
doesn't, this bug is causing it to take seconds when it should take
milliseconds.

By establishing a transaction per row instead of per batch, we retain
atomicity, without holding a lock for the whole batch. To keep
everything atomic, we need to update the global counter per row instead
of per batch.

This relies on the invariant that no rows in `version_downloads` get
touched once `date != CURRENT_DATE` other than by this script.

I believe this is also the root cause of #1304, and that we can maybe
revert #1312

### DIFF
--- a/src/bin/update-downloads.rs
+++ b/src/bin/update-downloads.rs
@@ -30,19 +30,17 @@ fn main() {
 
 fn update(conn: &PgConnection) -> QueryResult<()> {
     use version_downloads::dsl::*;
+
     let mut max = Some(0);
     while let Some(m) = max {
-        conn.transaction::<_, diesel::result::Error, _>(|| {
-            let rows = version_downloads
-                .filter(processed.eq(false))
-                .filter(id.gt(m))
-                .order(id)
-                .limit(LIMIT)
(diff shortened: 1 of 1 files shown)


sgrif:
bors: r+

sgrif:
bors: r-

(clippy lint failed, CI passed otherwise, just going to manually merge after fixing)

PR #1347, opened 2018-04-13
Add the ability to blacklist IPs

sgrif:
We're seeing some suspicious traffic that we think is the causing enough
load to be the source of our overall performance problems.

### DIFF
--- a/src/middleware/blacklist_ips.rs
+++ b/src/middleware/blacklist_ips.rs
@@ -1,0 +1,49 @@
+//! Middleware that blocks requests from a list of given IPs
+
+use super::prelude::*;
+
+use std::io::Cursor;
+use std::collections::HashMap;
+use std::net::IpAddr;
+
+// Can't derive debug because of Handler.
+#[allow(missing_debug_implementations)]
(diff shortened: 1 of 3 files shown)


src/middleware/blacklist_ips.rs
```diff
@@ -0,0 +1,51 @@
...
+impl AroundMiddleware for BlockIps {
+    fn with_handler(&mut self, handler: Box<Handler>) {
+        self.handler = Some(handler);
+    }
+}
+
+impl Handler for BlockIps {
+    fn call(&self, req: &mut Request) -> Result<Response, Box<Error + Send>> {
+        if self.ips.contains(&req.remote_addr().ip()) {
+            let body = "Due to unusual activity from your IP, it has \
```
ashleygwilliams: let's not say that we have blocked the IP. let's just say "We are unable to process your request at this time, please open an issue at <link>" and can we give them some identifying info to file with? an ID would be great

ashleygwilliams:
bors: r+

sgrif:
bors: r-

master broken, manually merging

Issue #1327, opened 2018-04-05
Bad cargo error message for adding owners if GitHub token is invalid


carols10cents:
Moved here from rust-lang/cargo#5297.

What happened:

Niko tried to add a team as a github owner for the lalrpop crate:

```bash
>  cargo owner --add github:lalrpop:publish
    Updating registry `https://github.com/rust-lang/crates.io-index`
error: failed to invite owners to crate lalrpop: failed to get a 200 OK response, got 502
headers:
    HTTP/1.1 100 Continue



    HTTP/1.1 502 Bad Gateway

    Connection: keep-alive

    Server: nginx

    Date: Thu, 05 Apr 2018 08:47:57 GMT

    Content-Type: text/html

    Content-Length: 166

    Strict-Transport-Security: max-age=31536000

    Via: 1.1 vegur



body:
<html>
<head><title>502 Bad Gateway</title></head>
<body bgcolor="white">
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

In the papertrail logs, I was able to see:

```
Apr 05 04:37:33 crates-io app/web.2: ERROR 2018-04-05T08:37:33Z: conduit_log_requests: 127.0.0.1:51462 [2018-04-05T08:37:33Z] Put /api/v1/crates/lalrpop/owners - 70ms 500: didn't get a 200 result from github, got 401 with: {"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"} didn't get a 200 result from github, got 401 with: {"message":"Bad credentials","documentation_url":"https://developer.github.com/v3"} 
Apr 05 04:37:33 crates-io app/web.2: 2018/04/05 08:37:33 [error] 100#0: *132269 upstream prematurely closed connection while reading response header from upstream, client: 10.69.109.160, server: _, request: "PUT //api/v1/crates/lalrpop/owners HTTP/1.1", upstream: "http://127.0.0.1:8888//api/v1/crates/lalrpop/owners", host: "crates.io" 
Apr 05 04:37:33 crates-io heroku/router: at=info method=PUT path="//api/v1/crates/lalrpop/owners" host=crates.io request_id=e9608c5b-92b9-4e6b-ac32-ce5f5c0ba811 fwd="209.6.50.191" dyno=web.2 connect=1ms service=90ms status=502 bytes=356 protocol=https 
[...]

PR #1349, opened 2018-04-13
Disable `conduit_log_requests` in production

sgrif:
Everything it outputs is redundant with heroku's router logging. We're
hitting our usage limits for log storage, so let's cut them down by
half.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -48,7 +48,7 @@ pub fn build_middleware(app: Arc<App>, endpoints: R404) -> MiddlewareBuilder {
         m.add(DebugRequest);
     }
 
-    if env != Env::Test {
+    if env == Env::Development {
         m.add(LogRequests(log::LogLevel::Info));
     }
 


jtgeibel:
I definitely support getting rid of the logging at this level since it includes a lot of redundant data when requests succeed.  Before we can do that, I think we need to first add logging of error results elsewhere as there have been some recent issues where this log output helped identify the root cause:

* 1296 - [500: entity not found No such file or directory (os error 2)](https://github.com/rust-lang/crates.io/issues/1296#issuecomment-377712934)
*  1327 - [500: didn't get a 200 result from github, got 401 with: ...](https://github.com/rust-lang/crates.io/issues/1327)

I've added a commit to #1301 which will log any unexpected errors propagated up the middleware stack.  Additionally it includes the request id provided by Heroku when the `X-Request-ID` header is present.

sgrif:
Heroku's built in logging includes both of those from what I can see.

jtgeibel:
For the log snippets included in those issues, the line provided by conduit_log_requests is the only one including the details of the error as far as I can see.  I don't have access to the full logs so maybe I'm missing something.

sgrif:
Ironically I just got paged for a spike in error rates, and you're right, heroku's router doesn't have the response body. I'm just going to change our log level to error instead.

jtgeibel:
> I'm just going to change our log level to error instead.

I'm not quite sure I follow.  Are you saying production was using something like `RUST_LOG=info` before?

sgrif:
It was. It isn't now

jtgeibel:
Okay, makes sense.  I've opened https://github.com/conduit-rust/conduit-log-requests/issues/4 as I think including the request ID from Heroku would be the last advantage to adding our own logging middleware for catching unexpected errors.

PR #1351, opened 2018-04-14
Don't re-index crates search all the time

sgrif:
I'm seeing a lot of this in the logs:

> Words longer than 2047 characters are ignored.
>  PL/pgSQL function trigger_crates_name_search() line 8 at assignment

It looks like the update-downloads script is causing this trigger to run
when it increments the download count. This restricts the trigger so it
only runs when the `updated_at` column is changed (which does not happen
when only the downloads column is touched)


jtgeibel:
Looks good to me.

bors: r+

PR #1352, opened 2018-04-14
Don't query `reverse_dependencies` on crates/show

sgrif:
This endpoint is one of our slowest, we don't need to hit it every time
someone visits this page.


jtgeibel:
Nice catch!

bors: r+

PR #1353, opened 2018-04-14
Ensure we return a well formed 500 on errors

sgrif:
rust-civet basically vomits if you ever return it an error. The
resulting response would cause nginx to flip out, and then for Heroku to
give us an H13 or otherwise respond with a 502 error.

This ensures that we always return an `Ok` response, no matter what.

### DIFF
--- a/src/middleware/ensure_well_formed_500.rs
+++ b/src/middleware/ensure_well_formed_500.rs
@@ -1,0 +1,29 @@
+//! Ensures that we returned a well formed response when we error, because civet vomits
+
+use super::prelude::*;
+
+use std::collections::HashMap;
+
+// Can't derive debug because of Handler.
+#[allow(missing_debug_implementations)]
+#[derive(Default)]
+pub struct EnsureWellFormed500;
+
+impl Middleware for EnsureWellFormed500 {
+    fn after(
+        &self,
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
the errors here are only clippy and cargo fmt; i am in general in favor of this so merge + deploy at will

jtgeibel:
I would prefer to see #1301 merged, which addresses the same issue. I also
exclude the new middleware in the test environment so that any errors there
don't become generic.

I do return a json encoding of the error, where as this returns a general
error. I could go either way on that.

sgrif:
We definitely should never be including any details about the error in the response body in production. We already have that information logged for debugging.

sgrif:
You're right that we should exclude this from the test environment, I'll make sure to do that. Happy to discuss further changes to this in #1301, as it looks like that addresses other behaviors, right now I'm just trying to cut down on the immediate log noise that is making debugging difficult.

jtgeibel:
Yeah, I understand. I won't have access to my laptop until later tonight so
do whatever you need to track down the production issues. I can always
rebase the other cleanups on top of this.

sgrif:
:+1: thanks for understanding.

PR #1354, opened 2018-04-14
Don't use shields.io's crate badges

sgrif:
Every single time someone loads a crates.io badge from shields.io, they
make a separate API call to our servers for each one, with no caching.
This means that right now when you go to the search page, you're going
to make 11 API calls when you should be making 1. The result is that shields.io
accounts for an unreasonably large amount of traffic to our systems.

We have all the information we need for this badge when we load the page,
so let's just do that and use their "badge with these words" API instead.

I've just removed it from the version show page since we already display
that info elsewhere on the page, and it's especially confusing when the
numbers don't match.

### DIFF
--- a/app/components/crate-badge.js
+++ b/app/components/crate-badge.js
@@ -1,0 +1,16 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+
+export default Component.extend({
+    classNames: ['vers'],
+
+    tagName: 'span',
+
+    color: computed('crate.max_version', function() {
+        if (this.get('crate.max_version')[0] == '0') {
+            return 'orange';
+        } else {
+            return 'blue';
+        }
+    }),
+});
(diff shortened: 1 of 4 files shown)


sgrif:
bors: r+

Turbo87:
> I'm unable to access it. Can you fix that?

I've sent you an invite. Once you've accepted can you check that the changes in https://percy.io/crates-io/crates.io/builds/656071 look correct to you?

Sidenote: The reason percy did not run for this PR was that our trial period had ended and we just got upgraded to a sponsored account a few hours ago, so hopefully we won't have that problem again :)

sgrif:
Oh hey, there's a bug!

CvX:
cc: @timhaines (have you considered a public read-only access option for projects on Percy?)

timhaines:
@CvX we would like to have a public read-only mode, and have had a few projects (including some large commercial ones) ask for it.  It's fairly tricky though, so is going to be a large project we plan for in the future.

Turbo87:
> we would like to have a public read-only mode

whoops, I was under the impression that read-only mode was already public 🙈 

I agree, that that is not a perfect situation... on the other hand having that feedback on the PRs for the maintainers is quite valuable so I'm a little hesitant to turning it off completely 🤔 

As @timhaines said, the percy feedback is not failing the PR, and bors will still merge fine even if percy is failing.

@timhaines you don't happen to have a rough estimate of when public read-only might become a thing?

PR #1355, opened 2018-04-14
Fix badges for crates with prerelease versions

sgrif:
Dashes are used to separate sections, and need to be escaped. I had
thought that `crate.max_version` never included prereleases, but I was
wrong.


sgrif:
bors: r+

sgrif:
(Pushing through since this is currently a bug in prod)

PR #1356, opened 2018-04-15
Add ember-cli-code-coverage.

CvX:
Run `npm run test-coverage` to get the code coverage report.

This PR uses a pre-release version of ember-cli-code-coverage (current master, git ref 77b5b92) which includes the latest fixes. We'll update to a new release as soon as it's available.

cc: @Turbo87

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -1,0 +1,3 @@
+{
+    "temp-directory": "coverage"
+}
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "start:live": "ember serve --proxy https://crates.io",
     "start:local": "ember serve --proxy http://127.0.0.1:8888",
     "start:staging": "ember serve --proxy https://staging-crates-io.herokuapp.com",
-    "test": "ember exam --split=2 --parallel"
+    "test": "ember exam --split=2 --parallel",
+    "test-coverage": "COVERAGE=true npm run test && ember coverage-merge && rm -rf coverage_* coverage/coverage-summary.json && nyc report"
   },
   "devDependencies": {
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
@@ -33,6 +34,7 @@
     "ember-cli-babel": "^6.8.2",
     "ember-cli-cjs-transform": "^1.1.1",
     "ember-cli-clipboard": "^0.9.0",
+    "ember-cli-code-coverage": "kategengler/ember-cli-code-coverage#77b5b92",
     "ember-cli-dependency-checker": "^2.0.1",
     "ember-cli-eslint": "^4.2.1",
     "ember-cli-htmlbars": "^2.0.3",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -69,6 +71,7 @@
     "eslint-plugin-ember": "^5.1.0",
     "loader.js": "^4.5.1",
     "normalize.css": "^7.0.0",
+    "nyc": "^11.6.0",
```
Turbo87: do we need that only for the command line reporting? can't we use a built-in command line reporter?

.nycrc
```diff
@@ -0,0 +1,3 @@
+{
+    "temp-directory": "coverage"
+}
```
Turbo87: we don't seem to be using this in our other projects that use code coverage. what's the reason for having this file?

package.json
```diff
@@ -22,7 +22,8 @@
     "start:live": "ember serve --proxy https://crates.io",
     "start:local": "ember serve --proxy http://127.0.0.1:8888",
     "start:staging": "ember serve --proxy https://staging-crates-io.herokuapp.com",
-    "test": "ember exam --split=2 --parallel"
+    "test": "ember exam --split=2 --parallel",
+    "test-coverage": "COVERAGE=true npm run test && ember coverage-merge && rm -rf coverage_* coverage/coverage-summary.json && nyc report"
```
Turbo87: why do we `rm -rf` here?

package.json
CvX: Problem here is that we use ember-exam to have parallel tests, so the process is as follows:

1. run ember tests with `COVERAGE=true` env-var (and `parallel: true` option in config/coverage.js)
2. do `ember coverage-merge` to merge the results from each parallel test run into a single './coverage' directory
3. display the code coverage results. here `nyc` is utilized. do you know how to display the results with the text reporter using only ember-cli-code-coverage/istanbul?

.nycrc
CvX: By default nyc reads coverage data from it's own temp directory (`./.nyc_output` [istanbul/nyc lib/config-util.js:215](https://github.com/istanbuljs/nyc/blob/2030e0c6721f574500a745277334b8bf20d58ccb/lib/config-util.js#L215)), while ember-cli-code-coverage writes to `./coverage`.

package.json
CvX: ember-cli-code-coverage, when called with `--parallel` option, outputs to multiple `coverage_[hash]` directories. These pile up quickly, since there new ones are created after each test run. Hence `coverage_*` removal.

`coverage-summary.json` is removed to prevent nyc issue (see: https://github.com/kategengler/ember-cli-code-coverage/pull/173#issuecomment-381274578)

CvX:
@Turbo87 "we should probably make npm run test-coverage the default for CI runs too 🤔"

Sure. It won't hurt to have coverage reports in the travis build logs, and it will be required when we'll integrate with an external code coverage tool (CodeClimate, Codecov, Coveralls, etc.)

PR #1359, opened 2018-04-16
Use /raw/ rather than /blob/ URLs for images

kornelski:
Images in READMEs don't render ([example](https://crates.io/crates/imgref)), because relative image URLs are converted to a URL to an HTML page about the image, rather than to the image data.

This fixes image embedding by changing `/blob/` to `/raw/` part of the URL for image URLs. The images are detected by file extension in the URL, because I haven't found a way to distinguish when URLs are generated for `` markup and other URLs.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -144,7 +144,7 @@ impl<'a> MarkdownRenderer<'a> {
                 let offset = new_url.len() - 5;
                 new_url.drain(offset..offset + 4);
             }
-            // Assumes GitHub's URL scheme. GitHub renders text and markdown
+            // Assumes GitHub’s URL scheme. GitHub renders text and markdown
             // better in the "blob" view, but images need to be served raw.
             new_url += if is_media_url(url) {
                 "raw/master"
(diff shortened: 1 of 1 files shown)


kornelski:
I don't understand Travis failure. It's fails only on stable, and only "errors" are about rustfmt disliking long lines, which *were there before*.

sgrif:
Have you tried rebasing onto master?

sgrif:
The lines it's complaining about are definitely lines you've added in this PR

kornelski:
Even if I fix my line, it still breaks because of lines already in master.

Rustfmt is such a clumsy bulldozer :(

src/render.rs
```diff
@@ -133,7 +144,13 @@ impl<'a> MarkdownRenderer<'a> {
                 let offset = new_url.len() - 5;
                 new_url.drain(offset..offset + 4);
             }
-            new_url += "blob/master";
+            // Assumes GitHub's URL scheme. GitHub renders text and markdown
```
jtgeibel: It looks like removing the apostrophe here fixes things.  It is also strange that these are reported as errors, even though we invoke with `cargo fmt -- --write-mode=diff` which should provide a diff.

Trying with rustfmt on beta also seems to work fine, so I don't think we need to report this upstream as it seems to be resolved there.

Maybe just reword a bit so that it can land now?

kornelski: OK, thanks

jtgeibel:
This looks good to me.  I've also confirmed that GitLab and Bitbucket follow the same blob vs raw convention.

bors: r+

PR #1362, opened 2018-04-23
Log more information

sgrif:
This introduces a new logger meant to replace both the heroku router
log output, and `conduit-log-requests`. It roughly follows the format
that the heroku router uses. It includes most information that we care
about from there, while excluding certain things we never care about
(response size in bytes [which we can't measure here], host, and
protocol specifically).

This logs some additional information that we want to be able to see,
such as user agent and referer. It also logs errors in the case of a
500, eliminating the need for conduit-log-requests. Unlike our debug
log code, this outputs everything on a single line, which is crucial for
dealing with production logs.

This also changes an nginx config to have it not mess with the
X-Forwarded-For header. Heroku always sets this header to the IP we care
about, and having this config just causes nginx to change it to
"{real_ip},{heroku_internal_routing_ip}", which we don't want.


src/middleware/mod.rs
```diff
@@ -55,7 +54,7 @@ pub fn build_middleware(app: Arc<App>, endpoints: R404) -> MiddlewareBuilder {
     }
 
     if env != Env::Test {
-        m.add(LogRequests(log::LogLevel::Info));
+        m.around(log_request::LogRequests::default());
```
jtgeibel: With around middleware, you may want to move this to the bottom of the file to ensure that it is the outermost around wrapper.  Otherwise, this won't log blacklisted ips or the static index.html served on non-api endpoints.

Alternatively, I think this can be converted to before/after Middleware.  If I understand correctly, the around middleware should only be necessary when we want to rewrite the request or to short-circuit other AroundMiddleware or the innermost handler by returning a response early.  A before handler can short-circuit by returning an `Err(_)`, but cannot return early with a response.  Here we only need to capture the start time on the way down the middleware stack.

sgrif:
bors: r+

PR #1363, opened 2018-04-24
Make recent downloads fast

sgrif:
Every aspect of crates.io is extremely fast except one. Calculating
recent downloads is a moderately expensive operation that we are doing
too often. It's not "slow" per-se. The query takes 500ms to complete on
average, which is not unreasonable.

However, this query does put a good bit of load on the DB server. The
root cause of #1304 was an irresponsible bot hitting /crates (which
includes recent downloads) enough that it caused our DB server's CPU
load to reach 100%. We're on a very low tier database server right now,
so we could fix this by just upgrading to the next tier if we wanted to.
However, as the number of crates grows, this problem will just get
worse.

Recent downloads appear in 3 places. Summary, search, and crate details.
Summary and search are the two endpoints that crawlers want to hit, so
we can't have slow queries on those pages.

There are other solutions we could take here. We could just not include
recent downloads on those pages, and include them only if a flag is set.
We could also move recent downloads to another endpoint which gets hit
separately. Both of these require an annoying amount of code though, and
we need a crawler policy that says "don't hit these endpoints", which
requires work for us to monitor those endpoints, etc.

Ultimately we don't need this data to be real time (it already is
slightly delayed anyway). We can just store this particular piece of
data in a materialized view, which we refresh periodically. Right now
I'm refreshing it whenever update-downloads runs, but I think we can go
even further and update it daily if this becomes a scaling problem in
the future.

Even though repopulating the view currently only takes ~500ms (same
amount of time as the query), I've opted to do it concurrently to avoid
locking the table for long periods of time as we continue to scale.

Crate details is still using the `crate_downloads` table, so it's a bit
closer to real time. When we're selecting this for a single crate, the
query is already fast enough. We may want to change this in the future.
We could also consider getting rid of `crate_downloads` entirely, and
populate the view from `version_downloads`. This will cause more CPU
load when we refresh the view, but it probably doesn't matter.

### DIFF
--- a/migrations/2018-04-24-145128_create_recent_crate_downloads/down.sql
+++ b/migrations/2018-04-24-145128_create_recent_crate_downloads/down.sql
@@ -1,0 +1,3 @@
+DROP FUNCTION refresh_recent_crate_downloads();
+DROP INDEX recent_crate_downloads_crate_id;
+DROP MATERIALIZED VIEW recent_crate_downloads;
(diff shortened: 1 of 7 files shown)


ashleygwilliams:
bors: r+

sgrif:
I can confirm it is indeed now fast.

PR #1364, opened 2018-04-24
Include query string in logs

sgrif:
Conduit doesn't provide any methods that give us the path and query
string if present, so we have to do some additional junk to make that
happen ourselves.


ashleygwilliams:
bors: r+

PR #1365, opened 2018-04-24
Tag slow requests in our logs

sgrif:
We occasionally see max response time spike to 10s or greater. I want to
be able to know exactly what that request was. Unfortunately, the time
that gets logged appears to be slightly different from what appears in
our metrics. I'd like to be able to find these requests in the logs more
easily.


jtgeibel:
bors: r+

PR #1367, opened 2018-04-24
Add config files for Diesel 1.3

sgrif:
This is mostly to help me test out some changes to Diesel in crates.io.
One of the big annoyances of this repo is having to re-run `diesel
print-schema` whenever a migration is added, and then modify `schema.rs`
to re-apply the differences between what we want and what Diesel
generates.

Most of those changes can now be automated, and the rest can be placed
in a `.patch` file which gets automatically applied.

This change will have no effect on people using Diesel 1.2 or earlier
(e.g. everyone except me since 1.3 isn't out)


jtgeibel:
This is a nice cleanup, looks good to me!

bors: r+

PR #1369, opened 2018-04-25
`cargo +beta fmt`

jtgeibel:
This PR prepares us for the next stable release of rust which will pull
in a new version of the rustfmt-preview component.  These changes have
been stable since at least April 11th, so I don't expect any changes
between now and the next release.

The changes consist of sorting use statements in alphabetical order,
and some minor tweaks to the formatting of macro definitions.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,8 +2,8 @@ extern crate diesel;
 extern crate diesel_migrations;
 extern crate dotenv;
 
-use diesel_migrations::run_pending_migrations;
 use diesel::prelude::*;
+use diesel_migrations::run_pending_migrations;
 use dotenv::dotenv;
 use std::env;
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -5,10 +5,10 @@ use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
+use curl::easy::Easy;
 use diesel::r2d2;
 use git2;
 use oauth2;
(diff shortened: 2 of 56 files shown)


jtgeibel:
We could definitely wait until closer to the release to merge.  Just trying to get out ahead of CI breakage, but there is also nothing in place to keep us from introducing formatting regressing over the next two weeks.

sgrif:
In an ideal world this wouldn't churn at all

jtgeibel:
These reordering options were recently changed to on by default but it looks like I can't disable them via configuration (the config option is nightly only).  I've filed rust-lang-nursery/rustfmt#2651 to see if the config option can be made stable now that it defaults to on.

jtgeibel:
@sgrif do you think we should merge this as-is (once I rebase)?  The other option I see is to switch back to a fixed nightly date (like we do for clippy) so that we can use the unstable config options to avoid reordering use statements for now.  I think there will be a bit of churn every 6 weeks on rustfmt-preview on the stable channel, but it is also really nice to know that `cargo +stable fmt` is always the go-to command to before committing.

sgrif:
We actually wouldn't even need to switch to a fixed nightly date anymore. rustfmt-nightly compiles on stable. But yeah, I agree that let's just stick with `cargo +fmt stable`. So I think you should merge this once 1.26 is stable

jtgeibel:
1.26 is out. 
bors: r+

PR #1371, opened 2018-04-26
Fix relevance sorting on search

### LINKED ISSUE #1370: relevance sort for search results is sorting alphabetically
After a search `https://crates.io/search?q=...`, when the relevance sort is selected, the crates are sorted alphabetically instead.

sgrif:
To sort by relevence, the backend expects the sort string to either be
not present, or set to the value "relevance". Right now the backend is
either sending the string `null` or an empty string. Neither of these do
what we want.

Fixes #1370


ashleygwilliams:
bors: r+

PR #1372, opened 2018-04-26
Do less work in update-downloads

sgrif:
I was noticing that query to increment the downloads column on `crates`
was getting run way more than I'd expect it to be. I also was seeing
logs about a lock contention issue with that query. Specifically, those
logs pointed me at one of two things wrong here.

The first is that we were running this query even if we were
incrementing the value by 0. Second, we have our cutoff set to 2 days
ago. I think the reason for both of these is that we wanted to make sure
the `processed` column got set to true. We can pretty easily avoid this
by just setting it to true for all rows at the very end when all our
preconditions are met.

This doesn't necessarily fix the lock contention problem, but it does
result in hitting far fewer rows, which should alleviate the problem.

### DIFF
--- a/src/bin/update-downloads.rs
+++ b/src/bin/update-downloads.rs
@@ -1,7 +1,6 @@
 #![deny(warnings)]
 
 extern crate cargo_registry;
-extern crate chrono;
 #[macro_use]
 extern crate diesel;
 
@@ -30,6 +29,7 @@ fn main() {
 }
 
 fn update(conn: &PgConnection) -> QueryResult<()> {
+    use diesel::dsl::now;
     use diesel::select;
     use version_downloads::dsl::*;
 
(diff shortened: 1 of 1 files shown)


src/bin/update-downloads.rs
```diff
@@ -38,13 +38,23 @@ fn update(conn: &PgConnection) -> QueryResult<()> {
...
             .load(conn)?;
         collect(conn, &rows)?;
         max = rows.last().map(|d| d.id);
     }
 
+    // Anything older than 24 hours ago will be frozen and will not be queried
+    // against again.
+    diesel::update(version_downloads)
+        .set(processed.eq(true))
+        .filter(date.lt(diesel::dsl::date(now)))
```
jtgeibel: The comment says older than 24 hours ago, but I read this more as "yesterday or older in UTC".  The behavior itself seems fine to me, but the comment confused me a bit at first.  (As you mentioned the old behavior was 2 days ago, so the comment was confusing then as well.)

Other than that, I'm r+.  This logic seems clearer and more efficient now and I think I finally understand how this portion of the logic works!

sgrif: Yeah, 24 hours is kinda wrong, but this matches what records will actually ever be touched again. After midnight UTC we never update old records.

sgrif:
bors: r+

PR #1373, opened 2018-04-27
Remove some unused indexes

sgrif:
The gh_id index has never been read at all. reserved_crate_names is too
small of a table for indexes to be useful. I've manually verified that
every query hitting version_downloads either doesn't use these indexes,
or has the same performance when not using them.


jtgeibel:
Looks good to me.  bors: r+

jtgeibel:
bors: r+

PR #1376, opened 2018-05-01
Update test in the update-downloads binary to work at all hours

jtgeibel:
The existing 2 hour offset causes tests failures between 0:00 and 2:00
UTC.  During this time period, the generated row is given a date of
"yesterday" and the new processing logic marks the row as processed.


sgrif:
bors: r+

PR #1378, opened 2018-05-01
Replace `civet` with `hyper` as http server

jtgeibel:
This is a work in progress implementation switching from `civet` to `hyper` as the base http server.

This is based on `conduit-hyper` which I've updated to `hyper 0.11`.  That source code is [here](https://github.com/jtgeibel/conduit-hyper/blob/master/src/lib.rs).  A huge shout-out to @sfackler and @drbawb as this is forked from their repos.  I'm curios if either of you are still using this code and if you have any feedback.  I plan to eventually publish this crate on crates.io, so please let me know if you would like to be added/removed as an author or if you have any objections.

There are several things I wish to investigate further, including:

* I'm not yet sure if hyper handles panics.  I see no results searching for `catch_unwind`, so I'm assuming panics need to be handled in this layer.
* Add some tests
* request.remote_addr() is deprecated, also see: https://github.com/hyperium/hyper/issues/1410#issuecomment-356115678

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,7 +29,7 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "html5ever 0.21.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "lazy_static 0.2.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "maplit 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "matches 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "tendril 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


jtgeibel:
Status update: I've rewritten `conduit-hyper` to be based on `hyper 0.12` which should be entering beta soon.  (The change isn't yet reflected in this PR.)  We use hyper in our test infrastructure, and I'm working on a separate PR to get us up to date there as well.

There are a few notes in the [README](https://github.com/jtgeibel/conduit-hyper#conduit-hyper) file but none of them should affect us here.  For instance, we don't care about the IP address of the request because it will always be the nginx reverse proxy so I haven't bothered to add the infrastructure to pull the connection IP address out of hyper.

sgrif:
We very much do care about the IP address of the request. It is used for logging as well as for blacklisting.

sgrif:
Oh I see what you're saying. We don't care about the remote addr, we're just pulling it from the x-forwarded-for header

jtgeibel:
Okay, I've published the `conduit-hyper` crate and updated this PR to use the released version.  More information can be found in the [README].  This allows us to remove a dependency on C code that we haven't updated in a long time.

I do have a [PR] against the crate that catches panics and converts them to a status 500 response, however the default hyper behavior of simply closing the connection on a panic matches our current `civet` backend.

Similar to the EnsureWellFormed500 middleware, `conduit-hyper` will turn application error results into a status 500 response.  I've decided to leave the existing middleware in place for now, because I expect we will eventually migrate away from `conduit` (and `conduit-hyper`) to a more complete web framework and I think it makes sense to defer such changes until then.

This PR should now be ready for review.

[README]: https://github.com/jtgeibel/conduit-hyper/#conduit-hyper
[PR]: https://github.com/jtgeibel/conduit-hyper/pull/1/files

PR #1383, opened 2018-05-03
Update the look of the install box

avrong:
Before:
cargo_before_cut

After:
cargo_after_cut

### DIFF
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -268,52 +268,64 @@
     }
 
     .install {
-        font-size: 120%;
         padding: 10px;
         background-color: darken($main-bg-dark, 8%);
-        @include display-flex;
 
-        .action {
-            display: none;
-            text-align: center;
-            padding-top: 18px;
-            padding-right: 10px;
-        }
-        code {
-            @include flex(1);
-            background: $main-color;
-            color: white;
-            padding: 20px;
-        }
-        button, button:active {
-            padding: 5px 0;
-            background-color: #FFFFFF;
-            border: none;
-            width: 60px;
-            cursor: pointer;
-        }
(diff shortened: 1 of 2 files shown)


sgrif:
Hi @avrong, thanks for the PR! Can you give some more context as to why you think this change is an improvement/needed?

avrong:
Hi @sgrif. Some my thoughts.

Old version have the non-centered big hint message which takes a lot of horisontal space (reduces the code line) and is useful for newbies only. Also, the "Copied!" message displays in another element and is moving content down when displaying.

In the new version we have the information line above the code line which takes a bit of vertical space. There also displayed "Copied!" message above the copy button.

sgrif:
> is useful for newbies only.

TBH I think this is the group we need to care most about here. Personally I find the new style very hard to read. I don't think horizontal space is something we're short of here. Is there a middle ground that keeps the `Cargo.toml` prominent, but doesn't move elements down when `copied` is displayed?

avrong:
Hi @sgrif. What so you think about this one?

install_box_new

carols10cents:
https://crates.io/crates/lock-free-multi-producer-single-consumer-ring-buffer is the longest crate name; it actually wraps in the install box and doesn't look *tooooo* terrible. I don't have real strong feelings about this, but in case this sways anyone:

Before:

screen shot 2018-08-22 at 7 02 15 pm

After:

screen shot 2018-08-22 at 7 02 08 pm

I'm not wild about the tab with the gradient; it doesn't feel like it fits into the current design (such as it is).

carols10cents:
I think we've decided not to take this change, we just never got around to closing this PR. Thank you though!

PR #1390, opened 2018-05-08
Fix Guide link in footer

wezm:
The [Guide link in the footer](https://doc.rust-lang.org/cargo/guide.html) is currently not found.


jtgeibel:
Thanks for the fix!

PR #1391, opened 2018-05-09
Bump dotenv

Eijebong:


jtgeibel:
Unfortunately we cannot update civet at this time. I plan to yank 0.10 and 0.11 due to compilation issues on Mac. We're also looking to migrate from civet to hyper.

Eijebong:
Removed the civet bump ;)

PR #1394, opened 2018-05-11
Update tests to use hyper 0.12

jtgeibel:
This PR updates the HTTP recording test infrastructure to use hyper 0.12.  The goal is to update hyper in the tests now to minimize any version changes when eventually moving production over to hyper (see companion PR #1378).  While we don't *have* to use the same version of hyper in production and our tests, having different versions would regress compile times.

Ignoring the crate additions/removals (as production code obviously don't depend on them), the only overlap I see with production is updating `libc` from 0.2.33 to 0.2.42.  As far as I can tell, the following changes do not affect production, even transitively (and if they do, are minimal):

* `net2` - 0.2.31 to 0.2.32
* `bytes` - 0.4.5 to 0.4.8
* `futures` - 0.1.17 to 0.1.21
* `tokio-core` - 0.1.12 to 0.1.17
* `tokio-io` - 0.1.3 to 0.1.6

It was necessary to lowercase the request header names in the recordings under `http-data` because the hyper client sending the request no longer uses title-case.


jtgeibel:
I've rebased and switched over to the newly released hyper 0.12.

sgrif:
bors: r+

jtgeibel:
Bors failed to merge this because another merge hit master while CI was running on the bors staging branch.  I've rebased on master and resolved the conflicts.

PR #1398, opened 2018-05-20
opensearch.xml not served if Accept header is text/html

### LINKED ISSUE #1358: opensearch.xml not served if Accept header is text/html
It seems that `opensearch.xml` is not served when the browser sends an `Accept` header containing `text/html`. Instead, it goes to the js frontend. This can be demonstrated by running `curl https://crates.io/opensearch.xml` vs `curl -H'Accept:text/html' https://crates.io/opensearch.xml`.

As this is the behavior of Firefox 59, it cannot add the search engine via its search bar. I suppose this is a failure of Firefox (Chromium doesn't send `Accept` headers when probing for search engine definitions), but it seems like crates.io should perhaps always serve the XML when that path is requested, regardless of headers.

ericmoritz:
According to #1358 Firefox incorrectly sends a Accept: text/html header when requesting
the opensearch.xml.

This fixes the bug, but I am not sure if it is the most ideal.

edit by JTG: Fixes #1358

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -75,6 +75,7 @@ pub fn build_middleware(app: Arc<App>, endpoints: R404) -> MiddlewareBuilder {
     if env != Env::Test {
         m.around(StaticOrContinue::new("dist"));
         m.around(EmberIndexRewrite::default());
+        m.around(StaticOrContinue::new("dist"));
         // Note: around middleware is run from bottom to top, so the rewrite occurs first
     }
 


jtgeibel:
Thanks for the contribution @ericmoritz!  I'm too am a little unsure about just checking for the ".xml" extension.  I don't think this would be valid in a crate name or version spec.  I don't think I've seen a github user or team name with a "." either, but I'm not certain if that is disallowed or not.  It looks like we don't allow "." in keywords either, so I think we are okay in all the *current* places I see where a URL can end in a user defined string.

I'm also guessing that Firefox requests "text/html" for `favicon.ico` too because the icon is missing in Firefox but present in Chrome.  It's possible some crawlers are not seeing our `robots.txt` as well.

I wonder if a more general solution would be to check against the `dist` folder before checking to see if a rewrite is necessary (swapping the order [here](https://github.com/rust-lang/crates.io/blob/60a96a133775dafeedfb16ec93a4dbfa6b37ff9a/src/middleware/mod.rs#L76-L77)).  Then the EmberIndexRewrite middleware would need to be tweaked to serve the static `index.html` directly (or we could check the `dist` folder twice, before and after the rewrite).

ericmoritz:
How about this design?

Remove the EmberIndexRewrite middleware,
Alter StaticOrContinue to pass through the request if the request path starts with `/api/`
If it doesn't start with `/api/` attempt the static_handler.
If the static handler fails with a 404, rewrite the request to `/index.html`

ericmoritz:
What if we just add a routes for `/crates/{name}` and pass those to the static handler as a request to `index.html` and serve those with the main handler?  That seems like a more explicit and straight forward way to serve up index.html.

Then if the main handler 404s, fallback to the static handler?

jtgeibel:
Unfortunately there are also a lot of [front-end routes] as well and if I understand your proposal correctly we would end up duplicating that list.

I'm really leaning towards just doing:

```rust
m.around(StaticOrContinue::new("dist"));
m.around(EmberIndexRewrite::default());
m.around(StaticOrContinue::new("dist"));
```

It looks weird but it aligns with what the old logic does and allows static files to be served first, a rewrite to `/index.html` if necessary, and then another pass at serving the static files in case the rewrite occurred.

Ideally we could just move the remaining 3 backend routes under `/api/` and drop the `wants_html` portion of the check, but we've unexpectedly broken some 3rd party tools in the past when moving some other routes.  While it is a bit hackish, duplicating the static middleware should be the lowest risk.

I'm going to be away from cell service for the long weekend, so as a heads up I likely won't be able to respond further until Tuesday.  I really appreciate your investigation and suggestions on this!

[front-end routes]: https://github.com/rust-lang/crates.io/blob/master/app/router.js

ericmoritz:
I'll try out your solution over the weekend.  Have a great weekend!

ericmoritz:
Adding the extra middleware does fix the bug and is the least disruptive change. 

I thought about merging the logic into a single middleware. My logic was: 

1. Call the static handler
2. If that 404s, call the fallback handler
3. If that 404s, call the static handler as `index.html`

That seemed to cover all the cases, and made the logic more explicit. However, I was too afraid of breaking something due to ignorance.  That logic removed to Accept check because it seemed uneccessary. I could easily add it back in on step 2.

How do you feel about using this logic?

ericmoritz:
Absolutely.  I'll squash it tomorrow night.

jtgeibel:
Thanks for squashing into 1 commit.  It looks like the CI failures are unrelated and I'll be fixing that soon so that bors can merge this.

jtgeibel:
bors: r+

PR #1402, opened 2018-05-23
Pull diesel_cli version from .diesel_version when installing on CI

jtgeibel:
The Diesel buildpack uses the version defined in `.diesel_version` when
running migrations.  This change ensures that we are also using the
same version on CI any time the version in that file is bumped.


sgrif:
bors: r+

PR #1408, opened 2018-05-31
Run `cargo update -p openssl` and fix deprecation

jtgeibel:
This is a semver compatible bump.  A bump to 0.10+ will be done in a future PR.


sgrif:
Just to clarify, this is not the version that we suspect was responsible for the previous segfaults, right?

jtgeibel:
Yes. The last PR jumped straight to 0.10.x.

I think the 0.10 and 0.9 series both pull in the same 0.9 series -sys
package. So if the bug was in the -sys then we may see it here too, but I
think the version of the -sys has advanced in the past few months so if it
was openssl-sys it may also have been fixed in the interim.

I believe this PR pulls in the most recent -sys.

sgrif:
:+1: I'd still like to wait to merge this until Monday so I can be around to watch stats when we deploy just in case

jtgeibel:
Sounds great to me.

sgrif:
bors: r+

Will deploy in the morning

PR #1409, opened 2018-05-31
`cargo update -p tar` and update one test

jtgeibel:
Some minor changes to header bytes in the tar crate have caused changes
to the test data.  It appears this is the only test that succesfully
uploads a tar file with file contents.


carols10cents:
oooops i forgot about bors :(

notriddle:
You can set GitHub to block pull requests that haven't been run through bors. [Basically](https://forum.bors.tech/t/a-plan-for-getting-rid-of-the-big-green-merge-button/141) configure it [like this](https://forum.bors.tech/uploads/default/original/1X/10afbfd663b1fd8a582a07375808b37fb58c59a8.PNG).

Issue #1001, opened 2017-08-22
Mark fields that are used by shields.io


dtolnay:
The shields.io service expects certain fields to exist in the JSON response of `'https://crates.io/api/v1/crates/' + crate` -- listed [here](https://github.com/badges/shields/blob/6534a2a1a821da91854d5d940f6e469c1c4851e2/server.js#L624-L663). In https://github.com/badges/shields/issues/1059 it looks like their license badge either never worked (?) or was broken at some point by a crates.io change.

Let's put a comment above fields that are used by shields.io so that if they need to change in the future, we know to follow up with a PR to shields.io.

kzys:
@carols10cents Can we close this issue now, since the PR has been merged?

PR #1421, opened 2018-06-06
#1001 Add a comment to fields used by shields.io

ericmoritz:
Fix for #1001. I guessed on the formatting of the comment. I can easily change it if needed.

### DIFF
--- a/src/views/mod.rs
+++ b/src/views/mod.rs
@@ -91,8 +91,10 @@ pub struct EncodableCrate {
     pub badges: Option<Vec<EncodableBadge>>,
     #[serde(with = "::util::rfc3339")]
     pub created_at: NaiveDateTime,
+    // NOTE: Used by shields.io, altering `downloads` requires a PR with shields.io
     pub downloads: i32,
     pub recent_downloads: Option<i64>,
+    // NOTE: Used by shields.io, altering `max_version` requires a PR with shields.io
     pub max_version: String,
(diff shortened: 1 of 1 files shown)


sgrif:
I'm not really sure that we need to specifically mark fields used by one client. We shouldn't be making any backwards incompatible changes to our API, as it is versioned and we know folks rely on it. (I'm not sure what specifically happened with #1001, I would guess that it never worked until it was fixed).

However, that issue has been open for a year so I'll defer to @carols10cents

Turbo87:
@sgrif I believe https://github.com/rust-lang/crates.io/pull/787 is the PR that introduced the problem

ericmoritz:
@sgrif I won't mind if we close this PR.  I was bored last night and grabbed a low hanging fruit.

carols10cents:
> I'm not really sure that we need to specifically mark fields used by one client. 

[We already mark the routes that cargo uses](https://github.com/rust-lang/crates.io/blob/master/src/router.rs#L16-L19); this doesn't feel much different.

> We shouldn't be making any backwards incompatible changes to our API, as it is versioned and we know folks rely on it.

Mistakes happen, we haven't documented the requests that our frontend uses, and we have been known to say "oh we're the only one using this, we can change it" and then been wrong. 

I like this change. Thank you @ericmoritz !!

bors: r+

PR #1426, opened 2018-06-07
WIP: When searching order crates by last update date

### LINKED ISSUE #1210: Add an option to order crates by last updated time

joaopapereira:
Closes #1210

### DIFF
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -4,11 +4,13 @@ use diesel_full_text_search::*;
 
 use controllers::helpers::Paginate;
 use controllers::prelude::*;
-use models::{Crate, CrateBadge, OwnerKind, Version};
+use models::{Crate, CrateBadge, OwnerKind, Version, User};
 use schema::*;
 use views::EncodableCrate;
 
 use models::krate::{canon_crate_name, ALL_COLUMNS};
+
+use std::collections::HashMap;
 
 /// Handles the `GET /crates` route.
 /// Returns a list of crates. Called in a variety of scenarios in the
@@ -32,7 +34,6 @@ use models::krate::{canon_crate_name, ALL_COLUMNS};
 /// function out to cover the different use cases, and create unit tests
 /// for them.
 pub fn search(req: &mut Request) -> CargoResult<Response> {
-    use diesel::sql_types::Bool;
 
     let conn = req.db_conn()?;
     let (offset, limit) = req.pagination(10, 100)?;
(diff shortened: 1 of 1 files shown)


src/controllers/krate/search.rs
```diff
@@ -188,19 +227,538 @@ pub fn search(req: &mut Request) -> CargoResult<Response> {
...
         .collect();
+    crates
+}
 
-    #[derive(Serialize)]
-    struct R {
-        crates: Vec<EncodableCrate>,
-        meta: Meta,
+#[cfg(test)]
+mod test {
```
sgrif: Should this live with the rest of the integration tests, rather than re-implementing half of it here?

src/controllers/krate/search.rs
```diff
@@ -32,15 +34,53 @@ use models::krate::{canon_crate_name, ALL_COLUMNS};
...
+
+    let crates = execute_search(
+        &conn,
+        offset,
+        limit,
+        params,
+        sort.to_string(),
+        current_user_id,
+    );
+    let total = crates.len() as i64;
```
sgrif: This is not correct. The query is running the equivalent of `SELECT crates.*, recent_downloads.downloads, COUNT(*) OVER ()`. `total` is the number of crates that could be returned without pagination, not the number returned by the query.

src/controllers/krate/search.rs
```diff
@@ -32,15 +34,53 @@ use models::krate::{canon_crate_name, ALL_COLUMNS};
...
+        total: i64,
+    }
+
+    Ok(req.json(&R {
+        crates,
+        meta: Meta { total },
+    }))
+}
+
+fn execute_search(
```
sgrif: I'm not sure why this needed to be split into a separate function.

src/controllers/krate/search.rs
joaopapereira: Are you talking about the builder or the tests?

sgrif: Both

carols10cents:

Thank you!

Issue #1393, opened 2018-05-11
add info on crate size


jaroslaw-weber:
It would be nice to have information about the crate size without downloading it (in the crate's page).

jtgeibel:
Yes, this would be a nice addition.  I recently hit the 10MB limit on a `-sys` crate that was pulling in a lot of code and resources that weren't actually needed for the build.  I probably would have noticed this and trimmed things down sooner if the page on crates.io had shown the size.

I don't think this is tracked anywhere in the database yet, but this is definitely something we could start tracking for newly published versions.  Would you be interested in putting together a PR for review?

jaroslaw-weber:
I will try to send a PR later :)

jtgeibel:
You're right, this has been resolved so I'm closing.  Thanks!

PR #1436, opened 2018-06-17
Add crate size on the crate detail page

PinkFloyded:
From #1393

Currently it looks like this:

crate_size_ui

I'm new to Rust and Ember. So, let me know if something can be done better.


carols10cents:
So I merged in upstream to resolve the merge conflicts, tweaked a little bit of the handlebars formatting, added some backend tests, and was about ready to merge when I realized-- `crate_size` should really be an attribute of a crate's `version`, not the crate itself, so that when you go to a different version's page, you see that version's size.

Are you up for moving the `crate_size` field to `version` instead? Or would you like me to take care of it? Either is fine with me!

PinkFloyded:
> crate_size should really be an attribute of a crate's version, not the crate itself, so that when you go to a different version's page, you see that version's size.

Makes sense. Hadn't thought of that.

I have moved `crate_size` to `versions`. Please take another look.

carols10cents:
Thank you for making that change!

I was again just about to merge, when I tried it out locally by publishing versions of a crate whose files had different amounts of data, and I noticed the size was always the same.

Turns out this code is really confusing, and there are a bunch of different "lengths" hanging around. The one that was being used was the length of the JSON metadata that describes the crate being uploaded, rather than the size of the tarball.

I wrote a failing test to capture this, and writing the test was also way harder than I expected to be; I think I'm going to have some more refactoring to do later inspired by this.

I also changed the variable names to hopefully be more descriptive, and I realized I had some knowledge about the format of the body of the publish request that wasn't written down anywhere, so I added some comments documenting this as well.

So thank you for exposing some problems in the codebase!! :) I'm going to make sure travis passes, approve the changes in percy, then I'll really be merging this in :)

carols10cents:
bors: r+

PR #1437, opened 2018-06-20
Prepare for new rustfmt in 1.27 release

jtgeibel:
This is the diff that results from `cargo +beta fmt`.  It will not pass
CI today, but should be mergable via bors after the release tomorrow.


jtgeibel:
bors: r+

PR #1445, opened 2018-06-27
Use a user's ID not their GH login to determine if they're an owner

### LINKED ISSUE #1443: Manage owners link shows if you have the same username as a deleted github account
I removed my GitHub account a while ago because I didn't need it anymore. I didn't think about the fact that I had some crates published.
Now I created it again and tried to access said crates (same email, same username).

While it does (kind of?) recognize me as owner, it does not allow me to modify the crate or publish a new version.

When clicking the avatar (see image below) I get forwarded to my new crates.io site (which shows 0 crates).

image

sgrif:
This code was likely written back when we used GH logins as a user's
unique identifier and never got changed when the backend did.

Fixes #1443


carols10cents:
phew.

bors: r+

PR #1457, opened 2018-07-10
Cargo.toml: Bump license-exprs to 1.4.0

### LINKED ISSUE #1278: Can't publish AGPL-3.0-or-later crate
Fails with `error: api errors: unknown license or other term: AGPL-3.0-or-later; see http://opensource.org/licenses for options, and http://spdx.org/licenses/ for their identifiers`, yet it is clearly listed on http://spdx.org/licenses/

wking:
And then update `Cargo.lock` with:

```console
$ cargo build
```

This bumps us from the SPDX License List 2.4 to 3.1.  Some 2.4 identifiers have since been deprecated by SPDX, but none of them have been removed, so it should be safe to deploy this before bumping Cargo and its docs (which mention 2.4 since rust-lang/cargo#4898).

Fixes #1278.


carols10cents:
Just tested this out locally, works great! I was able to publish a crate with license AGPL-3.0-or-later.

bors: r+

PR #1459, opened 2018-07-10
Don't consider yanked versions for `max_version` on Search

### LINKED ISSUE #1452: Clicking on crates.io search result leads to different page than the direct link.
If you go straight to the serde_json page from a this link: https://crates.io/crates/serde_json it brings up 1.0.22. If you search for serde_json from crates.io and click on the serde_json crate link, it brings up 1.0.23, which has been yanked.

This was really confusing for me, because I didn't find serde_json on crates.io through a direct link, so thought serde_json had been generally yanked, not just 1.0.23.

But the bigger issue is that the page you arrive to through a creates.io search result crate link is different than what the url actually shows if you go directly there.

sgrif:
Since this seems likely to creep up again, I've added a canonical
`versions` method to both `Crate` and `Vec<Crate>` which filters out
yanked versions, and a separate `all_versions` which includes yanked
versions. I've gone through all uses of `Version::belonging_to` to make
sure that these methods are used instead.

I'm tempted to remove the association entirely to force this function to
be used instead, but if I did that we'd lose the ability to use
`grouped_by`, which is a lot of code to replace.

In one case, this required adding an additional temporary, since the
trait appears to be confusing the borrow checker (I'm not sure why the
temporary is required now, but wasn't before -- the same lifetime is
appearing on a type in the same positions. I suspect it's a borrow
checker bug that would fall under the blanket of "fixed by MIR borrowck"
bugs)

Fixes #1452.

### DIFF
--- a/src/controllers/krate/downloads.rs
+++ b/src/controllers/krate/downloads.rs
@@ -7,7 +7,7 @@ use std::cmp;
 
 use controllers::prelude::*;
 
-use models::{Crate, Version, VersionDownload};
+use models::{Crate, CrateVersions, Version, VersionDownload};
 use schema::version_downloads;
 use views::EncodableVersionDownload;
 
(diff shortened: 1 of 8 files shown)


jtgeibel:
Looks good to me!

bors: r+

PR #1463, opened 2018-07-18
Add list of category slugs to docs menu

kyrias:
Previously the category slug list wasn't linked from anywhere on
crates.io.


carols10cents:
bors: r+

PR #1466, opened 2018-07-25
[DO NOT MERGE] WIP background job framework

### LINKED ISSUE #1384: Move all of our git logic off the web server
Note: I'm actively working on this, just documenting here to increase visibility and get outside feedback.

We have three endpoints that need to interact with git. `crates/new`, `versions/yank`, and `versions/unyank`. All three of these are slower than they should be (crate upload is slow for additional reasons which is outside the scope of this issue). Currently our behavior is:

- Do our git operation
- Attempt to push
- If pushing fails, `git fetch` and `git reset --hard`, then retry
- Retry at most 20 times

This has a lot of problems. The most obviously pressing for me is that we are doing an operation that takes seconds, and blocking sending a response on it. There's [a comment](https://github.com/rust-lang/crates.io/blob/674a57102715938901fdeb420989f4e773d932d0/src/git.rs#L129-L132) that talks about just spawning another thread for this, but that is very wrong and will cause major problems if we do that (what if it fails? what if the dyno is restarted before that thread completes?).

The second problem is that our retry behavior is limited to a very narrow scope. If any failure occurs other than failing to push, we abort. This is even more fragile, because we assume that our local checkout is up to date with the index. While this is almost always true (since the most common operation here is publishing a new crate, which shouldn't fail from being out of sync), it's not guaranteed, which leads to issues like #1296. 

What we need here is a proper background queue. For publishing a new crate, yanking, and unyanking, we should immediately update the database, and push updating the index into a queue to be completed by another machine in the future.

I feel very strongly that we should keep PostgreSQL as our only external dependency for as long as we possibly can. I don't want to introduce RabbitMQ, RocksDB (e.g. Faktory), etc as a dependency at this point.

For this reason, I've prototyped out a very simplistic background queue that uses PG's row locks to handle tracking whether a job is ready, taken, or done for us. You can see the current state of the code [here](https://github.com/sgrif/crates.io/commit/de9d9bf2aa177b64133ad7b1027616cfe5b458f5) (note: it will not compile right now, it relies on two features of Diesel 1.3).

The implementation I'm looking at assumes the following:

- All background jobs will be idempotent. To put it another way, they will work under "at least once" retry semantics.
- Every job should eventually complete, given sufficient retries
  - This implies that a job continuing to fail for long enough should page us. At best the failure is because GitHub is down (which means we should update our status page), and at worst it means a mismatch between our database and the index which needs manual resolution.
- Failures which don't result in the transaction returning `Err` (meaning in this implementation, the retries counter is not incremented) are things we don't care about, and it's fine to immediately try again.
  - This would be caused by panics from OOM, the dyno restarting, or a hardware failure
- We will only have one background worker for git related tasks
  - This isn't really ever actually assumed in the code, but it makes no sense for us to have more than one worker for this. If we have two they'll just constantly conflict with each other, and actually slow things down.

sgrif:
This PR is to make it easier to review the "guts" code as it gets worked out before we actually add the background jobs to fix #1384.

Since this was last reviewed I have fixed the race condition on failed jobs, and added some bare bones tests around the queueing infrastructure. I *think* this code is ready to go, but wanted to get some more eyes on it before I submit the final PR to move git logic over here.


sgrif:
@carols10cents This was an absolute nightmare to get proper tests for, but I think this should address your concerns?

src/job.rs
```diff
@@ -0,0 +1,247 @@
...
+            .skip_locked()
+            .first::<BackgroundJob>(conn)?;
+        increment_retries(conn, found_job.id);
+        job = Some(found_job);
+        Ok(())
+    })?;
+
+    if let Some(job) = job {
+        conn.transaction(|| {
+            delete(&job).execute(conn)?;
```
jtgeibel: If I understand the locking strategy here, the first transaction above ensures that no 2 threads will fetch the same job.  Once a job is fetched and the retry info is updated, the query will not fetch that row again for at least 1 minute.

Very shortly after the first transaction is completed (and that row lock released), the delete statement in this transaction will again lock the row for update such that even a job that runs for longer than a minute will not be fetched again by another thread until the row is unlocked.

Is my understanding here correct?  Is there an advantage/requirement to do these steps in 2 separate transactions?  I ask because it took me a little while to realize that this delete is re-locking the row and that it is important to issue this delete within the transaction *before* executing the job in the next line to ensure the row is locked while the job runs.

If somehow it took more than a minute for a thread to progress from line 72 to 76, and another thread fetched the same job, then I believe one of them would block on this delete (or maybe abort due to a detected deadlock).  If the other transaction completed successfully (deleting the row), then this would return an error.  If the other transaction failed, then the row would still exist and be unlocked so the other thread would immediately retry the job.  I don't think this is a reasonable failure mode, but I think a single execution of the job would still be guaranteed under this extreme scenario.

sgrif: > Is my understanding here correct?

Yes

> ? Is there an advantage/requirement to do these steps in 2 separate transactions? 

It's a requirement. We cannot update the retry counter in the same transaction as the job itself, since if the job fails the transaction may be aborted meaning we cannot commit those changes.

> If somehow it took more than a minute for a thread to progress from line 72 to 76

We will hit a TCP timeout before a minute passes. The only way it could take more than a minute to get from line 72 to 76 is if the OS has actually pre-empted our process and we receive no CPU resources for a full minute, which is not realistic.

But yes, you're correct that this behavior relies on less than a minute passing between the first transaction committing and the second transaction locking the row.

> and another thread fetched the same job, then I believe one of them would block on this delete (or maybe abort due to a detected deadlock).

Yeah, it'd block on the delete. I think you're right that we should add a sanity check to ensure the row was *actually* deleted before running the job as a sanity check. At the point where the `delete` line returns, we can be sure that we have an exclusive lock on the row, so that would be the point to make sure "no seriously we are the only thing that has access to this 100%". It's not a realistic failure scenario, but it doesn't hurt to handle it.

jtgeibel: > It's a requirement. We cannot update the retry counter in the same transaction as the job itself, since if the job fails the transaction may be aborted meaning we cannot commit those changes.

Ah, great point!  I knew I was missing some subtlety here.

Overall this looks really good to me.  While we'll only need one consumer for git operations, this is well future proofed for additional use cases.

sgrif: Yes, I've written this code to be extracted into its own library at some point. The only thing that should be coupled to crates.io is the use of `CargoResult`. To make this generically applicable to us I will specifically need to add a way to have a differing number of workers for different job types (having >1 worker for the git job makes no sense), but we don't have to deal with that for now since the git job will be our only job

sgrif:
I've updated this with what I think will be the final queuing code (I will be working on moving one of the git functions over to this tomorrow). The main change I've made here is not passing the runner's connection to the job, which means that we no longer have to juggle two transactions. I've also set us up to catch panics

sgrif:
Going to open a new PR with this and the git migration

PR #1467, opened 2018-07-26
Begin preparation for Rust 2018

jtgeibel:
This PR captures some of the results of my experimentation with `cargo fix`.  The first 2 commits are required in order to get `cargo fix` to apply cleanly.  In addition to these changes, there are 3 macros which cause problems in the automated conversion.  I have not incorporated the fix for these macros because the `crate` keyword cannot be used in paths on current stable.  I plan to file bugs against `cargo fix` for these 3 scenarios.

The final commit explicitly `use`s things in preference to using absolute paths within functions.  This commit isn't strictly required for a clean `cargo fix` run, but it will reduce unnecessary usage of the `crate` keyword one we run `cargo fix` for real.


carols10cents:
This is going to keep getting out of date and merge conflicty if we don't do it soon, and most of the other PRs already have merge conflicts sooooo

bors: r+

PR #1469, opened 2018-08-01
Allow a new macro hygiene lint until we get Diesel 1.4

jtgeibel:
Because we deny warnings, this new lint will cause our CI to fail once
it hits beta.  Allowing `unknown_lints` is necessary because
stable-1.28 isn't aware of this lint.

This patch can be reverted as part of upgrading to Diesel 1.4 once that
is released.  See the [upstream PR] for more information.

[upstream PR]: https://github.com/diesel-rs/diesel/pull/1787


jtgeibel:
I'm going to go ahead and merge this so that master builds cleanly on nightly.  The plan is to revert this commit as part of our next Diesel upgrade.

bors: r+

PR #1471, opened 2018-08-03
Add wildcard lifetimes where they were previously elided

jtgeibel:
This diff was created with the following commands:

    RUSTFLAGS="-W elided_lifetimes_in_paths" cargo +nightly fix
    cargo +stable fmt


sgrif:
bors: r+

PR #1476, opened 2018-08-15
Revert "Merge #1378"

jtgeibel:
This reverts commit ba6c689c2ebec7bcb8a9298de0750638af36c9ec, reversing changes made to 9b4a92cf16f712dbc4f3074fb40641b0f2b64c88.

This reverts the changes introduced in PR #1378.  The deploy has already been rolled back and this will get master back into a clean state so that deploys can be made if other issues arise before this bug can be resolved.  The code reverted by this PR was causing `cargo publish` to fail with a 404 error.

The issue appears to be triggered because `cargo publish` includes an additional `/` at the beginning of its route.  It is possible that civet removes this automatically before it gets to conduit's routing logic, and that hyper leaves the extra `/` intact, causing the routing logic to fail to find the correct handler.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,7 +29,7 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "html5ever 0.21.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


carols10cents:
thank yoouuu!!! 

bors: r+

PR #1477, opened 2018-08-16
[do not merge] Attempt to write a test that sends a request with two slashes

carols10cents:
@jtgeibel I tried to write a test that publishes to `//api/v1/crates/new` because that's what cargo does, but it fails even without your switch to hyper??? I don't know if our test infrastructure is different than the real stack or if I messed something up or what, but I wanted to give this to you in case it was useful at all.

### DIFF
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -2381,3 +2381,15 @@ fn new_krate_hard_links() {
     let body = ::new_crate_to_body_with_tarball(&::new_crate("foo", "1.1.0"), &tarball);
     bad_resp!(middle.call(req.with_body(&body)));
 }
+
+#[test]
+fn double_slash_in_publish_urls() {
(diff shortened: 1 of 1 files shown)


jtgeibel:
Interesting. I think even our tests under src/tests/ test our crate as if it was a library. I seem to remember wanting to make some things private in the lib crate but couldn't because those tests need to access them. I don't think we have anything that actually boots the application and sends real HTTP requests. Everything is a mock request injected into the router, skipping over the actual HTTP server  Maybe that sort of test would need to live in src/bin/server.rs, but we would probably need to build some infrastructure (or find an existing crate) to help with this.

sgrif:
Yup, our tests never spin up the actual web server -- they enter at conduit's router, at which point we expect the paths to have already been normalized. I'm not sure we can actually test this unless we go the full selenium/capybara route of spinning up a web server in another thread/process and sending requests over the network.

carols10cents:
Yeah. I'm going to close this for now.

Issue #1480, opened 2018-08-26
Deny rust keywords as crate names


pyfisch:
User @swmon recently registered a lot of crate names that happen to be rust keywords like  [if](https://crates.io/crates/if), [while](https://crates.io/crates/while) and [type](https://crates.io/crates/type).

These crates should not have been published in the first place as they can't be used because of the reserved word and the [cargo command line tool](https://github.com/rust-lang/cargo/blob/db48c63b3641a29a20b572ed516c58c84450e3bd/src/cargo/ops/cargo_new.rs#L131) already blacklists them.

I propose to blacklist the names in the back-end and remove the crates already published.

To implement this I think [`fn valid_name`](https://github.com/rust-lang/crates.io/blob/9b4a92cf16f712dbc4f3074fb40641b0f2b64c88/src/models/krate.rs#L251) needs to be changed.

(This is maybe a good "first PR" issue.)

terry90:
I'll do it then 👍

jtgeibel:
Closing based on the reasons outlined in https://github.com/rust-lang/crates.io/pull/1482#issuecomment-418565821.

89z:
https://github.com/rust-lang/crates.io/issues/1747

PR #1482, opened 2018-09-04
Blacklist reserved keywords as crates name

terry90:
Adresses #1480 

I'm not sure about the migration though, is there associated records to be manually deleted ? Do the foreign keys have a cascade delete thing ?


terry90:
Hi,
Thanks for the feedback.
I was unaware of this RFC, but I don't see why the blacklist should be any different than the Cargo's one ? I agree about including only strict keywords though.
Anyway if there is no technical reason as you say, maybe this PR shouldn't have been opened first and  #1480 can be closed too.

PR #1483, opened 2018-09-04
Test diesel migration to make sure the schema patch applies

carols10cents:
will explain this better in a bit but my kid just started crying

i'm hoping the travis build fails because i'm currently not able to deploy master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ matrix:
       script:
         - rustup component add rustfmt-preview
         - cargo fmt -- --check
+        - diesel migration run # check that the schema.patch file applies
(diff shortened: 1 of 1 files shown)


sgrif:
Diesel 1.4 is going to have a "verify-schema" flag which instead of regenerating the schema file will instead error if the file changed, which we should for sure enable on CI once that's added. I will try to prioritize implementing that.

carols10cents:
Sounds like this change is moot then, with the upcoming changes to diesel.

sgrif:
et viola https://github.com/diesel-rs/diesel/pull/1855

PR #1487, opened 2018-09-07
Handle renamed dependencies coming from Cargo

alexcrichton:
This commit is intended to be coupled with
https://github.com/rust-lang/cargo/pull/5993 where Cargo will start sending the
registry more information about locally renamed crates to persist into the
index.


alexcrichton:
Note that this likely shouldn't be merged until https://github.com/rust-lang/cargo/pull/5993 is decided on

Eh2406:
Sorry for being out of the loop, why does the index need to know what a crate chose to rename a dependency to?

alexcrichton:
Ah no worries, I thought the same as you! With features, however, it becomes significant because the implicit feature name for the dependency is named after the left-hand-side of what you write in `Cargo.toml`, which isn't the package name (which is what we're publishing right now)

Eh2406:
So if I have an optional dependency on `serde = "1"` that I have renamed to `serde1` then the feature will be called `serde1`? Has the feature part of the resolver ben tealt to read that correctly?

alexcrichton:
Correct yeah! To do that you'd do `serde1 = { version = "1", package = "serde" }`, and there should be tests yeah in Cargo for having this plumbed everywhere inside Cargo at least

alexcrichton:
Ok it looks like we'd like to merge https://github.com/rust-lang/cargo/pull/5993, in light of that r? on this PR to get it merged?

alexcrichton:
Now with a test!

carols10cents:
And I've deployed this!

PR #1488, opened 2018-09-07
Only use session tokens for auth for `GET` requests

sgrif:
Since we do not have any form of CSRF protection in place, we should not
be allowing session tokens to be used for non-get requests. We don't
currently have any CSRF vulnerabilities, as there are no `POST` requests
in our router today.

In the event that one does get added in the future, this will prevent
a CSRF vulnerability from appearing, without us having to remember this
detail in the future. It will also force us to properly add some form of
protection if we want to accept a POST request sent by an HTML form in
the future.

### DIFF
--- a/src/middleware/current_user.rs
+++ b/src/middleware/current_user.rs
@@ -1,5 +1,6 @@
 use super::prelude::*;
 
+use conduit::Method;
 use conduit_cookie::RequestSession;
 use diesel::prelude::*;
 
@@ -29,7 +30,7 @@ impl Middleware for CurrentUser {
 
         let conn = req.db_conn().map_err(std_error)?;
 
-        if let Some(id) = id {
+        if let (Some(id), Method::Get) = (id, req.method()) {
             // If it did, look for a user in the database with the given `user_id`
             let maybe_user = users::table.find(id).first::<User>(&*conn);
             if let Ok(user) = maybe_user {


jtgeibel:
We've definitely had issues like this in the past, for instance https://github.com/rust-lang/crates.io/commit/2a4684555bb62a99c73c52455d1f153e5baf9b16.

I believe we have some front-end PUT routes that use the session cookie.  For instance, `/crates/:crate_id/follow`, `/me/tokens`, `/confirm/:email_token`, and a few others.  (There some corresponding DELETE routes as well.)  Maybe we should explicitly test for the POST method and skip setting the user only in that case.

sgrif:
The browser is only capable of sending `GET` and `POST` requests from a form. Only `POST` requests are vulnerable to CSRF assuming any browser more recent than IE9

jtgeibel:
Yes, but the current PR only sets the session for GET, and I think the PUT routes will fail without a session.  The cargo routes (such as publish) will still get a user session from the "Authorization" header, but PUT routes used by the front-end, such as creating a new token, will not know which user to create a token for.

sgrif:
Are there any non-get requests sent through the browser?

sgrif:
Ah I guess token management is handled through the browser. I'll change this to exclude POST rather than include only GET

carols10cents:
> I'll change this to exclude POST rather than include only GET

Hi @sgrif, could you make this change please?

carols10cents:

PR #1489, opened 2018-09-07
Regenerate `schema.patch`

sgrif:
The PR adding an ltree column was originally opened before the updates
to Diesel CLI, and merged after. The changes to schema.rs affected the
context lines used by the patch file to anchor the
`recent_crate_downloads` view, causing the schema patch to fail to be
applied.

This regenerates that file, so we should avoid this problem in the
future.

There are a few longer term things that need to happen.

- Diesel CLI needs a flag to error if the schema changes, rather than
  regenerating the file. This would also error if the patch file failed
  to apply. This feature will be in 1.4, and I will try to prioritize
  implementing it. We will want to use this flag in CI and deploys
- Diesel CLI needs a better workflow for resolving merge conflicts. I'm
  not entirely sure what this should look like, but it should probably
  at least start with a `--regenerate-patch-file` flag or something
  along those lines to automate this when conflicts do occur.
- Diesel should have a better story for importing views, since that's
  the piece of this that is most likely to conflict.


sgrif:
@carols10cents This fixes the issue you were looking at with #1483 and will enable you to deploy again

PR #1492, opened 2018-09-19
Tweak a few things on CI

jtgeibel:
* Only install node on the stable branch
* Move rustfmt-preview installation to the before_install step
* Clear target directory on nightly (downloaded crates and the index will still be cached)


jtgeibel:
Everything is looking good on CI.

bors: r+

PR #1493, opened 2018-09-20
Ensure clippy is run for all targets and switch to clippy-preview on stable

jtgeibel:
`--all-targets` is now passed to clippy on CI.  Temporarily,
`-- -A renamed_and_removed_lints` is passed as well.  This can be
removed once the clippy::* attribute syntax is available in stable.

Additionally, clippy-preview is now installed and run on stable.
We were previously running clippy on an old nightly from late June and
no new warnings (except the one noted above) are raised on nightly so
switching to the -preview component on stable shouldn't introduce
churn.  Also, unlike with rustfmt, we can fix and merge any clippy
recommendations before they hit the stable channel, so we don't have to
sync with new releases.


sgrif:
If clippy shipped on stable is recommending something that is not possible on stable, we should open an issue on clippy for that.

jtgeibel:
It looks like this has been reported upstream in rust-lang/rust#54406.  It doesn't affect stable yet, but does affect beta and nightly.  I believe the intent is to stabilize this in time for the 2018 edition (see https://github.com/rust-lang/rust/issues/44690#issuecomment-418883928) so hopefully the allow attribute isn't needed for long.

jtgeibel:
I've reworked the implementation and updated the PR description. `--all-targets` is now passed to clippy and a few minor issues were addressed in the tests.  `-- -A renamed_and_removed_lints` is now applied only to the clippy command, since these warnings don't affect other build profiles.

carols10cents:
Woohoo! I'm excited to not have to keep a particular nightly around anymore :P 

bors: r+

PR #1494, opened 2018-09-20
Test code refactoring experiment

carols10cents:
Opening this to see what people think of this; eventually want to get rid of all the new_req* functions.

This is what I was talking about @jtgeibel

### DIFF
--- a/src/tests/all.rs
+++ b/src/tests/all.rs
@@ -794,6 +794,7 @@ fn new_crate_to_body_with_tarball(new_crate: &u::NewCrate, tarball: &[u8]) -> Ve
 /// A struct representing a browser session to be used for the duration of a test.
 /// Has useful methods for making common HTTP requests.
 struct BrowserSession {
+    app: Arc<App>,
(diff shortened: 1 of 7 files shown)


src/tests/all.rs
```diff
@@ -788,3 +775,245 @@ fn new_crate_to_body_with_tarball(new_crate: &u::NewCrate, tarball: &[u8]) -> Ve
...
+            t!(ar.finish());
+        }
+        empty_tarball
+    };
+}
+
+/// A builder for constructing a crate for the purposes of testing publishing. If you only need
+/// a crate to exist and don't need to test behavior caused by the publish request, inserting
+/// a crate into the database directly by using CrateBuilder will be faster.
+pub struct PublishBuilder {
```
jtgeibel: I really like the `PublishBuilder` and `CrateBuilder` distinction here.  We currently do some strange things in our existing crate creation logic.  Porting things over to this will definitely be an improvement!

jtgeibel:
LGTM, thanks for the updates!

bors: r+

PR #1496, opened 2018-09-21
Switch to use AWS signature v4

### LINKED ISSUE #1486: Crates.io uses deprecated AWS signature v2 for S3 upload
Currently crates.io is using signature v2 in the authorisation header for uploading crates to S3, however this is deprecated:

https://docs.aws.amazon.com/general/latest/gr/signature-version-2.html

This causes an issue while trying to create a private crates.io which is using an S3 bucket on one of the new regions or a private S3 which only support v4 (eg minio).

Is there a reason that crates.io can't be switched over to use signature v4?

cswindle:
The main changes in the PR are to add support for AWS signature v4 (rather than the deprecated v2) to fix #1486. It also includes the following minor changes that I made to make it easier to validate the fix:

 - allow git authentication using SSH keys
 - fix docker-compose commands


carols10cents:
It looks like the tests need to be updated too. Do you need any help with that?

cswindle:
Sorry this was stuck behind a few other changes that I wanted to get into cargo, hopefully shouldn't take too much to get the tests passing, so I will sort them out.

src/uploaders.rs
```diff
@@ -51,16 +71,18 @@ impl Uploader {
...
             } => {
                 let host = match *cdn {
                     Some(ref s) => s.clone(),
-                    None => bucket.host(),
+                    None => host.clone(),
                 };
+
                 let path = Uploader::crate_path(crate_name, version);
-                Some(format!("https://{}/{}", host, path))
+                Some(format!("https://{}/{}/{}", host, bucket, path))
```
jtgeibel: It looks like this would change the location of where crates are stored in production.  If this is necessary, it will need to be configurable.

src/uploaders.rs
```diff
@@ -74,16 +96,17 @@ impl Uploader {
...
                 ..
             } => {
                 let host = match *cdn {
                     Some(ref s) => s.clone(),
-                    None => bucket.host(),
+                    None => host.clone(),
                 };
                 let path = Uploader::readme_path(crate_name, version);
-                Some(format!("https://{}/{}", host, path))
+                Some(format!("https://{}/{}/{}", host, bucket, path))
```
jtgeibel: Same here for the readme path

src/uploaders.rs
```diff
@@ -148,7 +183,7 @@ impl Uploader {
             let mut body = Vec::new();
             LimitErrorReader::new(req.body(), maximums.max_upload_size).read_to_end(&mut body)?;
             verify_tarball(krate, vers, &body, maximums.max_unpack_size)?;
-            self.upload(&app.http_client()?, &path, body, "application/x-tar")?
+            self.upload(&path, body, "application/x-tar")?
```
jtgeibel: The removed call to `app.http_client()` is part of the test infrastructure to record and play back HTTP requests.  It looks like `upload` needs to add the proxy information as part of building the `PutObjectRequest`.

src/uploaders.rs
```diff
@@ -178,10 +208,23 @@ impl Uploader {
...
-                bucket.delete(&app.http_client()?, path)?;
+            Uploader::S3 {
+                ref client,
+                ref bucket,
+                ..
+            } => {
+                let req = DeleteObjectRequest {
+                    bucket: bucket.to_string(),
+                    key: path.to_string(),
+                    ..Default::default()
```
jtgeibel: Same thing here, this needs to go through the proxy when testing.

cswindle:
Thanks for the pointers @jtgeibel, I have now managed to get one of the tests working by using record, so should be able to start making progress on this again.

sgrif:
@cswindle Are you still interested in working on this?

sgrif:
This is blocked on #1507

cswindle:
@sgrif, we are currently running with a fork of crates.io to run internally as a private registry at work, so I am keen to gradually get all the changes merged back into crates.io codebase (which this is the first of a few changes), however I have been very busy with work, so not had any opportunity to get this work done. Likely be at least a month before I even think about picking this up again.

sgrif:
@cswindle Thanks for the update. Are you ok with closing this PR until you're more able to work on it?

carols10cents:
I'm going to close this PR for now; please feel free to reopen when you have time to work on this again @cswindle!

PR #1500, opened 2018-09-24
Rollup of #1498 and #1499

sgrif:
#1498 uses curl. #1499 removes curl. This PR fixes that. Don't review this PR, review #1498 and #1499 separately. And then review the last commit of this PR if you want to make sure I did the replacement properly.

But please merge this PR instead of those two.

Close #1498.
Close #1499.


carols10cents:
I believe this is moot now.

PR #1506, opened 2018-10-04
Policies: Document policy on maintainer-requested removals

fweimer:
Under the current removal policy, maintainers can only request removal
through Mozilla Legal or by introducing a deliberate Code of  Conduct
violation.  This is not a desirable situation, so this change adds
some language which suggest contacting help@crates.io.

### DIFF
--- a/app/templates/policies.hbs
+++ b/app/templates/policies.hbs
@@ -39,6 +39,20 @@ do what the law requires us to do, and address flagrant violations of the Rust
 Code of Conduct.
 </p>
 
+<h3 id='the-maintainer'><a href='#the-maintainer'>Upon Maintainer Request</a></h3>
+
+<p>
+If you uploaded a package by accident and it needs to be removed, contact
+<a href='mailto:help@crates.io'>help@crates.io</a> and request removal.
+</p>
+
+<p>
+Please note that this option is not intended for package versions which
+merely should not be used for some technical reason, and can remain published.
+For that, use
+<a href='https://doc.rust-lang.org/cargo/reference/publishing.html#cargo-yank'><code>cargo yank</code></a> instead.
+</p>
+
 <h3 id='squatting'><a href='#squatting'>Squatting</a></h3>
 
 <p>


sgrif:
Our policy is that we don't remove crates for any reason unless we are legally required to. If you were to contact us requesting a crate removal, you would get this reply:

> Thanks for reaching out! We currently do not delete crates, however. You can leave a readme/description there however to contact you if anyone would like the crate name and transfer it if a request comes.

fweimer:
> Our policy is that we don't remove crates for any reason unless we are legally required to. If you were to contact us requesting a crate removal, you would get this reply:

@sgrif, what about the [Code of Conduct](https://www.crates.io/policies#code-of-conduct)? That's a different category from legally-mandated removals. I find it counter-intuitive that you choose to remove offensive material, but refuse to act if the maintainer politely requests removal.

DDoSolitary:
@sgrif I would like to remind you that the crates contain the name and email of the uploader, which are considered as "Personal Data" according to GDPR. So you're almost always legally required to (at least partially) delete the crates when the maintainer asks you to do so.

Serentty:
Does the EULA include a provision giving Crates.io the right to perpetually redistribute all crates? Because if not, then that relies on that being allowed by the crate's licence, which doesn't seem like a given at all.

JanBeh:
Given the current defaults of tools like `cargo`, which easily can lead to accidental publications (see also my [post here](https://users.rust-lang.org/t/crates-io-my-personal-impression/66660)), I find it irritating that (obviously) accidentally published data can't get removed (which is what has been stated in this thread). I understand there is a need to keep long-lived and actively used crates (published under Open Source licenses that allow redistribution) available, but I guess it makes sense to have a team that can manage individual cases in a less rigid way and to communicate that. Maybe such a team exists, or maybe not. The opposite has been communicated here. I find it sad that this issue got closed as I don't think having to contact a legal department is a desirable interface for developers when they have a legitimate concern. Of course data protection laws may or may not require deletion anyway, but I would prefer if things don't need to be escalated like that unnecessarily.

PR #1511, opened 2018-10-05
Build on the new test infrastructure and convert more tests

jtgeibel:
This builds on the test infrastructure improvements made by @carols10cents.  In particular, it adds a `Response<T>` type that wraps the `conduit::Response` and a callback.  Tests can use the `good()` method which asserts that the response status is good and returns the deserialized `T`.  It is the equivalent of `::json(&mut ok_resp!(...))`. The `bad()` method is equivalent to `bad_resp!`.

The callback is called if the test calls `good()` on the response.  This allows helper methods to return a `Response<T>` with attached assertions that are run if the test expects the response to deserialize correctly, but are not run if the test expects an error response.

### DIFF
--- a/src/controllers/user/me.rs
+++ b/src/controllers/user/me.rs
@@ -8,7 +8,7 @@ use util::bad_request;
 
 use models::{Email, Follow, NewEmail, User, Version};
 use schema::{crates, emails, follows, users, versions};
-use views::{EncodablePrivateUser, EncodableVersion};
+use views::{EncodableMe, EncodableVersion};
 
 /// Handles the `GET /me` route.
 pub fn me(req: &mut dyn Request) -> CargoResult<Response> {
@@ -40,11 +40,7 @@ pub fn me(req: &mut dyn Request) -> CargoResult<Response> {
     let verification_sent = verified || verification_sent;
     let user = User { email, ..user };
 
-    #[derive(Serialize)]
-    struct R {
-        user: EncodablePrivateUser,
-    }
-    Ok(req.json(&R {
+    Ok(req.json(&EncodableMe {
         user: user.encodable_private(verified, verification_sent),
     }))
 }
--- a/src/tests/all.rs
+++ b/src/tests/all.rs
@@ -136,7 +136,8 @@ where
         good
     }
 
-    pub fn bad(&mut self) -> Bad {
+    pub fn bad_with_status(&mut self, code: u32) -> Bad {
+        assert_eq!(self.response.status.0, code);
         match ::bad_resp(&mut self.response) {
             None => panic!("ok response: {:?}", self.response.status),
             Some(b) => b,
@@ -873,7 +874,7 @@ impl MockUserSession {
         request
     }
 
-    /// Issue a GET request
+    /// Issue a GET request as the current user
     pub fn get<T>(&self, path: &str) -> Response<T>
     where
         for<'de> T: serde::Deserialize<'de>,
@@ -882,7 +883,7 @@ impl MockUserSession {
         Response::new(self.middle.call(&mut request))
     }
 
-    /// Issue a PUT request
+    /// Issue a PUT request as the current user
     pub fn put<T>(&self, path: &str, body: &[u8]) -> Response<T>
     where
         for<'de> T: serde::Deserialize<'de>,
@@ -890,6 +891,15 @@ impl MockUserSession {
         let mut builder = self.request_builder(Method::Put, path);
         let request = builder.with_body(body);
         Response::new(self.middle.call(request))
+    }
+
+    /// Issue a DELETE request as the current user
+    pub fn delete<T>(&self, path: &str) -> Response<T>
+    where
+        for<'de> T: serde::Deserialize<'de>,
+    {
+        let mut request = self.request_builder(Method::Delete, path);
+        Response::new(self.middle.call(&mut request))
     }
 
     /// Log out the currently logged in user.
--- a/src/tests/all.rs
+++ b/src/tests/all.rs
@@ -109,6 +109,13 @@ impl<T> Response<T>
 where
     for<'de> T: serde::Deserialize<'de>,
 {
+    pub fn new(response: ResponseResult) -> Self {
+        Self {
+            response: t!(response),
+            callback_on_good: None,
+        }
+    }
+
     pub fn with_callback(response: ResponseResult, callback_on_good: Box<Fn(&T)>) -> Self {
         Self {
             response: t!(response),
@@ -132,6 +139,10 @@ where
             None => panic!("ok response: {:?}", self.response.status),
             Some(b) => b,
         }
+    }
+
+    pub fn assert_status(&self, status: u32) {
+        assert_eq!(self.response.status.0, status);
     }
 }
 
@@ -807,10 +818,22 @@ struct MockUserSession {
     _bomb: record::Bomb,
     middle: conduit_middleware::MiddlewareBuilder,
     request: MockRequest,
-    _user: User,
+    _user: Option<User>,
 }
 
 impl MockUserSession {
+    pub fn anonymous() -> Self {
+        let (bomb, app, middle) = app();
+        let request = MockRequest::new(Method::Get, "/");
+        MockUserSession {
+            app,
+            _bomb: bomb,
+            middle,
+            request,
+            _user: None,
+        }
+    }
+
     /// Create a session logged in with an arbitrary user.
     pub fn logged_in() -> Self {
         let (bomb, app, middle) = app();
@@ -832,7 +855,7 @@ impl MockUserSession {
             _bomb: bomb,
             middle,
             request,
-            _user: user,
+            _user: Some(user),
         }
     }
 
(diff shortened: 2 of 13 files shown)


jtgeibel:
I've cleaned up the testing API a bit, converted some more tests, and rebased to resolve merge conflicts.

jtgeibel:
Okay, I've pushed my last big commit to this PR.  I think I've finally settled on an API that works well for all the tests I've attempted to convert so far.

One downside with the `MockUserSession` API was that it did not seem like it would be easy to extend it to issue requests using an API token.  Additionally, for tests that issue requests from multiple users it was necessary to keep multiple user values around to sign in with, in addition to the session itself.  Most tests (any?) don't actually use session cookies or exercise the login/logout routes, and the session based API appeared to give the opposite impression.

With the new API, there are 3 types that implement a new trait.  Any of these types can be used to issue requests, and each individual type uses a different scheme to handle authentication.  A test can obtain multiple `MockCookieUser` values and use them to issue requests in any sequence without going through the sign-in/logout dance.

For additional explanation, here is a copy of the documentation at the top of the new `src/tests/util.rs` file:

Tests start by using one of the `TestApp` constructors.  All constructors return at least a `TestApp` and `MockAnonymousUser`.  The `MockAnonymousUser` can be used to issue requests in an unauthenticated session.

A `TestApp` value provides raw access to the database through the `db` function and can construct new users via the `db_new_user` function.  This function returns a `MockCookieUser`, which can be used to generate one or more tokens via its `db_new_token` function, which in turn returns a `MockTokenUser`.

All three user types implement the `RequestHelper` trait which provides convenience methods for constructing requests.  Some of these methods, such as `publish` are expected to fail for an unauthenticated user (or for other reasons) and return a `Response<T>`.  The `Response<T>` provides several functions to check the response status and deserialize the JSON response.

`MockCookieUser` and `MockTokenUser` provide an `as_model` function which returns a reference to the underlying database model value (`User` and `ApiToken` respectively).

src/tests/all.rs
```diff
@@ -237,16 +237,15 @@ fn app() -> (
...
-    let env_result = env::var(s).ok().unwrap_or_default();
-
-    if env_result == "" {
-        panic!("must have `{}` defined", s);
+fn env(var: &str) -> String {
+    match env::var(var) {
+        Ok(ref s) if s == "" => panic!("environment variable `{}` must not be empty", var),
+        Ok(s) => s,
+        Err(_) => panic!(
+            "environment variable `{}` must be defined and valid unicode",
```
carols10cents: LOL this sounds like some personal experience in here...

src/tests/all.rs
```diff
@@ -855,9 +855,14 @@ impl MockUserSession {
...
     // }
 
-    /// Obtain the database connection
-    pub fn db_conn(&self) -> DieselConnection {
-        self.app.diesel_database.get().unwrap()
+    /// Obtain the database connection and pass it to the closure
+    ///
+    /// Within each test, the connection pool only has 1 connection so it is necessary to drop the
+    /// connection before making any API calls.  Once the closure returns, the connection is
+    /// dropped, ensuring it is available for any API calls.
```
carols10cents: This is nice! I think multiple people have hit this problem when starting to write new tests, and this API is much better than having to add curly braces for what might seem like a magical reason.

src/tests/all.rs
```diff
@@ -865,76 +855,69 @@ impl MockUserSession {
         f(&conn)
     }
 
-    /// For internal use only: make the current request
```
carols10cents: I wasn't totally happy with this API... and you've made it so much better. I don't know why I didn't consider just adding the authentication info to every request, but it makes so much sense now that you've implemented it!

src/tests/krate.rs
```diff
@@ -1511,9 +1511,11 @@ fn yank_not_owner() {
         CrateBuilder::new("foo_not", another_user.id).expect_build(conn);
     });
 
-    let mut response = session.yank("foo_not", "1.0.0");
-
-    ::json::<Bad>(&mut response);
+    let json = session.yank("foo_not", "1.0.0").bad();
+    assert_eq!(
+        json.errors[0].detail,
+        "crate `foo_not` does not have a version `1.0.0`"
```
carols10cents: 👍 👍 👍

src/tests/keyword.rs
```diff
@@ -38,7 +38,7 @@ fn index() {
 fn show() {
     let url = "/api/v1/keywords/foo";
     let session = MockUserSession::anonymous();
-    session.get::<()>(url).assert_status(404);
```
carols10cents: goodbye unit turbofish, we barely knew ye

src/tests/token.rs
```diff
@@ -32,51 +31,33 @@ macro_rules! assert_contains {
...
+static NEW_BAR: &[u8] = br#"{ "api_token": { "name": "bar" } }"#;
+
 #[test]
 fn list_logged_out() {
-    let (_b, _, middle) = app();
-    let mut req = req(Method::Get, "/api/v1/me/tokens");
-
-    let response = t_resp!(middle.call(&mut req));
-    assert_eq!(response.status.0, 403);
+    MockUserSession::anonymous().get(URL).assert_forbidden();
```
carols10cents: SO MUCH CLEARER!!!

src/tests/user.rs
```diff
@@ -23,37 +27,35 @@ pub struct UserShowPrivateResponse {
     pub user: EncodablePrivateUser,
 }
 
+#[derive(Deserialize)]
+struct CrateList {
```
carols10cents: I'm pretty sure these tests can use the `CrateList` defined in tests/all.rs, I'll try after finishing this review.

src/controllers/user/me.rs
```diff
@@ -8,7 +8,7 @@ use util::bad_request;
 
 use models::{Email, Follow, NewEmail, User, Version};
 use schema::{crates, emails, follows, users, versions};
-use views::{EncodablePrivateUser, EncodableVersion};
+use views::{EncodableMe, EncodableVersion};
```
carols10cents: There are a few changes in a few commits (like this change to extract `EncodableMe` in a commit about moving `MockUserSession` to `util::TestApp`) that don't go with their commit's message, but they're pretty small so I'm not going to make you fix them. Just wanted to state for the record that you didn't get anything in without me noticing ;P

src/tests/all.rs
```diff
@@ -88,73 +82,6 @@ macro_rules! bad_resp {
     }};
 }
 
-#[derive(Deserialize, Debug)]
```
carols10cents: Yay for tests/all.rs getting smaller instead of larger for once!! :guilty:

src/tests/krate.rs
```diff
@@ -67,20 +67,28 @@ struct SummaryResponse {
     popular_categories: Vec<EncodableCategory>,
 }
 
+impl ::util::MockTokenUser {
```
carols10cents: Eeeeeenteresting. I'm not sure how I feel about these test-file-local `impl`s. My worry is that there isn't one place to go to see all the actions that are nicely implemented for you to call in tests (and see what the method names are and how to call them), which could potentially lead to multiple implementations of the same functionality.... however, if there are actions that are really only relevant to one file of tests, then that gives you less you need to look through, if you look at the globally available actions and the locally available actions.

Typing this out, I'm wondering if maybe this will push us to better test organization-- create a file that's only about yanking crates, say, and this function will only be needed there, because not all the tests in this file have to do with yanking.

Hhhhmmmm. I'm mostly just rambling, let's keep this and see how it works out over time.

src/tests/version.rs
```diff
@@ -113,18 +114,18 @@ fn record_rerendered_readme_time() {
 
 #[test]
 fn version_size() {
-    let mut session = MockUserSession::logged_in();
+    let (_, _, user) = TestApp::with_user();
```
carols10cents: I'm not a huge fan of these places that we have to do `let (_, _, ...`, but I don't see enough of a pattern yet to have a suggestion for how to get rid of them :-/

src/controllers/user/me.rs
jtgeibel: Ha ha, yeah.  In general, I think there are more places we can extract view types in the main app and use the same types in the tests.  Maybe there is a reason we would want to have separate but identical types in the test, but I can't really think of problems that would help catch.

src/tests/krate.rs
jtgeibel: For the 2 files where I did this, I did a grep on the format strings and I'm pretty sure these won't be needed elsewhere, at least for our current suite of tests.  I agree we want to be careful not to duplicate these in multiple files.

The other downside is that the impl is on only one of the 3 concrete types.  So there is a lot of motivation to move this to the trait if the functionality is generally useful.

src/tests/version.rs
jtgeibel: Yeah, I see pros and cons here.  I do like that it is easy to look at the first line of a test and know pretty quickly that the test does not access the database directly (app is dropped), and makes only authenticated API request (anon is dropped).  It certainly does look strange in a few cases like this though.

I considered adding some sort of `SimpleTest` constructor that could hide some of this, but I'm not really sure what that should look like yet or if it is really necessary.  Hopefully as more tests are converted we will get a better feel for exactly what would be useful.

PR #1516, opened 2018-10-12
Refactor how we convert from `T: Error` to `dyn CargoError`

sgrif:
The way we handled this before with a `Shim` struct breaks any later
downcasting we may try to do, and can lead to errors slipping through in
a hard to debug way. It's much easier if we just directly implement
`CargoError` on those types (and lets us get rid of some trivial impls).

This new impl conflicts with `impl<T: CargoError> CargoError for
Box<T>`. It's fine to remove that impl. It means that
`Box<ConcreteType>` no longer implements `CargoError`, but we wouldn't
have boxed the error in the first place unless we needed `dyn
CargoError`.


jtgeibel:
LGTM!

bors: r+

PR #1518, opened 2018-10-12
Make `CargoError` implement debug

sgrif:
If you don't have the concrete error type you expect, or aren't sure
what you even have, figuring out the answer is incredibly frustrating
without this impl. We do not ever use the debug implementation outside
of tests.


jtgeibel:
bors: r+

PR #1519, opened 2018-10-13
Add `TestApp::init()` for tests that don't need HTTP recording

jtgeibel:
The unused `Uploader::NoOp` option is now `Uploader::Panic` which
panics on any usage.  The panic message directs users to initialize the
app with `TestApp::with_proxy()`.

While this adds more panics to the main lib, the application
configuration isn't dynamic and it should be obvious that
`Uploader::Panic` is not to be used to configure a production
application.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -103,8 +103,8 @@ impl App {
     /// Returns a client for making HTTP requests to upload crate files.
     ///
     /// The handle will go through a proxy if the uploader being used has specified one, which
-    /// is only done in test mode in order to be able to record and inspect the HTTP requests
-    /// that tests make.
+    /// is only done in tests with `TestApp::with_proxy()` in order to be able to record and
+    /// inspect the HTTP requests that tests make.
     pub fn http_client(&self) -> CargoResult<reqwest::Client> {
         let mut builder = reqwest::Client::builder();
         if let Some(proxy) = self.config.uploader.proxy() {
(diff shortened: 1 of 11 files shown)


jtgeibel:
The test suite currently has 33 (of 166 total) tests that need a thread running an HTTP recording proxy.  This PR avoids that overhead for 30 of the unnecessary cases.  On my machine (with an average of 3 runs each), this reduces my user time from 26.60 to 26.11 seconds.  An additional second or two can probably be saved once all test are converted.

carols10cents:
Nice! I'm into it! The only slight reservation I have is about the names. I'm trying to think of this from the point of view of a new contributor writing some new tests. They have the `empty`/`with_user`/`with_token` axis to choose from, which is mostly pretty clear I think, and this adds the `simple`/`full` axis that I'm not sure is as clear. 

The proxy change also affects tests that talk to github as well as those that talk to S3, right? The current doc comments on `simple` only talks about uploading; `full` talks about generic outgoing HTTP requests though, so that's good.

Some ideas for names, none of which I really like yet but maybe it'll help us get to something good:

- `proxy`/`no_proxy`
- `record_http`/`no_record_http`
- `makes_external_http_requests`/`no_external_http_requests`

Or could we make `simple` be the default and only require calling a method when the test makes http requests? That would avoid the `no_*` names above....

jtgeibel:
I renamed `simple` to `init` and `full` to `with_proxy`.  I rebased to fix merge conflicts and updated the tests converted in the other PR.  There are now 50 tests using `init` and 3 using `with_proxy`.

bors: r+

PR #1521, opened 2018-10-13
Convert http-data to pretty JSON

jtgeibel:
The includes the following changes to the test logic:

* http-data is now generated with `serde_json::to_string_pretty()`
* The User-Agent header is ignored so that it doesn't cause conflicts when reqwest is updated
* The body is now encoded as a base64 string because a byte slice is serialized with one byte per line.  This makes JSON responses from GitHub opaque, but that was the case with byte slices as well.

The script that was used to do the conversion is archived in the first commit, but removed in the second.


src/tests/record.rs
```diff
@@ -275,6 +276,10 @@ fn replay_http(
     mut exchange: Exchange,
     stdout: &mut Write,
 ) -> Box<Future<Item = hyper::Response<hyper::Body>, Error = hyper::Error> + Send> {
+    fn check_ignored_request_header(name: &str) -> bool {
```
sgrif: What do you think about making this a static vec instead, and calling `IGNORED_HEADERS.contains`?

sgrif:
LGTM other that comment

jtgeibel:
I've pushed a new commit switching to a static `&[&str]`.

Issue #1523, opened 2018-10-16
Short crate URLs


timvisee:
I'm accessing crate pages on `crates.io` quite a lot, by directly typing the `crates.io/crates/rand` URL in my browsers address bar. I think it's quite verbose/cumbersome, and type `create` a little too often.

What about implementing a short crate URL such as `crates.io/c/rand` in a similar fashion as reddit does with subreddits and user profiles. The short URL could be an alias, or a redirect to the full URL. Existing URLs won't break, of course.

I'm interested in what other think? It's a small but useful change I believe. I couldn't find a previous discussion on this topic. I'm happy to implement this through a PR if others can confirm such a change is desired.

sgrif:
I'd be fine with this as long as it was implemented as a redirect.

timvisee:
@sgrif Awesome! I've submitted a basic implementation with redirects, see #1526 .

carols10cents:
I've decided to close this, as browsers and search engines provide even shorter ways to do this:

- In Firefox, create a bookmark for crates.io with the location set to `https://crates.io/crates/%s` and the keyword set to `crate`. Then you can type "crate rand" into your address bar and it'll take you to https://crates.io/crates/rand.
- In Chrome, go to settings, search engine, manage search engines, then add a search engine with the keyword `crate` and the URL set to `https://crates.io/crates/%s`. Then you can type "crate rand" into your address bar and it'll take you to https://crates.io/crates/rand.
- [Duck Duck Go](https://duckduckgo.com) has a [bang search](https://duckduckgo.com/bang) for [crates.io](https://duckduckgo.com/bang?q=crates), so you can do `!crate rand` and it'll search crates.io for rand. Make DDG the default search in your browser and it'll work from your address bar as well.

timvisee:
Sure. Sorry that I want really able to come up with a proper implementation at the time. I agree this wouldn't be to useful anymore these days. Thanks for giving me the opportunity anyway!

PR #1526, opened 2018-10-17
Add crate(s) URL shorthand which redirects to full URL

timvisee:
Implements #1523 

This PR implements shorthand URLs for crates, to make them easier to type. These shorthand URLs are redirected to their long variant respectively.

The following redirects have been implemented:
- `/c` => `/crates`
- `/c/:crate_id` => `/crates/:crate_id`

I hope the implementation is correct and follows convention. I'm new to `ember`.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -26,6 +26,8 @@ Router.map(function() {
         this.route('docs');
         this.route('repo');
     });
+    this.route('c');
+    this.route('c', { path: '/c/:crate_id' });
     this.route('me', function() {
         this.route('crates');
         this.route('following');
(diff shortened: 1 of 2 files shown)


timvisee:
I just noticed that routes like `/c/rand/0.5.5` aren't redirected properly, the version number is dropped. The same happens for routes like `/c/rand/docs`/.

As I've said I'm not too familiar with `ember` yet. I'll attempt to figure out how to solve this though without redefining all possible crate routes with a shorthand.

sgrif:
This should have tests added as well

timvisee:
Sadly, I currently can't figure out how to properly implement this. Also, testing redirects seems to be broken at this moment, where the accessibility test client doesn't want to follow transitions/redirects.

I'll close this PR for now. Maybe I'm able to properly implement this at a later time.

PR #1527, opened 2018-10-17
Add a missing index on crates

sgrif:
Our most time consuming query is coming from crawlers on /crates ordered
by name. It turns out this query was not using any indexes, as the only
index we have on `crates.name` is `canon_crate_name(name)`. An
alternative would be to order by `canon_crate_name(name)` instead, but
that shouldn't actually be able to change the order. This is a low write
table, so the extra index isn't a major cost. Adding an index instead of changing
the query means the effects of this change will be easier to see.

The weird index name is because we already have an index called `index_crates_name`.


sgrif:
bors: r+

PR #1529, opened 2018-10-18
Remove dependencies on openssl 0.9

jtgeibel:
After recently updating to Ubuntu 18.10, I ran into the issue that `openssl 0.9` doesn't support OpenSSL 1.1.1. Fortunately `openssl 0.10` is already compatible.  This PR bumps our direct dependency on `openssl`.  It also bumps `hyper-tls` and `lettre` to update their transitive dependencies on `openssl`.

It is possible to build against a custom (non-system) version of OpenSSL, which is what I'm doing for now.  I can help with that if anyone else runs into this issue before this is merged.


jtgeibel:
I think this should be safe to merge, but it may still make sense to do the deploy-merge-deploy dance just in case there are issues in production.  Therefore, I'm holding off on merging this myself until someone is ready to deploy this in case it needs to be reverted.

Having said that, I believe this PR is low risk.  In the lockfile, this PR removes openssl 0.9.24, and makes the following updates: openssl 0.10.12 -> 0.10.13 and openssl-sys 0.9.36 -> 0.9.38.  We were already pulling in the newer openssl crate via our tests, and now the necessary production dependencies have been converted to the newer version as well.

On Heroku, incoming HTTPS connections are terminated in their routing layer, but we internally use openssl for: git and API interactions with GitHub, transmitting outbound email, publishing to S3, and calculating the tarball hash which is recorded in the index.

PR #1533, opened 2018-10-23
Rename blacklist_ips to block_ips

sgrif:


carols10cents:
there's another place i've been meaning to do this too

bors: r+

PR #1534, opened 2018-10-23
Require all traffic to provide a `User-Agent` header.

sgrif:
We want to be able to actually differentiate crawlers from each other,
so I've nudged them towards actually using a unique user agent (we
probably won't ever actually block generic UAs since folks sometimes do
actually use curl/wget from the command line).

Additionally, I've had a lot of cases lately where a crawler has been
outside of what we allow, but wasn't actually causing a service impact.
If I could contact those people without having to block their traffic, I
would. So I've also worded the message to try and nudge folks towards
including contact info, which most commercial bots already do.

### DIFF
--- a/src/middleware/mod.rs
+++ b/src/middleware/mod.rs
@@ -20,6 +20,7 @@ mod ember_index_rewrite;
 mod ensure_well_formed_500;
 mod head;
 mod log_request;
+mod require_user_agent;
 mod security_headers;
 mod static_or_continue;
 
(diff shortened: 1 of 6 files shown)


sgrif:
bors: r+

sgrif:
Yes, we learned that yesterday afternoon. This change has since been reverted and we're going to be discussing it during the next team meeting.

cuviper:
OK, thanks for the info!

Issue #630, opened 2017-03-15
Provide a databse snapshot to facilitate development.


LeopoldArkham:
It would be nice to have a snapshot of (a subset of) the database that someone wishing to contribute could load locally to test out the code on real data. You can fake crates, dependencies, downloads, etc., but that only gets you so far.

Eh2406:
I recall there being an issue for creating sanitized sql dumps of the database. Such a dump, would be useful for this aswell. Of course now I cannot find the issue.

carols10cents:
#410 is the closest one I know of, going to close that in favor of this one since we want to keep the index small. 

Notable info from that issue is that [rubygems.org provides a sanitized snapshot](https://rubygems.org/pages/data), i'll probably look into their infrastructure to see if there's something we could reuse.

The easiest thing to do sanitation-wise is probably to just not dump the users and follows tables. Actually we should probably make this a white list instead of a black list of tables to dump so that we don't leak data by adding a table that should be stripped out and forgetting to add it to the blacklist.

Eh2406:
That is the one I had in mind, thanks! When we have a game plan for this I'd love to help.

moore3071:
Would it be possible to use the repository [crates.io-index](https://github.com/rust-lang/crates.io-index) to create a quick development database? This would make it a lot harder to accidentally leak any private data.

icefoxen:
Do people still want this?  Not having to scrape the crates.io API for things like downloads and crate owners would be *great*.  I'll happily throw together a PR.

icefoxen:
Ok, after a little research it looks like heroku lets you make backups, pull backups to a local machine or such, and so on, but not do what we want directly, which is to be very careful to dump *only certain tables* to a backup.  Seems like the easy way would be to use `heroku pg:pull` to pull a remote database to a local postgres instance, call `pg_dump` and give it your whitelist of tables to dump, then upload the results to an S3 bucket or something and have a URL that redirects to that.  Not sure how big/unwieldy the database is, and I dunno how easy it is for crates.io to set up these sort of ad-hoc scripts; this is really not a programming problem as much as an infrastructure problem.

smarnach:
Given the recent discussion about squashing the index, I took a look at this issue. Here is a suggestion for a design.

## Goals

* Allow crawlers to download a single archive with all information instead of hitting crates.io thousands of times to retrieve all information on all crates.
* Facilitate local testing in crates.io development (e.g. query optimisation with a realistic dataset).
[...]

PR #1535, opened 2018-10-23
Publish an official crawler policy

sgrif:
This is a formalization of a policy that we've been informally enforcing
for some time now. The policy basically boils down to:

- Just use the index if you can
- Contact us to see if we can help in a way that doesn't require
  crawling if you can't.
- If you do crawl, limit to 1RPS
- You also have to provide a user agent, which should actually identify
  your crawler and have contact information
- We may still block you if you cause an impact on the integrity of the
  service.

I chose not to explicitly call out inflating a single crate's download
numbers as something that's forbidden, as it felt like doing that would
be an instance of ["Don't shove beans up your
nose"](https://en.wikipedia.org/wiki/Wikipedia:Don%27t_stuff_beans_up_your_nose)
That falls under the umbrella of "impacting the integrity of the
service" though, so this policy does give us an explicit reason to block
anyone engaging in that behavior.


ishitatsuyuki:
> If you do crawl, limit to 1RPS

I'd prefer something better to a hardcoded number. The idea I have now is to "not make concurrent requests".

[GitHub's abuse policy](https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits) may be a good reference. Some highlights are:

- They recommend requesting serially for a single user ID. (Maybe this is related to how they shard data?)
- They do recommend you to slow down requests at a specified rate but only for mutations.
- They emphasizes taking care of the `Retry-After` header which is a HTTP standard.

sgrif:
"Here are some guidelines to avoid hitting our abuse rate limits" isn't really comparable to a general crawler policy. GitHub is an entirely different kind of application, which is expecting most of its traffic to be from API clients. The policy is heavily based on a similar policy from NPM. The limit for bots is intentionally low, as we would much prefer that they use the index or reach out to us with their needs.

ishitatsuyuki:
I agree with you that we're different from GitHub.

However, I think that the reasons people are crawling is due to the index not containing enough data to run a mirror (like crates.rs) or aggregate stats. I will be working on #630 so we can eliminate most need of crawling. Also this is how we are currently different from NPM. NPM runs a CouchDB instance so that people can download and stream the full registry data anytime they want; therefore we don't need to hit the API.

Though, given that we only have 20k crates 1 request per second seems reasonable now. We probably want to revisit it when we implement generic rate limits and/or start to provide database dumps, but otherwise I have no objection to this as a general bot policy.

sgrif:
Yes, I agree that we should aim to provide better solutions to their needs. That doesn't change the fact that we have limited resources, and we need to prioritize user traffic over bots.

joshtriplett:
I would request that, rather than giving a specific rate limit, we instead tell people that the server may throttle excessive requests from a single user, and that they should not attempt to work around that throttling.

sgrif:
bors: r+

PR #1536, opened 2018-10-23
Rename documentation blacklist to blocklist

carols10cents:


sgrif:
Other than the fact that my eyes saw blacklist and blocklist as the same word at first, ltgm.

bors: r+

PR #1537, opened 2018-10-23
Bump semver

sgrif:
The amount of manual serialization/deserialization of semver stuff is
really starting to get in the way of refactoring. Semver supports both
Serde and Diesel on master (only serde in the released version, hence
the git dependency).

Conduit still expects an older version though. That project is
unmaintained and I've been having trouble getting added as an owner to
publish a new version (Alex is in Europe right now, not worth bothering
him) so we've got a little hacky workaround in the meantime.


jtgeibel:
Sounds good to me.

bors: r+

PR #1538, opened 2018-10-23
Try out the nginx realip module

sgrif:
On Heroku and any system that involves proxies, the remote_addr of the
request becomes useless, as it will be the proxy forwarding your request
rather than the client itself. Most proxies will set or append an
`X-Forwarded-For` header, putting whatever remote_addr it received at
the end of the the header. So if there are 3 proxies, the header will
look like

real_ip, proxy1, proxy2

However, if we're not careful this makes us vulnerable to IP spoofing. A
lot of implementations just look at the first value in the header. All
the proxies just append the header though (they don't know if they're
getting traffic from the origin or a proxy), so you end up with

spoofed_ip, real_ip, proxy1, proxy2

nginx, Rails, and many other implementations avoid this by requiring an
allowlist of trusted IPs. With this configuration, the realip only kicks
in if remote_addr is a trusted proxy, and then it will replace it with
the last non-trusted IP from the header.

This will simplify our own codebase, where we either have to do this
logic over and over, or in a few places we've forgotten and have bugs
(mostly logging related).

I haven't switched any code to use this yet, because I want to see our
logs saying I got this right before I do that.

### DIFF
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -2,6 +2,10 @@ daemon off;
 #Heroku dynos have at least 4 cores.
 worker_processes <%= ENV['NGINX_WORKERS'] || 4 %>;
 
+set_real_ip_from 127.0.0.1;
+real_ip_header X-Forwarded-For;
+real_ip_recursive on;
+
 events {
 	use epoll;
 	accept_mutex on;
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -2,9 +2,8 @@ daemon off;
 #Heroku dynos have at least 4 cores.
 worker_processes <%= ENV['NGINX_WORKERS'] || 4 %>;
 
-set_real_ip_from 127.0.0.1;
+set_real_ip_from 10.0.0.0/24;
 real_ip_header X-Forwarded-For;
-real_ip_recursive on;
 
 events {
 	use epoll;
@@ -53,6 +52,7 @@ http {
 			add_header Strict-Transport-Security "max-age=31536000" always;
             add_header Vary 'Accept, Accept-Encoding, Cookie';
 			proxy_set_header Host $http_host;
+			proxy_set_header X-Real-Ip $remote_addr;
 			proxy_redirect off;
 			if ($http_x_forwarded_proto != 'https') {
 				rewrite ^ https://$host$request_uri? permanent;
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -33,12 +33,13 @@ impl Handler for LogRequests {
 
         print!(
             "at={level} method={method} path=\"{path}\" \
-             request_id={request_id} fwd=\"{ip}\" service={time_ms}ms \
+             request_id={request_id} fwd=\"{ip}\" real_ip=\"{real_ip}\" service={time_ms}ms \
              status={status} user_agent=\"{user_agent}\" referer=\"{referer}\"",
             level = level,
             method = req.method(),
             path = FullPath(req),
             ip = request_header(req, "X-Forwarded-For"),
+            real_ip = req.remote_addr(),
             time_ms = response_time,
             user_agent = request_header(req, "User-Agent"),
             referer = request_header(req, "Referer"), // sic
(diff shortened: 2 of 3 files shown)


config/nginx.conf.erb
```diff
@@ -2,6 +2,10 @@ daemon off;
 #Heroku dynos have at least 4 cores.
 worker_processes <%= ENV['NGINX_WORKERS'] || 4 %>;
 
+set_real_ip_from 127.0.0.1;
+real_ip_header X-Forwarded-For;
+real_ip_recursive on;
```
jtgeibel: As far as I can tell, Heroku doesn't really have good docs on any of this.  Their [public docs](https://devcenter.heroku.com/articles/http-routing#heroku-headers) just say "`X-Forwarded-For`: the originating IP address of the client connecting to the Heroku router".  I found [this](https://stackoverflow.com/questions/18264304/get-clients-real-ip-address-on-heroku/18517550#18517550) on Stack Overflow, which says: "The router doesn't overwrite `X-Forwarded-For`, but it does guarantee that the real origin will always be the last item in the list."

Based on that, I think `real_ip_recursive` should be off (the default).  The [nginx docs](http://nginx.org/en/docs/http/ngx_http_realip_module.html#set_real_ip_from) say that with this off: "the original client address [...] is replaced by the last address sent in the request header".

config/nginx.conf.erb
```diff
@@ -2,6 +2,10 @@ daemon off;
 #Heroku dynos have at least 4 cores.
 worker_processes <%= ENV['NGINX_WORKERS'] || 4 %>;
 
+set_real_ip_from 127.0.0.1;
```
jtgeibel: Since Heroku can guarantee that all connections to us are coming from their router (and it's not possible to list out all their IP addresses), I think this should be `0.0.0.0/0`.

src/middleware/log_request.rs
```diff
@@ -33,12 +33,13 @@ impl Handler for LogRequests {
...
         print!(
             "at={level} method={method} path=\"{path}\" \
-             request_id={request_id} fwd=\"{ip}\" service={time_ms}ms \
+             request_id={request_id} fwd=\"{ip}\" real_ip=\"{real_ip}\" service={time_ms}ms \
              status={status} user_agent=\"{user_agent}\" referer=\"{referer}\"",
             level = level,
             method = req.method(),
             path = FullPath(req),
             ip = request_header(req, "X-Forwarded-For"),
+            real_ip = req.remote_addr(),
```
jtgeibel: This is going to be nginx connecting from `127.0.0.1`.  We still need to use the last value in `X-Forwarded-For` from the app.

jtgeibel: We could also add `proxy_set_header X-Real-IP $remote_addr;` to nginx if we want to pass it's interpretation along to the app.

config/nginx.conf.erb
sgrif: I figured I'd leave recursive on as we add more proxies in the future, but you're right that the behavior is the same with it on or off right now.

src/middleware/log_request.rs
sgrif: Ugh for some reason I thought that nginx would do some magic here. I like your X-Real-Ip idea.

config/nginx.conf.erb
sgrif: We will have to find a better solution when we put a CDN in front of the API, but I agree that this is fine in the short term.

sgrif: Whoops, you're right. You can find a little bit of documentation on it on the docs for internally routed apps, but mostly it's just experience and the fact that it wouldn't make sense for them to use any IP outside of the range laid out in RFC 1918

jtgeibel:
I'm having some trouble getting this working in staging :-(  First, I had to move the new directives under http, they don't seem to be valid at the global level.  When I deployed that, I'm seeing my IP address in the logs under `fwd` but the Heroku router address in `real_ip="10.148.41.22"`.  I tried moving the realip declarations under location as well, and it still doesn't seem to be working.

I considered trying `$real_remote_addr`, which the docs say "keeps the original client address".  So this doesn't sound like what we want, and the Heroku buildpack nginx is too old to try this anyway (added in 1.9.7).

carols10cents:
@sgrif could you investigate why this isn't working for @jtgeibel please?

carols10cents:

PR #1550, opened 2018-11-04
Ensure only exact name matches are added to the index

jtgeibel:
On the client side cargo checks dependency names against the index and
requires an exact match.  For instance, a user cannot depend on
`diesel-migrations` or `Diesel_migrations` as cargo generates an error
directing them to use the correct name `diesel_migrations`.

This check for an exact name match is now repeated on the server when a
crate is published.  This ensures that bugs in cargo or third-party
tooling cannot result in a dependency name that cargo is unable to
locate in the index.


sgrif:
bors: r+

Issue #159, opened 2015-06-01
Return search results for partial matches


eminence:
When you search for `ssh` you get the following results:
- ssh2 -- Bindings to libssh2 for interacting with SSH servers and executing remote commands, forwarding local ports, etc.
- fleet -- A client library for CoreOS's fleet.

When you search for 'libssh' you get the following results:
- libssh -- bindings for libssh

When you search for `libssh2` you get the following results:
- libssh2-sys -- Native bindings to the libssh2 library
- ssh2 -- Bindings to libssh2 for interacting with SSH servers and executing remote commands, forwarding local ports, etc

With the exception of 'fleet' I expected to see all 3 results when I search for just "ssh".   It seems as if the search text is being matched only on the description and not the package name?

DanielKeep:
I just wasted *hours* writing a crate that *already existed* because "tempfile" doesn't show up when you search for "temp" or "file".

azerupi:
I could be wrong, but think that at the moment crates are only returned on exact match with keywords, crate name, words in the description.

If partial matches were returned too, it would already solve a lot of issues.
If you search for `book` you get 6 results, but if you search for `boo` you get nothing.

Also I think exact matches with crate name should appear on top

AlisdairO:
Could use a little influence from number of downloads as well - arguably if you search for 'random' you'd probably want the 'rand' crate at least appearing in the top two results.  It's number 4 at the moment (so clearly not too bad!), but one could imagine other situations in which this would be worse..

nodakai:
@eminence It is working for you now?

```
$ cargo search ssh
    Updating registry `https://github.com/rust-lang/crates.io-index`
ssh (0.1.4)         Bindings to libssh, a library to write clients and servers for both version 1 and 2 of the protocol. It can also parse SSH conf…
mush (0.1.1)        Bindings to libssh2 for interacting with SSH servers and executing remote commands, forwarding local ports, etc. 
ssh2 (0.2.11)       Bindings to libssh2 for interacting with SSH servers and executing remote commands, forwarding local ports, etc. 
actiondb (0.6.1)    A safe and efficient unstructured text (log) parsing library.
fleet (0.2.0)       A client library for CoreOS's fleet.
$ cargo search libssh
    Updating registry `https://github.com/rust-lang/crates.io-index`
libssh (0.0.1)    bindings for libssh
[...]

PR #1560, opened 2018-11-24
loosen search

FreeMasen:
To address #159, 

It might be worth it to add a new flag for searching with the original behavior.

I have played around with adding a flag to the query params, `strict` to allow the original behavior. It is very much a viable option, I didn't want to make any UI changes w/o first checking in about what it should look like.


sgrif:
Hi @FreeMasen,

Thanks for the PR. This needs some more work before it can be merged. To start, you've got a very large `loose_search` function which is never used, and should be removed. The comparison should also be case insensitive here, so we should use `ilike` instead of like. This also doesn't have any tests, which will be needed (in particular we'd like to be sure these are at the bottom of the list when sorted by "relevance" -- we're discussing larger changes to search in general, but in the mean time we don't want results that only came in from `like` on the title to be higher than those that are ranked highly in the existing search).

Let me know if you're still interested in working on this, or if you need additional help. I'll do a more in depth review once these issues are addressed.

FreeMasen:
I added a test with one of the expectations from the original issue (#159). I am not sure what the expectation should be so let me know if you are looking for something specific.

FreeMasen:
It appears that I accidentally close this PR, reopening

src/models/krate.rs
```diff
@@ -235,6 +240,21 @@ impl<'a> NewCrate<'a> {
...
 
 impl Crate {
+    /// SQL filter with the `like` binary operator
+    /// ```sql
+    /// SELECT *
+    /// FROM crates
+    /// WHERE name like $1
+    /// ```
+    pub fn like_name(name: &str) -> LikeName<'_> {
+        canon_crate_name(crates::name).ilike(canon_crate_name(name))
```
sgrif: Just `like` is enough here, `canon_crate_name` already normalizes case

src/models/krate.rs
```diff
@@ -235,6 +240,21 @@ impl<'a> NewCrate<'a> {
 }
 
 impl Crate {
+    /// SQL filter with the `like` binary operator
+    /// ```sql
+    /// SELECT *
+    /// FROM crates
+    /// WHERE name like $1
```
sgrif: Do we need the SQL examples here?

src/tests/krate.rs
```diff
@@ -414,6 +414,69 @@ fn exact_match_on_queries_with_sort() {
...
+            .description("description")
+            .keyword("kw1")
+            .expect_build(conn);
+        // tempfile should appear 4th
+        let four = CrateBuilder::new("tempfile", user.id)
+            .readme("readme")
+            .description("description")
+            .keyword("kw1")
+            .expect_build(conn);
+        // mkstemp should appear 5th
```
sgrif: Can you put a bit more context in here? Why should this be 5th instead of tempfile?

src/tests/krate.rs
```diff
@@ -414,6 +414,69 @@ fn exact_match_on_queries_with_sort() {
     assert_eq!(json.crates[3].name, "other_sort");
 }
 
+#[test]
+fn loose_search_order() {
```
sgrif: This test is extremely long given how little code is being changed. Do you think this might be retesting things that are already tested elsewhere? Would it be worth trying to narrow the scope a bit to make it more clear what's being tested?

src/tests/krate.rs
FreeMasen: The only reason that is the case is because it was inserted later. I can remove that one.

src/tests/krate.rs
FreeMasen: General sort order is being addressed in other tests. I could reduce the number of creates to simplify the test. The basic idea is that we are now including crates that are partial matches. One of the previous comets about this PR was that the change should't impact current ordering so I tried to capture a few pieces of information that go into the current ordering like `_` working like white space or description matching superseding this new search. I can remove the non-matching crate and the extra partial match to shorten things up?

src/models/krate.rs
FreeMasen: I can change this back, the only reason I used `ilike` was from your previous comment.

src/models/krate.rs
FreeMasen: They can be removed.

src/models/krate.rs
sgrif: Sorry about that :(

carols10cents:
Hi @FreeMasen! I'm sorry about how long this has taken to review; thank you for your patience! 

This looks great to me, aside from one little thing-- because the query with `%`s added is only used in the `like_name` function, and is sort of an implementation detail of the substring matching anyway, I moved the `format!` into `like_name`. I went ahead and took care of that because of how long you've been waiting, I didn't want to have to make you wait any more to see this merged!

I forget if I have to wait for travis or if I can tell bors to merge now, let's find out and I'll come back later if this isn't valid:

@bors r+

FreeMasen:
@carols10cents Thank you so much for doing all that!

Issue #1478, opened 2018-08-20
Record who published a version of a crate


kornelski:
Currently neither the crates index nor the carates io api attribute published versions to users who published them. 

When a crate has multiple owners it's not possible to establish who published what. 

This attribution is needed for:

* Auditing and monitoring. In case something unexpected gets published it's important to know whose credentials were stolen or misused. 

* Evaluation. On crates.rs I'm experimenting with several ideas -  detecting abandoned crates (inactive authors) and finding trusted users based on relationships between crates, but the data is incomplete without knowing who exactly published each crate. 
 
Ideally I'd like to see ~~username~~ GitHub User ID for every published version. It could be as part of the crates index. If bloating of the index should be avoided, it'd be ok as crates.io API call as well. 

----

## Implementation instructions

(added by @carols10cents)

### Backend changes:

- [ ] Create a migration (using `diesel migration generate`, [as explained in this guide](http://diesel.rs/guides/getting-started/)) that adds a new column named `published_by` to the `versions` table. It should reference users.id and be nullable because we don't have the published by information for any versions published before this change.

- [ ] Add a published_by field to [the NewVersion struct](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/models/version.rs#L32-L40)

- [ ] Within the [crate publish transaction](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/controllers/krate/publish.rs#L69), [add to the new version record created](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/controllers/krate/publish.rs#L129-L139) a published_by field set to [user.id](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/controllers/krate/publish.rs#L42)

- [ ] Add to the [EncodableVersion](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/views/mod.rs#L184) type a field named `published_by` that's an instance of [EncodableOwner](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/views/mod.rs#L120-L128), which should allow for display in the frontend if we want and should give enough information to API users.

jtgeibel:
[...]

PR #1561, opened 2018-11-24
Record who published crate versions

Andy-Bell:
This PR is designed to address #1478.

I believe it answers all the check boxes listed there, but I am not 100% happy on how I have fixed the tests in the last commit here (there may not need to be the passing of the user_id values quite as much?).

Also the EncodableVersion just provides the ID of the Owner, not the full struct of EncodableOwner, as I could not work out how to do that cleanly - any insight would be appreciated


carols10cents:
Ugh github is being weird, it won't let me comment on lines. So here's my review:

- The published_by column actually needs to be nullable (so remove the `NOT NULL` constraint from the migration); we won't have this information for any versions published before this change gets deployed.

- That also means the `published_by` field on the `Version` struct should be type `Option<i32>`

- I think [I moved the `VersionBuilder` and `CrateBuilder` out from underneath you](https://github.com/rust-lang/crates.io/commit/a497d47c174d859f3bb8a87cdfe8d611c1cb946d); they're now in [src/tests/builders.rs](https://github.com/rust-lang/crates.io/blob/4eedb3ab6d7fe0716d5ac24f96b1b1e54c383047/src/tests/builders.rs). Mind moving your changes over there and removing the ones that are showing as added in this PR?

- Is there a reason you [pass `self.owner_id` from `CrateBuilder` `expect_build` to `CrateBuilder` `build`](https://github.com/rust-lang/crates.io/pull/1561/files#diff-56e85382157288e5e908b7c8ffd84363R491)? `build` can just use `self.owner_id` directly like it does [in this spot](https://github.com/rust-lang/crates.io/pull/1561/files#diff-56e85382157288e5e908b7c8ffd84363R435) rather than using the `user_id` parameter, right? I guess that's what you meant in the PR description by "(there may not need to be the passing of the user_id values quite as much?)", I think you're correct :)

I'm taking a look at what problems you might have encountered when trying to add the full struct of EncodableOwner to EncodableVersion now...

carols10cents:
Ok, so for including the information with the EncodableVersion:

- Taking a closer look, I think it actually makes more sense for published_by to be an EncodablePublicUser, not an EncodableOwner (because this will always be a person who typed `cargo publish`, not a team).
- As far as how to implement this, what about [taking a `User` as an argument to `Version` `encode`](https://github.com/rust-lang/crates.io/blob/4eedb3ab6d7fe0716d5ac24f96b1b1e54c383047/src/models/version.rs#L43), and then wherever we query to get versions to pass to `encode`, such as in [the krate metadata controller](https://github.com/rust-lang/crates.io/blob/4eedb3ab6d7fe0716d5ac24f96b1b1e54c383047/src/controllers/krate/metadata.rs#L217-L224), we also join to the users table on `published_by`?
- Then `Version` `encode` can call [`encodable_public` on that user](https://github.com/rust-lang/crates.io/blob/4eedb3ab6d7fe0716d5ac24f96b1b1e54c383047/src/models/user.rs#L185) to get an `EncodablePublicUser`.

Does that make sense? Would that solve whatever problems you were running into or no? Please ask more questions if I've guessed wrong!

carols10cents:
I think moving your changes on `CrateBuilder` and `VersionBuilder` to the builders.rs file should fix the travis failures.

carols10cents:
Hi @Andy-Bell, I hope you don't mind, but I'm planning on building off what you've started here for the time-sensitive requiring of verified email addresses to publish crates, so I finished the last piece of returning an `EncodablePublicUser` from the HTTP API. The queries were a bit tricky to get right actually! So no worries!

Because I've now contributed some stuff to this PR, I think someone else should review and merge this, at least my commits. Yours look great to me, thank you so much!

carols10cents:
I incorporated this PR into #1621, so it's going to get merged when that gets merged. Thank you @Andy-Bell!!!

PR #1562, opened 2018-11-28
Convert more tests

jtgeibel:
These commits contain minor test cleanups and convert some additional tests to the new infrastructure.

### DIFF
--- a/src/tests/user.rs
+++ b/src/tests/user.rs
@@ -1,13 +1,11 @@
-use std::sync::atomic::Ordering;
-
 use conduit::{Handler, Method};
 use diesel::prelude::*;
 
 use builders::{CrateBuilder, VersionBuilder};
-use models::{ApiToken, Email, NewUser, User};
+use models::{Email, NewUser, User};
 use util::RequestHelper;
 use views::{EncodablePrivateUser, EncodablePublicUser, EncodableVersion};
-use {app, logout, new_user, req, sign_in_as, OkBool, TestApp, NEXT_GH_ID};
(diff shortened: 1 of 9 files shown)


src/tests/user.rs
```diff
@@ -262,13 +267,9 @@ fn user_total_downloads() {
...
-    #[derive(Deserialize)]
-    struct Response {
-        total_downloads: i64,
-    }
-    let response: Response = ::json(&mut response);
-    assert_eq!(response.total_downloads, 30);
-    assert!(response.total_downloads != 32);
+    let stats: UserStats = ::json(&mut response);
+    assert_eq!(stats.total_downloads, 30);
+    assert!(stats.total_downloads != 32);
```
carols10cents: o.O i know you didn't do this, but we're asserting that `total_downloads` is 30 and then asserting that it's not 32????

src/tests/user.rs
```diff
@@ -154,138 +144,87 @@ fn following() {
...
-    let mut req = req(Method::Get, &format!("/api/v1/users/{}/stats", u.id));
-    let mut response = ok_resp!(middle.call(&mut req));
+    });
 
-    let stats: UserStats = ::json(&mut response);
-    assert_eq!(stats.total_downloads, 30);
-    assert!(stats.total_downloads != 32);
+    let url = format!("/api/v1/users/{}/stats", user.id);
+    let stats: UserStats = anon.get(&url).good();
+    assert_eq!(stats.total_downloads, 30); // instead of 32
```
carols10cents: OH PHEW!!!

carols10cents:
bors: r+

PR #1564, opened 2018-11-29
Rename structs in `krate_publish` with prefix `Encodable`

joshleeb:
This PR renames the structs in `views/krate_publish` to have the prefix `Encodable`, as per a `TODO` that was present.


joshleeb:
A small note: I didn't include any compatibility typedefs as I wasn't sure if that would be necessary, or if the general process is to rather bump the version in this instance. Happy to add them in if required 😄

jtgeibel:
Looks great, thanks @joshleeb!

bors: r+

PR #1565, opened 2018-11-29
Email verification warning

carols10cents:
This is the start of the implementation of https://github.com/rust-lang/crates-io-cargo-teams/issues/8. 

We can start warning now; if folks are using nightly, cargo will show it. Next release, beta will show it, and then the release after that, stable will show it. We have until the release after THAT to implement the hard error. I think. If I've counted correctly.


jtgeibel:
I tested this in staging with nightly cargo and got the expected warning.  Looks great!

bors: r+

Issue #1548, opened 2018-11-01
Audit trail for more actions


carols10cents:
Further steps beyond #1478. These changes wouldn't be exposed in the UI but would be in the database for us to use in determining who took what actions when.

- [ ] Add a column to the api_tokens table named `revoked` that's a BOOLEAN NOT NULL DEFAULT 'f'. Then instead of deleting tokens, mark them as revoked.

- [ ] Create a migration (using `diesel migration generate`, [as explained in this guide](http://diesel.rs/guides/getting-started/)) that adds a new table named `version_owner_actions` with the following columns:
  - id SERIAL PRIMARY KEY
  - version_id FOREIGN KEY REFERENCES versions (id)
  - owner_id FOREIGN KEY REFERENCES users (id)
  - owner_token_id FOREIGN KEY REFERENCES api_tokens (id)
  - action INTEGER NOT NULL (that would map to a Rust enum with variants publish, yank, unyank)
  - time TIMESTAMP NOT NULL DEFAULT now()

- [ ] Migrate the current `published_by` column on the `versions` table to be actions in the `version_owner_actions` table instead.

- [ ] Keep the [headers[0]](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/middleware/current_user.rs#L45) authorization header value accessible by holding onto it in the [`AuthenticationSource::ApiToken` variant](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/middleware/current_user.rs#L18)

- [ ] Within the [crate publish transaction](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/controllers/krate/publish.rs#L69), [after the new version record is created](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/controllers/krate/publish.rs#L129-L139), create a new version owner action record with action = "publish", version_id = [version.id](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/controllers/krate/publish.rs#L129-L139), owner_id = [user.id](https://github.com/rust-lang/crates.io/blob/edcde838405303cc1afe5a0e6ff895a69fef96fb/src/controllers/krate/publish.rs#L42), owner_token = req.authentication_source() (and then extract the 
token value)

[...]

PR #1567, opened 2018-11-30
Mark API tokens as revoked

joshleeb:
With this PR, when a user deletes an API token, rather than deleting it from the database, it will instead be marked as `revoked`.

That means that this PR also adds a migration to add the column `revoked` to the `api_tokens` table.

Ref. #1548 (Task 1)


carols10cents:
Thanks! Looks great!!!

bors: r+

PR #1571, opened 2018-12-07
Enable Rust 2018 Edition

jtgeibel:
The migration is split into three commits.  The first applies edition fixes.  The second enables the edition and cleans up `extern crate` and `use` statements in many places.  The final commit applies any remaining edition idioms.


carols10cents:
Hooray! 2018 is dead, long live 2018!

bors: r+

locks:
👏 👏

PR #1576, opened 2018-12-14
Add a whatsdeployed.io badge to README

notriddle:
[What's Deployed](https://whatsdeployed.io/s-9IG)

Right now, this only includes prod, and it basically only provides the difference between production and master. Staging, oddly enough, is *behind* prod right now, so it doesn't actually work. If @ag_dubs updated staging to master, it would be possible to do a compare between all three (which is where whatsdeployed really shines). Compare https://staging-crates-io.herokuapp.com/api/v1/site_metadata with https://crates.io/api/v1/site_metadata

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # crates.io
 
 [![Build Status](https://travis-ci.org/rust-lang/crates.io.svg?branch=master)](https://travis-ci.org/rust-lang/crates.io)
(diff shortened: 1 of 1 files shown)


carols10cents:
I've been trying to check in on this PR when prod is in various states and this never looks broken, sooo... merged!

PR #1578, opened 2018-12-14
Run `cargo clean` on CI when the version of rustc changes

jtgeibel:
Recently I've been monitoring where our CI time is spent.  Currently,
the build and test stages take about 14 minutes without a cache and
1.7 to 3.6 minutes when a cache is present.  The variation in the
latter comes from the `cargo install diesel_cli` step which is
partially rebuilt any time a new compatible version of a dependency
is published.

For each channel, we cache about 1GB of build artifacts.  This results
in a total of 4 minutes that are spent in caching related activities.
However, extracting the cache on Travis appears to be O(n^3) because
once a new version of rustc is released the size of the `target/`
directory doubles but cache extraction take 8 times as long and total
cache related time nearly quadruples to 13.5 minutes.  This effectively
wipes out any gains of caching build artifacts for our codebase,
especially on the beta and nightly channels.

This commit adds a shell script that is run before any build steps on
CI.  The script works by recording the version of rustc in a file under
`target/`.  Upon the next build, if the cache was created by the same
version of rustc then no action is taken.  However, if the version of
rustc has changed then `cargo clean` is run to clear out the stale
cached files.

This commit also sets the cache timeout (for  building and uploading
the cache) to 6 minutes.  The default is 3 minutes and this portion is
currently taking about 2.6 minutes.  The rationale here is to provide
some additional headroom on the stable channel.  As dependencies are
updated, the cache will gradually grow until a new release causes the
cache to be cleared.  The increased timeout should allow for this
natural growth while also enforcing a reasonable limit so that caching
overhead doesn’t grow too large.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ cache:
   cargo: true
   directories:
     - $HOME/.npm
-  timeout: 600
+  # Set timeout to 6 minutes (double the default of 3 minutes)
+  timeout: 360
 
 env:
   global:
@@ -26,6 +27,7 @@ env:
     - PERCY_PROJECT=crates-io/crates.io
 
 install:
+  - script/cargo-clean-on-new-rustc-version.sh
   - cargo install --force diesel_cli --vers `cat .diesel_version` --no-default-features --features postgres && export PATH=$HOME/.cargo/bin:$PATH
 
 before_script:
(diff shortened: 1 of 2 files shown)


jtgeibel:
I plan to test this on CI by forcing a rebuild of this PR after a new nightly release.

jtgeibel:
I'm seeing the expected results on CI with the new nightly.  I've cleared the caches and I'm merging directly to master to kick off that build.

PR #1581, opened 2018-12-20
Make crawler policy more readable

IsaacWoods:
I found the wall of text about the crawler policy less than readable, so I added some paragraph breaks and `<code>` tags in an attempt to make it slightly more readable. This doesn't change any of the text of the policy.


sgrif:
Would love a before/after screenshot. :)

sgrif:
bors: r+

sgrif:
bors: retry

Issue #1503, opened 2018-09-28
Staging area for crate uploads


d-e-s-o:
I believe it would be great to have sort of a staging area for crates.io. Currently it appears that once a crate is uploaded there is no going back/changing of it in any way, modulo publishing a new version of course.
While that is fine conceptually, for initial releases or changes impacting the README or other other things in major ways, it would be great if there was a way to try things out first, have an initial look with the possibility of adjusting things.
PyPi has a staging area for this purpose, which seemingly is just a different instance (different URL) that has the once-uploaded-no-more-changes restriction removed.

I am thinking that such a feature would benefit a great deal of people. Have I just not found it?

d-e-s-o:
Case in point, I just uploaded a new version of my crate to https://crates.io/crates/gui. Turns out that the relative link that I embedded into README.md for some reason gets expanded to a wrong and non-existent file on github.com. That is absolutely not what I wanted and a possibility to inspect the result beforehand (I could see the result on github.com, but the link works correctly there!) would have given me the chance to fix the problem. Now I have to live with it until a new version is ready.

dwijnand:
Another use case this solves is making multi-crate releases atomic.  See https://github.com/rust-lang/cargo/pull/6510#issuecomment-450894487 where CI observed a world where failure_derive v0.1.5 and failure v0.1.4 were the latest versions available, which broke CI because failure_derive v0.1.5 required a change released in failure v0.1.5.

PR #1588, opened 2019-01-03
Move index updates off the web server

sgrif:
This fundamentally changes our workflow for publishing, yanking, and unyanking crates. Rather than synchronously updating the index when the request comes in (and potentially retrying multiple times since we have multiple web servers that can create a race condition), we instead queue the update to be run on another machine at some point in the future.

This will improve the resiliency of index updates -- specifically letting us avoid the case where the index has been updated, but something happened to the web server before the database transaction committed.

This setup assumes that all jobs *must* complete within a short timeframe, or something is seriously wrong. The only background jobs we have right now are index updates, which are extremely low volume. If a job fails, it most likely means that GitHub is down, or a bug has made it to production which is preventing publishing and/or yanking. For these reasons, this PR includes a monitor binary which will page whoever is on call with extremely low thresholds (defaults to paging if a job has been in the queue for 15 minutes, configurable by env var). The runner is meant to be run on a dedicated worker, while the monitor should be run by some cron-like tool on a regular interval (Heroku scheduler for us)

One side effect of this change is that `cargo publish` returning with a 0 exit status does not mean that the crate can immediately be used. This has always technically been true, since S3 and GitHub both can have delays before they update as well, but it's going to consistently be a bit longer with this PR. It should only be a few seconds the majority of the time, and no more than a minute in the worst case. One enhancement I'd like to make, which is not included in this PR, is a UI to show the status of a publish. I did not include it here, as this PR is already huge, and I do not think that feature is strictly required to land this. In the common case, it will take longer to navigate to that UI than it will take for the job to complete. This enhancement will also go nicely with work on staging publishes if we want to add those (see #1503). There are also some low hanging fruit we can tackle to lower the job's running time if we feel it's necessary.

As for the queue itself, I've chosen to implement one here based on PostgreSQL's row locking. There are a few reasons for this vs something like RabbitMQ or Faktory. The first is operational. We still have a very small team, and very limited ops bandwidth. If we can avoid introducing another piece to our stack, that is a win both in terms of the amount of work our existing team has to do, and making it easy to grow the team (by lowering the number of technologies one person has to learn). The second reason is that using an existing queue wouldn't actually reduce the amount of code required by that much. The majority of the code here is related to actually running jobs, not interacting with PostgreSQL or serialization. The only Rust libraries that exist for this are low level bindings to other queues, but the majority of the "job" infrastructure would still be needed.

The queue code is intended to eventually be extracted to a library. This portion of the code is the `background` module, and is why a lot of the code in that module is written a bit more generically than crates.io specifically needs. It's still a bit too coupled to crates.io to be extracted right now, though -- and I'd like to have it in the wild for a bit before extracting it. The `background_jobs` module is our code for interacting with this "library".


sgrif:
> I see background_worker in Procfile, but how (and how often) does monitor get run?

It gets run by any cron-like tool, which is Heroku scheduler for us. I figure we'll start at every 5 minutes, and tweak if needed. It is not configured in the repository.

I disagree that throttling is going to be a concern from cloning, but you make a good point on disk space (I'd assume that `/tmp` is cleared if we're out of disk space but I'm not sure whether that happens and don't really want to find out). Ultimately the clone per worker was something I think is likely to cause more delay than we'd like, and I was planning on tackling it shortly after this was merged, so I'll fix it now.

> Finally, the retry semantics allow for jobs to be executed out of order.

This is intentional, but that's a good point that yanking/unyanking quickly can cause the index to not match the database. I need to give this some more thought. I don't think I agree that we need to guarantee in-order execution for operations affecting `(crate_name, version)`, but I do agree that the index not matching the DB would be a problem.

sgrif:
This is blocked on #1607

sgrif:
@jtgeibel I've updated the yanking logic to handle database updates in the background job instead of on the web server.

I have not changed the clone into a temp dir behavior, and I don't think we need to in this PR. The concern about disk space is actually already handled. We're using [`TempDir`](https://docs.rs/tempdir/0.3.7/tempdir/struct.TempDir.html) for this, which will already clean up the directory when it's dropped.

I do want to change this to reuse the same checkout over multiple runs for other reasons, but I think there's enough tradeoffs/discussions for that to be its own PR. The concern about disk usage is resolved, so I don't think it needs to block this on its own.

So I believe this is good to go if you want to give it a last look. If it is good to merge, I'd appreciate a comment letting me know but holding off on merging for now. I consider this a pretty high risk PR so I want to make sure it's deployed by itself when I have plenty of time to do some final checks on staging and have overridden myself to be on call.

sgrif:
Going to do some final testing on staging and deploy this next week

sgrif:
Staging revealed a few issues which I've pushed fixes for. I'm confident that this actually works, I'm going to quickly test on staging with the full index instead of an empty one before merging.

sgrif:
Delay for publishes was too long with the full index, so I've set it up to maintain a local checkout. This eliminates the delay once the runner has booted, but there may still be a delay if a publish is done immediately after the server boots, as the web server now boots instantly rather than waiting for the clone to complete.

sgrif:
Going to do one more test in staging and then deploy this.

@bors: r+

Issue #1579, opened 2018-12-17
Support for alternate registries (RFC 2141)


ehuss:
I'm looking at what it will take to stabilize alternate registries in Cargo. It looks like crates.io has not been updated for it.  I wanted to see if this has been considered, and if not what should be done.

RFC 2141 says that crates.io should check for a whitelist of registries before allowing publishing a new crate (to allow other people to use the crates.io codebase).

The question is, should it support this whitelist functionality?  If so, I had some design questions:
- How should the whitelist be configured? The only configuration support I see is the .env file, would that be sufficient?
- I suspect that it would preferable that dependencies listed on the website link to the correct registry (or maybe not link at all?).  This seems like it would require a lot of intrusive changes.  Roughly I think it would need a new column in the database, and updates to the front end to generate correct links.  There's probably a ton of other issues (I'm not familiar with the crates.io code).

Just rejecting alt-registry dependencies looks like it would be very easy.  Adding the whitelist functionality seems quite a bit more.  I'm willing to help, but wanted to see what people think.

cc @withoutboats

jtgeibel:
> How should the whitelist be configured?  The only configuration support I see is the .env file, would that be sufficient?

Yeah, I would expect that any whitelisted registries would be provided via environment variables.

>  I suspect that it would preferable that dependencies listed on the website link to the correct registry (or maybe not link at all?). This seems like it would require a lot of intrusive changes.

In addition to handling links from an alternative registry to a dependency that lives on crates.io, there will be some additional complexity in verifying crate metadata upon publishing.  When publishing a crate, we verify the provided metadata which includes making sure all dependencies exist.  (In the future, we may want to add additional checks such as ensuring the version constraint can be satisfied by at least 1 version of that dependency.)  On crates.io itself, we query the database for this information but an alternative registry that whitelists crates.io (or any other registry) will need to clone, update, and query against the index directly.

This would be functionality that crates.io itself doesn't use in production.  I'm a bit cautious of the team taking on the support and maintenance of code that we have no need to directly run ourselves.  At the same time, I think enabling alternative registries may be enough of a motivating use case to justify supporting such alternative deployments directly in the codebase.

withoutboats:
> This would be functionality that crates.io itself doesn't use in production. I'm a bit cautious of the team taking on the support and maintenance of code that we have no need to directly run ourselves.

[...]

PR #1589, opened 2019-01-08
Reject publishing of crates that depend on an alternative registry

jtgeibel:
See also #1579 and rust-lang/crates-io-cargo-teams#21.


sgrif:
LGTM once the test failure is fixed

PR #1595, opened 2019-01-13
Add an extra check to `verify_tarball()`.

icefoxen:
Check that no tarball entries contain `..` to refer to a parent directory.
Otherwise you could upload a tarball that contains, say,
 `foo-0.1.0/../../.cargo/bin/cargo` and overwrite user files when
extracted. The `tar` crate's `Archive::unpack()` method will refuse to
files of this form, but other clients/tools may not be so discerning.

### DIFF
--- a/src/uploaders.rs
+++ b/src/uploaders.rs
@@ -242,6 +242,19 @@ fn verify_tarball(
             return Err(human("invalid tarball uploaded"));
         }
 
+        // Check that no entries contain `..` to refer to a parent directory.
+        // Otherwise you could upload a tarball that contains, say,
+        // `foo-0.1.0/../../.cargo/bin/cargo` and overwrite user files when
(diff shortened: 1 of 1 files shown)


icefoxen:
...forgive the mess.  If people are okay with this change in general I'll squash it all to one commit.

pietroalbini:
While you're at it, can you check if it contains a `Component::RootDir` (unix) or a `Component::Prefix` (windows)?

icefoxen:
Good idea, added.

icefoxen:
Hah, the `tar` crate won't let me construct an archive that references a parent dir. We need a less well-written tar file library!

Not sure how to make a unit test for it then, though.  I don't really feel like hand-building a tar file.

icefoxen:
That's up to you guys!

PR #1596, opened 2019-01-16
Rate limit the crate publish endpoint

sgrif:
This adds a restriction to the number of crates that can be published
from a single IP address in a single time period. This is done per IP
instead of per-user, since that's what we can do with nginx alone. We
may want to do additional rate limiting per token to force spammers to
register additional accounts, and not just change their IP.

This will use what's called the "leaky bucket" strategy for rate
limiting. Basically every user has a bucket of tokens they use to
publish crates. They start with 10 tokens available. Every time they hit
this endpoint, they use one of the tokens. We give them a new token each
minute.

What this means is that you can upload 1 crate per minute, but we allow
you to upload up to 10 in a short period of time before we start
enforcing the rate limit.

When someone does hit the rate limit, they will receive a 429 response.
We could also allow it to instead just slow down the requests, refusing
to process them until a token is available (queueing a max of 10 requests).

This reserves 10 megabyte of memory for the IP table, which means we can
hold about 80000 IPs. When the table is full, it will attempt to drop
the oldest record, and if that doesn't give enough space, it'll give a
503. Keep in mind this is all in memory, not shared between our servers.
This means that it is possible (but not guaranteed) that someone can
upload 20 crates, and then send 2 requests per minute.

On Heroku and any system that involves proxies, the remote_addr of the
request becomes useless, as it will be the proxy forwarding your request
rather than the client itself. Most proxies will set or append an
`X-Forwarded-For` header, putting whatever remote_addr it received at
the end of the the header. So if there are 3 proxies, the header will
look like

real_ip, proxy1, proxy2

However, if we're not careful this makes us vulnerable to IP spoofing. A
lot of implementations just look at the first value in the header. All
the proxies just append the header though (they don't know if they're
getting traffic from the origin or a proxy), so you end up with

spoofed_ip, real_ip, proxy1, proxy2

nginx, Rails, and many other implementations avoid this by requiring an
allowlist of trusted IPs. With this configuration, the realip only kicks
in if remote_addr is a trusted proxy, and then it will replace it with
the last non-trusted IP from the header.

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,5 +1,5 @@
 https://github.com/Starkast/heroku-buildpack-cmake#a243c67
 https://github.com/emk/heroku-buildpack-rust#578d630
 https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/emberjs.tgz
-https://github.com/travis-ci/nginx-buildpack.git#2fbde35
+https://github.com/heroku/heroku-buildpack-nginx.git#fbc49cd
 https://github.com/sgrif/heroku-buildpack-diesel#f605edd
--- a/app/templates/policies.hbs
+++ b/app/templates/policies.hbs
@@ -25,6 +25,12 @@ them. If necessary, the team may reach out to inactive maintainers and help
 mediate the process of ownership transfer.
 </p>
 
+<p>
+Using an automated tool to claim ownership of a large number of package names
+is not permitted. We reserve the right to block traffic or revoke ownership
+of any package we determine to have been claimed by an automated tool.
+</p>
+
 <h2 id='removal'><a href='#removal'>Removal</a></h2>
 
 <p>
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -56,6 +56,9 @@ http {
 		proxy_set_header Host $http_host;
 		proxy_set_header X-Real-Ip $remote_addr;
 		proxy_redirect off;
+		if ($http_x_forwarded_proto != 'https') {
+			rewrite ^ https://$host$request_uri? permanent;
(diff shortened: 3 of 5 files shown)


sgrif:
I've given this some local testing to make sure everything is working properly (as best as I'm able to locally). I've confirmed that it is in fact setting the `X-Real-Ip` header to what it's expected to be, that the publish endpoint is rate limited, that all other endpoints are not rate limited, and that the changes to where the various proxy headers still works as before.

sgrif:
Note on above: I agree, but the wording is tricky and IMO should not block landing this

ashleygwilliams:
i always prefer to use PRs as a forcing function for docs- very happy to help wordsmith here if you'd like (or in PMs)

jtgeibel:
I've tested this in my staging area.  The IP address shows correctly in the logs and after making 10 requests to the publish endpoint I started getting responses with a 429 status.

This was on the heroku-16 stack.  The build pack pulls in nginx-1.9.5 on that stack.  I'm not sure what stack we are on in production, but it looks like Heroku updated the buildpack to nginx-1.14.1 on heroku-18, which is nice.

sgrif:
We talked a while back about policies, but it sorta stalled out. I've pushed up what I think would make the most sense here. I've put it under "Package Ownership" since this doesn't really have anything to do with crawlers. The wording "to claim ownership of a large number of packages" is deliberate, as it's the only wording I could come up with that separates the cases we care about from "we use a CI tool or a script to publish our suite of 10 crates" which we don't want to prevent.

jtgeibel:
I'm r+ on the policy language, and this PR overall.

PR #1601, opened 2019-01-22
Align crate icon with crate name in heading

p-jackson:
The crate heading text (with name and version) are being aligned by baseline using flex, however the icon doesn't line up because it doesn't have baseline information.

Manually pushing the icon down so aligns "correctly" with the crate name.

Before 🤢 🤮
image

After 🌹 🦄 
image

### DIFF
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -26,6 +26,12 @@
             @include flex-direction(column);
         }
     }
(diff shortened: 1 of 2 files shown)


p-jackson:
Oops, and the template change to 😂

jtgeibel:
Looks good to me.

@bors r+

PR #1604, opened 2019-01-24
Add migration and model for version_owner_actions table

joshleeb:
Add a migration to create the `version_owner_actions` table, as well as a `VersionOwnerAction` struct in `models/actions.rs`.

Ref. #1548 (Task 2)


migrations/2019-01-24-101412_create_version_owner_actions/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE version_owner_actions (
+    id SERIAL PRIMARY KEY,
+    version_id INTEGER,
+    owner_id INTEGER,
+    owner_token_id INTEGER,
```
joshleeb: I've left `REFERENCES api_token(id)` and similar off from `up.sql` as it was causing an error with applying `schema.patch` to `schema.rs`. 

Wasn't sure if that is something that is expected (and the foreign keys are defined in the model) of if there was something I was doing wrong.

sgrif: Nothing you're doing wrong. I'm surprised `patch` wasn't able to resolve this on its own. We probably need to pass it some more flags... This change will allow the patch to apply with your new foreign key.

```patch
diff --git a/src/schema.patch b/src/schema.patch
index 3ccdaee..7331e23 100644
--- a/src/schema.patch
+++ b/src/schema.patch
@@ -66,7 +66,7 @@ index df884e4..18e08cd 100644
  joinable!(version_authors -> users (user_id));
- joinable!(versions -> crates (crate_id));
+ joinable!(version_owner_actions -> api_tokens (owner_token_id));
  
 @@ -913,12 +935,13 @@ allow_tables_to_appear_in_same_query!(
      dependencies,
```

joshleeb: @sgrif seemed to do the trick, thanks!

markcatley:
I've used your changes, made some improvements and moved things a little further forward. I'd love your feedback. It is PR #1700.

Issue #1265, opened 2018-02-22
Tracking issue for updating dependencies


jtgeibel:
## API Changes

These require code changes to address API changes.

* [x] `tar` - a80193e
* [x] `openssl` part 1 - b4fc94f
* [x] `openssl` part 2 - 8036a78
* [x] `hyper` - #1289
* [x] `flate2` - bd0cd6d
* [x] `envlogger` - ac9d73d
* [x] `hex` - dccee42
* [x] `lettre` - 822735e
* [x] `oauth2` - #1842
* [x] `jemalloc` and `jemalloc-ctl` - to 0.3.2 - #1953 and #1956
* [x] `cookie` - 0.11 to 0.12, will also need to upstream a change to `conduit-cookie` #2208
* [x] `dotenv` - 0.11 to 0.15 - #1917
* [x] `hex` - 0.3 to 0.4 #2256

[oauth2-2.0-alpha]: https://github.com/jtgeibel/crates.io/commit/7de5a023cb211138b09485320bedc3e4f3a17fd2

## Version bump

These should just be a version bump in `Cargo.toml`.

* [ ] `url` - 1.7 to 2.1 (needs to wait for newer oauth2 3.0 alpha)
* [x] `comrak` - 0.4 to 0.7 - #2513
* [x] `dotenv` - 0.10 to 0.11 - #1391
* [x] `itertools` - 0.6 to 0.7
* [x] `base64` - 0.6 to 0.9
* [x] `log` - 0.3 to 0.4
* [x] `comrak` - 0.2 to 0.4 #1664
* [x] `ammonia` - 1.0 to 2.0 #1664
* [x] `git2` - 0.6 to 0.8 #1663
* [x] `env_logger` - 0.5 to 0.7 - #2236
* [x] `docopt` - 0.8 to 1.1 #1812
* [x] `toml` - 0.4 to 0.5 #2430
* [x] `git2` - 0.8 to 0.13 #2467
* [x] `ammonia` - 2.0 to 3.0 - #1876
* [x] `rand` - 0.6 to 0.7 - #2512
* [x] `base64` - 0.9 to 0.11 (also bump in s3) #2208
* [x] `parking_lot` - 0.7 to 0.10 - #2223
* [x] `handlebars` - 2 to 3 #2259

### `conduit-git-http-backend`

We could switch to `libflate` however it is no longer pulled in by `reqwest` by default and we've consolidated on a single version of `flate2` so this isn't a priority.

[...]

PR #1605, opened 2019-01-24
Update itertools to 0.7.0 and log to 0.4

joshleeb:
This PR updates dependency crates `itertools` to 0.7.0 and `log` to 0.4.

Ref. #1265


sgrif:
It looks like you accidentally updated a lot more than just those two dependencies, which has caused test failures. Did you maybe run `cargo update` on accident?

sgrif:
You might just want to make sure you've rebased on the latest master, then.

sgrif:
Scratch that, this isn't your fault. The author of `ring` has yanked all versions of their crate including the one that we depend on. Any attempts to change `Cargo.toml` in any way will now break our build. I'm looking into this now.

sgrif:
This is blocked on #1607

PR #1609, opened 2019-01-28
Update to the latest version of cookie

sgrif:
This is required for us to continue to make any changes to Cargo.toml,
since all versions of ring except the most recent have been yanked. We
only depend on ring through cookie. Moving from cookie 0.9 to 0.11
doesn't appear to require any code changes in either our code or
conduit-cookie.

We can't use `[patch]` here, since it will not force Cargo to resolve to a
yanked version of a crate (it'll still just try to resolve
conduit-cookie 0.8.4 off the index which inevitably depends on ring 0.11
so it'll fail and go back to the last version which used openssl)


alexcrichton:
Done!

PR #1611, opened 2019-01-28
Use the `--locked-schema` flag in Diesel 1.4

sgrif:
This flag tells Diesel CLI to error if a command would change
`src/schema.rs`, rather than updating it. By running this in CI and
production builds, we ensure that no PR will pass which changes schema
but doesn't modify this file (which has happened in the past)


jtgeibel:
I'm seeing this error out on Heroku on staging, but running locally succeeds (as no changes to src/schema.rs are needed).

```
2019-02-07T08:47:12.503215+00:00 heroku[web.1]: Starting process with command `bin/diesel migration run --locked-schema && bin/start-nginx ./target/release/server`
2019-02-07T08:47:15.387093+00:00 app[web.1]: Command would result in changes to src/schema.rs. Rerun the command locally, and commit the changes.
```

jtgeibel:
I'm thinking the issue is that we run `bin/diesel migration run --locked-schema` at application boot, but the source code is not available at that time.  The source is only available at build time.  So at the point we run the migration command, no `src/schema.rs` exists and it will thus *always* differ.

sgrif:
Let's go ahead and revert for now then

jtgeibel:
I'm working on a PR to remove this from just from the Procfile, unless you think we should revert the whole PR.

sgrif:
No that's fine

PR #1614, opened 2019-02-01
Split the two uses of version::deprecated::show into separate fns

carols10cents:
Discovered that /crates/:crate_id/:version is used in the show_version
test RequestHelper function, so it probably shouldn't be marked as
deprecated even though the frontend isn't using it.

Takes care of the FIXME comment.


jtgeibel:
I remember writing that FIXME, glad to see this resolved!  Great catch that this is used in the `yank` test.

@bors r+

PR #1615, opened 2019-02-02
Temporarily deny clippy::nursery and fix warnings

jtgeibel:
The warnings were ultimately from `clippy::needless_borrow` and
`clippy::redundant_clone`.


jtgeibel:
I decided to not leave the `#![deny(clippy::nursery)]` in the source code for now, but these lints are nice and didn't produce any false positives here.

Issue #1620, opened 2019-02-09
Require a valid email address to publish crates


carols10cents:
Just realized we didn't have a crates.io issue for this; the overall plan for implementing this is in https://github.com/rust-lang/crates-io-cargo-teams/issues/8.

Mirroring some of the content from that issue and from [users.rust-lang.org](https://users.rust-lang.org/t/a-verified-email-address-will-be-required-to-publish-to-crates-io-starting-on-2019-02-28/22425) here:

## Rationale

To comply with DMCA, we need a guaranteed way to contact publishers of content on crates.io.

## Implementation details

-    The verified email address is not associated at all to the email address that may optionally appear in the authors metadata in the crate’s Cargo.toml.
-    Your verified email address won’t be displayed anywhere publicly (unless you choose to place it in your Cargo.toml as well).
-    This email will only be used to contact you for crates.io operational needs and will never be shared with any third parties.
-    Only the crate owner running cargo publish will need to have their email address verified.
-    The email address will be saved with the particular version being published at publish time, so that if an owner is removed from the crate or removes their email address, it’s still available with the published content.

## Implementation plan

- [x] ~~Start publicizing this plan as soon as we agree on it~~ [Done](https://users.rust-lang.org/t/a-verified-email-address-will-be-required-to-publish-to-crates-io-starting-on-2019-02-28/22425)
- [x] ~~Add general warning display capability to Cargo and get it into nightly in this release cycle~~ [Done](https://github.com/rust-lang/cargo/pull/6303)
- [x] ~~Implement the warning in crates.io~~ [Done](https://github.com/rust-lang/crates.io/pull/1565)
- Warning capability would go into beta with Rust 1.32.0 on 2018-12-06
- Warning capability would be stable with Rust 1.32.0 on 2019-01-17
- We would warn for one release cycle
- Coinciding with the release of 1.33.0 on 2019-02-28, we would disallow publishing crates without a valid email address.
- [x] Start recording the verified email addresses of version publishers, if the user has a verified email
- [x] Implement the hard error in crates.io, possibly with date checks so we don't have to remember to merge+deploy code on a particular day

PR #1621, opened 2019-02-09
Record verified email, if present, of the publisher in the version record

carols10cents:
Connects to #1620. We can start recording emails if we have them, even though we're not requiring verified emails yet.

This builds on top of #1561.


sgrif:
Since we virtually never want this column outside of storing it for auditing purposes, does it make sense to store it in another table instead of as a column on versions?

carols10cents:
Great idea! That would certainly get rid of the complexity from working around https://github.com/diesel-rs/diesel/issues/1979.

Do the changes from #1561 look ok? Would you rather merge that one and then this one, or only merge this one?

sgrif:
#1561 looks fine overall. I'd probably opt to do a second query to load the publishers, but we can always make those changes later. I have no strong feelings on which we merge first, but I'd probably like to deploy at least once before merging either since we have a lot of code already in master for the next deploy.

carols10cents:
Ok, if both this and #1561 look fine, I'd recommend merging this one, which I think we're ready to do because you deployed today. I'll let you approve though, going to close #1561.

jtgeibel:
This looks good to me.  I've also tested in staging and confirmed that the publishing information is provided in the API and that the verified email address is recorded in the database (but not exposed publicly).

Thanks @Andy-Bell and @carols10cents for your work on this.

@bors r+

PR #1629, opened 2019-02-22
Require a verified email after 2019-03-01 00:00 UTC

carols10cents:
Because that's at least on 2019-02-28 (the date we announced) in all timezones,
and should definitely be after the release scheduled for that day.

As per the plan at https://github.com/rust-lang/crates-io-cargo-teams/issues/8

Needs to be merged and deployed anytime before 2019-03-01 00:00 UTC to be on time with our announced date.

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -3,6 +3,8 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
+use chrono::offset::TimeZone;
+use chrono::{DateTime, Utc};
 use hex::ToHex;
 
 use crate::git;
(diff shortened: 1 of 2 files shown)


carols10cents:
I opened #1630 for reference and to have it ready for when the tests start failing after 2019-03-01 00:00:00 UTC.

src/controllers/krate/publish.rs
```diff
@@ -267,3 +268,102 @@ fn parse_new_headers(req: &mut dyn Request) -> CargoResult<(EncodableCrateUpload
     let user = req.user()?;
     Ok((new, user.clone()))
 }
+
+lazy_static! {
+    static ref VERIFIED_EMAIL_REQUIRED_DATE: DateTime<Utc> =
+        { Utc.ymd(2019, 3, 1).and_hms(0, 0, 0) };
```
sgrif: Since this is only used in one place, and the functions being called have basically no overhead, do you think it's worth inlining this to not have to add lazy-static just to remove it next week?

carols10cents: Can do!

carols10cents:
Done!

jtgeibel:
LGTM!

@bors r+

PR #1630, opened 2019-02-22
DO NOT MERGE until after 2019-03-01; Remove date checks for verified email requirement

carols10cents:
This removes the date checks added in #1629 so that an email address is always required. The integration tests will start failing after 2019-03-01 00:00:00 UTC; this is the cure.


carols10cents:
Merging this to get tests passing again; sgrif already approved.

@bors r+

PR #1618, opened 2019-02-07
Switch to `conduit-hyper` in place of `civet`

jtgeibel:
`conduit-hyper` now handles normalizing paths, such that `//api/v1`
becomes `/api/v1` and doesn't break our router.

PR #1633, opened 2019-02-22
Revert "Auto merge of #1618 - jtgeibel:conduit-hyper-round2, r=sgrif"

sgrif:
This reverts commit 69368a6efe228f3a7cc91a3b8783ff03ec9e9d92, reversing
changes made to 6fa8ad32cc3470b3d320d4620ba305c8d220ff2d.

When deploying this change, we saw response times nearly triple, and our
CPU usage went through the roof. Additionally, this did not appear to
fix the memory issues we were hoping to address by removing civet.

/cc @jtgeibel

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,12 +153,12 @@ dependencies = [
  "base64 0.9.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "cargo-registry-s3 0.2.0",
  "chrono 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "civet 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "comrak 0.2.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "conduit 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


sgrif:
screen shot 2019-02-22 at 1 18 28 pm
screen shot 2019-02-22 at 1 18 31 pm

sgrif:
Note: I do think there's a possibility that we just had bad luck with noisy neighbors, and I do want to try deploying this again one more time when I can be around to keep an eye on it, but I'd like to get it off master so we don't block things until then

jtgeibel:
> I do think there's a possibility that we just had bad luck with noisy neighbors

That's definitely a possibility.  I had done some load/concurrently testing of some API endpoints and didn't see any noticeable difference in requests per second before and after the change.  However, I was only looking at RPS, not response time for individual requests and I'm now wondering if I was just effectively load testing the database.  I'll also do some load testing of non-API endpoints to see if serving the static ember HTML is impacted.

I agree, lets revert for now to keep master deployable.

@bors r+

sgrif:
:+1: I'm also fine just trying again as is later this week, just need to do it when I have some time to be around

PR #1634, opened 2019-02-23
Remove references to the `categories::path` column in Rust

sgrif:
This column is never used in Rust code, and we're doing a lot of
funkiness to work around its existence. Even if we wanted to use this
column in Rust code, we wouldn't be able to since the diesel_ltree crate
is using Diesel 0.16. There's a more recent version that uses Diesel
1.0, but I don't see us using this column in the future more than we are
today.

The column is still used in raw SQL, so we haven't removed it from the
actual database schema.


jtgeibel:
I'm really glad to see some older duplicate dependencies going away!

@bors r+

PR #1637, opened 2019-02-24
docker: use --locked-schema for diesel migrations

### LINKED ISSUE #1631: `docker-compose up` fails to run migrations
When I do a fresh clone of the crates.io repo, and then run `docker-compose up`, the backend fails to come up and instead gets stuck in a loop:

```
...
backend_1   | Read-only file system (os error 30)
backend_1   | Migrations failed, retrying in 5 seconds...
backend_1   | Read-only file system (os error 30)
backend_1   | Migrations failed, retrying in 5 seconds...
backend_1   | Read-only file system (os error 30)
backend_1   | Migrations failed, retrying in 5 seconds...
...
```

For new contributors, this should not happen and should instead come up without errors to ease the path to start working on the project.

---

From what I've been able to tell so far, `diesel migration run` is succeeding in the sense that it is running all of the migrations. (It lists the last migration that it is supposed to run, and I manually verified that that migration succeeded.)

Where it fails is that `diesel migration run` is trying to create `src/schema.rs.orig` on the source volume that is mounted as read-only, so when that fails it returns a non-zero exit code.

Two potential solutions I see:

1. Mount the source directory as read-write instead of read-only. This would be a very easy fix, and allow `diesel migration run` to write that file, and that file may be valuable (I'm not sure, I haven't worked with `diesel` before.)
2. Determine what we need to do to make `diesel migration run` not write that file. Do we need to re-sync `src/schema.rs`? Is there an option in `diesel` that prevents creating that file?

bryanburgers:
When using `docker-compose up` to bring up everything for development,
we mount the source directory as read-only so that changes can only
happen on the host machine and not because of something that the code
running in docker does.

This read-only mount was causing problems because `patch`, called by
`diesel migration run`, would try to write to the filesystem and fail.

When bringing up a docker environment for the first time from a fresh
clone, we expect that the schema.rs file is already pristine, and that
we just need to run migrations. For that reason, use `diesel migration
run`'s `--locked-schema` option to just run migrations.

This change makes it so that a new contributor can clone the repo, run
`docker-compose up`, and have an almost complete environment to poke at
and start exploring.

Fixes #1631

### DIFF
--- a/docker_entrypoint.sh
+++ b/docker_entrypoint.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # If the backend is started before postgres is ready, the migrations will fail
-until diesel migration run; do
+until diesel migration run --locked-schema; do
   echo "Migrations failed, retrying in 5 seconds..."
   sleep 5
 done


bryanburgers:
This merge request and #1632 which I also opened are similar in scope – getting the docker-compose stuff into a working and usable state.

Let me know if they should be combined into a single merge request.

jtgeibel:
LGTM, thanks!

@bors r+

PR #1642, opened 2019-02-28
Exclude yanked crates on user/team pages

### LINKED ISSUE #958: Yanked crates shouldn't be displayed on user/team pages

bryanburgers:
Add a URL parameter to the `/api/v1/crates` endpoint that specifies whether or not the endpoint should return crates that have been fully yanked - i.e., where every version has been yanked.

If not specified, this defaults to `true` because we have typically always returned all matching crates, whether or not their versions have been yanked.

This makes it possible to display only non-yanked crates at various places in the UI without changing Rust code.

Use this from the user and team pages to exclude yanked crates from them.

Closes #958

### DIFF
--- a/app/routes/user.js
+++ b/app/routes/user.js
@@ -15,6 +15,7 @@ export default Route.extend({
         return this.store.queryRecord('user', { user_id }).then(
             user => {
                 params.user_id = user.get('id');
+                params.include_yanked = 'n';
                 return RSVP.hash({
                     crates: this.store.query('crate', params),
                     user,
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -1,5 +1,6 @@
 //! Endpoint for searching and discovery functionality
 
+use diesel::dsl::*;
 use diesel::sql_types::{NotNull, Nullable};
 use diesel_full_text_search::*;
 
@@ -33,7 +34,6 @@ use crate::models::krate::{canon_crate_name, ALL_COLUMNS};
 /// function out to cover the different use cases, and create unit tests
 /// for them.
 pub fn search(req: &mut dyn Request) -> CargoResult<Response> {
-    use diesel::dsl::sql;
     use diesel::sql_types::{Bool, Text};
 
     let conn = req.db_conn()?;
@@ -44,6 +44,10 @@ pub fn search(req: &mut dyn Request) -> CargoResult<Response> {
         .map(|s| &**s)
         .unwrap_or("recent-downloads");
     let mut has_filter = false;
+    let include_yanked = params
+        .get("include_yanked")
+        .map(|s| s == "yes")
+        .unwrap_or(true);
 
     let selection = (
         ALL_COLUMNS,
@@ -152,6 +156,15 @@ pub fn search(req: &mut dyn Request) -> CargoResult<Response> {
                     .filter(follows::user_id.eq(req.user()?.id)),
             ),
         );
+    }
+
+    if !include_yanked {
+        has_filter = true;
+        query = query.filter(exists(
+            versions::table
+                .filter(versions::crate_id.eq(crates::id))
+                .filter(versions::yanked.eq(false)),
+        ));
     }
 
     if sort == "downloads" {
--- a/app/routes/team.js
+++ b/app/routes/team.js
@@ -16,6 +16,7 @@ export default Route.extend({
         return this.store.queryRecord('team', { team_id }).then(
             team => {
                 params.team_id = team.get('id');
+                params.include_yanked = 'n';
                 return RSVP.hash({
                     crates: this.store.query('crate', params),
                     team,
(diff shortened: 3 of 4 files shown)


app/routes/user.js
```diff
@@ -15,6 +15,7 @@ export default Route.extend({
         return this.store.queryRecord('user', { user_id }).then(
             user => {
                 params.user_id = user.get('id');
+                params.include_yanked = 'n';
```
sgrif: Should we just make this the default and not specify it in the URL here?

src/controllers/krate/search.rs
```diff
@@ -208,3 +221,11 @@ pub fn search(req: &mut dyn Request) -> CargoResult<Response> {
         meta: Meta { total },
     }))
 }
+
+/// Takes a string that might represent a boolean in a URL and converts it into a boolean.
```
sgrif: Do we need to generalize this much? Can we just assume a specific value?

bryanburgers: No, we really don't. I dropped the function and changed the option.map to just look for the string "yes".

app/routes/user.js
bryanburgers: Here was my thinking. If it's not the direction we want to go, I can make changes.

There are situations where we ***want*** to show yanked crates: https://github.com/rust-lang/crates.io/issues/958#issuecomment-351421886, https://github.com/rust-lang/crates.io/issues/145#issuecomment-281492212.

Given the following two URLs:

* `/api/v1/crates?search=foo`
* `/api/v1/crates?user_id=2576`

One option would be to add conditional logic to the controller like "if `user_id` is set, then don't show yanked crates; if `search` is set, then show yanked crates".

I shied away from that option because I think it would be confusing. A contributor focused on the front end would have to keep in their head which rules apply to which parameters. 

Given the following two URLs:

* `/api/v1/crates?search=foo&include_yanked=yes`
* `/api/v1/crates?user_id=2576&include_yanked=no`

It's much less surprising what will be returned from these endpoints.

However, I'm not sure we want to audit and edit every location where the front end calls the `/api/v1/crates` endpoint – and we can't audit third-party crawlers. Beacuse `/api/v1/crates` has historically defaulted to showing yanked crates, I opted to only add `&include_yanked=no` to the specific front end code where we don't want to show yanked crates.

* `/api/v1/crates?search=foo`
* `/api/v1/crates?user_id=2576&include_yanked=no`

This makes the URLs for the API endpoint a little bit more ugly for these pages. However, this largely only shows up in developer tools. This change doesn't add a query parameter to the URL for a user/team page, e.g. a user page continues to be `/users/sgrif`, not ~`/users/sgrif?include_yanked=no`~.

Am I making any sense? Hopefully I didn't answer a question you weren't asking!

bryanburgers:
*Note:* CI failures are because it's after 2019-02-28 and #1630 hasn't landed yet.

sgrif:
Just noting for the future -- this causes a roughly 30% performance regression on search, which brings us up to 265ms in the worst case (search for 's' ordered by relevance). I don't think it needs to be addressed in this PR, but I think we're pushing the limits on what's acceptable performance-wise.

We probably want to start thinking about introducing a materialized view that includes this yanked calculation and has already done the join on recent_crate_downloads, which can bring us to 160ms in the worst case -- or we may just want to stop doing the like check on single letter searches, since this will only continue to deteriorate in the future.

@bors: r+

bryanburgers:
Yikes. 30% is bad. I'm concerned about that. But we only pay that 30% when it's used, right? And right now, it won't get used on the search page?

sgrif:
Ah you're right, I missed that this only affects the user page by default

sgrif:
Looks like GH is having an outage that's breaking bors right now

sgrif:
Looks like beta broke the build, I'll fix it shortly

bryanburgers:
Boo! I cab look into the failure in a week or two.

sgrif:
I think you might just need to rebase

bryanburgers:
@sgrif Rebased and fixed.

PR #1645, opened 2019-03-05
Use a consistent format for numbers in dl graph

### LINKED ISSUE #1628: inconsistent number format in Stats Overview (point `.` with different meaning)
When I look at the download stats pop-over, the number format is incosistent.

- The 7-day average has a decimal point **1855.86** (one thouasand fifty five ... and a fraction)
- The daily downloads separates thousand by a point **2.140** (two thousand one houndred fourty)

The following example is from [here](https://crates.io/crates/pretty-assertions)

grafik

My HTTP-Request header is like this, if it plays any role:
```
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:65.0) Gecko/20100101 Firefox/65.0
Accept-Language: de,en-GB;q=0.7,en;q=0.3
```

bryanburgers:
When formatting download numbers in the download graph, use a consistent numbering format for both raw numbers and the rolling 7-day average.

There are cases where the user's local can make inconsistent formatting. Specifically, we use `Number.prototype.toFixed` to format the rolling 7-day average, which always results in a number format that uses a decimal point (e.g. "1234.56"). However, the raw download numbers end up using the locale, which in some places can end up formatting the number with a decimal as the thousands separator ("1.234").

To fix this, use google.visualization.NumberFormat for *both* values, so that they are always consistent.

Fixes #1628 

Note that I couldn't fully test this, because even after changing my browser's language to `de`, I would still see English number formatting. I did test it by [forcing the Google Visualization chart locale](https://developers.google.com/chart/interactive/docs/basic_load_libs#loadwithlocale), but I'm still not 100% sure if the numbers will show in a German format for a German viewer.

I am, however, sure that they will always show in a *consistent* format.

### DIFF
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -78,10 +78,20 @@ export default Component.extend({
 
         let myData = window.google.visualization.arrayToDataTable(data);
 
-        let fmt = new window.google.visualization.DateFormat({
+        let dateFmt = new window.google.visualization.DateFormat({
             pattern: 'LLL d, yyyy',
         });
-        fmt.format(myData, 0);
+        dateFmt.format(myData, 0);
+
+        // Create a formatter to use for daily download numbers
+        let numberFormatWhole = new window.google.visualization.NumberFormat({
+            pattern: '#,##0',
+        });
+
+        // Create a formatter to use for 7-day average numbers
+        let numberFormatDecimal = new window.google.visualization.NumberFormat({
+            pattern: '#,##0.0',
+        });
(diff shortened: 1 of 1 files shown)


bryanburgers:
screen shot 2019-03-04 at 9 32 31 pm
Before (de)

---

screen shot 2019-03-04 at 9 29 31 pm
After (de)

---

screen shot 2019-03-04 at 9 30 14 pm
After (en)

PR #1647, opened 2019-03-07
Switch to a RustConfig file to set the rustc version used by the buildpack

jtgeibel:
Using the `rust-toolchain` file is causing our CI to use the pinned
version for all three builds.  Additionally, the old version of our
buildpack doesn't support `rust-toolchain` yet, so I think this was
only working because the old compiler was cached.


carols10cents:
🤦‍♀️

Confirmed that the travis jobs are now being run with the version we intended; then deployed to staging and verified the buildback is using 1.32. 

@bors r+

PR #1649, opened 2019-03-07
Allow profiling jemalloc

sgrif:
This builds jemalloc with profiling enabled, and preserves debug symbols
in the release binary (note: `debug = true` does not disable
optimizations). This commit by itself doesn't actually perform any
profiling, but it allows it to be controlled by an environment variable.

At this point I cannot get any useful information about the memory issue
we're seeing in production locally. We also can't reproduce it in a
staging environment. There's *something* about our production traffic
patterns that we're missing, so the next step is to start looking at
dumps of production memory usage.

My intention is to start enabling sampling based profiling in
production, and copy some of the dumps off the server to examine
locally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2018"
 
 [profile.release]
 opt-level = 2
+debug = true
 
 [lib]
 name = "cargo_registry"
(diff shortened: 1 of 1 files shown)


jtgeibel:
These changes seem fine to me.

@bors r+

PR #1651, opened 2019-03-08
Bump tar dependency to the latest

alex:


alex:
This ensures that crates.io isn't vulnerable to https://rustsec.org/advisories/RUSTSEC-2018-0002.html

I think it'd be a good idea to use `cargo-audit` in CI.

jtgeibel:
Just to clarify, we manually deny any crate uploads containing a symlink or hardlink, and should not be vulnerable to this on the sever backend. This also protects older clients using older versions of the crate from downloading a bad crate.

alex:
Sorry, I should have been clear -- I didn't think crates.io was actually vulnerable to this, but I'm a big believer in security-through-resillience :-) I assume no one is in favor of keeping the vulnerable dependency around :P

PR #1653, opened 2019-03-09
Rename clippy lint to `cognitive_complexity`

jtgeibel:
This renamed lint started warning on nightly.  I've confirmed that the
new name works at least as far back as 1.32.0, our currently pinned
version in production.

### DIFF
--- a/src/tests/category.rs
+++ b/src/tests/category.rs
@@ -74,7 +74,7 @@ fn show() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


jtgeibel:
Well it looks like I was forgetting to include `--all-targets --all-features --all` when invoking clippy locally.  This will have to wait until 1.34 is stable, so I'm closing this for now.

PR #1655, opened 2019-03-10
Add a manual trigger to clean the cache on CI

jtgeibel:
When we bump a low-level dependency, such as `libc`, a large number of
crates are rebuilt and this can grow our cache to the point where
caching hits a timeout (and is not persisted).

By changing the value of `manual_stamp` in the script, the cache can be
cleared without waiting for the next compiler release.


jtgeibel:
Merging to add the stamp file to `auto` and `master` caches.

@bors r+

PR #1658, opened 2019-03-11
Remove the worker dyno

sgrif:
We're paying for a dyno that spends 90% of its time sleeping. A
repeating task like this is better handled by a cron-like tool (heroku
scheduler in our case), especially for something non-critical like
updating download counts.


jtgeibel:
This is great!  I rarely run this worker in my staging area, but can now add a scheduled task for this without running up dyno hours.

@bors r+

PR #1661, opened 2019-03-11
Move the background jobs to another crate

sgrif:
I wrote this code intending to extract it, and I'm planning to release
it after a few more changes which aren't relevant to crates.io. The code
right now is largely the same as what was here before, the main
difference being the use of a different error type.

The error handling will probably churn some more in the near future as I
figure out what I want the error handling story to look like in that
crate


jtgeibel:
I merged to resolve the merge conflicts so that I could r+ this, but it looks like the lockfile isn't storing the git SHA and the new build is picking up the latest changes on `swirl` master.

sgrif:
Locked to a single commit.

@bors r=jtgeibel

PR #1663, opened 2019-03-12
Bump `git2` and remove dependency on `cmake`

jtgeibel:
The `git2` crate has dropped its dependnecy on `cmake` and we can as
well.

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,4 +1,3 @@
-https://github.com/Starkast/heroku-buildpack-cmake#a243c67
 https://github.com/emk/heroku-buildpack-rust#578d630
 https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/emberjs.tgz
 https://github.com/heroku/heroku-buildpack-nginx.git#fbc49cd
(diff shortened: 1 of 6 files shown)


sgrif:
Excellent, cmake is the most common cause of build failures for me. I'm actually wondering if we should just drop git2 entirely and subshell out to git now that this is off the web server. The savings from being able to work with a shallow clone likely outweigh the cost of subshelling.

jtgeibel:
Rebased to resolve merge conflicts.

@bors r+

jtgeibel:
> The savings from being able to work with a shallow clone likely outweigh the cost of subshelling.

I didn't realize that `libgit2` doesn't support shallow clones.  While it looks like newer versions of `git` support committing to a shallow clone, I don't think pushing from a shallow clone is supported yet: "It is unclear how "push into/from a shallow repo" should behave." ([Documentation/technical/shallow.txt](https://github.com/git/git/blob/e902e9bcae2010bc42648c80ab6adc6c5a16a4a5/Documentation/technical/shallow.txt#L46))

PR #1665, opened 2019-03-12
application.hbs: Disable autocorrect/spellcheck on search input

xtian:


jtgeibel:
Thanks for the PR @xtian! The only attribute I'm not sure about it `autocorrect`.  It looks like this is [non-standard](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#Non-standard_attributes).

Would you try testing without this attribute on Safari to see if the attribute is necessary or not?

xtian:
@jtgeibel Yeah it's necessary. The autocorrect behavior is what led me to open this PR actually.

jtgeibel:
Okay, thanks!

@bors r+

PR #1672, opened 2019-03-16
Add support for graceful shutdown of the server

jtgeibel:
The server process now intercepts SIGINT and SIGTERM to initiate a
graceful shutdown of the server.

This is useful when tracking memory leaks locally, as both `hyper` and
`civet` are given a chance to return memory and shutdown cleanly.
However, this will not improve things in production as we also run an
instance of `nginx`, which will close all connections after receiving a
SIGTERM from Heroku.  From my preliminary investigation, it appears we
may need to customize the buildpack to change this behavior.

Additionally, the server will now briefly sleep before notifying Heroku
that it is ready to receive connections.  Also, when using `hyper` the
`Runtime` is configured to use 4 background threads.  This overrides
the default, which is one thread per CPU and provides consistency
between differently sized dynos.  The number of `conduit` background
threads are still configured via `SERVER_THREADS` and defaults to 50
in production.

### DIFF
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -1,19 +1,22 @@
 #![deny(warnings, clippy::all, rust_2018_idioms)]
 
 use cargo_registry::{boot, App, Env};
-use jemalloc_ctl;
 use std::{
     fs::File,
-    sync::{mpsc::channel, Arc},
+    sync::{mpsc::channel, Arc, Mutex},
+    thread,
+    time::Duration,
 };
 
 use civet::Server as CivetServer;
 use conduit_hyper::Service as HyperService;
+use futures::Future;
+use jemalloc_ctl;
 use reqwest::Client;
 
 enum Server {
     Civet(CivetServer),
-    Hyper(HyperService<conduit_middleware::MiddlewareBuilder>),
+    Hyper(tokio::runtime::Runtime),
 }
 
 use Server::*;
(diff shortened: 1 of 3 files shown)


src/bin/server.rs
```diff
@@ -66,19 +94,43 @@ fn main() {
...
-        let (_tx, rx) = channel::<()>();
-        rx.recv().unwrap();
+    // Block the main thread until the server has shutdown
+    match server {
+        Hyper(rt) => rt.shutdown_on_idle().wait().unwrap(),
+        Civet(server) => {
+            let (tx, rx) = channel::<()>();
+            ctrlc_handler(move || tx.send(()).unwrap_or(()));
+            rx.recv().unwrap();
+            drop(server);
```
markcatley: The drop seems unnecessary given it'll be dropped at the end of the block which immediately follows.

src/bin/server.rs
```diff
@@ -56,7 +59,32 @@ fn main() {
...
+
+        let (tx, rx) = futures::sync::oneshot::channel::<()>();
+        let server = server
+            .with_graceful_shutdown(rx)
+            .map_err(|e| log::error!("Server error: {}", e));
+
+        ctrlc_handler(move || tx.send(()).unwrap_or(()));
+
+        let mut rt = tokio::runtime::Builder::new()
+            .core_threads(4)
```
markcatley: Should this be configured with an environment variable?

src/bin/server.rs
```diff
@@ -56,7 +59,32 @@ fn main() {
...
+            .name_prefix("hyper-server-worker-")
+            .after_start(|| println!("Stared thread {}", thread::current().name().unwrap_or("?")))
+            .before_stop(|| {
+                println!(
```
markcatley: Would it be better to use the log macro? Some of the pre-existing printlns don't and some do. I am not sure what the rational is.

src/bin/server.rs
jtgeibel: I think it's fine to leave this hard-coded for now.  I added it mainly because sometimes we see the server boot with different levels of initial memory consumption and I'm trying to reduce potential variability (although I doubt this will really affect anything).  Long term, I expect we can switch back to all the defaults provided by `Runtime::new()`, which will set the number of core threads to the CPU count.

markcatley:
@jtgeibel The compile failure requires a `cargo fmt` and everything should be good.

Issue #1475, opened 2018-08-15
Proposal to add Cryptocurrencies category


kornelski:
Projects related to cryptocurrencies/blockchains are quite distinct, and Rust has some significant users in the world of cryptocurrencies. 

There already are more crates related to cryptocurrencies than: authentication, caching, configuration, wasm, http clients, macos apis, cargo plugins, build utilities, value formatting, date and time, internationalization, template engines, audio, video, text editors, emulators, simulators and visualization.

I suggest adding Cryptocurrencies as a subcategory of Cryptography. It could be described as "Libraries and tools for digital currencies, and distributed ledgers."

I've prototyped it by fuzzy matching keywords of existing crates to form a new category:

https://crates.rs/cryptography/cryptocurrencies

kzys:
@kornelski @carols10cents Resolving? (Why didn't @bors close the issue?)

PR #1674, opened 2019-03-17
Add Cryptocurrencies category

kornelski:
As described in #1475, there are many crates related to cryptocurrencies, which don't have a better category than cryptography.


carols10cents:
Thanks! 

@bors r+

PR #1675, opened 2019-03-18
Allow download counts to fail to be updated

### LINKED ISSUE #1387: Downloading a crate should continue to work if the database is down
The crate download endpoint only needs a database connection to update the download counter. We should have a lower timeout threshold for grabbing a connection on that endpoint, and just skip updating the downloads counter if it fails. People's builds should not break when our database is down for maintenance.

sgrif:
This ensures that the download endpoint still works even if counting the
download fails. The main case that we expect failure to occur is when
the application is in read only mode. However, even if an unexpected
failure occurs, we still want `cargo build` to succeed. Counting
downloads is always considered optional -- it's much more important that
people's builds succeed than having accurate download stats.

Note that we still require a database connection from the pool. In
theory, we could allow getting the version ID to fail as well, and just
blindly redirect to S3 no matter what, and rely on a 404 happening
there. However, this could result in successful builds in the event that
the index is out of sync with our DB, since we don't routinely clean up
the S3 bucket. Our current test behavior says that we 404 instead of
redirecting, so I've left things as is.

Once https://github.com/sfackler/r2d2/pull/73 is released, my plan is to
try to get a DB connection with a shorter timeout, and if we were able
to get a DB connection, do our current behavior, but if that fails just
blindly redirect to S3. We don't expect getting a connection to fail
unless we're being DDOS'd, nor do we expect requests to this endpoint
for crates that don't exist (since they should only come for things in
the index).

Fixes #1387

### DIFF
--- a/src/tests/krate.rs
+++ b/src/tests/krate.rs
@@ -1262,7 +1262,7 @@ fn download() {
 
 #[test]
 fn download_nonexistent_version_of_existing_crate_404s() {
-    let (app, anon, user) = TestApp::init().with_user();
+    let (app, anon, user) = TestApp::with_proxy().with_user();
     let user = user.as_model();
 
     app.db(|conn| {
--- a/src/controllers/version/downloads.rs
+++ b/src/controllers/version/downloads.rs
@@ -6,8 +6,6 @@ use crate::controllers::prelude::*;
 
 use chrono::{Duration, NaiveDate, Utc};
 
-use crate::Replica;
-
 use crate::models::{Crate, VersionDownload};
 use crate::schema::*;
 use crate::views::EncodableVersionDownload;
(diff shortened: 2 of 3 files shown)


bryanburgers:
I didn't test it locally, but this looks good to me.

src/tests/krate.rs
```diff
@@ -1262,7 +1262,7 @@ fn download() {
 
 #[test]
 fn download_nonexistent_version_of_existing_crate_404s() {
-    let (app, anon, user) = TestApp::init().with_user();
+    let (app, anon, user) = TestApp::with_proxy().with_user();
```
jtgeibel: This works fine for me locally with just `init()`, as nothing in this test connects to S3 or Github.  Otherwise, this PR looks great to me.

sgrif: Using `init` here is caring too much about the internals of the endpoint. This test shouldn't care whether we 404 before talking to S3 or not, just that something 404s at some point. While getting the 404 to come from S3 will certainly require some more changes (adding a recording), it shouldn't panic if we get to the uploader at all.

PR #1676, opened 2019-03-18
WIP: Rate limit the publish crate endpoint

bryanburgers:
A start to rate limiting the publish crate endpoint. Submitting to get
early feedback on whether this is the right direction to go.

---

**Why not use middleware?**

I originally wanted something declarative, and thought middleware would
be the ideal place for rate limiting. It looks like our framework does
not have per-route middleware, and routing does not happen until after
all of the middleware has run. I didn't see a clean way (other than
parsing the URL a second time and maintaining two parallel lists of URL
routes) to make this work.

**Can we still do declarative rate limiting?**

We can. In `src/router.rs`, I ended up with code like

```
api_router.push("/crates/new",
  c(krate::publish::publish).rl(rate_limit::RateLimitCategory::Publish));
```

However, the initial use case is to rate limit new crates differently
from new versions, and that will be imperative anyway, so I didn't
pursue this any further.

If this is valuable for other endpoints that don't require in-controller
logic, we can add it later.

---

Left to do:

- [ ] Diesel migration to create a table
- [ ] Complete the RateLimiterPostgres implementation
- [ ] Select which limiter to use based on the environment
- [ ] Write tests

### DIFF
--- a/src/util/errors.rs
+++ b/src/util/errors.rs
@@ -259,6 +259,31 @@ impl fmt::Display for Unauthorized {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct TooManyRequests;
+
+impl CargoError for TooManyRequests {
+    fn description(&self) -> &str {
+        "too many requests"
+    }
+
+    fn response(&self) -> Option<Response> {
+        let mut response = json_response(&Bad {
+            errors: vec![StringError {
+                detail: "too many requests have been seen recently".to_string(),
+            }],
+        });
+        response.status = (429, "Too Many Requests");
+        Some(response)
+    }
+}
+
+impl fmt::Display for TooManyRequests {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        "too many requests have been sent recently".fmt(f)
+    }
+}
+
 #[derive(Debug)]
 struct BadRequest(String);
 
--- a/src/util/rate_limit.rs
+++ b/src/util/rate_limit.rs
@@ -1,0 +1,167 @@
+use conduit::Request;
+use std::collections::HashMap;
+
+use crate::db::{DieselPool};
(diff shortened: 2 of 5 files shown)


bryanburgers:
@sgrif @jtgeibel I took a stab at an in-app rate limiter. Before I get too deep, is this the right direction?

src/util/rate_limit.rs
```diff
@@ -0,0 +1,167 @@
...
+#[derive(Debug, Clone)]
+pub struct RateLimitSettings {
+    /// The code for this category. Can be stored in a database, etc.
+    pub key: String,
+    /// The maximum number of tokens that can be acquired.
+    pub max_amount: usize,
+    /// How often we refill
+    pub refill_time: Duration,
+    /// The number of tokens that are added during a refill.
+    pub refill_amount: usize,
```
sgrif: I don't think we need to make this configurable. It'll always be 1 for the foreseeable future.

src/util/errors.rs
```diff
@@ -259,6 +259,31 @@ impl fmt::Display for Unauthorized {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct TooManyRequests;
```
sgrif: This should have a retry after field.

bryanburgers:
Thanks for the review! Also, I won’t be able to continue working on this until next week probably, so if anybody is already working on this, go right ahead! I won’t be offended.

sgrif:
Thanks for working on this @bryanburgers. This is a bit of a funky feature in that it's high priority for us, and it's something that can quickly shift in what we on the operations side want. As we've discussed here and privately, I've gone ahead with this on my own since we want to land this soon. Thanks again for taking a swing at this, and I hope this doesn't discourage you from tackling some of our other open issues.

Closing in favor of #1681

PR #1677, opened 2019-03-18
Update swirl

sgrif:
New features in this version:

- Codegen for jobs which reduces boilerplate when defining them
- Jobs no longer need to be registered with the runner
- Rework of how `run_all_pending_jobs` works so errors loading the job
  actually bubble up
- Configurable timeout for how long to wait for a job to begin running

We aren't fully taking advantage of the timeout yet, I'd eventually like
to try rebuilding the runner in-process a few times when an error
happens, and then crash the process if it fails a few times in a row.
This needs a bit more refactoring though (I'm not sure if we want to
keep the `Repository` in memory like we are now, and whether we want to
assume a full re-clone is needed on error). For now I've set it to
ensure our jobs don't hang forever though.

Warts in this version:

- Rust thinks imports only used in the job body are unused.
  (https://github.com/sgrif/swirl/issues/6)
  - Worked around for now by moving the imports into the job body
- `swirl::Job` has to be imported by calling code
- Codegen assumes that `serde::Deserialize` and `serde::Serialize` are
  available. (https://github.com/sgrif/swirl/issues/9)

Issues with this version:

- The new impl of `run_all_pending_jobs` will return an error if the DB
  is in read only mode. If the DB is read only, we couldn't have
  enqueued jobs anyway, so the workaround is "fine", but we need some
  more robust APIs in swirl to fix this. This only affects us in tests.
  - https://github.com/sgrif/swirl/issues/8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,9 +205,8 @@ dependencies = [
  "scheduled-thread-pool 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "semver 0.9.0 (git+https://github.com/steveklabnik/semver.git)",
  "serde 1.0.89 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.89 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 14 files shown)


sgrif:
Note: Diff stats are mostly just indentation changes. There's no substantial changes to the job bodies other than removing `self.`

PR #1678, opened 2019-03-18
Add monitoring for common spam patterns

sgrif:
We've noticed some common patterns in recent spam attacks. While our
response time on these has been ok, we can look for some of these common
patterns and page whoever is on-call earlier than we'd otherwise notice.
The exact patterns we look for is considered sensitive information, and
thus not in the repo and should not be discussed publicly.

Note that I've opted to look for crates that are likely spam, rather
than volume. Volume is more likely to have false positives, and is
better handled by more aggressive rate limiting.

This assumes that we consider a spam attack to be something we always
want to page for. Since we have better coverage of someone watching
discord most hours, we could alternatively have this post in a private
channel, and let whoever is awake determine if it's worth paging over.

If someone does get paged, it's assumed that this will get resolved
either by them taking action to remove the crates, or if the crate is
legitimate, by updating the config vars to remove that pattern.

### DIFF
--- a/src/bin/monitor.rs
+++ b/src/bin/monitor.rs
@@ -87,9 +87,7 @@ fn check_spam_attack(conn: &PgConnection) -> CargoResult<()> {
         .optional()?;
 
     if let Some(bad_crate) = bad_crate {
-        event_description = Some(
-            format!("Crate named {} published", bad_crate)
-        );
+        event_description = Some(format!("Crate named {} published", bad_crate));
     }
 
     let mut query = version_authors::table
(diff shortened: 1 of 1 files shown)


carols10cents:
Can you privately share the patterns you're thinking of setting these environment variables to initially?

jtgeibel:
What do you think about also adding a check to monitor the count of crates where `created_at` is within the last hour?  If that number is high, say over 60, then we could alert as well.  This could work along side the rate limiting, to catch people evading it with multiple accounts.

PR #1695, opened 2019-03-23
WIP: Remove `diesel_cli` on Travis

### LINKED ISSUE #1603: Update .travis.yml now that we are using Homu
It looks like Homu uses an `auto` branch instead of `staging`.  In our `.travis.yml` file we exclude builds on the `staging.tmp` branch.  This was previously required with `bors-ng`.

We can remove this exception now, but it's not yet clear to me if we need to add a similar rule for Homu.

jtgeibel:
WIP: testing to see if this actually works

Our build script should run all migrations before building tests and
installing `diesel_cli` takes build time and cache size.

This PR also closes #1603.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@ language: rust
 sudo: required
 dist: xenial
 
-# Ignore this branch per bors-ng documentation
-branches:
-  except:
-    - staging.tmp
-
 cache:
   cargo: true
   directories:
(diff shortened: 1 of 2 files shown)


jtgeibel:
I was hoping this would help reduce our cache sizes, but obviously we need `diesel_cli` to setup the database initially.  I'm closing this PR.

PR #1696, opened 2019-03-25
Require a User-Agent header, take 2

sgrif:
We deployed this back in October, and quickly realized that old versions
of Cargo (Rust 1.15 and earlier) didn't set a User-Agent header.
Notably, the version of Cargo used for bootstrapping the compiler back
then was this old, so we accidentally broke the build for rustc.

We still see build traffic for these old versions, and we likely will
never be willing to break `cargo build` for old versions of Cargo.
However, as we discussed back in early November, we're fine with doing
this for all other endpoints.

This change will break versions of Cargo prior to Rust 1.16 for all
operations that talk to crates.io, except for `cargo build`. We no
longer receive any traffic for publish, yank, unyank, or owners from
these old versions. (`cargo search` hits an endpoint that is also hit by
bots, so I can't say for sure if it's coming from cargo or random
crawlers, which is kinda the point of this requirement in the first
place)

### DIFF
--- a/src/tests/server.rs
+++ b/src/tests/server.rs
@@ -1,18 +1,28 @@
-// Note: Temporarily disabled because cargo-vendor doesn't include a
-// User-Agent header and Rust's CI broke. If this is still commented out
-// by Nov 7, 2018 ping the crates.io team.
+use conduit::Method;
 
-// use conduit::{Handler, Method};
-//
-// use {app, req};
-//
-//
-// #[test]
-// fn user_agent_is_required() {
-//     let (_b, _app, middle) = app();
-//
-//     let mut req = req(Method::Get, "/api/v1/crates");
-//     req.header("User-Agent", "");
-//     let resp = t!(middle.call(&mut req));
-//     assert_eq!(resp.status.0, 403);
-// }
+use crate::builders::*;
+use crate::util::*;
+
+#[test]
+fn user_agent_is_required() {
+    let (_app, anon) = TestApp::init().empty();
+
+    let mut req = anon.request_builder(Method::Get, "/api/v1/crates");
+    req.header("User-Agent", "");
+    let resp = anon.run::<()>(req);
+    resp.assert_status(403);
+}
+
+#[test]
+fn user_agent_is_not_required_for_download() {
+    let (app, anon, user) = TestApp::init().with_user();
+
+    app.db(|conn| {
+        CrateBuilder::new("dl_no_ua", user.as_model().id).expect_build(conn);
+    });
+
+    let mut req = anon.request_builder(Method::Get, "/api/v1/crates/dl_no_ua/0.99.0/download");
+    req.header("User-Agent", "");
+    let resp = anon.run::<()>(req);
+    resp.assert_status(302);
+}
--- a/src/middleware.rs
+++ b/src/middleware.rs
@@ -93,10 +93,8 @@ pub fn build_middleware(app: Arc<App>, endpoints: R404) -> MiddlewareBuilder {
         let ips = ip_list.split(',').map(String::from).collect();
         m.around(block_ips::BlockIps::new(ips));
     }
-    // Note: Temporarily disabled because cargo-vendor doesn't include a
-    // User-Agent header and Rust's CI broke. If this is still commented out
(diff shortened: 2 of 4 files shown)


src/tests/server.rs
```diff
@@ -1,18 +1,28 @@
...
-// }
+use crate::builders::*;
+use crate::util::*;
+
+#[test]
+fn user_agent_is_required() {
+    let (_app, anon) = TestApp::init().empty();
+
+    let mut req = anon.request_builder(Method::Get, "/api/v1/crates");
+    req.header("User-Agent", "");
```
bryanburgers: Is this testing when a User-Agent header exists, but is the empty string? If so, do we need a test where the header doesn't exist as well?

sgrif: Unfortunately, conduit-test doesn't give us a way to unset a header. We're setting the user agent to a non-empty string when the requests are first constructed so we don't have to set it for every test besides these two. We could refactor things to apply the default user agent later, so we have a place to construct a request without it here, but I've left it this way for two reasons:

- Explicitly setting `req.header("User-Agent", "")` makes it much more clear what this is actually testing
- It'd be very odd if we were treating a header with no value differently than a header not being sent at all (the only case I'm aware of where there's a semantic difference is when sending an HTTP/1.1 -> HTTP/2 upgrade request, which specifies that an `HTTP2-Settings` header must be present)

With that in mind, do you still think it's worth the additional test?

bryanburgers: Nope, I don't. 👍 LGTM.

If testing *without* the header was easy to do, I'd like to see it. But I'm with you that it would be very odd if the framework was treating a non-existent header differently than an empty header.

I just brought it up because I've been bitten in different places (not HTTP headers) where non-existent != empty string.

bryanburgers:
Pulled down and tested locally. Everything seems to working when using `curl -H 'User-Agent:'`.

```bash
bryan ~/projects/cratesio/ pr/1696 $ curl -i --head localhost:8888/api/v1/crates/bryan -H "User-Agent: test/1"
HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8
Set-Cookie: cargo_session=nFPRnOh/tCaZyaOecttwdaQd8auK817pCMXxVWQ0a+o=; HttpOnly; Path=/
Content-Length: 8621

bryan ~/projects/cratesio/ pr/1696 $ curl -i --head localhost:8888/api/v1/crates/bryan -H "User-Agent:"
HTTP/1.1 403 Forbidden
Content-Length: 631
Set-Cookie: cargo_session=nFPRnOh/tCaZyaOecttwdaQd8auK817pCMXxVWQ0a+o=; HttpOnly; Path=/

bryan ~/projects/cratesio/ pr/1696 $ curl -i --head localhost:8888/api/v1/crates/bryan/0.1.0/download -H "User-Agent: test/1"
HTTP/1.1 302 Found
Location: /crates/bryan/bryan-0.1.0.crate
Set-Cookie: cargo_session=nFPRnOh/tCaZyaOecttwdaQd8auK817pCMXxVWQ0a+o=; HttpOnly; Path=/

bryan ~/projects/cratesio/ pr/1696 $ curl -i --head localhost:8888/api/v1/crates/bryan/0.1.0/download -H "User-Agent:"
HTTP/1.1 302 Found
Location: /crates/bryan/bryan-0.1.0.crate
Set-Cookie: cargo_session=nFPRnOh/tCaZyaOecttwdaQd8auK817pCMXxVWQ0a+o=; HttpOnly; Path=/
```

Note that this **is** testing the non-existence of the header. From `curl`'s man-page:

> -H, --header <header>
> (HTTP) Extra header to include in the request when sending HTTP to a server. You may specify any number of extra headers. Note that if you should add a custom header that has the same name as  one  of the  internal  ones curl would use, your externally set header will be used instead of the internal one. This allows you to make even trickier stuff than curl would normally do. You should not replace internally set headers without knowing perfectly well what you're doing. **Remove an internal header by giving a replacement without content on the right side of the colon, as in:  -H  "Host:".**  If  you send the custom header with no-value then its header must be terminated with a semicolon, such as -H "X-Custom-Header;" to send "X-Custom-Header:".

Issue #1659, opened 2019-03-11
Have a way to yank a crate from the UI


carols10cents:
What needs to be done: This looks like a lot, but I tried to provide a lot of detail and examples to work from. The actual code changes shouldn't be very big.

- If the current user is an owner of the current crate, add a link or button in the sidebar with each version that says "yank" if the version is not yanked and "unyank" if the version is yanked. (Or is there a better place to put this control? I'm open to different UIs)
  - The HTML for the versions list in the sidebar is in https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/app/templates/crate/version.hbs#L195-L208
  - To tell if the current user is an owner, use the `isOwner` helper like the "Manage owners" section does here https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/app/templates/crate/version.hbs#L167-L173
  - Because we only show the latest 5 versions in the sidebar, the new link/button should be shown on the "All versions" page as well so any version can be yanked. The "All versions" page's HTML is here https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/app/templates/crate/versions.hbs

- The link/button should send a [`DELETE` request to `/api/v1/crates/[crate name]/[version num]/yank`](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/router.rs#L23) for yanking and a [`PUT` request to `/api/v1/crates/[crate name]/[version num]/unyank`](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/router.rs#L23) for unyanking.
  - A similar link/button is the "Remove" button for removing an owner https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/app/templates/crate/owners.hbs#L67
  - Which calls the ember `removeOwner` controller function: https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/app/controllers/crate/owners.js#L34-L49
  - Which calls the ember crate model `removeOwner` function: https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/app/models/crate.js#L48-L50
[...]

PR #1700, opened 2019-03-31
Add audit trail to the publish, yank and unyank transactions.

markcatley:
Contributes to [Issue 1548](https://github.com/rust-lang/crates.io/issues/1548)
Based on the commits from [PR 1604](https://github.com/rust-lang/crates.io/pull/1604)


src/models/action.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub time: NaiveDateTime,
+}
+
+#[derive(Copy, Clone, Debug, Insertable)]
+#[table_name = "version_owner_actions"]
+pub struct NewVersionOwnerAction {
+    pub version_id: i32,
+    pub owner_id: i32,
+    pub owner_token_id: i32,
+    pub action: i32,
```
markcatley: What trait to I have to implement to get diesel to do this conversion internally?

src/controllers/krate/publish.rs
```diff
@@ -149,6 +153,11 @@ pub fn publish(req: &mut dyn Request) -> CargoResult<Response> {
         )?
         .save(&conn, &new_crate.authors, &verified_email_address)?;
 
+        if let AuthenticationSource::ApiToken(token_id) = req.authentication_source()? {
```
markcatley: What should I be doing in the else block? Getting the token in some other way? Returning an unauthorized error?

src/controllers/version/yank.rs
```diff
@@ -36,6 +38,15 @@ fn modify_yank(req: &mut dyn Request, yanked: bool) -> CargoResult<Response> {
         return Err(human("must already be an owner to yank or unyank"));
     }
 
+    if let AuthenticationSource::ApiToken(token_id) = req.authentication_source()? {
```
markcatley: What should I be doing in the else block? Getting the token in some other way? Returning an unauthorized error?

markcatley:
What recommendations would you have on what should be unit tested?

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE version_owner_actions (
+    id SERIAL PRIMARY KEY,
+    version_id INTEGER REFERENCES versions(id) ON DELETE CASCADE,
```
markcatley: This is based on the previous commit, but my gut feeling is that version_id and owner_id should be NOT NULL. owner_token_id should possibly be also NOT NULL depending on other parts of this commit.

src/models/action.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub owner_id: i32,
+    pub owner_token_id: Option<i32>,
+    pub action: i32,
+}
+
+impl NewVersionOwnerAction {
+    pub fn new(
+        version_id: i32,
+        owner_id: i32,
+        owner_token_id: Option<i32>,
```
markcatley: This is None if it's session auth. Is that ok? Is there a better way to represent that?

src/models/action.rs
sgrif: `FromSqlRow`

markcatley: I'm already deriving that. If I change the type I get ```the trait `diesel::Expression` is not implemented for `models::action::VersionAction````

sgrif: Sorry, you're missing `AsExpression`

src/models/action.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub version_id: i32,
+    pub owner_id: i32,
+    pub owner_token_id: Option<i32>,
+    pub action: VersionAction,
+    pub time: NaiveDateTime,
+}
+
+#[derive(Copy, Clone, Debug, Insertable)]
+#[table_name = "version_owner_actions"]
+pub struct NewVersionOwnerAction {
```
sgrif: This struct doesn't ever look like it's meaningfully used. Can we just drop it and use the [the tuple form of insertion](http://docs.diesel.rs/diesel/dsl/fn.insert_into.html#using-a-tuple-for-values) instead?

src/models/token.rs
```diff
@@ -46,6 +46,28 @@ impl ApiToken {
             last_used_at: self.last_used_at,
         }
     }
+
+    pub fn find_by_api_token_and_revoked(
```
sgrif: This function always gets `false` for `revoked_`. What do you think about leaving this function defined as it was before?

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE version_owner_actions (
+    id SERIAL PRIMARY KEY,
+    version_id INTEGER REFERENCES versions(id) ON DELETE CASCADE NOT NULL,
+    owner_id INTEGER REFERENCES users(id) NOT NULL,
```
sgrif: Do you think it might be confusing to have a column called `owner_id` pointing to the `users` table and not the `crate_owners` table? What do you think about replacing `owner` with `user` here?

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
```diff
@@ -0,0 +1,8 @@
...
+    owner_token_id INTEGER REFERENCES api_tokens(id),
```
sgrif: What do you think about calling this `api_token_id` instead?

src/models/action.rs
```diff
@@ -0,0 +1,95 @@
...
+    serialize::{self, Output, ToSql},
+    sql_types::Integer,
+};
+use std::io::Write;
+
+use crate::models::{ApiToken, User, Version};
+use crate::schema::*;
+
+#[derive(Debug, Clone, Copy, FromSqlRow, AsExpression)]
+#[repr(u32)]
```
sgrif: What do you think about making this `#[repr(i32)]` instead? Then we can drop the `Into` impl and write `self as i32` instead.

sgrif:
I'd like to see some integration tests for this as well

markcatley:
Sorry that I left this alone for so long. I noticed that @carols10cents merged PR 1604 that this was based on so I've updated this PR and made the changes suggested.

Some of the changes @sgrif made are related to code that's already been merged. I have not made these, and @carols10cents and @sgrif will have to decide which of these should be made.

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
markcatley: This has been merged without NOT NULL constraints. Should we add them? Is it safe to add a NOT NULL constraint to a column?

src/models/action.rs
markcatley: Done.

src/models/token.rs
markcatley: Done.

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
markcatley: I agree with you but this is no longer in this PR. It's in a PR that's already been merged.

Would you like me to rename the field?

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
markcatley: I agree with you but this is no longer in this PR. It's in a PR that's already been merged.

Would you like me to rename the field?

src/models/action.rs
markcatley: Done

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
carols10cents: Whoops, yes, you're right-- these should be `NOT NULL` so that `VersionOwnerAction` doesn't need to have `Option` in `version_id`, `owner_id`, and `owner_token_id` columns. Nothing is adding data to this table right now, so it's safe to add `NOT NULL` constraints to these columns. Do you mind adding a migration doing that to this PR?

carols10cents: Actually, thinking about it more, we don't currently allow publishing/yanking/unyanking from the UI (that is, session auth) but [there have been requests for at least yanking/unyanking](https://github.com/rust-lang/crates.io/issues/1659). So `owner_token_id` should NOT have a `NOT NULL` constraint, and it will be null if performed via the UI/session auth.

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
carols10cents: I could go either way on renaming `owner_id` to `user_id`... up to you, @markcatley, if you'd like to add a migration to this PR to rename the field.

migrations/2019-01-26-090348_create_version_owner_actions/up.sql
carols10cents: Same with this spot, up to you whether you'd like to add a renaming migration to this PR.

markcatley:
Hi @carols10cents @sgrif,

I think it's probably a good idea to rename those fields. Never an easier time than when the table is unused (I hope). I believe all the changes have now been made.

Kind regards,
Mark

src/models/token.rs
```diff
@@ -46,6 +46,24 @@ impl ApiToken {
             last_used_at: self.last_used_at,
         }
     }
+
+    pub fn find_by_api_token(conn: &PgConnection, token_: &str) -> QueryResult<ApiToken> {
```
carols10cents: I like this code being on `ApiToken` rather than on `User` 👍

src/models/action.rs
```diff
@@ -1,26 +1,90 @@
...
 }
+
+impl VersionOwnerAction {
+    pub fn all(conn: &PgConnection) -> QueryResult<Vec<VersionOwnerAction>> {
+        version_owner_actions::table.load(conn)
+    }
+
+    pub fn by_version_id_and_action(
+        conn: &PgConnection,
+        _version_id: i32,
```
carols10cents: This is a really small consistency fix I forgot to mention last time I looked at this PR, but I think I'd like to see it fixed before merging. 

In Rust, the convention is to start parameter and variable names with underscores when they aren't used. So when I look at this signature, I think that `_version_id` and `_action` aren't going to be used in the body, but they are. I know we have to use some sort of name that doesn't conflict with the table's columns. In some other places, [like the `find_by_api_token` function](https://github.com/rust-lang/crates.io/pull/1700/files#diff-1cdb3b1876ff75ba470bd31a2dda81e0R50), we instead *end* the parameter name with an underscore.

Do you mind changing `_version_id` and `_action` in this function, and the 4 parameters in the `insert_version_owner_action` function, to end with underscores rather than starting with underscores?

migrations/2019-11-03-003946_tidy_up_version_owner_actions/up.sql
```diff
@@ -0,0 +1,20 @@
+-- I expect that it will be save to apply this migration in production as the code that adds records
```
carols10cents: Thank you so much for including this migration!

src/tests/krate.rs
```diff
@@ -1589,6 +1589,120 @@ fn publish_after_yank_max_version() {
...
+
+    let (app, anon, _, token) = TestApp::full().with_token();
+
+    app.db(|conn| assert!(VersionOwnerAction::all(&conn).unwrap().is_empty()));
+
+    // Upload a new crate, putting it in the git index
+    let crate_to_publish = PublishBuilder::new("fyk");
+    token.enqueue_publish(crate_to_publish).good();
+
+    // make sure it doesn't have any audit actions
```
carols10cents: I realize in general these tests are somewhat placeholders because the JSON API isn't returning audit actions yet, but I was a bit confused by this comment, and I'd like to leave clear instructions for whoever picks up the next step and adds in the API part. 

At this point in the test, there will be one audit action, because you can't publish a new crate without publishing at least one version of that crate. Also, no audit actions have taken place between these lines and the lines that have the comment "make sure it has one audit action".

So I think this comment should say "make sure it has one audit action", the next line saying "do this as a full integration test once the api is in place" is fine to stay, and then remove the comments on lines 1609 and 1610 because they'll be redundant with the comments here.

src/tests/krate.rs
```diff
@@ -1589,6 +1589,120 @@ fn publish_after_yank_max_version() {
...
+fn yank_records_an_audit_action() {
+    use cargo_registry::models::{VersionAction, VersionOwnerAction};
+
+    let (app, anon, _, token) = TestApp::full().with_token();
+
+    // Upload a new crate, putting it in the git index
+    let crate_to_publish = PublishBuilder::new("fyk");
+    token.enqueue_publish(crate_to_publish).good();
+
+    // make sure it doesn't have any audit actions
```
carols10cents: Same here, this is confusing because this version will have 1 audit action (publish) at this point...

src/tests/krate.rs
```diff
@@ -1589,6 +1589,120 @@ fn publish_after_yank_max_version() {
...
+            VersionOwnerAction::by_version_id_and_action(conn, version_id, VersionAction::Yank)
+                .unwrap()
+                .is_empty()
+        )
+    });
+
+    // yank it
+    token.yank("fyk", "1.0.0").good();
+
+    // make sure it has one audit action
```
carols10cents: ...and at this point, there will be 2 total audit actions, but one `Yank` audit action, so I think this comment should be updated to say "make sure it has one yank audit action" or "make sure it has one publish and one yank audit action" (since the API will probably return all audit actions rather than have a way to do filtering as we can do with the database here)

src/tests/krate.rs
```diff
@@ -1589,6 +1589,120 @@ fn publish_after_yank_max_version() {
...
+            VersionAction::Unyank
+        )
+        .unwrap()
+        .is_empty())
+    });
+
+    // unyank version 1.0.0
+    token.unyank("fyk", "1.0.0").good();
+
+    // make sure it has one audit action
```
carols10cents: And here, please update this comment to say "make sure it has one publish, one yank, and one unyank audit action".

carols10cents:
I've resolved the merge conflicts and made the small outstanding fixes so that we can merge this in, which I'll do once CI passes :) Thanks!

PR #1702, opened 2019-04-01
Fix new lint on nightly

sgrif:
`unused_imports` now triggers for redundant imports.
`diesel::Connection` is part of `diesel::prelude`, `mem::drop` is part
of `std::prelude`, and `InvitationResponse` was explicitly imported at
the top of the file


jtgeibel:
Are you there bors?

@bors r+

PR #1708, opened 2019-04-06
Our chat is now on Discord

kzys:
We no longer use irc.mozilla.org.


carols10cents:
Good catch! Thank you! ❤️ 

@bors: r+

PR #1712, opened 2019-04-09
Rename parsing category to parsing tools

kornelski:
The parsing and parser-implementations categories are commonly confused, since "parsing" correctly describes use of high-level parsers to parse specific formats, and doesn't sound related to low-level libraries for building parsers.


kzys:
Looks good to me.

PR #1714, opened 2019-04-10
Remove jQuery from Download Graph

kzys:
We may need to remove jQuery before enabling Fastboot.


sgrif:
This appears to be causing errors.

kzys:
Hmm, right. Let me take a look.

PR #1719, opened 2019-04-14
app: change let declarations to const

otaviopace:
Hello! 👋 🙂 

This Pull Request changes every `let` that could be a constant to a `const` declaration on the `app` folder, which contains the frontend for `crates.io`.

## But why?

If a variable **won't be changed** later, it **should be a constant**. So by using `const` instead of `let` on variables that won't change, **helps readability**. Since who ever is reading the code, will consider that certain variables won't change.

Also, if we keep **constants declarations with `let`**, the `JavaScript` **compiler will let us change that value later**. That is really bad and **can lead to human error**.

We are `Rust` developers, let's use compilers to our favor! 🎉

### DIFF
--- a/app/adapters/api-token.js
+++ b/app/adapters/api-token.js
@@ -6,9 +6,9 @@ export default DS.RESTAdapter.extend({
         return 'tokens';
     },
     createRecord(store, type, snapshot) {
-        let data = {};
-        let serializer = store.serializerFor(type.modelName);
-        let url = this.buildURL(type.modelName, null, snapshot, 'createRecord');
+        const data = {};
+        const serializer = store.serializerFor(type.modelName);
+        const url = this.buildURL(type.modelName, null, snapshot, 'createRecord');
 
         serializer.serializeIntoHash(data, type, snapshot, { includeId: true });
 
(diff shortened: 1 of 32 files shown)


Turbo87:
> We are `Rust` developers, let's use compilers to our favor! 🎉

yeah, except in JavaScript `const/let` has a different meaning than `let/let mut` in Rust. In JS the only thing that is const is the binding, so `const foo = {}; foo.bar = 'baz';` is still legal, which means that ultimately the `const` isn't helping much.

for more details on why most of the Ember.js community prefers `let` (except for actual constants) have a look at https://madhatted.com/2016/1/25/let-it-be

otaviopace:
Honestly I wasn't convinced to use `let` over `const` on JavaScript, I really disagree with this statement:

```
The semantics of const are more important than the behavior.
```

`const` declarations help:

- To avoid mutation, which can be confusing to debug, IMO. Immutability FTW.
- To avoid **unintentional** mutation, which can be **very bad**.
- To be more semantic (which kind of goes with the statement of `Matthew Beale` that I don't like, lol), since a variable won't be changed, why should we make it mutable?

**However** I do understand that apparently `let` is preferred over `const` for most cases for the `Ember.js` community.

Should we close this PR then? What do you guys think?

locks:
Hey @otaviopace!
Like Tobias mentioned, I'm also very partial for what the Ember community calls "liberally let", and we get the benefit of shared solutions/configs for tools like prettier, eslint, etc. I would vote for closing this :)

Issue #1598, opened 2019-01-18
Rename "Game engines" and "Games"


erlend-sh:
<img width="961" alt="screenshot 2019-01-18 at 10 43 40" src="https://user-images.githubusercontent.com/583842/51380516-2ffd4700-1b12-11e9-959a-058192ecb850.png">

A "Game engine" is a specific thing in the game industry. Most of what is listed in this category is just general game development tooling, not game engines. I suggest renaming this category to "Game Development".

"Games" on the other hand is very misleading, as is made evident by this category's description. I suggest renaming it to "Fun" or "For Fun".

erlend-sh:
I found precedent for this type of discussion in #488. Is there any particular process I’m supposed to follow here?

sunjay:
To leave it backwards compatible, maybe it could be renamed to "Game Development / Game Engines"?

erlend-sh:
Sure, as long as the new slug is `game development`. But I really wouldn't worry too much about backwards compatibility here as the gamedev ecosystem is still young and hasn't established any taxonomy conventions yet.

sunjay:
I'm specifically talking about backwards compatibility for those who have already listed the game engine category in their crates. They shouldn't get unlisted if the name changes.

chsm:
Not familiar how the categories are implemented, but can't game engines be included as a child category of game development? Then folks who currently list the game engine category for this create can continue to be listed under the parent category (game development) as well?

Lisoph:
What about renaming "Games" to "Entertainment"? Seems more fitting even with the current description.

Lokathor:
Having a Games category is totally fine and should stay that way.

Renaming the other category to Game Development I agree with.

erlend-sh:
How can this be moved forward? There appears to be unanimous support for a renaming to “Game Development”

kzys:
@erlend-sh I've just made a PR, but we need to move existing crates to this new category.

Another way of doing that is changing the display name to "Game Development" first to see how the Rust community reacts, and change the slug later.

kzys:
@erlend-sh @Lokathor How about just changing the display name first? It doesn't satisfy

> Sure, as long as the new slug is game development

but I got your 👍 :)

erlend-sh:
Yeah I’m on board with just changing the display name as a first step.

kzys:
From @carols10cents (https://github.com/rust-lang/crates.io/pull/1720#pullrequestreview-310781411)

[...]

PR #1720, opened 2019-04-16
Rename "Game engines" to "Game Development"

kzys:
See #1598 for the discussion.

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -227,7 +227,15 @@ Crates for dealing with files and filesystems.\
 [game-engines]
 name = "Game engines"
 description = """
-Crates for creating games.\
+For crates that try to provide a \"one-stop-shop\" for \
+all of your game development needs.\
+"""
+
+[game-development]
+name = "Game development"
+description = """
+For crates that focus on some individual part of accelerating \
+the development of games.\
 """
 
 [games]


kzys:
Ah, changing the slug means that we have to make some changes on the production database as well...

kzys:
So, we cannot merge the change as is. Renaming "Game Engines" to "Game Development" makes sense though.

jtgeibel:
Would it be better to just add a new slug?  I think it may be confusing if developers still have to add the "game-engine" slug to their `Cargo.toml` even though the category has been renamed.

sgrif:
I think it'd be even more confusing to have a separate "Game Development" category ("Game Engines" aren't for "Game Development"?). We definitely can't just change the slug though, even if we updated the existing data we'd unlist all existing crates when they published a new version.

We should probably have some form of migration strategy here rather than just assuming we'll always get categories exactly right the first time. We'll likely need to add some form of `aliases` field to categories and use that for lookup as well on publish. I agree with @jtgeibel that it'd be confusing to change the display name by itself.

src/boot/categories.toml
```diff
@@ -225,7 +225,7 @@ Crates for dealing with files and filesystems.\
 """
 
 [game-engines]
-name = "Game engines"
+name = "Game Development"
```
sgrif: None of the other categories capitalize words other than the first.

kzys:
Sorry, I wasn't aware the impact of this change.

Game Engines are for Game Development, but the description ("Crates for creating games.") and the existing creates are more aligned with "Game Development". Game engines usually refer full-fledged frameworks such as Unity and/or Unreal Engine.

We already have a few categories that are the subset of the existing categories, such as Algorithms vs. Compression. I think having separate "Game Engine" is okay, while we may ask the Rust community to migrate their crates from Game Engine to Game Development though.

Having aliases doesn't need this community effort, but adds some complexity on the crates.io.

carols10cents:
I'm sorry we've let this sit so long without feedback!

I think adding a new "Game development" category with a description that it's for not-engines and adjusting the description for "Game engines" so that it's clear what belongs there vs. "Game development" would be a good path forward.

kzys:
Thanks. Will update the TOML soon.

carols10cents:
Noticed this PR was updated; looks great now! Thank you!!

@bors r+

PR #1721, opened 2019-04-17
Ignore auto instead of staging.tmp on Travis CI

### LINKED ISSUE #1603: Update .travis.yml now that we are using Homu

kzys:
Fixes #1603.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,11 @@ language: rust
 sudo: required
 dist: xenial
 
-# Ignore this branch per bors-ng documentation
+# Ignore this branch per homu
+# https://github.com/barosl/homu
 branches:
   except:
-    - staging.tmp
+    - auto
 
 cache:
   cargo: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,6 @@ language: rust
 sudo: required
 dist: xenial
 
-# Ignore this branch per homu
-# https://github.com/barosl/homu
-branches:
-  except:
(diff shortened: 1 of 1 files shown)


carols10cents:
Thank you!!

jtgeibel:
Doesn't bors run its tests on auto? I'm not sure this will work.

carols10cents:
Oops. I guess not. @kzys could you change this to just remove the ignoring of `staging.tmp` and the reference to bors-ng, but not ignore `auto`? Thanks!

jtgeibel:
@kzys thanks for looking into this!  I agree with the changes Carol suggests.

It looks like Homu recommends running tests `only:` on the `auto` branch.  While it is rarely used, we currently allow merging via GitHub GUI and allow pushes directly to `master`.  Those could be locked down, but in both cases it looks like someone with `rust-lang` org permissions would need to do that.  So unless we change that, we should continue running tests on all branches for now.

Thanks again!

kzys:
> just remove the ignoring of staging.tmp and the reference to bors-ng, but not ignore auto?

Doesn't it make the entire branches: block empty?

jtgeibel:
LGTM!

@bors r+

PR #1722, opened 2019-04-20
Run Frontend's tests separately

kzys:
Run Frontend's tests and Backend's tests in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,9 +58,9 @@ matrix:
         - cargo clippy --all-targets --all-features --all
         - cargo test
     - name: Frontend
-      rust: stable
+      language: node_js
+      node_js: 10
       install:
-        - nvm install 10
         - npm ci
       before_script: skip
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,16 +48,26 @@ matrix:
   allow_failures:
     - rust: nightly
   include:
-    - rust: stable
+    - name: Backend
+      rust: stable
       before_install:
-        - nvm install 10
-        - npm ci
         - rustup component add rustfmt-preview
         - rustup component add clippy-preview
       script:
         - cargo fmt -- --check
         - cargo clippy --all-targets --all-features --all
         - cargo test
+    - name: Frontend
+      rust: stable
+      before_install:
+        - nvm install 10
+        - npm ci
+      # skip diesel
+      install:
+        - echo
+      before_script:
+        - echo
+      script:
         - npm run lint:js
         - npm test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -48,16 +48,26 @@ matrix:
...
-        - nvm install 10
-        - npm ci
         - rustup component add rustfmt-preview
         - rustup component add clippy-preview
       script:
         - cargo fmt -- --check
         - cargo clippy --all-targets --all-features --all
         - cargo test
+    - name: Frontend
+      rust: stable
```
Turbo87: do we need `rust` here at all? 

```yaml
language: node_js
node_js: 10
```
 should automatically invoke `nvm install`

.travis.yml
```diff
@@ -48,16 +48,26 @@ matrix:
...
+      before_install:
+        - nvm install 10
+        - npm ci
```
Turbo87: this should probably be moved to the `install` step while we're at it

.travis.yml
```diff
@@ -48,16 +48,26 @@ matrix:
...
+      # skip diesel
+      install:
+        - echo
+      before_script:
+        - echo
```
Turbo87: why the `echo` lines?

.travis.yml
kzys: Does it work. I thought that the `language:` wouldn't be applied at the Build Matrix level.

.travis.yml
kzys: Right. Then we can remove `echo` on install:

.travis.yml
kzys: To override the top-level install/before_script: but we probably can get rid of most of them. Let me update the PR.

.travis.yml
kzys: Travis CI's documentation regarding using multiple languages also doesn't use `language:` on Build Matrix.

https://docs.travis-ci.com/user/customizing-the-build#installing-a-second-programming-language

Turbo87: https://github.com/skylines-project/skylines/blob/43a68499581ea0ad04ee70e40a6f3a1f49683928/.travis.yml#L81-L82

seems to work though 😉

.travis.yml
Turbo87: you can use `install: skip` in those cases (see https://docs.travis-ci.com/user/build-stages/#build-stages-and-deployments)

kzys: ohh, didn't know that.

locks:
oops xD

PR #1728, opened 2019-04-22
Fix assertion error in Ember for version downloads

### LINKED ISSUE #1704: Downloads statistics seem broken
The downloads graph at https://crates.io/crates/crossbeam-channel displays zero 7-day avg downloads for the latest version (0.3.8). This must be wrong since the latest version is widely used and definitely has been downloaded at least once.

I suspect something is wrong due to yanking. Version 0.3.5 was yanked and all versions newer than 0.3.5 appear to be downloaded zero times, which can't be true. Looks like a bug.

bryanburgers:
Fix the following assertion error that occurs when running the frontend in debug mode, which cause graphs to not display at all.

> Assertion Failed: You must include an 'id' for version-download in an object passed to 'push'

This has been happening since 3c92ca2b46a6fd166b4ae556b3d925907c85250c landed.

The fix involves telling Ember how to create a unique ID for a version-download object, using the composite key.

I haven't been able to confirm, but I believe this is the reason that graphs in production are only ever counting/displaying a single version. When using this change locally in debug mode against the live website, the downloads graphs are rendering correctly.

Fixes #1704.

### DIFF
--- a/app/serializers/version-download.js
+++ b/app/serializers/version-download.js
@@ -1,0 +1,7 @@
+import DS from 'ember-data';
+
+export default DS.RESTSerializer.extend({
+    extractId(modelClass, resourceHash) {
+        return `${resourceHash.date}-${resourceHash.version}`;
+    },
+});


kzys:
Looks good to me. Can we write tests around the serializers? How about testing `normalize()` method?

Turbo87:
@kzys I agree that having tests for this would be a good idea. I'm not sure that unit testing the serializer is the best approach though. Maybe we can find a way to acceptance-test this to make sure that it works from end-to-end. (well, without the actual backend though...)

kzys:
Until we have the way to write end-to-end tests, I'd like to keep some tests against this serializer. Unit tests help us to identify root causes quickly, while it tends to couple with the implementation.

bryanburgers:
I don’t typically write much Ember, and I didn’t see any existing tests testing serializers that I could use an example. I’d welcome some advice on how to add useful tests to this!

sgrif:
I agree it'd be great to add more end-to-end tests here, but getting that set up is a large enough task for its own PR. I've confirmed this does fix a major UI bug locally, so I'm going to merge this.

@bors: r+

PR #1730, opened 2019-04-23
Use `ember-cli-dependency-lint` to warn about duplicate addons

Turbo87:
see https://github.com/salsify/ember-cli-dependency-lint#motivation for more information


kzys:
Looks good to me!

PR #1733, opened 2019-04-24
Update `ember-fetch` to v6.5.1

Turbo87:
see https://github.com/ember-cli/ember-fetch/blob/v6.5.1/CHANGELOG.md


kzys:
Looks good to me.

PR #1735, opened 2019-04-26
Clippy cleanups

jtgeibel:
This series first addresses a warning now produced on nightly.  Then `clippy::all` and `rust_2018_idioms` are explicitly denied for all lib, bin, and test artifacts.

By denying `clippy::all`, the RLS will automatically show clippy warnings in an IDE.  This only affects the RLS and `cargo clippy --all-targets` should still used in a command line workflow.

### DIFF
--- a/src/background_jobs.rs
+++ b/src/background_jobs.rs
@@ -44,7 +44,7 @@ impl Environment {
             .map(|(u, p)| (u.as_str(), p.as_str()))
     }
 
-    pub fn connection(&self) -> CargoResult<DieselPooledConn> {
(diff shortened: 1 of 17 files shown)


jtgeibel:
And of course I was forgetting to use `--all-targets` when testing against stable locally.  Oops.

Since the renamed lint is only used in tests, I've added a hack that can be removed after the next stable release.

kzys:
Do you need callgrind.out.11995 and stats.txt?

jtgeibel:
> Do you need callgrind.out.11995 and stats.txt?

Good catch, I've removed those files.

kzys:
Thanks. Looks good to me.

PR #1739, opened 2019-05-03
Fix import for serializeQueryParams from ember-fetch.

smarnach:
The Github login is currently broken. The bug was introduced by upgrading ember-fetch in #1733.

This fix imports `serializeQueryParams` from its new location, which fixes the Github login for me. To verify, I followed the instructions in `docs/CONTRIBUTING.md`.

Clearly test coverage is insufficient. I'm happy to add tests as desired, but would like to get feedback what kind of test you suggest first.


locks:
Thanks for the fix! I'm merging it as it is currently broken and we can work on the tests as a follow up.

PR #1753, opened 2019-05-23
Fix yank test

carols10cents:
This test wasn't set up correctly, so the error message that was being checked wasn't quite right.

Also, [based on this comment](https://github.com/rust-lang/crates.io/pull/1697#issuecomment-478740807) it seems like the names of these tests weren't as descriptive as they could be-- the previous test covers the success case of an owner being able to yank, and the next text checks that nonowners aren't allowed to yank.


jtgeibel:
CI should be green on beta now.

@bors r+

PR #1756, opened 2019-05-30
Run `cargo update`

jtgeibel:
I have several local branches updating dependencies and addressing any
breaking changes (towards closing out #1265).  I'd like to avoid
lockfile merge conflicts between those branches so I'm bumping all
semver compatible versions in this PR.

I'd like to let this bake in production for at least a few days before
opening PRs bumping major versions in `Cargo.toml`.

If this approach is too aggressive, or needs to be reverted, I can open
the other PRs more serially and look into breaking this upgrade into
smaller pieces.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,33 +15,36 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.6.3"
+version = "0.6.10"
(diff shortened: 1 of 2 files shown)


sgrif:
I've reverted this on master, it caused segfaults when deployed to production

PR #1757, opened 2019-05-31
Remove duplicate calls to `db_conn` in a single endpoint

jtgeibel:
These endpoints are obtaining two separate connections for each
request.  If there is a backlog of requests, then these endpoints will
wait in line multiple times for a database connection.

Any request with an auth token or cookie will also obtain and drop a
connection in the `CurrentUser` middleware.  We could consider caching
and reusing this database connection as well.

### DIFF
--- a/src/controllers/krate/follow.rs
+++ b/src/controllers/krate/follow.rs
@@ -3,16 +3,16 @@
 use diesel::associations::Identifiable;
 
 use crate::controllers::prelude::*;
+use crate::db::DieselPooledConn;
 use crate::models::{Crate, Follow};
 use crate::schema::*;
 
-fn follow_target(req: &mut dyn Request) -> CargoResult<Follow> {
+fn follow_target(req: &dyn Request, conn: &DieselPooledConn<'_>) -> CargoResult<Follow> {
     let user = req.user()?;
-    let conn = req.db_conn()?;
     let crate_name = &req.params()["crate_id"];
     let crate_id = Crate::by_name(crate_name)
         .select(crates::id)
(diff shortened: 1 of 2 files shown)


smarnach:
I agree that it feels "unfair" if a request needs to wait for a database connection multiple times, but I can't really tell what metric we are improving by reducing the number of connections obtained per request. The 99th percentile of latency, maybe? Do we have monitoring for that? Is the number of connections in the pool really a bottleneck that throttles processing in practice? If so, do we need to increase the pool size, rather than mitigating the impact of requests queuing up for database connections?

Retrieving a connection from the pool is a very fast operation. The connection pool already _is_ a cache for database connections. I would be reluctant to add the complexity of another level of cache per request unless we have data to prove that it is worth the trouble.

That said, the changes in this particular PR look good to me. They _may_ improve latency in some edge cases, and they don't really add complexity, so we can just as well merge them. I tested the endpoints manually for this branch merged into master, and they should also be covered by the tests.

jtgeibel:
I agree, this is a micro-optimization.  In production we did see an issue with a noisy neighbor that was saturating network activity, causing database requests to slow down.  We ended up with 50 requests/threads contending for ~25 database connections (and likely with more requests queued in nginx).  While I don't think this contention made the situation much worse, it does seem a bit odd to return a connection to the pool only to request a new one from the pool a few cycles later.

On the other hand, I expect that there are some places where we should be dropping the connection and obtaining a new one later.  Basically anywhere we block on an external service like GitHub or S3, allowing the connection to be reused rather than sit idle.

@bors r=smarnach

PR #1758, opened 2019-06-04
Return correct download link for requests for non-standard crate names.

### LINKED ISSUE #1687: Wrong redirect URL given for a download with -/_ canonicalization
I was poking around today at a crate I was working on and was confused that after publishing I was getting a 403 for the tarball for the crate that I wanted to manually inspect. I typed out everything below manually so the bug is definitely on me (not through Cargo or any tooling or anything like that), but figured it was still worth mentioning!

Specifically I was interested in the [`compiler_builtins`](https://crates.io/crates/compiler_builtins) crate and forgot whether it was `_` or `-` so I typed out:

```
$ curl -L https://crates.io/api/v1/crates/compiler-builtins/0.1.8/download | tar xzf -
```

but that gave me some error about it not being gzip. Next I ran `curl -I` and it prints out:

```
$ curl -I https://crates.io/api/v1/crates/compiler-builtins/0.1.8/download | grep Location
Location: https://static.crates.io/crates/compiler-builtins/compiler-builtins-0.1.8.crate
```

turns out this URL doesn't exist! It should actually be:

```
$ curl -I https://crates.io/api/v1/crates/compiler_builtins/0.1.8/download | grep Location
Location: https://static.crates.io/crates/compiler_builtins/compiler_builtins-0.1.8.crate
```

Is this intended to work? If not that's totally fine, but I figure if it was intended to work may be good to know!

smarnach:
Fixes #1687.

Crates are uploaded to S3 under the name they were first published as, but the download endpoint always uses the name as written in the request. If these names differ in their use of dashes vs. underscores, the download endpoint returns an invalid link.

To avoid adding another database request for every single crate download, this fix changes the database request that updates the download count to also return the original crate name. We will need to confirm that the performance impact of this change is negligible.

### DIFF
--- a/src/controllers/version/downloads.rs
+++ b/src/controllers/version/downloads.rs
@@ -18,13 +18,13 @@ pub fn download(req: &mut dyn Request) -> CargoResult<Response> {
     let crate_name = &req.params()["crate_id"];
     let version = &req.params()["version"];
 
-    increment_download_counts(req, crate_name, version)?;
+    let crate_name = increment_download_counts(req, crate_name, version)?;
 
     let redirect_url = req
         .app()
         .config
         .uploader
-        .crate_location(crate_name, version);
+        .crate_location(&crate_name, version);
 
     if req.wants_json() {
         #[derive(Serialize)]
@@ -39,7 +39,8 @@ pub fn download(req: &mut dyn Request) -> CargoResult<Response> {
 
 /// Increment the download counts for a given crate version.
 ///
-/// Returns an error if we could not load the version ID from the database.
+/// Returns the crate name as stored in the database, or an error if we could
+/// not load the version ID from the database.
 ///
 /// This ignores any errors that occur updating the download count. Failure is
 /// expected if the application is in read only mode, or for API-only mirrors.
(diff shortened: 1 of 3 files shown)


smarnach:
Looks like the query plan remains virtually the same with this change. Old version:
```
                                                               QUERY PLAN                                                                
-----------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.15..6.69 rows=1 width=4) (actual time=0.087..0.090 rows=1 loops=1)
   ->  Nested Loop  (cost=0.15..13.23 rows=2 width=4) (actual time=0.084..0.084 rows=1 loops=1)
         Join Filter: (versions.crate_id = crates.id)
         ->  Index Scan using index_version_num on versions  (cost=0.15..12.18 rows=2 width=8) (actual time=0.035..0.036 rows=1 loops=1)
               Index Cond: ((num)::text = '0.1.0'::text)
         ->  Materialize  (cost=0.00..1.02 rows=1 width=4) (actual time=0.044..0.044 rows=1 loops=1)
               ->  Seq Scan on crates  (cost=0.00..1.02 rows=1 width=4) (actual time=0.039..0.039 rows=1 loops=1)
                     Filter: (replace(lower((name)::text), '-'::text, '_'::text) = 'crate_with_dashes'::text)
                     Rows Removed by Filter: 1
 Planning Time: 0.802 ms
 Execution Time: 0.151 ms
```
New version:
```
                                                               QUERY PLAN                                                                
-----------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.15..6.69 rows=1 width=36) (actual time=0.091..0.094 rows=1 loops=1)
   ->  Nested Loop  (cost=0.15..13.23 rows=2 width=36) (actual time=0.088..0.089 rows=1 loops=1)
         Join Filter: (versions.crate_id = crates.id)
         ->  Index Scan using index_version_num on versions  (cost=0.15..12.18 rows=2 width=8) (actual time=0.036..0.036 rows=1 loops=1)
               Index Cond: ((num)::text = '0.1.0'::text)
         ->  Materialize  (cost=0.00..1.02 rows=1 width=36) (actual time=0.047..0.048 rows=1 loops=1)
               ->  Seq Scan on crates  (cost=0.00..1.02 rows=1 width=36) (actual time=0.037..0.037 rows=1 loops=1)
                     Filter: (replace(lower((name)::text), '-'::text, '_'::text) = 'crate_with_dashes'::text)
                     Rows Removed by Filter: 1
 Planning Time: 0.645 ms
 Execution Time: 0.173 ms
```

PR #1759, opened 2019-06-05
Show right-hand column for yanked versions to crate owners.

### LINKED ISSUE #1660: No way to manage ownership of a crate through the UI if all versions have been yanked
Bug report received via email.

smarnach:
Fixes #1660.

Without this change, the page for a yanked version looks like this:

image

With this change applied, crate owners see this page instead:

image

### DIFF
--- a/app/controllers/crate/version.js
+++ b/app/controllers/crate/version.js
@@ -33,6 +33,11 @@ export default Controller.extend({
     isOwner: computed('crate.owner_user', 'session.currentUser.id', function() {
         return this.get('crate.owner_user').findBy('id', this.get('session.currentUser.id'));
     }),
+    notYankedOrIsOwner: computed('model', 'crate.owner_user', 'session.currentUser.id', function() {
+        return (
+            !this.get('model').yanked || this.get('crate.owner_user').findBy('id', this.get('session.currentUser.id'))
+        );
+    }),
 
     sortedVersions: readOnly('crate.versions'),
 
--- a/app/templates/crate/version.hbs
+++ b/app/templates/crate/version.hbs
@@ -51,20 +51,18 @@
     </div>
 </div>
 
-{{#if currentVersion.yanked}}
-    <div class='crate-info'>
-        <div>
-        <p>This crate has been yanked, but it is still available for download for
-        other crates that may be depending on it.</p>
-        <p>You may wish to
-          {{#link-to 'crate.versions' crate}}view all versions{{/link-to}} to find one that
-          has not been yanked.
-        </p>
-        </div>
-    </div>
-{{else}}
 <div class='crate-info'>
     <div class='docs'>
+        {{#if currentVersion.yanked}}
+        <p>
+            This crate has been yanked, but it is still available for download for other crates that
+            may be depending on it.
+        </p>
+        <p>
+            You may wish to {{#link-to 'crate.versions' crate}}view all versions{{/link-to}} to find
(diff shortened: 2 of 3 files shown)


app/templates/crate/version.hbs
```diff
@@ -99,7 +97,9 @@
                 </div>
             {{/if}}
         {{/if}}
+        {{/if}}
```
sgrif: Looks like indentation might be off here

smarnach: The indendation was inconsistent to begin with, so I prioritized diff readability. I fixed the indentation now – please take another look.

app/controllers/crate/version.js
```diff
@@ -33,6 +33,11 @@ export default Controller.extend({
     isOwner: computed('crate.owner_user', 'session.currentUser.id', function() {
         return this.get('crate.owner_user').findBy('id', this.get('session.currentUser.id'));
     }),
+    notYankedOrIsOwner: computed('model', 'crate.owner_user', 'session.currentUser.id', function() {
```
efx: could we add unit tests for this computed?

smarnach: Sure we could, but I'm not familiar with Ember or Mirage, so I'm not sure what's the best way to do this. I've added an attempt as a new commit that tries to use Mirage to set up a new crate and a version. The test passes in its current form, but has a few problems:

* I was only able to use the Mirage factories when running the whole application with `setupApplicationTest()`. This seems unnecessary and makes the test much slower than it should be.
* Testing the other cases (yanked crate, with the user being owner or not) requires also faking an active session and user accounts. I couldn't find any existing infrastructure to do so in a unit test.

We may be able to test this with an acceptance test instead, but at least for me this would involve significant effort. Maybe someone more experienced with Ember could add this test?

efx: Apologies @smarnach, I should've added a bit more direction here. Ember can generate default unit tests that should avoid needing to use mirage. The problems you point out highlight the many concerns of acceptance / integration testing I hoped to avoid by suggesting a unit test. I'll open a PR against your branch with a simple unit test.

efx: I'm doing this in my fork, but you can create test files using `ember generate` commands. 

```sh
ember generate controller-test crates/version
```

efx: I was lazy and didn't fork your copy, but here is the working example of the unit test: https://github.com/rust-lang/crates.io/compare/master...efx:unit-test-controller?expand=1#diff-a30cef8546b1174ebf3c37a550e2357f

smarnach: Thanks a lot, this looks a lot more reasonable than my attempt! I guess I should learn a bit more about Ember before doing any more work on the frontend. :)

I've copied your test to this branch (and force-pushed, since my attempt isn't worth being preserved in the Git history).

efx: certainly @smarnach. Eh, learning on the go seems most efficient 😄. Grand!

kzys:
Looks good to me.

smarnach:
@locks @kzys Is there anything I can do to move this forward? It would be great if we could get this merged.

efx:
@locks any objections to merging this?

carols10cents:
The text wrapping is different now if you're not an owner, but I'm fine with it.

Screen Shot 2019-08-30 at 4 46 00 PM

@bors: r+

Issue #1641, opened 2019-02-28
Display build-dependencies on crate page


olson-dan:
I noticed that dependencies and dev dependencies are currently listed on individual crate pages, but build dependencies are not.

Cargo's build-dependencies often pull in external programs or libraries (gcc, cmake, pkg-config, libclang) and these programs are often missing on the machines of users attempting to build the crates.

It'd be helpful to list the build dependencies for each crate so that users have an idea of what 3rd party software must be installed to use it.  This could also help people vet crates for security purposes, as build.rs is not sandboxed in any way as far as I am aware.

critical-path:
May I try to address this issue?  Thank you!

carols10cents:
So I was just looking into this more, and looking at the PR https://github.com/rust-lang/crates.io/pull/1761 that was closed (sorry for the delay on reviewing!)

It turns out build dependencies *are* currently being displayed; they're listed as regular dependencies though, rather than being split out into their own category.

For example, https://crates.io/crates/lapin currently has listed:

```
Dependencies

    amq-protocol ^3.0
    amq-protocol-codegen ^3.0
    crossbeam-channel ^0.3
    log ^0.4
    mio ^0.6
    parking_lot ^0.9
    serde_json ^1.0
    futures-core-preview = 0.3.0-alpha.19 optional

Dev-Dependencies

    env_logger ^0.7
    futures-executor-preview ^0.3.0-alpha.19
    futures-util-preview ^0.3.0-alpha.19
    tcp-stream ^0.7
```

But looking at [the JSON returned from the API](https://crates.io/api/v1/crates/lapin/0.28.1/dependencies), amq-protocol is a normal dependency and amq-protocol-codegen is a build dependency, for example.

So the code in https://github.com/rust-lang/crates.io/pull/1761 needs to exclude build dependencies from the regular dependency display as well as adding the new build dependency section, and the test should assert that the build dependency only appears under the appropriate section.

Mordorreal:
Can I try to fix this?

smarnach:
@Mordorreal Sure, feel free to go ahead! You can ask questions here, or you can send a PR.

PR #1761, opened 2019-06-06
Display build-dependencies

critical-path:
@sgrif 

This is a first attempt at addressing #1641.  Please let me know whether I am on the right track.  I would be happy to make changes.  Thank you!


Turbo87:
can look into writing a test for this to make sure that we don't accidentally break this functionality?

critical-path:
Please let me know if you require any changes.  Thank you!

Issue #1625, opened 2019-02-14
Add dropdown box to allow selecting more results per page


therealprof:
As discussed at the RaH it would be really handy to allow displaying more results per page to make it easier to find stuff when browsing through a largish category.

@sgrif mentioned he would be okay with allowing changing the display to up to 100 entries per page to keep load on the servers at an acceptable level.

carols10cents:
As discussed in https://github.com/rust-lang/crates.io/pull/1762, we decided to increase the number of crates on one page from 10 to 50, rather than offering a dropdown. The increase was implemented in #1883, and I'm going to be deploying it in a little bit, so I'm closing this issue. Thanks!

PR #1762, opened 2019-06-09
Add dropdown box to allow selecting more results per page

daubaris:
#1625

### DIFF
--- a/app/controllers/crates.js
+++ b/app/controllers/crates.js
@@ -25,4 +25,16 @@ export default Controller.extend(PaginationMixin, {
             return 'Alphabetical';
         }
     }),
+
+    resultCount: computed('per_page', function() {
+        if (this.per_page === 10) {
+            return 10;
+        } else if (this.per_page === 20) {
+            return 20;
+        } else if (this.per_page === 50) {
+            return 50;
+        } else {
+            return 100;
+        }
+    }),
 });
--- a/app/routes/crates.js
+++ b/app/routes/crates.js
@@ -5,6 +5,7 @@ export default Route.extend({
         letter: { refreshModel: true },
         page: { refreshModel: true },
         sort: { refreshModel: true },
+        per_page: { refreshModel: true },
     },
 
     model(params) {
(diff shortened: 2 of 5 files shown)


sgrif:
Rather than adding a dropdown for this, could we just up the default sent by the frontend?

daubaris:
You mean to just increase the `per_page` value to 100 to be the default value instead of 10?
https://github.com/rust-lang/crates.io/blob/922a7c92abbb6e750f4d61595ac13d88c8b63fbf/app/controllers/crates.js#L11

sgrif:
I'd go with either 20 or 50 instead of 100, but yes

daubaris:
Wouldn't be better to leave the dropdown just to provide flexibility for the user or will this cause any performance issues? Just curious.

sgrif:
I'm not sure that a dropdown to configure this really adds that much
value, but I'm open to being convinced otherwise. Does anyone ever use
these to lower the number of records displayed?

On Fri, Jun 14, 2019 at 1:28 PM daubaris <notifications@github.com> wrote:
>
> Wouldn't be better to leave the dropdown just to provide flexibility for the user or will this cause any performance issues? Just curious.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub, or mute the thread.



-- 

Thanks,
Sean Griffin

kzys:
FYI, https://metacpan.org/search?size=20&q=foo is having a way to change the page size. https://rubygems.org/search?utf8=✓&query=foo, https://www.npmjs.com/search?q=foo and https://pypi.org/search/?q=foo don't provide the one.

Personally I'd like to increase the default page count to keep it simple.

locks:
Increasing the default page seems like a nice small improvement!

sgrif:
@daubaris Are you interested in amending this to increase the default?

carols10cents:
I'm sorry we've let this PR sit for so long! I'm going to close it since it's gotten out of date and since there's https://github.com/rust-lang/crates.io/pull/1883 that implemented the discussed modifications. Thank you!

PR #1763, opened 2019-06-10
Include `next_page` and `prev_page` metadata on search

sgrif:
As discussed in the May 30th team meeting
(https://discordapp.com/channels/442252698964721669/448525639469891595/583749739166695444),
this is the first step in deprecating/removing numbered pagination for
the "all crates" endpoint. The remaining steps are:

- Open issues/PRs on bots which have set a user agent that lets us
  identify them, have them start following this link
- Change our frontend to show "next/prev page" links on the all crates
  page
- Stop returning the "total" meta item when the next/prev page links
  will be cursor based
- Switch the links over to cursor based pagination when applicable
- Error when given pagination parameters that would cause very high
  offsets

Since the problem we're addressing only occurs when the offset is
greater than ~18k, the only place that we currently need to worry about
this is "all crates". There isn't currently any filter you can apply
that will return enough rows for this to be a problem (`?letter=` is
the most likely, but honestly that is probably the least useful
page/endpoint we have and it might just be worth removing it. Either way
AFAIK it's not hit by crawlers and humans aren't visiting page 900 in
their browsers, so even if we do end up with enough crates for the
offset to possibly be too high, it's unlikely anyone will ever hit it
there).

I'm not entirely sure how I want to structure the final logic for "give
me the next/previous page". I've just stuck it at the bottom of the
search function for now. We'll want something more structured in the
long term, but I'd like to see what this looks like with the cursor
based option before pulling it out into a more general abstraction.

### DIFF
--- a/src/controllers.rs
+++ b/src/controllers.rs
@@ -14,8 +14,8 @@ mod prelude {
     use std::collections::HashMap;
     use std::io;
 
-    use serde::Serialize;
     use indexmap::IndexMap;
+    use serde::Serialize;
     use url;
 
     pub trait RequestUtils {
--- a/src/controllers.rs
+++ b/src/controllers.rs
@@ -15,13 +15,14 @@ mod prelude {
     use std::io;
 
     use serde::Serialize;
+    use indexmap::IndexMap;
     use url;
 
     pub trait RequestUtils {
         fn redirect(&self, url: String) -> Response;
 
         fn json<T: Serialize>(&self, t: &T) -> Response;
-        fn query(&self) -> HashMap<String, String>;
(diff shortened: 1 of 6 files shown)


src/controllers.rs
```diff
@@ -31,7 +32,7 @@ mod prelude {
             crate::util::json_response(t)
         }
 
-        fn query(&self) -> HashMap<String, String> {
+        fn query(&self) -> IndexMap<String, String> {
```
jtgeibel: `IndexMap` defaults to the `RandomState` hasher from `std`, so this looks fine from a DOS perspective.

PR #1772, opened 2019-06-20
Fix checks and consistency in the API endpoints to manage ownership.

### LINKED ISSUE #1583: It's possible to permanently lock yourself out of crate ownership through teams
It's possible to lock yourself out of crate ownership by adding a team owner, and then removing yourself. Since teams cannot add or remove owners, there is no way to manage ownership from then on.

I have [several crates](https://crates.io/teams/github:hobbyistrust:owners) affected by this.

smarnach:
Fixes #1583.

The API endpoints to add and remove owners for a crate has multiple issues.

* The database interactions are not wrapped in a transaction, making them prone to race conditions.

* The lack of a transaction can also result in partially applied operations, e.g. when adding or
  removing multiple owners at once.

* The check whether the last owner was removed only triggers if the _original_ list of owners had a
  length of exactly one. This is wrong in at least two cases:

  * Teams do not have permission to manage ownership, so they should not be counted.

  * If multiple users are removed at once, we can end up with no owners even when there originally
    was more than one user.

This change fixes all mentioned problems, and adds tests for all bullet points except for the first
one (which can't be tested with reasonable effort).

### DIFF
--- a/src/tests/util.rs
+++ b/src/tests/util.rs
@@ -522,8 +522,8 @@ impl MockTokenUser {
         F: Fn(&MockTokenUser, &str, &[u8]) -> Response<OkBool>,
     {
         let url = format!("/api/v1/crates/{}/owners", krate_name);
-        let body = format!("{{\"owners\":[\"{}\"]}}", owners.join("\", \""));
-        method(&self, &url, body.as_bytes())
+        let body = json!({ "owners": owners }).to_string();
+        method(&self, &url, body.to_string().as_bytes())
     }
 
     /// Add a user as an owner for a crate.
--- a/src/tests/util.rs
+++ b/src/tests/util.rs
@@ -497,18 +497,33 @@ impl MockTokenUser {
         &self.token
     }
 
-    /// Add to the specified crate the specified owner.
-    pub fn add_named_owner(&self, krate_name: &str, owner: &str) -> Response<OkBool> {
-        let url = format!("/api/v1/crates/{}/owners", krate_name);
-        let body = format!("{{\"users\":[\"{}\"]}}", owner);
-        self.put(&url, body.as_bytes())
+    /// Add to the specified crate the specified owners.
+    pub fn add_named_owners(&self, krate_name: &str, owners: &[&str]) -> Response<OkBool> {
+        self.modify_owners(krate_name, owners, Self::put)
     }
 
-    /// Remove from the specified crate the specified owner.
+    /// Add a single owner to the specified crate.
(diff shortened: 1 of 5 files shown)


sgrif:
Can we split the transaction wrapping into its own commit? I don't anticipate any problems with this, but in the event we need to roll something back it's helpful to have each problem solved separately

smarnach:
@sgrif If I split this up, the transaction wrapping will necessarily be the first commit, since everything else depends on it. This means we can't roll back the transaction wrapping and keep any of the other changes, if this is what you have in mind, but we will still be able to roll back some of the other fixes and keep the transaction wrapping.

sgrif:
Yes, having the transaction wrapping first was what I had in mind

smarnach:
@sgrif I've split this up in three commits (to improve diff readability), and made sure the tests pass for each step (to facilitate partial rollbacks). Please take another look.

src/tests/util.rs
```diff
@@ -497,18 +497,33 @@ impl MockTokenUser {
...
+
+    fn modify_owners<F>(&self, krate_name: &str, owners: &[&str], method: F) -> Response<OkBool>
+    where
+        F: Fn(&MockTokenUser, &str, &[u8]) -> Response<OkBool>,
+    {
         let url = format!("/api/v1/crates/{}/owners", krate_name);
-        let body = format!("{{\"users\":[\"{}\"]}}", owner);
-        self.delete_with_body(&url, body.as_bytes())
+        let body = format!("{{\"owners\":[\"{}\"]}}", owners.join("\", \""));
+        method(&self, &url, body.as_bytes())
```
jtgeibel: It's probably worth switching to the `json!` macro here.

```rust
let body = json!({ "owners": owners });
method(&self, &url, body.to_string().as_bytes())
```

smarnach:
@jtgeibel Thanks for the review! Using the `json!` macro is certainly a lot more readable, and the changes are very low risk, so I can include them here.

I noticed another race condition I haven't fixed yet, though. When adding new owners, the code first reads the list of all current owners from the DB, then for each owner checks whether it occurs in the list of current owners, then creates a new owner invitation. If there already _was_ an owner invitation for an owner, and it gets accepted right after the code has checked for the list of current owners, we will create a new invitation for a user that is already an owner, which would normally be prohibited.

This race condition could be fixed by using the "serializable" isolation level for the transaction, but then we would also need to add code to retry the transaction if it fails due to a serialization error. I don't think this would be worth it – the above race condition should be exceedingly rare, and even when it gets triggered, the impact is negligible – there will be a spurious invitation, and accepting or rejecting it has no effect, since the invitee already is an owner.

PR #1780, opened 2019-07-09
Remove jQuery from hashchange.js and controllers

kzys:


app/initializers/hashchange.js
```diff
@@ -29,11 +27,17 @@ function hashchange() {
 }
 
 export function initialize() {
-    $(window).on('hashchange', hashchange);
+    if (typeof window.addEventListener === 'undefined') {
```
rwjblue: I think you will also need a `typeof window !== 'undefined'` guard here to work well in FastBoot. AFAICT you _basically_ never need to run this initializer in SSR modes.

kzys: Thanks. Will do.

PR #1784, opened 2019-07-18
Encapsulate the RequestProxy use cases

jtgeibel:
The RequestProxy value is constructed 3 places, overwriting 1 of 2
values.  The fields are made private and 2 new constructors give each
use case a name.

### DIFF
--- a/src/util/request_proxy.rs
+++ b/src/util/request_proxy.rs
@@ -1,25 +1,52 @@
 use std::{io::Read, net::SocketAddr};
 
-use conduit::Request;
+use conduit::{Method, Request};
 use conduit_hyper::semver;
+
+type RequestMutRef<'a> = &'a mut (dyn Request + 'a);
 
 // Can't derive Debug because of Request.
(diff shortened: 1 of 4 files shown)


src/util/request_proxy.rs
```diff
@@ -1,26 +1,53 @@
...
-    pub other: &'a mut (dyn Request + 'a),
-    pub path: Option<&'a str>,
-    pub method: Option<conduit::Method>,
+    other: RequestMutRef<'a>,
+    path: Option<&'a str>,
+    method: Option<conduit::Method>,
+}
+
+impl<'a> RequestProxy<'a> {
+    pub(crate) fn rewrite_path(req: RequestMutRef<'a>, path: &'a str) -> Self {
```
smarnach: You could consider adding a one-sentence doc comment here, and for the other constructor as well.

jtgeibel:
I've added some docs to the extracted logic.  Thanks @smarnach!

@bors r+

PR #1787, opened 2019-07-21
Adjust branding to reflect what crates.io actually does

### LINKED ISSUE #1777: <head> description of crates.io confusingly refers to Cargo
When you access https://crates.io, you find that the <title> says "Cargo: packages for Rust" and the <meta> description says "cargo is the package manager and crate host for rust" whereas at the top of the page it says "crates.io — Rust Package Registry".

I believe that the package host (crates.io) and the package manager (cargo) should be now distinguished clearly (although they can be linked together), so I would like the <head> description of https://crates.io to be properly updated.

rye:
To be clear that crates.io is a Rust package registry&mdash;in fact the _default_ crate host for Rust---this commit changes the `name`, `short_name`, and `description` fields in the manifest to reflect the nature of the project.  crates.io isn't really "Cargo", so it seems logical to me to not mention much about it, at least in the metadata for the site.

I also more explicitly mention that `crates.io` is the "default" package registry to be a bit more neutral; it's the baked-in default in Cargo, but there are others out there. I got the branding "Rust Package Registry" from the tagline in the visible top bar on the landing page.

Also, update the various acceptance tests that I found floating around.

Should resolve #1777, I think.

### DIFF
--- a/tests/acceptance/crates-test.js
+++ b/tests/acceptance/crates-test.js
@@ -1,4 +1,4 @@
-import { module, test } from 'qunit';
+yimport { module, test } from 'qunit';
 import { setupApplicationTest } from 'ember-qunit';
 import { click, currentURL, visit } from '@ember/test-helpers';
 import a11yAudit from 'ember-a11y-testing/test-support/audit';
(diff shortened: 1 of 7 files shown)


tests/acceptance/crates-test.js
```diff
@@ -1,4 +1,4 @@
-import { module, test } from 'qunit';
+yimport { module, test } from 'qunit';
```
smarnach: There's a stray "y" here, causing CI to fail.

rye: Oops, thanks for the catch! Pushed a fix.

smarnach:
@jtgeibel This looks rather uncontroversial to me – it makes the branding in the title match the branding we already have at the top of the page. The code changes look fine to me, so maybe we can merge this?

kzys:
Looks good to me. +1 for merging.

Issue #788, opened 2017-06-20
Crates.io returns mostly 404


kud1ing:
Maybe this is a duplicate of other 404-issues here, i am not sure.

`awesome-rust` uses `awesome_bot` to check links.
Unfortunately many (all?) links to crates.io return 404: https://travis-ci.org/rust-unofficial/awesome-rust

`curl https://crates.io/keywords/cassandra` receives a 404, too.
I am a web noob. Is there somwething that needs to be configured on the server or client?

carols10cents:
I think we are doing something nonstandard, but I'm not sure what. Until we figure this out, a workaround is specifying HTML as the content type you want:

```
curl -v 'https://crates.io/crates/assert_approx_eq' -H 'Accept: text/html'
```

returns a 200 for me. Hope that helps!

kud1ing:
Please give us a heads up when this is resolved.
`crates.io` was whitelisted in https://github.com/rust-unofficial/awesome-rust/pull/310
I'd like to revert this finally.

Nemo157:
Yeah, this is caused by https://github.com/rust-lang/crates.io/blob/a20eb96877d32bff6b5b9ac8ea3dfc497b5351b9/src/dist.rs#L38-L52

Interestingly requesting _any_ URL with `Accept: text/html` will return 200 since the server doesn't know what the valid routes are (e.g. `curl -H 'Accept: text/html' -I https://crates.io/foo/bar`)

I don't think there's any way to have a bot validate URLs on crates.io without executing the javascript and checking if it loads the page successfully. At least, while crates.io does client-side only rendering, if server-side rendering of at least the initial page is ever added that should then know whether the URL being requested is valid or not.

locks:
FYI, you can track server-side rendering efforts at https://github.com/rust-lang/crates.io/pull/819.

AaronFriel:
This is still not fixed.

![image](https://user-images.githubusercontent.com/788800/34024765-a455a552-e111-11e7-9e82-15c780a0e941.png)

carols10cents:
@AaronFriel Yes, that's why this issue is still open.

AaronFriel:
Would a commit altering the `wants_html` test to include wildcards be accepted? I'm just not sure why this has languished and I'm guessing there is something I'm missing.

carols10cents:
Yes, I am not sure of the scope of the solution needed or the effects of various solutions on the frontend and backend-- please give it a try and let me know if you have questions!

[...]

Issue #556, opened 2017-02-13
Missing packages should return HTTP 404


Rufflewind:
If `https://crates.io/crates/nonexistent_package` does not refer to a valid package, it should respond with HTTP 404 to avoid misleading crawlers and/or browsers into thinking that it's a valid page.

onur:
It is responding 404:

```
$ curl -D- https://crates.io/crates/nonexistent_package
HTTP/1.1 404 Not Found
Connection: keep-alive
Server: nginx
Date: Mon, 13 Feb 2017 12:34:09 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 35
Set-Cookie: cargo_session=--3Mz45cH3itfVkmY41+nVjWRm9+0=; HttpOnly; Secure; Path=/
Via: 1.1 vegur

{"errors":[{"detail":"Not Found"}]}
```

Rufflewind:
@onur It seems to be dependent on the `Accept` header:

    $ curl -v -H 'Accept: text/html' https://crates.io/crates/nonexistent_package
    < …
    < HTTP/1.1 200 OK
    < …

carols10cents:
Closing as a duplicate of #788.

PR #1788, opened 2019-07-25
Serve Ember bootstrap HTML for all non-api requests

jtgeibel:
The backend no longer checks for an "html" in the `Accept` header.
With the exception of 3 session related routes, all paths not starting
with "/api" will be redirected to the static Ember bootstrap page.

As a result of this change all non-api requests that don't contain
"html" in the `Accept` header will now unconditionally return `200`,
rather than `404`.  In a sense, this expands the scope of #556 to all
requests, not just those that set the header.  It also inverts the
problem described in #788, effectively turning it into a duplicate
of #556.

Fixes: #163

### DIFF
--- a/src/middleware/ember_index_rewrite.rs
+++ b/src/middleware/ember_index_rewrite.rs
@@ -25,21 +25,20 @@ impl AroundMiddleware for EmberIndexRewrite {
 
 impl Handler for EmberIndexRewrite {
     fn call(&self, req: &mut dyn Request) -> Result<Response, Box<dyn Error + Send>> {
-        // If the client is requesting html, then we've only got one page so
-        // rewrite the request.
-        let wants_html = req
-            .headers()
-            .find("Accept")
-            .map(|accept| accept.iter().any(|s| s.contains("html")))
-            .unwrap_or(false);
-        // If the route starts with /api, just assume they want the API
-        // response and fall through.
-        let is_api_path = req.path().starts_with("/api");
         let handler = self.handler.as_ref().unwrap();
-        if wants_html && !is_api_path {
-            handler.call(&mut RequestProxy::rewrite_path(req, "/index.html"))
-        } else {
+        let is_backend_path = match req.path() {
+            // Special case routes used for authentication
+            "/authorize" | "/authorize_url" | "/logout" => true,
+            // Paths starting with `/api` are intended for the backend
+            path if path.starts_with("/api") => true,
+            _ => false,
+        };
+
+        if is_backend_path {
             handler.call(req)
+        } else {
+            // Serve static Ember page to bootstrap the frontend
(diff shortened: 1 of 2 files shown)


src/middleware/ember_index_rewrite.rs
```diff
@@ -25,21 +25,20 @@ impl AroundMiddleware for EmberIndexRewrite {
...
-        // If the route starts with /api, just assume they want the API
-        // response and fall through.
-        let is_api_path = req.path().starts_with("/api");
         let handler = self.handler.as_ref().unwrap();
-        if wants_html && !is_api_path {
-            handler.call(&mut RequestProxy::rewrite_path(req, "/index.html"))
-        } else {
+        let is_backend_path = match req.path() {
+            // Special case routes used for authentication
+            "/authorize" | "/authorize_url" | "/logout" => true,
```
smarnach: Having to special-case these is a bit unfortunate. Would it be possible to instead do something along these lines:

* Try to serve the request with the backend.
    * On success, return the result.
    * If the backend returns a 404, check whether the path starts with `/api`.
        * If so, return a 404
        * Otherwise redirect to `index.html`.

Basically give the backend a chance to serve any request first, so we don't have to special-case any routes outside of the `/api` prefix. I haven't looked into the details, so I'm not sure whether this is feasible in this particular case, but it's roughly what I would do in other web frameworks.

sgrif: Is there a reason we can't just move these routes under the `/api` namespace? There's no reason we need the endpoint hit by the front end to be the same URL that the browser is currently on.

smarnach: If we move all backend routes into `/api`, we can also make nginx decide where to send requests:

* `/assets/` → serve from file system
* `/api/` → proxy to backend
* everything else → serve `index.html`

There isn't really a reason why requests for `index.html` should go to the backend, since the file is actually part of the frontend.

jtgeibel: > Is there a reason we can't just move these routes under the `/api` namespace?

I'd be fine with moving these.  The reason I didn't move them at this point is that they are involved in session management, and at least one endpoint needs to be changed in parallel with changes to GitHub oauth settings so that the oauth callback from GitHub hit the right endpoint.

It would probably be best if we add duplicate routes under `/api`, deploy, change GitHub oauth settings, and then land a commit removing the old routes.

kzys: +1 for moving them under /api. That also makes enabling FastBoot on the routes easier.

jtgeibel:
I'm closing this PR for now, as making this change will affect how crawlers see the site and a change to this logic may not make sense until we enable fastboot.

PR #1798, opened 2019-08-07
Move `update-downloads` to a background job

### LINKED ISSUE #1797: Monitor the update-downloads script
We've had multiple cases where the update-downloads script was erroring and no one caught it until someone mentioned their crate's download counts weren't updating days later.

Someone should get paged if this happens, which we could detect by:

- The update-downloads script aborting the transaction with an error
- The number of `version_downloads` records with `processed = 'f'` getting over some threshold

I'm not sure which would fit with our existing paging infrastructure best.

sgrif:
This replaces the `update-downloads` binary with a background job, and a
binary that is used to queue up a given job, which we will run from
Heroku scheduler. This accomplishes 2 things:

- It makes it easier to write tasks that need to run periodically (e.g.
  cleaning up stale rate limit buckets), since we don't need to create a
  new standalone binary.
- `update_downloads` and any future recurring tasks will automatically
  get monitoring if they fail, since we are already monitoring for
  background jobs not being successfully run.

Right now the intent is to have `enqueue-job update_downloads` get run
periodically by Heroku scheudler (and a similar scheduled task for any
future tasks that are added). Once swirl gains the ability to schedule
jobs to be run at arbitrary points in the future, we could instead have
these jobs re-queue themselves once they complete, and have the cron
task just look to see if any job is queued for each given type, queuing
it if not. That would have a bit less boilerplate, but a lot more
complexity.

Fixes #1797.

### DIFF
--- a/src/bin/enqueue-job.rs
+++ b/src/bin/enqueue-job.rs
@@ -1,0 +1,17 @@
+use cargo_registry::util::{CargoError, CargoResult};
+use cargo_registry::{db, tasks};
+use std::env::args;
+use swirl::Job;
+
+fn main() -> CargoResult<()> {
+    let conn = db::connect_now()?;
+
+    match &*args().nth(1).unwrap_or_default() {
+        "update_downloads" => tasks::update_downloads()
+            .enqueue(&conn)
+            .map_err(|e| CargoError::from_std_error(e))?,
+        other => panic!("Unrecognized job type `{}`", other),
+    };
+
+    Ok(())
+}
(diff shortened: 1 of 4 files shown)


smarnach:
For what it's worth, this looks good to me. I tested the code locally by manually enqueueing a job and seeing the background worker successfully picking it up and updating the download counts.

The command-line parsing and error reporting via panic is rather minimalistic, but it's definitely good enough for this use case.

Incidentally, I worked on a prototype for #630 in the last few days, and did something very similar to allow running the database dumps periodically. I will switch my prototype over to this implementation.

jtgeibel:
LGTM!

@bors r+

PR #1800, opened 2019-08-09
Prototype: Public database dumps

smarnach:
This is an unfinished prototype implementation of the design I proposed to implement #630 (see https://github.com/rust-lang/crates.io/issues/630#issuecomment-510380244 and https://github.com/rust-lang/crates.io/issues/630#issuecomment-511534102). I am submitting this for review to gather some feedback on the basic approach before spending more time on this.

This PR adds a background task to create a database dump. The task can be triggered with the `enqueue-job` binary, so it is easy to schedule in production using Heroku Scheduler.

### Testing instructions

To create a dump:

1. Start the background worker:

        cargo run --bin background-worker

1. Trigger a database dump:

        cargo run --bin enqueue-job dump_db

    The resulting tarball can be found in `./local_uploads/db-dump.tar.gz`.

To re-import the dump

1. Unpack the tarball:

        tar xzf local_uploads/db-dump.tar.gz

1. Create a new database:

        createdb test_import_dump

1. Run the Diesel migrations for the new DB:

        diesel migration run --database-url=postgres:///test_import_dump

1. Import the dump

        cd DUMP_DIRECTORY
        psql test_import_dump < import.sql

(Depending on your local PostgreSQL setup, in particular the permissions for your user account, you may need different commands and URIs than given above.)

### Author's notes

* The background task executes `psql` in a subprocess to actually create the dump. One reason for this approach is its simplicity – the `\copy` convenience command issues a suitable `COPY TO STDOUT` SQL command and streams the result directly to a local file. Another reason is that I couldn't figure out how to do this at all in Rust with a Diesel `PgConnection`. There doesn't seem to be a way to run raw SQL with full access to the result.

* The unit test to verify that the column visibility information in `dump_db.toml` is up to date compares the information in that file to the current schema of the test database. Diesel does not provide any schema reflection functionality, so we query the actual database instead. This test may spuriously fail or succeed locally if you still have some migrations from unmerged branches applied to your test database. On Travis this shouldn't be a problem, since I believe we always start with a fresh database there. (My preferred solution for this problem would be for Diesel to provide some way to introspect the information in `schema.rs`.)

### Remaining work

* [x] Address TODOs in the source code. The most significant one is to update the `Uploader` interface to accept streamed data instead of a `Vec<u8>`. Currently the whole database dump needs to be loaded into memory at once.

* ~~Record the URLs of uploaded tarballs in the database, and provide an API endpoint to download them.~~ Decided to only store latest at a known URL

* [x] Devise a scheme for cleaning up old dumps from S3. The easiest option is to only keep the latest dump.

* [x] Somewhere in the tar file, note the date and time the dump was generated

* [x] Verify that `dump-db.toml` is correct, i.e. that we don't leak any data we don't want to leak. Done via manual inspection. ~~One idea to do so is to reconstruct dumps from the information available via the API and compare to information in a test dump in the staging environment. This way we could verify what additional information will be made public.~~

* [x] The code needs some form of integration test. Idea from #1629: exporting some data, then trying to re-import it in a clean database.

* [x] Implement and document a way of re-importing the dumps to the database, e.g. to allow local testing of crates.io with realistic data.

* [x] Rebase and remove commits containing the first implementation

* [x] Document the existence of this dump, how often it's regenerated, and that only the most recent dump is available (maybe in the crawler policy/crawler blocked error message?)

* [x] Include the commit hash of the crates.io version that created the dump in the tarball


migrations/2019-08-08-191129_add_users_public/up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE users ADD COLUMN public BOOLEAN NOT NULL DEFAULT 'f';
+UPDATE users
+SET public = 't'
+WHERE
+    id IN (
```
sgrif: It seems like we can probably just use this select instead of adding this column to the table, or perhaps make a `public_users` view with this `WHERE` clause? Was there another reason for making this a persistent column?

migrations/2019-08-08-191129_add_users_public/up.sql
```diff
@@ -0,0 +1,9 @@
...
+        SELECT owner_id AS user_id FROM crate_owners WHERE NOT deleted AND owner_kind = 0
+        UNION
+        SELECT published_by as user_id FROM versions
```
sgrif: Is this information otherwise publicly available? We need it for DMCA purposes, but I'd rather not include it in the dump if it can't already be obtained through the API

src/bin/enqueue-job.rs
```diff
@@ -1,17 +1,18 @@
...
 fn main() -> CargoResult<()> {
-    let conn = db::connect_now()?;
-
     match &*args().nth(1).unwrap_or_default() {
-        "update_downloads" => tasks::update_downloads()
-            .enqueue(&conn)
-            .map_err(|e| CargoError::from_std_error(e))?,
-        other => panic!("Unrecognized job type `{}`", other),
-    };
+        "update_downloads" => enqueue(tasks::update_downloads()),
```
sgrif: Could this be `enqueue(job, &conn)` instead? I want to get rid of the need for the `.map_err` eventually, which would mean we could replace the custom function with `use swirl::Job::enqueue;`

src/tasks/dump-db.toml
```diff
@@ -0,0 +1,168 @@
+# Column visibility in database dumps. Only public columns are included in the dump.
+[api_tokens]
+id = "private"
```
sgrif: Do we need to specify `private` here? This should 100% for sure be an allow-list, so everything needs to be private by default anyway.

src/tasks/dump_db.rs
```diff
@@ -0,0 +1,185 @@
...
+    Ok(tarball_name)
+}
+
+fn upload_tarball(tarball: &Path, uploader: &Uploader) -> Result<(), PerformError> {
+    use std::io::Read;
+
+    let client = reqwest::Client::new();
+    let target_name = format!("dumps/{}", tarball.file_name().unwrap().to_str().unwrap());
+    let mut buf = vec![];
+    // TODO: find solution that does not require holding the whole database
```
sgrif: We can probably change `upload` to take `T: Read` to solve this

src/bin/dump-db.rs
```diff
@@ -0,0 +1,91 @@
...
+    let mut args = std::env::args().skip(1);
+    match args.next().as_ref().map(String::as_str) {
+        Some("setup") => db_setup(&conn),
+        Some("teardown") => db_teardown(&conn),
+        Some(_) | None => Err(human("unknown command")),
+    }
+}
+
+fn get_db_dump_credentials() -> CargoResult<(String, String)> {
+    let database_url = url::Url::parse(&env("DUMP_DATABASE_URL"))?;
```
sgrif: Can you clarify why we need a separate env var from `DATABASE_URL` for this?

src/bin/dump-db.rs
```diff
@@ -0,0 +1,91 @@
+//! Helper to set up the database role and permissions for creating database
```
sgrif: I'm a bit confused by this binary. It seems like we're handling visibility when generating the dump anyway? Why do we need to create a new role and permissions, when it's programmatically generated from the same data used to determine what columns we access anyway?

sgrif:
Thanks for working on this! I have a few more thoughts on the points listed in your PR description

> Another reason is that I couldn't figure out how to do this at all in Rust with a Diesel PgConnection. There doesn't seem to be a way to run raw SQL with full access to the result.

Yeah, Diesel doesn't support `COPY FROM` at all. It's very PG specific, and nobody has asked for support yet (thought TBH we could add it pretty easily for this). I am curious about the decision to go with CSV output for this though. My initial reaction is that a PG dump (either in SQL or binary format) would be more convenient for all parties, but now I'm wondering if that's actually true. This might be worth doing some user outreach on.

> This test may spuriously fail or succeed locally if you still have some migrations from unmerged branches applied to your test database. 

Seems fine, I don't think our test suite is expected to pass with random migrations applied, `dotenv 'diesel database reset --database-url=$TEST_DATABASE_URL'` is a pretty common command for me to run.

> Verify the column visibility configuration and filtering rules. One idea to do so is to reconstruct dumps from the information available via the API and compare to information in a test dump in the staging environment. This way we could verify what additional information will be made public.

I'm not sure I'm understanding this item properly, but if your point is to ensure all columns are listed in the TOML file, I think we can do this one of two ways:

- Reflect on the schema (yes, Diesel's code for this isn't public, but if you just need a list of table/column names it's a very simple query) and assert that all results are listed in the TOML file
- Just ignore it, since we don't want things to be included by default anyway. This would mean we would occasionally add a column and forget to include it in the dump, but I'm honestly not sure this is a common enough occurrence to worry about

> Record the URLs of uploaded tarballs in the database, and provide an API endpoint to download them.

> Devise a scheme for cleaning up old dumps from S3. The easiest option is to only keep the latest dump.

I think the easiest option is the best one here, and removes the need for an API endpoint at all -- we can just point people to `static.crates.io/database.dump` and be done with it.

migrations/2019-08-08-191129_add_users_public/up.sql
smarnach: The `versions.published_by` column is publicly available, see e.g. https://crates.io/api/v1/crates/rand/0.7.0. The `versions_published_by` _table_, on the other hand, is private, and is marked as private in the configuration.

src/tasks/dump-db.toml
smarnach: The idea is to have visibility information for all columns. There is a unit test that compares the schema of the test database to the configuration in this file, so when a new column is added, people are explicitly forced to decide whether it should be public or private. I arrived at this design in response to https://github.com/rust-lang/crates.io/issues/630#issuecomment-510642673.

src/bin/dump-db.rs
smarnach: We need a different user name and password to allow us to use row-level security to filter rows. I found it easiest to use a different URL for this, since different authentication may also mean we need to use a different way of connecting to the DB. In my local setup, for example, I usually connect via the socket, but need to connect via TCP for DB users that are not tied to my system user, so the URL is different.

src/bin/dump-db.rs
smarnach: The visibility information in the configuration is for _columns_. There are a few tables we want to filter by _row_, though, and these are configured here.

I don't particularly like this design, but it does allow autogenerating the script, and testing whtether the column visibility information is up to date. If we are OK with ditching the unit test, we can go back to the slightly simpler desing in https://github.com/rust-lang/crates.io/issues/630#issuecomment-510380244.

migrations/2019-08-08-191129_add_users_public/up.sql
smarnach: The reason I made this a persistent column is so we can use row-level security to filter the rows. The comments on #630 explain how I arrived at this design. The most important reason is [this comment by Carol](https://github.com/rust-lang/crates.io/issues/630#issuecomment-510642673).

src/bin/enqueue-job.rs
smarnach: Sure. The main reason I moved the connection inside the function is to save some typing (and maybe not to connect to the DB at all if the job name is invalid).

src/bin/dump-db.rs
sgrif: My issue is that we're using the same source both to generate these permissions as we are to generate the queries. If there's a bug in one there'll almost certainly be a bug in the other. I'm not sure that this is buying us any security for the complexity it gives.

sgrif:
r? @carols10cents 

Pretty much all the responses to the concerns I raised reference your comments, I don't want to step on your toes.

sgrif:
> We can't use pgdump, since it can only dump full tables. See #630 (comment) for more details.

My initial thought was copying into tables on a separate schema and dumping that, but that probably wouldn't make things any simpler

src/bin/dump-db.rs
smarnach: Using row security is not an _additional_ mechanism to filter the rows in this design. It's the _only_ mechanism to do so. The database role is not required to achieve some security gain, but to be able to use row security at all, since row security does not affect the owner of the tables.

So the reasoning is roughly
* We want to have a test to verify that the column information is up to date.
* Thus, we auto-generate the dump script from a column visibility configuration.
* Thus we need an independent mechanism to filter rows.
* Thus I came up with (ab)using row security to filter rows.

It may not be such a bright idea after all. From the review comments, it became clear to me that this design is far from obvious, and I also severely dislike having to create a new database role. Creating a role is a global change for the database server, instead of one that only affects the current database, so it can lead to interference between different databases on the same server, notably between the test and production databases in local installations.

I'll try to come up with a better TOML configuration format that also allows defining row filters, and that uses temporary views or subqueries to filter the rows.

smarnach:
> My initial thought was copying into tables on a separate schema and dumping that, but that probably wouldn't make things any simpler.

Yeah, I considered that in the comments to #630, but overall I believe it would make things much more difficult. Creating a consistent dump will become more difficult since we won't be able to use `ISOLATION LEVEL SERIALIZABLE READ ONLY DEFERRABLE` anymore, and restoring the data will also become more involved in many cases for various reasons. Overall I think CSV gives us the best balance here.

smarnach:
Based on the feedback, I cleaned the design up a bit. I removed the hacky use of row-level security in favour of SQL row filter expressions in the TOML file. This way, the whole information about what is included in the database dump is in a single place now, and we essentially don't need any changes to the existing code. The `dump-db` binary is gone, and the additional database role no longer needed.

I kept all the `"private"` columns in `dump-db.toml`, so we can still have a unit test that verifies whether the TOML config is up to date.

sgrif:
FWIW, having a hot replica for failovers is something I've been wanting to
do anyway so it could be worth having this go to a follower anyway

src/tasks/dump-db.toml
```diff
@@ -0,0 +1,186 @@
...
+[publish_rate_overrides.columns]
+user_id = "private"
+burst = "private"
+
+[readme_renderings.columns]
+version_id = "public"
+rendered_at = "public"
+
+[reserved_crate_names.columns]
+name = "public"
```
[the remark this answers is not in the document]
smarnach: This was my reasoning to include it.

src/tasks/dump-db.toml
```diff
@@ -0,0 +1,186 @@
...
+[publish_limit_buckets.columns]
+user_id = "private"
+tokens = "private"
+last_refill = "private"
+
+[publish_rate_overrides.columns]
+user_id = "private"
+burst = "private"
+
+[readme_renderings.columns]
```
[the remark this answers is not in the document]
smarnach: You are right, this shouldn't be included. I excluded all similar information, like `version_downloads.counted`, so this should be excluded as well.

src/tasks/dump-db.toml
```diff
@@ -0,0 +1,186 @@
...
+created_at = "public"
+downloads = "public"
+description = "public"
+homepage = "public"
+documentation = "public"
+readme = "public"
+textsearchable_index_col = "public"
+license = "public"
+repository = "public"
+max_upload_size = "public"
```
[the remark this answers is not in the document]
smarnach: I personally think it's OK to make this public, but it's up to you to make a final call. In general I would err on the side of keeping things private, but this one feels pretty safe.

src/tasks/dump_db.rs
```diff
@@ -0,0 +1,260 @@
...
+    let database_url = if cfg!(test) {
+        crate::env("TEST_DATABASE_URL")
+    } else {
+        crate::env("DATABASE_URL")
+    };
+    run_psql(&visibility_config, &database_url, &export_dir)?;
+    let tarball = create_tarball(&export_dir)?;
+    let target_name = format!("dumps/{}", tarball.file_name().unwrap().to_str().unwrap());
+    upload_tarball(&tarball, &target_name, &env.uploader)?;
+    // TODO: more robust cleanup
```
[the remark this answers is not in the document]
smarnach: If an error occurs in the middle of the process, we leak the tarball and the export directory, so we will fill up `/tmp` over time if there are persistent errors with the dump. I was thinking of using a wrapper with a `Drop` implementation for the cleanup, so it will be called even in the case of errors.

smarnach:
@carols10cents Thanks for the review!

> I downloaded a production snapshot and ran the task locally, it ran in 2 min 10 seconds.

What was the size of the resulting tarball?

> That feels like we should be able to schedule it for once a day whenever our traffic is usually lowest without impacting production query performance.

I don't expect too much of an impact given that the database is relatively small – we don't blow out all the disk caches with the dump, so it should hardly be noticeable. (I don't know the exact size of the database, nor what hardware we are using, nor the exact access patterns, but this is my gut feeling.)

> I have not tried re-importing the data; I'm happy to try that to test instructions or a tool for doing so once one of those exists.

The code that generates the export psql script can be easily modified to also generate an import script, which we can simply include in the tarball.

>  we should have the date+time the dump was generated in the tar somewhere since it won't be in the filename, so that people know how long ago it was generated (and can bug us if the dumps stop getting generated for some reason)

The tarbal contains a directory named after the timestamp, so this should already be covered.

carols10cents:
> What was the size of the resulting tarball?

138M. The dump of the entire production database snapshot I started with is 151M.

> nor what hardware we are using

We use heroku standard 1x dynos that have 512MB RAM (for where the task will run), and our database plan is heroku's standard 0, 4GB RAM.

> The tarbal contains a directory named after the timestamp, so this should already be covered.

Oh right! That'll do :)

smarnach:
I went through all the database columns again and tried fetching the information via the API. Here is a list of all columns currently marked as "public" that I could not retrieve via the API:

* `categories.path` – can be reconstructed by running crates.io locally.
* `crate_owners.created_at`, `crate_owners.updated_at` – currently not public information, as far as I can tell, but seems safe to publish.
* `crates.readme`, `crates.licence` – these columns are written when a new crate is published, but then never read again, as far as I can tell. Maybe we should remove these columns?
* `crates.textsearchable_index_col` – currently not public, but could be reconstructed from the crate contents, so seems safe to publish.
* `crates.max_upload_size` – already [discussed in the comments above](https://github.com/rust-lang/crates.io/pull/1800#discussion_r315383328).
* `reserved_crate_names.name` – already [discussed in the comments above](https://github.com/rust-lang/crates.io/pull/1800#discussion_r316373927).
* `teams.github_id`, `users.github_id` – available from the GitHub API, so already public.

The columns `version_downloads.counted` and `version_downloads.processed` are currently marked "private", but it may actually be useful to include them in the dump. This would allow to run the `update_downloads` task locally after importing the data.

smarnach:
I just hit a rather unexpected rabbit hole. When calling `psql`, the only way to figure out whether there were errors is to look at the output on stderr, since the exit status of `psql` is always 0. The output on stdout is irrelevant, so I essentially need

    let mut psql = Command::new("psql")
        .arg(database_url)
        .current_dir(export_dir)
        .stdin(Stdio::piped())
        .stdout(Stdio::null())
        .stderr(Stdio::piped())
        .spawn()?;

However, now I need to simultaneously write to stdin and read from stderr, and there doesn't seem to be a standard way of doing this in Rust. Options I can think of:

1. Ignore the problem and hope that either the psql script or the errors on stderr will be small enough to fit in the pipe buffers, so we will never actually deadlock. Then we could simply first write the input and then read the errors on stderr. There is a good chance that this will _actually_ never deadlock, but I don't really like this approach.

1. Spawn a thread to read from stderr, and write to stdin in the existing thread. I'm not sure background tasks are supposed to spawn new threads on the production server, since this could lead to memory spikes, but at least this approach is fairly simple.

1. Use `epoll()` (probably via `mio`). Would be rather cumbersome and low-level.

1. Use `tokio_process`. Seems like a rather big gun.

1. Use the [subprocess-communicate](https://crates.io/crates/subprocess-communicate) crate, which looks unmaintained and does not seem to be in widespread use.

Any advice on this? I would personally go with approach 2 (spawn a thread).

sgrif:
> Spawn a thread to read from stderr, and write to stdin in the existing thread. I'm not sure background tasks are supposed to spawn new threads on the production server, since this could lead to memory spikes, but at least this approach is fairly simple.

This is my preference. There's no specific reason a background job wouldn't be allowed to spawn additional threads (both in general across any lang, and also with swirl specifically). In theory this should probably spawn those threads on the job runner's thread pool, but there's no public API for that in swirl right now, and that would really only be a concern if this were high enough volume that it could overwhelm the worker process. That isn't the case here.

smarnach:
I've pushed a new version and would appreciate some feedback. Updates since the last review:

* I've made the database URL and the name of the upload configurable. This way, we can switch to dumping a read-only replica without any code changes, simply by changing the invocation of `enqueue-job` in Heroku Scheduler.

* The upload is streamed to S3, and never completely held in memory at once anymore. I still have an open question here – **what is the correct content type for a gzipped tarball?** According to some sources it should be `application/gzip`, but I remember that this has caused issues for crate downloads before.

* The `psql` invocation now has some error handling. As discussed above, I first used a thread for simultaneously writing to stdin and reading from stderr, but later switched to loading the psql script from a file, so the thread is not needed anymore.

* The psql scripts are now rendered as Handlebars templates. This adds a few dependencies, but I think it makes both the SQL generation code as well as the generated code a lot more readable.

* The code now also generates a script to reimport the dump into the database. This adds some complexities to the TOML file and the script generation code:
    * We need to import the tables in the correct order to make sure foreign key references are valid. 
    * We need to set defaults for columns that are non-nullable and don't have a default, but are not included in the dump.

* Both the import and the export scripts are now included in the dump tarball. This makes it relatively easy to import the dump. The export script is included for reference and debugging.

* I've added a first version of an integration test that creates a dump and then imports it to a new schema. This test is still incomplete. I needed to add `IF NOT EXISTS` to the two migrations that create extensions, since extensions are database-wide and not per schema. There are a few **open questions** for this test:
    * This test is rather slow. Do we think it adds enough value to keep it?
    * How should we generate test data to export and re-import? Keeping test data in the repo is cumbersome to maintain, so we probably want to write some code to generate the data. Suggestions welcome.
    * The test clutters the test output with messages about the migrations it runs for the new schema. How should we suppress this output? Options include adding a feature to `diesel_migrations` allowing to silence the output, or running `diesel migration run --database-url=...` in a subprocess, so we can suppress the output of that subprocess.
    * The temporary schema currently has a fixed name, so running the integration test twice in parallel against the same database will fail. I don't see much of a problem with this, but it's worth pointing out.

jtgeibel:
> I still have an open question here – what is the correct content type for a gzipped tarball? According to some sources it should be `application/gzip`, but I remember that this has caused issues for crate downloads before.

You're correct, based on [this], `Content-Type: application/gzip` is the correct header (as a tarball isn't an official media type).  A long time ago, we ran into an issue where we uploaded some crates to S3 with `Content-Encoding: gzip` and this did cause some issues (you can see where this was fixed in 53cbb26d022a88f7f311740f35ec5c6d4a5a9ab0).  The `Content-Encoding` indicates that some type of encoding/compression was applied to the file while in transit, and that the client is expected to decompress or undo this encoding.  As an example, nginx may dynamically choose to add a `Content-Encoding` and compress our JSON API responses if they are large enough for that to be beneficial.  In that case, the browser is expected to decompress the contents before interpreting it as JSON.

When we incorrectly set this header for crate uploads, some clients were decompressing the file and leaving the `*.crate` as a plain tar file.

As long as we stick to setting the `Content-Type` when uploading the database dump to S3, we should be fine.

[this]: https://superuser.com/a/960710/183933

carols10cents:
I'm still in the process of reviewing your update, but wanted to post this thought before I lost it:

> To re-import the dump
> 
> ...
>
> 3. Run the Diesel migrations for the new DB:
>
>        diesel migration run --database-url=postgres:///test_import_dump
>

So for [this goal that you articulated in 630](https://github.com/rust-lang/crates.io/issues/630#issuecomment-510380244):

> * Allow crawlers to download a single archive with all information instead of hitting crates.io thousands of times to retrieve all information on all crates.

The use of the data the crawlers are gathering might not have anything to do with crates.io at all. It strikes me as a bit unfortunate that we'd require a checkout of crates.io's code and an installation of diesel to use the database for analysis, etc. What do you think about including the results of `pg_dump -s` so that the schema would be included in the tar?

smarnach:
I actually expect the data to be used in different ways for the different use cases. If you want to run crates.io locally, either for development and testing, or to be able to hammer the API without any rate limits, then you of course need to import the data into the crates.io databse schema. The import script is intended for that use case.

If you don't want to run crates.io locally, I don't think you need to import the data into a Postgres database with the same schema that we are using. Depending on your use case, you may want a database with just the columns and tables you need or an in-memory SQLite database. You may simply want to load the CSV files into memory without using a database at all, or you may want something yet different.

I intentionally chose the CSV format to give people flexibility to use the data in different ways, and to avoid forcing them to run a Postgres server if they don't otherwise need it. People will certainly come up with their own solutions how to import the data, and we don't need to do all the work for them. I focused on the use case of re-importing the data to our schema for several reasons:

* It's actually the most _difficult_ way of using the data. We only have a partial dump and a complex schema. Postgres's `COPY` command does not pay attention to the column names in the CSV header, so you need to explicitly list them again in each import command. You need to import the tables in the right order, and you need to set a column default for one column.

* It gives people who are currently scraping the data from our API an easy upgrade path – download the dump, run crates.io locally and hammer the API with the code they already have. This would not only significantly speed up the process, since they are not limited to one request per second anymore; it would also remove all problems with data consistency they likely currently have, due to the data constantly changing while they are scraping it.

* We can use the import script for an integration test.

* It's the use case I am personally interested in. I want to be able to test changes locally with realistic data.

All that said, I actually like the idea of also including the output of `pg_dump -s`. It doesn't hurt at all to do so – it's just 6 KB once gzipped – and people may find it useful as a quick way to get started.

We also should include the commit hash of the crates.io version that created the dump in the tarball – people need to use the correct version to make sure the import script actually works.

I am also considering to implement a Rust library to work with this data directly, by loading it into memory with the CSV crate and providing a simple API to get all information for a crate. This would allow me to understand how easy it actually is to use the data in that way, and possibly make improvements.

I don't think we need to nail all the use cases before going ahead with this PR, though. We could announce the feature as "alpha" or "beta" for before we guarantee stability of the dump format, so we can collect feedback and iterate.

src/tasks/dump_db/readme_for_tarball.md
```diff
@@ -0,0 +1,30 @@
+# crates.io Database Dump
```
carols10cents: I love this change and the change moving the CSVs to a `data` directory! Really cleans up the `tar` file contents.

src/tasks/dump_db.rs
```diff
@@ -29,24 +29,57 @@ pub fn dump_db(
...
+        struct Metadata<'a> {
+            timestamp: &'a chrono::DateTime<chrono::Utc>,
+            crates_io_commit: String,
+            format_version: &'static str,
+        }
+        let metadata = Metadata {
+            timestamp: &self.timestamp,
+            crates_io_commit: dotenv::var("HEROKU_SLUG_COMMIT")
+                .unwrap_or_else(|_| "unknown".to_owned()),
+            format_version: "0.1",
```
carols10cents: I'm assuming you're envisioning this version number to increase its minor version with any backwards compatible schema change and increase its major version with any breaking schema change? So to give some information about compatibility with previous database dumps without needing to look up changes between git SHAs?

I'm a little concerned about the usefulness of this version number because I can very easily see us forgetting to update this with schema changes :-/ If it's not incremented meaningfully, then it's not going to be particularly useful, and people using the database dumps will need to look up what changed in the git history anyway.

Given that we recommend using the import script that does a destructive import anyway, rather than providing diffs between dumps say, is this going to provide enough value for people using the dumps to justify our maintenance of it?

smarnach: You are right, this would need a bit more specification to be useful.

I did not have in mind to increase this on every schema change, but rather everytime we change something more fundamentally to the format of the tarball, like adding a fundamentally new file.

The main goal was to allow us to make backwards-incompatible changes, and indicate these in the version number. However, this doesn't really work. If we change the dump in a way that breaks compatibility, people will notice without checking a version number, since there tool will break. And if they can't go back to the old version, this version number does not really help.

So the only way to break compatibility is to provide dumps in the new format at a different URL alongside the dumps in the old format. And that does not require a version number, so I'll remove it again.

smarnach:
To be able to proceed, I would need feedback on the following points:

* The [summary of additional columns we are about to publish](https://github.com/rust-lang/crates.io/pull/1800#issuecomment-524064670). I personally think these are fine, but I'd like to get a confirmation from someone else on the team as well.

* The [questions about the integration test](https://github.com/rust-lang/crates.io/pull/1800#issuecomment-529207899), in particular whether we want to keep the test in roughly its current form, and how we should generate test data. I would personally tend to keep the test, since it at least verifies that the export and import script are syntactically corrent, and I would try and use the existing factory functions to generate some rudimentary test data, but I'd keep it minimal, so it does not become too much of a maintenance burden.

carols10cents:
Ok, I deployed 620bb51 to staging and enqueued a dump-db job; the resulting tar is available at https://alexcrichton-test.s3.amazonaws.com/db-dump.tar.gz if you want to check it out (there's a small amount of data on https://staging-crates-io.herokuapp.com/ that has been used to test various things, so nothing production sized but also nothing important that we can't reset if we've leaked something).

The git commit is in the metadata, uploading to s3 went fine, I actually did the dump-db twice and the file was overwritten correctly.

I haven't tested passing arguments to the enqueueing of the dump-db task (but I'm not expecting that to have problems really); I have started the creation of a follower db on production so that we can use the follower once this is deployed.

carols10cents:
> To be able to proceed, I would need feedback on the following points:
>
> * The [summary of additional columns we are about to publish](https://github.com/rust-lang/crates.io/pull/1800#issuecomment-524064670). I personally think these are fine, but I'd like to get a confirmation from someone else on the team as well.

I think the way the columns are configured currently is fine as well. I also don't think trying to reconstruct dumps from the information available via the API and compare to information in a test dump in the staging environment (the idea in this PR's description) is worthwhile. Between the two of us, I think we've done a careful consideration of each column.

> * The [questions about the integration test](https://github.com/rust-lang/crates.io/pull/1800#issuecomment-529207899), in particular whether we want to keep the test in roughly its current form, and how we should generate test data. I would personally tend to keep the test, since it at least verifies that the export and import script are syntactically corrent, and I would try and use the existing factory functions to generate some rudimentary test data, but I'd keep it minimal, so it does not become too much of a maintenance burden.

(bringing the questions about integration tests from your linked column in)

> This test is rather slow (about 1.3 seconds wall time on my machine, but will be slower once we add more data). Do we think it adds enough value to keep it?

Yes, I think 1.3 s is worth it.

> How should we generate test data to export and re-import? Keeping test data in the repo is cumbersome to maintain, so we probably want to write some code to generate the data. Suggestions welcome.

I think it's much more likely that we'll mess up the schema such that we break the import rather than something about the data; I think it'd be ok to merge this as-is for now and add tests as we break this feature and people complain 🤷‍♀ That'll ensure there's value in the tests.

> The temporary schema currently has a fixed name, so running the integration test twice in parallel against the same database will fail. I don't see much of a problem with this, but it's worth pointing out. It would be easy to randomize the name of the schema if we think this is an issue.

I can't think of a scenario where I'd want to run the same test multiple times in parallel; I'm also fine categorizing this as "merge as-is and fix it if folks complain".

The remaining concern I have before merging this is documentation! I could potentially be convinced that we should merge this without documentation and do a "soft launch" in production, but my current concern is that we'll forget to come back to document the existence of these database dumps and they won't get used/people will complain that we haven't documented this when they find out it exists.

smarnach:
@carols10cents Thanks a lot for your reviews! No worries about the scattered reviews – I'm glad you managed to find the time for this big a review at all. I only summarized what answers I'm still waiting for since it was easy to lose track of what has already been answered, not to put pressure on you for quicker answers.

I would like to include documentation in this PR, but I first wanted to have everything else finished, to avoid having to rewrite the documentation all the time. Glad we are close to finalizing this. :)

smarnach:
@carols10cents I've added some documentation now. The link is not very prominent, and I marked the database dumps as "experimental" for the first launch.

I'd appreciate some language improvements for the docs by a native speaker.

carols10cents:
@bors r+

I think we're ready to go with this!!! Will work on deploying and setting up the scheduled job to enqueue the dump task once bors merges. I changed the docs to be 24 hours and I'm going to set up the task for once per day; I think that's a good starting place.

smarnach:
@carols10cents Oh, right, I meant to comment about the backup frequency, but forgot. Of course I'm fine with starting with once a day.

I suggested to do it every six hours because we are using a follower DB for the backups, so it shouldn't have any impact on performance, but it might increase the CloudFront bill if people are downloading the dumps more frequently.

Thanks a lot for guiding me through this process, I'm glad we got this done!

PR #1801, opened 2019-08-13
Bump `route-recognizer` to address build failure on nightly

jtgeibel:


smarnach:
Looks good to me. I reproduced the build failure on nightly and verified it is fixed by this change. The only change to route-recognizer between version 0.1.12 and 0.1.13 is to add `#[cfg(test)]` to a few benchmarks.

jtgeibel:
@bors r=smarnach

Thanks!

PR #1806, opened 2019-08-18
Allow blocking arbitrary traffic without redeploying

sgrif:
This generalizes the `block_ips` module to allow checking against any
header instead of just the `X-Real-Ip` header. This does not enable more
complex matching such as regexes, but it does let us apply the simple
string matching we have today to any header.

Additionally, which headers we check against is now configured by an
environment variable, which is a comma separated list in the form
`Header=VALUE_ENV_VAR`. So to keep the behavior we had today, one would
set `BLOCKED_TRAFFIC="X-Real-Ip=BLOCKED_IPS"`.

This means that when we see traffic patterns that we need to block which
can be handled by the existing logic, we can now begin applying it to
patterns we didn't previosuly anticipate without requiring a redeploy

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -43,6 +43,8 @@ impl Default for Config {
     /// - `GH_CLIENT_ID`: The client ID of the associated GitHub application.
     /// - `GH_CLIENT_SECRET`: The client secret of the associated GitHub application.
     /// - `DATABASE_URL`: The URL of the postgres database to use.
+    /// - `BLOCKED_TRAFFIC`: A list of headers and environment variables to use for blocking
+    ///.  traffic. See the `block_traffic` module for more documentation.
     fn default() -> Config {
         let checkout = PathBuf::from(env("GIT_REPO_CHECKOUT"));
         let api_protocol = String::from("https");
--- a/src/config.rs
+++ b/src/config.rs
@@ -18,6 +18,7 @@ pub struct Config {
     pub mirror: Replica,
     pub api_protocol: String,
     pub publish_rate_limit: PublishRateLimit,
+    pub blocked_traffic: Vec<(String, Vec<String>)>,
 }
 
 impl Default for Config {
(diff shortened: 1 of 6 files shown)


smarnach:
I'm wondering – why do we implement IP blocking and rate limiting in the backend server? Wouldn't it be easier to include this in the nginx configuration?

sgrif:
Most rate limiting is performed at the NGINX level. The only case that isn't is for publishing new crates. Determining whether `PUT /api/v1/crates/new` is a new crate or a new version of an existing crate requires connecting to the database, and for this case we also want to limit by user instead of IP, which requires connecting to the database (and also has the side effect of making this consistent regardless of how many servers we run).

We have another rate limit for publishes, which is less strict (and mostly to ensure we don't get rate limited by GH) which is handled in nginx, and I'm planning on opening a PR to add a few more at the nginx level soon.

As for why blocking is done in app code, that's mostly a function of familiarity amongst the team. Blocking traffic is something that primarily happens during incident response, so having it happen where folks who are on call are most familiar with things is important to make sure we can effectively respond to these incidents quickly. My understanding is that nginx isn't really meant for that use case, and while it can be used for IP blocking (it couldn't when we first introduced the app level since $remote_addr is not correct by default, but I believe we do have the appropriate extensions configured for it to handle it now) -- It still isn't really set up to do the sort of arbitrary blocking that I'm referring to here, especially not without a redeploy. That said, I'm open to being proven wrong and moving this to nginx if everyone handling ops is up to date

smarnach:
@sgrif Thanks for the detailed explanation. The rate-limiting setup completely makes sense to me now – if we need to connect to the DB, we can only do this in the backend, and the common case is already handled by nginx.

IP blocking can be configured in nginx using `deny`, but of course this would require to modify the configuration file, so we would need to redeploy on Heroku. It is possible to set up dynamic reconfiguration in nginx, but this would require further work.

So overall I understand the reasoning for implementing this ourselves, and it seems like the best approach for now. However, if we go ahead and implement more general HTTP filtering rules, we may get to the point where switching to a different reverse proxy like HAProxy would be worthwhile. We don't have a ton of logic in the nginx configuration, so the migration should be rather straightforward.

Somewhat related, do we have any ops documentation for crates.io? If you want to get more people on the on-call rota, we _should_ have documentation for common tasks.

sgrif:
>  However, if we go ahead and implement more general HTTP filtering rules, we may get to the point where switching to a different reverse proxy like HAProxy would be worthwhile.

Yes, I agree. This is very much a "simplest thing that works for our needs right now".

> Somewhat related, do we have any ops documentation for crates.io? If you want to get more people on the on-call rota, we should have documentation for common tasks.

A little bit, but it's sparse and scattered. Once we start bringing more folks on I'm going to make sure we have a playbook that they can reference

smarnach:
I think you closed this by accident.

sgrif:
Thank you

carols10cents:
@sgrif I pushed 2 commits to this branch; one adding a bit of docs and one adding a happy path test. If my new commits look good to you, feel free to merge this.

sgrif:
Looks great. Thanks!

@bors: r+

Issue #1805, opened 2019-08-17
Tracking issue for `cargo update`


jtgeibel:
We've tried twice now to land a broad `cargo update` against the backend, but both attempts resulted in segfaults in production.  This was first attempted in #1263 (April 2018), and again in #1756 (June 2019).

I've been examining the differences between that first attempt, current master, and their common ancestor.  Following is a list of crates which were bumped in that first attempt but which have not yet been superseded by a more recent version on master.  My hope is that the problematic change will be within this set of crates.

I've added columns to capture the version bump from the 2nd attempt, and the most recent (possibly semver incompatible) published version.  I've highlighted in bold the places where the version bump in the 2nd attempt was identical to the first.

|                | Master | 1st attempt | 2nd attempt | Most recent | Notes |
| -------------- | ------ | ----------- | ----------- | ----------- | ----- |
| ~aho-corasick~   | 0.6.3  | 0.6.4       | 0.6.10      | 0.7.6       | N/A after #1917 |
| ~backtrace~      | 0.3.3  | 0.3.5       | 0.3.26      | 0.3.34      | #1861 |
| ~docopt~         | 0.8.1  | 0.8.3       | **0.8.3**   | 1.1.0       | only used in render-readmes binary #1812 |
| ~filetime~       | 0.1.14 | 0.1.15      | **0.1.15**  | 0.2.7       | old version pulled in via conduit-static - #2089 |
| ~lazy_static~    | 0.2.9  | 0.2.11      | **0.2.11**  | 1.3.0       | pulled in via old semver (via conduit) - #2099 |
| ~maplit~         | 1.0.0  | 1.0.1       | **1.0.1**   | 1.0.1       | #2089 |
| ~num~            | 0.1.40 | 0.1.42      | gone        | 0.2.0       | #1917 |
| ~num-integer~    | 0.1.35 | 0.1.36      | 0.1.41      | 0.1.41      | #1917 |
| ~num-iter~       | 0.1.34 | 0.1.35      | gone        | 0.1.39      | #1917 |
| ~openssl-probe~  | 0.1.1  | 0.1.2       | **0.1.2**   | 0.1.2       | #2099 |
| ~r2d2~           | 0.8.1  | 0.8.2       | 0.8.4       | 0.8.6       | #1899 |
| ~regex~          | 0.2.2  | 0.2.6       | 0.2.11      | 1.2.1       | N/A after #1917 |
[...]

PR #1812, opened 2019-08-23
Bump docopt 0.8 -> 1.1.0

jtgeibel:
Docopt is only used directly in the `render_readme` binary, though this
does transitively update some crates that are used in production.

Refs #1265, #1805


sgrif:
@bors: r+

Will deploy on its own when I get home

PR #1813, opened 2019-08-27
Allow streaming uploads to S3.

smarnach:
Currently the interface in the `s3` crate accepts the contents of an upload as a byte slice, which
so far isn't much of a problem, since crate uploads are generally below 10MB. When uploading public
database dumps to S3 (#1800), however, we don't want to keep the whole dump in memory at once.

This change basically changes the type of the uploaded content from `&[u8]` to a generic type
implementing `std::io::Read + Send + 'static`, which are the trait bounds `reqwest` requires for the
request body.

The PUT request to S3 needs to include a Content-Length header. Since a `std::io::Read` does not
have a length, we need to add another parameter for the content length, and pass it on to `reqwest`.

### DIFF
--- a/src/uploaders.rs
+++ b/src/uploaders.rs
@@ -7,7 +7,7 @@ use crate::util::{human, internal, CargoResult, ChainError, Maximums};
 
 use std::env;
 use std::fs::{self, File};
-use std::io::{Read, Write};
+use std::io::{Cursor, Read};
 use std::sync::Arc;
 
 use crate::middleware::app::RequestApp;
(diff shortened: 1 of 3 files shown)


src/uploaders.rs
```diff
@@ -83,30 +83,29 @@ impl Uploader {
 
     /// Uploads a file using the configured uploader (either `S3`, `Local`).
     ///
-    /// It returns a a tuple containing the path of the uploaded file
-    /// and its checksum.
-    pub fn upload(
+    /// It returns the path of the uploaded file.
```
smarnach: As far as I can tell, the return value is never used, so we could just as well get rid of it.

PR #1814, opened 2019-08-27
Use docker-compose to run crates.io inside Travis

kzys:


smarnach:
The point I don't understand here is why we need docker-compose on Travis for end-to-end tests. We could also run both the frontend and the backend directly in the Travis container, without the additional level of containers.

(I'm not disagreeing with you, just curious how you got to this design.)

kzys:
Good point. We can!

The biggest reason I choose docker-compose here is that we already have `docker-compose.yml` and 2 Dockerfiles to run `backend` and `frontend` together. In other words, someone has chosen that before me :)

We also have a completely different set of configuration files (Procfile and buildpacks) for the production environment. Running frontend and backend by writing a small shell script is possible, but it will add another set of configuration files, which I'd like to avoid.

smarnach:
@kzys Thanks for the additional explanation.

I doubt it is feasible to make the production, test and development environments much more similar. In production we don't "run" the frontend – it's just a bunch of compiled static files, so I've got no idea what the "frontend" container would be supposed to be doing in production. In the development environment, we want to run the Ember test server, to detect changes and automatically recompile the JavaScript files. The test environment will have yet slightly different requirements.

I'd personally tend to avoid the complexities of using Docker containers for the tests, and certainly in production, since I don't see enough advantages, but I don't really know enough about our deployment for my opinion to bear much weight.

sgrif:
I agree with @smarnach here:

> I'd personally tend to avoid the complexities of using Docker containers for the tests, and certainly in production, since I don't see enough advantages

It looks like you're working on an alternative PR, so I'm going to close this in favor of that one

kzys:
@sgrif Can we keep this one open? #1825 is still work in progress and not working correctly. I'd like to have two working solutions first.

sgrif:
I'm not sure there's much reason to keep this open, given the consensus from two team members we don't want to go in this direction

PR #1815, opened 2019-08-28
Remove `license` column on `crates` table

rye:
Seemingly the last step in the checklist on #736, this PR removes the `license` column on the `crates` table, and all references thereto. (In fact, I think #736 can be closed once this PR is merged.)

As of #803, the front-end no longer really used this field, and the front-end was also changed to render the license of the _current_ version, instead of the value coming from this now-obsolete `license` column on the crate. Thus, this column likely contains stale data and, to my knowledge, isn't changed at all when a new version is uploaded.

I wasn't sure what requirements are enforced for backwards-compatibility, or whether my changes here would actually affect behavior of e.g. if there are old versions of `cargo` that will upload with these values. I am happy for these changes to be scrutinized until this is verified to be safe.


rye:
@sgrif, thanks! The rebase has been pushed… I merged the `cargo fmt` into the first commit as it was where the style fell out of consistency.

sgrif:
Thank you! This looks fine, but I'm going to hold off on merging until tomorrow so I can deploy it immediately after since this contains destructive migrations.

PR #1816, opened 2019-08-29
Make the users.gh_access_token column nullable.

smarnach:
We would like to be able to re-import public database dumps to a local deployment of crates.io, e.g.
for local testing with realistic data or to query the API locally. The only blocker for this is the
`gh_access_token` column – we obviously can't include it in the dump, but without it we can't import
the dump, since the column is not nullable (nor does it have a default value).

This change makes the column nullable. The code for creating new users is not changed, so for users
created in normal operation the presence of the ticket is still enforced, and the column can only be
null for users created by manual database manupulation.


smarnach:
I completely agree with you on the principles. Production should not suffer from this, and invariants should preferably be enforced at the database level.

However, I don't think this particular constraint is an invariant we care about. The GitHub token may be expired or revoked or invalid for other reasons, and we have to deal with these error conditions anyway. We don't add any fundamentally new failure modes by allowing the token to be absent.

In addition, we are still enforcing the invariant at the application level. In the `NewUser` struct and implementation, the token is still mandatory. Many important invariants are only enforced at the application level for practical reasons, which is absolutely fine. And this isn't even an important invariant.

I don't actually have the expectation that the database dump can always be imported without changes to the schema. It just so happens that it is possible to make this work without harming production in any meaningful way, so we may just as well take advantage of this coincidence.

That said, I considered all the other options you listed as well, and would be fine with simply adding the migration to the import script. It will look a bit out of place there, which is why I personally prefer the approach in this PR, but that's just a gut feeling, not an actual technical reason.

smarnach:
I'm closing this for now. On second thought, I can make it a first-class feature of the import script to set defaults on columns that need them. While I still think that this PR wouldn't hurt, it is probably better to set a precendent how this should be handled if cases like this one pop up in the future.

PR #1817, opened 2019-08-30
Two-Space Linting

CodingItWrong:
Updates the JS and HBS linters to use the standard two-spaces instead of four, at @locks' request. This should make working with blueprints easier in the future.

Other changes:

- Adds the standard `lint:hbs` NPM script and runs it as part of CI.
- Disable the `html-comments` lint rule, as there are HTML comments present.


kzys:
Can we merge #1715 before this one? Merging this change makes merging open pull requests harder...

CodingItWrong:
It's true; @locks as you requested the change do you have thoughts on how to make this change timing-wise?

locks:
I think it sounds good to wait for #1715 since it's a big PR. If there are other PRs affects, I'll personally update them to unblock them as necessary!

CodingItWrong:
@kzys I'm happy to update this PR after #1715 is merged 👍

smarnach:
Could we also get #1759 merged first? Not that that PR is particularly important, nor particularly big, but I believe that it is ready for merging, so it may save a bit of work to simply merge it.

sgrif:
You can go ahead and rebase now

CodingItWrong:
@sgrif thanks for coordinating—i should be able to do so tomorrow morning if that works ok

CodingItWrong:
@sgrif alright, this PR is all rebased and ready to go

PR #1824, opened 2019-09-04
update node versions to current stable release

locks:


jtgeibel:
I've been using a weird combo of node 10 and npm 5 from my distro, so this was a good excuse to switch to the official upstream packages.

I was able to reproduce the build error (in node-gyp) locally, and resolved it by upgrading `ember-cli-sass` to `^10.0.0`, then running `npm install` and `ember g ember-cli-sass`.  After that, everything seems to be running fine for me locally!

locks:
@jtgeibel thank you for investigating! I have followed your instructions to get the PR green.

PR #1827, opened 2019-09-05
Revert "Auto merge of #1715 - kzys:enable-fastboot, r=jtgeibel"

sgrif:
This reverts commit 111b7eb0f2a9e49c164d86e6aac1d62f27c47c16, reversing
changes made to 27bd9a802c31d619493844dd322bf4b53e9efc99.

This caused our memory usage to increase from 50MB to 350MB at boot, and
I was able to trivially cause the server to run out of memory by hitting
/policies repeatedly. This introduces a DOS vector that we can't afford.


sgrif:
@bors: r+

I've already rolled back the deploy in production, and this is purely a revert commit

Issue #1811, opened 2019-08-20
Tracking issue for Ember FastBoot


jtgeibel:
Recently I've been experimenting with the FastBoot PR, and below is my attempt at capturing the work that I think remains to enable FastBoot across the site.

##  Enable FastBoot for some static pages

As we previously discussed, the existing PR allows us to land support on a single static page to try this out in production:

- [x] `/policies` - added in #1900

## Groundwork for enabling FastBoot on dynamic pages

Some tasks that I think will be necessary support more dynamic cases:

- [x] Find a way to show the `<noscript>` message only on pages that haven't been converted yet. - #2101 tweaks the message and placement, but leaves the message present on all pages.
- [ ] Ensure a complete set of headers is sent, based on what we do in the backend middleware. - #2100
- [ ] For a logged in user, their cookie should be included in the request to the backend.
- [x] When JS is enabled, after the pre-rendered HTML is loaded and Ember boots, a new set of requests are sent from the client to the backend. - `ember-data-storefront` added in #2087 
- [ ] Requests to routes (`/does-not-exist`) or resources (`/crates/does-not-exist`) that don't exist, should respond with status 404 instead of 200.
- [ ] Update FastBoot logs to align with our typical log format, as described in https://github.com/rust-lang/crates.io/issues/1811#issuecomment-528419399 and https://github.com/rust-lang/crates.io/pull/1900#issuecomment-554496078
- [x] Pass User-Agent header to backend - fixed in #2048 

## Page specific bugs

Following are bugs I observed while testing with JS off:

- [x] `/` - Data doesn't load - fixed in #1937
- [ ] `/login` - TypeError: window.open is not a function
- [ ] `/logout` - completely blank page
- [ ] `github_login`, `/authorize/github` - Unclear how the login workflow should work without JS, or if we should even support that
- [ ] `/crates/test-crate`, `/crates/test-crate/0.1.5` - "Authors" and "Owners" sections are not populated
- [ ] `/crates/*/download` - I think this endpoint can be removed, it doesn't have an associated route file in the frontend
- [x] `/crates/*/{docs,repo} - Doesn't actually redirect when the metadata is set - added in #1912
- [ ] `/me/*`, `/dashboard` - redirects to `/`
- [ ] `/{categories,keywords,users,teams}/does-not-exist` - results in a 500 server error
- [ ] `/search` - doesn't load any search results
[...]

PR #1832, opened 2019-09-08
Enable FastBoot (again)

kzys:
Compared to #1715, this branch

- Runs 2 processes from Procfile, instead of using node-foreman.
- Makes the number of child processes `cluster` creates configurable, and the default value is one.

The memory consumption is at most ~197MB on my Heroku environment. Of course the environment is having only a few requests from my browser though. Would it be acceptable?

Untitled

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 release: bin/diesel migration run
-web: node fastboot.js & bin/start-nginx target/release/server & wait -n
+web: node --optimize_for_size --max_old_space_size=200 fastboot.js & bin/start-nginx target/release/server & wait -n
 background_worker: ./target/release/background-worker
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 release: bin/diesel migration run
-web: bin/start-nginx ./target/release/server
+web: node fastboot.js & bin/start-nginx target/release/server & wait -n
 background_worker: ./target/release/background-worker
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -33,6 +33,8 @@ body {
   @include align-items(center);
 }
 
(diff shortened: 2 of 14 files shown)


kzys:
Untitled 2

kzys:
Okay. The last commit (ebff9feb87d296e8741f398c897bca0fa7e2c46b) changed the memory consumption a lot.

Untitled

kzys:
I will update the branch once #1853 got merged, since this branch probably will conflict with that.

jtgeibel:
I think it would be beneficial to break out enabling FastBoot (in the Procfile, nginx configuration, and app-initialized file handling) into a separate PR.  That way we could land the other fixes in this PR ASAP, and then have a smaller PR that is easier to enable and revert if necessary.

It would be ideal if we could dynamically enable/disable FastBoot via an environment variable, similar to how hyper can be selected at boot time with the [USE_HYPER](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/bin/server.rs#L60) environment variable.  However, that seems tricky in this case, because we would need to dynamically configure nginx before starting it.

So I don't think we need to try to make this a dynamic configuation at this time, but I think it would make sense to break out the diff that actually enables FastBoot into its own PR and land the rest of these changes now.

kzys:
I think it is possible to implement the dynamic configuration, since the Nginx Buildpack is using ERB for templating.

Anyway, I've created #1886 which contains Fastboot-related changes, except for actually enabling Fastboot. Please take a look!

kzys:
Since we already have #1811 for tracking. I'm going to close this PR. We may have discussions on the tracking PR.

PR #1838, opened 2019-09-13
Support GitHub Actions Badges

CryZe:
Documented here https://help.github.com/en/articles/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

### DIFF
--- a/app/components/badge-github-actions.js
+++ b/app/components/badge-github-actions.js
@@ -45,12 +45,12 @@ export default Component.extend({
       .join('/');
   }),
   branch: computed('badge.attributes.branch', function() {
-    return encodeURIComponent(this.get('badge.attributes.branch') || '');
+    return encodeURIComponent(this.get('badge.attributes.branch') || 'master');
   }),
   event: computed('badge.attributes.event', function() {
     return encodeURIComponent(this.get('badge.attributes.event') || '');
   }),
   text: computed('badge', function() {
-    return `GitHub Actions workflow status for the ${this.workflow} workflow`;
+    return `GitHub Actions workflow status for the ${this.workflow} workflow on the ${this.branch} branch`;
   }),
 });
--- a/app/components/badge-github-actions.js
+++ b/app/components/badge-github-actions.js
@@ -5,8 +5,8 @@ import { alias } from '@ember/object/computed';
 export default Component.extend({
   tagName: 'span',
   repository: alias('badge.attributes.repository'),
-  imageUrl: computed('badge.attributes.{repository,workflow,branch,event}', function() {
-    const url = new URL(`https://github.com/${this.repository}/workflows/${this.workflow}/badge.svg`);
+  imageUrl: computed('badge.attributes.{repository,workflow_enc,branch,event}', function() {
+    const url = new URL(`https://github.com/${this.repository}/workflows/${this.workflow_enc}/badge.svg`);
 
     if (this.branch !== '') {
       url.searchParams.set('branch', this.branch);
@@ -19,6 +19,9 @@ export default Component.extend({
     return url.href;
   }),
   workflow: computed('badge.attributes.workflow', function() {
+    return this.get('badge.attributes.workflow');
+  }),
+  workflow_enc: computed('badge.attributes.workflow', function() {
     return this.get('badge.attributes.workflow')
       .split('/')
       .map(encodeURIComponent)
--- a/app/components/badge-github-actions.js
+++ b/app/components/badge-github-actions.js
@@ -18,19 +18,24 @@ export default Component.extend({
 
     return url.href;
   }),
-  url: computed('badge.attributes.{repository,workflow_enc,branch,event}', function() {
+  url: computed('badge.attributes.{repository,workflow,branch,event}', function() {
     const url = new URL(`https://github.com/${this.repository}/actions`);
 
-    if (this.workflow_enc !== '') {
-      url.searchParams.set('workflow', this.workflow_enc);
+    let query = '';
+    if (this.workflow !== '') {
+      query += `workflow:"${this.workflow}"`;
     }
 
     if (this.branch !== '') {
-      url.searchParams.set('branch', this.branch);
+      query += `branch:"${this.branch}"`;
     }
 
     if (this.event !== '') {
-      url.searchParams.set('event', this.event);
+      query += `event:"${this.event}"`;
+    }
+
+    if (query !== '') {
+      url.searchParams.set('query', query);
     }
 
     return url.href;
--- a/app/components/badge-github-actions.js
+++ b/app/components/badge-github-actions.js
@@ -1,0 +1,13 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { alias } from '@ember/object/computed';
+
+export default Component.extend({
+  tagName: 'span',
+  classNames: ['badge'],
+  repository: alias('badge.attributes.repository'),
+  workflow: alias('badge.attributes.workflow'),
+  text: computed('badge', function() {
+    return `GitHub Actions workflow status for the ${this.workflow} workflow`;
+  }),
+});
--- a/app/components/badge-github-actions.js
+++ b/app/components/badge-github-actions.js
@@ -6,7 +6,7 @@ export default Component.extend({
   tagName: 'span',
   classNames: ['badge'],
   repository: alias('badge.attributes.repository'),
-  imageUrl: computed('badge.attributes.repository', 'badge.attributes.workflow', 'badge.attributes.branch', 'badge.attributes.event', function () {
+  imageUrl: computed('badge.attributes.{repository,workflow,branch,event}', function() {
     const query = Object.entries({
       branch: this.branch,
       event: this.event,
@@ -23,19 +23,19 @@ export default Component.extend({
       return base;
     }
   }),
-  workflow: computed('badge.attributes.workflow', function () {
+  workflow: computed('badge.attributes.workflow', function() {
     return this.get('badge.attributes.workflow')
       .split('/')
       .map(encodeURIComponent)
       .join('/');
   }),
-  branch: computed('badge.attributes.branch', function () {
+  branch: computed('badge.attributes.branch', function() {
     return encodeURIComponent(this.get('badge.attributes.branch') || '');
   }),
-  event: computed('badge.attributes.event', function () {
+  event: computed('badge.attributes.event', function() {
     return encodeURIComponent(this.get('badge.attributes.event') || '');
   }),
-  text: computed('badge', function () {
+  text: computed('badge', function() {
     return `GitHub Actions workflow status for the ${this.workflow} workflow`;
   }),
 });
--- a/app/components/badge-github-actions.js
+++ b/app/components/badge-github-actions.js
@@ -7,21 +7,17 @@ export default Component.extend({
   classNames: ['badge'],
   repository: alias('badge.attributes.repository'),
   imageUrl: computed('badge.attributes.{repository,workflow,branch,event}', function() {
-    const query = Object.entries({
-      branch: this.branch,
-      event: this.event,
-    })
-      .filter(kv => kv[1] != '')
-      .map(kv => kv.map(encodeURIComponent).join('='))
-      .join('&');
+    const url = new URL(`https://github.com/${this.repository}/workflows/${this.workflow}/badge.svg`);
 
-    const base = `https://github.com/${this.repository}/workflows/${this.workflow}/badge.svg`;
-
-    if (query != '') {
-      return `${base}?${query}`;
-    } else {
-      return base;
+    if (this.branch !== '') {
+      url.searchParams.set('branch', this.branch);
     }
+
+    if (this.event !== '') {
+      url.searchParams.set('event', this.event);
+    }
+
+    return url.href;
   }),
   workflow: computed('badge.attributes.workflow', function() {
     return this.get('badge.attributes.workflow')
--- a/app/components/badge-github-actions.js
+++ b/app/components/badge-github-actions.js
@@ -6,13 +6,36 @@ export default Component.extend({
   tagName: 'span',
   classNames: ['badge'],
   repository: alias('badge.attributes.repository'),
-  workflow: computed('badge.attributes.workflow', function() {
+  imageUrl: computed('badge.attributes.repository', 'badge.attributes.workflow', 'badge.attributes.branch', 'badge.attributes.event', function () {
+    const query = Object.entries({
+      branch: this.branch,
+      event: this.event,
+    })
+      .filter(kv => kv[1] != '')
+      .map(kv => kv.map(encodeURIComponent).join('='))
+      .join('&');
+
+    const base = `https://github.com/${this.repository}/workflows/${this.workflow}/badge.svg`;
+
+    if (query != '') {
+      return `${base}?${query}`;
+    } else {
+      return base;
+    }
+  }),
+  workflow: computed('badge.attributes.workflow', function () {
     return this.get('badge.attributes.workflow')
       .split('/')
       .map(encodeURIComponent)
       .join('/');
   }),
-  text: computed('badge', function() {
+  branch: computed('badge.attributes.branch', function () {
+    return encodeURIComponent(this.get('badge.attributes.branch') || '');
+  }),
+  event: computed('badge.attributes.event', function () {
+    return encodeURIComponent(this.get('badge.attributes.event') || '');
+  }),
+  text: computed('badge', function () {
     return `GitHub Actions workflow status for the ${this.workflow} workflow`;
   }),
 });
--- a/app/components/badge-github-actions.js
+++ b/app/components/badge-github-actions.js
@@ -18,6 +18,23 @@ export default Component.extend({
 
     return url.href;
   }),
+  url: computed('badge.attributes.{repository,workflow_enc,branch,event}', function() {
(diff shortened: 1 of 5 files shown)


sgrif:
Can we call the badge `github_actions` instead of just `github`?

src/tests/badge.rs
```diff
@@ -432,6 +455,40 @@ fn travis_ci_required_keys() {
     assert_eq!(krate.badges(), vec![]);
 }
 
+#[test]
+fn github_required_keys() {
+    {
```
sgrif: Should this just be two tests?

CryZe:
Alright, I made those changes :)

CryZe:
Travis is having issues it seems.

CryZe:
This is done from my side.

jhpratt:
@sgrif ^^ I was actually just looking for this, and @CryZe says it should be good!

CryZe:
Isn't this true for all URLs in general? Shouldn't crates.io properly percentage encode all the bade URL values anyway? I'm not sure how this is different here. Anyways, I think I might need some help how and where I would do this. I don't really know the crates.io codebase, so I might need some mentoring.

I'll open a PR  in the cargo repo.

carols10cents:
[Here's an example](https://github.com/rust-lang/crates.io/blob/be8fc705724484921a6b7b2c79168394f172a638/app/components/badge-appveyor.js#L30) - for the appveyor badges, we have to replace underscores or periods with hyphens.

CryZe:
Alright, I implemented the whitespace escaping.

app/components/badge-github-actions.js
```diff
@@ -0,0 +1,15 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { alias } from '@ember/object/computed';
+
+export default Component.extend({
+  tagName: 'span',
+  classNames: ['badge'],
+  repository: alias('badge.attributes.repository'),
+  workflow: computed('badge.attributes.workflow', function() {
+    return this.get('badge.attributes.workflow').replace(' ', '%20');
```
charmander: This will only replace the first space – it’d have to be `/ /g`. But I agree that this is strange and something like `encodeURIComponent()` would seem to make more sense.

locks:
```suggestion
    return new URL(this.get('badge.attributes.workflow')).href;
```

CryZe:
This should work reasonably well now. ? encodeURIComponent` wouldn't work as / should not be escaped and `new URL(...)` doesn't seem to work for an arbitrary string that isn't really a URL.

charmander:
> `encodeURIComponent` wouldn't work as `/` should not be escaped

Is this trying to support specifying workflows by name, by path, or both? Because if a name includes `/`, it should be escaped.

charmander:
If both, I think the most sensible way to do it is to check [whether it ends with `.yml`/`.yaml`][1] (GitHub seems to do it case-sensitively), and choose between `encodeURIComponent(name)` or `name.split('/').map(encodeURIComponent).join('/')` accordingly.


[1]: https://help.github.com/en/github/automating-your-workflow-with-github-actions/configuring-a-workflow#creating-a-workflow-file

jhpratt:
For those that aren't aware, GitHub Actions will go to general availability this coming Wednesday, so it'd probably to best to get this resolved relatively soon.

CryZe:
Alright, I changed it to @charmander's suggestion (minus the check for whether we believe it's a path). I believe this is good enough to land like this. I think any remaining issues can be fixed in subsequent PRs (if anyone ever really runs into these weird edge cases).

charmander:
Including a slash in a workflow name doesn’t seem that weird, and there’s no way to work around it, since giving a workflow a name seems to block the path syntax. I guess it’s a quick enough change later, though.

Other remark: this can be convinced to do something it probably shouldn’t with `.` and `..` components, but it’s not a security risk or anything. (It’s unclear whether there’s a way to address the badge of workflows named `.` or `..`, but that one *is* definitely weird.)

carols10cents:
I'm setting up a repo with some actions right now to test this, and I see that the documentation for the badge says you can configure the badges to use a particular `branch` and a particular `event` as well. 

We support optional `branch` configuration on other badges, so I think we should definitely have support for that. 

I'm not sure how common `event` configuration in the badges is going to be desired, but it seems like we should offer all the possible parameters.

What do you think?

CryZe:
Oh yeah seems like they've been recently added. I don't think they existed when I created this PR.

carols10cents:
Hm, another thing is that a project can have multiple GitHub actions, but the way we've implemented badges, it's only possible to specify one `github-actions` badge. I'm not sure if this will be an issue? Maybe just something to note in the docs?

gnzlbg:
> Hm, another thing is that a project can have multiple GitHub actions, but the way we've implemented badges, it's only possible to specify one github-actions badge. I'm not sure if this will be an issue?

This would be an issue for libc, where we have different workflows for, e.g., publishing the website, and for running bors tests. I'm not sure which of them would be relevant to show on crates.io, or whether there is a way to query whether all workflows on a particular branch succeeded.

jhpratt:
Any updates?

CryZe:
This should be ready for review again now.

carols10cents:
o.O why is travis using rust beta 1.36??? https://travis-ci.com/rust-lang/crates.io/jobs/269657396#L325 looking into that...

smarnach:
@carols10cents I'm not sure I understand your question correctly. We always build on beta (in addition to building on stable and nightly), and the beta build uses the current beta version of Rust. You can see all build jobs here: https://travis-ci.com/rust-lang/crates.io/builds/142060297

smarnach:
I see, you are wondering about the beta version used. For me the linked lines in the logs read

```
$ rustc --version
rustc 1.41.0-beta.1 (eb3f7c2d3 2019-12-17)
$ rustup --version
rustup 1.21.1 (7832b2ebe 2019-12-20)
$ cargo --version
cargo 1.41.0-beta (626f0f40e 2019-12-03)
```

so it indeed seems to be the current beta version.

carols10cents:
> I see, you are wondering about the beta version used. For me the linked lines in the logs read

I reran the build, which overwrote what I was seeing, oops. Something was going wrong and then rustup was getting stuck on beta 1.36. [Maybe this link is stable?](https://travis-ci.com/rust-lang/crates.io/jobs/269657247#L321) in case it's not, here's what I saw in the logs, it kind of looked like a bug with rustup but I don't see anyone else with an issue like the error I'm seeing of `error: failure removing component 'rustc-x86_64-unknown-linux-gnu', directory does not exist: '"bin/rust-gdbgui"'`:

<details>

```
Installing Rust

8.06s$ curl -sSf https://build.travis-ci.com/files/rustup-init.sh | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION --profile=minimal -y

info: downloading installer

info: profile set to 'minimal'

info: default host triple is x86_64-unknown-linux-gnu

info: updating existing rustup installation

Rust is installed now. Great!

To get started you need Cargo's bin directory ($HOME/.cargo/bin) in your PATH

environment variable. Next time you log in this will be done

automatically.

To configure your current shell run source $HOME/.cargo/env

$ export PATH=${TRAVIS_HOME}/.cargo/bin:$PATH

9.09s$ rustup update

info: syncing channel updates for 'beta-x86_64-unknown-linux-gnu'

info: latest update on 2019-12-18, rust version 1.41.0-beta.1 (eb3f7c2d3 2019-12-17)

info: downloading component 'rustc'

info: downloading component 'rust-std'

info: downloading component 'cargo'

info: downloading component 'rust-docs'

info: removing previous version of component 'rustc'

info: rolling back changes

error: failure removing component 'rustc-x86_64-unknown-linux-gnu', directory does not exist: '"bin/rust-gdbgui"'

info: checking for self-updates

  beta-x86_64-unknown-linux-gnu update failed - rustc 1.36.0-beta.6 (a91f2e56c 2019-06-24)

info: cleaning up downloads & tmp directories

$ rustc --version

rustc 1.36.0-beta.6 (a91f2e56c 2019-06-24)
```

</details>

I cleared the travis cache for this PR and that seems to have fixed it 🤷‍♀ it looks like we have a different problem with nightly now though...

carols10cents:
OH meant to respond to this and never did @gnzlbg

> This would be an issue for libc, where we have different workflows for, e.g., publishing the website, and for running bors tests. I'm not sure which of them would be relevant to show on crates.io, or whether there is a way to query whether all workflows on a particular branch succeeded.

I don't see a way to configure a GitHub Actions badge to be the combination of multiple workflows, but it turns out rust and this repo need bors to be able to see if all workflows passed or not, so there's [one that checks if all the others completed successfully](https://github.com/rust-lang/crates.io/blob/45b8dbf2a3d4294ed5d60ed4fd85f7c4d679bbd1/.github/workflows/ci.yml#L192-L216). 

Given that workaround exists, I think it's fine for now to leave the badge implementation the way we have it. We'll keep this in mind to see if more cases of wanting multiple badges from the same service come up and re-evaluate then.

gnzlbg:
@carols10cents so if you are using bors, does that workflow only run in the auto branch ?

carols10cents:
> @carols10cents so if you are using bors, does that workflow only run in the auto branch ?

No, it runs on all PRs too (you can see it on this one).

djc:
We might want to follow this up with updates to the Cargo manifest documentation, here: https://github.com/rust-lang/cargo/blob/741f0e586a74f65c10200ae42b13bceea40d5957/src/doc/src/reference/manifest.md.

carols10cents:
@djc 

> We might want to follow this up with updates to the Cargo manifest documentation, here: [rust-lang/cargo:src/doc/src/reference/manifest.md@`741f0e5`](https://github.com/rust-lang/cargo/blob/741f0e586a74f65c10200ae42b13bceea40d5957/src/doc/src/reference/manifest.md).

There is already an open PR to do that https://github.com/rust-lang/cargo/pull/7547

flip1995:
ping @CryZe do you have time to wrap up this PR? If not, I'd be happy to help.

CryZe:
I unfortunately don't, so feel free to finish this. I sent you an invite for push access to the branch.

flip1995:
I'm a JS noob, so I hope I understood everything correctly.

I applied @charmander suggestion (https://github.com/rust-lang/crates.io/pull/1838#discussion_r362352443) in 9e458d6

> but the encoded version is also used for the alt and title text as well

fixed in 0daadf0

> The badge always links to /actions even if there are branch or event parameters specified.

addressed in c340a1c

> The alt/title text doesn't use the branch or event parameters;

fixed in 2e164b2. I left the workflow name in the text, since with GHA the workflow at hand is at least as important as the branch IMO. I can change this though.

app/components/badge-github-actions.js
```diff
@@ -0,0 +1,56 @@
...
+      url.searchParams.set('event', this.event);
+    }
+
+    return url.href;
+  }),
+  url: computed('badge.attributes.{repository,workflow_enc,branch,event}', function() {
+    const url = new URL(`https://github.com/${this.repository}/actions`);
+
+    if (this.workflow_enc !== '') {
+      url.searchParams.set('workflow', this.workflow_enc);
```
charmander: It looks like this’ll double-encode the workflow.

```suggestion
      url.searchParams.set('workflow', this.workflow);
```

(and the condition too for neatness even though it behaves the same)

app/components/badge-github-actions.js
```diff
@@ -0,0 +1,56 @@
...
+    return this.get('badge.attributes.workflow');
+  }),
+  workflow_enc: computed('badge.attributes.workflow', function() {
+    return this.get('badge.attributes.workflow')
+      .split('/')
+      .map(encodeURIComponent)
+      .join('/');
+  }),
+  branch: computed('badge.attributes.branch', function() {
+    return encodeURIComponent(this.get('badge.attributes.branch') || 'master');
```
charmander:
```suggestion
    return this.get('badge.attributes.branch') || 'master';
```

app/components/badge-github-actions.js
```diff
@@ -0,0 +1,56 @@
...
+  }),
+  event: computed('badge.attributes.event', function() {
+    return encodeURIComponent(this.get('badge.attributes.event') || '');
```
charmander:
```suggestion
    return this.get('badge.attributes.event') || '';
```

app/components/badge-github-actions.js
flip1995: According to [this doc](https://developer.mozilla.org/en-US/docs/Web/API/URL) `url.searchParams` is `readOnly`. Are you sure, that it can be used like this?

app/components/badge-github-actions.js
flip1995: I think we want to encode branch names, for branches with `/` in them.

flip1995: Ah wait, `/` works in queries. And we also don't want to display the branch name encoded in the (alt-)text.

app/components/badge-github-actions.js
charmander: The property is read-only:

```
> new URL('http://example.com/').searchParams = new URLSearchParams()
Uncaught:
TypeError: Cannot set property searchParams of [object URL] which has only a getter
```

but its value, the `URLSearchParams`, isn’t.

app/components/badge-github-actions.js
charmander: Branch names do need to be encoded, but `URLSearchParams` takes care of that. (It’s the name that needs the explicit `encodeURIComponent` because it’s a path component instead of a part of the params.)

app/components/badge-github-actions.js
flip1995: I just tested this and the format produced by the `searchParams` isn't accepted by GitHub. We need this format: `query=workflow%3A"Say+hello%2Fwhatever"+branch%3Abadges-test+event%3Apush`

flip1995:
ping @carols10cents. This should be ready for another round of review.

app/components/badge-github-actions.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    return url.href;
+  }),
+  url: computed('badge.attributes.{repository,workflow,branch,event}', function() {
+    const url = new URL(`https://github.com/${this.repository}/actions`);
+
+    let query = '';
+    if (this.workflow !== '') {
+      query += `workflow:"${this.workflow}"`;
```
charmander: For anyone who might have been wondering: quotes can be escaped as `\"`, but there doesn’t seem to be a way to represent a query where multiple quoted terms end with a backslash. Because out of these three, only workflow names can end with backslashes, it *is* possible to support every workflow name by:

- escaping double quotes
- putting the `workflow:` part last

I’m sure trailing backslashes aren’t considered important enough to support here, but how about ASCII double quotes in workflow and branch names?

CryZe:
Looks like the review fell under the radar. So consider this a friendly ping :)

djc:
@carols10cents friendly ping? Would be really nice to have this.

DarkEld3r:
Don't want to be rude, but is there a chance this is going to be merged anytime soon?

carols10cents:
Whether you're trying to be rude or not, you are being rude.

I'm not getting paid to work on crates.io right now, so I'm prioritizing paid work.

Also, there's a global pandemic going on. So you all need to chill.

jhpratt:
@carols10cents I don't think a person asking the status of things after roughly a month of no action is rude. We fully realize that neither you nor anyone else gets paid to work on crates.io. All we were seeking was a status update, whether from you or someone else. A simple statement that you (and others) are preoccupied with staying healthy (hopefully!) and your paid job would have sufficed. It's something we all understand, to say the least.

---

Somewhat related, I'm home myself and am more than willing to help out. If there are updates to code that need to be done, I can (likely) do that. I can also review if desired.

carols10cents:
> I don't think a person asking the status of things after roughly a month of no action is rude. We fully realize that neither you nor anyone else gets paid to work on crates.io. All we were seeking was a status update, whether from you or someone else.

All status changes are made to the issue. If there is no update here, there has been no change in status. I do think it's rude.

carols10cents:
Next time I get a chance to review PRs, I'll be reviewing them in "Least Recently Updated" order. My current queue: https://github.com/rust-lang/crates.io/pulls?q=is%3Aopen+is%3Apr+assignee%3Acarols10cents+sort%3Aupdated-asc

djc:
@carols10cents I've been thinking about your comments a lot since yesterday. Of course you decide how to use your unpaid time. That said, in my experience across many open source projects, asking for a status update or pinging a non-responsive maintainer (and I've been on that side plenty of times myself) is often considered fine -- sometimes issues get lost in the noise, and it's useful to highlight those that need action. Being explicit about a lack of time (whether paid or not) is useful and would likely have led to fewer comments about this, and could maybe also elicit discussion about whether or not another maintainer could take over your review.

However, I feel dismayed about your last comment. In saying that you'll review PRs in LRU order, I worry that you use your very scarce time for things that are arguably of lesser priority for crates.io users, which I feel would be unfortunate.

Dylan-DPC:
Just a visitor here (actually a subscriber who doesn't know how I landed on this issue), but I would recommend let's not discuss this things here. Thanks

BurntSushi:
**Moderation note:** @carols10cents has made it clear that asking for further status updates is unwanted. Similarly, engaging into a meta discussion about this is also not productive. Please respect the maintainer's wishes and stop.

JohnTitor:
I'm going to close this PR since we decided to deprecate the badges feature in the team meeting yesterday. See #2436 and related issues for more detail. Thanks for taking your time!

PR #1840, opened 2019-09-15
Add missing foreign key constraint for columns badges.crate_id.

smarnach:
Since we may already have entries in the `badges` table referring to crates that don't exist, we
need to delete these entries before creating the foreign key constraint. This is a destructive
operation in the sense that it cannot be reverted, but it should only delete rows that should have
been deleted together with the crates they belong to.

We have a test to verify that all columns called `crate_id` are actually foreign keys referring to
`crates.id`. However, the query for the relevant columns contains the filter `contype = 'f'`,
effectively limiting the result to columns that already have foreign key constraints. This change
fixes the query to also allow `contype IS NULL`.

In addition, I modified the query to only verify tables in the schema `public`. This is useful for
an integration test for the database dumps in #1800.

### DIFF
--- a/migrations/2019-09-15-194259_add_badges_crate_id_foreign_key/down.sql
+++ b/migrations/2019-09-15-194259_add_badges_crate_id_foreign_key/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE badges DROP CONSTRAINT fk_badges_crate_id;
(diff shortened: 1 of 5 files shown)


sgrif:
I've confirmed that all rows in production to be deleted are in fact rows that should have already been deleted.

@bors: r+

PR #1841, opened 2019-09-18
Merge prod

carols10cents:
There's a commit on heroku that's not on github, merging it back in.

Any reason not to do this @sgrif ?


sgrif:
The last commit is a merge commit of https://github.com/rust-lang/crates.io/pull/1815/commits/74f6f7d25e9661467f0921a45584993aa0df29e5, which means I probably forgot to put `-f` when deploying the second commit of that PR. I will do so now.

sgrif:
There should no longer be a commit on prod which is not on master

carols10cents:
Thanks!

PR #1843, opened 2019-09-19
Travis cache fixes

jtgeibel:
Build times have regressed and we are spending a lot of time doing caching stuff (and downloading/uploading 8GB of compressed data).  By manually wiping out the cache, stable and beta build times should align more closely with nightly (which is cleaned on every new nightly release).  This PR also fixes the list of pruned binaries:

* Remove update-downloads
* Add enqueue-job
* Sort alphabetically for easier comparison against `ls src/bin`

### DIFF
--- a/script/ci/prune-cache.sh
+++ b/script/ci/prune-cache.sh
@@ -7,7 +7,7 @@ du -hs target/debug
 
 crate_name="cargo-registry"
 test_name="all"
-bin_names="delete-crate delete-version populate render-readmes server test-pagerduty transfer-crates update-downloads background-worker monitor"
+bin_names="background-worker delete-crate delete-version enqueue-job monitor populate render-readmes server test-pagerduty transfer-crates"
 
 normalized_crate_name=${crate_name//-/_}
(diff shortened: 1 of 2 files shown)


jtgeibel:
Merging so that I can clear some caches on CI and then restart a job that is timing out on cache unpacking.

@bors r+

script/ci/prune-cache.sh
```diff
@@ -7,7 +7,7 @@ du -hs target/debug
 
 crate_name="cargo-registry"
 test_name="all"
-bin_names="delete-crate delete-version populate render-readmes server test-pagerduty transfer-crates update-downloads background-worker monitor"
+bin_names="background-worker delete-crate delete-version enqueue-job monitor populate render-readmes server test-pagerduty transfer-crates"
```
smarnach: We could generate this list to avoid having to manually maintain it, e.g.

    cargo metadata --format-version 1 |
        jq -r '.packages[] | select(.name == "cargo-registry") | .targets[] | select(.kind == ["bin"]) | .name'

PR #1844, opened 2019-09-19
Add indexes in migrations that have already been running in production

carols10cents:
@sgrif added these to the production database by hand for testing
purposes a while back; they seem to be performing well so they should
get added to the codebase to keep the schema in sync.

r? @sgrif, feel free to wait to merge this until you've checked the bloat you mentioned wanting to check; I'm happy to update this PR if needed. Just wanted to have something around as a reminder.


sgrif:
Both indexes look good, the brin index is way lossier than I'd like, but it's still providing some value and is basically free.

carols10cents:
ok, so is this good to r+?

PR #1845, opened 2019-09-19
Migration to remove an index already removed from prod by hand

carols10cents:
This index doesn't exist in production currently, but it still exists in a db created from this codebase because we never made a migration to remove it.

Should be a no-op on prod so can be merged and deployed at any time.


sgrif:
I suspect the reason it was removed is because it would be unused as it's redundant with the `unique_num` index. I'm not sure why that wasn't done with a migration, as it definitely should have been.

PR #1852, opened 2019-09-24
Fix GitHub relative SVGs not rendering properly

### LINKED ISSUE #1851: SVGs from GitHub aren't rendered correctly in README
By default, GitHub serves raw SVGs with the `Content-Type` header set to `text/plain`, which prevents the images from being displayed properly. This can be fixed easily by adding `sanitize=true` to the query string.

GitHub does it automatically when rendering READMEs, so the issue isn't present on their side and it might be a good idea to replicate that behaviour. (https://github.com/raftario/licensor vs https://crates.io/crates/licensor)

The fix is pretty trivial so I plan to open a PR for it tomorrow.

raftario:
Fixes #1851 where SVGs from GitHub aren't rendered in READMEs if they use relative links.

First PR, so if I did something wrong point it out!

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -148,15 +148,39 @@ impl SanitizeUrl {
     }
 }
 
+/// Groups media-related URL info
+struct MediaUrl {
+    is_media: bool,
+    add_sanitize_query: bool,
+}
+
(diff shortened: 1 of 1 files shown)


fundon:
Need merge it.

PR #1853, opened 2019-09-25
Update ember-prism dependency.

### LINKED ISSUE #1468: bug in markdown syntax highlighter
The highlighter seems to give keywords a higher precedence than strings. Take a look at the code block under "repositories" [here](https://crates.io/crates/hubcaps):

badly highlighted code

The source is:

```markdown
    ```rust
    let repo = github.repo("user", "repo");
    ```
```

But it seems to be highlighting `use` as a keyword instead of `"user"` as a string.

locks:
Fixes #1468.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5019,9 +5019,9 @@
       "dev": true
     },
     "clipboard": {
-      "version": "1.7.1",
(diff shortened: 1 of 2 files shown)


locks:
Should be updated and good to go!

Issue #1638, opened 2019-02-27
A wide table in the README breaks the layout


RazrFalcon:
Example: https://crates.io/crates/roxmltree

Chrome 72

![1551280638](https://user-images.githubusercontent.com/725494/53501597-1f4ad400-3ab5-11e9-9d51-ec49b468a60c.png)

andyfleming:
A similar example on https://crates.io/crates/dirs

![Screen Shot 2019-03-11 at 2 50 17 PM](https://user-images.githubusercontent.com/721038/54160529-058e8280-440d-11e9-9b35-ae65710f85de.png)

andyfleming:
A possible solution is to add `word-break: break-word;`.

```diff
 .crate-info .crate-readme table {
     border-collapse: collapse;
+    word-break: break-word;
 }
```

![Screen Shot 2019-03-11 at 2 56 19 PM](https://user-images.githubusercontent.com/721038/54161077-5652ab00-440e-11e9-97b1-746f4895b495.png)

Without changing the layout, we could also decrease the font size and padding to make it a little bit more readable.

```diff
 .crate-info .crate-readme table {
     border-collapse: collapse;
+    word-break: break-word;
+    font-size: 14px;
 }

.crate-info .crate-readme table td, .crate-info .crate-readme table th {
    border: 1px solid #dfe2e5;
-    padding: 6px 13px;
+    padding: 3px 6px;
}
```

![Screen Shot 2019-03-11 at 3 03 31 PM](https://user-images.githubusercontent.com/721038/54161265-dc6ef180-440e-11e9-98c3-e4fc5bb5d4a8.png)

I'm not sure it's the ideal solution, but it may be a good short-term fix.

hbina:
Hello! I would like to work on this.
So I think what I have to do is to find the place where crates place the CSS stuff and sneak in the changes @andyfleming suggested?
Any pointers are appreciated.
Thanks!

hbina:
I have clone and set up my mirror of crates.io here https://akarin-crates.herokuapp.com/ but its empty, do i have to manually publish crates that I want to test out. Or did I do something wrong?

smarnach:
@hbina Thanks for looking into this issue!

[...]

PR #1856, opened 2019-09-28
Updated crate.scss

hbina:
1. Fixed table from overflowing when the width is too large.
2. Removed a pointless CSS property. See https://developer.mozilla.org/en-US/docs/Web/CSS/float

`inline-block` is ignored due to the `float`.
If `float` has a value other than `none`, the box is floated and `display` is treated as `block`.

Fixes: #1638

### DIFF
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -343,7 +343,6 @@
             border-collapse: collapse;
             display: block;
             overflow-x: auto;
-            white-space: wrap;
 
             th, td {
                 border: 1px solid #dfe2e5;
@@ -469,6 +468,7 @@
     .small { margin-left: 20px; display: inline-block; }
     a.arrow {
         display: inline-block;
+        float: right;
     }
 }
 
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -341,6 +341,9 @@
 
         table {
             border-collapse: collapse;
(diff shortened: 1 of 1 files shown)


app/styles/crate.scss
```diff
@@ -466,7 +469,6 @@
     .small { margin-left: 20px; display: inline-block; }
     a.arrow {
         display: inline-block;
-        float: right;
```
smarnach: Removing this does change the semantics – the arrow won't be a float anymore, and it's not moved to the right side of the containing element anymore. It does not appear to make a difference on modern browsers because the containing element has `justify-content: space-between;`, which also results in moving the arrow to the right, but I'd be careful and just leave this in.

app/styles/crate.scss
```diff
@@ -341,6 +341,9 @@
 
         table {
             border-collapse: collapse;
+            display: block;
+            overflow-x: auto;
+            white-space: wrap;
```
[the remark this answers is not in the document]
hbina: Fixed.

app/styles/crate.scss
hbina: Fixed.

hbina:
Should I merge these commits?

jtgeibel:
LGTM!

@bors r+

PR #1857, opened 2019-09-29
Fixed versions suggestion

hbina:
Previously, if it had searched everything and found nothing. However, this also means it cannot find `maxVersion`, which was what it was set to before.
Thus, this commit's solution is to provide another fallback when there are no stable and non-yanked versions available. It will simply find anything that is not yet yanked even if it is unstable.
If even this fails, then we truly have nothing to offer thereby offering an empty page.

The code could be improved imo, but I am not especially well-versed in JavaScript.

### DIFF
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -11,7 +11,7 @@ export default Route.extend({
 
   flashMessages: service(),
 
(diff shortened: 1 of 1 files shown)


locks:
Thanks for the help! The code reads straightforward and I appreciate the comments so I'm approving :)

PR #1864, opened 2019-10-10
Switch to async/await for bin/server and tests

jtgeibel:
This PR series converts the future combinations to async/await syntax in the server binary (when `USE_HYPER=1` is set) and the recording HTTP proxy used in tests.

This change requires the update of several dependencies at once (see the list below).  It is not strictly necessary to update `reqwest` at this time, but I'm doing so in this series to avoid the build time regression of pulling in `hyper 0.12`, `hyper 0.13` and their associated transitive dependencies.

Blockers:

- [x] async/await is stable in 1.39.0 :tada:
- [x] Bump the rust release in `RustConfig`
- Stable release of underlying stack
  - [x] `futures 0.3.1`
  - [x] `tokio 0.2.9`
  - [x] `hyper 0.13.1` and `hyper-tls 0.4.1`
  - [x] `reqwest 0.10.1`
  - [x] `conduit-hyper 0.2.0`


jtgeibel:
:tada: I've rebased this branch and now consider it ready for review. :tada: 

There is currently a missing feature in conduit-hyper which should be resolved before we try rolling out `USE_HYPER=1` to production again.  That can be handled in a future PR and I think we can land this now for the improvements to the test harness.

r? @carols10cents

jtgeibel:
Rebased to resolve merge conflicts, and added a commit to switch from git master to the now released `reqwest v0.10.0`.

jtgeibel:
The build failure is a regression on nightly and is already reported as rust-lang/rust#68264.

Cargo.toml
```diff
@@ -75,21 +75,22 @@ conduit-router = "0.8"
...
 handlebars = "2.0.1"
 
 [dev-dependencies]
 conduit-test = "0.8"
-hyper-tls = "0.3"
+hyper-tls = "0.4"
 lazy_static = "1.0"
-tokio-core = "0.1"
 diesel_migrations = { version = "1.3.0", features = ["postgres"] }
+tower-service = "0.3.0-alpha.2"
```
carols10cents: 0.3.0 non-alpha is out! https://crates.io/crates/tower-service/0.3.0

jtgeibel: Good catch.  I've pushed a new commit.

I had already bumped the `tower-service` version upstream, so no change to the lockfile from that.  I've also decided to drop the alpha from `conduit-hyper`, a release that includes a new test!  I was considering a few more tweaks to the API, but there is always a `0.3.0` if needed.

carols10cents:
I'm having trouble signing in locally when I'm on this branch rebased on master at 70723050585e70018c92e422988626a9583d78e3 (and I verified that when I'm on 70723050585e70018c92e422988626a9583d78e3 locally I *can* sign in), even without `USE_HYPER=1`. 

It appears to be failing [when requesting the user info from GitHub](https://github.com/rust-lang/crates.io/blob/c6d13eba5b35cbc585e84f5ee744c03ecf51f995/src/github.rs#L16-L32); the request is returning 403. I'm able to print out the `Authorization` token value that I'm trying to use locally (`println!("auth secret = {}", auth.secret());` [as in here](https://github.com/rust-lang/crates.io/blob/c6d13eba5b35cbc585e84f5ee744c03ecf51f995/src/github.rs#L26)), and use it in a `curl` request to GitHub that returns successfully:

```
curl -H "Accept: application/vnd.github.v3+json" -H "Authorization: token [redacted]" https://api.github.com/user
```

Soooo 🤷‍♀ ???

jtgeibel:
@carols10cents excellent catch!  So it turns out reqwest no longer sets a default user-agent header.  I've rebased and then added a commit which sets the user-agent to our URL and updates the tests to always check for this value.

I'm pushing the changes to staging now, and will also set `USE_HYPER=1` there.

carols10cents:
2 issues we investigated tonight:

- The default number of threads used in development mode needs to be increased to make the site usable locally (I tried 4 and it seemed to work ok)
- Using `thread::sleep` in a controller function appeared to be blocking other requests, and this wasn't the expected behavior when using `tokio::task::spawn_blocking`

carols10cents:
If those two issues get tracked somewhere and block turning on hyper for everyone everywhere, feel free to r=me :)

jtgeibel:
I've opened #2204 as a tracking issue for any remaining blockers for enabling `USE_HYPER=1` in production.

@bors r=carols10cents

PR #1870, opened 2019-10-15
Allow rowspan and colspan in markdown table rendering

### LINKED ISSUE #1863: README rendeder scrubs out layout-affecting tags
Table in readme for [psm 0.1.5](https://crates.io/crates/psm/0.1.5) has had its colspan and rowspan attributes removed, breaking the table layout.

These attributes should not be removed.

jtescher:
Allow `rowspan` and `colspan` attributes of forms in READMEs.

Fixes #1863

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -67,6 +67,8 @@ impl<'a> MarkdownRenderer<'a> {
             ("a", hashset(&["href", "id", "target"])),
             ("img", hashset(&["width", "height", "src", "alt", "align"])),
             ("input", hashset(&["checked", "disabled", "type"])),
+            ("th", hashset(&["rowspan", "colspan"])),
+            ("td", hashset(&["rowspan", "colspan"])),
         ]);
         let allowed_classes = hashmap(&[(
             "code",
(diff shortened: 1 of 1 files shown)


smarnach:
@jtescher Thanks for your contribution! This change looks good to me, but I don't have access to merge and deploy it – some other member of the team will do so soon.

The only thing I'm wondering about is why we don't use [ammonia's defaults](https://github.com/rust-ammonia/ammonia/blob/v2.0.0/src/lib.rs#L257-L329) – they look reasonable to me, and would avoid a bunch of other rendering issues. This is of course unrelated to this PR, and in my opinion we can merge this as it is.

CC @jtgeibel

sgrif:
I'm actually a bit confused about why percy is failing this, this PR definitely shouldn't have affected download count graphs

jtgeibel:
I restarted the Frontend job and Percy is happy now, so I'm guessing this was a transient issue on their end.

@smarnach Looking just at the `tag_attributes` variable, it looks like we additionally allow `id` and `target` on `a` tags.  I think `id` is there to allow hyperlinks to work within the README file.  `target` allows the README to open a link in a new tab.  Maybe it would be more reasonable to start with the defaults and then explicitly allow (and document) any such deltas.  Would you like to open an issue to track this?

I agree these changes are good to merge.
@bors r+

smarnach:
I was assuming that the differences in the chart were caused by changes to Google Charts. We use the deprecated `jsapi` loader that always pulls the latest version of Google Charts, so I wasn't surprised there might be small changes to how the chart looks.

@jtgeibel Yes, we currently do allow a few additional things, e.g. the `input` element. Since the ammonia API allows amending the defaults with functions like [`add_tag_attributes()`](https://docs.rs/ammonia/2.0.0/ammonia/struct.Builder.html#method.add_tag_attributes), it would be easy to add these items, but we don't need to put the full list in our code.

I will open an issue for this for further discussion.

PR #1871, opened 2019-10-16
Upload all content to s3 as public and immutable

### LINKED ISSUE #1826: static.crates.io cache control
Crate files on static.crates.io don't have any explicit cache lifetime set. Please consider adding caching information, such as `Cache-control: max-age=31536000, immutable`.

Even though Cargo itself has a non-HTTP cache and doesn't use that information, it would be helpful for users who put a caching proxy in front of crates.io infrastructure.

```sh
curl -I 'https://static.crates.io/crates/winapi/winapi-0.2.5.crate'
```

```text
HTTP/2 200
content-type: application/x-tar
content-length: 478113
date: Thu, 05 Sep 2019 13:21:39 GMT
last-modified: Mon, 09 Nov 2015 21:19:48 GMT
etag: "83c818a44aed9494e7a9f8fa52acb789"
x-amz-version-id: null
accept-ranges: bytes
server: AmazonS3
x-cache: Miss from cloudfront
via: 1.1 21e3028cc4bd3f3af310d648f3748004.cloudfront.net (CloudFront)
x-amz-cf-pop: LHR61-C2
x-amz-cf-id: Qp_nTqbJJXOt85twQ53LRqUcMPu00Vl_D3cajrAItIgF-h-6jtFSBQ==
```

jtescher:
Adds cache control header with value `public,max-age=31536000,immutable` to all future crate and README uploads.

Fixes #1826

### DIFF
--- a/src/bin/render-readmes.rs
+++ b/src/bin/render-readmes.rs
@@ -133,7 +133,14 @@ fn main() {
                 let readme_path = format!("readmes/{0}/{0}-{1}.html", krate_name, version.num);
                 config
                     .uploader
-                    .upload(&client, &readme_path, content, content_length, "text/html")
+                    .upload(
+                        &client,
+                        &readme_path,
+                        content,
+                        content_length,
+                        "text/html",
+                        None,
+                    )
                     .unwrap_or_else(|_| {
                         panic!(
                             "[{}-{}] Couldn't upload file to S3",
--- a/src/s3/lib.rs
+++ b/src/s3/lib.rs
@@ -45,6 +45,7 @@ impl Bucket {
         content: R,
         content_length: u64,
         content_type: &str,
+        extra_headers: Option<header::HeaderMap>,
     ) -> reqwest::Result<reqwest::Response> {
         let path = if path.starts_with('/') {
             &path[1..]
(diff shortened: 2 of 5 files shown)


jtescher:
Ah thanks for the context @smarnach! I'll update this to set the appropriate headers for both scenarios.

jtescher:
@smarnach I reworked the PR to be more conservative, currently only marking crate data immutable. It seems like rendered readme file data may also be altered at a later point via [render-readmes.rs](https://github.com/rust-lang/crates.io/blob/master/src/bin/render-readmes.rs) (or are the cloudfront caches purged currently when this occurs? If so I can add the header there too)

If you know the frequency at which the database dump job is run I'm happy to add the expires header as well.

smarnach:
I tested this on a setup similar to production, and everything is working as expected. There are two things currently blocking me from merging this:

* We are currently making some changes to our infrastructure that are probably orthogonal to this change, but I would like to avoid making this change at the same time.
* I need to figure out whether there are legal implications. We sometimes get requests to take down crates for legal reasons. With this change applied, these crates may still be cached for another year. We probably need to invalidate the file on the CloudFront edge caches in these cases, but I need to discuss this with the team.

PR #1873, opened 2019-10-18
Show crate last updated at on search results

### LINKED ISSUE #1597: Show last updated timestamp in search results
Currently it's cumbersome to check whether a crate is still being actively developed or not when searching. To see when the crate was last updated, you have to move away from the search results page and do that for every crate individually.

jtescher:
Adds the crate's last updated at time to search results on the right:

image

And with hover state to display exact date and time:

image

Happy to move it to a different location if preferrable, reasonable alternatives could be at the end of the quick links? or under the crate title?

Fixes #1597

### DIFF
--- a/app/styles/crate.scss
+++ b/app/styles/crate.scss
@@ -155,6 +155,15 @@
         @include display-flex;
         @include align-items(center);
     }
+    .updated-at {
+        padding-top: 5px;
+        @include display-flex;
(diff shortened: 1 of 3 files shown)


carols10cents:
This looks great, thank you so much! Sorry for the delay on reviewing :)

@bors: r+

PR #1874, opened 2019-10-19
Fix error response payload parsing

### LINKED ISSUE #1547: Ownership invitations to blocked groups don't error in UI
When adding an owner in the form of `github:org:team`, you may get an error on the command-line such as:

```
error: failed to add owners to crate mbedtls: api errors: It looks like you don't have permission to query a necessary property from Github to complete this request. You may need to re-authenticate on crates.io to grant permission to read github org memberships. Just go to https://crates.io/login
```

However, if you do the same action using the UI, you just get a notification that "an invitation has been sent".

jtescher:
Caught request error objects do not have a payload property, this should start properly showing error details instead of the generic fallback message.

Fixes #1547

### DIFF
--- a/app/components/email-input.js
+++ b/app/components/email-input.js
@@ -100,9 +100,9 @@ export default Component.extend({
         await ajax(`/api/v1/users/${user.id}/resend`, { method: 'PUT' });
         this.set('disableResend', true);
       } catch (error) {
-        if (error.payload) {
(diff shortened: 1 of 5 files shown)


jtgeibel:
Thanks @jtescher, this looks good to me!  Thanks for the PR.

@bors r+

PR #1875, opened 2019-10-21
Clear search input on index route

### LINKED ISSUE #1230: Search field not cleared back when going back into history
## Steps to reproduce

 1. Go to crates.io;
 2. Type `rayon` and press <kbd>Enter</kbd>;
 3. Go to previous page.

## Expected result

The search field is empty.

## Actual result

The search field still contains `rayon`.

jtescher:
Not sure what the correct product choice is here, but this clears the search input when navigating back to index route if that is in fact the behavior that is preferred.

Fixes #1230 

(for some context both npm and rubygems.org do NOT clear the index if you search and then navigate back via browser, but both clear if you navigate forward to the index route via top left icon clicks)

### DIFF
--- a/app/routes/index.js
+++ b/app/routes/index.js
@@ -14,6 +14,7 @@ export default Route.extend({
   },
 
   setupController(controller) {
+    this.controllerFor('application').set('searchQuery', null);
     controller.dataTask.perform();
   },
 });
--- a/app/routes/search.js
+++ b/app/routes/search.js
@@ -14,6 +14,7 @@ export default Route.extend({
   },
 
   setupController(controller, params) {
+    this.controllerFor('application').set('searchQuery', params.q);
     controller.dataTask.perform(params);
   },
 });
--- a/tests/acceptance/search-test.js
+++ b/tests/acceptance/search-test.js
@@ -62,6 +62,20 @@ module('Acceptance | search', function(hooks) {
     assert.dom('[data-test-search-nav]').hasText('Displaying 1-8 of 8 total results');
   });
 
+  test('clearing search results', async function(assert) {
+    this.server.loadFixtures();
+
+    await visit('/search?q=rust');
+
+    assert.equal(currentURL(), '/search?q=rust');
+    assert.dom('[data-test-search-input]').hasValue('rust');
+
+    await visit('/');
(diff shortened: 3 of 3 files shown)


carols10cents:
Hm, I'm not sure this is what we want. The behavior you described is working as you intend, but it's leading to some unexpected results. When trying this out locally with `npm run start:local`, if i:

- Visit http://localhost:4200/
- Search for "test", which navigates me to http://localhost:4200/search?q=test, and "test" is in the search query box as expected
- Use the browser back button, which navigates me to http://localhost:4200/ and empties the search query box as expected with this PR's change
- Use the browser forward button, which navigates me to http://localhost:4200/search?q=test but the search query box remains empty. I expected "test" to be in the search box again.

Another unexpected scenario:

- Visit http://localhost:4200/
- Search for "test", which navigates me to http://localhost:4200/search?q=test, and "test" is in the search query box as expected
- Click on the "Browse all crates" link at the top of the page, which navigates me to http://localhost:4200/crates, and "test" is still in the search query box-- with the intention of this PR's change, I kind of expected the search query box to be emptied

Is it possible to set the search query to what's in the current URL's `q` parameter (if there is one) instead of `null` unconditionally? And is it possible to set the search query in this manner for all routes, not only `/`?

What do you think?

jtescher:
I can do that if it the right product choice, could also decide to close this pr and the underlying issue if it is working as intended currently.

carols10cents:
I think we should make a change to the current behavior, if you're up for continuing to work on this. Thanks!

jtescher:
@carols10cents ok great! rebased, now explicitly setting the `searchQuery` both index and search seems to fix the issues you were bringing up. When I run this locally I can now navigate back and forward and the search terms are as expected. Navigating through clicking top left clears the query, also reloading the page at any point does not mess things up.

Let me know if there are other things you'd like to see here 👍

carols10cents:
Awesome!!! This looks great, thank you so much!

I tested this in Firefox, Chrome, and Safari and it's working as I would expect. Firefox Nightly sometimes is only changing the URL and not the page when I navigate back, but I expect it'll work tomorrow 😆

@bors r+

Issue #1872, opened 2019-10-16
Base list of allowed HTML tags and attributes on the Ammonia defaults


smarnach:
Our readme renderer uses [Ammonia](https://docs.rs/ammonia/2.0.0) to sanitise the HTML. We currently explicitly [list all HTML tags and attributes we allow in the source code](https://github.com/rust-lang/crates.io/blob/af26bb3ab2dceb4810f0edce145d43ec484b7028/src/render.rs#L25-L91). To simplify this code, and to allow some additional safe attributes we currently don't allow, we could rely on [Ammonia's defaults](https://github.com/rust-ammonia/ammonia/blob/v2.0.0/src/lib.rs#L257-L329) instead, and [add any additional items we may need](https://docs.rs/ammonia/2.0.0/ammonia/?search=add).

Suggested steps:

* Consider upgrading to the latest Ammonia release.
* Compare the list of currently allowed tags in crates.io to the defaults in Ammonia.
* Update the code to in `MarkdownRenderer::new()` (linked above) to only add these additional tags to the defaults.

AZanellato:
I can try this :smile:

PR #1876, opened 2019-10-22
Ammonia 3.0 upgrade

AZanellato:
Closes https://github.com/rust-lang/crates.io/issues/1872.

I have followed the instructions on the issue. Basically removing that list of allowed elements and relying on ammonia to do the job and also upgrading ammonia to 3.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,15 +31,15 @@ dependencies = [
 
 [[package]]
 name = "ammonia"
-version = "2.0.0"
+version = "3.0.0"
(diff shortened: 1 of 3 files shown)


smarnach:
Thanks for the pull request! Could you also look into the list of tag attributes? The motivating bug was that we did not allow some attributes that are relevant for table layout, but there are more attributes that we could safely allow.

AZanellato:
Of course! Will do that later today :smile:

AZanellato:
@jtgeibel done! :)

PR #1880, opened 2019-10-26
Replace jQuery logic with native DOM

locks:


app/components/rl-dropdown.js
```diff
@@ -15,26 +14,17 @@ export default Component.extend({
...
-  propagateClicks: true,
-
   click(event) {
-    let closeOnChildClick = this.closeOnChildClick;
-    let propagateClicks = this.propagateClicks;
-    let $target = $(event.target);
-    let $c = this.$();
+    let closeOnChildClick = 'a:link';
+    let $target = event.target;
+    let $c = this.element;
```
kzys: Both of them are not jQuery objects anymore. It would be better to remove `$` from the names.

locks: I left them as is to minimize the changes in the code. Also, this code will likely be changed or moved in a subsequent pass.

carols10cents:
Looks great! Locally, in Firefox, I tested: 

- API token input focusing 
- Sorting dropdown opening and closing
- Readme syntax highlighting
- Readme anchors

and everything worked as expected. Thanks!!!

@bors r+

PR #1881, opened 2019-10-26
replace triple mustache with html-safe helper

locks:


smarnach:
What is the goal of this change? There is [another occurrence of triple braces here](https://github.com/rust-lang/crates.io/blob/b7951cdc7d9ae36d0c2b56de4a122f15d18481e4/app/templates/crate/owners.hbs#L60); should that be replaced as well?

locks:
I have replaced the other occurrence :) The idea is that we should move away from triple mustaches. Since we still need to interpolate some HTML–which we kinda control anyway–the general advice is to use the `htmlSafe` utility instead.

locks:
I believe this is a lint starting with some version of ember.js, hence my change.

app/templates/crate/owners.hbs
```diff
@@ -57,7 +57,11 @@
         {{/link-to}}
       </div>
       <div class='stats'>
-        {{{ if user.email user.email "&nbsp;" }}}
+        {{#if user.email}}
+          {{html-safe user.email}}
```
smarnach: Does the user email really need `html-safe`? I believe the triple braces were only there to prevent `&nbsp;` from being escaped, and if you expand it the way you did we don't need it anymore.

app/templates/components/crate-readme.hbs
```diff
@@ -1 +1 @@
-{{{rendered}}}
+{{html-safe rendered}}
```
smarnach: Nit: Please keep the newline at the end of file here. Many editors will add it back, resulting in spurious diffs. (Probably not so relevant for this particular file, but well.)

app/helpers/html-safe.js
```diff
@@ -0,0 +1,8 @@
+import { helper } from '@ember/component/helper';
+import { htmlSafe as markAsSafe } from '@ember/template';
+
+export function htmlSafe([content] /*, hash*/) {
```
smarnach: Can we use array destructuring on all browsers we support? Or do we compile the JS files somehow?

locks: `config/targets.js` tells ember-cli-babel which browsers we support so it knows which features to compile, and I think our listed browsers support array destructuring. I'd have to check UCAndroid.

app/templates/crate/owners.hbs
locks: Hm, good point. I tried to keep the exact same behaviour as I wasn't sure, but if we don't need it here then even better!

locks:
> Sorry for posting so many comments on such a simple PR.

No worries, that's what reviews are for 😁 it also helps me make sure I'm aligned.

app/templates/crate/owners.hbs
smarnach: Thinking about it, marking the email address as "safe" even appears to be actively harmful, and we don't do so in other places we show email addresses. This currently doesn't _actually_ introduce an XSS vulnerability since the only email address we show is the address of the current user, so people could only attack themselves, but I'd prefer if you remove `html-safe` here.

locks: Done!

smarnach:
Thanks for the updates, looks good to me now!

@bors r+

PR #1882, opened 2019-10-28
Use deploy keys for git operations

pietroalbini:
At the moment crates.io relies on Personal Access Tokens to be able to push changes to the index, but this is not great from a security point of view as it's not possible to scope a token to a single repository.

That means if an attacker managed to compromise the server hosting crates.io they'd be able to compromise other, unrelated parts of the Rust project infrastructure.

This PR changes crates.io to optionally support using deploy keys instead of tokens to push new commits to GitHub. Deploy keys are scoped to a single repository and can be either read-only or read-write (in our case we need write access).

The key must be stored in the `GIT_SSH_KEY` environment variable, base64-encoded. It has to be encoded as that way it's easier to copy-paste it, and the [setup-deploy-keys tool in rust-lang/simpleinfra](https://github.com/rust-lang/simpleinfra#setup-deploy-keys) generates them that way.

r? @carols10cents

### DIFF
--- a/src/bin/background-worker.rs
+++ b/src/bin/background-worker.rs
@@ -59,7 +59,8 @@ fn main() {
 
     println!("Cloning index");
 
-    let repository = Repository::open(&config.index_location).expect("Failed to clone index");
+    let repository =
+        Repository::open(&config.index_location, &credentials).expect("Failed to clone index");
 
     let environment = Environment::new(
         repository,
--- a/src/bin/background-worker.rs
+++ b/src/bin/background-worker.rs
@@ -12,7 +12,7 @@
 
 #![deny(warnings, clippy::all, rust_2018_idioms)]
 
-use cargo_registry::git::Repository;
+use cargo_registry::git::{Credentials, Repository};
 use cargo_registry::{background_jobs::*, db};
 use diesel::r2d2;
 use std::thread::sleep;
@@ -34,9 +34,17 @@ fn main() {
 
     let username = dotenv::var("GIT_HTTP_USER");
     let password = dotenv::var("GIT_HTTP_PWD");
-    let credentials = match (username, password) {
-        (Ok(u), Ok(p)) => Some((u, p)),
-        _ => None,
+    let ssh_key = dotenv::var("GIT_SSH_KEY");
+    let credentials = match (username, password, ssh_key) {
+        (Ok(username), Ok(password), Err(_)) => Credentials::Http { username, password },
+        (Err(_), Err(_), Ok(encoded_key)) => Credentials::Ssh {
+            key: String::from_utf8(
+                base64::decode(&encoded_key).expect("failed to base64 decode the ssh key"),
+            )
+            .expect("failed to convert the ssh key to a string"),
+        },
+        (Ok(_), Ok(_), Ok(_)) => panic!("can't configure both http and ssh authentication"),
+        _ => Credentials::Missing,
     };
 
     let job_start_timeout = dotenv::var("BACKGROUND_JOB_TIMEOUT")
--- a/src/bin/background-worker.rs
+++ b/src/bin/background-worker.rs
@@ -36,14 +36,19 @@ fn main() {
     let password = dotenv::var("GIT_HTTP_PWD");
     let ssh_key = dotenv::var("GIT_SSH_KEY");
     let credentials = match (username, password, ssh_key) {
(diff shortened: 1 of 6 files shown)


src/bin/background-worker.rs
```diff
@@ -34,9 +34,22 @@ fn main() {
...
     let password = dotenv::var("GIT_HTTP_PWD");
-    let credentials = match (username, password) {
-        (Ok(u), Ok(p)) => Some((u, p)),
-        _ => None,
+    let ssh_key = dotenv::var("GIT_SSH_KEY");
+    let credentials = match (username, password, ssh_key) {
+        (extra_user, extra_pass, Ok(encoded_key)) => {
+            if let (Ok(_), Ok(_)) = (extra_user, extra_pass) {
+                println!("warning: both http and ssh credentials to authenticate with git are set");
+                println!("note:note:  ssh credentials will take precedence over the http ones");
```
carols10cents: Did you mean to duplicate this "note:note:" here?

pietroalbini: Pressed the wrong keybinding probably :sweat_smile: Amended.

pietroalbini:
Ok, confirmed everything works and deployed the changes to staging.

Other than the two new commits, I had to make a few changes to the configuration:

* The `GIT_REPO_URL` environment variable has to be changed to `ssh://git@github.com/rust-lang/crates.io-index`, as of course using the HTTP protocol with SSH keys doesn't work (doh).
* `git2` doesn't support ed25519 SSH keys (at least in the version we have currently deployed). To work around this I generated an RSA key instead, which can be done by passing the `--rsa` flag to `setup-deploy-keys`.

carols10cents:
>     * The `GIT_REPO_URL` environment variable has to be changed to `ssh://git@github.com/rust-lang/crates.io-index`, as of course using the HTTP protocol with SSH keys doesn't work (doh).

Aha. That makes deploying a bit trickier then, since the SSH URL won't work with the HTTP username/password, so we can't update the environment variables and then push this code as crate publishes won't work while the code deploy is happening.

What do you think about having a new environment variable, `GIT_SSH_REPO_URL`, that goes with the new `GIT_SSH_KEY` repo?

Once I started thinking about that, it started to feel strange that the logic to decide what environment variables to use was in src/bin/background_worker.rs rather than in src/git.rs. 

I feel like I've already asked you to change a bunch, so I'm going to push up a commit rearranging the code along these lines to see what you think.

The good news is that publishing to staging worked!

carols10cents:
Ok, with the change I pushed that adds a new environment variable and moves all the (now rather complicated) logic determining which variables should get used, I'm able to set all the environment variables on staging and publish crates whether staging is running the code on master or the code in this PR.

So @pietroalbini, if you could review my last commit, then make a deploy key for the production index, let me know and I'll merge and deploy!

src/git.rs
```diff
@@ -80,17 +80,83 @@ pub struct Dependency {
...
 
 impl Repository {
-    pub fn open(url: &Url, credentials: &Credentials) -> CargoResult<Self> {
+    pub fn open(repository_config: &RepositoryConfig) -> CargoResult<Self> {
         let checkout_path = TempDir::new("git")?;
+
         let repository = git2::build::RepoBuilder::new()
-            .fetch_options(Self::fetch_options(credentials))
-            .clone(url.as_str(), checkout_path.path())?;
+            .fetch_options(Self::fetch_options(&repository_config.credentials))
```
pietroalbini: Now that the credentials are stored in the `Repository` struct we can make `fetch_options` a method and call it like:

```rust
.fetch_options(self.fetch_options())
```

pietroalbini:
By the way, other than the comment I left your changes looks good!

src/git.rs
carols10cents: Hm except in this spot, we don't have a `Repository` struct yet, we're in the process of creating one, and we can't create it without having the `git2::Repository`, which we can't create without the `fetch_options` here? Unless I'm misunderstanding?

pietroalbini: Oh, right, missed that.

pietroalbini:
Ok this looks great on my end. I think we can deploy it.

carols10cents:
@pietroalbini could you create a deploy key for the production index and add it to production heroku when you have a chance, since my simple-infra install wasn't working? I just added you as a collaborator in heroku so I think you have permissions.

Issue #1888, opened 2019-11-07
Cleanup: remove `email` field on `User` in Rust code


carols10cents:
This is for someone who likes fixing compiler errors and knows or wants to learn about diesel :)

As of [this PR](https://github.com/rust-lang/crates.io/pull/1045), I'm *pretty* sure we don't use the `email` field on `User` anywhere directly anymore (and if we aren't, we shouldn't be). Instead, we have an `emails` table that joins to `users`.

So we can clean this field up and be less confusing about where the user's email is.

The first step is removing all usage of this field from the Rust code, but leaving the column on the `users` table in the database. Once the Rust code is no longer using the column, we can safely migrate the database to drop that column.

But this issue is for just the first step. The way I'd go about it, which isn't the only way so feel free to go about this the way you'd like, is to remove [this line](https://github.com/rust-lang/crates.io/blob/0a6d3fb4aa75fbe504cb0c5a8ce1d853ddd244eb/src/models/user.rs#L16) from the `User` struct definition and try to compile.

A lot of the error messages are going to be because the default behavior of diesel is to select all columns of a table and expect the struct definition to match. Removing the email field from the struct will break that assumption. We need to change the Rust code to never select the email field from the database. We have this problem with krate too, which has a column having to do with full text search that we never want to select in the Rust code. 

The way the krate code works is it defines a type alias and a constant named `ALL_COLUMNS`: https://github.com/rust-lang/crates.io/blob/0a6d3fb4aa75fbe504cb0c5a8ce1d853ddd244eb/src/models/krate.rs#L50-L76

And then everywhere that we select a krate from the database, use that constant instead. For example, selecting all crates happens with the `all` function: https://github.com/rust-lang/crates.io/blob/0a6d3fb4aa75fbe504cb0c5a8ce1d853ddd244eb/src/models/krate.rs#L245-L247 which uses the `ALL_COLUMNS` constant and another type alias named `All`: https://github.com/rust-lang/crates.io/blob/0a6d3fb4aa75fbe504cb0c5a8ce1d853ddd244eb/src/models/krate.rs#L81

Other convenience functions like `Crate::by_name`: https://github.com/rust-lang/crates.io/blob/0a6d3fb4aa75fbe504cb0c5a8ce1d853ddd244eb/src/models/krate.rs#L237-L239 chain `filter` calls off `all` to avoid having to duplicate the `.select(ALL_COLUMNS)` bit.

[...]

PR #1891, opened 2019-11-12
Drop Email column (merge and deploy the fix for #1888 first!)

### LINKED ISSUE #1889: Cleanup: drop the `email` column on the `users` table
Can only be done after https://github.com/rust-lang/crates.io/issues/1888 is fixed. That issue is to remove all uses of `email` field on `User` from the Rust code.

What needs to be done:

- Generate a migration using the diesel CLI (search [this guide](https://diesel.rs/guides/getting-started/) for "diesel migration generate" for more instructions)
- In `up.sql`, drop the `email` column from the `users` table.
- In `down.sql`, leave a comment that this migration isn't really reversible since we're deleting data [like this migration has](https://github.com/rust-lang/crates.io/blob/master/migrations/20170728002039_fix_keywords/down.sql), but also restore the column with nulls.
- Run the migration and commit the schema changes along with the new migration.

hogum:
#### Fixes #1889

This cleans up the uses of the field `email` from `User` by:
- dropping the `email` field from `User`
- clearing references to the dropped email field
- migrating and updating the schema changes


hogum:
How should the `database visibility configuration` be updated after modifying the schema?
Its test fails with:
```
---- tasks::dump_db::gen_scripts::tests::check_visibility_config stdout ----
thread 'tasks::dump_db::gen_scripts::tests::check_visibility_config' panicked at 'The visibility configuration does not match the database schema:
Column users.email does not exist in the database.', src/tasks/dump_db/gen_scripts.rs:204:9
```

carols10cents:
> How should the `test database visibility configuration` be updated after modifying the schema?
> It fails with:
> 
> ```
> ---- tasks::dump_db::gen_scripts::tests::check_visibility_config stdout ----
> thread 'tasks::dump_db::gen_scripts::tests::check_visibility_config' panicked at 'The visibility configuration does not match the database schema:
> Column users.email does not exist in the database.', src/tasks/dump_db/gen_scripts.rs:204:9
> ```

Removing [the users.email line from the dump-db.toml file](https://github.com/rust-lang/crates.io/blob/23f187ac703cd2e4e36621729f550f4c6f8be9de/src/tasks/dump_db/dump-db.toml#L168) should do it. The dump-db.toml file is configuration for the public database dumps we make, and this test ensures the database schema is always in sync with this configuration so that we don't accidentally leak data. 

I'll make the error message for that test a bit clearer to point at what needs to be done.

hogum:
Thanks @carols10cents! That sorted it out.
Dumping to the emails table originally required the `user: email` column - what would be the reason for getting a unique key violation on `emails: user_id` after dropping the `users: email` column. 

https://github.com/rust-lang/crates.io/blob/23f187ac703cd2e4e36621729f550f4c6f8be9de/migrations/20170804200817_add_email_table/up.sql#L16-L17

carols10cents:
> Dumping to the emails table originally required the `user: email` column - what would be the reason for getting a unique key violation on `emails: user_id` after dropping the `users: email` column.

I'm confused, are you getting a unique key violation? Can you give me steps to reproduce?

carols10cents:
Oh I see the errors in travis now. Investigating!

migrations/2019-11-11-162609_drop_email_from_user/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
```
carols10cents: You can remove this comment :)

migrations/2019-11-11-162609_drop_email_from_user/up.sql
```diff
@@ -0,0 +1,2 @@
...
+ALTER TABLE  users DROP COLUMN email CASCADE;
```
carols10cents: This is super picky, but can you remove the extra space between `TABLE` and `users`?

src/controllers/user/me.rs
```diff
@@ -38,10 +38,10 @@ pub fn me(req: &mut dyn Request) -> CargoResult<Response> {
 
     let verified = verified.unwrap_or(false);
     let verification_sent = verified || verification_sent;
-    let user = User { email, ..user };
+    let user = User { ..user };
```
carols10cents: I think this line can be completely removed now, actually! I think `user = User { ..user }` is doing a no-op.

src/models/user.rs
```diff
@@ -179,19 +185,21 @@ impl User {
     /// Converts this `User` model into an `EncodablePrivateUser` for JSON serialization.
     pub fn encodable_private(
         self,
+        email: Option<String>,
+
         email_verified: bool,
         email_verification_sent: bool,
-    ) -> EncodablePrivateUser {
+    ) -> CargoResult<EncodablePrivateUser> {
```
carols10cents: Can you explain why this function needed to be changed to return a `Result`? It doesn't look like it's necessary to me because I don't see anywhere returning an `Err` variant. Am I missing something?

src/models/user.rs
```diff
@@ -13,20 +13,19 @@ use crate::views::{EncodablePrivateUser, EncodablePublicUser};
...
     pub id: i32,
-    pub email: Option<String>,
     pub gh_access_token: String,
     pub gh_login: String,
     pub name: Option<String>,
     pub gh_avatar: Option<String>,
     pub gh_id: i32,
 }
 
+/// Represents a new user record insertible to the `users` table
```
carols10cents: Small fix: can you change `insertible` to `insertable` in this comment please?

src/tests/util.rs
```diff
@@ -135,9 +135,10 @@ impl TestApp {
         use diesel::prelude::*;
 
         let user = self.db(|conn| {
-            let mut user = crate::new_user(username).create_or_update(conn).unwrap();
-            let email = "something@example.com";
-            user.email = Some(email.to_string());
+            let email = "cow@mammals.milk";
+            let user = crate::new_user(username)
+                .create_or_update(Some(email), conn)
```
carols10cents: Aha, so the unique constraint that's causing the tests to fail is that here in the test setup you're creating a new user with an email set, then [lines 142-149 right after this](https://github.com/rust-lang/crates.io/pull/1891/files#diff-2bda8a5d16907d0b0ef38311d9dccb43R142-R149) the test setup is inserting the email into the table directly, which violates [the constraint we have that each user may only have one email](https://github.com/rust-lang/crates.io/blob/21e04f3028624485621d577359008560728a37d3/migrations/20170804200817_add_email_table/up.sql#L7).

To fix this, you can pass `None` to `create_or_update` here and the next part will take care of inserting the email (and setting it to verified, which is important for the tests).

src/models/user.rs
hogum: It isn't necessary. I was previously reading this `email` variable from the database with an `Err`

https://github.com/rust-lang/crates.io/blob/3544b91086e822607cfeadda57b261e66112ba52/src/models/user.rs#L86

variant result before realizing [that was already "done"](https://github.com/rust-lang/crates.io/blob/3544b91086e822607cfeadda57b261e66112ba52/src/controllers/user/me.rs#L28-L37) and the email could simply be passed to the `src/models/users/encodable_private` function as an argument.
I definitely forgot to undo that modification to the return type - will do so now

migrations/2019-11-11-162609_drop_email_from_user/up.sql
hogum: :)

src/tests/user.rs
```diff
@@ -510,12 +514,13 @@ fn test_confirm_user_email() {
 
     // Simulate logging in via GitHub. Don't use app.db_new_user because it inserts a verified
     // email directly into the database and we want to test the verification flow here.
+    let email = "cow@mammals@milk";
```
carols10cents: I'm curious, why is the value different here? Any particular reason, and any reason why it's not a valid email address (two `@`s)?

hogum:
Resolved. The invalid email address was a mistake I hadn't noted. Is it not supposed to be caught? I'm surprised it passes the test

hogum:
Resolved. The invalid email address was a mistake I hadn't noted. Is it not supposed to be caught? I'm surprised it passes the test

carols10cents:
Thank you so much!! This looks great, but I'm not going to merge it yet-- we need to merge and deploy https://github.com/rust-lang/crates.io/issues/1888 to remove all the *uses* of the email column first, so that there aren't any connections using the `email` field when we deploy this PR.

carols10cents:
I removed some more of the temporary code that was needed to deploy the code that stopped using the email column (and that has been deployed). I also introduced more merge conflicts by merging another PR just now; I'll be fixing those in a few hours.

carols10cents:
I think this is ready to go now!! Thank you!!!

@bors r+

Issue #1791, opened 2019-07-29
Can we have a "mathematics" category slug?


ornamentist:
Crates.io category slugs includes a very broad "science" slug, covering a huge range of fields and disciplines.

Could we have a "mathematics" slug for crates that are specifically mathematically oriented (like my crate https://crates.io/crates/un_algebra)? 

I understand this might lead to requests for, for example, "physics" or "chemistry" slugs, although "mathematics", "physics", "chemistry" etc are still very broad terms.

carols10cents:
Hi @ornamentist ! Sorry this has taken so long, but @mindriot101 has opened https://github.com/rust-lang/crates.io/pull/1893 for this. They'd like some feedback/suggestions on what the description for the category should be-- do you have any thoughts? Thanks!

carols10cents:
Fixed by https://github.com/rust-lang/crates.io/pull/1893

PR #1893, opened 2019-11-13
Add mathematics category

mindriot101:
This change adds the additional mathematics category, as well as removing "math" from the list of  fields covered by the "science"
category.

I think having a separate category makes a lot of sense, as mathematics related items can be related to e.g. simulations or games, which might be worth extracting it out from "science".

I would however like some input on the description. As a non-mathematician I'm not too aware of the range of mathematics crates that are available.

#1791

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -272,6 +272,12 @@ Crates to help adapting internationalized software to specific \
 languages and regions.\
 """
 
+[mathematics]
+name = "Mathematics"
+description = """
+Crates related to mathematics or related topics.
+"""
+
 [memory-management]
 name = "Memory management"
 description = """
@@ -398,7 +404,7 @@ Rust.\
 [science]
 name = "Science"
 description = """
-Crates related to solving problems involving math, physics, chemistry, \
+Crates related to solving problems involving physics, chemistry, \
 biology, machine learning, geoscience, and other scientific fields.\
 """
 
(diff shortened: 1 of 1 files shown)


carols10cents:
Thank you for this! I'm not sure what it should be either, let's see if the requester of this category has some ideas!

ornamentist:
> Thank you for this! I'm not sure what it should be either, let's see if the requester of this category has some ideas!

I should point out I'm not a mathematician either, so take what I say with a grain of salt.

There seems to be multiple ways to look at mathematics as a discipline, e.g. "pure" .vs. "applied", "symbolic" .vs. "numeric", or the ontologies used by mathematics journals and encyclopedias. This situation reflects the amazing depth, breadth and history of the discipline.

My _un-algebra_ crate (https://crates.io/crates/un_algebra) is concerned with abstract algebra and is more on the "pure" end of the spectrum. The very useful _num_ crates have both pure and applied aspects. The various optimized linear algebra crates could equally well have "gaming", "GIS" or "graphics" tags as they're based on mathematics but are optimized for those kinds of projects.

I'd like to understand more how other systems-programming or general-purpose languages handle library classification for complex domains like "mathematics" or "science" and see what lessons we can learn.

All of which is saying that I believe we need at least top-level tags for "mathematics", "science", etc but that this is the start of a longer process of ensuring the tag system can grow with Rust's expansion into more complex application domains.

carols10cents:
Ok, I'm fine going with "Crates with a mathematical aspect". Could you make that change please @mindriot101 ?

mindriot101:
@carols10cents done

carols10cents:
Great, thank you!! 

@bors r+

PR #1894, opened 2019-11-13
Update swirl

sgrif:
The latest revisions to swirl include:

- Concrete error types everywhere instead of `Box<dyn Error>`, which
  simplifies conversion for us.
- Automatic handling of connection pool configuration
  - Swirl can now just take a DB url, and defaults to a connection pool
    with 2x the thread count connections, which is what we want
  - Jobs can take a connection as an argument, meaning we don't need to
    carry around a pool manually in the environment
- Updated dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -219,9 +219,9 @@ dependencies = [
  "reqwest 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "scheduled-thread-pool 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "semver 0.9.0 (git+https://github.com/steveklabnik/semver.git)",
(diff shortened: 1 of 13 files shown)


sgrif:
Note: This is still pointing at a local path instead of git, https://github.com/sgrif/swirl/pull/20 and https://github.com/sgrif/swirl/pull/21 need review so they can be merged at which point I will point this at the merge commit on master and take it out of draft mode

sgrif:
(There's no reason this PR can't be reviewed, it just can't be merged)

jtgeibel:
I haven't looked at any of the upstream changes, but as a consumer of the API this PR looks good so far.

carols10cents:
Closing in favor of #2269

PR #1897, opened 2019-11-14
Add Cargo.toml copy button to crate rows

### LINKED ISSUE #1587: Copyable version on search result page
The usual workflow to get a copyable Cargo.toml entry for a given crate one is required to go through four steps:  
- go on crates.io
- search for a crate
- click on search result entry
- copy Cargo.toml entry  via "Copy to clipboard" button

Oftentimes I know what crate I want to use but don't know the exact name of the crate.  
I really would like to have the "Copy to clipboard" button already on the search result page.

There already is a version badge for each crate on the search result page from what I know so I guess the necessary information should already be available.

DSpeckhals:
To copy the most recent version of a crate to the clipboard, users would have to go to the crate's individual page. However, it would probably be convenient to have the same control on the search results page. To accomplish this, a new component was created to encapsulate the copy control and its notifications (success/failure). This new component was added to the search result page; the crate page was reworked to use it as well.

For this first run at the UI, I added the same copy button to the left of the name/link for the result. I'm open to any other ideas too!

Here are some screenshots to illustrate the changes:

### Crate Page

Screenshot from 2019-11-14 12-33-09

Screenshot from 2019-11-14 12-32-03

### Results Page
Screenshot from 2019-11-14 14-24-57

Fixes #1587

### DIFF
--- a/app/components/crate-row.js
+++ b/app/components/crate-row.js
@@ -1,7 +1,11 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
 
 export default Component.extend({
   classNames: ['crate', 'row'],
+  crateTomlText: computed('crate.name', 'max_version', function() {
+    return `${this.get('crate.name')} = "${this.get('crate.max_version')}"`;
+  }),
 
   'data-test-crate-row': true,
 });
--- a/app/components/crate-toml-copy.js
+++ b/app/components/crate-toml-copy.js
@@ -1,0 +1,32 @@
+import Component from '@ember/component';
+import { later } from '@ember/runloop';
+
+export default Component.extend({
+  classNames: ['crate-toml-copy'],
+  copyText: '',
+  showSuccess: false,
+  showNotification: false,
+  toggleClipboardProps(isSuccess) {
+    this.setProperties({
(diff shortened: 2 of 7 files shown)


DSpeckhals:
Thank you for the input! Now that I see the copy button to the left of the crate name, I think I like that better. As for the location of the "Copied!" text, I know the problem and I'll correct it.

DSpeckhals:
What do you think of these changes? I made the icon a little smaller to better associate it with the crate name; another option would be to make it closer to the height of the badges. I personally like it smaller like the text, but it's not a strong opinion.

image

carols10cents:
I like it, thank you!! The badges not being the same height as the text is definitely a pre-existing condition 😥 

The "Copied!" text on exact match search results is now overlapping the badges a bit though, not sure if that was supposed to be fixed with this latest commit or not:

Screen Shot 2019-11-18 at 1 19 16 PM

DSpeckhals:
I fixed the notification positioning. It was incorrect in exact matches (like your screenshot).

carols10cents:
Looks great!!! Thank you so much!!

@bors r+

Issue #169, opened 2015-07-18
Enforce that external URIs in crate metadata must be absolute


nagisa:
[core-utils](https://crates.io/crates/core-utils) crate didn’t specify URI scheme when defining its repository link, therefore it is interpreted as `https://crates.io/github.com/0X1A/core-utils`. This can lead to various issues by crafting malicious URIs to achieve unintended actions on crates.io system (trivial example would be following a package).

carols10cents:
What could we normalize the URIs to, though? I think publishing should reject non-absolute URIs instead...

kzys:
+1 for rejecting non-absolute URIs.

carols10cents:
Instructions for implementing this:

- In the [`parse_new_headers`](https://github.com/rust-lang/crates.io/blob/a898c389f284af995c2457b504a714ca5f24e32b/src/controllers/krate/publish.rs#L217) function that interprets the crate metadata sent with a publish request, check the [`EncodableCrateUpload`'s](https://github.com/rust-lang/crates.io/blob/a898c389f284af995c2457b504a714ca5f24e32b/src/controllers/krate/publish.rs#L227) [`homepage`](https://github.com/rust-lang/crates.io/blob/a898c389f284af995c2457b504a714ca5f24e32b/src/views/krate_publish.rs#L23), [`documentation`](https://github.com/rust-lang/crates.io/blob/a898c389f284af995c2457b504a714ca5f24e32b/src/views/krate_publish.rs#L24), and [`repository`](https://github.com/rust-lang/crates.io/blob/a898c389f284af995c2457b504a714ca5f24e32b/src/views/krate_publish.rs#L33) fields for valid, absolute URLs by parsing them with the `url` crate's [`parse` function](https://docs.rs/url/2.1.0/url/struct.Url.html#method.parse).
- I think it would be nice to report all invalid URLs at once, so rather than saying "your repository url is invalid", then they fix that and get "your documentation url is invalid" etc, gather the fields that have invalid URLs and return one error that says "the following fields have invalid URLs: repository, documentation, homepage". This'll look similar to the required field checks here: https://github.com/rust-lang/crates.io/blob/a898c389f284af995c2457b504a714ca5f24e32b/src/controllers/krate/publish.rs#L234-L252
[...]

PR #1898, opened 2019-11-15
Check that URLs begin with `http://` or `https://`

jtgeibel:
This switches to a manual check of the string prefix, because
`Url::parse` may sanitize URLs as they are parsed, making it
difficult to ensure both slashes are present.

Fixes: #169


smarnach:
This looks good to me.

@bors r+

PR #1899, opened 2019-11-15
Bump r2d2

jtgeibel:
`cargo update -p r2d2`

Refs: #1805

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -554,7 +554,7 @@ dependencies = [
  "chrono 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "diesel_derives 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "pq-sys 0.4.4 (registry+https://github.com/rust-lang/crates.io-index)",
- "r2d2 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


jtgeibel:
We should do the deploy, merge, deploy dance on this one, just to be safe.

smarnach:
This pulls in two new versions of `parking_lot` (in addition to the two versions we've already been using), so there is definitely some potential for things going wrong there.

I can't deploy, so I'm randomly selecting
r? @carols10cents 
as new assignee.

jtgeibel:
> This pulls in two new versions of parking_lot (in addition to the two versions we've already been using)

Yeah, this is unfortunate.  (Although the duplication isn't quite that bad, as one pair is `parking_lot` and the other `parking_lot_core`.)  Long term, I'm hoping to get this under control, but its difficult because we effectively can't run `cargo update` at this time.

In terms of build time, it especially bad that we pull in 3 or 4 versions of `syn` and `quote`.  Fortunately, these are both 1.0 now so as we update other things those should consolidate.

carols10cents:
I'm good with this, but I don't want to do this at the same time that we're switching to cloudfront, so I'll do the deploy-merge-deploy dance on Wednesday.

carols10cents:
Dance time!!!

@bors r+

Issue #1895, opened 2019-11-14
Email all individual owners when a new version is published


carols10cents:
Extracted from #815 

Now that we've been requiring verified email addresses for publishing, it's more likely this feature will work well.

**This is a big task that's more complicated than it looks!** Please ask questions if anything is confusing! This task would be great for someone looking to get experience with just about every part of crates.io: the frontend, the backend API, the database, email sending.

What should happen:

- When any owner of a crate (either by being an individual owner or having permission through a team) publishes a new version of a crate they own,
- Email all individual owners of that crate that have verified email addresses to notify them of the new version
  - We could email all team owners as well, but we'd have to query GitHub for that each time and try to match up team members to crates.io accounts, seems complex. Let's start with individual owners.
  - **We could email the person who published the crate as well if they have permissions because of a team, so that they're notified more quickly if they didn't authorize that publish with their credentials. However, there'd be no way to turn this email off because we don't have a record on which to store that preference? I'm torn about whether we say "the person who did the publish always gets an email even if you've turned notifications off" or whether we say "people who have permissions via teams never get emails even if they're the one who published the crate". Open to arguments.**
  - The email should include the name of the crate, the version published, who published it, the date and time they published it, instructions on what to do if the publish wasn't authorized, and a link to the page to opt out of the emails. [Here's an example of an email from rubygems.org](https://github.com/rubygems/rubygems.org/issues/2101#issuecomment-526008525):
<img width="1264" alt="rubygems.org email screenshot" src="https://user-images.githubusercontent.com/193874/68871076-e34eb200-06c9-11ea-9966-61467847a3ea.png">

- Let each owner of a crate opt out of getting these notifications for each crate they're an individual owner of. Rubygems.org has this preference page:

<img width="898" alt="rubygems.org preferences page screenshot" src="https://user-images.githubusercontent.com/193874/68870811-7cc99400-06c9-11ea-8a17-6beff0806388.png">

## Implementation instructions

Roughly trying to separate these into pieces that can be submitted and reviewed as separate PRs; let me know if you run into any problems with the way I've split the tasks up.

- [x] Backend: email preference storing (addressed in https://github.com/rust-lang/crates.io/pull/1901)
[...]

PR #1901, opened 2019-11-16
backend: Add user email preference storing

DSpeckhals:
This PR addresses the first step in implementing #1895: "Backend: email preference storing."

This commit is the first step in the process to send crate owners an email notification when a new version of one of their crates is published. A database migration adds a `email_notifications` column to the `crate_owners` table, and thus the property was added to the corresponding `CrateOwner` struct. This new property is defaulted to `true`.

Because a user may not want to receive a version publish notification for all of their crates, an API endpoint was added to allow them to toggle email notifications for each crate. The front end implementation will be in a forthcoming commit, as well as the actual sending of these notifications.


src/controllers/user/me.rs
```diff
@@ -36,12 +36,27 @@ pub fn me(req: &mut dyn Request) -> CargoResult<Response> {
         ))
         .first::<(User, Option<bool>, Option<String>, bool)>(&*conn)?;
 
+    let owned_crates = crate_owners::table
+        .inner_join(crates::table)
+        .filter(crate_owners::owner_id.eq(user_id))
+        .select((crates::id, crates::name, crate_owners::email_notifications))
+        .order(crates::name.asc())
```
carols10cents: Good call on the ordering!!

src/views.rs
```diff
@@ -149,9 +149,17 @@ pub struct EncodableApiTokenWithToken {
     pub last_used_at: Option<NaiveDateTime>,
 }
 
+#[derive(Deserialize, Serialize, Debug)]
+pub struct OwnedCrate {
```
carols10cents: I just realized I said "tuple" in the implementation instructions and you correctly inferred that this info should be in a struct instead like everything else 😅 Great work figuring out what I meant instead of what I said!

src/controllers/user/me.rs
```diff
@@ -36,12 +36,27 @@ pub fn me(req: &mut dyn Request) -> CargoResult<Response> {
         ))
         .first::<(User, Option<bool>, Option<String>, bool)>(&*conn)?;
 
+    let owned_crates = crate_owners::table
+        .inner_join(crates::table)
+        .filter(crate_owners::owner_id.eq(user_id))
```
carols10cents: I forgot to mention this in the instructions and the code doesn't make this very clear, but this query needs to also have a filter for the `crate_owners::owner_kind` being `OwnerKind::User` because records for teams owning crates are also stored in this table. Here's an example of a similar query: https://github.com/rust-lang/crates.io/blob/ea248e161b22dd0e7f132825333d0e4383d301c1/src/controllers/user/other.rs#L39

I'm starting to think we should split team and user ownership records into separate tables, but that's a totally separate issue 😅

src/controllers/user/me.rs
```diff
@@ -212,3 +227,38 @@ pub fn regenerate_token_and_send(req: &mut dyn Request) -> CargoResult<Response>
...
+    let mut body = String::new();
+    req.body().read_to_string(&mut body)?;
+    let updates: Vec<CrateEmailNotifications> =
+        serde_json::from_str(&body).map_err(|_| human("invalid json request"))?;
+
+    let user = req.user()?;
+    let conn = req.db_conn()?;
+
+    for to_update in updates {
+        update(CrateOwner::belonging_to(user))
```
carols10cents: So this query will need a filter on `owner_kind` too, like the query above. Sorry again for forgetting to include this in the instructions!

I also realize I didn't communicate clearly about what I was thinking for the updates here-- I suggested updating the records unconditionally (rather than checking to see if they needed to be updated as you're doing here with the `filter(email_notifications.ne(to_update.email_notifications))`) so that instead of doing one update statement per crate owned, we can do one `INSERT... ON CONFLICT DO UPDATE` (aka "upsert") statement including all the crate ownership records for this user. There are [some people who own a lot of crates](https://crates.io/users/alexcrichton), so while in most cases doing an UPDATE per record would be fine, the edge cases worry me!

[Here's an example of doing an upsert](https://github.com/rust-lang/crates.io/blob/2efe33df7183c2822e8b1b53e1090a510b6b6fce/src/boot/categories.rs#L114-L123) with multiple records like needs to be done here, are you up for making that change? Please let me know if more details would be helpful!

src/tests/user.rs
```diff
@@ -569,3 +590,121 @@ fn test_existing_user_email() {
...
+            crate_owners::table
+                .select(crate_owners::email_notifications)
+                .filter(crate_owners::crate_id.eq(not_my_crate.id))
+                .first::<bool>(&*conn)
+        })
+        .unwrap();
+
+    // There should be no change to the `email_notifications` value for a crate not belonging to me
+    assert!(email_notifications);
+}
```
carols10cents: Great tests!!!

src/controllers/user/me.rs
DSpeckhals: I pushed a commit to address this one. The one downside that I think the `updated_at` column would be updated regardless of whether it was changed or not.

src/controllers/user/me.rs
```diff
@@ -36,12 +38,28 @@ pub fn me(req: &mut dyn Request) -> CargoResult<Response> {
         ))
         .first::<(User, Option<bool>, Option<String>, bool)>(&*conn)?;
 
+    let owned_crates = crate_owners::table
+        .inner_join(crates::table)
+        .filter(crate_owners::owner_id.eq(user_id))
+        .filter(crate_owners::owner_kind.eq(OwnerKind::User as i32))
```
carols10cents: Writing up a response to your other comment made me realize that in this spot we also need a filter here that `crate_owners::deleted.eq(false)`. And since I missed remembering 2 filters now, I'm thinking we should have a function for this so that the logic is in one spot that we're sure is correct. Looking around, I see a few other places where we're selecting "crates currently owned by a user, not a team":

- https://github.com/rust-lang/crates.io/blob/36cfbe31b3a81d3a80782299673e3925edde4758/src/controllers/krate/search.rs#L121-L125
- Sort of here: https://github.com/rust-lang/crates.io/blob/36cfbe31b3a81d3a80782299673e3925edde4758/src/models/krate.rs#L397-L404
- https://github.com/rust-lang/crates.io/blob/36cfbe31b3a81d3a80782299673e3925edde4758/src/models/user.rs#L132-L139

Are you up for a little refactoring? Just a little bit of scope creep? 😇

src/controllers/user/me.rs
carols10cents: > The one downside that I think the `updated_at` column would be updated regardless of whether it was changed or not.

I don't think that's the case -- updating the `updated_at` column [is implemented as a `BEFORE UPDATE` per-row trigger](https://github.com/rust-lang/crates.io/blob/36cfbe31b3a81d3a80782299673e3925edde4758/migrations/20151126095136_dumped_migration_103/up.sql#L47-L49), and if I'm reading [the postgres docs](https://www.postgresql.org/docs/12/trigger-definition.html) correctly, these only apply to *affected* rows:

> With a per-row trigger, the trigger function is invoked once for each row that is affected by the statement that fired the trigger.

So inserting with a conflict but the same values that are already there wouldn't affect that row. I did some testing locally and the behavior I saw was consistent with my understanding of the docs.

HOWEVER, this does raise the issue of whether we want the `updated_at` column to be changed when someone DOES update this preference... and now I'm conflicted (pun not intended)!

- The updated_at column on crate_owners is currently public information [at least through the database dumps](https://github.com/rust-lang/crates.io/blob/36cfbe31b3a81d3a80782299673e3925edde4758/src/tasks/dump_db/dump-db.toml#L69) when [the record isn't `deleted`](https://github.com/rust-lang/crates.io/blob/36cfbe31b3a81d3a80782299673e3925edde4758/src/tasks/dump_db/dump-db.toml#L62). There's a possibility one could figure out if you updated your email notification settings if we changed the updated_at. While probably unlikely and probably not valuable, I'd like to not leak info if we can help it.
- The main purpose of that `updated_at` column is to help in auditing when we're trying to figure out why something unexpected happened with who owns a crate. This is fairly unhelpful, though, as we aren't recording *who* updated the record. [We're working on adding more auditing](https://github.com/rust-lang/crates.io/issues/1548) but that's not done yet. Once that auditing is implemented, I'm not sure that we care about this column at all, but I need to figure that out some more.
- Maybe this is another sign that this table is the wrong place for storing this preference...... but I'm not sure where the right place would be. A new table? Just for the "off" settings because we're defaulting to "on"? Do you have any thoughts? I'm going to bounce some ideas around with some folks and I'll update here with what I'm thinking in the next few days.

src/controllers/user/me.rs
DSpeckhals: I can pull that out into another function...no problem!

src/controllers/user/me.rs
DSpeckhals: The ON CONFLICT logic is helpful to know. Thanks for researching that.

I don't personally like the idea of a "don't send me notifications for this crate" table. How it is seems okay, except for the new column not making sense for teams. I'll be curious to know what the general consensus is.

If we stick with the status quo for this PR, would it make sense to remove the updated_at column from the DB dump?

src/controllers/user/me.rs
DSpeckhals: I just pushed a commit that refactors as mentioned. I hope it's what you're looking for!

DSpeckhals: BTW, I also made a little change to the way the "upsert" works. When building the inserts, I filtered out the records that won't change. Perhaps this is unnecessary based on what you've stated above, but it seems nice to only try to update the records that need updating. I'm fine with whatever, though.

src/controllers/user/me.rs
carols10cents: > I don't personally like the idea of a "don't send me notifications for this crate" table. How it is seems okay, except for the new column not making sense for teams. I'll be curious to know what the general consensus is.

I chatted with @jtgeibel at a meetup last night, and he didn't like that this column didn't make sense for teams. But that's kind of an unrelated problem-- the solution is to split this table up into teams only and users only, and that doesn't have anything to do with this change.

I think removing the `updated_at` column from the database dump makes sense, and we'll get to see if anyone has complaints about us removing columns! (We have said the database dumps are experimental, so they can't be too mad). Let's go with that. Thanks!

src/models/owner.rs
```diff
@@ -22,6 +23,21 @@ pub struct CrateOwner {
...
+
+impl CrateOwner {
+    /// Returns a base crate owner query filtered by the owner kind argument. This query also
+    /// filters out deleted records.
+    pub fn by_owner_kind(kind: OwnerKind) -> BoxedQuery<'static> {
+        use self::crate_owners::dsl::*;
+
+        crate_owners
+            .filter(deleted.eq(false))
+            .filter(owner_kind.eq(kind as i32))
```
carols10cents: Oh this is so much nicer!! I love that this function encapsulates the casting of `OwnerKind` too :)

src/controllers/user/me.rs
carols10cents: > Perhaps this is unnecessary based on what you've stated above, but it seems nice to only try to update the records that need updating. I'm fine with whatever, though.

Yeah I think it's probably unnecessary, but it's also fine :)

src/controllers/user/me.rs
DSpeckhals: I made `updated_at` private in the latest commit.

src/controllers/user/me.rs
DSpeckhals: I removed it for sake of simplicity in the latest commit.

carols10cents:
Looks great!!! Thank you so much!!

@bors r+

PR #1902, opened 2019-11-16
Improve build caching on CI

jtgeibel:
These changes started based on a discussion over on [internals](https://internals.rust-lang.org/t/evaluating-github-actions/11292/8?u=jtgeibel).  Basically, on every single build we end up touching some files and invaliding the caches.  This forces Travis to compress and upload a new cache on every single build.

This PR includes several changes on CI:

* Incremental compilation is disabled.  It appears that incremental files can change even if nothing in the codebase changes.
* Drop the `target/.rustc_info.json` file before caching, as it can change on each build.
* Running `cargo install` for the diesel CLI acts similar to a `cargo update` causing changes to the index and some caches maintained under `$HOME/.cargo`.  `cargo install` is now only run if the diesel binary is not on the path.  The downside to this is that if we bump the `.diesel_version` file then Travis will not be using the exact same version as Heroku until we clear the CI cache.  Since we only use this binary to run migrations and we haven't bumped versions since January, I think this change is low risk compared to the improvement.
* Add `RUSTFLAGS="-C debuginfo=0"` to disable debug info, which helps shrink the cache size.

Based on my observations while putting together this PR, it looks like disabling incremental compilation has an impact on compile times, but this is small relative to the overall win on caching behavior.  Additionally, our cache size has been reduced from 3948.04MB to 2439.19MB.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
     - DATABASE_URL=postgres://postgres:@localhost/cargo_registry_test
     - TEST_DATABASE_URL=postgres://postgres:@localhost/cargo_registry_test
     - CARGO_TARGET_DIR=target
+    - CARGO_INCREMENTAL=0
     - PERCY_PARALLEL_TOTAL=2
     # Percy secrets are included here to enable Percy's GitHub integration
     # on community-submitted PRs
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ env:
     - JOBS=1 # See https://git.io/vdao3 for details.
     - DATABASE_URL=postgres://postgres:@localhost/cargo_registry_test
     - TEST_DATABASE_URL=postgres://postgres:@localhost/cargo_registry_test
-    - CARGO_TARGET_DIR=target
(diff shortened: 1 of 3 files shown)


jtgeibel:
@bors: r+

Merging because the cache sizes on `master` and `auto` have grown to 9+ gigs each and this should resolve that.

carols10cents:
Now I'm seeing 2157.20MB cache for master, and I don't see auto at all, so... yay? Not sure if auto being gone is expected (I guess we haven't run r+ since this was merged?)

Screen Shot of travis caches as I see them now

carols10cents:
Ok now that I've r+ed something I see auto at 560.20MB.

jtgeibel:
Yeah, I ended up manually clearing the caches for auto and master.  The PR had shrunk their size, but they were still around 7G each for some reason.  My best guess is that something changed on Travis that caused the hash in cache filenames to change and that those branches still had old cache files that would never be used again.

In any case, hopefully our cache size will now hover aroud 2.1GB per branch/PR, instead of 4GB.

Issue #1896, opened 2019-11-14
Consider adding a configuration option for maximum metadata length


carols10cents:
Split off from https://github.com/rust-lang/crates.io/issues/1522. Relevant discussion:

Justin:

 * We should add a configuration entry for the global max size of the metadata (we currently use max tarball size several places)

Carol:

Do we really need a separate setting for this? Isn't max content length an effective maximum on metadata, because you could theoretically have a tarball that's 0 bytes and metadata that takes up the rest of the space? Just thinking in terms of what we would set this configuration option to if we had it!

I suppose that gets into how we want to resolve [this comment](https://github.com/rust-lang/crates.io/blob/23f187ac703cd2e4e36621729f550f4c6f8be9de/src/controllers/krate/publish.rs#L107-L109) and how we want to communicate this limit to someone whose crate is getting rejected who has a `.crate` file below our stated limits.

Justin:

I think it makes sense to add a separate config item here, but I don't know what we should set it to.  Maybe we should add some logging to get a feel for some typical metadata sizes.  The publish endpoint is fairly attractive from a DOS perspective, and the metadata could kick off a lot of database activity.  I could potentially see a scenario where it would be nice to drop this limit quickly via an environment variable.  Although I haven't really put much serious though into what such an attack might look like and if this would be an effective mitigation.

smokytheangel0:
I can see in src/controllers/krate/publish.rs, where we could get just the metadata size of the upload by getting the difference between content_length and file_length, but I am wondering what maximums you had in mind?

carols10cents:
Hi @smokytheangel0 ! There is not another issue for logging. Good catch, I mostly was concentrating on extracting the relevant info out of the other issue and I didn't really read the text I was copypasting very closely 😅 

The logging is implemented here: https://github.com/rust-lang/crates.io/blob/2efe33df7183c2822e8b1b53e1090a510b6b6fce/src/middleware/log_request.rs and when you run `cargo run --bin server` and make requests locally, you'll see lots of stuff printed for local dev. The relevant lines that we log in production are the ones that start with `at=info`.

[...]

PR #1905, opened 2019-11-18
conditionally add request metadata length to log output (no test)

smokytheangel0:
This pull request implements the first step in determining a maximum metadata size for uploads as described in issue 1896 [https://github.com/rust-lang/crates.io/issues/1896](url). There was no test added for this change as log output is not automatically tested.

### DIFF
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -31,10 +31,16 @@ impl Handler for LogRequests {
         let response_time =
             response_time.as_secs() * 1000 + u64::from(response_time.subsec_nanos()) / 1_000_000;
 
+        let metadata_length = req
+            .extensions()
+            .find::<u64>()
+            .map_or(String::new(), |l| format!(" metadata_length={}", l));
+
         print!(
             "at={level} method={method} path=\"{path}\" \
              request_id={request_id} fwd=\"{ip}\" service={time_ms}ms \
-             status={status} user_agent=\"{user_agent}\"",
+             status={status} user_agent=\"{user_agent}\"\
+             {metadata_length}",
             level = level,
             method = req.method(),
             path = FullPath(req),
@@ -43,6 +49,7 @@ impl Handler for LogRequests {
             user_agent = request_header(req, "User-Agent"),
             request_id = request_header(req, "X-Request-Id"),
             status = response_code,
+            metadata_length = metadata_length,
         );
 
         if let Err(ref e) = res {
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -217,6 +217,8 @@ pub fn publish(req: &mut dyn Request) -> CargoResult<Response> {
 fn parse_new_headers(req: &mut dyn Request) -> CargoResult<(EncodableCrateUpload, User)> {
     // Read the json upload request
     let metadata_length = u64::from(read_le_u32(req.body())?);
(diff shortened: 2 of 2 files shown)


smokytheangel0:
waiting on a full local build to manually test the addition.

src/middleware/log_request.rs
```diff
@@ -31,19 +31,37 @@ impl Handler for LogRequests {
...
-            level = level,
-            method = req.method(),
-            path = FullPath(req),
-            ip = request_header(req, "X-Real-Ip"),
-            time_ms = response_time,
-            user_agent = request_header(req, "User-Agent"),
-            request_id = request_header(req, "X-Request-Id"),
-            status = response_code,
-        );
+        if let Some(metadata_length) = req.extensions().find::<u64>() {
```
carols10cents: This adds a lot of duplication. What do you think about, before the `print`, doing something like:

```
let metadata_length = req.extensions().find::<u64>().map_or(String::new(), |n| format!(" metadata_length={}", n));
```

which will set `metadata_length` to an empty string if there isn't a length, and will set `metadata_length` to ` metadata_length=32` if the length is `Some(32)`.

Then the `print` can unconditionally print out the value in `metadata_length`.

Does that make sense? If not, please ask questions!

smokytheangel0: this makes perfect sense, I was wondering myself how to reduce the duplicity of the branches, and I am not very familiar with map/reduce type things, so this will be an excellent change

smokytheangel0:
I think I may have handled git wrong here, I reverted some other changes and added crates.io upstream so that I could stay level with it in the future, I don't see anything wrong with the latest commit though, so please let me know if I messed anything up.

src/middleware/log_request.rs
```diff
@@ -31,10 +31,16 @@ impl Handler for LogRequests {
...
+            .extensions()
+            .find::<u64>()
+            .map_or(String::new(), |l| format!(" metadata_length={}", l));
+
         print!(
             "at={level} method={method} path=\"{path}\" \
              request_id={request_id} fwd=\"{ip}\" service={time_ms}ms \
-             status={status} user_agent=\"{user_agent}\"",
+             status={status} user_agent=\"{user_agent}\"\
+             metadata_length={metadata_length}",
```
carols10cents: When I test this out locally, I'm seeing this on requests that don't have a `metadata_length`:

```
at=info method=GET path="/api/v1/summary" request_id= fwd="" service=44ms status=200 user_agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:72.0) Gecko/20100101 Firefox/72.0"metadata_length=
```

And this on requests that do:

```
at=info method=PUT path="/api/v1/crates/new" request_id= fwd="" service=1060ms status=200 user_agent="cargo 1.39.0 (1c6ec66d5 2019-09-30)"metadata_length= metadata_length=734
```

I think we can remove the `metadata_length=` here now because it's duplicated with [this](https://github.com/rust-lang/crates.io/pull/1905/files#diff-3c16efcf7a504a2aae66b1d607874710R37) ` metadata_length=`.

Could you make that change please? Once that's done, I think this is good to go! And don't worry about the commits, I can fix them up 👍

smokytheangel0: I definitely should have noticed that, thanks.

carols10cents:
Thank you so much!!!

@bors r+

PR #1906, opened 2019-11-19
Add CloudFront IP ranges as trusted for real_ip

jtgeibel:
This change will allow nginx to recurse past the CloudFront IP
addresses and store the correct client IP address in `$remote_addr`.
It is important that the correct client IP is available for rate
limiting on the publish endpoint and logging.

This list will need to be updated periodically.

cc @pietroalbini - This should probably be deployed before we enable CloudFront on production.


sgrif:
I'd like to just replace this with something that pulls from its API at boot time rather than having to remember to update the list periodically, but this is fine for now

Issue #1360, opened 2018-04-17
Send email on crate owner invitation


carols10cents:
If there's an email address for the person being invited (which we don't require yet)

It'd be awesome if the email included a one-click link to accept the invitation without logging in (so with a token like the email verification works now)

Related:

* https://github.com/rust-lang/crates.io/issues/924 (the owner invitation system so far)
* https://github.com/rust-lang/crates.io/issues/1116 (showing a notification on crates.io)

## Instructions

Invitations get created in the [krate.owner_add](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/models/krate.rs#L422) method. 

Currently, if you add someone as an owner multiple times, it just updates the one existing invitation record. We don't want to send multiple emails in this situation, as this would be an easy vector for spamming someone. 

You can tell whether a row was inserted or not when you say `on conflict do update` by adding a `returning` method call to request that the query returns some of the data, getting the result with the `get_result` method, and calling the `optional` method to tell diesel it's ok if it doesn't get a result. If the query did nothing (because that record already existed), nothing will be returned and this diesel statement will result in a `None` value. [Here's an example where we do this](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/models/krate.rs#L185-L192): 

So after [creating an invitation](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/models/krate.rs#L436-L443), and only if that invitation record was created (rather than only updated because it already existed), send the invited user an email.

Get the email address by first getting the [`User`](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/models/user.rs#L14) ([`owner` here](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/models/krate.rs#L435) is an `Owner::User`, change the underscore in that match pattern to a variable name to be able to use the inner `User` [that `Owner::User` holds](https://github.com/rust-lang/crates.io/blob/adf28ad82b4ec2aa8af95c03e42f9f361296278f/src/models/owner.rs#L34)) 

[...]

PR #1909, opened 2019-11-20
Send email on crate owner invitation

nkanderson:
If the recipient of a crate ownership invitation has a verified email address, they will be sent an email indicating that a specified user has invited them to become an owner of a named crate. Only sends the email on the first instance of an invitation.

Completes the first part of #1360.

### DIFF
--- a/src/email.rs
+++ b/src/email.rs
@@ -92,7 +92,7 @@ https://crates.io/confirm/{}",
 /// the database and the user will see the invitation when they visit
 /// https://crates.io/me/pending-invites/.
 pub fn send_owner_invite_email(email: &str, user_name: &str, crate_name: &str) {
-    let subject = "Please confirm crate ownership";
+    let subject = "Crate ownership invitation";
     let body = format!(
         "{} has invited you to become an owner of the crate {}!\n
 Please visit https://crates.io/me/pending-invites to accept or reject
--- a/src/email.rs
+++ b/src/email.rs
@@ -86,6 +86,23 @@ https://crates.io/confirm/{}",
     send_email(email, subject, &body)
 }
 
+/// Attempts to send a crate owner invitation email. Swallows all errors.
+///
+/// Whether or not the email is sent, the invitation entry will be created in
+/// the database and the user will see the invitation when they visit
+/// https://crates.io/me/pending-invites/.
+pub fn send_owner_invite_email(email: &str, user_name: &str, crate_name: &str) {
+    let subject = "Please confirm crate ownership";
+    let body = format!(
+        "{} has invited you to become an owner of the crate {}!\n
+Please visit https://crates.io/me/pending-invites to accept or reject
+this invitation.",
+        user_name, crate_name
+    );
+
+    let _ = send_email(email, subject, &body);
+}
+
 fn send_email(recipient: &str, subject: &str, body: &str) -> CargoResult<()> {
     let mailgun_config = init_config_vars();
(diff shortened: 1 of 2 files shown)


src/models/krate.rs
```diff
@@ -427,14 +428,28 @@ impl Crate {
...
                         invited_by_user_id: req_user.id,
                         crate_id: self.id,
                     })
                     .on_conflict_do_nothing()
-                    .execute(conn)?;
+                    .get_result::<CrateOwnerInvitation>(conn)
+                    .optional()?;
+
+                if maybe_inserted.is_some() {
+                    if let Owner::User(user) = &owner {
```
carols10cents: Because we're in [this arm of the match](https://github.com/rust-lang/crates.io/pull/1909/commits/8838932b407dad8bf51bc8e0807e6bcdbcaa40f6#diff-1307fe21b99862533118a71d3c111feaR430), we know this `if let` will always be true. We should be able to:

- Remove this line
- Change the pattern for this arm to `Owner::User(user)` (I thought we could get both with `owner @ Owner::User(user)` but I just tried it and that doesn't work, but we don't actually need both, keep reading)
- Change [this line](https://github.com/rust-lang/crates.io/pull/1909/commits/8838932b407dad8bf51bc8e0807e6bcdbcaa40f6#diff-1307fe21b99862533118a71d3c111feaR433) to `invited_user_id: user.id` because that's [what `owner.id()` does anyway](https://github.com/rust-lang/crates.io/blob/9042ebf469e18c068ddeadfbef842d482c87fa21/src/models/owner.rs#L80)
- Change [this line](https://github.com/rust-lang/crates.io/pull/1909/commits/8838932b407dad8bf51bc8e0807e6bcdbcaa40f6#diff-1307fe21b99862533118a71d3c111feaR455) to `user.gh_login`, because that's [what `owner.login()` does anyway](https://github.com/rust-lang/crates.io/blob/9042ebf469e18c068ddeadfbef842d482c87fa21/src/models/owner.rs#L73)

That'll let us get rid of a conditional that will always be true and the extra indentation that required. Could you make these changes please?

src/email.rs
```diff
@@ -86,6 +86,23 @@ https://crates.io/confirm/{}",
     send_email(email, subject, &body)
 }
 
+/// Attempts to send a crate owner invitation email. Swallows all errors.
+///
+/// Whether or not the email is sent, the invitation entry will be created in
+/// the database and the user will see the invitation when they visit
+/// https://crates.io/me/pending-invites/.
+pub fn send_owner_invite_email(email: &str, user_name: &str, crate_name: &str) {
+    let subject = "Please confirm crate ownership";
```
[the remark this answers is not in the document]
nkanderson: No problem, please be as picky as you'd like :) And I can see how that original subject sounded like it may refer to an existing ownership confirmation. I've updated it to "Crate ownership invitation".

src/models/krate.rs
nkanderson: Definitely! Yeah, I had attempted the match arm update you'd initially suggested in the issue description, and I think hit the same as you.

I made those updates and ran through a quick test, seems to be working as expected.

nkanderson:
I made the requested changes as an additional commit, but if preferred, I can squash that down to just the one.

nkanderson:
Whoops, looks like I had a formatting issue initially, should be ok now though.

carols10cents:
Looks great now!! Thank you so much!!

@bors r+

PR #1911, opened 2019-11-21
Preparation to drop `email` in User in Rust codes

hbina:
Related: #1888 

This commit(s) prepares the migration to dropping `email` column from the `User` table.
Here, I introduce a new intermediate type called `UserNoEmailType`.

Copied from my latest commit:

Every queries out of `User`'stable should now be converted into an intermediate type called `UserNoEmailType`.
This is done by selecting only the appropriate columns which is defined in `user::ALL_COLUMNS` (notice that it excludes `email`) 
This could be renamed, I guess.
Conversion from this type to `User` is simply a copy except that `email` returned will always be `None`.

I think its possible to remove `email` from User directly because its not actually the mirror of `User`'s table.
The mirror struct is actually `NewUser`.

### DIFF
--- a/src/controllers/krate/metadata.rs
+++ b/src/controllers/krate/metadata.rs
@@ -5,9 +5,11 @@
 //! `Cargo.toml` file.
 
 use crate::controllers::prelude::*;
+use crate::models::user;
+use crate::models::user::UserNoEmailType;
 use crate::models::{
     Category, Crate, CrateCategory, CrateKeyword, CrateVersions, Keyword, RecentCrateDownloads,
-    User, Version,
+    Version,
 };
 use crate::schema::*;
 use crate::views::{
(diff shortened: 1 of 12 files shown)


src/controllers/krate/metadata.rs
```diff
@@ -220,18 +222,21 @@ pub fn reverse_dependencies(req: &mut dyn Request) -> CargoResult<Response> {
 
     let version_ids: Vec<i32> = rev_deps.iter().map(|dep| dep.version_id).collect();
 
+    //  FIXME   ::  Another one of these silly transformation that is necessary to test
+    //              whether or not this function uses the column email from the user
+    //              table. Here, we simply set it to None.
```
carols10cents: Does this comment still apply, since the transformation it's talking about was removed in https://github.com/rust-lang/crates.io/pull/1911/commits/ea51376af883fdc9b12374eda436841a5e15dd09?

hbina: You are right, I don't think this applies anymore. Do you want me to merge these commits into one?

carols10cents:
Thank you so much! I introduced some conflicts by merging some other PRs, so I resolved those and squashed for you while I was in there. Great work!!

@bors r+

hbina:
Thank you!

PR #1907, opened 2019-11-19
Add a `USE_FASTBOOT=staging-experimental` mode

jtgeibel:
This adds a new mode to the `USE_FASTBOOT` environment variable.  The logic is now:

* When deployed on Heroku (staging or production - via `script/start-web.sh`):
  * If `USE_FASTBOOT=staging-experimental` then non-backend requests are sent to FastBoot.
  * If `USE_FASTBOOT` is set and non-zero length, FastBoot is enabled only for allowed paths.
  * Otherwise (`USE_FASTBOOT` is not set, or is zero length), FastBoot is disabled.
* When doing local development (via `script/ember.sh`):
  * If `USE_FASTBOOT` is set, all (initial) frontend requests are served via FastBoot.  This is equivalent to `staging-experimental` above.
  * If `USE_FASTBOOT` is not set, FastBoot is disabled.

Because our development environment doesn't go through nginx, it isn't possible to support the "FastBoot is enabled only for allowed paths" mode.  However, once the frontend has booted, all further requests hit the backend (unless JS is disabled, then every client navigation results in a new request served via FastBoot).

PR #1912, opened 2019-11-22
Make /install work under FastBoot

kzys:
Will add nginx.conf change after #1907. Early feedback is welcome!

### DIFF
--- a/app/routes/crate/docs.js
+++ b/app/routes/crate/docs.js
@@ -3,13 +3,14 @@ import { inject as service } from '@ember/service';
 
 export default Route.extend({
   flashMessages: service(),
+  redirector: service(),
 
   redirect() {
     let crate = this.modelFor('crate');
 
     let documentation = crate.get('documentation');
     if (documentation) {
-      window.location = documentation;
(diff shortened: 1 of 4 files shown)


jtgeibel:
What do you think about converting `app/routes/crate/repo.js` and `app/routes/crate/docs.js` in this PR as well?  It seems like the `_redirectTo` logic could be shared between these routes.

(`app/routes/github-login.js` sets `window.location` as well, but supporting login without JS is a whole other thing, if we decide to support that use case.)

kzys:
Sure. I can do that. But please deploy/enable `/policies` change first.

jtgeibel:
> But please deploy/enable `/policies` change first. We should not enable three or four FastBoot-enabled routes at once, especially since we don't have any yet.

I agree, we should enable that first.  That change is deployed now, but I don't have the ability to enable it on production, and I can't speak for @carols10cents on when that will be possible.

jtgeibel:
Looks good to me!

This should be safe to deploy at any time, as no changes are made to nginx.

@bors r+

PR #1913, opened 2019-11-22
front-end: Add email notification preferences

DSpeckhals:
This commit adds the front-end capabilities to coincide with the API endpoint that was built on the server. There's also a small bug fixed in the api-token-row; it was throwing an error every time the "/me" route was loaded.

This PR addresses the second item in #1895.

I like how the UI ended up, but it may very well be creators bias :smile:. If this doesn't work, then I'll be happy to iterate!

### Screenshot of a crate with notifications on and one that's off
image

### DIFF
--- a/app/components/api-token-row.js
+++ b/app/components/api-token-row.js
@@ -8,7 +8,7 @@ export default Component.extend({
 
   didInsertElement() {
     let input = this.element.querySelector('input');
-    if (input.focus) {
+    if (input && input.focus) {
       input.focus();
     }
   },
--- a/app/templates/me/index.hbs
+++ b/app/templates/me/index.hbs
@@ -25,6 +25,39 @@
   {{email-input type='email' value=model.user.email user=model.user}}
 </div>
 
+<form id='me-email-notifications' {{ action 'saveEmailNotifications' on='submit' }} >
+  <h2>Email Notification Preferences</h2>
+
+  <p>
+    To aid detection of unauthorized crate changes, we email you each time a new version of a crate you own is pushed.
+    By receiving and reading these emails, you help protect the Rust ecosystem.
+  </p>
+
+  <div class='row'>
+    <ul>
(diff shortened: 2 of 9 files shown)


app/components/api-token-row.js
```diff
@@ -8,7 +8,7 @@ export default Component.extend({
 
   didInsertElement() {
     let input = this.element.querySelector('input');
-    if (input.focus) {
+    if (input && input.focus) {
```
carols10cents: Great catch on this! I wish I had seen this before this weekend; someone reported revoking tokens wasn't working so I reverted a commit because I didn't know what the fix was right away. [I'm glad you came up with the same fix I eventually did](https://github.com/rust-lang/crates.io/pull/1923)!

DSpeckhals: Nice. Once your PR is merged, the current merge conflict should resolve itself. Thanks.

app/templates/me/index.hbs
```diff
@@ -25,6 +25,39 @@
   {{email-input type='email' value=model.user.email user=model.user}}
 </div>
 
+<form id='me-email-notifications' {{ action 'saveEmailNotifications' on='submit' }} >
+  <h2>Email Notification Preferences</h2>
+
+  <p>
+    To aid detection of unauthorized crate changes, we email you each time a new version of a crate you own is pushed.
+    By receiving and reading these emails, you help protect the Rust ecosystem.
```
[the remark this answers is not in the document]
DSpeckhals: It wasn't my wording originally. Most of it was taken directly from the RubyGems preferences page, so all credit is due to them :smile:. Your wording clarification fits nicely. I'll add that.

DSpeckhals:
Here's what Select/Deselect All button could look like based on the most recent commit:

image

carols10cents:
This looks great!! Thank you so much!! 

@bors r+

PR #1917, opened 2019-11-23
Update chrono, dotenv, num-*, migrations_macros, and derive_deref

jtgeibel:
Overall, this drops 16 old versions of dependencies.  The `num-*` crates are on the list at #1805, but overall I think this is low risk to deploy together.

* `cargo update -p num -p num-iter -p num-traits --aggressive`
* `cargo update -p chrono`
* `cargo update -p migrations_macros -p derive_deref`
* Update to latest `dotenv`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,7 +77,7 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "0.1.2"
+version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


jtgeibel:
I've added a commit bumping `chrono`, because the newer version drops a few deps on old `num-*` crates and doesn't affect the rest of the dependency tree.

jtgeibel:
I've added a final commit that bumps `migrations_macros` and `derive_deref`.  This drops some old versions of `quote`, `syn`, `synom`, and `unicode-xid`!  I would have landed this as a separate PR, except there was one small conflict that would need to be rebased and overall I think this PR is still low risk to deploy.

carols10cents:
Nice! 

@bors r+

PR #1918, opened 2019-11-23
Move 3 backend routes to `/api/private/session`

jtgeibel:
This PR moves 3 remaining backend routes so that all backend routes are now scoped under `/api/`.  Upon further examination, I've determined that the initial OAuth callback is handled by the `/authorize/github` frontend route, which has not moved.  Therefore, this PR does not need to be synchronized with any changes to the GitHub OAuth settings!


carols10cents:
@bors r+

I can log in and out locally and the code looks good, I'll definitely be deploying to staging and testing there before deploying to prod :)

PR #1920, opened 2019-11-23
Begin refactoring error handling/fallback logic

jtgeibel:
This PR is best reviewed as individual commits.  The first commits rename some types and functions without changing anything.  The final commits remove some unused logic and lay some groundwork for future refactoring.

The overall intent of this PR is to more clearly document the existing behavior.  In particular, the top level error types no longer include `Cargo` in the name as not all error types are intended for cargo endpoints.

r? @smarnach


jtgeibel:
> I'd also be fine with naming the error and result times `Error` and `Result`, which is a common pattern, but this would create quite a few conflicts we would have to fix manually, which doesn't seem worth it.

My thoughts as well.

@bors r=smarnach

PR #1921, opened 2019-11-24
Remove conversion from hyphen to underscore

hskang9:
resolves cargo's [#2775](https://github.com/rust-lang/cargo/issues/2775) issue

### DIFF
--- a/migrations/20150319224700_dumped_migration_93/up.sql
+++ b/migrations/20150319224700_dumped_migration_93/up.sql
@@ -1,1 +1,2 @@
 CREATE FUNCTION canon_crate_name(text) RETURNS text;
+
--- a/migrations/20150319224700_dumped_migration_93/up.sql
+++ b/migrations/20150319224700_dumped_migration_93/up.sql
@@ -1,4 +1,1 @@
-CREATE FUNCTION canon_crate_name(text) RETURNS text AS $$
-                    SELECT replace(lower($1), '-', '_')
-                $$ LANGUAGE SQL
-            ;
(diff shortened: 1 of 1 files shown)


carols10cents:
Hi, sorry, but I think there's a misunderstanding. Fixing https://github.com/rust-lang/cargo/issues/2775 means that whether a Cargo.toml file says `crate-name` or `crate_name`, it will resolve to the one crate with `crate_name` on crates.io. This PR would have the effect of allowing the publishing of *both* `crate-name` and `crate_name` as *different* crates, which would be potentially confusing.

hskang9:
Well making - and _ different name was what I was trying to do, but if that creates a confusion then, I should not make this pr. But here is the confusion happening in cargo side. For example, I uploaded cargo package ‘hyphen-example’ but crates.io considers it as ‘hyphen_example’. So I should be able to import package writing “hyphen_example” but running cargo install ‘hyphen_example’ cannot find the package, but when I run cargo install ‘hyphen-example’ then it successfully installs the package. Should I add same conversion logic in cargo as crates.io applies?

Eh2406:
> Should I add same conversion logic in cargo as crates.io applies?

Short answer: Yes.
Long answer: We need to find a way to do the "conversion logic" for crates.io and not for other registries, or decide to break existing functionality for other registries. We also need to find a way to best support older cargos that do not have the "conversion logic", or decide to break existing users. Neither will be straightforward https://github.com/rust-lang/cargo/issues/2775#issuecomment-557188111

PR #1923, opened 2019-11-24
Redo api row

carols10cents:
@locks I think this fixes the revoke button -- what I saw was `TypeError: input is null` on page load, when there isn't any input because a row hasn't been inserted yet.

This reverts the reverts and adds a null check.

r? @locks

### DIFF
--- a/app/components/api-token-row.js
+++ b/app/components/api-token-row.js
@@ -8,7 +8,7 @@ export default Component.extend({
 
   didInsertElement() {
     let input = this.element.querySelector('input');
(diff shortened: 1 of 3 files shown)


locks:
Reviewing! I'm getting sidetracked due to unrelated test failures locally in my machine, on this branch and master.

carols10cents:
I'm going to merge this because DSpeckhals came up with the same fix in #1913 and I want to get this fix in separately from that PR.

@bors r+

Issue #1915, opened 2019-11-22
Add cache-control headers for readme files.


smarnach:
We added cache-control headers for `.crate` files in #1871. We should also add cache-control headers for readme files. While it is rare, readme files can change when we regenerate them, so we can't mark them as immutable, and should limit the cache time to some lower value, say one week.

hbina:
So I assume https://github.com/rust-lang/crates.io/blob/1ccd73057330af607067a0fec84044f424c87432/src/uploaders.rs#L90 should no longer wraps `extra_headers` in `Option<T>`?

rye:
Do I have it right that #1924 resolved this?

smarnach:
@rye Yes, this is fixed now.

PR #1924, opened 2019-11-24
Added cache control to readme files

hbina:
Related issue:  #1915 

`extra_headers` parameter is no longer wrapped in `Option<T>`

### DIFF
--- a/src/bin/render-readmes.rs
+++ b/src/bin/render-readmes.rs
@@ -21,7 +21,7 @@ use chrono::{TimeZone, Utc};
 use diesel::{dsl::any, prelude::*};
 use docopt::Docopt;
 use flate2::read::GzDecoder;
-use reqwest::Client;
+use reqwest::{header, Client};
 use tar::{self, Archive};
 
 const DEFAULT_PAGE_SIZE: usize = 25;
@@ -137,7 +137,7 @@ fn main() {
                         content,
                         content_length,
                         "text/html",
-                        None,
+                        header::HeaderMap::new(),
                     )
                     .unwrap_or_else(|_| {
                         panic!(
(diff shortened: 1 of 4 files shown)


src/bin/render-readmes.rs
```diff
@@ -137,7 +137,7 @@ fn main() {
                         content,
                         content_length,
                         "text/html",
-                        None,
+                        header::HeaderMap::new(),
```
smarnach: Shouldn't we set a TTL here as well?

(For context, this is a binary that re-renders all readme files. It can be manually run if we make updates to the readme rendering code.)

src/uploaders.rs
```diff
@@ -163,13 +163,18 @@ impl Uploader {
         let path = Uploader::readme_path(crate_name, vers);
         let content_length = readme.len() as u64;
         let content = Cursor::new(readme);
+        let mut extra_headers = header::HeaderMap::new();
+        extra_headers.insert(
+            header::CACHE_CONTROL,
+            CACHE_CONTROL_IMMUTABLE.parse().unwrap(),
```
smarnach: The variable `CACHE_CONTROL_IMMUTABLE` is set to `public,max-age=31536000,immutable`, i.e. the files are tagged to never change and to live for one year in the cache. However, the readme files _might_ change, so let's cache them only for a week, and let's not tag them as immutable. (You may want to read the [documentation of the cache-control header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)).

hbina: I see, sorry for misunderstanding the original Issue. So I assume the only field that is required is `max-age=<seconds>`?

smarnach: I think you can keep `public` as well. Not sure whether it has any effect when setting `max-age` at the same time, but it certainly doesn't hurt.

smarnach:
@bors r+

This should be low risk to deploy.

PR #1927, opened 2019-11-25
Missed a conversion from UserNoEmailType to User

carols10cents:
When reviewing #1911, I missed a spot. I reverted the PR merge on master, so this PR reverts the revert and adds the fix.

I tried to write a test that would have caught this, but I utterly failed at making a real-enough app/request that would have `.user()` but would let me set the `session()` :(

Also a lot of this code is temporary and will be undone once I've rebased #1891, but this needs to be deployed separately from #1891 to ensure there aren't any uses of the email column.

r? @jtgeibel

### DIFF
--- a/src/bin/transfer-crates.rs
+++ b/src/bin/transfer-crates.rs
@@ -7,7 +7,7 @@
 
 use cargo_registry::{
     db,
-    models::{Crate, OwnerKind, User},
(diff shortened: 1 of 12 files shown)


jtgeibel:
LGTM!

@bors r+

PR #1929, opened 2019-11-28
Error response refactoring step 2

jtgeibel:
This is a continuation of #1920.  This series captures several changes:

* The controller prelude is split to provide helpers for either cargo or frontend endpoints.  Eventually this will be unnecessary, but I see it as a temporary guard rail to encourage the correct usage.
* Several user and session frontend endpoints are updated to return a response with the correct HTTP status.  Tests were updated as necessary.
* A `util::errors::http` module is added for concrete error types associated with HTTP status codes.  There are more types to be moved here, but for review purposes that can be done in another PR.
* Once `cargo_err` was refactored to use `http::Ok`, all of the description and fallback logic in the `AppError` trait and `ConcreteAppError` type was no longer necessary.

Here is what I have in mind for the next steps in this refactor:

* Add a `not_found` helper that allows a 404 status with an error message for the client.  (TBD if this can be merged with the existing `NotFound` type or if that functionality should remain independent.)
* Finish moving types to `http` as applicable.
* Review remaing existing usage of helpers within controllers.

After that groundwork, my final step is to investigate the ability to sanitize error response for cargo at the router level so that `cargo_err` can be removed.  It isn't really practical currently to use the correct helper type from within models, since model functionality can be used from either type of endpoint (or both).

The router seems like a strange place for that behavior, but it's where the [error to response conversion](https://github.com/rust-lang/crates.io/blob/4a1542cb17ae50c2bfd0f7e85876d82617e30415/src/router.rs#L139) already happens, and the middleware layer only sees a generic `dyn Error`, not our `AppError`.

With a central place to sanitize errors for old cargo versions, we could even enable it only for old (and empty) user agents.

r? @smarnach

### DIFF
--- a/src/util/errors/http.rs
+++ b/src/util/errors/http.rs
@@ -11,10 +11,6 @@ pub(super) struct Ok(pub(super) String);
 pub(super) struct ServerError(pub(super) String);
 
 impl AppError for Ok {
-    fn description(&self) -> &str {
-        self.0.as_ref()
-    }
-
     fn response(&self) -> Option<Response> {
         Some(json_response(&Bad {
             errors: vec![StringError {
@@ -31,10 +27,6 @@ impl fmt::Display for Ok {
 }
 
 impl AppError for ServerError {
-    fn description(&self) -> &str {
-        self.0.as_ref()
-    }
-
     fn response(&self) -> Option<Response> {
         let mut response = json_response(&Bad {
             errors: vec![StringError {
--- a/src/util/errors/http.rs
+++ b/src/util/errors/http.rs
@@ -6,7 +6,29 @@ use super::{AppError, Bad, StringError};
 use crate::util::json_response;
 
 #[derive(Debug)]
+pub(super) struct Ok(pub(super) String);
+#[derive(Debug)]
 pub(super) struct ServerError(pub(super) String);
+
+impl AppError for Ok {
+    fn description(&self) -> &str {
+        self.0.as_ref()
+    }
+
+    fn response(&self) -> Option<Response> {
+        Some(json_response(&Bad {
+            errors: vec![StringError {
+                detail: self.0.clone(),
+            }],
+        }))
+    }
+}
+
+impl fmt::Display for Ok {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
 
 impl AppError for ServerError {
     fn description(&self) -> &str {
--- a/src/util/errors/http.rs
+++ b/src/util/errors/http.rs
@@ -1,0 +1,31 @@
+use std::fmt;
+
+use conduit::Response;
+
+use super::{AppError, Bad, StringError};
+use crate::util::json_response;
+
+#[derive(Debug)]
+pub(super) struct ServerError(pub(super) String);
+
+impl AppError for ServerError {
+    fn description(&self) -> &str {
+        self.0.as_ref()
+    }
+
+    fn response(&self) -> Option<Response> {
+        let mut response = json_response(&Bad {
+            errors: vec![StringError {
+                detail: self.0.clone(),
+            }],
+        });
+        response.status = (500, "Internal Server Error");
+        Some(response)
+    }
+}
+
+impl fmt::Display for ServerError {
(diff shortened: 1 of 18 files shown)


jtgeibel:
For a bit further context, the tests for the generic error to response behavior are [here](https://github.com/rust-lang/crates.io/blob/4a1542cb17ae50c2bfd0f7e85876d82617e30415/src/router.rs#L186-L229).  Basically, the error handling behavior is:
* If `AppError::response` returns `Some(_)` then that response is sent to the user.  When the error is converted to a response like this, it is no longer available for logging in the middleware layer (beyond seeing the HTTP status code, or extracting details from the JSON).
* If `AppError::response` returns `None` then the error is converted to a `Box<dyn Error>`.  This allows for more detailed logging, and is eventually converted to a generic status 500 response by the `EnsureWellFormed500` middleware.  (This applies to things like `internal`, and the generic `AppError` impl for `Error`s.)

jtgeibel:
I've pushed an additional commit to resolve an issue in logging chained errors that was uncovered by this branch.

Returning the following from a controller

```rust
None.chain_error(|| internal("inner error")).chain_error(|| internal("middle error")).chain_error(|| internal("outer error"))
```

results in logging only the outer error on `master`:

```
[...] error="outer error"
```

With the changes in this PR to go through `fmt::Display`, now two implementations became active:

```
[...] error="outer error caused by middle error caused by inner error
Caused by: middle error caused by inner error
Caused by: inner error"
```

With the final commit here, we now see the expected log output:

```
[...] error="outer error caused by middle error caused by inner error"
```

I've added it to this PR since it doesn't really fix anything as a standalone fix on master.

jtgeibel:
I've rebased to resolve merge conflicts.

I also have a series of commits [building on this branch](https://github.com/jtgeibel/crates.io/compare/fix-some-error-responses...error-refactor-step3).  I'll open a separate PR for that once this lands and I have a chance to add more documentation.

smarnach:
@jtgeibel Apologies for not the delay in reviewing this – I'll get to this in the weekend.

src/util/errors/http.rs
```diff
@@ -0,0 +1,45 @@
+use std::fmt;
+
+use conduit::Response;
+
+use super::{AppError, Bad, StringError};
+use crate::util::json_response;
+
+#[derive(Debug)]
+pub(super) struct Ok(pub(super) String);
```
smarnach: Why is this `pub(super)`? I assumed the super module can see this anyway, and `pub(super)` is only needed if access from sibling modules is required.

jtgeibel: Without `pub(super)` I would have to move the `cargo_err`, `server_error`, etc helpers into this file, and reexport them.  I'll probably move everything there eventually, but for now I'd like to keep these helpers and their documentation/comments at the top of `error.rs`.

The parent module cannot see these without the `pub(super)`.  Submodules can see private types/fields from parent modules, but not the other way around.

jtgeibel:
> The endpoints you changed the status codes for look like they are not used by Cargo to me, but I don't know all the things that Cargo can do and have not double-checked in the Cargo code base.

My understanding is that with the exception of download, all interactions with crates.io are via the `crates-io` crate in the [cargo source tree](https://github.com/rust-lang/cargo/tree/master/crates/crates-io).

> It would be great if we could add integration tests for the interaction with Cargo, so we can be more confident about changes that might cause regressions.

That would be awesome!

> Would it be possible to add a test that error chains are logged correctly? Preferably in a fuzzy way, e.g. only making sure that the error message of each error in the change appears exactly once in the log.

I've pushed some commits that add tests for the `chain_err` behavior.  I've added comments to identify the combinations where a user facing error wraps a more concrete error, causing no error to be logged (because a valid response was sent to the user).

I don't think its worth adding anything more complicated at the moment (such as integrating with the logging middleware to capture the actual log output).  Right now most uses of `chain_err` in the codebase are to convert a `None` into a user facing error.  Also, after my next set of commits, it may make sense to impl `chain_err` on the new concrete `Error` enum type.  Thus eliminating any misuse of `chain_err` to wrap an error already intended for the user.

src/util/errors/http.rs
smarnach: You are right, I got confused there.

smarnach:
Looks good to me.

@bors r+

Issue #310, opened 2016-04-13
Just Updated column lists greatest, not the latest version


bluss:
Right now the Just Updated column lists `ndarray (0.5.0-alpha.1)` even though the new upload that triggered it was `0.4.9`. [ndarray @ crates.io](https://crates.io/crates/ndarray)

carols10cents:
Yep, [this is where the version is coming from](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L727) for the /summary JSON that's used on the home page.

carols10cents:
Ok, so currently the way we get the "just updated" list is:

- [Query](https://github.com/rust-lang/crates.io/blob/21e04f3028624485621d577359008560728a37d3/src/controllers/krate/metadata.rs#L47-L52) for crate records where `created_at` != `updated_at` (because if they were equal, that would mean the first version of the crate was just published), order by `updated_at` descending, and take the first 10.
- Using the same [`encode_crates` closure](https://github.com/rust-lang/crates.io/blob/21e04f3028624485621d577359008560728a37d3/src/controllers/krate/metadata.rs#L29-L40) as the other queries do, load all the version records for the crates in this list and pick the largest semver.

There are a number of possible solutions here; I'm not sure which is best.

- A solution that doesn't change the database queries at all would be that for the "just updated" list only, use a different function than the `encode_crates` closure. Instead of picking the `Version::max`, pick the version that has `created_at` equal to the crate's `updated_at`.
- Along the same lines but implemented by changing the queries, instead of getting ALL versions for the 10 "just updated" crates, query for only the versions of the crate whose `created_at` is equal to the crate's `updated_at`. Probably a tiny bit more efficient, but adds complexity to the queries, and I don't think we're worried about the performance of the existing queries.
- Change the query to sort the *versions* table by created_at descending instead of the *crates* table, and exclude versions that have their created_at equal to their crate's created_at to exclude new crates. This query seems to get me the same crates as the original:

```
select crates.name, versions.id, versions.num, crates.updated_at, versions.created_at 
from versions 
join crates on crates.id = versions.crate_id 
and crates.created_at != versions.created_at 
order by versions.created_at desc 
limit 10;
```

I kind of like this solution because it does one query instead of 2, but it diverges the "just updated" query the most from the other queries, which adds maintenance complexity. I'm also not sure off the top of my head how to translate this to diesel.

[...]

PR #1931, opened 2019-12-01
Newest updated

bruceadams:
This PR is working toward addressing #310. This is incomplete. There is a relevant test commented out because it fails.


bruceadams:
This is pretty far from correct 😞 … some discussion on #310.

bruceadams:
I'm much happier with this code. Pretty much everywhere that used to have some concept of max_version now has top_versions which contains both the highest semver version and the most recently updated version.

What is not yet here is _using_ the most recently updated version in the user interface. (I don't really know where this is in the code, but I haven't really looked either.)

src/models/krate.rs
```diff
@@ -206,7 +214,7 @@ impl<'a> NewCrate<'a> {
 }
 
 impl Crate {
-    /// SQL filter based on whether the crate's name loosly matches the given
+    /// SQL filter based on whether the crate's name loosely matches the given
```
carols10cents: ❤️

src/tests/util.rs
```diff
@@ -303,7 +303,7 @@ impl TestAppBuilder {
     }
 }
 
-/// A colleciton of helper methods for the 3 authentication types
+/// A collection of helper methods for the 3 authentication types
```
carols10cents: ❤️

src/controllers/krate/metadata.rs
```diff
@@ -1,13 +1,10 @@
 //! Endpoints that expose metadata about a crate
 //!
-//! These endpoints provide data that could be obtained direclty from the
+//! These endpoints provide data that could be obtained directly from the
```
carols10cents: ❤️

src/models/version.rs
```diff
@@ -39,6 +39,25 @@ pub struct NewVersion {
...
+/// The highest version (semver order) and the most recently updated version.
+/// Typically used for a single crate.
+#[derive(Debug, Clone)]
+pub struct TopVersions {
+    pub highest: semver::Version,
+    pub newest: semver::Version,
+}
+
+/// A default semver value, "0.0.0", for use in TopVersions
+fn default_semver_version() -> semver::Version {
```
carols10cents: Nice extraction!

jtgeibel:
This looks really good to me so far!  A few things I noticed running locally:

* The Just Updated column isn't showing a version, just empty `()`.  I think this may be causing the test failure as well.
* Everything still links to the top-level `crates/crate_name` route.  When a version is displayed, it would be nice if the link went directly to that version's page.

To test the end-to-end behavior, I'm looking to publish the following set of crates to staging (`Cargo.toml` below).  Let me know if you think of any additional edge cases that should be covered.

```toml
[workspace]

# The comments show the intended publish order
members = [
    "prerelease_only",        # 0.0.1-alpha.0
    "prerelease_then_normal", # 0.0.1-alpha.0, 0.0.1
    "normal_then_prerelease", # 0.0.1, 0.1.0-alpha.0
    "normal_release_only",    # 0.0.1
    "normal_releases",        # 0.0.1, 0.1.0, 0.1.1, 0.2.0
    "normal_and_backport",    # 0.1.0, 0.2.0, 0.2.1, 0.1.1
    "normal_and_backport2",   # 0.1.0, 0.2.0, 0.1.1, 0.2.1
]
```

bruceadams:
I see the empty version for Just Updated and struggled to understand how `max_version` worked and `newest_version` did not. I think I have _that_ detail fixed at this point 🎉. I like suggestion about linking straight to a specific version when we display a specific version. I may not get to that today.

app/templates/index.hbs
```diff
@@ -38,19 +38,19 @@
...
   </section>
   <section id='just-updated' data-test-just-updated aria-busy="{{dataTask.isRunning}}">
     <h2>Just Updated</h2>
-    {{crate-list crates=model.just_updated}}
+    {{crate-list-newest crates=model.just_updated}}
   </section>
   <section id='most-recently-downloaded' data-test-most-recently-downloaded aria-busy="{{dataTask.isRunning}}">
     <h2>Most Recent Downloads</h2>
-    {{crate-list crates=model.most_recently_downloaded}}
+    {{crate-list-max crates=model.most_recently_downloaded}}
```
carols10cents: The "Most Recent Downloads" list is for any version of the crate, like "Most Downloaded" is, so I think this should use `crate-list-name-only` too? What do you think?

app/templates/components/crate-list-newest.hbs
```diff
@@ -0,0 +1,12 @@
+<ol>
+  {{#each crates as |crate index|}}
+    <li>
+      {{#link-to 'crate.version' crate.id crate.newest_version class='name' data-test-crate-link=index}}
```
carols10cents: So the "Just Updated" list shows a version and links to that version, and the "Most downloaded" lists don't show a version and don't link to a version... I wonder if, for consistency, we should make the "New Crates" list items link to the version because it displays a version? It probably doesn't matter in practice because new crates are only likely to have one version, but what do you think?

bruceadams: I wondered about this. It felt weird to link to a specific version when we expect there to be only one. What do you think of not displaying the version number in the "New Crates" list? The meaning of _version number_ isn't very deep when there is only one.

app/templates/index.hbs
bruceadams: Ah! Right. I hadn't managed to see this. Thanks!

bruceadams:
So… I'm happy with the state of the code and the user interface in this pull request as it currently stands. The main thing I'm uncomfortable about is that the scope of this change has grown well beyond the original issue #310. I am more than willing to trim this back to a narrow scope that only fiddles with the *Just Updated* column and leaves everything else as it is today. I can imagine some in between places as well. Preferences?

app/templates/components/crate-list-newest.hbs
carols10cents: Hmmm. Hmmm... the only slight pause I have is that the version number a new crate starts with is a small signal... if it's 0.0.1, it's probably someone's experiment... if it's 1.0.0, it's the launch of something someone's probably been working on for a while and wants to signal that people should use it... is that an interesting enough reason to keep the display of the version number for new crates? I don't know!

bruceadams: Yes. I can see your point about version number being a signal for new crates. In addition, having the version number visible for new crates doesn't strike me as detrimental on the crates.io home page. So: my remaining question is just: what do we link to?
- Linking to the specific version aligns with the change in this PR makes for _Just Updates_, making link behavior more consistent.
- Linking to the front page of for the crate seems simpler; it gives a shorter url and a smidge less complexity. If someone sat looking at the crates.io front page for a while, then clicked on a new crate, there is a chance that crate had already published another version. Linking to the crate home would make that newer version more visible.

bruceadams: I went with linking to the specific version under *New Crates*. That feels most right to me: consistent with *Just Updated* and I think my stated concern about yet newer versions isn't worth worrying about. The link the user clicks on takes them to what the text says, which, I think, meets the principle of least astonishment.

bruceadams:
Sorry about the test failure 😞. Running the front end test locally stopped working and I haven't figured out why.

tests/acceptance/crate-test.js
```diff
@@ -66,7 +66,7 @@ module('Acceptance | crate page', function(hooks) {
     await visit('/');
     await click('[data-test-just-updated] [data-test-crate-link="0"]');
 
-    assert.equal(currentURL(), '/crates/nanomsg');
+    assert.equal(currentURL(), '/crates/nanomsg/6.8.9');
```
carols10cents: Is this test one that's failing for you locally? I think it's because this test uses the factories, and the version number isn't guaranteed to be the same every time the factory gets called?

This can be fixed by explicitly setting a `newest_version` for this particular instance, sort of like [this test explicitly sets a `max_version`](https://github.com/rust-lang/crates.io/blob/3e5dd28127284e64ff8e4b106ca5dc3472398e5a/tests/acceptance/crate-test.js#L29).

bruceadams: Running the tests locally crashes fairly badly. I don't have much experience with npm stuffs, so I don't have much clue what is going on. It _did_ work for me. I don't know of having changed anything (although I'm a compulsive `brew upgrade`r and that might've broken something).

bruceadams:
I will be at the Rust coffee this evening 🎉  @jtgeibel and I looked at the Percy failures. It seems to be related to auto generated stuff that has interesting cross cutting impact. In some bigger picture, it'd be great to stabilize that stuff.

I'm fine with test cleanup. I'm not going to be able to really look at code for several hours.

tests/acceptance/crate-test.js
carols10cents: Yikes! I have occasional luck with removing `node_modules` and re-`npm install`ing? We can look at it tonight though :)

jtgeibel: Locally, I've replaced this test with the following.  This should eliminate the dependence on faker data and make the test more robust to future changes.  Would you add a final commit with this or a similar update?

```js
  test('visiting a crate page from the front page', async function(assert) {
    this.server.create('crate', { id: 'nanomsg', newest_version: '0.6.1' });
    this.server.create('version', { crate: 'nanomsg', num: '0.6.1' });

    await visit('/');
    await click('[data-test-just-updated] [data-test-crate-link="0"]');

    assert.equal(currentURL(), '/crates/nanomsg/0.6.1');
    assert.equal(title(), 'nanomsg - crates.io: Rust Package Registry');

    assert.dom('[data-test-heading] [data-test-crate-name]').hasText('nanomsg');
    assert.dom('[data-test-heading] [data-test-crate-version]').hasText('0.6.1');
  });
```

bruceadams:
@jtgeibel Your test adjustment makes sense; I added it. The first rebase yesterday took me a while! I was happy the second one was easy!

jtgeibel:
Thanks @bruceadams! I hate to say it, but this needs another rebase now, hopeful another easy one though.  If you get around to it today, ping me and I'll r+ ASAP.  There's been a lot of work on the frontend recently, sorry for the high rate of churn.

bruceadams:
I learned more about Ember.js, including that the [documentation is very good](https://guides.emberjs.com/release/routing/linking-between-routes/#toc_the-linkto--component) even for a reader like me who barely has a clue about Ember.js.

jtgeibel:
Thanks Bruce!

@bors: r+

PR #1932, opened 2019-12-01
Add `-D warnings` to `RUSTFLAGS` on CI

jtgeibel:
By denying warnings explicitly on CI, we don't need to deny warnings in each library or binary.  `#![warn(clippy::all)]` is used to automatically enable clippy in RLS enabled IDEs and `rust_2018_idioms` is now applied consistently.  Finally, clippy allow attributes that are no longer needed have been removed.


carols10cents:
Works great!! Thank you, this should hopefully remove some friction of local development!

@bors r+

PR #1939, opened 2019-12-05
Move check mark SVG to static file

### LINKED ISSUE #1938: Content security policy disallows data SVG in CSS
The [pretty checkmarks](https://github.com/rust-lang/crates.io/blob/master/app/styles/me.scss#L211) that @DSpeckhals added in https://github.com/rust-lang/crates.io/pull/1913 worked locally but are being blocked in production because of our content security policies:

```
Content Security Policy: The page’s settings blocked the loading of a resource at data:image/svg+xml,%3Csvg width='32' hei… (“img-src”).
```

I know it's tiny, but is there a reason we can't just put this as an SVG in [public/assets](https://github.com/rust-lang/crates.io/tree/master/public/assets) @DSpeckhals ?

DSpeckhals:
The content security policy of the production website was set up to not allow data URL's. Moving the logo to a static SVG policy fixes this.

Fixes #1938

### DIFF
--- a/app/styles/me.scss
+++ b/app/styles/me.scss
@@ -208,7 +208,7 @@
             &:after {
                 background-color: #cfc487;
                 border-color: #cfc487;
-                background-image: url('assets/check-mark.svg');
+                background-image: url('/assets/check-mark.svg');
                 background-repeat: no-repeat;
                 background-position: 3px 4px;
             }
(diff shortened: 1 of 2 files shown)


carols10cents:
I tried this out locally and it was trying to load from `/assets/assets/check-mark.svg`, so it looks like the path needed to start with a slash. I went ahead and made that change, tested it out on staging to be sure, and this works! Thank you for the quick fix!

@bors r+

PR #1940, opened 2019-12-05
Use `CrateOwner::by_owner_kind` everywhere to exclude deleted ownerships

carols10cents:
I found some places I missed checking for where we should be using the new `CrateOwner::by_owner_kind` helper method added in https://github.com/rust-lang/crates.io/pull/1901/commits/6d94155b4f80272adedc0c0fe3dbec77cf82f7e1 that adds the filter for `deleted` being false.

I've also added one test that currently passes and two tests that failed before this fix to hopefully cover everywhere that selects crate owners; the other places already have tests.


smarnach:
LGTM

@bors r+

Issue #1036, opened 2017-09-02
Download breakdown by compatible versions


bluss:
Imagine [a crate](https://crates.io/crates/itertools) with some history.

We want to answer the question, is _someone still using_ `0.4.x` and _should it get bugfixes_?

It would be very useful if the download statistics could show statistics grouped by cargo's default compatible version cohorts, meaning that all 0.4.* downloads are one line, 0.5.* are one line, etc.

I don't consider it so useful to have to go through the *20* current versions inside 0.4.* manually.

mindriot101:
I'm happy to work on this issue if it's still wanted? I'm guessing something like the attached image (sorry for the poor quality, I'm not a designer!) breaking down the number of downloads for each minor version of a crate. I've assumed _over 90 days_ is the sort of metric you're interested in? Maybe worth adding this for crate owners only?
![crates-download-by-minor-version](https://user-images.githubusercontent.com/59756/70093788-12957680-1619-11ea-9b7b-87dc050097e5.png)

bluss:
This still seems useful to me! I can see it has not gotten much attention in terms of +1's but it might be a quite invisible issue.

PR #1941, opened 2019-12-05
Recent download stats by minor version

mindriot101:
Implement an implementation of the design suggested and sketched out in #1036.

This PR implements the basic functionality of showing a downloads graph, but there are still a few issues:

1. the data available _cheaply_ from the existing ember `model` shows _total_ downloads, not _downloads over the last 90 days_ which is likely more useful for the purposes suggested in #1036;
2. the graph has not been tested for crates with an extreme number of `major.minor` versions, e.g. >10. There may need to be some filtering if this is the case, when the number of crates is high, or only return the top 10 downloaded versions even if they are not consecutive;
3. it might be good to let the user filter based on some parameters, e.g. let them:
    * select the timespan desired, e.g. "all time", or "recent"
    * select the type of grouping, e.g. `major`, `major.minor`

I have added a basic test for the title of the plot, however I have not implemented a test for the graph itself. I don't think there is a test for the graph on the crate page, so perhaps this is not so bad?

---

I would appreciate some guidance on how to get the last 90 days downloaded stats from the models. I presume as it's not in the `crate` model itself, either:

1. we add it to the crate model, update the API to return it, or
2. we add an additional network fetch, or
3. something I've not thought of.

Option 1 feels like returning a lot of nearly duplicate data that is only used in one place. This also adds scope creep to the API endpoint. Option 2 would add additional complexity of further async network requests that may not be required.


mindriot101:
For context the page looks like this:

Screenshot from 2019-12-06 08-33-33

My original draft design is in the issue #1036

carols10cents:
This is really neat!! Thank you for working on this!!!

I tried out a few crates and the graphs are showing some interesting info, and they even look great on mobile widths!

mobile view of rand crate's graph

To do some stress testing, I found some crates that have a large number of versions-- in particular, rustc-ap-serialize has released 631 major versions, so its graph looks like this:

rustc-ap-serialize's graph

This makes me wonder if we should limit the number of versions at some point to a number that can be displayed in a visible way on the graph?

I also noticed in your mockups on the issue that the labels on the y-axis are "0.4.x", "0.3.x", etc but the labels on the y-axis using this PR are "0.4.0", "0.3.0", etc-- which can be a little confusing when, say, only "0.0.1" exists but "0.0.0" doesn't actually exist:

Screen Shot 2019-12-09 at 9 27 33 AM

I think changing the labels to be what you have in the mockup would be great, is that something you just haven't had a chance to do yet?

> I would appreciate some guidance on how to get the last 90 days downloaded stats from the models. I presume as it's not in the `crate` model itself, either:
> 
>     1. we add it to the crate model, update the API to return it, or
> 
>     2. we add an additional network fetch, or
> 
>     3. something I've not thought of.
> 
> Option 1 feels like returning a lot of nearly duplicate data that is only used in one place. This also adds scope creep to the API endpoint. Option 2 would add additional complexity of further async network requests that may not be required.

I was refreshing my memory about what the existing download graph does-- it shows downloads over the last 90 days, and uses the data returned from `http://localhost:4200/api/v1/crates/{crate-name}/downloads`. It's only the top 5 versions, and the download numbers are broken down by day.

I think to add a graph similar to the one you've added but for the last 90 days, we should create a new API endpoint that returns just the information you need-- all downloads from the last 90 days for all versions of the crate, grouped by minor version (further grouping those by major version could be done on the frontend).

We could make it such that this network request is only performed if you visit the `http://localhost:4200/crates/{crate-name}/versions` page, so I don't think that one extra network request for providing useful information when it's being shown should be a problem.

What do you think?

All these thoughts were from playing around with this PR; going to take a look at the code now :) Thank you so much!

app/components/versions-graph.js
```diff
@@ -39,7 +39,10 @@ export default Component.extend({
       let minor = semver.minor(v.num);
       let downloads = v.downloads;
 
-      let key = `${major}.${minor}`;
+      // XXX ugly hack to get semver to parse the version correctly later on.
+      // We want to do a semver-aware sort, but the `semver.lt` function only
+      // understands version triples, not doubles.
+      let key = `${major}.${minor}.0`;
```
carols10cents: Aha, this is why there's a `.0` in the label :)

app/components/versions-graph.js
```diff
@@ -55,7 +58,7 @@ export default Component.extend({
     ];
 
     // Update plotData with rows in the correct format for google visualization library
-    for (var [key, value] of downloadsPerVersion) {
+    for (var [key, value] of sortIncreasingBySemver(downloadsPerVersion)) {
       plotData.push([key, value, '#62865f', value]);
```
carols10cents: Maybe in here we could change the trailing `.0` to `.x` before pushing into `plotData`?

mindriot101:
> This is really neat!! Thank you for working on this!!!
> 
> I tried out a few crates and the graphs are showing some interesting info, and they even look great on mobile widths!
> 

Great, glad to hear it! I had not specifically tweaked this or targeted this use case, it is a consequence of copying your crate downloads graph which is responsive already.

> To do some stress testing, I found some crates that have a large number of versions-- in particular, rustc-ap-serialize has released 631 major versions, so its graph looks like this:
> 
> This makes me wonder if we should limit the number of versions at some point to a number that can be displayed in a visible way on the graph?

Yes I had considered this as well. We can choose between the most downloaded `N` versions, or the `N` most recent versions. I think the `N` most downloaded would suit the best as this serves the original use case of "which versions can I stop maintaining", and is probably the most interesting to people.

An extension could be to let the user choose what the `N` is in this case, though initially I think we should pick.

> I also noticed in your mockups on the issue that the labels on the y-axis are "0.4.x", "0.3.x", etc but the labels on the y-axis using this PR are "0.4.0", "0.3.0", etc-- which can be a little confusing when, say, only "0.0.1" exists but "0.0.0" doesn't actually exist:
> 
> I think changing the labels to be what you have in the mockup would be great, is that something you just haven't had a chance to do yet?

Yeah as you saw in the code, I needed to add a ".0" section to get semver to parse it correctly. I will give it a go to see if it parses `0.1.x` for example. Otherwise it is a presentation problem and I can update the presentation code.

> > I would appreciate some guidance on how to get the last 90 days downloaded stats from the models. I presume as it's not in the `crate` model itself, either:
> > ```
> > 1. we add it to the crate model, update the API to return it, or
> > 
> > 2. we add an additional network fetch, or
> > 
> > 3. something I've not thought of.
> > ```
> > 
> > 
> > Option 1 feels like returning a lot of nearly duplicate data that is only used in one place. This also adds scope creep to the API endpoint. Option 2 would add additional complexity of further async network requests that may not be required.
> 
> I was refreshing my memory about what the existing download graph does-- it shows downloads over the last 90 days, and uses the data returned from `http://localhost:4200/api/v1/crates/{crate-name}/downloads`. It's only the top 5 versions, and the download numbers are broken down by day.
> 
> I think to add a graph similar to the one you've added but for the last 90 days, we should create a new API endpoint that returns just the information you need-- all downloads from the last 90 days for all versions of the crate, grouped by minor version (further grouping those by major version could be done on the frontend).
> 
> We could make it such that this network request is only performed if you visit the `http://localhost:4200/crates/{crate-name}/versions` page, so I don't think that one extra network request for providing useful information when it's being shown should be a problem.
> 
> What do you think?

That sounds like a good plan. I think you're right, the scope of the new endpoint will be fairly limited so only people that view the versions page will be accessing that endpoint.

I will try to get something going fairly soon.

> All these thoughts were from playing around with this PR; going to take a look at the code now :) Thank you so much!

Glad to help, I'm super excited to give back to the Rust community!

app/components/versions-graph.js
mindriot101: I've implemented this change in the latest pushed commit.

mindriot101:
If you're happy to wait a little bit, I will give the additional endpoint a go. Do you think it makes sense to have such a graph? Would you prefer a graph for all time, or the last 90 days?

bjorn3:
Maybe group `1.0.x` and `1.1.x` together as `1.x.y`, because a `1.0.x` requirement is satisfied by a `1.1.x` version.

bjorn3:
Also `0.0.x` and 0.0.y` are not compatible.

mindriot101:
I've just pushed some changes implementing a backend API endpoint. [I'd appreciate a good review of it](https://github.com/rust-lang/crates.io/blob/b1065745f715f38009716271db38211b4a2edc85/src/controllers/krate/downloads.rs#L72) for what can be improved as I'm not too familiar with Diesel. I've tried to take inspiration from existing code. [I've also added a test, which I'm not confident I've done very well.](https://github.com/rust-lang/crates.io/blob/b1065745f715f38009716271db38211b4a2edc85/src/tests/krate.rs#L633).

carols10cents:
@bjorn3 

> Maybe group `1.0.x` and `1.1.x` together as `1.x.y`, because a `1.0.x` requirement is satisfied by a `1.1.x` version.

While that is true for the default version specification without any operator (which has the same behavior as using a caret), library authors might be interested in people using [tilde requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#tilde-requirements).

> Also `0.0.x` and `0.0.y` are not compatible.

You are correct-- I thought they were, according to Cargo when using caret specifications, but it turns out [there's an exception when the minor version is `0`.](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements). Given that Cargo's default version when running `cargo new` is 0.1.0, I expect this situation to be relatively rare in practice (I can run some queries to be sure). I'm not saying we *shouldn't* break these out separately for correctness's sake, just that if it turns out to be prohibitively complex or expensive, we should consider the rarity of this case. I'm going to take a look at the current code now and see. Any thoughts on this @mindriot101 ?

src/controllers/krate/downloads.rs
```diff
@@ -67,3 +67,70 @@ pub fn downloads(req: &mut dyn Request) -> AppResult<Response> {
...
+    // XXX I am not sure how to do this in the database yet, with Diesel's API, so for the time
+    // being, perform this aggregation in Rust.
+    let downloads = VersionDownload::belonging_to(versions.as_slice())
+        .filter(version_downloads::date.gt(date(now - ndays.days())))
+        .load(&*conn)?
+        .grouped_by(versions.as_slice())
+        .into_iter()
+        .map(|grouped_versions: Vec<VersionDownload>| {
+            let total_downloads = grouped_versions.iter().map(|v| v.downloads).sum();
+            // XXX(perf) this is slow, iterating over the versions array looking for the matching
```
carols10cents: I think you can zip the results of the query with `versions` to get the matching version for each group, like [the search query does with crates and versions and other items here](https://github.com/rust-lang/crates.io/blob/3e5dd28127284e64ff8e4b106ca5dc3472398e5a/src/controllers/krate/search.rs#L177-L192)? There's some documentation with more explanation [here](https://docs.diesel.rs/diesel/associations/index.html) (search for `grouped_by` to get to the relevant section).

src/tests/krate.rs
```diff
@@ -618,6 +629,55 @@ fn versions() {
...
+            .first(conn)
+            .unwrap();
+
+        // Update the download count for some of these versions
+        VersionDownload::create_or_increment(v1, conn).unwrap();
+        VersionDownload::create_or_increment(v1, conn).unwrap();
+        VersionDownload::create_or_increment(v1, conn).unwrap();
+
+        VersionDownload::create_or_increment(v2, conn).unwrap();
+        VersionDownload::create_or_increment(v2, conn).unwrap();
```
carols10cents: In general, I think this test setup is fine, but it's a bit verbose and a bit hard to match up with the assertions.

We have [a function on the crate builder](https://github.com/rust-lang/crates.io/blob/3e5dd28127284e64ff8e4b106ca5dc3472398e5a/src/tests/builders.rs#L212-L217) that sets recent downloads on a crate ([by adding downloads to an arbitrary version of the crate](https://github.com/rust-lang/crates.io/blob/3e5dd28127284e64ff8e4b106ca5dc3472398e5a/src/tests/builders.rs#L260-L270)), I think we could do something similar on the version builder to enable adding downloads to a particular version?

`CrateBuilder::version` can also take a `VersionBuilder` rather than just a version string, so the test setup could get rid of these highlighted lines and do something like this instead:

```
let _krate = CrateBuilder::new("foo_versions", user.id)
            .version(VersionBuilder::new("0.5.1").recent_downloads(3))
            .version(VersionBuilder::new("1.0.0").recent_downloads(2))
            .expect_build(conn);
```

I'm also wondering if we should be testing that this endpoint doesn't include non-recent download stats, so we'd need a way to set older downloads. The crate builder isn't a great example for this as it kind of cheats. `VersionDownload` also doesn't have a nice function for setting downloads for a particular date because usually we only want to set downloads for today. The [populate](https://github.com/rust-lang/crates.io/blob/master/src/bin/populate.rs) utility script for setting up test data has an example of doing this, though.

What do you think? Does that make sense?

bjorn3:
> I expect this situation to be relatively rare in practice

[`goblin`](https://crates.io/crates/goblin/versions) has been using `0.0.x` until recently.

mindriot101:
@carols10cents thank you so much for this feedback, and suggestions on how to continue. This will really help me understand the `crates.io` code base better.

src/controllers/krate/downloads.rs
```diff
@@ -67,3 +67,70 @@ pub fn downloads(req: &mut dyn Request) -> AppResult<Response> {
...
+    struct Meta<'a> {
+        #[serde(rename = "crate")]
+        krate: &'a str,
+        ndays: i32,
+    }
+
+    // Now get the grouped versions for the last `ndays` days.
+    //
+    // XXX I am not sure how to do this in the database yet, with Diesel's API, so for the time
+    // being, perform this aggregation in Rust.
```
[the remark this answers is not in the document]
mindriot101: Thanks for the pointer, I've implemented it with diesel functions in the latest version. The result is much cleaner, and has improved my understanding of diesel some more which I'm in favour of!

src/controllers/krate/downloads.rs
mindriot101: This has been taken care of with the latest changes switching over to using diesel queries.

src/tests/krate.rs
mindriot101: I've updated the tests to try and add clarity to what is going on. The example you gave was close to working, what is the correct api is:

```rust
let _krate = CrateBuilder::new("foo_versions", user.id)
            .version(VersionBuilder::new("0.5.1")).recent_downloads(3)
            .version(VersionBuilder::new("1.0.0")).recent_downloads(2)
            .expect_build(conn);
```

however this only adds downloads to the most recently added downloaded version (i.e. 1.0.0). I had to use diesel to insert rows into the `version_downloads` table manaully, though I tried to add clarity.

> I'm also wondering if we should be testing that this endpoint doesn't include non-recent download stats, so we'd need a way to set older downloads. The crate builder isn't a great example for this as it kind of cheats. VersionDownload also doesn't have a nice function for setting downloads for a particular date because usually we only want to set downloads for today. The populate utility script for setting up test data has an example of doing this, though.

I also added this to the test, so now it sets up two versions within 90 days and two versions older than 90 days. I _also_ added some checking that the versions come back sorted, which should make the tests stable and reliable.

src/controllers/krate/downloads.rs
```diff
@@ -67,3 +67,70 @@ pub fn downloads(req: &mut dyn Request) -> AppResult<Response> {
         meta,
     }))
 }
+
+/// Handles the `GET /crates/:crate_id/recent_downloads` route.
+pub fn recent_downloads(req: &mut dyn Request) -> AppResult<Response> {
+    use diesel::dsl::*;
+
+    let ndays = 90;
```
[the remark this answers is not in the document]
mindriot101: I don't have time to look into this now unfortunately, but yes I think this would be a good idea to extract. I'm happy to take a look.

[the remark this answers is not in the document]
mindriot101: I've now moved this constant into the `Config` struct which seems a reasonable place to put it. We don't want the user to change it, but it is something that can be accessed from all request handlers.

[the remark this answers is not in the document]
mindriot101: I've now moved this constant into the `Config` struct which seems a reasonable place to put it. We don't want the user to change it, but it is something that can be accessed from all request handlers.

mindriot101:
Yup not hooked up yet I'm afraid. I wanted to have a think about @bjorn3's thoughts. Sorry I haven't had time yet

carols10cents:
Oh no problem! Just checking :)

app/components/versions-graph.js
```diff
@@ -0,0 +1,105 @@
+import semver from 'semver';
+import Component from '@ember/component';
+
+// Component to plot downloadeds split by semver minor version.
+export default Component.extend({
+  resizeHandler: undefined,
```
locks: Why is this `undefined`?

app/components/versions-graph.js
```diff
@@ -0,0 +1,105 @@
...
+  didRender() {
+    this._super(...arguments);
+
+    let data = this.data;
+
+    // Early exit if the google plotting libraries do not exist
+    let show = data && window.google && window.googleChartsLoaded;
+    this.element.style.display = show ? '' : 'none';
+    if (!show) {
+      console.error('cannot show graph due to missing google dependencies');
```
locks: In which situations would the dependencies not be present?

app/components/versions-graph.js
```diff
@@ -0,0 +1,105 @@
...
+      let major = semver.major(v.num);
+      let minor = semver.minor(v.num);
+      let downloads = v.downloads;
+
+      // XXX ugly hack to get semver to parse the version correctly later on.
+      // We want to do a semver-aware sort, but the `semver.lt` function only
+      // understands version triples, not doubles.
+      let key = `${major}.${minor}.0`;
+      if (downloadsPerVersion.has(key)) {
+        const old = downloadsPerVersion.get(key);
```
locks:
```suggestion
        let old = downloadsPerVersion.get(key);
```

app/components/versions-graph.js
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    // Build up the plotting data
+    let plotData = [
+      // Headings and the nature of additional parameters
+      ['Version', 'Downloads', { role: 'style' }, { role: 'annotation' }],
+    ];
+
+    // Update plotData with rows in the correct format for google visualization library
+    for (var [key, value] of sortIncreasingBySemver(downloadsPerVersion)) {
```
locks:
```suggestion
    for (let [key, value] of sortIncreasingBySemver(downloadsPerVersion)) {
```

app/components/versions-graph.js
```diff
@@ -0,0 +1,105 @@
...
+    // Draw the plot into the current element
+    let chart = new window.google.visualization.BarChart(this.element);
+    chart.draw(myData, options);
+  },
+});
+
+function sortIncreasingBySemver(downloadsMap) {
+  const items = Array.from(downloadsMap.entries());
+  // Sort by semver comparison
+  items.sort((a, b) => {
```
locks:
```suggestion
  items.sort(([versionA], [versionB]) => {
```

app/components/versions-graph.js
```diff
@@ -0,0 +1,105 @@
...
+    // Index 0 is the version string in the array.
+    //
+    // We use `lt` here as we want the array to be sorted in reverse order
+    // (newest at the top)
+    return semver.lt(a[0], b[0]);
```
locks:
```suggestion
    return semver.lt(versionA, versionB);
```

app/components/versions-graph.js
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  // Update the labels to show e.g. `0.1.x` instead of `0.1.0` which is
+  // required by semver comparisons
+  return items.map(i => {
```
locks:
```suggestion
  return items.map(([version, count]) => {
```

app/components/versions-graph.js
```diff
@@ -0,0 +1,105 @@
...
+    return [i[0].replace(/\.0$/, '.x'), i[1]];
```
locks:
```suggestion
    return [version.replace(/\.0$/, '.x'), count];
```

app/components/versions-graph.js
bjorn3: When javascript is disabled.

mindriot101:
>> Maybe group 1.0.x and 1.1.x together as 1.x.y, because a 1.0.x requirement is satisfied by a 1.1.x version.

> While that is true for the default version specification without any operator (which has the same behavior as using a caret), library authors might be interested in people using tilde requirements.

I agree with this. Given the youth of most Rust crates (only 13% of crate versions are >=1.0.0), I think it's worth separating by minor version. Otherwise most crates will have two bars, one for <1.0.0 and one for 1.0.0+. Also this works with the tilde requirement.

>> Also 0.0.x and 0.0.y are not compatible.

> You are correct-- I thought they were, according to Cargo when using caret specifications, but it turns out there's an exception when the minor version is 0.. Given that Cargo's default version when running cargo new is 0.1.0, I expect this situation to be relatively rare in practice (I can run some queries to be sure). I'm not saying we shouldn't break these out separately for correctness's sake, just that if it turns out to be prohibitively complex or expensive, we should consider the rarity of this case. I'm going to take a look at the current code now and see. Any thoughts on this @mindriot101 ?

So it turns out that ~10% of crate versions on crates.io have versions < 0.1.0 which is not a small number! I would prefer for the time being to continue to split by minor version going forward, and update the logic if this does not give the correct information.

app/components/versions-graph.js
mindriot101: I was using `download-graph.js` as inspiration and that's what happens there. We can initialise `resizeHandler` at this location and it seems to work ok. Which do you prefer?

app/components/versions-graph.js
mindriot101: I presume you mean using NoScript selectively. The entire site does not work without Javascript, so we're not handling that use case. Again this is taking influence from `download-graph.js`, should the graph plotting scripts fail to download then we don't bother making the extra network requests.

carols10cents:
(Again, apologies if I'm looking at this before you're ready, I saw there were more commits and wanted to check in)

I accidentally tried this out on a snapshot of production that I downloaded over 90 days ago, so that no crate's versions had any recent downloads! In this case, the recent_downloads API returns:

```
{"downloads":[],"meta":{"crate":"some_crate","ndays":90}}
```

and on the page, instead of the graph, I see an error message that says `Data column(s) for axis #0 cannot be of type string`. 

Because we have crawlers and mirrors and such, it's probably rare that no version in the past 90 days would have any downloads, but we should probably handle this situation a bit nicer juuuust in case.

Now to go get a more recent snapshot so I can actually test this ;)

carols10cents:
Ahh, I see some `TODO`s in there. Carry on!

mindriot101:
Hi @carols10cents thank you for testing this edge case I had not considered. I am still working on this, but happy for any feedback. I am unlikely to get time this week though, but next week I have some time.

carols10cents:
No problem!! Not trying to rush you, I'm just trying to make sure we don't leave any PRs without reviews for too long :) Take as much time as you need.

mindriot101:
Hi all, I'd love to get some help on how to implement a test for @carols10cents's example where no downloads have been performed within 90 days. In particular [this commented out test](https://github.com/rust-lang/crates.io/pull/1941/files?file-filters%5B%5D=.js#diff-842e3f2cad381ca6f428a69468c2c04dR137). I'm not familiar enough with the javascript testing ecosystem. I think I need to update some of the fixtures in the `mirage` dir. Can anyone give me any pointers?

carols10cents:
> Hi all, I'd love to get some help on how to implement a test for @carols10cents's example where no downloads have been performed within 90 days. In particular [this commented out test](https://github.com/rust-lang/crates.io/pull/1941/files?file-filters%5B%5D=.js#diff-842e3f2cad381ca6f428a69468c2c04dR137). I'm not familiar enough with the javascript testing ecosystem. I think I need to update some of the fixtures in the `mirage` dir. Can anyone give me any pointers?

Hi! What do you think about something like this, in the mirage config?

```
  this.get('/crates/:crate_id/recent_downloads', (schema, request) => {
    let crate = request.params.crate_id;
    
    var downloads = [
      { version: '0.1.0', downloads: 32 }, 
      { version: '0.2.0', downloads: 128 }
    ];  

    if (crate == 'norecentdownloads') {
      downloads = [];
    }
    
    return { downloads: downloads, meta: { crate: crate, ndays: 90 } };
  });

```

app/templates/crate/versions.hbs
```diff
@@ -12,7 +12,7 @@
 
 <div class="graph">
   <h4>Downloads per minor version in the last 90 days</h4>
-  {{versions-graph model=@model}}
+  {{versions-graph model=model}}
```
locks: This should be reverted, along with the linting configuration.

mindriot101: Done

tests/acceptance/crate-test.js
```diff
@@ -117,6 +117,27 @@ module('Acceptance | crate page', function(hooks) {
...
+    this.server.loadFixtures();
+
+    await visit('/crates/nanomsg');
+    await click('[data-test-all-versions-link]');
+
+    assert.dom('.graph').hasText(/Downloads per minor version/);
+    assert.dom('.graph svg g g g g text[text-anchor=end]').exists({ count: 2 });
+  });
+
+  /* TODO
```
locks: Would using [QUnit.todo](https://api.qunitjs.com/QUnit/todo) make sense here?

mindriot101:
Hi all, sorry for the lack of input recently. I'm having trouble creating a good test that exercises the situation when no recent downloads are found. I think this is an edge case, however, I'd prefer a test for it.

The expected behaviour is that it should gracefully show text saying that no recent downloads are available. So far I have handled this case (i.e. it does not throw an unhandled exception) but nothing further. I'd like a hand implementing the test.

The advice in [carols10cents's post](https://github.com/rust-lang/crates.io/pull/1941#issuecomment-570436134) gets me close I think, but other tests fail. I think this is an issue with my lack of javascript and ember testing knowledge. I would really appreciate some help.

The alternative is that we do not have an automated test for this case and rely on manual testing. I can edit my local database to cause this case for a particular crate for example.

mindriot101:
Sorry it's been so long with this, but IIRC I was having trouble with the JS tests. Can anyone help? cc @carols10cents ?

I realise that my PR needs to be rebased, which I can do later.

Thanks!

src/controllers/krate/downloads.rs
```diff
@@ -67,3 +69,53 @@ pub fn downloads(req: &mut dyn Request) -> AppResult<Response> {
         meta,
     }))
 }
+
+/// Handles the `GET /crates/:crate_id/recent_downloads` route.
+pub fn recent_downloads(req: &mut dyn Request) -> AppResult<Response> {
```
Turbo87, line 74: @jtgeibel any thoughts on this? it might conflict with our plans to move the historical download data to S3 🤔

Turbo87:
hi @mindriot101!

first of all, sorry for the long radio silence on this PR. the team has been quite busy with just keeping the service running at all, and there hasn't been much capacity for implementing and reviewing new features.

we did however talk about this PR today in the team meeting. the code generally looks good, but we had a lot of performance issues with the version downloads database table lately and it looks like this feature would increase the load on that table even further. since this is more of a nice-to-have feature and not something that we would consider mission-critical, we will unfortunately have to reject this feature for the reasons above 😞 

and please don't take this personally, we would love more PRs from you in the future, but we unfortunately need to be mindful with our limited server resources right now.

mindriot101:
That's absolutely fine @Turbo87 . Thanks for the update 👍

Issue #1903, opened 2019-11-18
Try out GitHub Actions


carols10cents:
This issue is great for someone looking to learn about how GitHub Actions and crates.io's CI work!

[rust-lang/rust is working on moving to GitHub Actions](https://internals.rust-lang.org/t/evaluating-github-actions/11292); I'd love for us to be on a similar CI setup. I meant to try Azure Pipelines here as well when rust-lang/rust moved, but seeing as they're moving again (even though aiui GitHub Actions is Azure under the hood), I don't see a reason for us to try Azure Pipelines if the eventual goal is consistency.

Pietro says GitHub actions should feel faster than Travis, so let's give it a try!

Implementation instructions:

- [Follow the documentation for GitHub Actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions). I actually don't know too much about it so I can't be much more specific here, but if you have any questions, please ask them here. I can take a look too and we can figure it out together! [These example Rust/Cargo workflows look relevant](https://internals.rust-lang.org/t/evaluating-github-actions/11292/4?u=carols10cents).
- Open a PR with the GitHub Actions configuration and try to get it to have the same result that Travis gets. Ping me if any of the repo's settings need to be changed.

weihanglo:
I'd love to work on this!

smarnach:
@weihanglo Feel free to go ahead! If you have any questions, you can ask here.

weihanglo:
I’ve opened a pull request and got some question to figure out. Please take a look.
https://github.com/rust-lang/crates.io/pull/1942/files

PR #1942, opened 2019-12-07
Transit CI to GitHub Actions

weihanglo:
# What

Transit the CI process to GitHub Actions.

Close https://github.com/rust-lang/crates.io/issues/1903

[Live GitHub Actions of this PR](https://github.com/weihanglo/crates.io/actions?query=workflow%3ACI+)

# Details

- Create two independent jobs, Frontend and Backend, to run in parallel.
- The pre-installed Rust in the environment of Github Actions is installed under root privileges. That makes tarball cache fails to restore. Instead, we install Rust via rustup script manually.
- GitHub Actions has not yet support `allow_failures` option, so we set` fail-fast` to false to prevent  failure on nightly to cancel the entire CI process.
- Update all documents mentioned Travis-CI to GitHub Actions.

# Need helps

- I have no experience of handling Percy token. Need some help to check the correctness.
- Not sure how to configure GitHub with bors correctly. I followed [this](https://github.com/rust-osdev/x86_64/blob/master/bors.toml) and set the status to job names.


.github/workflows/ci.yml
```diff
@@ -51,3 +51,141 @@ jobs:
...
+          mkdir -p /etc/postgresql/11/main
+          sudo tee /etc/postgresql/11/main/pg_hba.conf <<END
+          local   all             all                                     trust
+          host    all             all             127.0.0.1/32            trust
+          END
+          sudo systemctl restart postgresql@11-main
+
+      - name: Setup database
+        run: |
+          which diesel || cargo install diesel_cli --force --vers $(cat .diesel_version) --no-default-features --features postgres
```
carols10cents: I don't think we need `--force` here anymore because we'll only install `diesel_cli` if it's not present? I don't think it's hurting anything either though... [we removed it from the travis config recently](https://github.com/rust-lang/crates.io/commit/f1afb8f95b563f2b47793158ec0d7202096bb03e#diff-354f30a63fb0907d4ad57269548329e3). What do you think?

.travis.yml
```diff
@@ -1,77 +0,0 @@
-language: rust
```
carols10cents: I think we'll want to run our tests on *both* travis and github actions for a little while, to make sure everything is configured correctly. Could you remove this commit please, so that the travis config is still in the repo?

docs/ARCHITECTURE.md
```diff
@@ -87,7 +87,7 @@ for another platform.
 
 These files are mostly only relevant when running crates.io's code in development mode.
 
-* `bors.toml` - Configure our instance of [bors-ng][] for continous integration
+* `bors.toml` - Configure our instance of [bors-ng] for continous integration
```
carols10cents: Oops-- we haven't been using bors-ng for a while, but we forgot to remove the config files. [I've opened a PR to fix that](https://github.com/rust-lang/crates.io/pull/1944) which will probably cause a merge conflict here; you can discard these changes.

Thank you so much for remembering to update the documentation now so that we don't have this problem again in the future!

.github/workflows/ci.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+      - trying
```
carols10cents: I don't think we need this commit at all because, as I mentioned in other comments, we aren't actually using bors-ng.

carols10cents:
@jtgeibel I know you've worked on the CI config lately, so I'd love for you to review this PR too!

.github/workflows/ci.yml
```diff
@@ -51,3 +51,141 @@ jobs:
...
+      - name: Cache rustup
+        uses: actions/cache@v1
+        with:
+          path: ~/.rustup
+          key: ${{ runner.os }}-rustup-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-key: |
+              ${{ runner.os }}-rustup-${{ matrix.rust }}-
+              ${{ runner.os }}-rustup-
+
+      - name: Cache cargo binaries
```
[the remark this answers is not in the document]
weihanglo: The implement was just simply followed [the guide in cargo book](https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci). Not sure if caching these gets optimized, but I am more concerned about the [cache limits GitHub provided](https://github.com/actions/cache#cache-limits) (only 400MB per cache).

.github/workflows/ci.yml
weihanglo: Removed :)

.travis.yml
weihanglo: Done. I'll keep it.

weihanglo:
Percy builds in GitHub Actions has already integrated with the token. Seems no changes need. And [this build](https://percy.io/crates-io/crates.io/builds/3414605) is from the last commit b188188. 

However it [failed here](https://github.com/weihanglo/crates.io/pull/1/checks?check_run_id=340504037#step:7:215) and I cannot figure out why. It should be exact the [same build](https://percy.io/crates-io/crates.io/builds/3414614) from [travis-ci](https://travis-ci.com/rust-lang/crates.io/jobs/265288044#L578). I've also test upgrade Chrome in CI environment but the [build still got an error](https://github.com/weihanglo/crates.io/pull/2/checks?check_run_id=340576160#step:8:215) and [differed](https://percy.io/crates-io/crates.io/builds/3415214)

carols10cents:
Hm what is the binary file named 3a8c2f23 that was added?

carols10cents:
> Percy builds in GitHub Actions has already integrated with the token. Seems no changes need. And [this build](https://percy.io/crates-io/crates.io/builds/3414605) is from the last commit [b188188](https://github.com/rust-lang/crates.io/commit/b1881886f435dfe1a1a2775b16ac1b02a7c4a4ef).
> 
> However it [failed here](https://github.com/weihanglo/crates.io/pull/1/checks?check_run_id=340504037#step:7:215) and I cannot figure out why. It should be exact the [same build](https://percy.io/crates-io/crates.io/builds/3414614) from [travis-ci](https://travis-ci.com/rust-lang/crates.io/jobs/265288044#L578). I've also test upgrade Chrome in CI environment but the [build still got an error](https://github.com/weihanglo/crates.io/pull/2/checks?check_run_id=340576160#step:8:215) and [differed](https://percy.io/crates-io/crates.io/builds/3415214)

Strange! In the GitHub Actions logs you linked to where the percy API call failed, if [I scroll back up a bit I see](https://github.com/weihanglo/crates.io/pull/2/checks?check_run_id=340576160#step:8:201):

```
[percy] Finalizing build...
[percy] Visual diffs are now processing: https://percy.io/crates-io/crates.io/builds/3415214
```

When the frontend tests run in Travis, I also see [logs like that in the middle of the test run](https://travis-ci.com/rust-lang/crates.io/jobs/265288044#L563-L565), but I also see them [again near the end](https://travis-ci.com/rust-lang/crates.io/jobs/265288044#L591-L592), so twice total, which also happens to be [the number of parallel splits we're running the tests with](https://github.com/rust-lang/crates.io/blob/master/package.json#L28).

[Percy's documentation mentions that the percy client has automatic support for parallel test runners on Travis CI](https://docs.percy.io/docs/parallel-test-suites). 

To test this theory, could you try removing the `--split=2 --parallel` from the test command and see if GitHub actions is able to send all the screenshots to percy?

If that works, another thing to try would be to set [the percy environment variables](https://docs.percy.io/docs/parallel-test-suites#section-manual-configuration-with-environment-variables) for the test job-- like `PERCY_PARALLEL_NONCE=`date +%s` PERCY_PARALLEL_TOTAL=2 ember exam --split=2 --parallel` or similar.

Could you please try those two things? If neither work, I'll do some more investigating. 😅  Let me know!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,191 @@
...
+              rm -v "$HOME/.rustup"
+              rm -v "$HOME/.cargo"
+          fi
+          echo "::add-path::$HOME/.cargo/bin"
+
+      - name: Cache rustup
+        uses: actions/cache@v1
+        with:
+          path: ~/.rustup
+          key: ${{ runner.os }}-rustup-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
```
jtgeibel: I'm not sure if it is necessary to include the `-${{ hashFiles('**/Cargo.lock') }}` portion for this directory and for the `~/.cargo/` paths below.  (It definitely makes sense for the `target/` directory.)  My reasoning is that the contents of these directories shouldn't really be affected by the contents of the lock file.

By the same reasoning, `-${{ matrix.rust }}` may also not be necessary for many of these.

If you're unsure about removing it though, I definitely don't see this as a blocker.  This is probably a micro-optimization given the expected size of these directories.

jtgeibel:
For some reason, the beta branch (and only beta) has failed twice on the same test now.

```
 thread 'krate::new_krate' panicked at 'Could not run jobs: NoMessageReceived', src/libcore/result.rs:1165:5
```

The test isn't failing on Travis, so that's a bit strange. :confused:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,191 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
pietroalbini: `try` as well then.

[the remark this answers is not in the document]
weihanglo: Would fix.
Just curious where are documentations of the infra.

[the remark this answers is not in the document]
pietroalbini: Well, there isn't much documentation at the moment, even though we're working on improving that.

weihanglo:
> To test this theory, could you try removing the `--split=2 --parallel` from the test command and see if GitHub actions is able to send all the screenshots to percy?
> 
> If that works, another thing to try would be to set [the percy environment variables](https://docs.percy.io/docs/parallel-test-suites#section-manual-configuration-with-environment-variables) for the test job-- like `PERCY_PARALLEL_NONCE=`date +%s` PERCY_PARALLEL_TOTAL=2 ember exam --split=2 --parallel` or similar.
> 
> Could you please try those two things? If neither work, I'll do some more investigating. 😅 Let me know!
@carols10cents 

I've tried but still failed on parallel one.

- Without any parallel tests, it passed breezily. [Percy build](https://percy.io/crates-io/crates.io/builds/3461997)
- With a `PERCY_PARALLEL_NONCE`, it seems that CI uploaded correct snapshots but still failed to fetch the image on `/crates/:crate`. [Percy build](https://percy.io/crates-io/crates.io/builds/3462405)

Maybe there are network issues on GitHub Actions?

.github/workflows/ci.yml
weihanglo: Thanks. We need figure out when to invalidate old cache. Maybe some trick like [script//ci/cargo-clean-on-new-rustc-version.sh](https://github.com/rust-lang/crates.io/blob/710ad1edc03c5ac3542c16821232ef6b365f098a/script/ci/cargo-clean-on-new-rustc-version.sh), but I am not sure how to implement that.

weihanglo:
@pietroalbini  

> There is a binary file named `3a8c2f23`.

Removed. Thanks!

> On Crater and (soon) on rustc the approach we've taken is to add a "success" and "failure" jobs with the same display name: both of them have all the other jobs as `needs:`, and one is `if: success()` while the other is `if: !success()`. That allows to gate on the `bors build finished` check.
> 
> https://github.com/rust-lang/crater/blob/a1930d2acd71382b66ce68d9e45ce57f350fae1a/.github/workflows/bors.yml#L125-L149

Current Github Actions does not support [`allow_failures`](https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail) config on job/matrix. Though we can let it fail silently but it  is not the right way to handle it in my opinions.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,196 @@
...
+      - auto
+      - try
+    paths-ignore:
+      - 'docs/**'
+      - README.md
```
pietroalbini: So, we need to run CI all the time when something is pushed to `auto` and `try`, even if that only changes the documentation. If we don't do that, we won't be able to merge those PRs with bors.

Also, now that I think about it, we don't need to run CI on the `master` branch, `auto` and `try` are enough.

carols10cents:
>     * With a `PERCY_PARALLEL_NONCE`, it seems that CI uploaded correct snapshots but still failed to fetch the image on `/crates/:crate`. [Percy build](https://percy.io/crates-io/crates.io/builds/3462405)

Yeah, I'm seeing the download graph not load in Percy [on a different PR](https://percy.io/crates-io/crates.io/builds/3454116?utm_campaign=crates-io&utm_content=crates.io&utm_source=github_status_public), so I think that's unrelated to GitHub Actions. I'm going to investigate that a bit today.

.github/workflows/ci.yml
weihanglo: Jobs frontend and backend are added. Also all `path-ignore` and `master` branch restrctions are now removed :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,216 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - auto
+      - try
```
jtgeibel: On travis we need to include the `master` branch, because it's (well technically the repo `HEAD`) cache is used as the fallback for new branches and PRs.  That doesn't seem to apply to the caching logic on GH Actions, so I think its okay to drop it here.

PR #1944, opened 2019-12-09
Remove bors-ng config; we now use rust-lang's main bors

carols10cents:
We've been using rust-lang's bors since
[2019-01](https://github.com/rust-lang/rust-central-station/pull/124)
and just never removed the references to bors-ng, oops


smarnach:
LGTM.

@bors r+

PR #1945, opened 2019-12-09
Query GitHub teams directly by name

iliana:
I was having trouble adding the `github:awslabs:tough` team to [one of our new crates](https://crates.io/crates/tough) because the awslabs org has over 100 teams.

This allows users to use the 101st-created team on their GitHub organization for staggeringly large organizations by querying the team directly.

Test data was updated by hand per guidance in #crates-io, based on similar queries I made manually, but I would greatly appreciate if someone with access to actually make these API calls tested things work. :)

### DIFF
--- a/src/models/team.rs
+++ b/src/models/team.rs
@@ -135,26 +135,18 @@ impl Team {
 
         #[derive(Deserialize)]
         struct GithubTeam {
-            slug: String,         // the name we want to find
             id: i32,              // unique GH id (needed for membership queries)
             name: Option<String>, // Pretty name
         }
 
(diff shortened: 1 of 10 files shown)


carols10cents:
I've deployed this; please try adding the team again and let me know if it doesn't work :) Thank you for the fix!!

iliana:
```
$ cargo owner --add github:awslabs:tough olpc-cjson
    Updating crates.io index
       Owner team github:awslabs:tough has been added as an owner of crate olpc-cjson
```

Perfect!

PR #1948, opened 2019-12-12
Try to figure out some percy problems

carols10cents:
I'm fixing some issues with our tests and trying to figure out some percy weirdness in the process.

### DIFF
--- a/tests/acceptance/crate-test.js
+++ b/tests/acceptance/crate-test.js
@@ -12,17 +12,6 @@ module('Acceptance | crate page', function(hooks) {
   setupApplicationTest(hooks);
   setupMirage(hooks);
 
-  test('is accessible', async function(assert) {
-    assert.expect(0);
-
-    this.server.create('crate', 'withVersion', { id: 'nanomsg' });
-
-    await visit('/');
-    percySnapshot(assert);
-
-    await a11yAudit(axeConfig);
-  });
-
   test('/crates/:crate is accessible', async function(assert) {
     assert.expect(0);
 


carols10cents:
> The search test wasn't actually going to the search page 😰

and the search page actually has some accessibility problems 🤦‍♀

carols10cents:
I really don't understand what the problem could be.

The chart for `/crates/:crate` *always* fails to load on travis now (and github actions on that PR), but *never* fails to load locally for me, either while running the tests locally or visiting the page.

The chart on `/crates/:crate/:version` always *successfully* loads on travis, so it's not a "google charts js isn't accessible on travis anymore" problem.

carols10cents:
I found these docs on how to just hide an element from percy: https://docs.percy.io/docs/percy-specific-css#section-percy-specific-css

Also learned [here](https://docs.percy.io/docs/percy-platform-basics#section-snapshot-rendering-and-asset-discovery) that percy takes DOM snapshots, not screenshots, and then renders the page on its own infrastructure. That page also says:

> We skip assets that are hosted on a different domain from what the snapshot came from.

so I'm not sure how the Google charts ever worked, or work on some pages but not others!!

carols10cents:
Ok, the download graph was removed as I expected for `Acceptance | crate page | /crates/:crate is accessible`

But I don't understand why it still shows up for `Acceptance | crate page | /crates/:crate/:version is accessible` and percy reports "no changes" 😭

carols10cents:
I'm going to close this and redo it once we figure out the direction we want to go with percy.

PR #1950, opened 2019-12-13
Add support for badges for GitHub workflows

scattenlaeufer:
This PR adds support for GitHub workflow badges to be defined in `Cargo.toml` in the same style as other badges are.

Once this PR is merged, I will also create a PR to [rust-lang/cargo](https://github.com/rust-lang/cargo) to add this badge to the manifest documentation in the Cargo Book.

### DIFF
--- a/app/components/badge-github.js
+++ b/app/components/badge-github.js
@@ -1,0 +1,15 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { alias } from '@ember/object/computed';
+
+export default Component.extend({
+  tagName: 'span',
+  classNames: ['badge'],
+  repository: alias('badge.attributes.repository'),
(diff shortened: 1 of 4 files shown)


charmander:
#1838

smarnach:
@scattenlaeufer As already commented, this seems to duplicate #1838. Did you see that other PR? Is the scope really the same? (I haven't looked at the code at all so far.)

scattenlaeufer:
No, I didn't see #1838 and the only major difference between our PR is that I made the workflow name optional and let it default to `Rust`.

So this PR should probably be closed since it's a duplicate.

carols10cents:
Even though #1838 still needs some changes, I think that one is closer to what we want eventually than this one is, so I am going to close this PR as a duplicate of that one.

Thank you for your work, though!!

PR #1951, opened 2019-12-14
Update `ember-template-lint` to v1.12.1

Turbo87:
see https://github.com/ember-template-lint/ember-template-lint/blob/v1.12.1/CHANGELOG.md ✨


.template-lintrc.js
```diff
@@ -7,7 +7,9 @@ module.exports = {
 
   rules: {
     'img-alt-attributes': false,
-    'triple-curlies': false,
-    'html-comments': false,
+    'no-html-comments': false,
+    'no-triple-curlies': false,
```
locks: We don't have triples anymore, I think we can remove this.
```suggestion
```

Issue #19, opened 2014-10-28
Make API key more easily available after creating a new account


Gankra:
Not very discoverable, especially since this is probably the first thing you're interested in getting after registering. Resetting it should definitely be in there, but that's certainly not where I would expect to go for "what is my API key".

Possible solution: redirect new logins and/or registrations to there.

carols10cents:
What happens today when you're a new user of crates.io:

You visit crates.io and see this:

<img width="1016" alt="crates.io homepage" src="https://user-images.githubusercontent.com/193874/68893893-1ad25400-06f4-11ea-96d8-6c376b4952a7.png">

So you click "Log in with GitHub", do the oAuth dance, and you're taken... back to the homepage, only now your icon is in the upper right:

<img width="1003" alt="crates.io homepage but logged in" src="https://user-images.githubusercontent.com/193874/68894160-b368d400-06f4-11ea-81ba-3a04de5600c1.png">

There's not really a call to action, you have to notice there's a little triangle next to your icon that opens this menu:

<img width="1006" alt="user menu open" src="https://user-images.githubusercontent.com/193874/68894057-769cdd00-06f4-11ea-9f54-a527083c46d0.png">

and the menu item you want is "account settings", which is the crates.io/me page where you can verify your email and create API tokens:

<img width="1000" alt="account settings page" src="https://user-images.githubusercontent.com/193874/68894120-9f24d700-06f4-11ea-8c25-fc71f53aeeb4.png">

The smallest change is probably adding a notice at the top of the page that says something like "Welcome to crates.io! Visit [account settings](https://crates.io/me) to verify your email address and create an API token!". 

Ideally, we'd only show that if we just created an account rather than updating one in the [`NewUser` `create_or_update` method](https://github.com/rust-lang/crates.io/blob/21e04f3028624485621d577359008560728a37d3/src/models/user.rs#L55). I *think* we can send that info to the frontend [here](https://github.com/rust-lang/crates.io/blob/21e04f3028624485621d577359008560728a37d3/app/routes/github-authorize.js#L22-L23) from [the backend controller action associated with `/authorize`](https://github.com/rust-lang/crates.io/blob/21e04f3028624485621d577359008560728a37d3/src/controllers/user/session.rs#L46-L105).


Other changes I could see that seem more complex:

- Always redirecting to /me after logging in (not sure if that's the best experience either)
- Customizing the message shown based on whether the user needs to verify their email, create an API token, or both (account completion percentage sort of?)
[...]

PR #1952, opened 2019-12-14
Add flash message with info after login

Mordorreal:
- Add new types for flash message
- Show info type flash message after login via github
- Add missed tests

Small initial step to address #19


app/routes/login.js
```diff
@@ -67,6 +68,12 @@ export default Route.extend({
       let user = this.store.push(this.store.normalize('user', data.user));
       let transition = this.get('session.savedTransition');
       this.session.loginUser(user);
+
+      const messageAfterLogin = htmlSafe(
+        "Welcome to crates.io! Visit <a href='me'>account settings</a> to verify your email address and create an API token!",
+      );
+      this.flashMessages.show(messageAfterLogin, { type: 'info' });
```
Turbo87: do you want to show this message to **all** users? even if they have verified their data already? 🤔

Mordorreal: For first version yes. Currently trying  to understand how I can pass data from backend and what parameters it will be. I think about `{ "email_verified": false, "api_key_verified": true }` but backed currently is a real mess for me. No good structure and abstractions after rails. Still trying to learn how to handle that.

carols10cents: Hi @mordorreal! Thank you for this pull request!

I'm sorry our backend is such a mess; there isn't really a Rails for Rust yet and crates.io was written even before any of the experimental web frameworks existed!

There is a [router.rs file](https://github.com/rust-lang/crates.io/blob/4b1aa086e702bfd741773b5265cf149fe2c6a6b7/src/router.rs) that is similar (in purpose at least) to Rails' routes.rb. There are two backend routes that will be relevant for this PR:

- [`/api/private/session/authorize`](https://github.com/rust-lang/crates.io/blob/4b1aa086e702bfd741773b5265cf149fe2c6a6b7/src/router.rs#L118) is how the frontend gets information about the current user from the backend immediately after logging in with GitHub. It [calls](https://github.com/rust-lang/crates.io/blob/4b1aa086e702bfd741773b5265cf149fe2c6a6b7/src/router.rs#L119) the [`user::session::authorize` controller function](https://github.com/rust-lang/crates.io/blob/f82b3926e4fbe787228d462b9aef0a8b734c9494/src/controllers/user/session.rs#L46-L105). After interacting with info from GitHub, it [calls](https://github.com/rust-lang/crates.io/blob/f82b3926e4fbe787228d462b9aef0a8b734c9494/src/controllers/user/session.rs#L104) the `user::me::me` controller function (discussed more next)
- [`/api/v1/me`](https://github.com/rust-lang/crates.io/blob/4b1aa086e702bfd741773b5265cf149fe2c6a6b7/src/router.rs#L79) is how the frontend gets information about the current user from the backend if there's already an existing session for a logged in user. It also calls the [`user::me::me` controller function](https://github.com/rust-lang/crates.io/blob/4b1aa086e702bfd741773b5265cf149fe2c6a6b7/src/controllers/user/me.rs#L14-L61), so this URL returns the same data as the session authorize URL.

The data returned currently is defined by the "view" struct [`EncodableMe`](https://github.com/rust-lang/crates.io/blob/f82b3926e4fbe787228d462b9aef0a8b734c9494/src/views.rs#L159-L163). Because the backend is only a JSON API, there are no HTML views, so all the views are structs that get serialized to JSON with serde.

`EncodableMe` contains an [`EncodablePrivateUser`](https://github.com/rust-lang/crates.io/blob/f82b3926e4fbe787228d462b9aef0a8b734c9494/src/views.rs#L165-L178), which has an [`email_verified` field](https://github.com/rust-lang/crates.io/blob/f82b3926e4fbe787228d462b9aef0a8b734c9494/src/views.rs#L172), so you should be able to use that!

We can add a field to `EncodablePrivateUser` called `has_tokens` or similar that is `true` if the user has at least one token and `false` if they have none. We'd need to add a query to `user::me::me` that gets that information. The query should use `ApiToken::belonging_to` sort of like [this code does](https://github.com/rust-lang/crates.io/blob/4b1aa086e702bfd741773b5265cf149fe2c6a6b7/src/controllers/token.rs#L13-L14) and diesel's `exists` function sort of like [this code does](https://github.com/rust-lang/crates.io/blob/4b1aa086e702bfd741773b5265cf149fe2c6a6b7/src/models/krate.rs#L163-L169). I'm actually not sure off the top of my head exactly how to combine those two! [Diesel's docs](http://docs.diesel.rs/) is where I'd look to figure that out. Want to give that a try? 

Let me know if you have more questions, I hope that helps to pick the relevant pieces out of our mess at least!

Mordorreal: Hi @carols10cents! Thank you for very detailed answer! You're awesome! I'll try to continue on the issue in this PR.

Mordorreal:
@carols10cents Update with all fixes and improvements. Feedback is welcome!

Mordorreal:
Can someone with permissions also restart CI? Looks like there some problems with testing environment.

app/routes/me/index.js
```diff
@@ -5,12 +5,10 @@ import AuthenticatedRoute from '../../mixins/authenticated-route';
...
-        .setProperties({
-          emailNotificationsSuccess: false,
-          emailNotificationsError: false,
-        })
-        .clear();
+      this.controller.setProperties({
+        emailNotificationsSuccess: false,
+        emailNotificationsError: false,
+      });
     },
```
Turbo87: why did you change the formatting here?

app/templates/catch-all.hbs
```diff
@@ -1,4 +1,4 @@
-<p id='flash' class='shown'>Oops, that route doesn't exist!</p>
+<p id='flash' class='show'>Oops, that route doesn't exist!</p>
```
Turbo87: why did you change the class name?

tests/integration/components/welcome-message-test.js
```diff
@@ -0,0 +1,73 @@
...
+  test('it show reminder about email only if user has tokens', async function(assert) {
+    assert.expect(2);
+    const user = this.server.create('user', 'withTokens');
+
+    this.session = this.owner.lookup('service:session');
+    this.session.loginUser(user);
+
+    await render(hbs`{{welcome-message}}`);
+
+    assert.equal(
```
Turbo87: these checks should use qunit-dom assertions

app/services/flash-messages.js
```diff
@@ -4,8 +4,9 @@ export default Service.extend({
   message: null,
   _nextMessage: null,
 
-  show(message) {
+  show(message, options = { type: 'warning' }) {
     this.set('message', message);
+    this.set('options', options);
   },
```
Turbo87: are these changes still necessary? it looks like we're not using flash messages anymore

app/routes/login.js
```diff
@@ -68,6 +68,7 @@ export default Route.extend({
       let user = this.store.push(this.store.normalize('user', data.user));
       let transition = this.get('session.savedTransition');
       this.session.loginUser(user);
+
```
Turbo87: ?

app/components/welcome-message.js
```diff
@@ -0,0 +1,26 @@
...
+
+    const textArray = [
+      !user.email_verified && 'verify your email address',
+      !user.email_verified && !user.has_tokens && ' and ',
+      !user.has_tokens && 'create an API token',
+      '!',
+    ].filter(e => !!e);
+
+    return textArray.join('');
+  }),
```
Turbo87: I think I would prefer it if we moved that code into the index controller and not instantiate the component at all if we don't have anything to display

app/components/welcome-message.js
```diff
@@ -0,0 +1,26 @@
...
+  text: computed('session.currentUser.{email_verified,has_tokens}', function() {
+    const user = this.get('session.currentUser');
+    if (!user || (user.email_verified && user.has_tokens)) return '';
+
+    const textArray = [
+      !user.email_verified && 'verify your email address',
+      !user.email_verified && !user.has_tokens && ' and ',
+      !user.has_tokens && 'create an API token',
+      '!',
+    ].filter(e => !!e);
```
Turbo87: the code logic here seems a bit hard to read. do you think it could be simplified in some way?

app/routes/me/index.js
Mordorreal: There was an error in console. Please, see commit comment.

app/templates/catch-all.hbs
Mordorreal: For me, the noun is made more sense here, because this is what this css class doing. But English is not my native language and maybe I'm wrong.

tests/integration/components/welcome-message-test.js
Mordorreal: ok. thanks. Will fix.

app/services/flash-messages.js
Mordorreal: If the project has no plan to bring better user exp with right info or success messages then I can remove it.

app/routes/login.js
Mordorreal: Nice catch. will remove.

app/components/welcome-message.js
Mordorreal: This logic has nothing in common with the index controller. This is common arch mistake in all MVC frameworks. Trying to move business logic inside of controllers fail by design.

app/components/welcome-message.js
Mordorreal: I tried some variants before come to this. For example, a variant with ternary operators looks more ugly:
```
const verifyText = user.email_verified ? '' : 'verify your email address ';
const andText = user.email_verified && user.has_tokens ? '' : 'and ';
const tokenText = user.has_tokens ? '' : 'create an API token';

return `${verifyText}${andText}{$tokenText}`;
```
Moving this calculation logic to template is not right.
I'm open to any ideas on how to improve this code.

tests/integration/components/welcome-message-test.js
Mordorreal: Done.

app/routes/login.js
Mordorreal: Done

Mordorreal:
@carols10cents Sorry for disturbing but can you please review the pr?

Mordorreal:
@carols10cents, @Turbo87 Hi. Can you please review? I don't know why label is saying that pr waiting for the author because it's not true. PR is waiting for review.

locks:
@Mordorreal you have multiple conflicts at the moment. Can you rebase and resolve them?

Mordorreal:
@locks Ohhhh. Rebased. That was hard. Can you please review pr? Not sure that I will have willing to rebase this pr again after one month of silence.

Mordorreal:
@Turbo87 Are this PR still needed? If no, please close. If yes, I have time for it now.

Turbo87:
closing due to https://github.com/rust-lang/crates.io/issues/19#issuecomment-1914335696 :)

PR #1953, opened 2019-12-15
Upgrade `jemallocator`

jtgeibel:
Background threads are now enabled for all artifacts by enabling the
`background_threads` feature.

Reviewing `jemalloc-sys` shows the underlying `jemalloc` version remains
unchanged at 5.1, so this is not expected to affect runtime behavior in
production.

Refs: #1265

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,8 +194,7 @@ dependencies = [
  "hyper 0.12.25 (registry+https://github.com/rust-lang/crates.io-index)",
  "hyper-tls 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "indexmap 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "jemalloc-ctl 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "jemallocator 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


smarnach:
> Background threads are now enabled for all artifacts by enabling the
> `background_threads` feature.

This looks like a questionable design choice, since you basically don't have any way to make sure _no_ background threads are used – features are additve and could be enabled by any transitive dependency. It may make sense in this particular case, and it's definitely not a problem for us, since making sure that a feature i enabled is easy.

@bors r+

carols10cents:
Errrr when I try to run the server locally with this PR on macOS, I get:

```
$ cargo run --bin server
    Finished dev [unoptimized + debuginfo] target(s) in 0.51s
     Running `target/debug/server`
<jemalloc>: option background_thread currently supports pthread only
memory allocation of 40 bytes failedAbort trap: 6
```

Looking into this now to see if this is a personal problem...

ehuss:
`background_threads` is not supported on macOS (https://github.com/jemalloc/jemalloc/issues/1433).  Perhaps jemalloc could ignore the feature on that platform?   Or make it an non-default feature in the crates.io code and turn it on for deployment?

jtgeibel:
Thanks @ehuss, that's alright.  The biggest thing that surprised me is that jemalloc appears to behave very differently at runtime (ignoring the option on macOS) than with the same option compiled into the binary (always an error on macOS).

PR #1955, opened 2019-12-16
1360 crate ownership invitation token

nkanderson:
Completes #1360


src/controllers/crate_owner_invitation.rs
```diff
@@ -41,23 +41,46 @@ pub fn handle_invite(req: &mut dyn Request) -> AppResult<Response> {
...
+/// Handles the `PUT /me/crate_owner_invitations/accept/:token` route.
+pub fn handle_invite_with_token(req: &mut dyn Request) -> AppResult<Response> {
+    let conn = req.db_conn()?;
+    let req_token = &req.params()["token"];
+
+    let crate_owner_invite: CrateOwnerInvitation = crate_owner_invitations::table
+        .filter(crate_owner_invitations::token.eq(req_token))
+        .first::<CrateOwnerInvitation>(&*conn)?;
+
+    // TODO: Return 404 if no crate_owner_invite was found
```
carols10cents: I think the `first`...`?` should return 404 already -- [diesel's `first`](https://docs.diesel.rs/diesel/query_dsl/trait.RunQueryDsl.html#method.first) returns `Err(NotFound)` if no records are returned, and [this code converts `Diesel::NotFound`](https://github.com/rust-lang/crates.io/blob/efeb9d0cef99b8575f97715342794a9518199f69/src/util/errors.rs#L72) to [crates.io's `NotFound`](https://github.com/rust-lang/crates.io/blob/efeb9d0cef99b8575f97715342794a9518199f69/src/util/errors.rs#L232), which [returns 404](https://github.com/rust-lang/crates.io/blob/efeb9d0cef99b8575f97715342794a9518199f69/src/util/errors.rs#L245). That conversion happens as part of the `?` operator.

When I'm on this branch and I visit `http://localhost:4200/accept-invite/foo`, I see the request to `http://localhost:4200/api/v1//me/crate_owner_invitations/accept/foo` get a 404. So I think this TODO is TODONE! :)

migrations/2019-12-13-053745_add_crate_owner_invitation_token/down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE crate_owner_invitations DROP COLUMN token;
+ALTER TABLE crate_owner_invitations DROP COLUMN token_generated_at;
+DROP FUNCTION crate_owner_invitations_set_token_generated_at();
```
carols10cents: Just found another issue-- I needed to roll this migration back locally so that I could switch to a branch that doesn't have these code changes. The migration needs to drop the trigger as well, and then the changes should be undone in the reverse order in which they were done, so drop the trigger, then the function, then drop the columns.

app/routes/accept-invite.js
```diff
@@ -0,0 +1,28 @@
...
+  // session: service(),
+
+  async model(params) {
+    try {
+      await ajax(`/api/v1//me/crate_owner_invitations/accept/${params.token}`, { method: 'PUT', body: '{}' });
+
+      // TODO: determine if user model reload is necessary here
+      // The crate invitation acceptance draws heavily on prior art from the email confirmation, which
+      // included the following block as a way to reload the user model after updating. It doesn't seem
+      // to be necessary for this route, but it's unclear how best to verify that.
```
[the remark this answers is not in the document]
nkanderson: Ok, just tested the flow you outlined, and it seems to be good!

src/controllers/crate_owner_invitation.rs
nkanderson: 😁 🎉

migrations/2019-12-13-053745_add_crate_owner_invitation_token/down.sql
nkanderson: Yikes, good catch! I made the changes you suggested and ran the following to test:
```
diesel migration run
diesel migration redo
```
Both seemed to complete successfully.

I also ran `diesel migration revert` and manually verified that the expected columns were not present in the `crate_ownership_invitations` table, before re-running the migration. Is there anything else you'd recommend for checking the migrations?

nkanderson:
Whew, thanks for your patience @carols10cents! I've been a bit busy with the holidays and things the last couple of weeks.

I think you've answered the questions I had, so I made edits as needed and rebased onto master. Looks like CI is passing, so I'm going to mark this as ready for review.

carols10cents:
Thank *you* for *your* patience! I saw one tiny spot where a URL had 2 slashes and went ahead and fixed that since it was the only thing I saw. This looks great to me! Thank you!!

@bors r+

PR #1959, opened 2019-12-17
Use `ember-auto-import` to import npm modules

Turbo87:
... instead of using the complicated `app.import()` API


Turbo87:
npm... 😩

PR #1960, opened 2019-12-17
dependency-lint: Disable `generateTests` option

Turbo87:
We already run `npm run lint:deps` as part of CI, there is no need to also generate tests for this and slow down the build doing so.


locks:
Sounds good!

@bors r+

PR #1971, opened 2019-12-17
Add dependabot configuration file

Turbo87:
see https://dependabot.com/docs/config-file/

as discussed via Discord, I generally prefer the configuration files as it is more explicit and transparent than configuration via the dashboard. the other advantage is that not only organization admins can configure it now via PRs. :)

/cc @carols10cents


carols10cents:
Looks great! Thank you!!!

@bors r+

PR #1983, opened 2019-12-18
Migrate `faker` imports

Turbo87:
`ember-cli-mirage` v1.0.0 removed its `faker` dependency in favor of importing via `ember-auto-import`. This PR adds an explicit dependency on `faker`, and migrates all the existing import to import from the `'faker'` module directly.


locks:
Looks good!

@bors r+

PR #1991, opened 2019-12-18
Limit the branches that travis builds on

carols10cents:
Because branches that become PRs are built twice :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,12 @@ language: rust
 sudo: required
 dist: xenial
 
+branches:
+  only:
+  - auto
+  - try
+  - master
+
 cache:
   cargo: true
   directories:


.travis.yml
```diff
@@ -2,6 +2,12 @@ language: rust
 sudo: required
 dist: xenial
 
+branches:
+  only:
+  - auto
+  - try
```
smarnach: What is this branch? We don't currently seem to have a branch with this name.

pietroalbini: It's used when you do `@bors try`.

jtgeibel:
I saw it mentioned somewhere that we don't need to run tests on `master`, so I wanted to mention that we do want to continue building on master because Travis uses the "default branch" ([docs](https://docs.travis-ci.com/user/caching/#fetching-and-storing-caches)) as the fallback cache for new branches and PRs.  I don't see any UI for this in the Travis settings, and I don't think it makes sense to switch `HEAD` to point to `auto` instead of `master` on GitHub.

@bors r+

smarnach:
Have we reached some quota limit for Travis builds? This didn't even start building on Travis yet, same for #1929.

Turbo87:
@smarnach I'm not sure if it's on the TravisCI side since the PR jobs for Travis seem to run fine. does @bors have some sort of quota limit? 🤔

Turbo87:
hmm, or @bors can run X number of jobs in parallel and the jobs that "timed out" are still running and blocking new jobs from being accepted 🤔

PR #1994, opened 2019-12-19
Link directly to cargo install page

jtgeibel:
This fixes a paper-cut where after clicking on the button, clicking the
back button causes the user to immediately redirect again.

Additionally, on Firefox 71, it was not possible to long-click on the
back button to return to crates.io (as the only entry in the list was
the redirect.  On Chromium 79 I had no such issue.

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -2,10 +2,10 @@
   <h1>The Rust community&rsquo;s crate registry</h1>
 
   <div class='links'>
-    {{#link-to "install" class='yellow-button' data-test-install-cargo-link=true}}
+    <a href="https://doc.rust-lang.org/cargo/getting-started/installation.html" class='yellow-button' data-test-install-cargo-link>
       {{svg-jar "button-download"}}
       Install Cargo
-    {{/link-to}}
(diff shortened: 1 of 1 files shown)


app/templates/index.hbs
```diff
@@ -2,10 +2,10 @@
   <h1>The Rust community&rsquo;s crate registry</h1>
 
   <div class='links'>
-    {{#link-to "install" class='yellow-button' data-test-install-cargo-link=true}}
+    <a href="https://doc.rust-lang.org/cargo/getting-started/installation.html" class='yellow-button' data-test-install-cargo-link='true'>
```
Turbo87:
```suggestion
    <a href="https://doc.rust-lang.org/cargo/getting-started/installation.html" class='yellow-button' data-test-install-cargo-link>
```

jtgeibel:
I've amended the commit to drop the `='true'`.

> do we still use the install route afterwards? was it done that way to not break existing links to https://crates.io/install?

Yes, I left it in case there are external pages linking to that URL.  I don't think there are any remaining internal uses of it.

@bors r=Turbo87

PR #2000, opened 2019-12-19
Configure percy to hide the downloads graph (upgrade to percy 2.0.0 beta)

carols10cents:
Trying just this change by itself


carols10cents:
Don't merge this yet-- I expected this to have 2 differences in the snapshots, but there aren't any. I'm trying to get in touch with Percy support to see if I messed something up or if my expectations are incorrect.

carols10cents:
The percy team mentioned they just released a beta that might fix the problems and that would support the percy yaml config. The travis logs contain lines that say percy finished the build, but the percy page still says it's waiting.... so I've contacted support again with this info. I think we'll probably be better off with #2016 for now.

carols10cents:
Ok, so this seems to work too! I think I still slightly prefer #2016 until @percy/ember is out of beta, interested to see what @jtgeibel or anyone else thinks.

I still have to re-submit the changes that remove the duplicate tests and fix the search tests....

jtgeibel:
I've merged #2016.  I also think it makes sense to pursue this further once it is out of beta.

PR #2016, opened 2019-12-20
Use the percy media query method to hide the downloads graph from percy

carols10cents:
This is a different attempt to fix the percy problems; the percy yaml config doesn't work with the ember-percy version we're currently on.

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -359,3 +359,9 @@ abbr[title] {
     text-decoration: none;
     border-bottom: 1px dotted;
 }
+
+@media only percy {
+    #crate-downloads .graph {
+       display: none;
+    }
+}


carols10cents:
WOO ok this works at least, hides the download graph in both cases!!

It's not super great that this CSS would be in prod, but it's only a tiny bit sooo 🤷‍♀ 

r? @jtgeibel 

wdyt?

smarnach:
There probably is a way to only include the CSS snippet on Travis, but even if we spend the time to figure it out it doesn't buy us much, while at the same time increasing the risk of bugs that only happen in production, so I wouldn't bother.

jtgeibel:
LGTM!

@bors: r+

PR #2026, opened 2019-12-20
DEPRECATION: Passing actions to components as strings is deprecated.

carols10cents:
DEPRECATION: Passing actions to components as strings (like `{{input
enter="search"}}`) is deprecated. Please use closure actions instead (`{{input
enter=(action "search")}}`). ('cargo/templates/catch-all.hbs' @ L6:C2)
[deprecation id: ember-component.send-action] See
https://emberjs.com/deprecations/v3.x#toc_ember-component-send-action for more
details.

### DIFF
--- a/app/templates/catch-all.hbs
+++ b/app/templates/catch-all.hbs
@@ -6,6 +6,6 @@
   {{input type="text" class="search"
           placeholder="Search"
           value=search
-          enter="search"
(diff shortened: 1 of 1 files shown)


Turbo87:
perfect, thanks! 👍 

@bors r+

Issue #2028, opened 2019-12-20
Upgrading ember-source from 3.1.2 to 3.15.0 breaks logging in


carols10cents:
When I try to log in, I get an invalid state parameter error.

See https://github.com/rust-lang/crates.io/pull/2001 and https://github.com/rust-lang/crates.io/pull/2027.

I'm going to be updating this with more details as I figure them out.

---

Ok, I don't know how much of this is related or not, but I'm going to err on the side of over-posting.

---

## fsevents can't install (not likely related)

When I run `npm install` lately (on master as well), I see a bunch of errors and ultimately it says `npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.11`. Seeing as this is an optional dependency, I don't think this is causing the problem, but I haven't been seeing these errors until recently so I decided to list it just in case. 

Might be related to an update to macOS I installed recently, but `brew doctor` isn't complaining about anything. Skimming through fsevents' issues seems like the recommended solution is upgrading to fsevents 2+.

<details>
<summary>Full error logs</summary>

stderr:

```
$ npm install

> fsevents@1.2.11 install /Users/carolnichols/rust/crates.io/node_modules/fsevents
> node-gyp rebuild

No receipt for 'com.apple.pkg.CLTools_Executables' found at '/'.

No receipt for 'com.apple.pkg.DeveloperToolsCLILeo' found at '/'.

No receipt for 'com.apple.pkg.DeveloperToolsCLI' found at '/'.

gyp: No Xcode or CLT version detected!
gyp ERR! configure error
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/Users/carolnichols/.nvm/versions/node/v12.13.1/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:351:16)
gyp ERR! stack     at ChildProcess.emit (events.js:210:5)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:272:12)
gyp ERR! System Darwin 19.2.0
gyp ERR! command "/Users/carolnichols/.nvm/versions/node/v12.13.1/bin/node" "/Users/carolnichols/.nvm/versions/node/v12.13.1/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /Users/carolnichols/rust/crates.io/node_modules/fsevents
[...]

PR #2027, opened 2019-12-20
Revert "Auto merge of #2001 - rust-lang:dependabot/npm_and_yarn/ember…

carols10cents:
…-source-3.15.0, r=Turbo87"

This reverts commit 97ecbf3f1e4c957f25abe0fd7f8decee3c50ddfb, reversing
changes made to 6cfc3bfd61a1e1d1bc4d99bbf724b5196b5257f8.

@Turbo87 I did manual testing of logging in because you mentioned you weren't able to test it; when I was on face1f37 logging in didn't work for me.

I did a git bisect and it pointed to the ember-source upgrade https://github.com/rust-lang/crates.io/pull/2001 which is... all of ember? I'm not really sure how to go about debugging this further, but I'm going to try and I'll open a new issue with more details as I figure them out.

Reverting the ember-source upgrade lets me log in again, so this should make master deployable.

I really really wish we could have a full-stack automated test for logging in with GitHub, but I don't see that happening anytime soon!!

PR #2029, opened 2019-12-20
routes/github-authorize: Fix `queryParams` code

Turbo87:
Accessing `queryParams` on the `transition` object was never officially public API in Ember.js, and has been deprecated in Ember.js v3.6 (see https://deprecations.emberjs.com/v3.x#toc_transition-state). Since it was only considered "intimate" API its removal was done in v3.9 instead of v4.0. Since we skipped both of these versions we unfortunately never saw the corresponding deprecation warning... 😞 

But even if we did upgrade to v3.8 LTS in between, we probably still would have missed it, because the `github-authorize` route is only called in a popup window where the usual DevTools console is not attached, and we unfortunately don't have any test for the login flow (yet!).

Closes https://github.com/rust-lang/crates.io/pull/2027, Resolves https://github.com/rust-lang/crates.io/issues/2028

@carols10cents r?

### DIFF
--- a/app/routes/github-authorize.js
+++ b/app/routes/github-authorize.js
@@ -18,7 +18,7 @@ import { serializeQueryParams } from 'ember-fetch/utils/serialize-query-params';
 export default Route.extend({
   async beforeModel(transition) {
     try {
-      let queryParams = serializeQueryParams(transition.queryParams);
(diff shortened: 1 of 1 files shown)


carols10cents:
Works for me locally!!! Thank you thank you thank you!!

@bors r+

PR #2031, opened 2019-12-21
Enable debug MALLOC_CONF options on CI

jtgeibel:
On top of the production configuration, this commit adds two options:

* `abort:true` - most jemalloc warnings become fatal ([opt.abort])
* `junk:true` - both allocated and deallocated memory will be
  initialized to 0xa5. ([opt.junk])

Combined with the existing `abort_conf:true`, this should be equivalent
to enabling the `debug` feature of `jemallocator`.  We cannot enable
this feature in `Cargo.toml` because that would affect performance in
production.

[opt.abort]: http://jemalloc.net/jemalloc.3.html#opt.abort
[opt.junk]: http://jemalloc.net/jemalloc.3.html#opt.junk

r? @carols10cents

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,7 @@ jobs:
       TEST_DATABASE_URL: postgres://postgres:@localhost/cargo_registry_test
       CARGO_INCREMENTAL: 0
       RUSTFLAGS: "-C debuginfo=0 -D warnings"
+      MALLOC_CONF: "background_thread:true,abort_conf:true"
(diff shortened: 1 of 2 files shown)


jtgeibel:
Hopefully https://github.com/rust-lang/crates.io/pull/2027#issuecomment-568095773 was just a fluke.  If not, this should make things worse I guess, lol.  The tests run fine for me locally.

carols10cents:
LGTM! 🤷‍♀ 

@bors r+

PR #2032, opened 2019-12-21
Add a concrete `util::Error` type (step 3 of refactoring)

jtgeibel:
This PR may be easier to review commit by commit.  Overall this series continues the previous refactoring work in #1920 and #1929.  In particular, a concrete `util::Error` error type is introduced.  The code is updated to follow a style where:

* The new concrete type `util::Error` is great for code that is not part of the request/response lifecycle.  Binaries have been converted to `fn main() -> Result<(), Error>` where applicable.  This avoids pulling in the unnecessary infrastructure to convert errors into a user facing JSON responses (relative to `AppError`).
  * There are now 13 results across 6 files for `, Error>`.  Usage occurs in `src/bin/`, `src/boot/`, and `src/uploaders.rs`.
* `diesel::QueryResult` - There is a lot of code that only deals with query errors.  If only one type of error is possible in a function, using that specific error is preferable to the more general `util::Error`.  This is especially common in model code.
  * There are now 49 results across 16 files for `QueryResult<`.  Most usage is in `src/models/`, with some usage in `src/publish_rate_limit.rs`, `src/tasks/`, and `src/tests/`.
* `util::errors::AppResult` - Some failures should be converted into user facing JSON responses.  This error type is more dynamic and is box allocated.  Low-level errors are typically not converted to user facing errors and most usage is within the models, controllers, and middleware layers (85 of 102 results below).
  * There are now 102 results across 36 files for `AppResult<`.  Most usage is in `src/controllers/` and `src/models/`, with additional usage in `src/middleware/`, `src/{email,github,publish_rate_limit,router,uploaders}.rs`, and `src/{tests,util}/`.

Finally, a `middleware::Result` type is introduced for use in middleware and router code.

The behavior of `dyn AppError` and `chain_error` is a remaining area for improvement.  There may be cases where we want to wrap an `internal` error in a user facing error (like `cargo_err` or `server_err`) and still have the internal error make it up to the logging middleware.  However, I don't see an easy way to propagate this up the middleware stack (without encoding it in an internal response header that we later remove, :sweat:).

r? @smarnach

### DIFF
--- a/src/background_jobs.rs
+++ b/src/background_jobs.rs
@@ -2,7 +2,7 @@ use std::panic::AssertUnwindSafe;
 use std::sync::{Arc, Mutex, MutexGuard, PoisonError};
 
 use diesel::r2d2::PoolError;
-use swirl::errors::PerformError;
+use swirl::PerformError;
 
 use crate::db::{DieselPool, DieselPooledConn};
 use crate::git::Repository;
--- a/src/background_jobs.rs
+++ b/src/background_jobs.rs
@@ -1,21 +1,22 @@
 use std::panic::AssertUnwindSafe;
 use std::sync::{Arc, Mutex, MutexGuard, PoisonError};
-use swirl::PerformError;
+
+use diesel::r2d2::PoolError;
 
 use crate::db::{DieselPool, DieselPooledConn};
 use crate::git::Repository;
 use crate::uploaders::Uploader;
-use crate::util::errors::{AppErrToStdErr, AppResult};
+use crate::util::errors::AppResult;
 
 impl<'a> swirl::db::BorrowedConnection<'a> for DieselPool {
     type Connection = DieselPooledConn<'a>;
 }
 
 impl swirl::db::DieselPool for DieselPool {
-    type Error = AppErrToStdErr;
+    type Error = PoolError;
 
     fn get(&self) -> Result<swirl::db::DieselPooledConn<'_, Self>, Self::Error> {
-        self.get().map_err(AppErrToStdErr)
+        self.get()
     }
 }
 
@@ -56,10 +57,8 @@ impl Environment {
         }
     }
 
-    pub fn connection(&self) -> Result<DieselPooledConn<'_>, PerformError> {
-        self.connection_pool
-            .get()
-            .map_err(|e| AppErrToStdErr(e).into())
+    pub fn connection(&self) -> Result<DieselPooledConn<'_>, PoolError> {
+        self.connection_pool.get()
     }
 
     pub fn lock_index(&self) -> AppResult<MutexGuard<'_, Repository>> {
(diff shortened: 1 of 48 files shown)


smarnach:
I plan to review this on 27 December, but this plan is subject to change…

smarnach:
I'm sorry I did not get to this today, and I won't have time until 2020-01-05.

jtgeibel:
> I'm sorry I did not get to this today, and I won't have time until 2020-01-05.

No problem at all.  When you have a chance, would you start with PR #2071.  It's a small fix and I'd like to get it deployed, just want a 2nd set of eyes on it.  Thanks!

src/bin/enqueue-job.rs
```diff
@@ -1,29 +1,20 @@
...
         "dump_db" => {
             let database_url = args.next().unwrap_or_else(|| env("DATABASE_URL"));
             let target_name = args
                 .next()
                 .unwrap_or_else(|| String::from("db-dump.tar.gz"));
-            tasks::dump_db(database_url, target_name).enqueue(&conn)
+            Ok(tasks::dump_db(database_url, target_name).enqueue(&conn)?)
         }
-        other => Err(cargo_err(&format!("Unrecognized job type `{}`", other))),
+        other => Err(Error::from(format!("Unrecognized job type `{}`", other))),
```
smarnach: You could use `?` instead of `Error::from` here as well, as you did in the other match arms. (I don't have a strong opinion either way.)

I'm glad you could get rid of the helper trait here. :)

src/background_jobs.rs
```diff
@@ -1,21 +1,22 @@
 use std::panic::AssertUnwindSafe;
 use std::sync::{Arc, Mutex, MutexGuard, PoisonError};
-use swirl::PerformError;
+
+use diesel::r2d2::PoolError;
+use swirl::errors::PerformError;
```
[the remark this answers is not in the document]
jtgeibel: It looks like I removed it at one point in the series, and then added it back in, so I probably copied the longer path from an error message.

I added a commit converting 3 places to use the short path (now 6) for consistency.

src/bin/enqueue-job.rs
jtgeibel: It looks like clippy doesn't like `Err(format!("Unrecognized job type `{}`", other))?` (the `clippy::try_err` lint), and the suggestion is to use `.into()`.  I don't have a strong opinion between `Error::from` and `into` here, so I'll leave it as-is.

jtgeibel:
> It would be great if you could introduce some high-level documentation in a doc comment at the beginning of error.rs, explaining the use cases for the main error or result types – basically a summary of what you put in the PR description.

I had been holding off on adding high-level documentation until things settled down, but I think this is now a good time to add these recommendations as documentation.  (Especially since "I'll add some docs in a future PR" always takes longer than expected to get back around to.)  I've rebased the PR, and added a commit adapting part of the PR description into module docs.

Thanks for the review and feedback!

@bors r+

PR #2037, opened 2019-12-21
Replace `<XSelect>` component with regular `<select>` element

Turbo87:
We've only been using `XSelect` in one specific place and v3 of the addon was causing deprecation warnings due to their use of the `this.$()` API.

Since we were not using any of the more advanced features from `XSelect` it was relatively straight-forward to replace with just a regular `<select>` element.

r? @locks


app/helpers/eq.js
```diff
@@ -0,0 +1,7 @@
+import { helper } from '@ember/component/helper';
```
NullVoxPopuli: it's gonna be so good to have this helper as part of the framework :)

PR #1741, opened 2019-05-08
Fingerprinting SVGs

markcatley:
In the production build all the SVGs are copied into dist/assets without fingerprints. All other assets have fingerprints. Since all files in dist/assets are served with far-future cache headers there would be cache issues if the files were ever updated. This is not major issue as these files are not used in normal operations.

Closes #1575

PR #2045, opened 2019-12-22
Fix an icon link that was generating bad URLs

jtgeibel:
On current staging, but not production, navigating directly to a path
such as `/crates/crate-name` results in a request sent to
`/crates/cargo-835dd6a18132048a52ac569f2615b59d.png` (a 404 response
status).

It appears that the `src` field is now treated as a relative path.
Using an absolute URL works for me locally.

r? @Turbo87

### DIFF
--- a/config/manifest.js
+++ b/config/manifest.js
@@ -11,7 +11,7 @@ module.exports = function(/* environment, appConfig */) {
     theme_color: '#f9f7ec',
     icons: [
       {
-        src: 'cargo.png',
+        src: '/cargo.png',
         sizes: '227x227',
         type: 'image/png',
       },


jtgeibel:
I've pushed this to staging and this issue is gone there.

I've also noticed that navigating directly to a crate or version page results in a spinner in place of the follow button.  For example: https://staging.crates.io/crates/egg-man

Things work after navigating within the app.  It seems no request is sent to the backend endpoint when navigating directly.

Turbo87:
I assume this is related to https://github.com/zonkyio/ember-web-app/pull/144, which got merged in https://github.com/rust-lang/crates.io/pull/2002.

I'm wondering though if this is the correct resolution or if the image actually should be fingerprinted. I'll have a look.

Turbo87:
Since https://zonkyio.github.io/ember-web-app/docs/manifest/available-properties#icons is also using absolute paths I'll go ahead and merge this 👍 

@bors r+

jtgeibel:
@Turbo87 from what I've seen, these files do end up hashed in production.  I think we were good as of #1741, unless we've regressed somewhere.  For instance, `cargo.png` is deployed as https://staging.crates.io/cargo-835dd6a18132048a52ac569f2615b59d.png.

Turbo87:
yep, saw that when I was trying out your branch locally :)

PR #2046, opened 2019-12-22
Remove some unnecessary eslint overrides

jtgeibel:
I was experimenting with the eslint setup, and it appears that some lint overrides are no longer needed.

This PR removes all results for `git grep "eslint-disable "` and removes some overrides in `.template-lintrc.js`.  There are 10 remaining needed `eslint-disable-next-line` usages.

r? @Turbo87


Turbo87:
nice, thanks! 

@bors r+

Issue #2030, opened 2019-12-20
Running locally (for testing) requires setting GIT_REPO_URL


bruceadams:
Running the `background-worker` locally fails to find the local index git repo, like this:

```bash
$ cargo run --bin background-worker
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/background-worker`
Booting runner
Using local uploader, crate files will be in the local_uploads directory
Cloning index
thread 'main' panicked at 'Failed to clone index: Error { code: -1, klass: 2, message: "failed to resolve path \'file://./tmp/index-bare\': No such file or directory" }', src/libcore/result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

A workaround, described in https://github.com/rust-lang/crates.io/pull/1996#issuecomment-567920787, is to define GIT_REPO_URL to the correct, absolute path on the local machine, like this:

```bash
$ GIT_REPO_URL=file:///Users/bruce/github/crates.io/tmp/index-bare cargo run --bin background-worker
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/background-worker`
Booting runner
Using local uploader, crate files will be in the local_uploads directory
Cloning index
Runner booted, running jobs
```

Ideally, this would manage to _just work_ using relative paths.

carols10cents:
Closing because #2051 was merged; looks like the auto-close stuff didn't work for some reason?

smarnach:
The PR stated

> This fixes issue #2030

GitHub only looks at the word immediately before the bug number to see whether it's [one of the trigger words](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references), but "issue" isn't one of them.

bruceadams:
Oops! I guess I don't speak Github very well. 🙊

PR #2051, opened 2019-12-22
Use absolute path for GIT_REPO_URL file:// URL

bruceadams:
file:// URLs do not support relative paths. This change uses $PWD to make an absolute path based on where the user is running Crates on their local machine.

This fixes issue #2030

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -6,7 +6,7 @@ export DATABASE_URL=
 # If you are running a mirror of crates.io, uncomment this line.
 # export MIRROR=1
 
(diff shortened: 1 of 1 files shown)


smarnach:
Thanks for the fix! I wasn't sure whether environment variables are expanded during processing of `.env` (it's not executed by the shell after all), but apparently this works fine.

A `file:` URL including the scheme does not support relative paths, but the `git2` crate could reasonably accept a relative path without the scheme (like `"tmp/index-bare"`) and interpret it as a relative file path (like the `git` command-line client does), but it doesn't.

@bors r+

PR #2057, opened 2019-12-23
Fix frontend test issues

carols10cents:
This fixes frontend test problems *uncovered* by Percy, but hopefully no longer *impacted* by Percy problems 🤞 

r? @jtgeibel

### DIFF
--- a/app/templates/search.hbs
+++ b/app/templates/search.hbs
@@ -79,17 +79,17 @@
     {{/each}}
   </div>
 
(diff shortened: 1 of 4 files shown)


carols10cents:
One of the percy values is expected, the other is because it looks like the Azure badge changed the text when you aren't actually using Azure 🤣

jtgeibel:
LGTM!

@bors r+

Issue #2059, opened 2019-12-23
Tracking issue: Gate bors on GH Actions instead of Travis


jtgeibel:
This is a tracking issue for the proposal to gate bors on GH Actions instead of Travis.

Implementation:
- [ ] Change bors configuration

Open issues:
* [x] `panicked at 'Could not run jobs: NoMessageReceived'` - Backtraces enabled in #2058 so we can begin to debug
* Caching behavior isn't always reliable
  * [x] Some builds fail to get the cargo bin cache and have to compile `diesel_cli` in order to setup the database.
  * [x] Unclear if we should be caching compilers at all, although it would be nice to avoid downloading compilers from upstream for every build.  See docs added in #2058.
  * [x] Many builds seem to recompile lots of dependencies, even though the cache of `target/` was restored and `Cargo.lock` hasn't changed. [I think this was resolved in #2532]
* [x] bors needs a single job to gate on.  Currently on Travis, we ignore failures on the nightly channel.  With our current `end-success` job on GH Actions, failures caused by a nightly regression will block us from merging any PRs.
* [x] If a single job fails, GH only offers the ability to restart the entire workflow.  When this happens, the old `end-failure` job remains in the UI with a red X, but it's not clear to me yet if that would cause bors to also fail a retry. **Update:** I haven't seen this in a while, this may be resolved. 
* [x] I've seen an occasional failure on the frontend with `Error: Browser failed to connect within 30s. testem.js not loaded?`  See [this build](https://github.com/rust-lang/crates.io/commit/8b1536854a604c5673da2086a959ceb980457623/checks?check_suite_id=417610695) for an example.  I don't recall ever seeing this occur on Travis. - timeout bumped to 120 seconds in #2158
* [x] Over in #2184 it appears that GH Actions may have issues displaying the correct build status badge, since `master` builds show up under `auto` because the builds are against the same commit hash. - #2059 

Alternatives:

* Do nothing, we continue to gate bors on Travis instead of switching to GH Actions.

Turbo87:
We currently have an issue with the GitHub Action for Rust nightly failing CI. In Travis we have set that job to be allowed to fail, but there does not seem to be a corresponding option in GitHub Actions.

I guess this makes the switch for bors from Travis to GitHub Actions a little more complicated 🤔

jtgeibel:
I've updated the issue with a few other things I've seen and closed out things I think we've resolved.

[...]

PR #2060, opened 2019-12-23
CI: Enable GH Actions on master branch

jtgeibel:
Based on the documented cache [restrictions], pull reqeusts are failing
to access the cache.  One solution is to alway build on master, because
all builds should be able to access caches created by the default
branch.

r? @jtgeibel

[restrictions]: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - auto
+      - master
       - try
 
   pull_request:


jtgeibel:
I'm going to r+ my own PR, because I'd like to land this ASAP.  I don't really see any obvious alternatives, and we can always drop it if we find a better long-term solution.

cc #2059 

@bors r+

PR #2062, opened 2019-12-24
CI: Switch to `actions-rs/toolchain@v1`

jtgeibel:
The caches for `~/.rustup` and `~/.cargo/bin` have been removed.  A
cache for `~/.cargo/bin/diesel` is added.  The cache key is derived from
version specified in `.diesel_version` and there is no `restore-key`
fallback for this step.


jtgeibel:
Closing in favor of #2142

Issue #1928, opened 2019-11-25
Only accept integers for `page` and `per_page` params; return 400 otherwise


carols10cents:
Right now, I see occasional 500 errors in our logs that look like this (some info elided):

```
Nov 25 07:18:41 crates-io app/web.1 at=error method=GET path="/api/v1/crates?page=1&per_page=100%22%EF%BC%8Cexception" service=1ms status=500 error="invalid digit found in string"
```

I have a suspicion that somehow *we're* generating links that contain (decoded) `"，exception`, but I haven't figured out how yet.

Whether we're causing these incorrectly formed URLs or not, the backend should be responding with 400 Bad Request, not 500 because of an invalid digit in the string.

## Implementation instructions

I've [created a branch](https://github.com/integer32llc/crates.io/tree/failing-test) with a commit that adds [some failing tests](https://github.com/integer32llc/crates.io/commit/bdcda54668fb4d71739e26168b0a363764ac1bd6); please include those tests and get them to pass as part of your PR :) (editing the tests might be necessary, and bonus points if you add tests in the pagination controller helper for the happy paths and other error cases!)

The problem is in the pagination controller helper:

- [`page` parsing](https://github.com/rust-lang/crates.io/blob/f93908984bbd02eb8ed79668ce9afa0703b4979d/src/controllers/helpers/pagination.rs#L19)
- [`per_page` parsing](https://github.com/rust-lang/crates.io/blob/f93908984bbd02eb8ed79668ce9afa0703b4979d/src/controllers/helpers/pagination.rs#L47)

To return 400 instead of 500, you'll need to add some `map_err` calls to return [`BadRequest` errors](https://github.com/rust-lang/crates.io/blob/f93908984bbd02eb8ed79668ce9afa0703b4979d/src/util/errors.rs#L264). [Here's an example of someplace we do this](https://github.com/rust-lang/crates.io/blob/f93908984bbd02eb8ed79668ce9afa0703b4979d/src/controllers/token.rs#L57); search for `bad_request` to see other examples.

Let me know if you have any questions or if any of the instructions don't make sense!

andreheringer:
I would like to tackle this issue if if possible

smarnach:
@andreheringer Sure! If you have any questions, you can ask here. Alternatively, you can send a PR of your work in progress.

PR #2065, opened 2019-12-25
Convert pagination ISE into a Bad Request response

jtgeibel:
Errors in parsing the pagination query string parameters can be safely returned to the client and do not need to propagate up to the logging middleware.

The following have been observed in production logs:

* error="cannot parse integer from empty string" for `/api/v1/crates?page=&per_page=50&sort=downloads` (empty `page=`)
* error="invalid digit found in string" for `/api/v1/crates?page=1&per_page=100%22%EF%BC%8Cexception"` (very invalid `per_page`)

Additionally, 2 existing pagination errors are converted to use `bad_request()` in place of `cargo_err()`.  It is not necessary to send a status=200 response to support old versions of cargo for pagination errors.

Fixes:  #1928

r? @smarnach

### DIFF
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -1,5 +1,5 @@
 use crate::models::helpers::with_count::*;
-use crate::util::errors::{cargo_err, AppResult};
+use crate::util::errors::{bad_request, AppResult};
 use diesel::pg::Pg;
 use diesel::prelude::*;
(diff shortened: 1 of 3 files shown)


carols10cents:
This will fix #1928 :)

jtgeibel:
Thanks!  I've updated the description to reference that issue, and cherry-picked the tests provided there.

PR #2066, opened 2019-12-26
Add more logging to background jobs

jtgeibel:
r? @smarnach

### DIFF
--- a/src/bin/enqueue-job.rs
+++ b/src/bin/enqueue-job.rs
@@ -5,7 +5,11 @@ use diesel::PgConnection;
 fn main() -> AppResult<()> {
     let conn = db::connect_now()?;
     let mut args = std::env::args().skip(1);
-    match &*args.next().unwrap_or_default() {
+
+    let job = args.next().unwrap_or_default();
+    println!("Enqueueing background job: {}", job);
+
+    match &*job {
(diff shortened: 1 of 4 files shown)


carols10cents:
I'm a tiny bit concerned about how much this will increase our papertrail log volume... the update_downloads task runs every 10 min :-/ We can change our papertrail settings to filter out those log lines, but then what would the purpose be?

jtgeibel:
I think the additional log volume will be small compared to our web traffic logging.  We can always scale it back if some of the lines end up being unhelpful.  Mainly, I'm trying to investigate the occasional jumps in response times and while I don't think its related to background work, it would be nice to have timestamps to see the precise sequence of these events relative to web traffic.

src/bin/enqueue-job.rs
```diff
@@ -5,7 +5,11 @@ use diesel::PgConnection;
 fn main() -> AppResult<()> {
     let conn = db::connect_now()?;
     let mut args = std::env::args().skip(1);
-    match &*args.next().unwrap_or_default() {
+
+    let job = args.next().unwrap_or_default();
+    println!("Enqueueing background job: {}", job);
```
smarnach: When called locally without a command-line argument, the eror output isn't great, but I guess it's good enough – this isn't a tool meant to be used by many end users.

PR #2067, opened 2019-12-26
tests: Add basic "successful/failed login" test cases

Turbo87:
This PR adds two basic application tests for the login button and the `login` route. Please note that this does not test the `github-login` and `github-authorize` routes yet, which we will do in a follow-up PR.

r? @locks 

/cc @carols10cents

### DIFF
--- a/app/routes/login.js
+++ b/app/routes/login.js
@@ -1,5 +1,6 @@
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
+import window from 'ember-window-mock';
 
 /**
  * This route will open a popup window directed at the `github-login` route.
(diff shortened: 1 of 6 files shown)


carols10cents:
Catching up on what's been merged, this is awesome!!! Thank you so much!!!

smarnach:
Yup, this is one of the tests that were most sorely missing from the frontend. Actually, the first time I tried to set up crates.io on my machine [the login was broken](https://github.com/rust-lang/crates.io/pull/1739), and it took me a while to figure out what's wrong since the error output goes to the ephemeral new window, and all tests pass. These tests would have caught the error with an easily accessible error message, and the broken change would have been fixed within minutes before it was even merged.

PR #2071, opened 2019-12-26
Fix cache control headers set for S3 uploads

jtgeibel:
The headers intended for crates and readme uploads were swapped.

r? @smarnach


smarnach:
Snap, how did we manage to get _htis_ wrong? I guess I was the one who let this slip in the review.

It's a bit unfortunate the we uploaded readme files with an infinite lifetime, since they may theoretically linger around in some cache until the end of time. We should issue a CloudFront invalidation for the affected files to at least get them out of CloudFront.

PR #2072, opened 2019-12-26
Remove some usage of `unwrap()`

jtgeibel:
This commit reduces the usage of unwrap within the request/response
lifecycle, and documents a few remaining cases.  Usage of unwrap in
tests, binaries, or server boot code is generally okay and was not
reviewed.

Remaining work:

* Some usage of unwrap in `git.rs` and `src/tasks/` background jobs.
   This code is not directly in the request/response lifecycle, but
  should be reviewed.
* Current usage in middleware seems okay, but could be documented.
  In most cases, such as with the handler in `AroundMiddleware`, the
  conduit API ensures a handler is set.

### DIFF
--- a/src/s3/Cargo.toml
+++ b/src/s3/Cargo.toml
@@ -6,6 +6,7 @@ authors = ["Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-lang/crates.io"
 description = "Interaction between crates.io and S3 for storing crate files"
+edition = "2018"
 
 [lib]
 
--- a/src/s3/error.rs
+++ b/src/s3/error.rs
@@ -3,6 +3,7 @@ use std::fmt;
 use openssl::error::ErrorStack;
 use reqwest::Error as ReqwestError;
 
+#[derive(Debug)]
 pub enum Error {
     Openssl(ErrorStack),
     Reqwest(ReqwestError),
@@ -28,3 +29,5 @@ impl fmt::Display for Error {
         }
     }
 }
+
+impl std::error::Error for Error {}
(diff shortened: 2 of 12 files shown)


src/s3/Cargo.toml
```diff
@@ -6,6 +6,7 @@ authors = ["Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-lang/crates.io"
 description = "Interaction between crates.io and S3 for storing crate files"
+edition = "2018"
```
carols10cents: I wasn't expecting a conversion to the 2018 edition in this PR, but it's not very much, so I'll allow it ;)

src/s3/error.rs
```diff
@@ -0,0 +1,30 @@
+use std::fmt;
+
+use openssl::error::ErrorStack;
+use reqwest::Error as ReqwestError;
+
+pub enum Error {
```
carols10cents: What do you think about implementing `std::error::Error` on this type? I think it'll reduce surprise if someday we decide to do this in something that returns `Box<dyn std::error::Error>`. I could see an argument for YAGNI too though.

jtgeibel:
I've push a commit to add impls for `Debug` and `std::error::Error`.  It felt like I was missing something.  I took a look at #2032 and it looks like I implemented these traits for the error type introduced there, so I think it makes sense to add it here as well, even if not currently needed in the codebase.

> I wasn't expecting a conversion to the 2018 edition in this PR, but it's not very much, so I'll allow it ;)

Yeah, looks like it was missed before and I has already removed the `extern crate`s.  Thanks for letting it slide this time!

@bors r=carols10cents

Issue #1930, opened 2019-11-28
Crate name / search term should be in the page title


upsuper:
When I'm looking for some crates, I end up getting several crates.io pages which basically look the same from the title, which is unfortunate.

![crates-io-screenshot](https://user-images.githubusercontent.com/333750/69807020-59a9f300-1238-11ea-809b-5178e966dc9a.png)

I suggest that the page title should include the crate name / search term etc., and ideally at the beginning of the page title, which would make it easier to find the desired page.

smarnach:
I agree with this suggestion. I think a title in the form "<search term> · Search results · crates.io" or "<crate name> · crates.io" would make sense. This is in line with what Wikipedia and many other web pages do.

hbina:
Any related files that I can look at?

phlopsi:
Someone changed the title generation over the course of the last two (?) months. Previously, when I bookmarked a crate page it'd always show, e.g. "serde - Cargo: packages for Rust", "lalrpop - Cargo: packages for Rust", etc. The current title being the same for every crate ("crates.io: Rust Package Registry") is a clear feature regression. It's annoying, that I have to edit every bookmark to contain the crate name, now. Rather than an enhancement, this should be treated as a bug, because we already had this feature and someone removed it.

If this feature gets re-implemented I'd suggest, that a should be test added, that checks, if the crate name is somewhere inside the page title. Clearly, if someone removes a test, it must be noticed and questioned during code review, i.e. a test should prevent any future PRs to remove this feature, **again**.

smarnach:
@phlopsi There has been a change to the branding in #1787, but this really only changed the strings from "Cargo" to "crates.io". But it is true that this used to work correctly, and we even have a [test for it that still passes](https://github.com/rust-lang/crates.io/blob/4a1542cb17ae50c2bfd0f7e85876d82617e30415/tests/acceptance/crate-test.js#L70). Clearly something has regressed, so I will tag this as a bug instead.

carols10cents:
Interesting, so I think the tests are passing because of [this commit](https://github.com/rust-lang/crates.io/commit/034fd451e3c66ea09217600629c5ff45c4e86ade), which was trying to work around [Testem](https://www.npmjs.com/package/testem) adding its own title element. The [testem team feels that testing the title should be done at an integration test level](https://github.com/testem/testem/issues/195).

[...]

PR #2074, opened 2019-12-27
restore crate prefixes to browser-tab titles

alsuren:
I'm not sure when it got broken, but all of the crates.io brower tabs I have open currently say `crates.io: Rust Package Registry` when they previously said things like `mio - crates.io: Rust Package Registry` or `Search Results for 'mio' - crates.io: Rust Package Registry`. This patch fixes this.

There is a bunch of code that sets the page title dynamically,
but it is all being overridden by a title tag in the main
app/index.html template (the DOM is only allowed to have
a single title tag in it, so subsequent title tags are
ignored by all browsers).

There is probably a better way to do this (using fastboot
to fill in the correct title on the backend would be ideal)
but this patch should be enough to restore the dynamic
titles for browsers with javascript enabled.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -5,7 +5,9 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
-    <title>crates.io: Rust Package Registry</title>
+    <noscript>
+      <title>crates.io: Rust Package Registry</title>
+    </noscript>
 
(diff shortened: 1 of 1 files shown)


locks:
Thanks for the help! I also think it's worth it to get this in and see about fastboot afterwards.
@bors r+

carols10cents:
This PR fixes https://github.com/rust-lang/crates.io/issues/1930! Thank you!

PR #2076, opened 2019-12-28
Respect `@media (prefers-color-scheme: dark)` by adding a "dark theme"

JonathanHelianthicusDoe:
This adds some new `@media` queries to change the look of crates.io to a dark theme that respects the original design, any time the user agent has `prefers-color-scheme: dark`.

See [the relevant MDN entry](https://developer.mozilla.org/docs/Web/CSS/@media/prefers-color-scheme). Also note that in Firefox you can set `ui.systemUsesDarkTheme` to `1` (as an integer, not a string nor a boolean) to cause `prefers-color-scheme` to have the value `dark`.

The changes made are based on [a usercss that I created](https://github.com/JonathanHelianthicusDoe/crates-io-dark). The usercss itself is actually CC BY-SA 4.0+ licensed, but just to be clear, I do of course license the contents of this PR dually under the MIT license and version 2.0 of the Apache license.


locks:
Thanks for contributing! I haven't discussed this work with the team, but if the decision is to move forward on this, we might want to figure out percy tests so we don't have regressions later on.

smarnach:
A dark theme sounds like a good idea! I haven't reviewed this PR, but in principle I'm in favour of merging this. Percy tests for this would be awesome as well.

locks:
Thank you for updating the PR!

Turbo87:
@JonathanHelianthicusDoe hey, sorry this has been lingering for so long. the plan is to significantly revamp the whole design of the page in the next months and this dark scheme support would conflict with those plans at the moment. we can revisit this idea after the new design has landed though. due to these reasons I'll go ahead and close this PR for now. sorry! 😞

teohhanhui:
> the plan is to significantly revamp the whole design of the page in the next months

That sounds like the greatest opportunity to think dark theme-first.

kotx:
> @JonathanHelianthicusDoe hey, sorry this has been lingering for so long. the plan is to significantly revamp the whole design of the page in the next months and this dark scheme support would conflict with those plans at the moment. we can revisit this idea after the new design has landed though. due to these reasons I'll go ahead and close this PR for now. sorry! 😞

It's been a while, any updates on a dark theme?
Related: #4348

PR #2077, opened 2019-12-28
Defer user authentication until requested by endpoint

jtgeibel:
Many endpoints do not require user authentication, and by moving the
logic out of the middleware layer these endpoints now avoid obtaining a
database connection from the pool and 1 or 2 queries.

Due to lifetime issues a small middlware component remains.  Further
details are described in the middleware's documentation.

r? @carols10cents

### DIFF
--- a/src/controllers/krate/follow.rs
+++ b/src/controllers/krate/follow.rs
@@ -8,15 +8,12 @@ use crate::models::{Crate, Follow};
 use crate::schema::*;
 
 fn follow_target(req: &dyn Request, conn: &DieselPooledConn<'_>) -> AppResult<Follow> {
-    let user = req.user()?;
+    let user_id = req.authenticate(conn)?.user_id();
     let crate_name = &req.params()["crate_id"];
     let crate_id = Crate::by_name(crate_name)
         .select(crates::id)
         .first(&**conn)?;
-    Ok(Follow {
-        user_id: user.id,
-        crate_id,
-    })
+    Ok(Follow { user_id, crate_id })
 }
 
 /// Handles the `PUT /crates/:crate_id/follow` route.
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -8,7 +8,7 @@ use crate::controllers::cargo_prelude::*;
 use crate::git;
 use crate::models::dependency;
 use crate::models::{
-    insert_version_owner_action, Badge, Category, Keyword, NewCrate, NewVersion, Rights, User,
+    insert_version_owner_action, Badge, Category, Keyword, NewCrate, NewVersion, Rights,
     VersionAction,
 };
 
@@ -39,9 +39,11 @@ pub fn publish(req: &mut dyn Request) -> AppResult<Response> {
     // - Then the .crate tarball length is passed to the upload_crate function where the actual
     //   file is read and uploaded.
 
-    let (new_crate, user) = parse_new_headers(req)?;
+    let new_crate = parse_new_headers(req)?;
 
     let conn = app.diesel_database.get()?;
+    let ids = req.authenticate(&conn)?;
+    let user = ids.find_user(&conn)?;
 
     let verified_email_address = user.verified_email(&conn)?;
     let verified_email_address = verified_email_address.ok_or_else(|| {
@@ -150,7 +152,7 @@ pub fn publish(req: &mut dyn Request) -> AppResult<Response> {
             &conn,
             version.id,
             user.id,
-            req.authentication_source()?.api_token_id(),
+            ids.api_token_id(),
             VersionAction::Publish,
         )?;
 
@@ -223,9 +225,8 @@ pub fn publish(req: &mut dyn Request) -> AppResult<Response> {
 /// Used by the `krate::new` function.
 ///
 /// This function parses the JSON headers to interpret the data and validates
-/// the data during and after the parsing. Returns crate metadata and user
-/// information.
-fn parse_new_headers(req: &mut dyn Request) -> AppResult<(EncodableCrateUpload, User)> {
+/// the data during and after the parsing. Returns crate metadata.
+fn parse_new_headers(req: &mut dyn Request) -> AppResult<EncodableCrateUpload> {
     // Read the json upload request
     let metadata_length = u64::from(read_le_u32(req.body())?);
     req.mut_extensions().insert(metadata_length);
@@ -264,6 +265,5 @@ fn parse_new_headers(req: &mut dyn Request) -> AppResult<(EncodableCrateUpload, 
         )));
     }
 
-    let user = req.user()?;
-    Ok((new, user.clone()))
+    Ok(new)
 }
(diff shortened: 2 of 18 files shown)


src/middleware/current_user.rs
```diff
@@ -1,104 +1,37 @@
...
-
-            drop(conn);
-
-            if let Some((api_token, user)) = user_auth {
-                // Attach the `User` model from the database and the API token to the request
-                req.mut_extensions().insert(user);
-                req.mut_extensions().insert(api_token);
-            }
+        if let Some(id) = req.session().get("user_id").and_then(|s| s.parse().ok()) {
+            req.mut_extensions().insert(TrustedUserId(id));
```
Turbo87: out of curiosity, why is it not possible to put this into the authenticate call too?

src/controllers/krate/follow.rs
```diff
@@ -8,15 +8,12 @@ use crate::models::{Crate, Follow};
...
+    let user_id = req.authenticate(conn)?.user_id();
     let crate_name = &req.params()["crate_id"];
     let crate_id = Crate::by_name(crate_name)
         .select(crates::id)
         .first(&**conn)?;
-    Ok(Follow {
-        user_id: user.id,
-        crate_id,
-    })
+    Ok(Follow { user_id, crate_id })
```
carols10cents: I like that this abstraction gets rid of a `User` lookup in cases like this spot when it's not needed 👍

src/controllers/krate/publish.rs
```diff
@@ -40,9 +40,11 @@ pub fn publish(req: &mut dyn Request) -> AppResult<Response> {
     // - Then the .crate tarball length is passed to the upload_crate function where the actual
     //   file is read and uploaded.
 
-    let (new_crate, user) = parse_new_headers(req)?;
+    let new_crate = parse_new_headers(req)?;
 
     let conn = app.diesel_database.get()?;
+    let ids = req.authenticate(&conn)?;
+    let user = ids.find_user(&conn)?;
```
carols10cents: LOVE that this is out of `parse_new_headers` now!!!!

src/controllers/version.rs
```diff
@@ -5,15 +5,17 @@ pub mod yank;
 
 use super::prelude::*;
 
+use crate::db::DieselPooledConn;
 use crate::models::{Crate, CrateVersions, Version};
 use crate::schema::versions;
 
-fn version_and_crate(req: &mut dyn Request) -> AppResult<(Version, Crate)> {
+fn version_and_crate(req: &dyn Request) -> AppResult<(DieselPooledConn<'_>, Version, Crate)> {
```
carols10cents: This function is a little strange now that it returns a connection too, but 🤷‍♀ it's not exactly the cleanest abstraction to begin with anyway given that half the time we don't even use the returned crate. Always more to do.

src/middleware/current_user.rs
jtgeibel: > I think this req.session() is the call that needs the mutable reference to the request

Yes, that's exactly the problem I ran into.  `session()` would probably be better called `session_mut()`, since it calls `mut_extensions()` and `find_mut()` [internally](https://github.com/conduit-rust/conduit-cookie/blob/master/src/session.rs#L100-L105) to return a `&mut HashMap<_, _>`.

src/middleware/current_user.rs
```diff
@@ -1,104 +1,37 @@
+//! Middleware that captures the `user_id` from the signed session cookie
+//!
+//! Due to lifetimes it is not possible to call `mut_extensions()` while a reference obtained from
+//! `extensions()` is still live.  The call to `conduit_cookie::RequestSession::session` needs
+//! mutable access to the request and its extensions, and there is no read-only alternative in
+//! `conduit_cookie` to access the session cookie.  This means that it is not possible to access
```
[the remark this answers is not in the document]
jtgeibel: > Would our code be better if there was a read-only alternative to get the session cookie?

It would be nice to support that upstream in `conduit-cookie`.  We only use the session cookie here and during the login oauth workflow, and in both cases we don't need a database connection at that particular moment so it hasn't been a problem, yet.  Although I don't really see us adding more complex cookie logic anytime soon.

> It looks like the write access is used for caching the results of decoding the cookie.

As I understand it, the middleware layer from `conduit-cookie` decodes the cookie before the request is processed by the endpoint, and then the middleware adds the (possibly updated) cookie back into every response.  So the `&mut HashMap<_, _>` returned here is what provides the API for mutating the session cookie.

There are some changes (especially with respect to bumping old deps, and more especially wrt our very old dependency on `semver`) that I would like make to the `conduit-*` ecosystem.  Last I looked we (crates.io) are the only public users of these crates, so it would be awesome if we could make some improvements and release new versions of these crates.  Longer term, it would be nice if we could add async/await support in our middleware stack.

For now, I think its fine to document the workaround.  This also provides an easy way for the test harness to simulate cookie authenticated requests, although it would certainly be nice to construct real cookies in the test harness instead.

src/controllers/util.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    pub fn find_user(&self, conn: &PgConnection) -> AppResult<User> {
+        User::find(conn, self.user_id())
+            .chain_error(|| internal("user_id from cookie or token not found in database"))
+    }
+}
+
+impl<'a> UserAuthenticationExt for dyn Request + 'a {
+    /// Obtain `CurrentUserIds` for the request or return an `Unauthorized` error
```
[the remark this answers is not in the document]
jtgeibel: Yeah, that was the name at one point.  Updating to `AuthenticatedUser`.

src/controllers/util.rs
```diff
@@ -0,0 +1,44 @@
+use super::prelude::*;
+
+use crate::middleware::current_user::TrustedUserId;
+use crate::models::{ApiToken, User};
+use crate::util::errors::{internal, AppError, AppResult, ChainError, Unauthorized};
+
+#[derive(Debug)]
+pub struct AuthenticatedUser(i32, Option<i32>);
```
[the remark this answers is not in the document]
jtgeibel: Good point.  I'll keep the getters to enforce read-only access, but converting to a regular struct.

src/controllers/version.rs
jtgeibel: > This function is a little strange now that it returns a connection too

I would be cleaner to obtain a connection and pass it to this function, but it just ~~isn't~~ wasn't possible because of the lifetimes here.  As I write this, I now see that I eventually dropped the `mut`ness of the `&dyn Request`, so it should now be possible to pass in a `&PgConnection` as well, at least I think.

> Always more to do.

Yup!

jtgeibel:
I've rebased to resolve merge conflicts, and added a 2nd commit that addresses the review comments.

jtgeibel:
Rebased to resolve merge conflicts.  I believe all review comments have been addressed.  I'd like to deploy to staging tonight, and possibly production tomorrow.

@bors r=carols10cents

PR #2087, opened 2020-01-04
Don't make duplicated requests through Ember Data

kzys:
ember-data-storefront provides FastbootAdapter which caches XHRs in
Fastboot's Shoebox.

This will prevent duplicated requests on /crates and other
Ember Data-powered pages.

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -4,8 +4,23 @@ import { inject as service } from '@ember/service';
 
 export default RESTAdapter.extend({
   fastboot: service(),
+  fetcher: service(),
 
   namespace: 'api/v1',
+
+  ajax(url, type, options) {
+    if (type === 'GET') {
+      let cache = this.fetcher.get(url);
+      if (cache) {
+        return cache;
+      }
+    }
+
+    return this._super(url, type, options).then(resp => {
+      this.fetcher.put(url, resp);
+      return resp;
+    });
+  },
 
   headers: computed('fastboot.{isFastBoot,request.headers}', function () {
     if (this.fastboot.isFastBoot) {
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -10,14 +10,14 @@ export default RESTAdapter.extend({
 
   ajax(url, type, options) {
     if (type === 'GET') {
-      let cache = this.fetcher.get(url);
+      let cache = this.fetcher.get(url, options);
       if (cache) {
         return cache;
       }
     }
 
     return this._super(url, type, options).then(resp => {
-      this.fetcher.put(url, resp);
+      this.fetcher.put(url, options, resp);
       return resp;
     });
   },
--- a/app/services/fetcher.js
+++ b/app/services/fetcher.js
@@ -7,16 +7,17 @@ const KEY = 'ajax-cache';
 export default class FetcherService extends Service {
   @service fastboot;
 
-  get(url) {
+  get(url, options) {
     let shoebox = this.fastboot.shoebox;
     if (!shoebox) {
       return;
     }
     let cache = shoebox.retrieve(KEY) || {};
-    return cache[url];
+    let key = cacheKey(url, options);
+    return cache[key];
   }
 
-  put(url, obj) {
+  put(url, options, obj) {
     let fastboot = this.fastboot;
     let shoebox = this.fastboot.shoebox;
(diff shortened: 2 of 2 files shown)


app/adapters/application.js
```diff
@@ -1,9 +1,10 @@
 import RESTAdapter from '@ember-data/adapter/rest';
 import { inject as service } from '@ember/service';
 import { computed } from '@ember/object';
+import FastbootAdapter from 'ember-data-storefront/mixins/fastboot-adapter';
 
-export default RESTAdapter.extend({
-  fastboot: service(),
+export default RESTAdapter.extend(FastbootAdapter, {
+  fetcher: service(),
```
kzys: Nah, I don't need fetcher but need and fastboot should be kept. Will fix.

jtgeibel:
Based on some concerns raised on Discord, lets reassign.

r? @locks

kzys:
From [Discord](https://discordapp.com/channels/442252698964721669/448525639469891595/663097199990800424)

> @kzys I asked for second opinions on https://github.com/rust-lang/crates.io/pull/2087 and I think we shouldn't bring in ember-data-storefront just for this

I don't have strong opinion on ember-data-storefront. If we don't want to add the package, can we inject `fetch()` into RESTAdapter instead? I'm not familiar about Ember's dependency injection mechanism.

https://github.com/emberjs/data/blob/94aef4dd61d8697636d23631441c19ff96e38f65/packages/adapter/addon/rest.js#L1057

kzys:
Ping @locks?

locks:
Looking into it today!

kzys:
@locks We could inject something that queries Fastboot's Shoebox on adapters from initializers, but Ember's Adapter API doesn't provide a good hook point which we can utilize. Both `ajax` and `_ajaxRequest` are declared as private.

https://api.emberjs.com/ember-data/3.4/classes/DS.RESTAdapter?show=private

Surprisingly even ember-cli-fastboot is overriding _ajaxRequest.

https://github.com/ember-fastboot/ember-cli-fastboot/blob/6a5d6c6e38c30e5afda20db34de2ff5745b1416f/fastboot/initializers/ajax.js#L33

kzys:
Ping @locks?

locks:
Sorry, busy days lately. If I can't address this by Monday I will step out of the way.

kzys:
I have removed `ember-data-storefront` on the last revision.

kzys:
@jtgeibel @smarnach - Can you take a look since @locks is busy?

jtgeibel:
@kzys I had some trouble testing this in my staging area at https://jtgeibel-staging-crates-io.herokuapp.com/.  (This PR is currently deployed there with `USE_FASTBOOT=staging-experimental`.)  I was running into issues on both master and this branch.

So I focused on getting a better local environment set up.  I've opened #2215 and I've got things working there now.  I'd like to do a bit more testing, but this PR seems to work well for me locally (with and without fastboot) when merged with this PR.

I'm not sure what the issue with `staging-experimental` is when deployed to Heroku.  In that environment I think maybe the requests from the fastboot server to the backend are being routed in a strange way.  I'm not sure if they are going directly to the backend, locally through `nginx`, or externally through the Heroku router.

app/adapters/application.js
jtgeibel: Does this comment still apply?

kzys:
> I'm not sure if they are going directly to the backend, locally through nginx, or externally through the Heroku router.

I think it should be directly request the backend, but need to check in my environment as well.

locks:
I just found out about https://github.com/cardstack/ember-data-fastboot, which should help with the boilerplate.

kzys:
@locks It is somewhat coupled with JSONAPISerializer, which we don't use.

> Right now, this addon assumes that your application is using DS.JSONAPISerializer as your application's default serialization method.

app/adapters/application.js
kzys: No. I need both `fetcher` and `fastboot` now.

kzys:
I cannot keep up with all the refactorting the team is doing. Ping me once that is over...

Turbo87:
In #7301 we've removed most of the unused fastboot code, since it doesn't really seem well supported by Ember.js these days. Since this PR is building on top of fastboot, I'll close it for now. Thank you for your work anyway! :)

PR #2090, opened 2020-01-05
Add computer-vision slug

vadixidav:
Right now the closest thing in the list to computer vision is `science::robotics`. Computer vision algorithms are not always used for robotics. This slug also allows computer vision crates to be found easier instead of being lost among the other robotics crates.

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -81,7 +81,7 @@ Algorithms for making data smaller.\
 """
 
 [computer-vision]
-name = "Computer Vision"
+name = "Computer vision"
 description = """
 Crates for comprehending the world from video or images.\
 """
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -80,6 +80,12 @@ description = """
 Algorithms for making data smaller.\
 """
 
+[computer-vision]
+name = "Computer Vision"
+description = """
+Crates for comprehending the world from video or images.\
+"""
+
 [config]
 name = "Configuration"
 description = """


src/boot/categories.toml
```diff
@@ -80,6 +80,12 @@ description = """
 Algorithms for making data smaller.\
 """
 
+[computer-vision]
+name = "Computer Vision"
```
smarnach: To be consistent with the other category names, the "v" in "vision" should be lower case.

carols10cents:
@vadixidav do you know how many crates would fit into this category today?

vadixidav:
@carols10cents Some crates:

## Ones i know of because I have used them or someone in rust-cv used them
* https://crates.io/crates/akaze
* https://crates.io/crates/akaze-util
* https://crates.io/crates/opencv
* https://github.com/nebgnahz/cv-rs (not currently published on crates.io)
* https://crates.io/crates/visual-odometry-rs
* https://crates.io/crates/levenberg-marquardt
* https://crates.io/crates/cv-core
* https://crates.io/crates/p3p
* https://crates.io/crates/sample-consensus
* https://crates.io/crates/arrsac
* https://crates.io/crates/ndarray-vision
* https://crates.io/crates/vision

## Ones that I found on crates.io just now that look fleshed out
* https://crates.io/crates/edge-detection
* https://crates.io/crates/aperature
* https://crates.io/crates/openni2
* https://crates.io/crates/superpixel
* https://crates.io/crates/dssim

## Unfinished or empty crates that probably would have used the computer-vision slug
* https://crates.io/crates/simplecv
* https://crates.io/crates/xifeng

I am also very actively working on adding as many computer vision algorithms as possible, so this is a growing number. I don't know the amount of crates under other slugs, but computer vision is pretty broad. I have not included some crates that are related to computer vision (or were made for computer vision), but have other uses (such as https://crates.io/crates/hnsw).

vadixidav:
One alternative could be to put the `computer-vision` slug under the `science` one. I felt that science wasn't the right place for it, but `robotics` is also under science, despite that some of the crates there were never intended for or related to science (including most of the computer vision crates, which are also in `science::robotics`).

smarnach:
I'm not convinced this deserves a separate category yet. We already have a few rather specialized ones like "Text editors", but the vast majority of categories is more general. I'd be reluctant to add another quite narrow one. I'll leave this open for a while to see who else wants to chime in.

carols10cents:
What do you think about a top-level "machine learning" category (and remove the mention of "machine learning" from the description of the science category) with a "computer vision" subcategory?

vadixidav:
@carols10cents So far almost of the crates we have made would not fall under machine learning. Most of them are related to solving or iteratively improving camera or 3d point positions and camera parameters based on points detected and matched with other crates. For instance, most of the functionality in OpenCV is not related to machine learning. Additionally, OpenMVG, which is where many of our ports are sourced from, doesn't utilize any machine learning.

Although machine learning and computer vision have started to overlap greatly in the last several years, I still wouldn't say it falls under machine learning. I feel like computer vision crates that relate to machine learning are more likely just to be considered "machine learning" crates anyways. Perhaps my view is skewed though, as I am not a machine learning specialist, so I am more focused on the other parts of computer vision. We should probably get the opinion of a second person, at least.

vadixidav:
So, it doesn't seem there is much activity here. I think what we should do is agree upon what slug would be a good fit for different computer vision crates.

Some are related to image processing. This might involve, for instance, the [Canny edge detector](https://en.wikipedia.org/wiki/Canny_edge_detector), denoising, gaussian blur/rescale, and other image operations. The closest category for these kinds of crates seems to be the `graphics` slug, but that doesn't seem appropriate considering the description: "Crates for graphics libraries and applications, including raster and vector graphics primitives such as geometry, curves, and color." It doesn't seem to be related to image processing, so perhaps there should be an `image-processing` slug. Technically, there is a `multimedia` tag with the description: "Crates that provide audio, video, and image processing or rendering engines." I don't think that this description intends to include things such as the canny edge detector or other image analysis and processing routines used in computer vision.

Some are related to photogrammetry & multi-view geometry. This will involve crates such as [`akaze`](https://crates.io/crates/akaze). These currently are best suited to go in `science::robotics`, because they are sometimes used in [SLAM](https://en.wikipedia.org/wiki/Simultaneous_localization_and_mapping). These descriptors are sometimes used in robot navigation via SLAM, but not always. Faster descriptors (ORB, AKAZE less-so, definitely not SIFT) tend to be used for robotics, while the more robust ones (such as [SIFT](https://en.wikipedia.org/wiki/Scale-invariant_feature_transform)) are rarely used in robotics, but they are used in SfM (https://en.wikipedia.org/wiki/Structure_from_motion). Crates that are relevant to SfM don't really have a category, but they are used to do things like inspections, geographical surveys, or creating 3d models to preserve historical sites or buildings. SfM itself falls under the concept of computer vision, but it is basically summed up as reconstructing 3d information from 2d images.

There are also crates that do operations on images to prepare them for some of these processes. This would be, for instance, image undistortion. For instance, [here](https://opencv-python-tutroals.readthedocs.io/en/latest/py_tutorials/py_calib3d/py_calibration/py_calibration.html) is an OpenCV tutorial on this topic in OpenCV with Python. I don't know which category these go into, but the `graphics` or `multimedia` slugs seem to be the closest matches.

You also have depth map estimation. This is related to robotics in some cases, so it could go in that slug, but sometimes it is not. Python OpenCV tutorial on this topic is [here](https://docs.opencv.org/master/dd/d53/tutorial_py_depthmap.html). I think, at present, this would best go under `science::robotics`.

Finally, there are the most well-known of all computer vision tasks: convolutional neural networks. They have made a large impact on modern society, but I would argue that since traditional machine learning frameworks (Tensorflow, PyTorch, etc) include these networks as built-in primitives, there is little reason to create a category specifically for them.

I have covered some of the largest components of computer vision here. I believe that while we may not create a `computer-vision` slug, we might be able to break down the different categories that we see here:

* Some things seem to be involved in processing imagery, such as the [`imageproc`](https://crates.io/crates/imageproc) crate. These crates would probably be best served by having a separate category for `image-processing`.
* Some things work with turning 2d information into 3d information. These crates can probably continue to go into both `science` and `science::robotics` for the time being.
* Some things are common to all the above, mostly having to do with camera models, epipolar geometry, focal length, distortion, etc. Perhaps these things can also continue to exist in `science` and `science::robotics`, but if there was a category such as `science::optics`, that would work pretty well for these kinds of things. I guess I am thinking that perhaps the core of "computer vision" might be able to live in a broader category involving cameras or optics, but I don't know what the name of said category would be.

Does anyone have any other thoughts on this? I think we should create a direction for which categories things like this should go, particularly as Rust becomes more relevant in this field.

jtgeibel:
@vadixidav I think you make a good case for adding this at the top level, and not under a new machine-learning category.  I've added to the agenda for Friday to bring up with the team (since we discussed previously but missed this angle).

vadixidav:
@jtgeibel That is good to hear. It is difficult to find places for many of these crates. I appreciate the time spent discussing this issue in meetings.

jtgeibel:
We discussed merging this yesterday, and this is good to go as-is.  I should be able to deploy this in the next hour or 2.

@bors r+

jtgeibel:
Deployed.  This category should now be available to new publishes.

PR #2091, opened 2020-01-06
Fix fallback to unstable version when looking for unyanked versions

charmander:
See the commit “Fix fallback to unstable version” for the behaviour change.

- [ ] find out how to write a test for this

### DIFF
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -61,7 +61,7 @@ export default Route.extend({
                 }
               });
 
-              if (latestStableVersion == null) {
+              if (latestUnyankedVersion == null) {
                 // There's not even any unyanked version...
                 params.version_num = maxVersion;
               } else {
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -45,9 +45,7 @@ export default Route.extend({
           .then(versions => {
             const latestStableVersion = versions.find(version => {
               // Find the latest version that is stable AND not-yanked.
-              if (!isUnstableVersion(version.get('num')) && !version.get('yanked')) {
-                return version;
-              }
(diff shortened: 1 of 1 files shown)


charmander:
Actually, this might not fix anything:

> If `max_version` is `0.0.0` then all versions have been yanked

suggests that `latestUnyankedVersion == null` is supposed to be impossible here, inside https://github.com/rust-lang/crates.io/blob/a251368963d55ff12ecb65249cfe96950deb867b/app/routes/crate/version.js#L41. I’ll try to find out for sure.

Turbo87:
@charmander is there a GitHub issue for the thing that you're fixing here? do you think it would be possible to write a test case for it?

carols10cents:
Hi @charmander, we deliberately changed this in https://github.com/rust-lang/crates.io/pull/1857 to fix https://github.com/rust-lang/crates.io/issues/1144. If you can open a bug first to describe what you're trying to fix, we can work on the best way to fix it and test it! Thank you!

charmander:
@carols10cents This is a fix to a typo in that patch – the second check was supposed to be for the new `latestUnyankedVersion` instead of the already-checked `latestStableVersion` – but I’m still trying to figure out how to create a crate affected by it or prove that one doesn’t exist. Sorry for the wait.

charmander:
Actually, I don’t understand how #1857 could have changed any behaviour without this typo fix… maybe #1144 was really fixed by #1684?

charmander:
Okay, it looks like `max_version` comes from `Crate::top_versions`, which uses `CrateVersions::versions`, which only includes unyanked versions. So making a test for this isn’t possible and it’s just cleanup. Changes pushed accordingly!

PR #2093, opened 2020-01-06
Update rustc used in production

carols10cents:
This will make the deploy take longer.


jtgeibel:
I'm going to deploy this to staging, and then production to prime the build caches.

@bors r+

PR #2101, opened 2020-01-07
Tweak <noscript> text and placement

kzys:
After having FastBoot, JavaScript would not be "required" but still
necessary for full functionality of crates.io.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -23,14 +23,13 @@
     <link rel="search" href="/opensearch.xml" type="application/opensearchdescription+xml" title="Cargo">
   </head>
   <body>
-    {{content-for 'body'}}
     <noscript>
-        <div id="main">
-            <div class='noscript'>
-                This site requires JavaScript to be enabled.
-            </div>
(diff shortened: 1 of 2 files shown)


locks:
We might want to still have some kind of message, as some functionality will not be available if you don't have JS enabled. What do you think?

jtgeibel:
How about changing the message to something like: 

> Some functionality may not be available with JavaScript disabled.

Most users won't see the message at all, and with functionality like the download charts we may not ever reach complete feature parity.

We could also move the message to a better place with better styling, as it is currently located in a strange place below the footer.

kzys:
If so, we could probably put the message in app/index.html instead of adding this new file. What do you think?

jtgeibel:
Yeah, I agree.  I'd like to avoid adding another file that we'll need to keep synchronized with app/index.html.

jtgeibel:
LGTM!  I've updated the PR subject and text to reflect the changes.

@bors r+

PR #2105, opened 2020-01-08
Add 'unsafe-inline' for style-src CSP

jtgeibel:
This fixes the following error seen in the console:

> Content Security Policy: The page’s settings blocked the loading of a resource at inline (“style-src”). format+en,default+en,ui+en,corechart+en.I.js:596:193

r? @carols10cents

### DIFF
--- a/src/middleware/security_headers.rs
+++ b/src/middleware/security_headers.rs
@@ -39,13 +39,15 @@ impl SecurityHeaders {
         // to load scripts. Remove 'unsafe-eval' if google fixes the issue:
         // https://github.com/google/google-visualization-issues/issues/1356
         // or if we switch to a different graph generation library.
+        //
+        // 'unsafe-inline' for style-src is necessary for Google Charts
         headers.insert(
             "Content-Security-Policy".into(),
             vec![format!(
                 "default-src 'self'; \
                  connect-src 'self' https://docs.rs https://{}; \
                  script-src 'self' 'unsafe-eval' https://www.google.com; \
-                 style-src 'self' https://www.google.com https://ajax.googleapis.com; \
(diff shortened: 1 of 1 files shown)


jtgeibel:
I have this currently deployed in staging, and the only remaining console error appears to be from the Ember Inspector addon.  That error could be resolved by adding 'unsafe-inline' to `script-src` as well, but that's probably unnecessary (as I have no clue what potential Ember Inspector functionality this might be blocking).

I had hoped to also remove 'unsafe-eval' from `script-src`, but ran into errors.  The issue linked in the comment is closed and [this announcment](https://groups.google.com/forum/#!topic/google-visualization-api/hGys_iI99bE) also seems to indicate that this should no longer be necessary when using Google Charts.  It sounds like it should work when loading via the older jsapi, but maybe migrating to the [new loader](https://developers.google.com/chart/interactive/docs/basic_load_libs#update-library-loader-code) would work.

smarnach:
This error has been showing in the console forever, but I didn't notice any negative effect (i.e. no broken styling in the graph). The error message is not a problem by itself, so I don't think we should weaken our Content-Security-Policy header just to make this error message go away.

jtgeibel:
> This error has been showing in the console forever, but I didn't notice any negative effect (i.e. no broken styling in the graph).

I agree, I'm not aware of any functional improvements here, beyond removing the error from the console.

> The error message is not a problem by itself, so I don't think we should weaken our Content-Security-Policy header just to make this error message go away.

I'm not too worried about adding unsafe-inline to style-src.  I would gladly trade it in place of removing the unsafe-eval from script-src, if I can get that working.  I'll do some experimentation with the new loader to see if that works.

locks:
I didn't manage to update the PR with the rebase so I opened a new one here https://github.com/rust-lang/crates.io/pull/2292. Please continue discussion there!

PR #2108, opened 2020-01-09
Remove default null setting of the all_keywords parameter

### LINKED ISSUE #2106: Inconsistent search results between initial search and later search
I came across some strange behavior today, where a crate with the exact name match didn't appear in a search result. Reloading the page however did make the result appear.

Steps to reproduce:

- Go to `https://crates.io/`
- Search for `cargo_metadata` -> Match does not appear
- Reload the page -> Match does appear

Looking into the requests made, there appears to a slight difference in requests made between the search request done during site load, and any subsequent searches.

Initial search request: `https://crates.io/api/v1/crates?page=1&per_page=10&q=cargo_metadata`
Subsequent requests: `https://crates.io/api/v1/crates?all_keywords=&page=1&per_page=10&q=cargo_metadata`

carols10cents:
Fixes #2106, I think.

I don't think we actually need this parameter set to null here?

Alternatively, the backend could be changed to treat "presence but empty" differently than "presence with value", which currently we do not :(

### DIFF
--- a/app/controllers/application.js
+++ b/app/controllers/application.js
@@ -29,7 +29,6 @@ export default Controller.extend(EKMixin, {
         queryParams: {
           q: this.searchQuery,
           page: 1,
-          all_keywords: null,
         },
       });
     },


jtgeibel:
Works fine for me on staging.  Will deploy to prod as well since this is high impact.

@bors r+

PR #2127, opened 2020-01-14
mirage/factories: Use predictable values instead of random fake values

Turbo87:
This generally is easier to work with in my experience. If we want random values we can still generate them on a case-by-case basis, but by default we will have somewhat stable test fixtures now.

This should unblock #2120 :)

r? @locks


locks:
I agree with the above sentiment!
@bors r+

PR #2138, opened 2020-01-16
mirage: Calculate `crates_cnt` properties

Turbo87:
These should be consistent with the data that we have in the mock backend database.

r? @locks


locks:
Hm, we have CI failures.

Turbo87:
> we have CI failures.

yeah... they do look unrelated though 🤔

Issue #2134, opened 2020-01-15
Provide a better explanation for end users of what the maintenance intention badge means


jhinch:
Currently, if you are browsing crates.io, some crates display a maintenance intention. For example if you look at https://crates.io/crates/lazy_static you'll see a badge 'maintenance: passively-maintained'. When you hover over it, the alt text simple says 'Maintenance intention for this crate' and you cannot click through to any page to see what it means. Most other badges you can click through to something, such as a CI build, to get more information on what the badge means.

What made this worse for me is that I am running Privacy Badger, a popular addon which blocks trackers, protecting users privacy. It turns out that img.shields.io which hosts the image badge uses tracking cookies and as a result, the badge was blocked from loading and the alt text was the only thing which I would see. It took me some time to figure out that this was suppose to be a badge _and_ what the value of the badge actually meant. I managed to find the definitions in an RFC and the Cargo reference documentation

https://github.com/rust-lang/rfcs/blob/master/text/1824-crates.io-default-ranking.md#maintenance
https://doc.rust-lang.org/cargo/reference/manifest.html

I'm proposing that the experience here could be improved. Specifically I think that the alt text could be modified to reference what the maintenance intention actually is (e.g. 'Maintenance intention: passive maintenance') which has the added benefit of assisting people using screen readers. Secondly, I would suggest that the badge should link to the definitions for a more thorough explanation of what the maintenance intention means. This could be in the form of a page similar to https://crates.io/category_slugs where the maintenance intentions are lists and described, or to the Cargo documentation (linking to the RFC probably doesn't make sense as the definitions are only a subsection of a greater RFC).

I would be happy to explore implementing this if the crates.io team agree it is a good idea.

carols10cents:
I think your idea sounds great! Please let us know if you have any questions while implementing!!

jhinch:
I believe this ticket has been addressed now as part of the PR which was merged

Turbo87:
yep, thank you! :)

PR #2140, opened 2020-01-17
Improve the maintenance badge user experience

jhinch:
* Change the alt-text to be a little more descriptive
  and screen-reader friendly
* Link to the Cargo documentation from the badge which describes what the maintenance intentions are

This addresses #2134. Have a read through of my summary for the motivations on the ticket to get a little context of what this is trying to achieve.

### DIFF
--- a/app/components/badge-maintenance.hbs
+++ b/app/components/badge-maintenance.hbs
@@ -1,9 +1,11 @@
 <span ...attributes>
   {{#unless this.none}}
-    <img
-        src="https://img.shields.io/badge/maintenance-{{this.escapedStatus}}-{{this.color}}.svg"
-        alt="{{this.text}}"
-        title="{{this.text}}">
+    <a href="https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata">
+      <img
+          src="https://img.shields.io/badge/maintenance-{{this.escapedStatus}}-{{this.color}}.svg"
+          alt="{{this.text}}"
+          title="{{this.text}}">
+    </a>
   {{/unless}}
 
 </span>
--- a/app/components/badge-maintenance.hbs
+++ b/app/components/badge-maintenance.hbs
@@ -1,6 +1,6 @@
 <span ...attributes>
   {{#unless this.none}}
-    <a href="https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata">
+    <a href="https://doc.rust-lang.org/cargo/reference/manifest.html#the-badges-section">
       <img
           src="https://img.shields.io/badge/maintenance-{{this.escapedStatus}}-{{this.color}}.svg"
           alt="{{this.text}}"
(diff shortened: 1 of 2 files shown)


jhinch:
I'm happy to add tests but I'm unsure exactly what they would look like or if there are some examples I can look at. I'm a little out of practice with my frontend dev

jhinch:
I have rebased my changes onto master to fix the broken tests which appear to be fixed on master. I am not familiar percy.io or what is needed to get it to pass.

jhinch:
When I looked into the percy.io failure, it seems like the visual difference actually fixes a bug in terms of the vertical alignment (at least when I look at the vertical line spacing in comparison with other badges.

locks:
@jhinch thank you for helping out! Can you check the current conflict?

jhinch:
Thanks for the response! I'll take a look at resolving the conflict

jtgeibel:
@jhinch we discussed these changes in the team meeting yesterday, and the general consensus is that we should avoid duplicating documentation between crates.io and cargo.  So far, we've preferred to link to cargo documentation unless the documentation is specific to crates.io, such as the policies page.

Would you mind splitting the documentation improvement part of this PR into a PR against the cargo documentation?  I'd like to merge the improvements to the badge text and alt-text as part of this PR and then land a separate PR to link to the improved documentation.  There are already some upstream changes to the cargo documentation that are currently in beta (linked below).  Relative to the current stable the beta changes split `[badges]` into its own section.  Maybe the documentation improvements from this PR can be applied there with a new header added under that section so that the maintenance badge section can be linked to directly.

The downside to this approach is that the documentation changes then take 6 to 12 weeks to hit stable.  That still seems preferable to duplicating documentation between the two repos, and we could temporarily link to nightly/beta temporarily until the changes land on stable.

Okay, having said all that, there is a reasonable argument that the `[badges]` section is special.  It's entirely possible that alternative registries offer different (or even no) badges.  So maybe the badges section in cargo's documentation should just mention that the section is registry-specific and link to a page on crates.io.  Moving all the badges documentation to this repo would be a larger change and we'd want to make sure the cargo team agrees with that change in approach as well.

So in summary, would you mind trimming this PR down to just the alt-text and screen reader improvements and then opening either a PR against cargo to improve the documentation or opening an issue to discuss migrating the badges documentation?

Docs Source (master): https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/manifest.md#the-badges-section
Rendered (beta): https://doc.rust-lang.org/beta/cargo/reference/manifest.html#the-badges-section

jhinch:
Thanks for the detailed response. It's much appreciated. I agree that keeping the documentation in one place is much better if the Cargo documentation can be expanded on. For this PR, should I keep the link in place but link to the Cargo documentation, or would you like me to remove the link as well?

jtgeibel:
Linking to the cargo docs in this PR sounds fine, thanks!

jhinch:
PR has been updated with feedback addressed. Apologies for the delay.

PR #2142, opened 2020-01-18
CI: Include rustc version in key for `target/` cache

jtgeibel:
Based on a solution suggested by @pietroalbini.

I plan to push a few more things to this branch to test things and will probably drop the cache of rustc.  So for now, 
r? @ghost

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,10 +95,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: ~/.cargo/bin
-          key: ${{ runner.os }}-cargo-bin-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
-          restore-key: |
-            ${{ runner.os }}-cargo-bin-${{ matrix.rust }}-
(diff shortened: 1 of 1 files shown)


jtgeibel:
This is ready for review, but for some reason the caching behavior is still sub-optimal.  On this branch the 2nd build does restore the build cache from build 1, however, the lint and test steps still recompile about half of the crates.  Same for build 3.  For some reason, the cache seem to not be working for some dependencies, even though the files in `target/` are restored.  I don't see any obvious pattern in the list of affected crates either.

We're using the same `CARGO_INCREMENTAL: 0` and `RUSTFLAGS: "-C debuginfo=0 -D warnings"` that helped improve caching on Travis.  I think this PR improves the caching situation, but I'm not sure how to troubleshoot the remaining odd behavior.

r? @pietroalbini

pietroalbini:
This looks good to me!

@bors r+

jtgeibel:
> Insufficient privileges: Not in reviewers

How are you not on that list?  Here, let me help:
@bors r=pietroalbini

PR #2154, opened 2020-01-23
Update `version_downloads.counted` last to minimize lock contention

jtgeibel:
If we defer the update to `version_downloads.counted` until immediately
before the commit, it should eliminate the contention between counting
new downloads and propagating previous downloads.

Refs: #2153
See also: https://github.com/rust-lang/crates.io/issues/2153#issuecomment-577504032

### DIFF
--- a/src/tasks/update_downloads.rs
+++ b/src/tasks/update_downloads.rs
@@ -52,11 +52,6 @@ fn collect(conn: &PgConnection, rows: &[VersionDownload]) -> QueryResult<()> {
         let amt = download.downloads - download.counted;
 
         conn.transaction::<_, diesel::result::Error, _>(|| {
-            // increment the number of counted downloads
-            update(version_downloads::table.find(download.id()))
-                .set(version_downloads::counted.eq(version_downloads::counted + amt))
-                .execute(conn)?;
-
             // Update the total number of version downloads
             let crate_id = update(versions::table.find(download.version_id))
                 .set(versions::downloads.eq(versions::downloads + amt))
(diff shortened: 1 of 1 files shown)


jtgeibel:
Lets do r? @sgrif

jtgeibel:
I expect this PR will completely eliminate the occurrence of slow download requests.  Here's a timeline showing the progress so far:

* 14th - switched to performance dynos
* 15th - 49 slow downloads
* 16th - 56 slow downloads
* 17th - 117 slow downloads
* 18th - moved some expensive queries to the read-only replica; 5 slow downloads (2 before, 3 after)
* 19th - 9 slow downloads
* 20th - 3 slow downloads
* 21st - 2 slow downloads
* 22nd - 7 slow downloads
* 23rd - 1 slow downloads
* 24th - 1 slow downloads
* 25th - 2 today so far

sgrif:
I'd be really surprised that this was specifically causing lock contention, since the database will log any locks that are held for long enough to cause a slow request.

That said, this change seems completely reasonable on its own

jtgeibel:
> I'd be really surprised that this was specifically causing lock contention, since the database will log any locks that are held for long enough to cause a slow request.

The example I found in the logs from the 19th (https://github.com/rust-lang/crates.io/issues/2153#issuecomment-577504032) seem to be the only occurrence of the database complaining about this loudly.  Postges appears to have a 1 second threshold for generating log output about slow locks.  Searching the logs for `program:app/postgres waiting` shows a few other interesting bits of log output:

* `program:app/postgres.7060`, `program:app/postgres.11521`, and `program:app/postgres.5638` - slow at inserting into `version_downloads`
* `program:app/postgres.10814` - slow at inserting into `emails` :confused:
* `program:app/postgres.19904` - several interesting messages at once, but this appears to align with a deploy so maybe a migration locked something for a moment.

Overall, the events logged here are far fewer than the number of slow download requests we've observed.  The only explanation I have for that at the moment is that the download endpoint also issues a query joining `versions` and `crates`.  Maybe slow requests end up spending some extra time in each query, but not enough to log either query as >1 second by postgres.

sgrif:
I think this will only have a positive impact no matter what

jtgeibel:
I should be able to deploy this shortly.

@bors r=sgrif

Issue #2143, opened 2020-01-18
Linking directly to a new version from the front page shows no downloads


jtgeibel:
Under #1931 we made the decision to link directly from the front page to a version if we were showing that version number in parenthesis next to the crate name.  This seemed like a reasonable change to make at the time, but in retrospect it can cause confusion.  (In fact I confused myself because I just deployed a change, and wasn't sure why I wasn't seeing any downloads when clicking on crates from the front page.)

When clicking on a link under "Just Updated", you are taken directly to the newest version of that crate.  Because that version was just published, it will not have any downloads.  Even if the crate has old versions which are frequently downloaded, the version specific page shows only the downloads for that version.  This is unfortunate, because the "Just Updated" links are prominent on the front page and clicking on them gives the immediate impression that a crate has no history of downloads.

This issue is confounded by the fact that there is no easy way to navigate from a specific version of a crate, to the overview page for the crate.  These pages look almost identical, with the only differences I can see being the exact contents of the readme, and the download graph.

## Proposed solution

We should update the front page components to point to the crate overview pages like we did previously.  Optionally, we should provide an easy and prominent way to navigate from a version page back to the crate overview page.  We could also make it clearer on the version page that the graph is showing download for that version only.

This would be a great first PR if anyone is interested.

Zexbe:
Ok, I'm going to try

Zexbe:
I am done. The contribution steps didn't work for me.
I had to do figure out some extra steps to make npm work:
`npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe`
`npm install -g ember-cli`

Turbo87:
Resolved by #2155

PR #2155, opened 2020-01-23
Resolves issue 2143

Zexbe:
I made three changes to improve the experience.

    The homepage links go to the overview instead of a specific version
    I added the version next to the stats, so you can tell what they are for.
    I added a link back to the overview page next to the stats.

On the overview page it looks like this "Stats Overview"
On a specific version it looks like this "Stats Overview for 0.2.4 (see all)"

Resolves issue #2143

Remaking to run job again

### DIFF
--- a/app/components/crate-list-newest.hbs
+++ b/app/components/crate-list-newest.hbs
@@ -1,7 +1,7 @@
 <ol>
   {{#each @crates as |crate index|}}
     <li>
-      <LinkTo @route="crate.version" @models={{array crate.id crate.newest_version}} class="name" data-test-crate-link={{index}}>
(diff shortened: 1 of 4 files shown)


Zexbe:
Darn still failed, will have to try again later :(

Zexbe:
I tried to add an anchor, it didn't work. LinkTo does not have anything to add anchor. I also tried to create a link with an anchor without LinkTo, but because the page needs to be rendered, the anchor doesn't exist yet, so the browser does not scroll to it.

jtgeibel:
Thanks for digging into that @Zexbe!  There's probably a way to check for the fragment and scroll from JS once the page renders, but I'm not too worried about it.

@bors r+

PR #2167, opened 2020-01-27
Fix auth

### LINKED ISSUE #2166: Logging in for the very first time always fails (in production), works on 2nd try
What happened:

- I cleared all cookies out of Chrome
- I visited crates.io
- I logged in
- Instead of loading `/me`, I got the error "Failed to log in: must be logged in to perform that action"
- If I reload the page and log in again, it works

On preliminary investigation, it looks like the `TrustedUserId` extension isn't being set in time before being read, but you get like half logged in and the cookie gets set so that the next time you log in, the extension is set. Investigating further now.

carols10cents:
Set the TrustedUserId in the extensions when logging in

Fixes #2166.

In #2077 (d0b9bccc4cf80d3e311696a7a9e62ee469cce892), we changed the type of the authentication data stored in the extensions from `User` to `TrustedUserId`, but we missed this spot :( We're allowed to insert anything we want into the extensions, but if we use a different type than what we use when looking up data in the extensions, it won't find what we inserted.

I checked all the other locations of `mut_extensions` and `extensions` and I think they're good, but overall this system feels a bit loosey-goosey to me. Not sure how to improve it, though.

r? @jtgeibel

### DIFF
--- a/src/controllers/user/me.rs
+++ b/src/controllers/user/me.rs
@@ -13,17 +13,6 @@ use crate::views::{EncodableMe, EncodableVersion, OwnedCrate};
 
 /// Handles the `GET /me` route.
 pub fn me(req: &mut dyn Request) -> AppResult<Response> {
-    // Changed to getting User information from database because in
-    // src/tests/user.rs, when testing put and get on updating email,
-    // request seems to be somehow 'cached'. When we try to get a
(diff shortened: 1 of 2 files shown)


jtgeibel:
Great catch, thanks!

@bors r+

PR #2176, opened 2020-02-06
Use Option::as_deref to convert Option<String> to Option<&str>

jtgeibel:
Option::as_deref was introduced in rustc 1.40.0.  There is a new clippy lint on nightly that recommended the changes in the first commit.  The second commit converts 2 other equivalent syntax forms.  The old forms were:

* `as_ref().map(String::as_str)`
* `as_ref().map(|s| &**s)`
* `as_ref().map(|s| &s[..])`


locks:
Nice cleanup!

@bors r+

PR #2177, opened 2020-02-06
Fixing issues with owner invites and renamed accounts

carols10cents:
Pushing commit by commit to demonstrate test failures and fixes

### DIFF
--- a/src/models/owner.rs
+++ b/src/models/owner.rs
@@ -72,6 +72,7 @@ impl Owner {
         } else {
             users::table
                 .filter(users::gh_login.eq(name))
+                .filter(users::gh_id.ne(-1))
                 .first(conn)
                 .map(Owner::User)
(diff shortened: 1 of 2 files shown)


sgrif:
Since you're looking into this, https://github.com/rust-lang/crates.io/pull/1586 is still open and goes after the deeper problem, which is more than just owner invites

carols10cents:
Yep, it'd be great if you'd rebase that PR and address the comments smarnarch left! AFAICT the test I've added in this PR would still fail on that branch, though, so to fix the immediate problem a user is experiencing, that PR won't help.

carols10cents:
I'm happy to squash the format commit in before merging, but didn't want to mess up the test result demonstrations I already did. This is ready for review now; see email for context.

jtgeibel:
Okay, I finally understand the subset of accounts this applies to and how it relates to the other PR.  This looks good to me.

@bors r+

PR #2184, opened 2020-02-13
Revert status badge to travis

JohnTitor:
~~Prevents unexpected failures from being displayed.~~
Since we don't gate on GH Actions currently, so display the Travis badge until we resolve everything in #2059.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # crates.io
 
-[![Build Status](https://github.com/rust-lang/crates.io/workflows/CI/badge.svg)](https://github.com/rust-lang/crates.io/actions?query=workflow%3ACI)
+[![Build Status](https://travis-ci.com/rust-lang/crates.io.svg?branch=master)](https://travis-ci.com/rust-lang/crates.io)
 [![What's Deployed](https://img.shields.io/badge/whatsdeployed-prod-green.svg)](https://whatsdeployed.io/s-9IG)
 
 Source code for the default [Cargo](http://doc.crates.io) registry. Viewable


README.md
```diff
@@ -1,6 +1,6 @@
 # crates.io
 
-[![Build Status](https://github.com/rust-lang/crates.io/workflows/CI/badge.svg)](https://github.com/rust-lang/crates.io/actions?query=workflow%3ACI)
+[![Build Status](https://github.com/rust-lang/crates.io/workflows/CI/badge.svg?branch=auto&event=push)](https://github.com/rust-lang/crates.io/actions?query=workflow%3ACI)
```
jtgeibel: This should probably point to the `master` branch (which should be the default), since bors does its testing on `auto`.

JohnTitor: But we don't run any Actions on master, right? Actually we use same queries in Clippy: https://github.com/rust-lang/rust-clippy/commit/fb59043621f9297d641b16d3a7151c69e70f515e

jtgeibel: We do run against master, but in the Actions UI it shows as `auto` (with a link for "other workflows run on this commit").  The commit message you link to seems to indicate that this isn't just an issue with the UI, but also with the status badges.  Do you know if we've actually seen this report incorrect status for our repo in the past?  It seems like the downside to this change is that we will show a build failure if a bors merge fails on auto and thus never makes it to master.

Since we don't gate on GH Actions currently, maybe it makes sense to revert back to displaying the Travis badge until we resolve everything in #2059.  (I'll add a note about this there as well.)

JohnTitor: > Do you know if we've actually seen this report incorrect status for our repo in the past?

IIRC default link sometimes shows unexpected workflows but I have no reference :/

> Since we don't gate on GH Actions currently, maybe it makes sense to revert back to displaying the Travis badge until we resolve everything in #2059. (I'll add a note about this there as well.)

I feel it makes sense. Would you mind my doing so in this PR?

jtgeibel: > I feel it makes sense. Would you mind my doing so in this PR?

Sounds fine to me :+1:

JohnTitor: Done!

jtgeibel:
Great!

@bors r+

PR #2186, opened 2020-02-13
Replace `tempdir` with `tempfile`

JohnTitor:
replaces `tempdir` which is deprecated with `tempfile`


jtgeibel:
@JohnTitor thanks for the PR!  Would you drop the last 2 commits of this series and scope this to just on the change to `tempfile`?  Unfortunately we cannot merge a full-scale update to the lockfile at this time.

r? @jtgeibel

JohnTitor:
Sure, dropped!

locks:
Thank you!

@bors r+

PR #2187, opened 2020-02-13
Remove use of private items

JohnTitor:
`serde_json` tweaks the visibility of `ErrorCode` and `syntax()` so they are not available with further version(1.0.45). I don't update deps at this point but this tweak should be useful for further work.


jtgeibel:
Thanks for resurrecting this commit from your other PR.

@bors r+

PR #2190, opened 2020-02-14
Remove unused `version_authors.user_id`

### LINKED ISSUE #2056: Clean up the unused version_authors.user_id column
We can consider removing it from the database schema.

JohnTitor:
Fixes #2056

r? @carols10cents


Turbo87 (approved):

src/controllers/version/metadata.rs
```diff
@@ -43,20 +43,15 @@ pub fn authors(req: &mut dyn Request) -> AppResult<Response> {
...
     struct R {
-        users: Vec<EncodablePublicUser>,
         meta: Meta,
     }
     #[derive(Serialize)]
     struct Meta {
         names: Vec<String>,
     }
     Ok(req.json(&R {
-        users: vec![],
```
Turbo87, line 59: @JohnTitor we discussed this in the team meeting yesterday. we would prefer for this to stay in the API response for now to avoid potentially breaking changes. the database changes are fine, but it would be great it you could remove this API change from the PR :)

Turbo87:
I've rebased this and removed the commit that modified the API response. This should be good to go now. 

r? @jtgeibel

jtgeibel:
Thanks for the PR @JohnTitor and for the rebase @Turbo87. I've confirmed on staging and production that `select count(*) from version_authors where user_id is not null;` is 0.

@bors r+

PR #2192, opened 2020-02-15
Revert changes to update_downloads task in #2157

jtgeibel:
It appears that the additional wrapper transaction added around the
update_downloads task causes delays and timeouts to download requests
whenever the background job is run.  Reverting so that master can be
deployed.

r? @ghost

### DIFF
--- a/src/tasks.rs
+++ b/src/tasks.rs
@@ -3,8 +3,3 @@ mod update_downloads;
 
 pub use dump_db::dump_db;
 pub use update_downloads::update_downloads;
-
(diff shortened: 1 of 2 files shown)


jtgeibel:
After observing this PR in production, the revert solves the slow and timed-out download requests that were observed after the last deploy.

@bors r+

jtgeibel:
Travis is doing that thing again, where it waits just long enough before spawning jobs so that the builds finish just moments after the timeout.

@bors retry

PR #2193, opened 2020-02-16
Make crate stats look more consistent

azolotko:
Screenshot 2020-02-16 at 19 20 05


locks:
Thank you for the help!
@bors r+

PR #2194, opened 2020-02-17
Make owner invites case-insensitive

### LINKED ISSUE #2081: Owner invite doesn't work when username is specified with different casing
I'm trying to invite https://crates.io/users/matherunner to my crate https://crates.io/crates/hltas. I tried adding `Matherunner`, `matherunner` as well as his email address, multiple times, always getting "An invite has been sent to matherunner". However, he never got any emails and his owner invites are empty.

## Implementation instructions

- [Start from this branch](https://github.com/rust-lang/crates.io/tree/failing-test-2081) that has a test that fails because the username of the invited owner is specified in a different case than it is in the database.
- Get the test to pass by changing [this `filter`](https://github.com/rust-lang/crates.io/blob/7020cb2218b0458f97a03757edd167be4d8097cf/src/models/owner.rs#L74) to be case insensitive using the `crate::lower` function like [this code does](https://github.com/rust-lang/crates.io/blob/7020cb2218b0458f97a03757edd167be4d8097cf/src/controllers/user/other.rs#L11-L14) (I'm not sure why that code chooses to lowercase the provided name in Rust and then use the SQL `crate::lower` function on the `gh_login`... I think both should be done with SQL in this case?) We have [an index on `lower(gh_login)`](https://github.com/rust-lang/crates.io/blob/7020cb2218b0458f97a03757edd167be4d8097cf/migrations/2017-10-09-135625_add_lower_gh_login_index/up.sql#L2) so this query will be fast.

JohnTitor:
Fixes #2081

r? @carols10cents

### DIFF
--- a/src/controllers/user/other.rs
+++ b/src/controllers/user/other.rs
@@ -9,7 +9,7 @@ use crate::views::EncodablePublicUser;
 pub fn show(req: &mut dyn RequestExt) -> EndpointResult {
     use self::users::dsl::{gh_login, id, users};
 
-    let name = &req.params()["user_id"].to_lowercase();
+    let name = crate::lower(&req.params()["user_id"]);
     let conn = req.db_conn()?;
     let user: User = users
         .filter(crate::lower(gh_login).eq(name))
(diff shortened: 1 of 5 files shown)


Turbo87:
LGTM 👍 

@bors r+

PR #2205, opened 2020-02-20
Introduce new lockfile format

JohnTitor:
Introduce new lockfile format, as suggested in #2195.

r? @jtgeibel


jtgeibel:
Looks good to me!  I'm going to hold off on merging for a day or 2 just in case we need to revert #1864 for some reason.  Current master is deployed to staging and I don't expect any issues.  I should be able to deploy that to production after work tomorrow and expect to r+ this a few hours after that.

PR #2207, opened 2020-02-20
Don't return `Ok` if update handler isn't run

sgrif:
We're losing jobs to publish crates. It's not clear why. The requests
are returning 200, which implies they are being queued correctly. There
is no log to indicate that these jobs failed, so right now I believe
that this job is erroneously returning `Ok` when it failed. Given GitHub
is having a partial outage, that seems to reinforce this theory.

From looking at the code, this looks like the only way we could have
returned `Ok` when the publish failed. I believe we got some sort of
error response (or lack of one) from GitHub that caused this callback
not to be run.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -203,6 +203,7 @@ impl Repository {
 
         // git push
         let mut ref_status = Ok(());
+        let mut callback_called = false;
         {
             let mut origin = self.repository.find_remote("origin")?;
             let mut callbacks = git2::RemoteCallbacks::new();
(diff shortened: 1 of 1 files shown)


src/git.rs
```diff
@@ -212,7 +212,9 @@ impl Repository {
             callbacks.push_update_reference(|refname, status| {
                 assert_eq!(refname, "refs/heads/master");
                 if let Some(s) = status {
-                    ref_status = Err(format!("failed to push a ref: {}", s).into())
+                    ref_status = Some(Err(format!("failed to push a ref: {}", s).into()))
+                } else {
+                    ref_status = Some(Ok(()));
```
pietroalbini: I think this could still mask some errors.

[The documentation of git2](https://docs.rs/git2/0.11.0/git2/struct.RemoteCallbacks.html#method.push_update_reference) says *"Set a callback to get invoked for **each** updated reference on a push."* (emphasis mine): if the first ref has an error while the second ref was pushed successfully, this would override the error of the first ref.

PR #2208, opened 2020-02-21
Bump to alpha releases of conduit-* crates

jtgeibel:
These upstream packages have been updated to prune old dependencies and
include no functional changes at this time.  This PR removes 14
dependencies.

The updated version of `civet` is based on the 0.9 release we have been
pinned to.  The 0.12 release abandons the attempts to update to the
underlying library which resulted in build failures on macOS for the
0.10 and 0.11 releases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,22 +108,12 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96434f987501f0ed4eb336a411e0631ecd1afa11574fe148587adc4ff96143c9"
-dependencies = [
- "byteorder",
- "safemem 0.2.0",
-]
-
-[[package]]
-name = "base64"
 version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


JohnTitor:
Btw, this also looks good to me, feel free to r=me if you want.

jtgeibel:
@JohnTitor do you have macOS to confirm this builds/tests there?  If not, hopefully @carols10cents can give it a quick test (although I'm pretty confident I dropped the problematic changes from upstream master).

Note to self: When deploying this to staging, I want to make sure I'm signed-in and then deploy.  Then ensure the existing cookie can be read by `cookie 0.12`.  Then test logout and login.  I think breaking changes are unlikely here, but want to be safe when it comes to session management.

JohnTitor:
Ah, sorry I missed that! Let's wait then.

jtgeibel:
@carols10cents can you verify this PR still builds for you on macOS?  I don't expect any problems, but I want to verify before possibly breaking your build.

jtgeibel:
I've rebased to resolve conflicts.  I'm fairly confident I didn't break things on macOS and I'd like to drop these old dependencies so I'm going to merge.

@bors r+

PR #2209, opened 2020-02-21
Refactor `RlDropdown` components

Turbo87:
This PR applies a bunch of refactorings to the `RlDropdown` components. Ultimately, this removes their reliance on jQuery and replaces it with a dependency on `ember-click-outside` and the already existing `ember-keyboard` dependency.

r? @locks


Windvis:
This is probably a silly question (I have no context about this project 😄), but can't this be replaced with ember-basic-dropdown instead?

Turbo87:
yeah, it could. but I've tried and it's significantly more work due to styling issues and some other incompatibilities. 😅

this PR brings the api closer to the one of basic dropdown though, so it will be easier to switch in the future.

PR #2211, opened 2020-02-22
Revert #2201 to fix broken new token button

jtgeibel:
I've bisected the issue with the broken "New Token" button to this merge.  What is very strange is that the original commit 28a03bfd that implemented the change does not appear to have this issue.  Therefore it seems the issue is some sort of conflict between multiple pull requests somewhere in https://github.com/rust-lang/crates.io/compare/c970817d...663259cb.

I'm deploying to staging now and will deploy to production as soon as I test the change there.

cc @locks
r? @ghost

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36425,9 +36425,9 @@
       }
     },
     "ember-source": {
-      "version": "3.16.2",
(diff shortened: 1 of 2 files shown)


jtgeibel:
Works fine on staging.  Deploy to production in process.

@bors r+

PR #2213, opened 2020-02-22
Aggressively update conduit-hyper and reqwest

jtgeibel:
This series does an aggressive update for `conduit-hyper` and then `reqwest`.  This portion of the dependency tree has been updated recently and we aren't using `conduit-hyper` in production yet, so this should be fairly low risk even though it bumps a lot of crates.

The commit messages show the detailed output of each command.

cc #1805
r? @JohnTitor


jtgeibel:
> I would r+ if I could.

I think I can set r=you for you.  Let's see.

@bors r=JohnTitor

I plan to deploy this today and monitor for any issues.

PR #2215, opened 2020-02-24
Add backend proxy to aid local fastboot testing

jtgeibel:
To avoid the need to set up a local nginx configuration during
development, this commit adds a simple proxy for frontend requests when
`USE_FASTBOOT=staging-experimental` is set.

When enabled, developers should browse to http://localhost:8888/
(instead of port 4200 when using `npm run start:local`).


jtgeibel:
I've rebased onto the new http types, which simplified this logic a lot.  This is still intended for local development only.

r? @JohnTitor

PR #2223, opened 2020-02-25
Update `parking_lot` to 0.10

JohnTitor:
Update `parking_lot` to 0.10 (and `r2d2` to 0.8.8 from 0.8.6). This leaves only one version of `parking_lot` in the dependencies and removes some packages.

r? @jtgeibel


jtgeibel:
This is great, it looks like this drops at least 10 old dependencies!

@bors r+

PR #2224, opened 2020-02-25
Remove jQuery

Turbo87:
Following #2209 there does not seem to be any code left that relies on jQuery being available. This PR removes jQuery from the build which leads to 30kB (~10%; after gzip) less JS code shipped to the clients.

@jtgeibel @carols10cents we might want to do a round of QA on the staging server for this to ensure that I haven't missed anything 🙏


Turbo87:
hmm, looks like I got a test failure to solve first... 😅

Turbo87:
the failing CI test was caused by Ember Data using `fetch()` instead of `$.ajax()` now, and having slightly different error handling code in this case. Previously it would automatically try to parse the error response as JSON, but with the `fetch()` implementation it no longer does that.

I've adjusted the `application` adapter to bring that functionality back for now :)

PR #2225, opened 2020-02-25
Use `Vec::with_capacity()` as possible

JohnTitor:
It's a minor cleanup, uses `Vec::with_capacity()` as possible to prevent reallocation.


src/controllers/krate/publish.rs
```diff
@@ -245,7 +245,9 @@ fn parse_new_headers(req: &mut dyn Request) -> AppResult<EncodableCrateUpload> {
     fn empty(s: Option<&String>) -> bool {
         s.map_or(true, String::is_empty)
     }
-    let mut missing = Vec::new();
+
+    // It can have up to three elements as below conditions.
```
aclonegeek:
```suggestion
    // It can have up to three elements per below conditions.
```

jtgeibel:
LGTM!

@bors r+

Issue #2226, opened 2020-02-26
Add log message at the start of the index update process


pietroalbini:
Currently the background job for pushing a new crate to the git index doesn't log anything, making it harder to investigate outages and incidents affecting git. We should add a log message at the start of the process, showing the crate name.

jtgeibel:
I believe this is resolved by #2231.  We now log the crate name and version both before and after updating the index for all index operations (publish, yank, and unyank).

PR #2231, opened 2020-02-26
Add a log message to add_crate background job

rye:
Per #2226; is this sufficient to resolve that issue?

This PR simply adds a log message (`println!`) right before carrying out the final `commit_and_push` operation, which could otherwise fail silently if the Git host is experiencing an outage.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -267,12 +267,11 @@ pub fn add_crate(env: &Environment, krate: Crate) -> Result<(), PerformError> {
     serde_json::to_writer(&mut file, &krate)?;
     file.write_all(b"\n")?;
 
-    println!("Updating crate `{}#{}`", krate.name, krate.vers);
+    let message: String = format!("Updating crate `{}#{}`", krate.name, krate.vers);
 
-    repo.commit_and_push(
-        &format!("Updating crate `{}#{}`", krate.name, krate.vers),
-        &repo.relative_index_file(&krate.name),
-    )
+    println!("{}", message);
+
+    repo.commit_and_push(&message, &repo.relative_index_file(&krate.name))
 }
 
 /// Yanks or unyanks a crate version. This requires finding the index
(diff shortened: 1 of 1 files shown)


src/git.rs
```diff
@@ -267,6 +267,8 @@ pub fn add_crate(env: &Environment, krate: Crate) -> Result<(), PerformError> {
     serde_json::to_writer(&mut file, &krate)?;
     file.write_all(b"\n")?;
 
+    println!("Updating crate `{}#{}`", krate.name, krate.vers);
```
jtgeibel: Let's use `format!` here to get a `String`, and then use that to print and in the call to `commit_and_push` below.

rye:
I think I may just try moving the `println!` statement(s) _into_ the `commit_and_push` function, since in all invocations we pass a Git commit message that describes what the commit does. Then we don't need a wrapper function; we just print out one (or two) messages containing the commit message and that'll probably be sufficient.

src/git.rs
```diff
@@ -227,6 +231,8 @@ impl Repository {
             ref_status = Err("update_reference callback was not called".into());
         }
 
+        println!("commit_and_push: Finished with \"{}\"", msg);
```
jtgeibel: I think this final step should still be extracted into a wrapper function.  If any of the early returns or `?`s produce an error, then nothing will be logged.  I think a wrapper like the following would be an easy way to address that:

```rust
fn commit_and_push(&self, msg: &str, modified_file: &Path) -> Result<(), PerformError> {
    self.commit_and_push_inner(msg, modified_file)
        .map(|_| println!("commit_and_push: Finished with \"{}\"", msg))
        .map_err(|err| {
            println!("commit_and_push: Finished with \"{}\"", err);
            err
        })
}
```

src/git.rs
```diff
@@ -201,6 +203,8 @@ impl Repository {
         self.repository
             .commit(Some("HEAD"), &sig, &sig, &msg, &tree, &[&parent])?;
 
+        println!("commit_and_push: Pushing \"{}\"", msg);
```
jtgeibel: I think we can drop this output, and have just 2 log entries generated per operation.  One at the beginning and a success or failure at the end.

src/git.rs
rye: I think I've settled on a good solution very similar to this, but with the initial println! also in the wrapper method and slightly different naming.

jtgeibel:
LGTM, thanks!

@bors r+

PR #2234, opened 2020-02-29
Include crate name and version in the publish API's log entries

### LINKED ISSUE #2228: Include the crate name in the publish API's log entry
Currently the log entry for a newly published crate looks like this:

```
Feb 30 25:61:61 crates-io app/web.1 at=info method=PUT path="/api/v1/crates/new" request_id=GGGGGGGG-GGGG-GGGG-GGGG-GGGGGGGGGGGG fwd="999.999.999.999" service=0ms status=200 user_agent="cargo 2.0.0 (000000000 1970-01-01)" metadata_length=1234
```

Such entry lacks the crate name though, which makes investigating which crate that call added much harder than necessary. We should add the crate name *in the same log entry*.

pietroalbini:
This PR includes the crate name and version in the log entries of the publish API, to fix #2228 and to aid debugging issues in production. This is done in three steps.

First, the log entry generation code was refactored to easily allow adding new fields to it. Now, instead of hardcoding the list of fields in a `format!` macro, there is a builder with methods to add individual fields.

~~Then, a way to attach custom log fields to a response was added. This is implemented by setting HTTP headers in the response starting with `X-CratesIo-Log-Metadata-`, and by changing the middleware responsible for request logging to strip those headers and add their content to the log line instead.~~

Then, a way to attach custom log fields to a response was added. This is implemented as a "request extension", with a convenience method in the `ResponseExt` trait to add the custom log fields to all responses generated by that request.

Finally, this new feature was used to attach the crate name and version to the publish API's log entries.

r? @sgrif

### DIFF
--- a/src/controllers.rs
+++ b/src/controllers.rs
@@ -38,6 +38,8 @@ mod prelude {
         fn query(&self) -> IndexMap<String, String>;
         fn wants_json(&self) -> bool;
         fn query_with_params(&self, params: IndexMap<String, String>) -> String;
+
+        fn log_metadata<V: std::fmt::Display>(&mut self, key: &'static str, value: V);
     }
 
     impl<'a> RequestUtils for dyn Request + 'a {
@@ -75,6 +77,10 @@ mod prelude {
                 .extend_pairs(params)
                 .finish();
             format!("?{}", query_string)
+        }
+
+        fn log_metadata<V: std::fmt::Display>(&mut self, key: &'static str, value: V) {
+            crate::middleware::log_request::add_custom_metadata(self, key, value);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


jtgeibel:
Looks like a great approach overall to me @pietroalbini.  The only suggestion I have (and this seems kind of backwards but I think it makes sense) is that instead of using phantom `Response` headers, it may be easier to attach the metadata to the extensions `TypeMap` in the `Request` which will propagate back up the middleware stack.

Then there could either be a wrapper type for each field, or a `LoggingMetadata(HashMap<&'str, String>)` that fields get added to.  It looks like this is already how the "metadata_length" is propograted (although it is less clear because it just uses a bare `u64` type).

This change would allow the metadata to be captured and attached to the request early in an endpoint.  That way if the endpoint returns early due to an error then the logging metadata has already been captured and will not be lost.  I think the current implementation will not log the additional metadata if the publish fails early with a user-facing or internal server error.

pietroalbini:
That's a great point!

pietroalbini:
@jtgeibel implemented the change!

r? @jtgeibel

jtgeibel:
This looks great!

@bors r+

PR #2235, opened 2020-03-01
Log inner chained errors

jtgeibel:
This builds on the logging infrastructure recently added by @pietroalbini.  Whenever a user facing error is chained around an internal error, the router now captures that message and adds it to the log metadata.

In particular, this should allow us to identify errors sent to cargo via `cargo_err` which are sent via a status 200 response and left no indication in the logs that an error has occurred.

r? @pietroalbini

### DIFF
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -98,10 +98,6 @@ impl Display for RequestLine<'_> {
             }
         }
 
-        if let Some(len) = self.req.extensions().find::<u64>() {
-            line.add_field("metadata_length", len)?;
-        }
-
         if let Err(err) = self.res {
             line.add_quoted_field("error", err)?;
         }
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -58,6 +58,16 @@ pub fn add_custom_metadata<V: Display>(req: &mut dyn Request, key: &'static str,
     }
 }
 
+#[cfg(test)]
+pub(crate) fn get_log_message(req: &dyn Request, key: &'static str) -> String {
+    for (k, v) in &req.extensions().find::<CustomMetadata>().unwrap().entries {
+        if key == *k {
+            return v.clone();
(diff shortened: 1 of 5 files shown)


src/middleware/log_request.rs
```diff
@@ -58,6 +58,16 @@ pub fn add_custom_metadata<V: Display>(req: &mut dyn Request, key: &'static str,
...
 }
 
+#[cfg(test)]
+pub(crate) fn get_log_message(req: &dyn Request, key: &'static str) -> String {
+    for (k, v) in &req.extensions().find::<CustomMetadata>().unwrap().entries {
+        if key == *k {
+            return v.clone();
+        }
+    }
+    String::new()
```
pietroalbini: Why a `String::new()`? Since this is for tests, if the key is missing I'd expect a panic.

src/router.rs
```diff
@@ -143,10 +143,15 @@ impl Handler for C {
         let C(f) = *self;
         match f(req) {
             Ok(resp) => Ok(resp),
-            Err(e) => match e.response() {
-                Some(response) => Ok(response),
-                None => Err(std_error(e)),
-            },
+            Err(e) => {
+                if let Some(cause) = e.cause() {
+                    req.log_metadata("cause", cause.to_string())
```
pietroalbini: This will not log nested causes (like error -> cause -> cause). Is this something we expect to happen?

src/middleware/log_request.rs
jtgeibel: Good suggestion.  I've been working on some upstream `conduit` changes and have been thinking in production mode for a lot of that.  I've converted this to a panic.

src/router.rs
jtgeibel: This actually already works!  I've added a commit to demonstrate that in the test.

pietroalbini:
Looks great!

@bors r+

PR #2236, opened 2020-03-01
Update `env_logger` to 0.7

JohnTitor:
This removes one dependency (`wincolor`), there should be no fallout.


jtgeibel:
Thanks!

@bors r+

PR #2237, opened 2020-03-01
Clarify messages for index commits and logs

jtgeibel:
Align with the version syntax used by cargo on the command line.

r? @pietroalbini


pietroalbini:
Is there a reason for this change? I'm worried someone might rely on the commit messages.

jtgeibel:
There is nothing specific driving this change, just a minor nit that we're expanding to our logs as well.  I hope no one relies on the index commit messages for anything, but if someone only needs to follow newly published versions then it's certainly possible.

I'm fine with closing this PR if you think that's best.  I'd say this is a low-risk, low-reward change.

pietroalbini:
I don't think it's worth the potential breakage.

PR #2246, opened 2020-03-04
Run `cargo update -p web-sys --aggressive`

JohnTitor:
Mainly this updates wasm related packages and removes some packages.

r? @jtgeibel


jtgeibel:
This bump to `web-sys` is a great find!  This has minimal overlap with production.

Would you mind including the output of `cargo update` in the commit message for future PRs?  The downside of the new lockfile format is that it isn't as easy to see a glace (at the end of the old format) what the changes are.  It's easy to forget to capture it, but for aggressive updates it is especially nice.

@bors r+

JohnTitor:
> Would you mind including the output of cargo update in the commit message for future PRs?

Sure thing, next time!

PR #2250, opened 2020-03-05
Pin quote version to avoid compilation failure

JohnTitor:
quote modified private-ish API in v1.0.3 but failure_derive depends on that therefore compilation will fail when we update quote v1. This avoids that failure and accidental updates by pinning quote v1 version.
ref: https://users.rust-lang.org/t/failure-derive-compilation-error/39062

r? @jtgeibel


JohnTitor:
Now, that issue has been resolved, closing.

Issue #1617, opened 2019-02-07
Show github teams in "Manage owners" view


killercup:
Currently, the "manage ownership" view doesn't show the Github teams that own the crate.

For example, if I go to <https://crates.io/crates/rustfix>, I can see `github:rust-lang-nursery:rustfix` as an owner of the crate (as well as Oli and me). If I click on "Manage owners" (the `/crates/rustfix/owners` route), the Github team isn't in the list. Additionally, the input field on that page is labelled "username", so I can't tell if adding teams is possible (e.g. using the same syntax as in `cargo owner --add github:rust-lang-nursery:rustfix`).

frewsxcv:
this appears to be resolved https://crates.io/crates/rustfix/owners

PR #2253, opened 2020-03-06
Display teams on the crate owners page

RobbieClarken:
This change updates the crate owners page (`/crates/<crate-name>/owners`) to display teams which are set to be crate owner. It also adds a "Remove" button for each team which allows removing teams from owning the crate in the same way that user type owners can be removed.

I have displayed the team as `<organisation>/<team>` rather than `github:<organisation>:<team>` because I think it looks cleaner (eg `rust-lang/core` vs `github:rust-lang:core`). Happy to change it if people disagree.

Screen Shot 2020-03-06 at 5 17 50 pm

I haven't touched adding teams via the owners page in this PR. From what I can tell, it is currently possible although the message that is returned is misleading. I think that should be tackled in a separate PR.

Addresses #1617.

### DIFF
--- a/app/models/team.js
+++ b/app/models/team.js
@@ -14,7 +14,7 @@ export default Model.extend({
     let login_split = login.split(':');
     return login_split[1];
   }),
-  display_name: computed('org_name', function() {
+  display_name: computed('name', 'org_name', function() {
     return `${this.org_name}/${this.name}`;
   }),
 });
--- a/app/models/team.js
+++ b/app/models/team.js
@@ -15,6 +15,7 @@ export default Model.extend({
     return login_split[1];
   }),
   display_name: computed('name', 'org_name', function() {
-    return `${this.org_name}/${this.name}`;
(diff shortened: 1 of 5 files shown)


smarnach:
@locks Could you take this review? I'm too slow at frontend reviews to find time for them at the moment. For what it's worth, in principle I'm in favour of adding this feature.

app/models/team.js
```diff
@@ -14,4 +14,7 @@ export default Model.extend({
     let login_split = login.split(':');
     return login_split[1];
   }),
+  display_name: computed('org_name', function() {
```
locks:
```suggestion
  display_name: computed('name', 'org_name', function() {
```

app/models/team.js
```diff
@@ -14,4 +14,7 @@ export default Model.extend({
...
+    return `${this.org_name}/${this.name}`;
```
locks:
```suggestion
    let { name, org_name } = this.getProperties('name', 'org_name');
    return `${org_name}/${name}`;
```

PR #2256, opened 2020-03-06
Update hex to 0.4

JohnTitor:
Update hex to 0.4, API is changed, no extra dependencies change.
/cc #1265
r? @jtgeibel


locks:
Neat cleanup!
@bors r+

Issue #2252, opened 2020-03-06
Error accessing dashboard


dbr:
I logged in, and tried to click on Dashboard link under my username. It directs to
https://crates.io/me
and gives the following error. Same happens with "Owner Invites" page in same menu, but Account Settings works

```
Ember Data Request GET /api/v1/crates returned a 403 Payload (application/json; charset=utf-8) [object Object]

  r@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3958:19
r@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3961:100
handleResponse@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:4027:17
handleResponse@https://crates.io/assets/cargo-122a2a85ee957cc7845ce8a377e9170f.js:6:340
r@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:1925:9
o@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3997:14
ajax/</<@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:4033:8
ajax/<@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:4033:19
_@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3640:12
y@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3638:128
invoke@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3219:139
flush@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3211:74
flush@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3222:292
_end@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3273:9
V/this._boundAutorunEnd@https://crates.io/assets/vendor-d56f71d6bbd0a4e94bf420b72ba26f0e.js:3231:605
```

jtgeibel:
[...]

PR #2258, opened 2020-03-07
Add some error chaining to log more on auth failure

jtgeibel:
cc #2252
r? @pietroalbini

### DIFF
--- a/src/middleware/block_traffic.rs
+++ b/src/middleware/block_traffic.rs
@@ -47,7 +47,8 @@ impl Handler for BlockTraffic {
             .iter()
             .any(|value| self.blocked_values.iter().any(|v| v == value));
         if has_blocked_value {
-            super::log_request::add_custom_metadata(req, "cause", "blocked");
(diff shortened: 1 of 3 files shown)


src/middleware/block_traffic.rs
```diff
@@ -47,6 +47,7 @@ impl Handler for BlockTraffic {
             .iter()
             .any(|value| self.blocked_values.iter().any(|v| v == value));
         if has_blocked_value {
+            super::log_request::add_custom_metadata(req, "cause", "blocked");
```
pietroalbini: I would rename the cause to `blacklisted header`, and add `header_name` and `header_value`.

jtgeibel: I've added the `header_name` to the message.  I didn't add the header value to the message, because capturing that would make the code more complicated at least for now all the headers we block on should already be captured in the logs.

pietroalbini:
@bors r+

Thanks!

PR #2266, opened 2020-03-10
Ensure the update_downloads job doesn't run concurrently

jtgeibel:
This is an improved implementation of #2157.  The previous design
relied on a transaction based lock to manage the lifetime of the lock.
The wrapper transaction caused the `update_downloads` job to interfere
with incoming download requests, and the changes had to be reverted.

This implementation uses a session lock which is automatically released
even if the callback panics.

r? @sgrif

### DIFF
--- a/src/tasks/util/advisory_lock.rs
+++ b/src/tasks/util/advisory_lock.rs
@@ -1,0 +1,102 @@
+use std::error::Error;
+
+use diesel::prelude::*;
+use diesel::sql_types::BigInt;
+use diesel::PgConnection;
+
+sql_function!(fn pg_try_advisory_lock(key: BigInt) -> Bool);
+sql_function!(fn pg_advisory_unlock(key: BigInt) -> Bool);
+
+/// Run the callback if the session advisory lock for the given key can be obtained.
+///
+/// If the lock is not already held, the callback will be called and the lock will be unlocked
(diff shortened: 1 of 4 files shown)


src/tasks/util/advisory_lock.rs
```diff
@@ -0,0 +1,102 @@
...
+    conn: &'a PgConnection,
+    key: i64,
+}
+
+impl<'a> Drop for DropGuard<'a> {
+    fn drop(&mut self) {
+        match diesel::select(pg_advisory_unlock(self.key)).get_result(self.conn) {
+            Ok(true) => (),
+            Ok(false) => println!(
+                "Error: job advisory lock for key {} was not locked",
```
jtgeibel: Maybe this should be a panic, to forbid the callback from releasing the lock on its own.

src/tasks/util/advisory_lock.rs
```diff
@@ -0,0 +1,102 @@
...
+                self.key
+            ),
+            Err(err) => println!("Error unlocking advisory lock (key: {}): {}", self.key, err),
```
jtgeibel: Similar here.  We could do something (or just panic) to ensure the connection is closed and the lock unlocked.

jtgeibel:
Closing as superseded by #2539. We've now seen that fix work correctly several times in production.

PR #2267, opened 2020-03-10
Create a new DB pool when rebuilding the worker

sgrif:
When we fail to *start* (not run) new jobs, which could happen because
all our worker threads are saturated by a job that is taking too long,
we will rebuild the runner up to 5 times in an attempt to recover before
killing the process. Importantly, if this occurred because of a slow
job, this should give those slow jobs a chance to finish without
blocking the rest of the queue.

However, we were sharing a database pool even when we rebuilt the
runner. This means that if the slow jobs held a database connection,
our retries will just fail immedaitely, and we kill the process (along
with any hopes of the slow job catching up). This creates a new
connection pool whenever we restart the runner.

r? @jtgeibel

### DIFF
--- a/src/background_jobs.rs
+++ b/src/background_jobs.rs
@@ -50,8 +50,22 @@ impl Environment {
         uploader: Uploader,
         http_client: Client,
     ) -> Self {
+        Self::new_shared(
+            Arc::new(Mutex::new(index)),
+            connection_pool,
+            uploader,
+            http_client,
+        )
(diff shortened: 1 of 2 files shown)


jtgeibel:
I cherry-picked the commit on top of current production (556e2b75) to avoid pulling in other changes from master.  I'll deploy this (4c5fa5b, pre-merge) soon.

@bors r+

PR #2269, opened 2020-03-10
Update swirl

sgrif:
This removes some of the boilerplate we had to add last night for
manually managing connection pools with swirl. Swirl now can
automatically manage its own connection pool. It can be given either a
database URL, a pool builder, or a fully built pool (which is used in
tests). If it's given one of the first two, it will configure the
connection pool to have a size of 2x the thread count. I've also let it
go up to the default number of threads, which is the number of CPU
cores. Since this would mean the pool size grows pretty dramatically,
I've configured it to not try to keep any idle connections (meaning it
will try to establish the connection when asked for one, not before).

Additionally, we no longer need the connection pool to live on our
environment, we can take either a connection or a pool as an argument to
a background job, and swirl will automatically take it from the pool. In
the future I'm going to try and have swirl make some more intelligent
decisions on this, so we could do something like have 2 threads only run
jobs that don't need a connection (like publish!)

Heroku doesn't guarantee the number of logical cores on standard dynos,
but it appears to have consistently been 4 for a while. This means we'll
be doubling the number of background workers, and potentially using up
to 8 database connections on worker. Now that we've moved some web
traffic to the replica, we may want to lower the pool size for the
primary db on the web servers to compensate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2479,11 +2479,12 @@ checksum = "032c03039aae92b350aad2e3779c352e104d919cb192ba2fabbd7b831ce4f0f6"
 [[package]]
 name = "swirl"
 version = "0.1.0"
-source = "git+https://github.com/sgrif/swirl.git?rev=de5d8bb#de5d8bbcfbf7878aa2c248c105938567b32ed8a8"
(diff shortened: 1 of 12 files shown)


jtgeibel:
LGTM

@bors r+

PR #2291, opened 2020-03-21
Migrate routes to native classes

locks:


app/routes/application.js
```diff
@@ -1,17 +1,17 @@
...
-  },
+  }
 
-  actions: {
-    didTransition() {
-      this.flashMessages.step();
-    },
-  },
-});
+  @action
```
Turbo87: I'm a little worried about this because I've had issues with the `@action` decorator on routes in the past 🤔

locks: Do you remember what the issue was? I tried checking the literature and this seemed the correct way to upgrade :/

Turbo87: nope, unfortunately I don't remember the specifics, but I'd sleep better if I knew that we had better test coverage for these things :-/

PR #2304, opened 2020-03-22
Remove unused `id=crates` assignments

Turbo87:
There does not seems to be any code that relies on these IDs being set, so I think it should be save to remove them.

r? @locks

### DIFF
--- a/app/templates/category/index.hbs
+++ b/app/templates/category/index.hbs
@@ -87,7 +87,7 @@
   </div>
 </div>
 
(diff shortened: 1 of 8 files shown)


locks:
Hm, if someone was linking to `crates.io/crates/#crates` it will no longer scroll down, correct?

Turbo87:
yes, that's correct. though I wonder how many people are aware and using that id since it doesn't seem to be linked anywhere in our app.

PR #2316, opened 2020-03-24
version: Move `quick-links` CSS class into CSS module

Turbo87:
r? @locks

### DIFF
--- a/app/styles/crate/version.module.scss
+++ b/app/styles/crate/version.module.scss
@@ -5,6 +5,26 @@
     top: 4px;
 }
 
+.quick-links {
+    display: flex;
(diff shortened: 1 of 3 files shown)


app/styles/crate/version.module.scss
```diff
@@ -5,6 +5,26 @@
     top: 4px;
 }
 
+ul.quick-links {
```
locks: do you need the `ul` specifier?

Turbo87: not strictly necessary. do you want me to remove it?

Turbo87: removed

PR #2368, opened 2020-03-30
Extract `results-meta` CSS module classes

Turbo87:
... which replaces the `id="results"` assignments. While they all look similar there are sometimes slight differences. If it's considered worth it, we can try to deduplicate these styles at a later point.

r? @locks

### DIFF
--- a/app/styles/categories.module.scss
+++ b/app/styles/categories.module.scss
@@ -1,5 +1,12 @@
 /* Styles for the /categories page */
 
+.results-meta {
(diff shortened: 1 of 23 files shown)


locks:
It seems a lot of repetition of `.results-meta` declarations 🤔

Turbo87:
> While they all look similar there are sometimes slight differences. If it's considered worth it, we can try to deduplicate these styles at a later point.

😉

PR #2373, opened 2020-03-30
Remove unnecessary `this.get()` calls

Turbo87:
With the newer Ember.js versions this is no longer necessary in the majority of cases 🎉 

r? @locks

### DIFF
--- a/app/components/owned-crate-row.js
+++ b/app/components/owned-crate-row.js
@@ -13,7 +13,7 @@ export default Component.extend({
 
   actions: {
     toggleEmailNotifications() {
-      this.set('emailNotifications', !this.get('emailNotifications'));
+      this.set('emailNotifications', !this.emailNotifications);
     },
   },
 });
--- a/app/controllers/category/index.js
+++ b/app/controllers/category/index.js
@@ -21,7 +21,7 @@ export default Controller.extend(PaginationMixin, {
       return 'Alphabetical';
     } else if (this.sort === 'new') {
       return 'Newly Added';
-    } else if (this.get('sort') === 'recent-updates') {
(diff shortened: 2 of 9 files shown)


app/controllers/crate/version.js
```diff
@@ -34,7 +34,7 @@ export default Controller.extend({
     return this.get('crate.owner_user').findBy('id', this.get('session.currentUser.id'));
   }),
   notYankedOrIsOwner: computed('model', 'crate.owner_user', 'session.currentUser.id', function () {
-    return !this.get('model').yanked || this.get('crate.owner_user').findBy('id', this.get('session.currentUser.id'));
+    return !this.model.yanked || this.get('crate.owner_user').findBy('id', this.get('session.currentUser.id'));
```
locks: I tried tracing the logic and I think `this.model` comes from Ember Data so we might want to leave the `get` here.

app/controllers/me/index.js
```diff
@@ -22,7 +22,7 @@ export default Controller.extend({
   emailNotificationsSuccess: false,
 
   setAllEmailNotifications(value) {
-    this.get('ownedCrates').forEach(c => {
+    this.ownedCrates.forEach(c => {
```
locks: `ownedCrates` comes from `ownedCrates: alias('model.ownedCrates')`, no get necessary in this file?

app/controllers/crate/version.js
Turbo87: removed

app/controllers/me/index.js
Turbo87: removed for now

PR #2381, opened 2020-04-01
Convert promise chains to async/await

Turbo87:
async/await is significantly easier to reason about in JavaScript than Promise chains.

r? @locks


locks:
Agreed!
@bors r+

jtgeibel:
> Test was successful, but fast-forwarding failed: 422 Update is not a fast forward

I've never seen this before.  Lets try again.

@bors retry

PR #2383, opened 2020-04-02
Move request logging near the top of the middleware stack

jtgeibel:
By converting to a regular `Middleware` (instead of `AroundMiddleware`),
the logging can now sit near the top of the middleware stack.
Previously logging occurred at the top of the around stack, which sits
under the layers added via `add()`.

This change ensures that middleware sitting below the request logger
in the stack are able to add logging metadata in a call to `after()`.

r? @JohnTitor


JohnTitor:
Looks good! @bors r+

PR #2387, opened 2020-04-03
Set YAML indent size to 2

JohnTitor:
Current config makes my editor confused.


Turbo87:
looks like that matches all the YAML files that we have in this repo, so I guess that's a reasonable change 👍 

@bors r+

PR #2389, opened 2020-04-05
Use http types from `http` crate

jtgeibel:
In the upstream conduit crates, I've dropped the custom HTTP related types in favor of types from the `http` crate.  This reduces the amount of code to maintain upstream and generally provides a cleaner API, such as named `StatusCode::*` constants instead of naked integers.

This should also make it easier and more efficient to support both `civet` and `hyper` as web servers, as they (and our middleware stack) now use the same types and avoid unnecessary conversions.

r? @JohnTitor

### DIFF
--- a/src/tests/server.rs
+++ b/src/tests/server.rs
@@ -1,4 +1,4 @@
-use conduit::Method;
+use conduit::{header, Method};
 
 use crate::builders::*;
 use crate::util::*;
@@ -7,10 +7,10 @@ use crate::util::*;
 fn user_agent_is_required() {
     let (_app, anon) = TestApp::init().empty();
 
-    let mut req = anon.request_builder(Method::Get, "/api/v1/crates");
-    req.header("User-Agent", "");
+    let mut req = anon.request_builder(Method::GET, "/api/v1/crates");
+    req.header(header::USER_AGENT, "");
     let resp = anon.run::<()>(req);
-    resp.assert_status(403);
+    resp.assert_status(StatusCode::FORBIDDEN);
 }
 
 #[test]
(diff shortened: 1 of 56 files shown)


jtgeibel:
I'm also pushing this to staging for some testing there.  The deploy should finish within the next 15 minutes.  (The deployed SHA is available in the [site metadata](https://staging.crates.io/api/v1/site_metadata) which can be used to confirm staging still points to the latest commit in this PR, currently `5d0906e`.)

src/tests/server.rs
```diff
@@ -57,17 +57,25 @@ fn block_traffic_via_arbitrary_header_and_value() {
...
-    let mut req = anon.request_builder(Method::Get, "/api/v1/crates/dl_no_ua/0.99.0/download");
+    let mut req = anon.request_builder(Method::GET, "/api/v1/crates/dl_no_ua/0.99.0/download");
     // A request with a header value we want to block isn't allowed
-    req.header("User-Agent", "1");
-    req.header("X-Request-Id", "abcd"); // Needed for the error message we generate
+    req.header(header::USER_AGENT, "1");
+    req.header(
+        // Needed for the error message we generate
+        // FIXME: Simplify once we hit `conduit-test 0.9.0-alpha.2`
+        header::HeaderName::from_lowercase(b"x-request-id").unwrap(),
```
JohnTitor, line 66: It'd be nice the http crate had that header..

JohnTitor: FYI: https://github.com/hyperium/http/pull/411

jtgeibel: Awesome!  Thanks for opening the upstream PR.

PR #2400, opened 2020-04-09
crate.version: Simplify `model()` hook

Turbo87:
Best reviewed commit-by-commit 😅 

r? @locks

### DIFF
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -13,7 +13,7 @@ export default Route.extend({
   async model(params) {
     const requestedVersion = params.version_num;
     const crate = this.modelFor('crate');
-    const maxVersion = crate.get('max_version');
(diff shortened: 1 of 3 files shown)


app/routes/crate/version.js
```diff
@@ -13,7 +13,7 @@ export default Route.extend({
   async model(params) {
     const requestedVersion = params.version_num;
     const crate = this.modelFor('crate');
-    const maxVersion = crate.get('max_version');
+    const maxVersion = crate.max_version;
```
locks, line 16: I'd prefer to keep `.get` for (ED) models since they're proxy objects.

Turbo87: that's not quite correct AFAIK. async relationships are proxy objects, but models and their regular attributes are just regular things and don't need to use `.get()` anymore.

PR #2410, opened 2020-04-10
WIP: Use `ember-asset-size-action` to report asset size changes to PRs

Turbo87:
r? @locks

### DIFF
--- a/.github/workflows/asset-size.yml
+++ b/.github/workflows/asset-size.yml
@@ -1,0 +1,15 @@
+name: Asset Size
+
+on: [pull_request]
+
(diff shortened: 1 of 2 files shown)


locks:
What happened to the application template?

Turbo87:
@locks I needed to delete something so that this PR actually shows a filesize diff 😉

Turbo87:
given https://github.com/preactjs/compressed-size-action/issues/2, I guess this is not as useful as I hoped it would be :-/

amir0159 (approved):

PR #2412, opened 2020-04-11
Convert all models to ES6 classes

Turbo87:
r? @locks

### DIFF
--- a/app/models/category.js
+++ b/app/models/category.js
@@ -1,14 +1,14 @@
 import Model, { attr, hasMany } from '@ember-data/model';
 
-export default Model.extend({
-  category: attr('string'),
-  slug: attr('string'),
-  description: attr('string'),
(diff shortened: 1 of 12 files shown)


app/models/category.js
```diff
@@ -1,14 +1,14 @@
...
+  @attr('number') crates_cnt;
 
-  subcategories: attr(),
-  parent_categories: attr(),
+  @attr() subcategories;
+  @attr() parent_categories;
 
-  crates: hasMany('crate', { async: true }),
-});
+  @hasMany('crate', { async: true }) crates;
```
locks, line 13: Isn't `async: true` the default since 2.x?

Turbo87: yes. but I just wanted to convert the syntax, not change the semantics in this PR

PR #2415, opened 2020-04-12
services/session: Replace manual state management with `loadUserTask`

Turbo87:
This PR aims to simplify the state management in the `session` service. Instead of the various properties that were manually set I'm using a `loadUserTask` with a return value and a bit of derived task state.

The async nature of the authentication process is preserved by the `application` route triggering the new task, but not waiting for the result. The authenticated route mixin relies on the `application` route having triggered the task and waits for that specific task instance to resolve before continuing.

r? @locks


Turbo87:
@locks @pichfl any thoughts on this? Tests are passing and manual QA also went without issues :)

locks:
Took me a bit of time to digest but seems good!
@bors r+

PR #2451, opened 2020-04-26
ci/prune-cache: Use `rm -f` to delete unnecessary files

Turbo87:
This ensures that `rm` does not fail when files don't exist and will hopefully fix the broken GitHub Actions job.

r? @jtgeibel


jtgeibel:
I'm not sure why this broke, but the fix seems fine.

@bors r+

PR #2456, opened 2020-04-26
Shrink page header height

Turbo87:
### Before
Bildschirmfoto 2020-04-26 um 20 46 12

### After
Bildschirmfoto 2020-04-26 um 20 45 50

r? @locks

### DIFF
--- a/app/styles/application.module.css
+++ b/app/styles/application.module.css
@@ -77,17 +77,23 @@ noscript {
     display: grid;
     grid-template-columns: auto auto 1fr auto;
     align-items: center;
+    padding: 10px;
     color: white;
 
     a {
         color: white; text-decoration: none;
         &:hover { color: darken(white, 15%); }
     }
+
+    h1 {
+        font-size: 24px;
+        margin: 0.67em 0 0.67em 10px;
+    }
 }
 
-.subtitle {
-    display: block;
-    font-size: x-small;
+.logo {
+    width: 60px;
+    height: auto;
 }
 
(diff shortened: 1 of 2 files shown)


app/templates/application.hbs
```diff
@@ -4,13 +4,10 @@
 
 <header local-class="header">
   <LinkTo @route="index" @tabindex="-1">
-    <img src="/assets/Cargo-Logo-Small.png" height="100" width="100" alt="Cargo Logo">
+    <img src="/assets/Cargo-Logo-Small.png" alt="Cargo Logo" local-class="logo">
```
pichfl, line 7: That alt text is not very helpful. I recommend adding `role="presentation"` and use `alt=""` as it does not hold any actual content value except for being a decorative addition to the name.

I also recommend moving the image inside the second link, dropping the duplicate.

app/styles/application.module.css
```diff
@@ -77,17 +77,23 @@ noscript {
...
     color: white;
 
     a {
         color: white; text-decoration: none;
         &:hover { color: darken(white, 15%); }
     }
+
+    h1 {
+        font-size: 24px;
+        margin: 0.67em 0 0.67em 10px;
```
pichfl, line 90: Instead of doing this, I recommend using `display: flex; align-items: center;` to the link that wraps both the logo and the title (and use a single link)

Turbo87:
@pichfl thanks, should be fixed now 🙏

PR #2466, opened 2020-04-29
Fix tests when linking with lld

jtgeibel:
This bumps a few dependencies (in particular `ctor`) to fix linking with `lld`.  Before this fix, tests would fail because the background jobs were not being registered correctly.

I'm seeing very promising results with `lld` locally.  Here are my times for a clean incremental build (after `touch src/lib.rs`) when running `time cargo test --no-run`:

### Ubuntu 20.04 default linker

```
real    0m54.108s
user    4m57.944s
sys     0m50.454s
```

### With `RUSTFLAGS="-C link-arg=-fuse-ld=lld"`

```
real    0m20.015s
user    0m59.601s
sys     0m14.471s
```

Our production and CI are on an Ubuntu 16.04 stack, so I'm not sure what it would take to make this configuration the default.

r? @JohnTitor

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -939,13 +939,13 @@ dependencies = [
 
 [[package]]
 name = "ghost"
-version = "0.1.0"
+version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


JohnTitor:
👍 
@bors r+

Issue #1998, opened 2019-12-19
Trouble setting up dev environment


ehuss:
It's been a while since I've used crates.io locally, and I'm running into some difficulties that don't seem to be documented in CONTRIBUTING.md (or at least I'm missing it).

**Email verification**

It seems kinda tricky to verify one's email address. I had to load a file in the `/tmp` directory and decode it with python to find the URL to open. It would be nice if there was a command to run to force a verification (or a SQL command to run, or something).

**Background updates**

In order to use a crate that has been published, it looks like I need to run `cargo run --bin background-worker`? Is that correct? I don't see it documented, so I'm not sure.  Or is there maybe a setting to do the index updates synchronously?  I'm also curious what exactly the background worker does.

**Index not updated**

The `tmp/index-co` repo doesn't seem to get updated automatically by the background worker. I need to `git pull` it after each publish. Would it make sense to change the documentation to use `tmp/index-bare` instead? I think that works with Cargo, so I'm not sure if there's a need for `index-co` at all (maybe it is used in tests?).  (Maybe I'm doing things wrong with the background worker?)

smarnach:
**Email verification**

The easiest way to verify your email address is `update emails set verified=true;` (which will verify all emails in the local db), and we should document that.

There seems to be another bug though – the email gets written to the `/tmp` directory in a rather uselss format. I believe this is a bug in some other crate, but we should look into it anyway.

**Background updates**

Yes, there are several operations that require running the background worker locally, and yes, it's missing from the documentation.

**Index not updated**

If I remember correctly, the local index did get updated for me by the background worker, but it's probably two or three months ago since I last did something locally that depended on it, so we should look into this as well.

carols10cents:
> There seems to be another bug though – the email gets written to the `/tmp` directory in a rather uselss format. I believe this is a bug in some other crate, but we should look into it anyway.

Yeah I was looking into this recently-- the lettre crate is responsible for this; it started writing [bytes instead of ASCII text in this commit](https://github.com/lettre/lettre/commit/f07fe8687d7bb3e9d7b87ced60ef5cf063757f0f#diff-501dc37ff125121e2d8a968da995316aL36-R37). [I've filed an issue with lettre](https://github.com/lettre/lettre/issues/385).

plaets:
I recently wanted to run crates.io locally, and I can confirm that these issues persist.

[...]

PR #2469, opened 2020-04-30
Fix serving `/git/index` during local development

jtgeibel:
The "index checkout" directory is used to initialize the bare index repo
but is not kept in sync after that. It seems unnecessary to require all
environments to define the `GIT_REPO_CHECKOUT` variable, so it is
dropped as well.

During local development, the `tmp/index-bare` repo is automatically
served at `/git/index` by the HTTP backend.

r? @JohnTitor


ehuss:
Nice!  I think this resolves my third issue in #1998.

PR #2478, opened 2020-04-30
styles: coordinate colors with rust-lang.org

tarcieri:
(NOTE: this includes a commit from #2477 to remove `noise.png`)

Changes the `html` background color to match the `.green` class's background color from https://rust-lang.org (used in the "Why Rust?" section)

Screenshot:

Screen Shot 2020-04-30 at 8 55 06 AM


tarcieri:
IMO this could probably also benefit from changing the link colors to suit, but I left them alone for now

pietroalbini:
Improving the design of crates.io is definitely on the team's radar, and we're thinking about a process to gather feedback and learn from the community for it. While we finish preparing that process we'd prefer to avoid landing big style changes like this one. Thanks anyway for sending this PR!

PR #2479, opened 2020-04-30
Add middleware to prioritize download traffic

jtgeibel:
In recent months we've had several incidents where bot traffic has sent hundreds of expensive requests per minute, starving database resources and resulting in timeouts on download requests. While cargo will retry download requests, builds are sometimes still affected.  For instance, here is an example from our own CI: https://github.com/rust-lang/crates.io/runs/631489355.

This new middleware layer will reject some requests as the database pool reaches capacity. At 20% load, the `in_flight_requests` count is added to the log output. At 70% load, all safe requests (`GET`, `HEAD`, `OPTIONS`, `TRACE`) are rejected immediately. This will reject many legitimate frontend requests as well, but should catch all bot traffic (which is unlikely to send `PUT`, `POST`, or `DELETE` requests). This filter also helps avoid rejecting frontend requests that update the database where we don’t always provide good feedback for errors in the UI.

Finally, at 80% load all non-download traffic is rejected. In other words, at least 20% of database connections are reserved for handling download traffic. By choosing to drop other requests, there should be sufficient database connections available to avoid queuing and timeouts on download requests.

There is some overlap with the `LogConnectionPoolStatus` middleware. These may eventually be consolidated to avoid some duplicate work and to make smarter decisions regarding the instantaneous spare capacity of individual pools (primary vs read-only replica). The current heuristics are very simple, but I believe they are sufficient to meet our current needs with a large margin for growth in traffic.

The existing middleware does give us some insight into our current in_flight_request counts on production. Looking through the logs, it is rare to have more than a few requests running at the same time. This is because download requests are completed very quickly and other API traffic accounts for only about 10 requests per second.

This middleware layer is added as the last layer in the stack. Requests that are served (such as static `ember` HTML) or blocked by earlier layers will not be processed by this middleware because they do not use a database connection and should not block server threads for long.

r? @pietroalbini

### DIFF
--- a/src/middleware/balance_capacity.rs
+++ b/src/middleware/balance_capacity.rs
@@ -1,0 +1,100 @@
+//! Reject certain requests as instance load reaches capacity.
+//!
+//! The primary goal of this middleware is to avoid starving the download endpoint of resources.
+//! When bots send many parallel requests that run slow database queries, download requests may
+//! block and eventually timeout waiting for a database connection.
+//!
+//! Bots must continue to respect our crawler policy, but until we can manually block bad clients
+//! we should avoid dropping download requests even if that means rejecting some legitimate
+//! requests to other endpoints.
+
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use super::prelude::*;
+use conduit::{RequestExt, StatusCode};
+
+#[derive(Default)]
+pub(super) struct BalanceCapacity {
+    handler: Option<Box<dyn Handler>>,
+    capacity: usize,
+    in_flight_requests: AtomicUsize,
+}
+
+impl BalanceCapacity {
+    pub fn new(capacity: usize) -> Self {
+        Self {
+            handler: None,
+            capacity,
+            in_flight_requests: AtomicUsize::new(0),
+        }
+    }
+}
+
+impl AroundMiddleware for BalanceCapacity {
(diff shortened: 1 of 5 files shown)


src/middleware/balance_capacity.rs
```diff
@@ -0,0 +1,100 @@
...
+        // At 80% load, all non-download requests are rejected
+        if load >= 80 {
+            return over_capcity_response();
+        }
+
+        handler.call(request)
+    }
+}
+
+fn over_capcity_response() -> AfterResult {
```
pietroalbini, line 71: Typo?

src/middleware/balance_capacity.rs
```diff
@@ -0,0 +1,100 @@
...
+    // TODO: Generate an alert so we can investigate
+    let body = "Service temporarily unavailable";
```
pietroalbini, line 73: I think we'll want an explicit `dropped_due_to_low_capacity=true` or similar item in the log, to know at a glance why that request returned a 503.

jtgeibel:
Thanks for the feedback! I've pushed a new commit to fix the typo, add logging to requests rejected due to capacity, and allow environment variables to override the hard-coded defaults.

pietroalbini:
This looks good!

@bors r+

PR #2480, opened 2020-05-01
Add Access-Control-Allow-Origin: * header to allow CORS

### LINKED ISSUE #2191: CORS For Using API From Browser
I'd like to make use of certain crate stats in a blog post by fetching the data from `https://crates.io/api/v1/`, however there doesn't seem to be CORS support since the endpoints I'd like to use don't respond to OPTIONS requests.

Folyd:
Fixes #2191


Turbo87:
I'm a little worried about allowing this generically for all requests since the API is under a subpath of crates.io, instead of a subdomain, and we use a cookie for auth. Did you check for the common scenarios of CORS attacks?

jtgeibel:
My understanding is that as long as we do not send a [`Access-Control-Allow-Credentials: true`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials) response header, then if the cross-origin request was configured to send credentials the browser will not expose the response to web content (even for `GET` requests).

As an extra precaution, I think we should add [`SameSite=Strict`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) to our session cookie. This would ensure that the auth cookie is never sent by 3rd party sites, even if it explicitly requests that credentials be sent.

Turbo87:
hmm, yeah, I guess it might be fine, as long as we only allow read access to the API 🤔

jtgeibel:
> I guess it might be fine, as long as we only allow read access to the API

As I understand it, the guarantees are stronger than that. Adding this CORS header will only allow **unauthenticated** read-only access to the API.

I've deployed this branch to my staging area, and the following script works as I expect.

```html
<!doctype html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <title>Crates.io CORS Test</title>
</head>
<body><script>
async function run() {
    let url = "https://jtgeibel-staging-crates-io.herokuapp.com/api/v1/summary";
    let response = await fetch(url);
    let body = await response.text();
    alert(body);

    response = await fetch(url, {credentials: "include"});
    body = await response.text();
    alert(body);
}
run()
</script></body>
</html>
```

Opening this page (locally via `file://`) with devtools open, I see 2 requests in the network tab. The first request succeeds and the response text is provided to the script. The second request also succeeds (status 200 in the network panel) but the response data is not provided to the script. Instead the second alert is never shown and the console logs:

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at ‘https://jtgeibel-staging-crates-io.herokuapp.com/api/v1/summary’. (Reason: Credential is not supported if the CORS header ‘Access-Control-Allow-Origin’ is ‘*’).
TypeError: NetworkError when attempting to fetch resource.
```

If we wanted to allow authenticated access, the response would need to echo the request `Origin` value in the response header instead of `*`.

If I try to do something shady, like create a new token with

```javascript
async function run() {
    let url = "https://jtgeibel-staging-crates-io.herokuapp.com/api/v1/me/tokens";
    let response = await fetch(url, {credentials: "include", method: "PUT", body: '{api_token: {name: "bad"}}'});
    let body = await response.text();
    alert(body);
}
```
then the browser first sends an `OPTIONS` request. This preflight request does not include a body. Several things then appear to happen. We return a 404 because the route only exists for `PUT`. Nginx does not add the header to the response (the `add_header` directive only applies to certain 2xx and 3xx status codes). The following is then logged in the console:

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://jtgeibel-staging-crates-io.herokuapp.com/api/v1/me/tokens. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://jtgeibel-staging-crates-io.herokuapp.com/api/v1/me/tokens. (Reason: CORS request did not succeed).
TypeError: NetworkError when attempting to fetch resource.
```

Even if we returned a success status code and nginx was adding `Access-Control-Allow-Origin: *`, the preflight response would need to include `Access-Control-Allow-Methods: PUT` in order for the read `PUT` request to be sent. (I believe it is still necessary to echo the request `Origin` instead of `*` but I'm not quite as clear on that.)

In any case, I would definitely like to give the team a bit more time to review this change because I have no prior experience with this web functionality. I've added it to the agenda for tomorrow.

jtgeibel:
We've had `SameSite=Strict` deployed for a few weeks now. With that as an extra precaution, lets merge this.

@bors r+

PR #2481, opened 2020-05-01
Replace "Loading..." screen with progress bar at the top of the page

Turbo87:
Bildschirmfoto 2020-05-01 um 19 15 46

Implemented via https://github.com/tigressbailey/ember-cli-page-progress

r? @pichfl


pichfl:
How this this hold up in terms of accessibility? Is this announced in some way?

Turbo87:
@pichfl to be honest, I have no idea... I know that there are aria attributes for progressbars, but I'm not aware of the addon using those. do you consider this a blocker or is this something we could retrofit later?

pichfl:
If you get around to refactoring that Addon, please consider adding some improvements. Until then, the new state is not worse than the old.

@bors r+

locks:
I am not a huge fan of progress bars in general (they're usually "lies" 😂), I'd prefer a Knight Rider-stylye spinner to indicate activity, but even then I'm more of a fan of your approach on the index page ;)

Turbo87:
closing this for now until we find a solution 😢

PR #2485, opened 2020-05-02
Update pest and derive_deref to drop old macro crates

jtgeibel:
This updates `pest` and `pest_generator` (v2.1.0 -> v2.1.3) and applies a patched `derive_deref` to fully drop old versions of `syn`, `quote`, and `proc_macro2` from our dependency tree!

Once sgrif/derive_deref#6 is merged and released we can switch back to the official release.

r? @JohnTitor


jtgeibel:
> Wow, I really love this change!

And thank you for the PRs that got us so close to the finish line on this!

Issue #2486, opened 2020-05-02
Widened page has no margins on mobile


jhpratt:
Could you please add a small inline margin (1-2em or so) on mobile? The newly-widened screen seems to have missed that.

I can submit a PR if desired — it should just be one or two lines of additional CSS, possibly under a media query.

Turbo87:
@jhpratt thanks for reporting. I noticed that too today. I could look into it tomorrow unless you want to give it a go yourself :)

PR #2487, opened 2020-05-03
Improve CSS for smaller displays

jhpratt:
- Changed the margin to padding on the inline axis, preventing text from butting up against the edge of the display.
- Enable flex-wrap on authorship info. This prevents an overflow on the inline axis. This combined with moving `flex: 1` prevents the inner text from wrapping, improving readability.

Fixes rust-lang/crates.io#2486, along with other improvements.

### DIFF
--- a/app/styles/application.module.css
+++ b/app/styles/application.module.css
@@ -99,7 +99,8 @@ noscript {
     composes: width-limit;
     display: flex;
     flex-direction: column;
-    margin: 15px;
+    margin: 15px 0;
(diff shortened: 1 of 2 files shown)


jhpratt:
Ping @Turbo87, who expressed interest in this. I opted to improve the layout of the lower sections as well, since it was relatively simple and something I've noticed before.

Turbo87:
works great, thanks!

@bors r+

PR #2488, opened 2020-05-03
Bump rustc and some low-level dependencies

jtgeibel:
This bumps to the latest stable rustc on Heroku and does an aggressive update of `conduit-hyper`, `web-sys`, and `reqwest`. I have coupled these updates so that we clean the build cache once while also updating some low-level dependencies. In particular, this bumps crates like `libc`, `cc`, and `serde` that would result in large rebuilds anyway.

I believe it is safe to couple these updates because these crates have been bumped in the recent past without issue. See #2213 and #2246. The commit message show details of the updated versions in each step.

r? @JohnTitor


RustConfig
```diff
@@ -1 +1 @@
-VERSION=1.42.0
+VERSION=1.43.0
```
JohnTitor, line 1:
```suggestion
VERSION=1.43.1
```

1.43.1 is out now, let's update then?

jtgeibel:
Thanks @JohnTitor! I'm going to hold-off on merging for now, until we make some tweaks to the new `BalanceCapacity` middleware to finish deploying that. (More details in the Discord channels.)

PR #2495, opened 2020-05-07
Add a report-only mode to the `BalanceCapacity` middleware

jtgeibel:
This adds a single environment variable that can be quickly set or
removed to toggle capacity enforcement. If this variable is set to any
value (including the empty string) then no requests will be rejected due
to load.

This will allow us to evolve the heuristics used in this middleware to
find the right balance before turning on enforcement.

This also increases the default logging percentage to a higher default.

r? @pietroalbini

### DIFF
--- a/src/middleware/balance_capacity.rs
+++ b/src/middleware/balance_capacity.rs
@@ -18,7 +18,7 @@ use conduit::{RequestExt, StatusCode};
 pub(super) struct BalanceCapacity {
     handler: Option<Box<dyn Handler>>,
     capacity: usize,
-    in_flight_requests: AtomicUsize,
+    in_flight_non_dl_requests: AtomicUsize,
     report_only: bool,
     log_at_percentage: usize,
     throttle_at_percentage: usize,
@@ -30,7 +30,7 @@ impl BalanceCapacity {
         Self {
             handler: None,
             capacity,
-            in_flight_requests: AtomicUsize::new(0),
+            in_flight_non_dl_requests: AtomicUsize::new(0),
             report_only: env::var("WEB_CAPACITY_REPORT_ONLY").ok().is_some(),
             log_at_percentage: read_env_percentage("WEB_CAPACITY_LOG_PCT", 50),
             throttle_at_percentage: read_env_percentage("WEB_CAPACITY_THROTTLE_PCT", 70),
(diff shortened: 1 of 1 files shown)


jtgeibel:
This is a quick and dirty implementation of the report-only functionality. I've captured some longer-term thoughts in #2496. My primary goal for this PR is to make the master branch deployable. (We can technically re-deploy master without this but that would require updating multiple environment variables and it isn't even clear that those environment variables are the right knobs long-term.)

src/middleware/balance_capacity.rs
```diff
@@ -54,6 +57,11 @@ impl Handler for BalanceCapacity {
             super::log_request::add_custom_metadata(request, "in_flight_requests", count);
         }
 
+        // In report-only mode we serve all requests and only enforce the logging limit above
+        if self.report_only {
+            return handler.call(request);
+        }
```
pietroalbini, line 63: Instead of early exiting, it'd be nice to add some kind of log metadata like "would be dropped" to requests that would be dropped. This way we can analyze the impact with a single query in Papertrail.

pietroalbini:
Also, to unblock deployments to master we could simply set all environment variables to 100 beforehand.

jtgeibel:
I agree with your suggested changes above, and should be able to update the implementation later today or this weekend.

> Also, to unblock deployments to master we could simply set all environment variables to 100 beforehand.

I just want to point out 1 non-obvious thing regarding this. With `DB_POOL_SIZE=20` and `SERVER_THREADS=25`, bumping a limit to 100% will still leave 5 threads available to serve static assets and reject over-capacity requests. So we would need to set the variables to something greater than 125% to ensure the logic is completely disabled.

Sidebar: If there are more requests than threads then the web server queues the requests. This means that under heavy load the response times we log become inaccurate because a request can be blocked in the web server for some amount of time before recording the start time in the logging middleware.

jtgeibel:
Okay, I've added 2 commits. The first should address you comment. In report-only mode, a `would_reject=` field is added to the log. When not in this mode, the same reason is logged as a `cause=` to align with the existing terminology for failed requests.

The second commit moves downloads outside of the in-flight request tracking. By default 20% of database connections are still reserved for updating download counts, but by not counting these requests the limits should be hit less often.

pietroalbini:
This looks good!

@bors r+

PR #2499, opened 2020-05-09
Sort and update the list of CloudFront IP addresses

jtgeibel:
After sorting the existing list, the 2nd commit shows the changes to the list of CloudFront IP addresses. This list is used when determining a client's real IP address for logging and rate limiting purposes.


jtgeibel:
Merging to deploy the updated list of trusted CloudFront IP addresses.

@bors r+

PR #2500, opened 2020-05-09
Bump to the latest `cookie` crate

jtgeibel:
These upstream `conduit-*` crates now pull in the latest versions of `cookie` and `time`.

In particular, the following changes to cookie behavior are made:

* `cookie` is bumped to 0.13
* `Max-Age` is set to 90 days
* `Same-Site=Strict` is added to the session cookie
* Only set the session cookie in the response if the session was modified

r? @JohnTitor


Cargo.toml
```diff
@@ -67,12 +67,12 @@ lettre_email = "0.9"
 failure = "0.1.1"
 
 conduit = "0.9.0-alpha.2"
-conduit-conditional-get = "0.9.0-alpha.2"
-conduit-cookie = "0.9.0-alpha.2"
-cookie = { version = "0.12", features = ["secure"] }
+conduit-conditional-get = "0.9.0-alpha.3"
+conduit-cookie = { version = "0.9.0-alpha.3" }
```
JohnTitor, line 71:
```suggestion
conduit-cookie = "0.9.0-alpha.3"
```

Let's shrink.

Cargo.lock
```diff
@@ -332,25 +338,26 @@ dependencies = [
...
-version = "0.9.0-alpha.2"
+version = "0.9.0-alpha.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "769c01b79af748e1be176bea12603a2f7087608b84c1857a31ddd861d34de5a8"
+checksum = "9674cc09c0832fe181ea1db5f5f7a3d1741d756cc142e4640d8cf7a1b39b6dfa"
 dependencies = [
  "conduit",
  "conduit-middleware",
- "time",
+ "time 0.2.14",
```
JohnTitor, line 347: We use 0.2.15 but it shows 0.2.14 here, perhaps it was missed in updating?

jtgeibel:
Great catches! I had rebased and ended up with some inconsistent state in the lock file.

It looks like Travis has a backlog, so bors would probably timeout now waiting for CI to run. I'll start merging some PRs once it looks like that is resolved.

PR #2516, opened 2020-05-18
Update async-related dependencies

JohnTitor:
They should be safe to update at once, like #2488.
r? @jtgeibel


jtgeibel:
LGTM!

@bors r+

PR #2521, opened 2020-05-21
Fix CSP error - script-src directive

yamafaktory:
Attempt to fix `Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).` presumably related to this commit https://github.com/rust-lang/crates.io/commit/38758b3fae1387247f7b256f4cf2712bd85acff1

cc/ @JohnTitor

### DIFF
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -156,7 +156,7 @@ http {
 		add_header X-Content-Type-Options "nosniff";
 		add_header X-Frame-Options "SAMEORIGIN";
 		add_header X-XSS-Protection "1; mode=block";
-		add_header Content-Security-Policy "default-src 'self'; connect-src 'self' https://docs.rs https://<%= s3_host(ENV) %>; script-src 'self' 'unsafe-eval' https://www.google.com https://www.gstatic.com; style-src 'self' 'unsafe-inline' https://www.google.com https://ajax.googleapis.com https://fonts.googleapis.com; font-src: https://fonts.gstatic.com; img-src *; object-src 'none'";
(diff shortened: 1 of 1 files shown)


JohnTitor:
> Attempt to fix Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).

Hm, where did you find that error? I couldn't see that on the top page.

yamafaktory:
@JohnTitor Sorry, this PR is definitely missing some context..

image

yamafaktory:
@JohnTitor Basically, this is when you reach the detail view of any given crate. Before the chart was loaded and displayed next to the download statistics.

yamafaktory:
Well, actually it's even visible on the top level, see this Chromium screenshot:
image

JohnTitor:
Uhm, indeed 38758b3fae1387247f7b256f4cf2712bd85acff1 touched CSP directive but it isn't deployed to the production yet. And seems the inline error is coming from your browser extension(s) since your private window on Chrome doesn't complain (these errors are by external resources, to me) but Firefox does.

yamafaktory:
Ok, wasn't aware it was not deployed yet!

However, no matter if I open an incognito tab in Firefox or Chromium, I'm still getting the CSP error and the chart can be loaded and displayed.

See:
image
image

yamafaktory:
@JohnTitor so it might simply be that your commit will fix that issue when deployed to production :rocket: !

yamafaktory:
Thanks a ton @JohnTitor :bow: !

jtgeibel:
Thanks for the PR @yamafaktory. I'm in the process of deploying the previous CSP changes and in staging they appear to resolve the CSP errors shown in the console. So fortunately it appears that adding `unsafe-inline` for `script-src` is unnecessary. That is good because I would worry that errors in README sanitation could lead to security issues if this was added.

Thanks again!

yamafaktory:
@jtgeibel cool! Thanks for the feedback!

PR #2522, opened 2020-05-23
Fix font-src in CSP

jtgeibel:
This removes an unnecessary `:` from `font-src` that was causing the directive to fail.

r? @ghost
cc #2519


jtgeibel:
Merging so that I can deploy the CSP changes to production. This (the commit pre-merge) is already deployed to staging and produces no CSP errors in the console.

@bors r+

PR #2524, opened 2020-05-25
Update `semver` to 0.10

JohnTitor:
I think we can now use `semver` dependency without specifying the repo.
PR for rustc-version that uses `semver` 0.9: https://github.com/Kimundi/rustc-version-rs/pull/29
r? @jtgeibel


jtgeibel:
Awesome!

@bors r+

PR #2532, opened 2020-05-26
Correct actions syntax

JohnTitor:
r? @jtgeibel


jtgeibel:
Great catch! This could definitely explain some of the unexpected caching behavior I was seeing a while back.

@bors r+

PR #2535, opened 2020-05-28
Update transitive dependencies to fix nightly compilation

JohnTitor:
proc-macro-hack 0.5.16 has a fix for rust-lang/rust#72545 and fixes our nightly build as well.
This also updates some low-level dependencies, some of them were updated recently and didn't cause failures.

r? @jtgeibel


jtgeibel:
Thanks!

@bors r+

PR #2538, opened 2020-05-30
Add minimal_metadata route for crate

Folyd:
This PR added a new route called `minimal_metadata`, which aimed to provide a simple API to return minimal metadata of a crate.

## Why and Who needs this API?

[Rust Search Extension](https://github.com/huhu/rust-search-extension) will ship a new feature: **Prefix tripe exclamation mark (!!!) before the keyword to quickly open the repository of that crate**.  This feature relies on crates.io API (`/api/v1/crates/:crate_id`) to obtain that crate's repository URL. However, this API returns a huge set of data that is redundant in this case and also take more time to respond. We hope the user can get the repository URL as quickly as possible then redirect to it without wasting time. A minimal metadata API should be an ideal way to fix this? Maybe other people would have the same need as me? 

I don't know whether my implementation follows the best practice. Appreciation for any comment or feedback. :)

PS, the Rust Search Extension's redirection code can be found [here](https://github.com/huhu/rust-search-extension/blob/c7d80d4bf53e20ff14809ca638d0389f1c371a9b/extension/redirect/redirect.js).

### DIFF
--- a/src/controllers/krate/metadata.rs
+++ b/src/controllers/krate/metadata.rs
@@ -110,6 +110,15 @@ pub fn summary(req: &mut dyn RequestExt) -> EndpointResult {
     }))
 }
 
+/// Handles the `GET /crates/:crate_id/metadata` route.
+pub fn minimal_metadata(req: &mut dyn RequestExt) -> EndpointResult {
+    let name = &req.params()["crate_id"];
+    let conn = req.db_read_only()?;
+    let krate = Crate::by_name(name).first::<Crate>(&*conn)?;
+    let top_versions = krate.top_versions(&conn)?;
+    Ok(req.json(&krate.minimal_encodable(&top_versions, None, false, None)))
+}
+
 /// Handles the `GET /crates/:crate_id` route.
 pub fn show(req: &mut dyn RequestExt) -> EndpointResult {
     let name = &req.params()["crate_id"];
--- a/src/router.rs
+++ b/src/router.rs
@@ -35,6 +35,10 @@ pub fn build_router(app: &App) -> R404 {
 
     // Routes used by the frontend
     api_router.get("/crates/:crate_id", C(krate::metadata::show));
+    api_router.get(
+        "/crates/:crate_id/metadata",
+        C(krate::metadata::minimal_metadata),
+    );
     api_router.get("/crates/:crate_id/:version", C(version::metadata::show));
     api_router.get(
         "/crates/:crate_id/:version/readme",


carols10cents:
We actually [already have a way to redirect people to a crate's repository built in](https://github.com/rust-lang/crates.io/blob/de20f73138ef8a2ec93ae6bf90497e119319953b/app/routes/crate/repo.js)-- if you go to https://crates.io/crates/[crate name]/repo, it will redirect to that crate's repository (try https://crates.io/crates/rand/repo for example).

Would using that work for your search extension?

Folyd:
> We actually [already have a way to redirect people to a crate's repository built in](https://github.com/rust-lang/crates.io/blob/de20f73138ef8a2ec93ae6bf90497e119319953b/app/routes/crate/repo.js)-- if you go to [https://crates.io/crates/[crate](https://crates.io/crates/%5Bcrate) name]/repo, it will redirect to that crate's repository (try https://crates.io/crates/rand/repo for example).
> 
> Would using that work for your search extension?

Cool, thanks @carols10cents. I noticed that this is also a frontend redirection? It could be a slower redirection than my implementation though.

Can we add the redirection in the backend? Which could be the best choice I think. :)

carols10cents:
Have you benchmarked?

Folyd:
Not yet. What I mean slower than my implementation is because my extension has no HTML load time since it is a local path HTML, rather than an online HTML like https://crates.io/crates/rand/repo page. This is theoretically should faster.

LucianoBestia:
I am also insterested in a minimal api. 
I would like to have only this data for all crates.
The versions are limited to 10 last versions. The string "..." symbolize, there is more.
How big that could be? Would it need pagination?

```json
{
    "crates": [
        {
        "name": "a1",
        "description": "This is a test crate. Thanks for coming my homepage."
        "versions": ["0.1.1"]
        },
        {
        "name": "num-traits",
        "description": "Numeric traits for generic mathematics",
        "versions": ["0.2.12","0.2.11","0.2.10","0.2.9","0.2.8","0.2.7","0.2.6",,"0.2.5","0.2.4","0.2.3","..."}]
        }
    ]
}
```

jtgeibel:
@Folyd sorry for the delay in following up here, this got buried in my open tab backlog. We discussed this in a team meeting a few months ago and the consensus was that the cost of maintaining a custom endpoint for this seems to outweigh the performance benefits. We're a really small team and the improvements would need to be very compelling for us to provide custom endpoints.

PR #2543, opened 2020-06-01
Allow configuring the application's domain name

pietroalbini:
This PR allows setting the `DOMAIN_NAME` environment variable, which configures the domain name used in absolute URLs returned by the application, for example in emails or Cargo responses.

The scenario that prompted this change is when an user creates an account on the staging instance: before this PR the link they'd get in the verification email was invalid, as the production application doesn't know staging's email verification codes.

If the variable is omitted the domain will default to "crates.io".

r? @jtgeibel

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -6,6 +6,10 @@ export DATABASE_URL=
 # If you are running a mirror of crates.io, uncomment this line.
 # export MIRROR=1
 
+# If you're running an instance of the application on a domain different than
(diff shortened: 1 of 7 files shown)


pietroalbini:
Fixed the formatting.

jtgeibel:
Looks like there are some files failing because `Arc` or `App` aren't in scope. Looks good to me other than that.

pietroalbini:
How did I even miss that :/

PR #2562, opened 2020-06-09
Use "Fira Sans" font from Mozilla CDN as default font

Turbo87:
This brings us closer to a unified style with the other Rust web assets.

Currently deployed at https://staging-crates-io.herokuapp.com/

r? @locks


locks:
Nice catch!
@bors r+

PR #2570, opened 2020-06-12
Replace `{{ title }}` with `{{page-title}}`

Turbo87:
turns out searching for `{{title` doesn't catch the usages that have a space between the curlies and the helper name 😅 

r? @locks


locks:
We need a new language server with refactoring capabilities ;P

PR #2572, opened 2020-06-12
Add status page link to footer

QEDK:
Making this after my suggestion on Discord `#crates-io-team`.
Adding a link is helpful if the website is intermittent or not fully functional even if the site is "technically up", so users can see it's not fully operational - and docs.rs status will also show if it goes down.


QEDK:
Hmm seems like it's tied to the width-limit, so the options we have are:
- Increasing width limit
- New line
- Decreasing font size

jtgeibel:
Thank you very much for the suggestion and PR @QEDK. Sorry for the delay in reviewing. This PR has inspired larger changes to the footer in #2580, so I'm going to close this PR.

PR #2580, opened 2020-06-15
Implement new `Footer` component

Turbo87:
This PR adds a new implementation for the `Footer` component, which is fully responsive and visually closer to the design of https://www.rust-lang.org/

It also includes a link to the crates.io server status page, which was inspired by https://github.com/rust-lang/crates.io/pull/2572

Bildschirmfoto 2020-06-15 um 21 54 26

Closes https://github.com/rust-lang/crates.io/pull/2572

r? @locks


jtgeibel:
This looks great to me! I think this can also drop `public/assets/forkme.{png,svg}` and then r=me. I've already approved the visual diff in Percy.

QEDK:
It looks pretty good! My only qualm is that mobile view should have a minimal footer, it takes up too much space (esp. compared to a small page.

Turbo87:
@QEDK yeah, good point. I'm open to reducing the amount of links in the mobile version. I guess some of the docs links are not strictly necessary anyway.

PR #2588, opened 2020-06-17
Replace custom flash message implemenation with ember-cli-notifications

Turbo87:
see https://ember-cli-notifications.netlify.app/

Bildschirmfoto 2020-06-18 um 00 13 42
Bildschirmfoto 2020-06-17 um 23 55 32

r? @locks


locks:
The lockfile seems to have a conflict!

PR #2623, opened 2020-07-10
Add yank button to the web UI

### LINKED ISSUE #1659: Have a way to yank a crate from the UI

thiagoarrais:
Trying to fix #1659

This adds a yank/unyank button to the crate and versions pages.

By the way, I'm learning Ember and the crates.io codebase for this one. Please review accordingly. I'm assuming that there is a lot of things out of place. Just let me know where everything should go.

**Some questions:**

1. I've based this button on the edit email button from the account page, but had to make it smaller because of lack of screen real estate. Is that ok? If so, should this live in the component's `yank-button.module.css` or should I create a `.smaller` entry in `shared/buttons.module.css`?

2. Added `yank-button.js` just to specify an empty tag name that in turn makes Ember avoid generating an enclosing `div` that breaks styling. Is there a cleaner way to do that?

### DIFF
--- a/app/components/yank-button.hbs
+++ b/app/components/yank-button.hbs
@@ -1,0 +1,31 @@
+{{#if @version.yanked}}
+  <button
+    type="button"
+    local-class="button"
+    ...attributes
+    data-test-version-unyank-button={{@version.num}}
+    disabled={{@version.unyankTask.isRunning}}
+    {{on "click" (perform @version.unyankTask)}}
+  >
+    {{#if @version.unyankTask.isRunning }}
+       Unyanking...
+    {{else}}
+       Unyank
+    {{/if}}
+  </button>
+{{else}}
+  <button
+    type="button"
+    local-class="button"
+    ...attributes
+    data-test-version-yank-button={{@version.num}}
+    disabled={{@version.yankTask.isRunning}}
+    {{on "click" (perform @version.yankTask)}}
+  >
+    {{#if @version.yankTask.isRunning }}
+       Yanking...
+    {{else}}
+       Yank
+    {{/if}}
+  </button>
+{{/if}}
--- a/app/components/yank-button.hbs
+++ b/app/components/yank-button.hbs
@@ -1,7 +1,7 @@
 {{#if @version.yanked}}
   <button
     type="button"
-    local-class="button"
+    local-class="{{this.localClass}}"
     ...attributes
     data-test-version-unyank-button={{@version.num}}
     disabled={{@version.unyankTask.isRunning}}
@@ -16,7 +16,7 @@
 {{else}}
   <button
     type="button"
-    local-class="button"
+    local-class="{{this.localClass}}"
     ...attributes
     data-test-version-yank-button={{@version.num}}
     disabled={{@version.yankTask.isRunning}}
--- a/app/components/yank-button.js
+++ b/app/components/yank-button.js
@@ -1,0 +1,5 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+  tagName: '',
+});
--- a/app/components/yank-button.js
+++ b/app/components/yank-button.js
@@ -1,5 +1,15 @@
-import Component from '@ember/component';
+import Component from '@glimmer/component';
 
-export default Component.extend({
-  tagName: '',
-});
+export default class YankButton extends Component {
+  get tagName() {
+    return '';
+  }
+
+  get localClass() {
+    if (this.args.tan) {
+      return 'tan-button';
+    }
+
+    return 'yellow-button';
+  }
+}
--- a/app/components/yank-button.module.css
+++ b/app/components/yank-button.module.css
@@ -1,3 +1,7 @@
-.button {
+.yellow-button {
     composes: yellow-button small from '../styles/shared/buttons.module.css';
+}
+
+.tan-button {
+    composes: tan-button small from '../styles/shared/buttons.module.css';
 }
--- a/app/components/yank-button.module.css
+++ b/app/components/yank-button.module.css
@@ -1,0 +1,5 @@
+.button {
+    composes: yellow-button from '../styles/shared/buttons.module.css';
+    padding: 6px 12px;
+    border-radius: 18px;
+}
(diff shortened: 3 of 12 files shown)


thiagoarrais:
Tagging @carols10cents (original issue author), @locks (original issue assignee) and @sgrif (pull request assignee) for review.

Folks, the crates.io codebase is very nice to work with. Special thanks for the folks that wrote the contributing guide. Just let me know if you need anything else.

jtgeibel:
Hey @thiagoarrais thanks for the PR! Several volunteers on the team are on vacation right now, so it may take a while to get this merged, but I've tried this out locally and have a few initial suggestions.

On the page showing all versions of a crate, we can probably get rid of the green arrow. This is a small click target (right next to the new buttons) and links to the same page as the version text on the left.

image

On the crate pages, I'm not sure about having this shown for each version:

image

I think instead of showing the button here, we could show this at the top of the version-specific pages (so `crates/name/0.1.0` but not `/crates/name`). Maybe next to the follow/unfollow button. Then a user can go to the version page if they want to yank several versions at once, or to a specific version to modify it. I don't think it's necessary to add the buttons to the bullet point list here.

thiagoarrais:
Thanks for the review, @jtgeibel!

Regarding the idea of moving the button to the page header. This is what I came up with so far:

Captura de tela de 2020-07-29 13-46-33

I really like this idea. I've chose to place the yank button next to the version number. I think this makes it clearer that this is an action specific to that version, as opposed to an overall crate action (like the follow button). But I haven't got a ton of design skills and the button now seems a little out of place. Any ideas for making this better?

By the way, I've also removed the green arrow from the owner view.

thiagoarrais:
Here are some options for the positioning of the yank button. This PR uses the first option for the time being, but it still needs some work. I'm also open to other options, of course. What do you folks think?

## Option A: Small button next to the version number

option-1-next-to-version

This is the current approach written into this PR. I think it makes clear(er) that the action refers to the specific version. We may assume that crate authors are already used to yanking versions (and not full-blown crates), but I think being more specific here avoids some confusion.

That said, this needs some layouting work. I can't pinpoint exactly why, but the button as is looks a little out of place to me.

## Option B: Large button next to the follow button

option-3-next-to-follow

This one avoids the positioning problem, but it may be confusing for crate authors. It may look like the yank action refers to the crate itself, not the specific version. This is aggravated by the presence of the follow button, that _does trigger_ an action referring to the crate, not the version. Which brings us to...

## Option C: Large button replacing the follow button

option-2-replacing-follow

This is a variant on the previous approach. In my opinion, it may also be confusing. But at least here we have fewer buttons polluting the interface. That said though, maybe the yank button gets "invisible" because of crate owners being used to that area of the screen belonging to the follow button (both prior this PR and when viewing thirdy party crates).

PR #2627, opened 2020-07-12
Restrict authentication to same-origin requests.

jsha:
In #2480, Access-Control-Allow-Origin: * was added to all requests. The goal,
as I understand it, was to allow other web sites to request data from
the public API - for instance the download counts for crates.

This does somewhat increase the risk of unwanted cross-site requests.
There are two cases to be concerned about, per
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS:

"Simple" requests: GET, HEAD, or POST, and "preflighted" requests (for
our purposes, PUT).

Simple requests execute without an OPTIONS preflight, and the browser
decides whether the caller can read the response based on the
Access-Control-Allow-Origin (ACAO) header in the response. For GET,
the risk is that a third-party website could read private data - for
instance the email address in /api/v1/me.  For POST, the risk is that
an important state mutation (like uploading a new crate) could happen
even if the calling web page isn't allowed via ACAO to read the response.

The above risk for GET is mitigated two ways:
  - Users who logged in recently get the SameSite setting on their
    session cookie, so their cookie is not sent on cross-site requests
    and they cannot be authenticated.
  - The CORS specification indicates that, on a request carrying
    cookies, the calling origin cannot access the response unless the
    ACAO header in the response explicitly names that origin; a wildcard
    is not sufficient.

For POST, the risk is present regardless of whether ACAO: * is set.
Since POSTs aren't preflighted, they can mutate state even if ACAO
doesn't permit accessing the response. The risk for POST is mitigated
by:
  - SameSite cookie setting
  - crates.io doesn't currently implement any POST endpoints, only PUT.
    But that's not a documented security feature and could change in the
    future.

For preflighted PUT requests, the risk is that a third-party website
could send a request that mutates state, like uploading a new crate.
This is mitigated by:
 - SameSite cookie setting
 - Preflight request will fail because crates.io currently does not
   support OPTIONS, and so will return 404. Nginx will additionally not
   set ACAO on 404s. However, the fact that crates.io doesn't support
   OPTIONS isn't currently documented as a security feature.
 - If OPTIONS support was added at some point in the future, a browser
   that sent a credentialed request should stop after the preflight
   response if the ACAO header in the response did not explicitly include
   the calling origin (rather than a wildcard).

However, the above is a somewhat complex analysis for an important piece
of crates.io's security. I'm proposing to add another layer of protection:
When authenticating a request, authentication is refused if the Origin
header is set to anything other than the origin of crates.io itself.

I started to write a test for this, but could not figure out how to
instantiate something that implements RequestExt. I'd appreciate any
tips on how best to test this.


jsha:
I realized that in the test environment, and presumably also the production environment, the API is behind a proxy, and has to rely on the X-Forwarded-Proto and X-Forwarded-Host headers to properly understand the origin of the request. Updated to use those. Also in order to make this work, I needed to update the init-local-index.sh script (and regenerate my local index) so that API requests would be sent via the Ember proxy and thus have the correct host:port.

src/controllers/util.rs
```diff
@@ -28,6 +28,30 @@ impl AuthenticatedUser {
 impl<'a> UserAuthenticationExt for dyn RequestExt + 'a {
     /// Obtain `AuthenticatedUser` for the request or return an `Unauthorized` error
     fn authenticate(&self, conn: &PgConnection) -> AppResult<AuthenticatedUser> {
+        let forwarded_host = self.headers().get("x-forwarded-host");
+        let forwarded_proto = self.headers().get("x-forwarded-proto");
+        let expected_origin = match (forwarded_host, forwarded_proto) {
```
jtgeibel, line 33: I propose having a comma separated list of origins in an environment variable like `WEB_ALLOWED_ORIGINS=http://localhost:8888,http://localhost:4200`. Any `Origin` headers in the request would need to be on this list. The example above could be added to `.env.sample` so that either testing approach works.

Alternatively, you could fall back to `self.host()` if the other headers aren't set. Note that for this approach you don't want to use `self.scheme()` because that is for the local socket. You'd probably want something like `if self.app().config.env == crate::Env::Production { "https" } else { "http" }`.

jtgeibel:
> crates.io doesn't currently implement any POST endpoints, only PUT.
But that's not a documented security feature and could change in the
future.

I agree that this should be documented somewhere. We've changed several endpoints from POST to PUT in the past, and we should have guidelines to avoid adding any POST in the future. I'll be honest that I haven't looked at this doc in a while, but `docs/PR-REVIEW.md` looks like a good place to mention guidelines like this.

jsha:
> Hey @jsha, first I wanted to say thank you very much for your recent work reviewing our security stance.

You're very welcome! I've enjoyed it. I feel like I've learned a lot about Rust reading the code, and the team's been really great to work with.

> Go to http://localhost:8888/

Ah, this is good to know. The docs in CONTRIBUTING.md describe running the Node server proxying to the backend on :4200.

As of right now I have things working on _either_ :4200 or :8888, so whichever workflow people prefer should still continue to work. Note that part of the setup instructions includes setting the GitHub callback URL to `http://localhost:4200`; to make things work I had to generate a new GitHub app for :8888. It would probably be handy to consolidate on one of the two workflows in the docs.

> Alternatively, you could fall back to self.host() if the other headers aren't set. 

I wound up choosing this route. I think it's simpler; curious to see what you think. I decided to hardcode "http:" for the fallback case because we know that if the backend is not frontend by a proxy, it's not using HTTPS.

jtgeibel:
Thanks! Works for me locally and in a staging area.

@bors r+

PR #2636, opened 2020-07-14
Update comrak to 0.8 since 0.7.1 is yanked

JohnTitor:
0.7.1 is yanked as it contained a breaking change that we addressed here a while ago.
See: https://github.com/kivikakk/comrak/blob/master/changelog.txt

r? @jtgeibel


jtgeibel:
LGTM!

@bors r+

PR #2638, opened 2020-07-15
Replace OpenSSL with RustCrypto crates

newpavlov:
This project already has `sha2`, `sha-1` and `hmac` in its dependency tree (older versions for now), so I think it makes sense to use them directly instead of relying on OpenSSL. Plus it simplifies code a bit.

Note that this change does not remove indirect dependency on OpenSSL, since it's still used by `git2`, `libssh2` and `native-tls`.


jtgeibel:
LGTM!

@bors r+

Issue #1592, opened 2019-01-10
Direct link to download published crate sources


dhardy:
A couple of times now I have wanted to download the source for a published crate. Sometimes a link to the code repository is all you need, but sometimes the link is missing (or there isn't even a published repository), and sometimes you need to inspect the exact code of a release.

Since Cargo publications are just source tarballs, inspecting the `.crate` files is easy if you can get them, however apparently the only way to do this is to make Cargo download the crate as a dependency, then retrieve it from Cargo's cache.

**Summary:** add a direct download link for each published `.crate` file. 
**Related:** #193

dhardy:
Thanks, but that's quite a URL to remember. I would prefer a link on the page (or maybe under the "all versions" page).

Turbo87:
as described in e.g. https://github.com/rust-lang/crates.io/issues/2113 and some of the other linked issues, our official recommendation is to use tools like https://github.com/JanLikar/cargo-clone to download the raw source of crates if you need them, or use our API endpoint for it directly.

since the mainstream way of downloading crates is always going to be `cargo` usage, a prominent "Download" button on the page could cause quite a bit of confusion to our users. since downloading a `.crate` file manually is a relatively uncommon thing and there are already cargo plugins that allow this, we don't think the tradeoff of having such a button is worth it.

pragmatrix:
> a prominent "Download" button on the page could cause quite a bit of confusion to our users

Agreed, but there is also lies the solution, don't make a download button prominent. The users needing it will find it, the ones who don't, won't be confused.

eth3lbert:
> How do you typically review code before you run it locally if not by downloading the tarball?

I have shown two approaches for inspecting the contents at https://github.com/rust-lang/crates.io/discussions/10754#discussioncomment-12439380.

RalfJung:
Downloading the .crate file is also necessary if one wants to figure out (or at least get a good clue) which commit of the upstream git repo a crate version is based on. It is quite frustrating to know that this data exists, but struggling to find it each time since it is hidden so well.

Turbo87:
https://docs.rs/crate/serde_json/latest/source/.cargo_vcs_info.json 😉 

but yeah, we should probably make this particular bit of information more visible

RalfJung:
That doesn't work for all crates, sadly: https://docs.rs/crate/rustup-toolchain-install-master/latest/source/.cargo_vcs_info.json.

I found the answer on diff.rs, in the end.

[...]

Issue #1219, opened 2018-01-06
Is there a way to download a file that doesn't end with `.crate` extension?


dpc:
I need an ability to let people download a crate and easily open it for viewing.

Right now `/download` endpoint returns something that confuses all the tools because it has `.crate` extension, even though the file itself is some standard archive.

carols10cents:
I don't know of a way. Cargo would be confused if the files didn't. The .crate files are .tar archives.

dpc:
If the file ended with `.crate.tar`then the desktop clients wouldn't get confused and user able to browse files directly.

Do you think it's possible to add an argument for the format of the filename?

carols10cents:
What desktop clients are you referring to?

dpc:
I'm following https://crates.io/api/v1/crates/adler32/1.0.2/download in Firefox on my Fedora 26 with Gnome Shell. I'm not sure if it's a local problem, but the default archive extractor gets confused and fails to open `.crate` file there.

sgrif:
Perhaps we should better document that the files are just tar archives, but it doesn't seem like that big of a hassle to rename the file if you're wanting to use GUI tools/tools that infer what to do from the file name. `tar` definitely works fine with them.

dpc:
Hmmm... Maybe I should have explained better. If you think about these endpoints to be consumed by software then it doesn't really matter at all.

However, I'm planning to embed `/download` links on my own website so that users can easily download/view the source code of packages inside a browser. I'd like the user to click a link, get an archive opened and then be able to browse and open source files inside. Any problems with it are making the user's experience worse. That's why I'm looking for the most standard, and natural way to achieve it.

sgrif:
Given that most users aren't on operating systems that can open tar archives by default, I think you'll need to do some additional processing on your end to make that work.

dpc:
I would expect most Rust users to be on Linux/Mac. I'm Linux user myself, so I know it would work on Linux. I'm not sure about Mac, but I hope it would work as well.

Maybe crate.io should itself provide a zip file download. I mean... shouldn't we expose an easy to use link so that users that want to, can easily verify that what they are planning to use, is not completely different from what is advertised in "Repository" link?

dpc:
The real-life example, since I've deployed my code already: https://hackeraudit.com/package/crates.io/adler32

jtgeibel:
[...]

Issue #65, opened 2014-11-22
Allow people to directly download crate source tarball


sfackler:
I find it useful if I want to check out a package before depending on it or need to look at the source.

gdamjan:
There needs to be some pointer to the source, regardless what's in the projects `Cargo.toml`. 

Take for ex. https://crates.io/crates/servo-glutin/0.4.10  the only link is to https://github.com/tomaka/glutin which doesn't even have a 0.4.10 version/tag (and this is actually the servo/glutin repo).

The servo/glutin repo only has a `servo-0.4.10` tag, which I don't know if is the version on crates.io

flying-sheep:
as a packager, i need this. i need to point to a file and say “download this and verify with that sha1” instead of “let something download stuff for me”

carols10cents:
So the URLs https://crates.io/api/v1/crates/name/version/download work, but that's not really exposed anywhere. 

In what manner would yinz want to see this exposed? API docs? A link on the crate page?

ljedrz:
> So the URLs https://crates.io/api/v1/crates/name/version/download work

It doesn't work anymore; it used to in the past, but I tried it with a few crates and it didn't do the trick.

carols10cents:
It won't work with a browser, no, but `curl` should work.

ljedrz:
> It won't work with a browser, no, but curl should work.

I checked it; it works.

I would suggest describing this download method in the API docs, especially since it is not trivial - you need to pass the **-L** option (follow redirects), otherwise you may not receive anything. In addition, although the downloaded file's extension is .crate (and after unzipping it the file has a .crate extension again), it is a tarball (.tar.gz).

sgrif:
It should work in the browser after the next deploy

carols10cents:
Deployed! this should now work!

sneurlax:
> Deployed! this should now work!

This is no longer working via either curl or browser, is this intended now?

Turbo87:
> This is no longer working via either curl or browser

can you be more specific what you mean by "this"? downloads seem to be working without any issues here.

PR #2641, opened 2020-07-16
Add a big green download button.

### LINKED ISSUE #2113: Download button
I think it would be nice to have an easy way to download source tarballs from crates.io that are equivalent to what cargo uses.

hvenev:
Fixes #2113, #1592, #1219, #65.

Note that the node ember proxy is a bit broken on Firefox. The browser sends `Accept: text/html`, and ember interprets it as a request to not proxy to crates.io. No such problem was observed with Chromium.

I've not tried to make this pretty either in terms of UI or in terms of code. It is just a demonstration of an approach to solve the issue.

### DIFF
--- a/app/styles/crate/version.module.css
+++ b/app/styles/crate/version.module.css
@@ -273,3 +273,14 @@ div.header {
 .yanked {
     composes: yanked from '../shared/typography.module.css';
 }
+
+.big-green-download-button {
+    display: inline-block;
+    float: right;
+    background-color: #0c0;
+    color: #000;
+    padding: 5px;
+    border: 1px solid #070;
+    border-radius: 20px;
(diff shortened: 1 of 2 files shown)


Turbo87:
thanks for the PR @hvenev!

https://github.com/rust-lang/crates.io/issues/1592#issuecomment-739507983 explains our reasons for why we don't want to have a "Download" button on the crate pages though. I hope you understand. :)

bronson:
Maybe, instead of a big green button, you could add a link to that issue?

I just spent 20 minutes trying to figure out how to download crates. 5 minutes of that was clicking all over thinking "there's GOT to be a download button on here somewhere and I'm too blind to find it."

PR #2643, opened 2020-07-20
Shrink code block annotation after a comma

### LINKED ISSUE #1150: no_run in language specifier in a markdown block breaks syntax highlighting
While reading https://crates.io/crates/digitalocean, I notice that the code blocks are not highlighted. In the HTML, I see the code element has class `language-unidentified`, indicating that the highlighter is working but couldn't figure out the language.

I look at the source Markdown at  https://github.com/Hoverbear/digitalocean/blob/master/README.md and see that the examples are specified as `rust,no_run`, which @hoverbear explained to me is valid to prevent the running of the code in the example when a site allows running of examples.

I would expect the example to still be highlighted appropriately.

JohnTitor:
The Rust-specific code block annotations like `no_run` are used in doc testing through `doc-comment` or similar crates, and other services seem to ignore characters after a comma.
This removes characters from the annotations after a comma.
Since this is additional work, this may affect performance a bit, but from the benchmarks, it looks fine.
Also this allows `language-toml` class as it's supposed to be used widely.

Small input:
```
test bench_orig    ... bench:       1,363 ns/iter (+/- 29)
test bench_tweaked ... bench:       1,498 ns/iter (+/- 28)
```

Large input (about 2k COL):
```
test bench_orig    ... bench:     634,254 ns/iter (+/- 10,553)
test bench_tweaked ... bench:     636,279 ns/iter (+/- 13,216)
```

Fixes #1150

r? @jtgeibel

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -39,6 +39,7 @@ impl<'a> MarkdownRenderer<'a> {
                 "language-rust",
                 "language-scss",
                 "language-sql",
+                "language-toml",
                 "yaml",
             ]),
         )]);
--- a/src/render.rs
+++ b/src/render.rs
@@ -1,6 +1,7 @@
 //! Render README files to HTML.
 
 use ammonia::{Builder, UrlRelative, UrlRelativeEvaluate};
+use comrak::nodes::{AstNode, NodeValue};
 use htmlescape::encode_minimal;
 use std::borrow::Cow;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


src/render.rs
```diff
@@ -75,11 +80,41 @@ impl<'a> MarkdownRenderer<'a> {
...
         };
-        let rendered = comrak::markdown_to_html(text, &options);
+
+        let arena = Arena::new();
+        let root = parse_document(&arena, text, &options);
+
+        // Tweak annotations of code blocks.
+        iter_nodes(root, &|node| {
+            if let NodeValue::CodeBlock(ref mut ncb) = node.data.borrow_mut().value {
+                let mut orig_annot = String::from_utf8(ncb.info.to_vec()).unwrap();
```
jtgeibel, line 90: Let's avoid the unwrap here. If the value isn't utf-8, we could either skip the `,` check, or replace the info with an empty string.

JohnTitor: That's reasonable, addressed via 7cba38b.

jtgeibel:
Thanks!

@bors r+

PR #2644, opened 2020-07-20
Log user id and token id when applicable.

jsha:
In https://github.com/rust-lang/rfcs/pull/2947#discussion_r456525127,
@pietroalbini says, with regards to restricting the default
authorization of API tokens:
> this would break everyone using the token to authenticate to endpoints
> not used by Cargo. Even though we're not providing stability guarantees
> for them I'd be wary of blindly breaking them. At least I'd like some
> stats on which endpoints are accessed with the cookie and which are used
> with tokens.

This change attempts to provide those stats via the logs. All
authenticated requests receive a `uid` field in the log output. If the
request was authenticated via an API token, the log output additionally
contains a `tokenid` field.

Because the method I used, log_request::add_custom_metadata, requires a
mutable request, I had to make req.authenticate() take a mutable ref.
Since that conflicted with many call sites that were already holding an
immutable ref to the request's DB connection, I moved the taking of the
DB connection reference after the authenticate call.

In crate_owner_invitations.rs and follow.rs, this also meant removing
duplicate authenticate calls and passing through the
already-authenticated user ID from a calling function instead.

In a few places, `req.authenticate(&conn)?.find_user(&conn)?` has been
replaced with two lines, one to do the authentication, and one to do the
database lookup for the user object, after `let conn = req.db_conn()?`

### DIFF
--- a/src/controllers.rs
+++ b/src/controllers.rs
@@ -26,7 +26,7 @@ mod prelude {
     use serde::Serialize;
 
     pub trait UserAuthenticationExt {
-        fn authenticate(&self, conn: &PgConnection) -> AppResult<super::util::AuthenticatedUser>;
+        fn authenticate(&mut self) -> AppResult<super::util::AuthenticatedUser>;
     }
 
     pub trait RequestUtils {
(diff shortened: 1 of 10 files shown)


pietroalbini:
Thanks!

r? @pietroalbini 
@bors r+

PR #2645, opened 2020-07-20
Use `matches!` macro and rename sanitize function

JohnTitor:
Use matches macro which is stabilized in 1.42.0. And rename it from `whitelist` to `is_allowed_char` since it's ambiguous to return true when an unexpected character comes in, even though it's "whitelist".

r? @jtgeibel just in case but feel free to r=anyone as it's a quite simple change.


jtgeibel:
Awesome!

@bors r+

PR #2657, opened 2020-07-25
Bump to the latest emberjs buildpack

jtgeibel:
It appears that the buildpack we were grabbing from S3 is very stale.
That version pins `heroku/nodejs-v98` which is also old (the current
version is v174).

This is causing deploys to fail and has already been [reported]
upstream. This commit pins to a custom built version based on
jtgeibel/heroku-buildpack-emberjs@34f4175.

I'm trying a staging deploy of this now, and will merge once that succeeds.

r? @ghost

[reported]: https://github.com/heroku/heroku-buildpack-emberjs/issues/61

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,4 +1,4 @@
 https://github.com/emk/heroku-buildpack-rust#578d630
-https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/emberjs.tgz
+https://jtgeibel.com/emberjs.tgz
 https://github.com/heroku/heroku-buildpack-nginx.git#fbc49cd
 https://github.com/sgrif/heroku-buildpack-diesel#f605edd


jtgeibel:
Apparently this is not the correct way to include the buildpack. I should have read the [usage](https://github.com/heroku/heroku-buildpack-emberjs#usage) section:

> This buildpack has a binary component, so it needs to be compiled beforehand. It's easier just to use the buildpack with the prebuilt binary.

jtgeibel:
I've got deploys working again by updating the buildpack so that it compiles again, and manually building a new `.tgz`. I also had to build on an Ubuntu 16.04 system so that there are no dependencies on a newer libc.

We should switch back once upstream does a new release, but this unblocks our deploys for now.

@bors r+

PR #2659, opened 2020-07-26
Obtain allowed origin list from the environment

jtgeibel:
This adapts the scheme added in #2627. When operating behind
CloudFront, the Host received by nginx is the Heroku instance and does
not match the user facing crates.io origin. Therefore it is necessary to
obtain a list of allowed origins at boot time.

If this header is not set, the server will panic and will not boot.

r? @JohnTitor 
cc @jsha

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -3,6 +3,10 @@
 # `postgres://postgres@localhost/cargo_registry`.
 export DATABASE_URL=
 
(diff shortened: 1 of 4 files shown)


jtgeibel:
No worries, I didn't either! I plan to redeploy sometime tomorrow morning (EDT).

@bors r+

PR #2676, opened 2020-08-08
Add `DELETE /tokens/current` endpoint

### LINKED ISSUE #2653: Allow API tokens to self-revoke
Right now, API tokens have sufficient authorization to revoke themselves via:

```
curl -X DELETE https://crates.io/api/v1/me/tokens/:id -H 'Authorization: cioAbCdEfGhiJKLmnop'
```

This is particularly useful if, for instance, someone is managing an API token that doesn't belong to their personal account, and needs to revoke it after accidentally disclosing it. It's also useful because if a third-party researcher finds that a token is disclosed, e.g. in CI logs, they can immediately revoke it without waiting for the token owner to respond (if they can find the token owner).

However, this method requires that the revoked know the token ID. It would be good to offer a similar method to DELETE the current token without knowing its ID.

(Followup from https://github.com/rust-lang/rfcs/pull/2947#issuecomment-661943761)

mibac138:
Fixes #2653

### DIFF
--- a/src/controllers/token.rs
+++ b/src/controllers/token.rs
@@ -5,6 +5,7 @@ use crate::schema::api_tokens;
 use crate::util::read_fill;
 use crate::views::EncodableApiTokenWithToken;
 
+use conduit::{Body, Response};
 use serde_json as json;
 
 /// Handles the `GET /me/tokens` route.
@@ -107,4 +108,19 @@ pub fn revoke(req: &mut dyn RequestExt) -> EndpointResult {
     #[derive(Serialize)]
     struct R {}
     Ok(req.json(&R {}))
+}
+
+/// Handles the `DELETE /tokens/current` route.
+pub fn revoke_current(req: &mut dyn RequestExt) -> EndpointResult {
+    let authenticated_user = req.authenticate()?;
+    let api_token_id = authenticated_user
+        .api_token_id()
+        .ok_or_else(|| bad_request("token not provided"))?;
+
+    let conn = req.db_conn()?;
+    diesel::update(api_tokens::table.filter(api_tokens::id.eq(api_token_id)))
+        .set(api_tokens::revoked.eq(true))
+        .execute(&*conn)?;
+
+    Ok(Response::builder().status(204).body(Body::empty()).unwrap())
 }
--- a/src/tests/token.rs
+++ b/src/tests/token.rs
@@ -272,6 +272,74 @@ fn revoke_token_success() {
 }
 
 #[test]
+fn revoke_current_token_success() {
+    let (app, _, user, token) = TestApp::init().with_token();
+
+    // Ensure that the token currently exists in the database
+    app.db(|conn| {
+        let tokens: Vec<ApiToken> = assert_ok!(ApiToken::belonging_to(user.as_model())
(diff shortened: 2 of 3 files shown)


mibac138:
ping @sgrif

src/controllers/token.rs
```diff
@@ -110,3 +110,21 @@ pub fn revoke(req: &mut dyn RequestExt) -> EndpointResult {
...
+pub fn revoke_current(req: &mut dyn RequestExt) -> EndpointResult {
+    let authenticated_user = req.authenticate()?;
+    let api_token_id = authenticated_user
+        .api_token_id()
+        .ok_or_else(|| bad_request("token not provided"))?;
+    let conn = req.db_conn()?;
+    diesel::delete({
+        use self::api_tokens::dsl::*;
+        api_tokens.filter(id.eq(api_token_id))
+    })
```
Turbo87, line 124: @mibac138 was there a reason why you chose to delete the token instead of setting `revoked: true` like the `revoke()` function above does it? It would probably be best to use the same behavior for both implementations unless there are strong reasons against it.

src/controllers/token.rs
```diff
@@ -110,3 +110,21 @@ pub fn revoke(req: &mut dyn RequestExt) -> EndpointResult {
     struct R {}
     Ok(req.json(&R {}))
 }
+
+/// Handles the `DELETE /tokens/current` route.
+pub fn revoke_current(req: &mut dyn RequestExt) -> EndpointResult {
+    let authenticated_user = req.authenticate()?;
+    let api_token_id = authenticated_user
+        .api_token_id()
+        .ok_or_else(|| bad_request("token not provided"))?;
```
Turbo87, line 119: could you add a test case for this "token not provided" error? 🙏

Turbo87 (changes requested):

src/controllers/token.rs
mibac138: I think I just did that by mistake, fixed now

src/controllers/token.rs
mibac138: Added as `revoke_current_token_invalid`

src/tests/token.rs
```diff
@@ -274,6 +274,64 @@ fn revoke_token_success() {
...
+fn revoke_current_token_invalid() {
+    let (_, _, user) = TestApp::init().with_user();
+
+    user.delete::<RevokedResponse>("/api/v1/tokens/current")
+        .bad_with_status(StatusCode::BAD_REQUEST)
+        .assert_error("token not provided");
+}
+
+#[test]
+fn revoke_current_token_fail() {
```
Turbo87, line 311: what is the difference between this test and `revoke_current_token_invalid()`?

Turbo87 (approved):
I've slightly adjusted the code to return an empty response with HTTP 204 (No Content), instead of the empty JSON object, and I've fixed the merge conflicts on the tests by rebasing them on master.

From my side this is good to go now.

@pietroalbini @jtgeibel any objections from your side?

hi-rustin (approved):
LGTM

jtgeibel:
LGTM!

@bors r+

PR #2677, opened 2020-08-10
Update and clean-up locked dependencies

JohnTitor:
~~Seems some dependabot PRs fail due to missing the `ajv` package (e.g. https://travis-ci.com/github/rust-lang/crates.io/jobs/369873253).~~
This updates/de-duplicates locked deps.

r? @Turbo87


Turbo87:
hmm... not quite sure I understand how that package can be missing. is some subdependency missing a dependency declaration for it? I'm not aware of us ourselves using it directly 🤔

JohnTitor:
> 
> 
> hmm... not quite sure I understand how that package can be missing. is some subdependency missing a dependency declaration for it? I'm not aware of us ourselves using it directly 🤔

Oops, I thought I saw the warning about it but it has gone after updating. Yeah, we can drop the last commit.

PR #2683, opened 2020-08-10
Fix button alignment on mobile / narrow widths

### LINKED ISSUE #2654: Buttons are misaligned on mobile / narrow widths
Looks like this with a narrow browser window:

image

I also see it on my Android phone configured with small display size (i.e. low DPI value):

image

ansmonjol:
Hello,

This PR aims to fix the alignment issue on the two main button, on mobile and narrow widths.

This issue was due to [a media query](https://github.com/rust-lang/crates.io/blob/master/app/styles/index.module.css#L17) and [a `vertical-align`](https://github.com/rust-lang/crates.io/blob/master/app/styles/shared/buttons.module.css#L43) rule conflict between `570px` and `447px`, highlighted by the second link margin-top.

Note that the `yellow-button` class is extended 9 times in the app, that's why I've preferred a `vertical-align` "reset" on the `hero-button` definition.

Didn't found a better way. Let me know if you think about something else.

BEFORE: 

image

AFTER:

image

Fixes #2654

### DIFF
--- a/app/styles/index.module.css
+++ b/app/styles/index.module.css
@@ -21,6 +21,7 @@
 
 .hero-button {
     composes: yellow-button from '../styles/shared/buttons.module.css';
+    vertical-align: initial;
 }
 
 .blurb {


Turbo87:
@bors r+

thanks @ansmonjol :)

PR #2700, opened 2020-08-16
Set the default r2d2::Pool::max_size to 3

paolobarbolini:
While working locally with the crates.io backend I got stuck for some time. Requests to `/api/v1/crates/new` would hang for 30 seconds and then a 500 error would be returned. The error was from r2d2, which timed out while trying to acquire a database connection from the pool. I finally figured out the default max pool size had been set to 1. Strangely this seems to have been this way for 3 years now, so I don't know if I'm doing something wrong here, but this fixed the issue for me.


smarnach:
I think locally you run only a single thread by default, so you shouldn't need more than one DB connection at a time? I'm not really sure about this, but there is no harm in increasing this number, in particular since it has caused problems for you.

@bors r+

Issue #2619, opened 2020-07-10
Deprecation of GitHub teams API URL format that we use


jtgeibel:
While investigating #2617, I noticed a response header linking to this blog post about [deprecation of the API URL we are using](https://developer.github.com/changes/2020-01-21-moving-the-team-api-endpoints/). It looks like we have until February 1st, 2021 to migrate to the new API path.

* [Legacy URL](https://docs.github.com/en/rest/reference/teams#get-team-membership-for-a-user-legacy) - here is [our existing usage](https://github.com/rust-lang/crates.io/blob/3dbbcb0c15f529a39fcbf5d6e95a686b439489e5/src/models/team.rs#L223) of that API.
* [New API, scoped by org](https://developer.github.com/v3/teams/members/#get-team-membership-for-a-user)

There are 2 forms of the new URL. The slug version `GET /orgs/:org/teams/:team_slug/memberships/:username` uses the org and team slugs, which we currently store in the database as `teams.login` in the form `github:org:team`. However my understanding is that with this format we'll probably also want to issue a separate API call to ensure `teams.github_id` matches the current team.

For the id based format, `GET /organizations/:org_id/team/:team_id/memberships/:username`, I don't believe that we currently track the `org_id` anywhere. Perhaps we could start tracking this and use the existing API (before it goes away) to backfill all the `org_id`s in the database.

thiagoarrais:
I'd like to tackle this in case anyone else isn't trying already. I can dedicate some time on an off over the coming weeks. I'm sure it's doable before the deprecation date.

I think I'll start by writing a PR for collecting the org_id when the teams are inserted into the database. Then we can work on backfilling the existing teams. Is that ok?

StylusEater:
@thiagoarrais I'm happy to work on this with you. I too planned to start work on it. Let me know your thoughts. I'm located in US/Eastern and could work with you over Discord this weekend.

thiagoarrais:
Nice, StylusEater. I can't work on the weekends, though. Maybe we can work together but asynchronously?

Anyway, I've opened a [(draft) PR for the initial change](https://github.com/rust-lang/crates.io/pull/2725). I'd really love to get some feedback on it. Tests are failing locally, but I'm not sure if this isn't a configuration issue.

Another possible course of action you can take is writing the backfill procedure based on this PR.

thiagoarrais:
Quick update:

[...]

PR #2725, opened 2020-08-28
Start collecting organization IDs for teams

thiagoarrais:
This is an initial change for fixing #2619. It records the extra parameter needed by the new API endpoint whenever a team owner is added to a crate.

We can't currently change the `team_with_gh_id_contains_user` function since the database doesn't hold the org_id for the old teams. For that we'll need to backfill the org_id (probably by using the deprecated [`/teams/{team_id}`](https://docs.github.com/en/rest/reference/teams#get-a-team-legacy) endpoint). After all org_ids are in place, a future PR can do that.


thiagoarrais:
Rebased to current master.

jtgeibel:
Thanks @thiagoarrais, looks great to me! I've tested this in my staging area and the newly created team included the `org_id` as expected.

@bors r+

PR #2726, opened 2020-08-29
Save lockfile state after running `npm install`

jtgeibel:
I think this is the first time I've run `npm install` locally since the switch to Renovate, but I'm not certain where this drift came from.

r? @Turbo87

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4638,7 +4638,8 @@
     "array-equal": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
-      "integrity": "sha1-jCpe8kcv2ep0KwTHenUJO6J1fJM="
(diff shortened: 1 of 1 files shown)


Turbo87:
hmm... interesting. what npm version are you running locally? maybe that could be causing the diff 🤔

jtgeibel:
I've been running node 14.x since April, but this is the first I've ended up with local changes. I'm currently on node v14.9.0 and npm 6.14.8. (The latest LTS release, node 12.8.3, appears to come with npm 6.14.6, so no much difference there.)

Turbo87:
strange. OTOH I've found several issues where people were complaining about similar issues (`dev` flags changing between `npm install` runs). never had any such issues with yarn so far. I'll see if npm has something similar to `yarn install --frozen-lockfile`.

I'll merge this for now and keep monitoring it.

@bors r+

PR #2727, opened 2020-08-29
Address new clippy suggestion

jtgeibel:
r? @JohnTitor


Turbo87 (approved):
I assume this is better because it doesn't cause a `Vec` allocation?

jtgeibel:
Yeah, I think that is the intent of the lint. The lint is `clippy::needless_collect` which looks like it landed in late 2018, so I think it is now detecting a new variation. Performance of this test though is clearly not a concern.

Turbo87:
yeah, makes sense. feel free to r=@Turbo87 if you want.

PR #2728, opened 2020-08-29
Bump to latest stable rustc and add some #[track_caller]

jtgeibel:
This should improve diagnostics of test failures. The attribute was added to one non-test function, `env()`.

r? @JohnTitor


JohnTitor (approved):
Looks good!

PR #2751, opened 2020-09-05
Fix mismatch in crate versions between pages

### LINKED ISSUE #2710: Crate version in search results doesn't match version on crate page
See for example with `clap`:

## Search Results

image

## /crates/clap

image

The version on the search page is a beta, while the version on the crate page (/crates/clap) is the latest stable release.

camelid:
Fixes #2710.

For example with `clap`, in the search results (/search?q=clap) the
crate version was shown as `3.0.0-beta.1`, whereas on the crate's page
(/crates/clap) it was shown as `2.33.3`.

The issue was that the search results were showing the *maximum* version,
not the newest version, which I believe is what the crate page was
showing.

Also added a missing "the" in a hover title.

---

@rustbot modify labels: A-frontend A-versions C-bug

### DIFF
--- a/app/components/crate-row.hbs
+++ b/app/components/crate-row.hbs
@@ -9,8 +9,8 @@
       >
         {{ @crate.name }}
       </LinkTo>
-      <span local-class="version" data-test-version>v{{@crate.max_version}}</span>
-      <CrateTomlCopy @copyText='{{@crate.name}} = "{{@crate.max_version}}"' @inline={{true}} local-class="copy-button" />
+      <span local-class="version" data-test-version>v{{@crate.newest_version}}</span>
+      <CrateTomlCopy @copyText='{{@crate.name}} = "{{@crate.newest_version}}"' @inline={{true}} local-class="copy-button" />
     </div>
     <div local-class="description" data-test-description>
       {{ truncate-text @crate.description }}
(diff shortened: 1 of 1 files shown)


camelid:
I'm new to both Ember and working on crates.io, so I would appreciate it if you could make sure that I'm understanding the crates.io versioning models correctly!

camelid:
The version template which represents the crate page (/crates/clap) seems to be referencing an attribute on the controller called `currentVersion` which is an alias for `model.version`, but I could only find `max_version` and `newest_version` in the `crate` model:

https://github.com/rust-lang/crates.io/blob/e308ecfceab4727efb59a55f6888531025812781/app/models/crate.js#L11-L12

app/components/crate-row.hbs
```diff
@@ -28,7 +28,7 @@
     <div local-class="updated-at" >
       {{svg-jar "latest-updates" height="32" width="32"}}
       <span>
-        <abbr title="The last time crate was updated">Updated:</abbr>
+        <abbr title="The last time the crate was updated">Updated:</abbr>
```
camelid, line 31: Or this maybe to match the other hover titles?

```suggestion
        <abbr title="Last time crate was updated">Updated:</abbr>
```

camelid:
It looks like there's still a version display mismatch with `inkwell`:

image

image

camelid:
So this probably shouldn't be `newest_version`, it should be whatever the crate page is using. Does the crate page display the maximum version that's not alpha/beta/etc.?

PR #2760, opened 2020-09-09
Log `in_flight_total` when the request count exceeds a threshold

jtgeibel:
The parent of this PR is what is currently deployed so we should be able to merge this and deploy `843a975` (not the merge commit) without deploying untested changes on master.

r? @pietroalbini

### DIFF
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -55,7 +55,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             if config.env == Env::Development {
                 5
             } else {
-                50
+                // A large default because this can be easily changed via env and in production we
+                // want the logging middleware to accurately record the start time.
+                500
             }
         });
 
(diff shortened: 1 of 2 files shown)


src/bin/server.rs
```diff
@@ -55,7 +55,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             if config.env == Env::Development {
                 5
             } else {
-                50
+                // A large default because this can be easily changed via env and in production we
+                // want the logging middleware to accurately record the start time.
```
jtgeibel, line 59: Medium term, we can capture the request start time in either the `conduit-hyper` or `civet` server and add it to the request extensions.

Longer term, we can support async middleware that runs before the current sync middleware. This would allow us to drop the thread count, but I don't really expect issues with high thread counts. I don't think there is a ton of work to make this happen, but low priority.

PR #2782, opened 2020-09-11
Cargo.toml: Sort dependencies alphabetically

Turbo87:
Having the dependencies arbitrarily grouped makes it hard to know where to put new dependencies. This PR sorts them alphabetically, which should resolve the problem.

r? @jtgeibel

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ cargo-registry-s3 = { path = "src/s3", version = "0.2.0" }
 chrono = { version = "0.4.0", features = ["serde"] }
 civet = "0.12.0-alpha.4"
 comrak = { version = "0.8", default-features = false }
+
 conduit = "0.9.0-alpha.3"
 conduit-conditional-get = "0.9.0-alpha.3"
 conduit-cookie = "0.9.0-alpha.4"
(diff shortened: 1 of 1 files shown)


JohnTitor (commented):
If the current is not separated by any intention, it feels like a reasonable change. But I'd like to add some break lines. For instance, the conduit-* crates should be separated by break lines so that we can find them easily.

Turbo87:
> the conduit-* crates should be separated by break lines so that we can find them easily

isn't that solved already by sorting the deps alphabetically? 🤔

JohnTitor:
> isn't that solved already by sorting the deps alphabetically? 🤔


Nah, this removes the blank lines.

Turbo87:
yeah, my point is: if the deps are sorted alphabetically, then finding the conduit crates is already pretty easy and we don't need this grouping via empty lines anymore 😉

JohnTitor:
> if the deps are sorted alphabetically, then finding the conduit crates is already pretty easy and we don't need this grouping via empty lines anymore

Hm I don't think so, they are definitely helpful for me.

Turbo87:
@JohnTitor I've added blank lines around the `conduit` crates :)

JohnTitor (approved):
Thanks, now it looks better to me! I'd like to defer to Justin's decision then.

PR #2783, opened 2020-09-11
Use `anyhow` for error handling

Turbo87:
see https://github.com/dtolnay/anyhow

This seems easier than defining our own `Error` enum, and it allows us to easily add additional context to the errors when they bubble up the stack. 

Since this is "application code" I assume that https://github.com/dtolnay/anyhow#comparison-to-thiserror applies and that `anyhow` is the better solution for us, compared to using something like `thiserror`.

r? @jtgeibel 

/cc @dtolnay


jtgeibel (approved):
This looks great to me! When I split these errors out from `AppError` I hoped to eventually pick a good library for this. This seems like a fine solution.

I'm going to wait to merge until after the next deploy, because we already have a huge backlog pending.

PR #2784, opened 2020-09-12
renovatebot: Enable `rust` updates

Turbo87:
Since it seems to have been working well for the frontend dependencies I guess we could also give it a try for the backend. If we figure out that it doesn't work as well we can always disable it again.

r? @jtgeibel

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,9 +10,6 @@
     "docker": {
         "enabled": false
     },
-    "rust": {
-        "enabled": false
-    },
     "postUpdateOptions": ["npmDedupe"],
     "packageRules": [{
         "groupName": "embroider",


jtgeibel:
I don't think we're ready to try this yet. My biggest concern is that when low-level dependencies are bumped (like libc) then basically the entire dependency tree is rebuilt. This regresses our build times during deploy and grows the size of the build cache. I think we should do just 1 or 2 `cargo update`s during a 6 week rustc cycle, and try to synchronize somewhat with when we bump to the latest stable rustc. This way we have as few slow builds as possible, and regularly (manually) prune the build cache when bumping rustc.

Unless there is a fine-grained way to schedule when the bot opens PRs, I'm afraid we will end up with a large backlog of PRs that we keep open for weeks at a time, and then when we're ready to deploy, a single `cargo update` PR would be easier to manage than individual PRs per dependency.

Turbo87:
fine by me :)

though I do think it's a bit unfortunate that our hosting situation is restricting us like that 😞

PR #2785, opened 2020-09-12
Double nginx worker connections

jtgeibel:
This should double our burst capacity during large influxes of download
traffic and reduce the occurance of H13 errors due to hitting the worker
connection limit.

With more queued connections we could potentially bump into limits on
the number of open file handles. It looks like Heroku dynos have a
ulimit on open files of:
* 10,000 for Standard-1X dynos
* 1,048,576 for Performance-M dynos

Based on these numbers, we have a margin of 1,808 file handles for
database connections, serving static files, etc. even on a Standard-1X
dyno.

r? @pietroalbini

### DIFF
--- a/src/middleware/require_user_agent.rs
+++ b/src/middleware/require_user_agent.rs
@@ -19,7 +19,8 @@ impl AroundMiddleware for RequireUserAgent {
 
 impl Handler for RequireUserAgent {
     fn call(&self, req: &mut dyn RequestExt) -> AfterResult {
-        let has_user_agent = request_header(req, header::USER_AGENT) != "";
+        let agent = request_header(req, header::USER_AGENT);
+        let has_user_agent = agent != "" && agent != "Amazon CloudFront";
         let is_download = req.path().ends_with("download");
         if !has_user_agent && !is_download {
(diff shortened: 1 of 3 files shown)


src/middleware/require_user_agent.rs
```diff
@@ -19,7 +19,8 @@ impl AroundMiddleware for RequireUserAgent {
 
 impl Handler for RequireUserAgent {
     fn call(&self, req: &mut dyn RequestExt) -> AfterResult {
-        let has_user_agent = request_header(req, header::USER_AGENT) != "";
+        let agent = request_header(req, header::USER_AGENT);
+        let has_user_agent = agent != "" && agent != "Amazon CloudFront";
```
pietroalbini, line 23: Can we put this into an environment variable?

pietroalbini (approved):

src/middleware/require_user_agent.rs
jtgeibel: @pietroalbini I've added a commit moving the string into an environment variable. If the variable is unset, it defaults to our current CDN. The CDN check and be disabled by setting `WEB_CDN_USER_AGENT` to the empty string.

PR #2787, opened 2020-09-13
Fix discord channel name in README.md

yerke:
I just tried to follow instructions to join discord channel and noticed that there is no #crates-io channel, but rather #crates-io-team.


JohnTitor (approved):
Good point, thanks!

PR #2790, opened 2020-09-13
Show snippet of 'description' in reverse dependency list

### LINKED ISSUE #1574: Show snippet of 'description' in reverse dependency list
It would be helpful if the reverse dependency list could show a snippet of the `description` in addition to the name, to give some idea of what each rev dep package does. I was looking at reverse dependencies for a few crates to try to find idiomatic code "in the wild" (ex just now: ndarray, nalgebra, and a few similar packages). The purpose of some packages can be inferred just from the name, but for others it can not be.

yerke:
Closes #1574 

Show snippet of 'description' in reverse dependency list

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 /target
-/.cargo
-/.vscode
-/.idea
+.cargo
+.vscode/
 
 # compiled output
 /dist
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /target
-.cargo
-.vscode/
+/.cargo
+/.vscode
+/.idea
 
 # compiled output
 /dist
--- a/src/views.rs
+++ b/src/views.rs
@@ -307,6 +307,7 @@ mod tests {
         let ver = EncodableVersion {
             id: 1,
             krate: "".to_string(),
+            description: Some("".to_string()),
             num: "".to_string(),
             dl_path: "".to_string(),
             readme_path: "".to_string(),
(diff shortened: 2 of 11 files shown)


yerke:
Before changes:
Screen Shot 2020-09-13 at 4 37 43 PM

After changes:
Screen Shot 2020-09-13 at 4 35 14 PM

.gitignore
```diff
@@ -1,6 +1,7 @@
 /target
-.cargo
-.vscode/
+/.cargo
+/.vscode
+/.idea
```
Turbo87, line 4: these things belong in your global gitignore file (see https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)

yerke: reverted

src/views.rs
```diff
@@ -202,6 +202,7 @@ pub struct EncodableVersion {
     pub id: i32,
     #[serde(rename = "crate")]
     pub krate: String,
+    pub description: Option<String>,
```
Turbo87, line 205: instead of adding this additional property on the JSON object that does not correspond to the database layout at all, I think it would be better to include the full `crate` resources in the response. the `version` resource already has a relationship to the `crate` in the line above this one. the only thing to look out for is to not do a N+1 database fetch when collecting the data for the response.

Turbo87 (changes requested):

Issue #2794, opened 2020-09-17
Find good placement for yank button on crate page


jtgeibel:
Per https://github.com/rust-lang/crates.io/pull/2623#issuecomment-666522853 we need to find good UI for including the yank button on a crate page.

cc @thiagoarrais

thiagoarrais:
Maybe option B from https://github.com/rust-lang/crates.io/pull/2623#issuecomment-666522853, but with the version in the button label. It may make it clearer that the action refers to the specific version, not the whole crate.

I'm a little short on time right now, but will try to make a mockup later.

jtgeibel:
> Maybe option B

@thiagoarrais I forgot to mention it, but option B was my favorite, and I like how you have added the version for clarification! I like the new mockup.

PR #2795, opened 2020-09-17
Temporarily remove yank button from crate header

jtgeibel:
Temporarily remove the yank button from the crate page until we find better placement for the button. The buttons added to the all versions page in #2623 remain.

I would like to merge this as a temporary fix to unblock deploying to production. I've opened #2794 to track adding this back.

r? @locks
cc @thiagoarrais


thiagoarrais:
I'm all for this. Haven't ever been 100% happy with this header button since I've introduced it.

jtgeibel:
I skipped the failing test and rebased on master so that I could deploy.

@bors r=thiagoarrais

PR #2800, opened 2020-09-17
Add Sentry error tracking for the frontend

Turbo87:
Up until now we've essentially been running blind. This PR adds error tracking via sentry.io to our frontend code.

- [x] This is marked as work-in-progress for now until we have updated our privacy policy document to reflect this. (https://github.com/rust-lang/core-team/issues/22)
- [x] The CSP configuration of the prod and staging servers also needs to be adjusted to allow connections to the sentry servers.

r? @pietroalbini

### DIFF
--- a/app/sentry.js
+++ b/app/sentry.js
@@ -1,0 +1,21 @@
+import * as Sentry from '@sentry/browser';
+import { Ember } from '@sentry/integrations/esm/ember';
+
+import config from './config/environment';
+
+export function init() {
+  let integrations = [];
+  if (config.environment === 'production') {
+    integrations.push(new Ember());
+  }
+
+  let isProd = location.hostname === 'crates.io';
+  let isStaging = location.hostname === 'staging.crates.io';
+  let environment = isProd ? 'production' : isStaging ? 'staging' : 'unknown';
+
+  Sentry.init({
+    environment,
+    ...config.sentry,
+    integrations,
+  });
+}
--- a/config/environment.js
+++ b/config/environment.js
@@ -70,10 +70,15 @@ module.exports = function (environment) {
     // here you can enable a production-specific feature
     delete ENV['ember-cli-mirage'];
 
+    ENV.sentry = {
+      dsn: 'https://dafb34e83c054b9289dc733ff336cabc@o242099.ingest.sentry.io/5430576',
+    };
+
     // Heroku Git Hash support
     if (process.env.SOURCE_VERSION) {
       let hash = process.env.SOURCE_VERSION.substr(0, 7);
       ENV['ember-cli-app-version'] = { version: hash };
+      ENV.sentry.release = hash;
     }
   }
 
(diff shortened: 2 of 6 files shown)


config/environment.js
```diff
@@ -70,10 +70,15 @@ module.exports = function (environment) {
     // here you can enable a production-specific feature
     delete ENV['ember-cli-mirage'];
 
+    ENV.sentry = {
+      dsn: 'https://dafb34e83c054b9289dc733ff336cabc@o242099.ingest.sentry.io/5430576',
```
pietroalbini, line 74: Is there any way to embed this in an environment variable? I'd *really* prefer not to have it in the source code.

app/sentry.js
```diff
@@ -0,0 +1,21 @@
...
+
+export function init() {
+  let integrations = [];
+  if (config.environment === 'production') {
+    integrations.push(new Ember());
+  }
+
+  let isProd = location.hostname === 'crates.io';
+  let isStaging = location.hostname === 'staging.crates.io';
+  let environment = isProd ? 'production' : isStaging ? 'staging' : 'unknown';
```
pietroalbini, line 14: I'm thinking about more robust ways to make this work, but not much came to mind :(

Turbo87: we could use an env var for this, if we can set one that is available during the build process on heroku

config/environment.js
Turbo87: we could, but it won't make much of a difference since the value will have to be sent to the browser anyway and thus can't be a secret. If we were talking about using Sentry for the backend I'd totally agree though.

if you want to prevent this DSN from being used on other pages you can set "Allowed Domains" on https://sentry.io/settings/rust-lang/projects/crates-io-frontend/ to restrict on which origins the DSN can be used.

Turbo87: btw if you look at https://sentry.io/settings/rust-lang/projects/crates-io-frontend/install/javascript-ember/ you can see that this is essentially how Sentry itself recommends to use this :wink:

pietroalbini: I'm not that worried about leaking the value, but more about being able to change it without tweaking the source code.

Turbo87: alright, just keep in mind that you will have to rebuild the frontend so that it can write the new value in the assets.

any suggestion for the env var? `SENTRY_DSN_WEB`?

pietroalbini: :+1:

Turbo87: done

app/sentry.js
jtgeibel: We could possibly check if `HEROKU_APP_NAME` contains `staging` (the variables should be available to the buildpack, but I don't know if the nodejs buildpack adds them to the environment). This would have the slight benefit of working in my personal staging environment, though I have no plans to enable Sentry there at this time so I'm fine with this check too.

jtgeibel (commented):
This looks great to me! Marked as `S-blocked` pending policy changes.

pietroalbini:
Blocked on https://github.com/rust-lang/core-team/issues/22 before deploying this to production.

jtgeibel:
Merging now that this is unblocked!

@bors r+

PR #2806, opened 2020-09-20
Use nodejs buildpack directly (instead of via the emberjs buildpack)

jtgeibel:
Since #2657 we build the frontend twice during a deploy. It seems that
the nodejs buildpack (pulled in by the emberjs buildpack) is running
our build script so lets just depend on that directly because it is
actively supported by Heroku.

r? @JohnTitor

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,4 +1,4 @@
 https://github.com/emk/heroku-buildpack-rust#578d630
-https://jtgeibel.com/emberjs.tgz
+https://github.com/heroku/heroku-buildpack-nodejs#v176
 https://github.com/heroku/heroku-buildpack-nginx.git#fbc49cd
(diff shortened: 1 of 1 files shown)


jtgeibel:
This also appears to drop the size of the compiled slug from ~270M to ~90M!

JohnTitor (approved):
> 
> 
> This also appears to drop the size of the compiled slug from ~270M to ~90M!

Wow, that's a great improvement! Looks reasonable to me.

jtgeibel:
> > This also appears to drop the size of the compiled slug from ~270M to ~90M!
> 
> Wow, that's a great improvement!

Strangely, I did not observe a similar drop on staging. I looked into some recent deploys, more info in #2809.

PR #2808, opened 2020-09-20
Add TOML syntax highlighting and fix YAML syntax highlighting

### LINKED ISSUE #2761: Add TOML syntax highlighting
That way it will look nicer when crate READMEs have something like this:

image

camelid:
Fixes #2761.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -40,7 +40,7 @@ impl<'a> MarkdownRenderer<'a> {
                 "language-scss",
                 "language-sql",
                 "language-toml",
-                "yaml",
+                "language-yaml",
             ]),
         )]);
         let sanitize_url = UrlRelative::Custom(Box::new(SanitizeUrl::new(base_url)));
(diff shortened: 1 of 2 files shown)


camelid:
## Before

image

## After

image

src/render.rs
```diff
@@ -40,7 +40,7 @@ impl<'a> MarkdownRenderer<'a> {
                 "language-scss",
                 "language-sql",
                 "language-toml",
-                "yaml",
+                "language-yaml",
```
camelid, line 43: My local dev environment isn't set up for working on the backend, so I haven't tested this, but I think this is what it should be.

camelid:
Cc @JohnTitor

src/render.rs
Turbo87: this change is unrelated to "Add TOML syntax highlighting", right?

camelid: Correct. I saw this while I was looking at the TOML stuff and I figured I would bundle them into one PR.

camelid: Before this change, YAML highlighting wasn't working:

image

And indeed the `language-yaml` class was missing.

PR #2810, opened 2020-09-21
Move away from deprecated team membership API

### LINKED ISSUE #2619: Deprecation of GitHub teams API URL format that we use

thiagoarrais:
Endpoint `/teams/:team_id/memberships/:user_name` is fading away as per https://developer.github.com/changes/2020-01-21-moving-the-team-api-endpoints/

This moves away from that in favor of the new endpoint: `/organizations/:org_id/team/:team_id/memberships/:username`.

Fixes #2619


thiagoarrais:
Created as draft to prevent acidental merging since it depends on backfilling the database with the missing organization IDs.

See #2796

@jtgeibel: any regards here?

thiagoarrais:
Rebased to latest master and removed the (now obsolete) org id backfilling binary. Let me know if you need anything else.

jtgeibel:
This is now deployed to [staging](https://staging.crates.io/). We'll need a team that we can add to a crate, and a user on that team that can test publishing a new version.

If you want to publish a crate to staging and add the team `github:conduit-rust:test` then I can try publishing. I've also added this to tomorrow's weekly meeting as we might be able to coordinate something there as well.

thiagoarrais:
I've created [the crate](https://staging.crates.io/crates/issue-2619) but can't add the team because "could not find the github team conduit-rust/test"

jtgeibel:
Oops, membership on that team was set to private, so you need to be a member to add the team. I've made that test team public and sent you an invite so you can join if that doesn't work for some reason.

thiagoarrais:
Ok. I've added the team to the crate [issue-2619](https://staging.crates.io/crates/issue-2619) now.

jtgeibel:
Alright, I've been able to publish a [0.1.1](https://staging.crates.io/crates/issue-2619/0.1.1) as a team member on staging. LGTM! Thank you so much @thiagoarrais for the series of PRs to get this fixed!

I plan to deploy to production tomorrow.

@bors r+

PR #2815, opened 2020-09-21
me.index: Remove unnecessary `created_at` assignment

Turbo87:
This attribute isn't read by the backend code, and there is no reason for us to explicitly set it.

r? @locks


Turbo87:
turns out this is used for sorting... I'll close this for now and will reopen once the creation form is no longer part of the list

PR #2819, opened 2020-09-22
Move frontend dependencies to devDependencies

jtgeibel:
The only remaining dependencies are for experimental fastboot support.

r? @Turbo87

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4955,6 +4955,7 @@
       "version": "6.26.0",
       "resolved": "https://registry.npmjs.org/babel-code-frame/-/babel-code-frame-6.26.0.tgz",
       "integrity": "sha1-Y/1D99weO7fONZR9uP42mj9Yx0s=",
+      "dev": true,
       "requires": {
         "chalk": "^1.1.3",
         "esutils": "^2.0.2",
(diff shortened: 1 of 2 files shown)


jtgeibel:
Screwed up commas in JSON twice now with this PR :crossed_fingers:

jtgeibel:
As expected, moving these two packages doesn't really have a noticeable reduction in slug size.

Turbo87:
I've started using this convention where all packages that are imported via ember-auto-import are put into `dependencies` to signal that they are technically not build-time dependencies. The way the build currently works they will still be build-time dependencies and not imported from `node_modules` at runtime, but that might change in the future.

> As expected, moving these two packages doesn't really have a noticeable reduction in slug size.

If it's not noticeable, if the change really worth it? 😅

jtgeibel:
> If it's not noticeable, is the change really worth it?

At this time, I don't think so, so closing this PR.

> I've started using this convention where all packages that are imported via ember-auto-import are put into dependencies to signal that they are technically not build-time dependencies.

I'm interested in hearing more about this. At first glance, I don't see a difference in how `pretty-bytes` and `copy-text-to-clipboard` are used compared to other dependencies, but I'm probably missing something.

Turbo87:
> I'm interested in hearing more about this. At first glance, I don't see a difference in how `pretty-bytes` and `copy-text-to-clipboard` are used compared to other dependencies, but I'm probably missing something.

the difference is that they are regular JS modules that are imported in the `app` folder, vs. using import paths that are supplied by Ember addons. but as I said, both should work fine, I've just used it as a convention in several projects so far :)

PR #2842, opened 2020-09-29
Use `fill: currentColor` in `lock` icon

Turbo87:
This makes the icon better reusable and makes it easier to introduce e.g. hover states

A PR about locks for @locks 😁 

?r @locks


locks:
😍 I'm flattered.

@bors r+

PR #2843, opened 2020-09-29
Refactor login code to use `postMessage()` instead of polling loop

Turbo87:
I got nerd-sniped by the following comment in the codebase:

> For the life of me I cannot figure out how to do this other than polling

This PR simplifies the implementation and now uses the recommended way to communicate between parent and child window: `window.postMessage()`

see https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage for more information

r? @jtgeibel


Turbo87:
FYI I'm currently deploying this to the staging environment so that we can test it out before merging

jtgeibel:
Glad to see this implemented in the right way! LGTM, and works fine on staging.

@bors r+

PR #2846, opened 2020-09-30
Remove `login` route

Turbo87:
There is no reason for this to be implemented in a dedicated route that needs to use `transition.abort()` to avoid users from actually entering it. Directly running the code without the route transition seems significantly easier.

I assume this was originally implemented this way so that we're able to use regular links in the menus, but since CSS allows us to style buttons to look like the regular menu items this is not a relevant limitation anymore. This change also prevents people from opening the login link "in a new tab", which never really worked anyway.

Deploying this to the staging env right now for QA check.

r? @jtgeibel


jtgeibel:
LGTM! @bors r+

PR #2851, opened 2020-09-30
Cleanup `OwnedCrateRow` component

Turbo87:
This PR is best reviewed commit-by-commit 😅 

Ultimately it simplifies the implementation by converting it to a Glimmer component and using a more conventional approach for the styling. The component should still look and work the same as before from the users point of view though.

r? @locks


locks (approved):
Thanks for the clean commits, made reviewing the changes much easier.
@bors r+

PR #2854, opened 2020-09-30
README: Remove broken "What's Deployed" badge

Turbo87:
This page only shows "HTTPError" so the value that this badge brings is currently rather questionable 😉

r? @jtgeibel


jtgeibel:
For me this redirects to https://whatsdeployed.io/s/9IG/rust-lang/crates.io and shows everything merged since the last deploy.

I don't personally use this, but some people may find it useful. I wonder why you're seeing an error, maybe just a temporary issue.

Turbo87:
> I wonder why you're seeing an error, maybe just a temporary issue.

looks like it, yeah 😅

PR #2857, opened 2020-10-01
Add "Email Change" acceptance tests

Turbo87:
This PR also includes a small bugfix that resets the `email` property of the user back to its original value in case the request fails.

r? @locks


jtgeibel:
LGTM! @bors r+

PR #2858, opened 2020-10-01
fix: prevent yarn usage

Gumichocopengin8:
- prevent to use yarn instead of npm
- these days, many people use yarn, so it can be possible to make a mistake to use yarn instead of npm
- in order to avoid the annoying mistake, I created this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,6 +115,7 @@
   },
   "engines": {
     "node": "^12.9.1",
-    "npm": "^6.10.2"
+    "npm": "^6.10.2",
+    "yarn": "please_use_npm"
   }
 }


Gumichocopengin8:
@bors @Turbo87 
Hello, hope you're doing well.
I'd appreciate it if you'd review this PR.
Thank you.

package.json
```diff
@@ -115,6 +115,7 @@
   },
   "engines": {
     "node": "^12.9.1",
-    "npm": "^6.10.2"
+    "npm": "^6.10.2",
+    "yarn": "please_use_npm"
```
Turbo87, line 119: what's the effect of this?

technically using yarn should be fine. the only downside would be that the lockfile is not taken into account when installing through yarn.

Gumichocopengin8: I think this helps you - https://stackoverflow.com/questions/57328593/prevent-yarn-install-from-running-in-project-i-e-force-npm-install
In my opinion , we should stick to yarn or npm, not both because of lock file. So far, this repo is using npm, so I just added a restriction that prevents using yarn.

locks: I like the idea of this PR, but the implementation seems to add even more confusing error message. Currently, if you run yarn you get a warning at the top about the presence of an npm lockfile.

Gumichocopengin8: @locks 
I don't think it is confusing. https://github.com/yarnpkg/yarn/issues/4895#issuecomment-545644733 suggests this too.
I think a warning is not sufficient to prevent yarn because some ppl don't pay attention to warnings.
 
What do you think, @Turbo87?

Turbo87: even though I'm using `npm` to install the dependencies I usually still use things like `yarn test` and `yarn start:live` because I'm used to using yarn from other projects and have several shell aliases set up for things like this.

while I appreciate the idea and effort, I think I would prefer to not merge this 😞

Gumichocopengin8: @Turbo87 
I see. That makes sense. Thank you for reviewing it anyways!
I'll close this PR.

PR #2859, opened 2020-10-01
test: add more tests for format-num

Gumichocopengin8:
- added more tests for format-num
- currently format-num test does not cover larger values and decimal values. 
- Some countries, such as India use this format `12,34,567`. However, we want to use the format like `1,234,567`. 
- That's why I added a larger value test.


Gumichocopengin8:
@bors @Turbo87 
Hello, hope you're doing well.
I'd appreciate it if you'd review this PR.
Thank you.

Turbo87:
> Some countries, such as India use this format 12,34,567. However, we want to use the format like 1,234,567.

yeah, we should at some point think about formatting in the users preferred locale instead, but for now this is fine :)

@bors r+

PR #2860, opened 2020-10-02
Use notification service for the email notification success message

Turbo87:
The `confirm` route previously only contained a single sentence and the layout looked a little broken. This PR changes the route to redirect back to the frontpage and show a success notification using our regular notification system.

Bildschirmfoto 2020-10-02 um 09 31 14

r? @locks


locks:
This feels like a cleaner approach, let's do it,
@bors r+

PR #2861, opened 2020-10-02
helpers/format-num: Format numbers in the user's preferred locale

Turbo87:
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_identification_and_negotiation 😉 

r? @locks


locks:
Nice change!
@bors r+

PR #2866, opened 2020-10-03
Move `logout` route code into `Header` component

Turbo87:
There is no reason for this to be implemented in a dedicated route. Directly running the code without the route transition seems significantly easier.

r? @jtgeibel


Turbo87:
you're right. I've cherry picked the wrong version of the commit and this one has a typo 🤦 

I'm fixing it right now and writing a test for it to prevent it from regressing in the future.

Turbo87:
https://github.com/rust-lang/crates.io/pull/2875 should fix it :)

PR #2873, opened 2020-10-04
services/google-charts: Ensure that `loadScript()` rejects with an `Error` instance

Turbo87:
Up until now the function would have rejected with an `Event` instance, which shows up in a non-useful way on Sentry. https://developer.mozilla.org/de/docs/Web/API/HTMLScriptElement suggests to use `event.target.src` instead to convert the event to an `Error` instance.

r? @jtgeibel


jtgeibel:
LGTM! @bors r+

PR #2880, opened 2020-10-05
Improve Google Charts error handling

Turbo87:
According to Sentry we have quite a few users that have trouble loading the Google Charts code. This PR improves the error handling in two ways:

1. The preloading code in the `application` route is simply ignoring any errors, since it's irrelevant for the preloading and the errors would trigger again when loaded through the `DownloadGraph` component

2. The `DownloadGraph` component is adjusted to show an error message if the loading of the Google Charts API failed, and it will no longer rethrow `ExternalScriptError` errors, which means they won't show up on Sentry anymore.

This should fix https://sentry.io/organizations/rust-lang/issues/1935278146/

r? @jtgeibel

### DIFF
--- a/app/components/download-graph.hbs
+++ b/app/components/download-graph.hbs
@@ -1,5 +1,12 @@
-<div
-  {{did-insert this.renderChart}}
-  {{did-update this.renderChart @data}}
-  ...attributes
-></div>
+{{#if this.loadTask.last.error}}
+  <span data-test-google-api-error>
+    There was an error loading the Google Charts code.
+    Please try again later.
+  </span>
+{{else}}
+  <div
+    {{did-insert this.renderChart}}
+    {{did-update this.renderChart @data}}
+    ...attributes
+  ></div>
+{{/if}}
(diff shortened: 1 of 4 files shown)


jtgeibel:
I've confirmed that if I block `https://www.google.com/jsapi` or `https://www.gstatic.com/charts/loader.js` (the former is a permanent redirect to the latter) in dev-tools then I see the error message. Allowing those requests, further resources from `www.gstatic.com` are loaded and if those are blocked it results in a missing chart with no error message. I see no issue with that, unless it's easy to intercept those errors as well.

If you want to include it in this PR, I think we should load `https://www.gstatic.com/charts/loader.js` directly. That matches the current [quick start](https://developers.google.com/chart/interactive/docs/quick_start) documentation and would avoid the request to `google.com`. (It's possible some people block `google.com` but not `gstatic.com`, and the cart should then load for them.)

Turbo87:
> If you want to include it in this PR, I think we should load https://www.gstatic.com/charts/loader.js directly.

I'd prefer to keep this PR focused on just the error display :)

Will have a look at opening a follow-up PR though. Another thing to consider is moving away from Google Charts entirely at some point, but that's a slightly bigger task 😅

jtgeibel:
> I'd prefer to keep this PR focused on just the error display :)

Sounds good to me. I've opened @2885 for your review.

@bors r+

PR #2881, opened 2020-10-05
Refactor login code

Turbo87:
This PR significantly refactors the login code on the frontend. While the diff may look a bit chaotic, when reviewed commit-by-commit it should be manageable. I'm happy to split this into multiple PRs if it seems beneficial.

The main change is that the code that is running inside the login popup window is reduced, which should ultimately speed up the login process. I've also added loading spinners to both the login and logout buttons/menu items.

One thing I couldn't do myself is related to the "Add faster `/authorize/github` alternative" commit. This introduces an additional HTML file which we can use as the redirect target for the GitHub OAuth app and has the same JS behavior as the current `/authorize/github` route of the Ember app. The advantage of the HTML file is that is does not need to boot the full Ember app, just to perform the `postMessage()` that the parent window waits for. I've kept the `/authorize/github` route for now to not introduce any breaking changes. Once this has been tested and shipped to prod we can eventually remove the route.

I'm currently deploying this branch to the staging environment for QA.

r? @jtgeibel


jtgeibel:
This looks awesome! This is currently deployed on staging (which also merges #2880).

Everything looks good in terms of code review. I ran into 2 issues during testing. When logging in, if I hit escape when the window opens and then close it, the login button remains disabled with the spinner.

Next, in order to test the static OAuth callback page, I've also deployed this to https://jtgeibel-staging-crates-io.herokuapp.com/ and changed the oath callback to `https://jtgeibel-staging-crates-io.herokuapp.com/github-redirect.html`. This appears to fail due to CSP:

> Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).

To avoid extracting the tiny script into a separate request, I think we can fix this by putting a special case in `nginx.conf.erb`. I'm deploying this to my staging environment to test it now:

```
location = /github-redirect.html {
	add_header X-Content-Type-Options nosniff;
	add_header Cache-Control public;
	root dist;
	expires 1d;
}
```

jtgeibel:
I've pushed a commit with the nginx change for your review. This is working fine for me at https://jtgeibel-staging-crates-io.herokuapp.com/

Turbo87:
> When logging in, if I hit escape when the window opens and then close it, the login button remains disabled with the spinner.

good catch! let me have a look :)

Turbo87:
@jtgeibel should be fixed by the last commit. unfortunately I had to introduce `window.closed` polling again for this... 😅

jtgeibel:
@Turbo87 this is currently deployed to staging. The error notification always triggers for me (Firefox 81), even if authorization completes.

Turbo87:
@jtgeibel I removed the `window.close()` call from the `github-redirect.html` code, and put it into the `loginTask` instead to avoid that conflict. I've tested this locally with Chrome, Firefox and Safari and it seems to work for me now. :)

jtgeibel:
LTGM! I've tested in my staging area which has been converted to use the static HTML redirect file, and on staging which has not yet been converted. Both work fine for me!

@pietroalbini or @carols10cents, would you update the settings for `staging` under GitHub Developer Settings -> OAuth Apps? It should change to `https://staging.crates.io/github-redirect.html` to enable the more efficient static redirect.

@bors r+

pietroalbini:
> @pietroalbini or @carols10cents, would you update the settings for staging under GitHub Developer Settings -> OAuth Apps? It should change to https://staging.crates.io/github-redirect.html to enable the more efficient static redirect.

Done!

jtgeibel:
> Done!

Thanks @pietroalbini! I'm preparing a production deploy now. I figure we can give that 24 hours or so, to make sure there are no unexpected issues, and then make the change there as well.

PR #2890, opened 2020-10-07
Update `node` to v14.9.0

Turbo87:
I've been using this version locally, and it seems that Heroku also supports it, so I guess there is not much reason to stay on v12 :)

r? @jtgeibel


jtgeibel:
Sounds great! I've been running with the v14 series locally since April (at the time I think the v12 packages wouldn't install on Ubuntu 20.04 beta). It looks like this goes into active LTS on 2020-10-27, so I'd say we're close enough.

@bors r+

PR #2892, opened 2020-10-08
Remove use of deprecated `set-env` and `add-path`

JohnTitor:
GitHub now warns about the use of `set-env` and `add-path`: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
This replaces them with new commands and just run `npx percy` directly instead of using percy's action since they haven't updated it (https://github.com/percy/exec-action/pull/18).
Also, I noticed the current config uploads snapshots even if it comes from a forked repo during debugging. 45866ff prevents it.

r? @jtgeibel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,10 @@ jobs:
         if: github.repository == 'rust-lang/crates.io'
         run: npx percy exec -- npm run test-coverage
 
+      - name: test-coverage
+        if: github.repository != 'rust-lang/crates.io'
+        run: npm run test-coverage
+
   backend:
     name: Backend
     runs-on: ubuntu-16.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,11 +45,10 @@ jobs:
           npm run lint:deps
 
       - name: Add nonce for parallel tests
-        run: echo ::set-env name=PERCY_PARALLEL_NONCE::`date +%s`
+        run: echo "PERCY_PARALLEL_NONCE=`date +%s`" >> $GITHUB_ENV
 
-      - uses: percy/exec-action@v0.3.0
-        with:
-          command: npm run test-coverage
+      - name: Run percy exec
+        run: npx percy exec -- npm run test-coverage
 
   backend:
     name: Backend
@@ -98,7 +97,7 @@ jobs:
               rm -v "$HOME/.rustup"
               rm -v "$HOME/.cargo"
           fi
-          echo "::add-path::$HOME/.cargo/bin"
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
 
       # Cache `diesel` binary
       #
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,9 +44,6 @@ jobs:
           npm run lint:js
           npm run lint:deps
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -45,11 +45,11 @@ jobs:
           npm run lint:deps
 
       - name: Add nonce for parallel tests
-        run: echo ::set-env name=PERCY_PARALLEL_NONCE::`date +%s`
+        run: echo "PERCY_PARALLEL_NONCE=`date +%s`" >> $GITHUB_ENV
```
Turbo87, line 48: now that we're using `percy exec`, I wonder if we actually still need the `PERCY_PARALLEL_NONCE` 🤔

.github/workflows/ci.yml
```diff
@@ -45,11 +45,11 @@ jobs:
...
 
-      - uses: percy/exec-action@v0.3.0
-        with:
-          command: npm run test-coverage
+      - name: Run percy exec
+        if: github.repository == 'rust-lang/crates.io'
```
Turbo87, line 51: why do you want to run this only on this repo? wouldn't that mean that it doesn't run for PRs coming from forked repos?

JohnTitor: Note that this rejects any builds on forked repo, not here. I usually run CI on my fork before submitting PR to make sure things but it seems its CI triggers percy, see https://github.com/JohnTitor/crates.io/pull/17 and https://percy.io/crates-io/crates.io/builds/7205884. I think it's better not to trigger the check in such a case.

Turbo87: hmm, yeah, I'm surprised it still ran on this PR, but maybe I misunderstood how GitHub does things 😅 

I'm wondering, should we still run `npm run test-coverage` if `if: github.repository != 'rust-lang/crates.io'`?

JohnTitor: `if: github.repository == 'rust-lang/crates.io'` means that "if the org name is `rust-lang` and the repo name is `crates.io`, run the below command" so this should work as expected. `if: github.repository != 'rust-lang/crates.io'` won't trigger percy here :P

Turbo87: > so this should work as expected

yeah, but I didn't expect that condition to be true for a pull request CI job from a forked repository, like for this PR

> if: github.repository != 'rust-lang/crates.io' won't trigger percy here

exactly, but it not only does not trigger Percy currently, it doesn't run the test suite at all anymore, which seems bad, no?

JohnTitor: > yeah, but I didn't expect that condition to be true for a pull request CI job from a forked repository, like for this PR

Nah, this PR is opened on **this repository** so GHA is triggered here, not on forked repo. If a PR is opened on a forked repo, this doesn't trigger percy.

jtgeibel: > I'm wondering, should we still run `npm run test-coverage` if `if: github.repository != 'rust-lang/crates.io'`?

Yeah, I think so. It seems like we should run these tests even when not triggering percy. I'm not as worried about the travis config, since I think we're close to gating on GH Actions.

JohnTitor: Okay, addressed via c8fa8ff.

jtgeibel:
Thanks! @bors r+

PR #2893, opened 2020-10-08
Implement email confirmation API calls using ember-api-actions

Turbo87:
r? @locks

### DIFF
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -13,4 +13,23 @@ export default class User extends Model {
   @attr kind;
 
   stats = memberAction({ type: 'GET', path: 'stats' });
+
+  async changeEmail(email) {
+    await this.#changeEmail(email);
+    this.store.pushPayload({
+      user: {
+        id: this.id,
+        email,
+        email_verified: false,
+        email_verification_sent: true,
+      },
+    });
+  }
+
+  #changeEmail = memberAction({
+    type: 'PUT',
+    before(email) {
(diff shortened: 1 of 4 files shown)


tests/acceptance/email-change-test.js
```diff
@@ -185,7 +185,7 @@ module('Acceptance | Email Change', function (hooks) {
 
       await click('[data-test-email-input] [data-test-resend-button]');
       assert.dom('[data-test-email-input] [data-test-resend-button]').isEnabled().hasText('Resend');
-      assert.dom('[data-test-email-input] [data-test-error]').hasText('Unknown error in resending message');
+      assert.dom('[data-test-email-input] [data-test-error]').hasText('Error in resending message: [object Object]');
```
Turbo87, line 188: This is not exactly pretty, but I didn't touch the formatting code, I just ensured that we actually return an Ember Data error from the method call. I have another PR queued up already that fixes the formatting and error display.

app/models/user.js
```diff
@@ -13,4 +13,28 @@ export default class User extends Model {
...
+      user: {
+        id: this.id,
+        email,
+        email_verified: false,
+        email_verification_sent: true,
+      },
+    });
+  }
+
+  #changeEmail = memberAction({
```
locks, line 29: fancy!

Turbo87: it worked surprisingly well 😅

PR #2894, opened 2020-10-08
crate.version: Fix `latestUnyankedVersion` bug

Turbo87:
`latestStableVersion` is already guaranteed to be `null` because of the previous condition. I assume the original intention here was to check for `latestUnyankedVersion` instead.

It looks like this bug was already present in the original implementation in https://github.com/rust-lang/crates.io/pull/1857

r? @jtgeibel


jtgeibel:
Good catch. @bors r+

PR #2901, opened 2020-10-10
Switch from npm to yarn

Turbo87:
As discussed in the meeting on friday, yarn currently has better UX and allows us to use `resolutions` to work around transitive dependency version conflicts. This should unblock https://github.com/rust-lang/crates.io/pull/2699, for example.

r? @jtgeibel


jtgeibel:
LGTM, but needs to be updated for #2909. Feel free to r=me after that.

PR #2929, opened 2020-10-20
styles: Prepend `normalize.css` code

Turbo87:
This should be included as the very first part of our CSS code so that the following code can override it if necessary

r? @locks


locks:
Sounds good!

@bors r+

PR #2930, opened 2020-10-20
robots.txt: Disallow robots from indexing our raw API resources

Turbo87:
see http://www.robotstxt.org/robotstxt.html for more info on this topic

/cc @Mark-Simulacrum 

r? @jtgeibel

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,3 +1,3 @@
 # http://www.robotstxt.org
 User-agent: *
-Disallow:
+Disallow: /api


locks:
This seems like a good idea.

Turbo87:
I've found conflicting information on this, but some of the pages did indicate that blocking `/api` here would also disallow it for fetch requests on the legitimate pages. I'm closing this for now unless somebody is able to confirm that it's not the case or we can find a better/working way to implement this.

PR #2932, opened 2020-10-20
crate.version: Navigate to `index` route if version can't be found

Turbo87:
Staying on an empty page seems bad, so let's navigate the user to the `index` route instead.

r? @locks


Turbo87:
nevermind, this doesn't make sense 😅

PR #2941, opened 2020-10-21
styles/application: Set minimum height for the `body` element

Turbo87:
This causes our footer to always appear at the bottom of the page, irregardless of how long the content is:

Bildschirmfoto 2020-10-21 um 16 38 46

r? @locks


locks:
The new Percy screenshots look much better, great tweak!
@bors r+

PR #2944, opened 2020-10-21
Extract and clean up `Settings::ApiTokens` component

Turbo87:
Until now the "Create API Token" form was rendered as a list item in the same list as all the regular API tokens. This PR extracts the form from the list to make the list rendering more maintainable.

r? @locks


locks:
Nice cleanup!

@bors r+

Issue #1504, opened 2018-09-28
Compile to a monolithic binary


ishitatsuyuki:
This project is currently in the structure of a gigantic library plus a few manage commands.

The issue is that the binaries are very fat and having multiple of them eats up disk space quickly. `libcargo_registry` on debug build takes over 100MB on its own, and it's linked to 8 different binaries. My target directory is easily over 2GB.

Also, this approach is no good for deployment as well. It leads to larger uploaded bundle, although gzip should help here.

Instead, we should have one monolithic binary and handle all the subcommands with [clap](https://clap.rs). As a random note, Clap 3.x in development allows more modular code because it [removes the facade between App and Subcommand](https://github.com/clap-rs/clap/issues/1037).

jtgeibel:
I think consolidating into two binaries (maybe `server` and `ops`) would be a good goal, though we can make incremental progress by combining just a few binaries at a time.

Right now, when running `cargo test`: `lib.rs`, all binaries, and `all.rs` are built in both debug and test profiles. (Cargo assumes that tests may exec any binary, so regular binaries are build in addition to the test binaries.) None of our binaries include tests so this is a lot of extra build time that doesn't add useful feedback to the development cycle.

Even when linking with `lld` via `RUSTFLAGS="-C link-arg=-fuse-ld=lld"`, all this extra linking adds to the build time.

By consolidating into fewer binaries, it should improve build time and RAM consumption during local development.

PR #2945, opened 2020-10-22
WIP: Refactor workspace layout

Turbo87:
Having a single `src` folder for one package, but then another package inside of it (`s3`), and then also two bins that rely on a module in a subfolder of `src/bin` has confused me quite a bit. This PR is an attempt to move to a more standardized workspace layout, with a single top-level `src` folder that contains the `cargo-registry`, `cargo-registry-s3` and `on-call` packages.

This should allow us to easily add more internal packages in the future. One example would be to pull the database/schema-related code into its own package, so that the `on-call` package does not have to pull in all the unrelated code from `cargo-registry`.

r? @jtgeibel


Turbo87:
looks like *something* is still missing to make the tests work. not sure what though... 😅

JohnTitor:
Note that this causes a lot of conflicts and diff-noise. I _really_ oppose this unless you strongly need it.
The current structure is a common pattern enough (putting the _main_ crate to `src` and binary utils as `bin`, though `s3` is indeed unclear a bit, but it can be addressed by the below).

> two bins that rely on a module in a subfolder of src/bin

It makes sense for me as the `on_call` module is for bin utils.

> This should allow us to easily add more internal packages in the future. 

I don't think so. We can add them easily enough even now, like:

```toml
[workspace]
members = [
  ".", # `cargo-registry`
  "yet_another_crate", # `./yet_another_crate`
  "s3" # we could also move the `s3` dir to the root and do this, if you want. Currently, it's equal to `src/s3`.
]
```

> but then another package inside of it (s3)

This can be resolved by the above workspace config, we don't have to do huge moving like this.

jtgeibel:
I have some concerns with this as well. I recall reading that IDE support for workspaces was incomplete, but I can't remember if that was RLS and/or rust-analyzers. It's also possible those issues have been addressed by now, as my recollection is probably from 1+ year ago.

> two bins that rely on a module in a subfolder of src/bin

Yeah, this is a little strange. I think this is best addressed by our plan to consolidate info fewer bins. Linking all these bins uses lots of RAM and is slow even when using lld (`RUSTFLAGS="-C link-arg=-fuse-ld=lld"`). I think `server` and `ops` binaries would be sufficient. (See also #1504 on the developer experience roadmap.)

> but then another package inside of it (`s3`)

It would be nice if we could switch to an external crate for this functionality. Alternatively, I'd be fine absorbing this file into our main library. I don't think there is much benefit to keeping this as a separate crate within our tree.

> One example would be to pull the database/schema-related code into its own package [...]

I think this is a great idea. Given the database dumps, it's possible there would be benefit to publishing this as a public crate that consumers of our exports could use.

I agree with @JohnTitor and would prefer to avoid moving all the existing code under `src/cargo-registry`. In particular, so that it is easier to back this out if we do run into issues with IDE support. I think an initial trial with the `s3` sub-crate would be a good start.

Turbo87:
sounds reasonable. thanks for the details :)

PR #2946, opened 2020-10-22
Remove turbofishes from diesel calls

Turbo87:
This makes the code a little more readable and helps e.g. `intellij-rust` in understanding the types of the variables better

r? @jtgeibel


jtgeibel:
I poked around a little bit with rust-analyzer and don't see any regressions in type inference there. In a few of these cases I like the turbofish version a bit more, but consistency and good IDE experiences seems more important.

@bors r+

PR #2948, opened 2020-10-22
Extract `on_call` and `tools` subprojects

Turbo87:
This PR is roughly similar to https://github.com/rust-lang/crates.io/pull/2945 and potentially replaces it. 

@JohnTitor does this address your concerns? :)

r? @jtgeibel


src/tools/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
JohnTitor, line 1: You said:
> Having a single src folder for one package, but then another package inside of it (s3)

So, if you're confused by putting all the crates in `src`, shouldn't `tools` be at the root? And I'd avoid `src/tools/src/bin/*.rs` (notice two `src`s!).

I imagine:
```
.
├── s3 (if you want, but it integrates `cargo-registry` so I think it's better to be in `src`)
├── src (a.k.a. `cargo-registry`)
└── tools (`src/bin` previously. I guess `utils` is better? `src/bin` is a common structure, but if we have multiple crates, it might be better to be at root for clarify.)
	└── on_call(if you want to separate it. It should be under `src/bin` since it's only for bin utils.)
```

src/tools/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "tools"
+version = "0.0.0"
+authors = ["Alex Crichton <alex@alexcrichton.com>"]
```
JohnTitor, line 4:
```suggestion
authors = ["The Rust Project Developers"]
```

nit: Should be?

src/tools/Cargo.toml
Turbo87: we can also move them to the root folder, but I wanted to avoid cluttering the root folder with all those package folders for now.

JohnTitor: Then, it may be good to have `crates/` dir, like:
```
.
└── crates
	├── cargo-registry
	├	└── s3
	└── tools (or `utils`)
		└── on_call
```

src/tools/Cargo.toml
Turbo87: idk, I just copied it from the other `Cargo.toml` files 😉

jtgeibel:
I commented on the other PR as well, before seeing this. Just wanted to reiterate that I think it would be better to consolidate these tools into a single binary, rather than extract them into a separate crate.

What I have in mind is a `src/bin/ops.rs` binary, with subcommands in `src/bin/ops/*.rs`.

Turbo87:
alright, closing this for now while working on the new direction :)

Issue #1162, opened 2017-11-08
Add tarpulin for code coverage


vignesh-sankaran:
Just came across this tool [tarpulin](https://github.com/xd009642/tarpaulin) that is an in-development code coverage tool built in Rust. It still has a lot of ground to cover, but perhaps it's worth a try and we could report issues upstream?

PR #2951, opened 2020-10-23
CI: Run tests via `cargo-tarpaulin`

Turbo87:
This PR adds https://github.com/actions-rs/tarpaulin to our CI pipeline to display a test coverage report.

Resolves https://github.com/rust-lang/crates.io/issues/1162

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,10 +197,8 @@ jobs:
           cargo fmt -- --check
           cargo clippy --all-targets --all-features --all
 
-      - name: Run cargo-tarpaulin
-        uses: actions-rs/tarpaulin@v0.1.3
-        with:
-          version: '0.17.0'
+      - run: cargo install cargo-tarpaulin
+      - run: cargo tarpaulin
 
       - name: Prune unnecessary cache
         run: script/ci/prune-cache.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,8 +197,17 @@ jobs:
           cargo fmt -- --check
           cargo clippy --all-targets --all-features --all
 
-      - run: cargo install cargo-tarpaulin
-      - run: cargo tarpaulin
+      - name: Install cargo-tarpaulin
+        if: matrix.rust == 'stable'
+        run: which cargo-tarpaulin || cargo install cargo-tarpaulin
+
+      - name: Run tests (with coverage report)
+        if: matrix.rust == 'stable'
+        run: cargo tarpaulin
+
+      - name: Run tests
+        if: matrix.rust != 'stable'
+        run: cargo test
 
       - name: Prune unnecessary cache
         run: script/ci/prune-cache.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
       DATABASE_URL: postgres://postgres:postgres@localhost/cargo_registry_test
       TEST_DATABASE_URL: postgres://postgres:postgres@localhost/cargo_registry_test
       CARGO_INCREMENTAL: 0
-      RUSTFLAGS: "-C debuginfo=0 -D warnings"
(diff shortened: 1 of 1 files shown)


jtgeibel:
cc @pietroalbini as I think you had some concerns about running 3rd party actions on CI.

pietroalbini:
In general I'd prefer if we avoid using third-party actions, as those could be yet another avenue to inject malicious code into the build.

Turbo87:
can we make them first-party actions?

JohnTitor:
If the issue is to use "the third party actions", we could use `cargo-tarpaulin` directly.

pietroalbini:
Without using the external action this looks good on my side!

Turbo87:
unfortunately this still has two issues:

1) it doesn't actually work (yet?): 0.00% coverage, 0/2661 lines covered

2) it currently takes 12 minutes to install tarpaulin on CI and then run the tests in it, which seems like quite a lot 😞

JohnTitor:
> it doesn't actually work (yet?): 0.00% coverage, 0/2661 lines covered

Hmm, I'm not sure at all.

> it currently takes 12 minutes to install tarpaulin on CI and then run the tests in it, which seems like quite a lot

So, in general, releases don't happen that often and it'd solve the issue to save the binary in the cache.

Turbo87:
@xd009642 any clue what might be the problem here? 🙏

xd009642:
oh sorry I realise I've neglected the issue my end. I remember when I tried it last I was struggling to get the backend tests working on fedora but I've got an ubuntu laptop here so I'll try again. I'll tag you on https://github.com/xd009642/tarpaulin/issues/601 if I have any questions just to avoid filling up the PR comments :+1:

xd009642:
Couple of comments on the PR from working on sorting the coverage issue in github actions:

1. Either add `--force` to the cargo install or check for the existence of tarpaulin first otherwise the install fails 
2. You probably only want to run coverage for one specific toolchain, otherwise you'll have multiple reports per commit and I'm not sure if any services like coveralls or codecov support that

Turbo87:
looks like this is ready to go now! thanks again @xd009642 :)

r? @pietroalbini

.github/workflows/ci.yml
```diff
@@ -195,7 +195,16 @@ jobs:
           cargo fmt -- --check
           cargo clippy --all-targets --all-features --all
 
-      - name: Test
+      - name: Install cargo-tarpaulin
+        if: matrix.rust == 'stable'
+        run: which cargo-tarpaulin || cargo install cargo-tarpaulin
```
JohnTitor, line 200: This workaround is still needed on CI? I think recent cargo doesn't require it anymore.

xd009642: It failed for me running the CI twice consecutively yesterday on my test project to debug the 0% coverage https://github.com/xd009642/crates-io-test/actions

JohnTitor: Hm, that's interesting, I guess cargo recognizes it as an untracked file since it's introduced by cache (or another reason) 🤔. Thanks for clarifying, then I'm fine as-is.

simonsan: Wouldn't it be reasonable to cache `tarpaulin`? There is no need to rebuild it CI run I feel?

https://github.com/marketplace/actions/cache#skipping-steps-based-on-cache-hit
https://github.com/actions/cache/blob/main/examples.md#rust---cargo

Turbo87: AFAICT https://github.com/rust-lang/crates.io/blob/2b3bdf116729f34b8d767199018beb5efc5ce3c1/.github/workflows/ci.yml#L119-L123 should already cache the binaries including `diesel-cli` and `cargo-tarpaulin`

simonsan: Ahh, sorry then. Must have overlooked that part :/

jtgeibel:
Alright, lets try this out on master and we'll see how well caching works on new PRs.

@bors r+

PR #2953, opened 2020-10-23
Update `oauth2` and remove `failure` dependency

Turbo87:
This PR updates the `oauth2` dependency to the latest v4 pre-release (see https://github.com/ramosbugs/oauth2-rs/issues/99). This allows us to finally get rid of the `failure` dependency, because `oauth2` v3 was our last dependency using it.

r? @jtgeibel


pietroalbini:
r? @pietroalbini 

I'm slightly worried about upgrading to a pre-release for a library related to authentication. Is there a reason for this other than just removing failure?

Turbo87:
> I'm slightly worried about upgrading to a pre-release for a library related to authentication. Is there a reason for this other than just removing failure?

nope, that was my only reason. and I generally share your concern about the pre-release version. I looked at https://github.com/ramosbugs/oauth2-rs/issues/99 and saw that the list of v4 was almost done though, and since the update was building fine and the test suite passing I was somewhat confident in this. we can also move this to WIP status though and wait for their final release.

pietroalbini:
Yeah let's wait until the final release.

Turbo87:
closing for now until a stable oauth2 release exists

PR #2961, opened 2020-10-26
Report errors and slow requests on the backend to Sentry

Turbo87:
This PR implements initial Sentry integration for our backend API via https://github.com/getsentry/sentry-rust.

It will automatically report panics to Sentry, as well as internal server errors (unfortunately without stacktrace for now) and slow requests, similar to our current `println!()` logging.

r? @pietroalbini

### DIFF
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -2,12 +2,11 @@
 //! information that we care about like User-Agent
 
 use super::prelude::*;
+use crate::middleware::current_user::TrustedUserId;
 use crate::util::request_header;
 use conduit::{header, Host, RequestExt, Scheme, StatusCode};
-use sentry::protocol::IpAddress;
 use sentry::Level;
 use std::fmt::{self, Display, Formatter};
-use std::net::IpAddr;
 use std::time::Instant;
 
 const SLOW_REQUEST_THRESHOLD_MS: u64 = 1000;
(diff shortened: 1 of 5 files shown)


src/middleware/log_request.rs
```diff
@@ -60,6 +65,104 @@ pub fn add_custom_metadata<V: Display>(req: &mut dyn RequestExt, key: &'static s
...
+                None => Some(req.remote_addr().ip()),
+            };
+
+            let user = sentry::User {
+                ip_address: ip_addr.map(IpAddress::Exact),
+                ..Default::default()
+            };
+
+            scope.set_user(Some(user));
+        }
```
pietroalbini, line 113: We should not send the IP over to Sentry. Maybe let's set the id and username when the user is logged in?

src/middleware/log_request.rs
```diff
@@ -60,6 +65,104 @@ pub fn add_custom_metadata<V: Display>(req: &mut dyn RequestExt, key: &'static s
...
+
+        {
+            let headers = req
+                .headers()
+                .iter()
+                .map(|(k, v)| (k.to_string(), v.to_str().unwrap_or_default().to_string()))
```
pietroalbini, line 119: We should scrub sensitive headers out of the data we send to sentry. Headers that come to mind are `X-Real-Ip`, `Authorization` and `Cookie`.

pietroalbini (changes requested):
The integration code looks good! I think we should make a couple changes to avoid sending sensitive data over to Sentry, and I want to test in staging how it interacts with CloudFront (as it'd be really useful to be able to search events by CloudFront Request ID).

src/middleware/log_request.rs
Turbo87: okay. I just added it because it's currently being logged to the console too, but that should be easy to remove. I'll see if we can add the user id/name instead.

src/middleware/log_request.rs
Turbo87: done

src/middleware/log_request.rs
Turbo87: done

src/middleware/log_request.rs
```diff
@@ -60,6 +64,104 @@ pub fn add_custom_metadata<V: Display>(req: &mut dyn RequestExt, key: &'static s
...
+            let user = sentry::User {
+                id,
+                ..Default::default()
+            };
+
+            scope.set_user(Some(user));
+        }
+
+        {
+            let filtered_headers = vec!["Authorization", "Cookie", "X-Real-Ip"];
```
pietroalbini, line 112: Let's move this to a constant at the top of the file, like `SENTRY_REMOVED_HEADERS: &[&str]`.

pietroalbini (approved):
r=me with the nit fixed!

src/middleware/log_request.rs
Turbo87: done

PR #2962, opened 2020-10-26
Use `claim` to simplify assertions

Turbo87:
This PR introduces https://github.com/svartalf/rust-claim/ for our tests to replace our custom `t!()` macro and simplify/improve a few other `assert!()` usages.

`t!()` is certainly shorter than `assert_ok!()`, but I think the latter expresses the intent way better than the former. With `t!()` is was always thinking that this might have something to do with translations... 😅 

r? @jtgeibel

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,6 +271,7 @@ dependencies = [
  "cargo-registry-s3",
  "chrono",
  "civet",
+ "claim",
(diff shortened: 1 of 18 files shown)


jtgeibel:
I especially like `assert_some_eq!`, LGTM!

@bors r+

PR #2971, opened 2020-10-29
Fix version logic when there are only pre-releases

jtgeibel:
It looks like fixing the bug in #2894 exposed this bug. When there are
only pre-release versions of a crate, the version number needs to be
extracted so that the version object can be located again a few lines
later. This could probably be refactored, but for now this removes the
spurious error notification shown in the UI.

r? @Turbo87


Turbo87:
ouch... sorry about that. I thought I had written tests for this part of the code but apparently I missed this case 🤦 

@bors r+

PR #2977, opened 2020-10-29
Cleanup crate version downloads processing code

Turbo87:
This gets rid of a few instances of unnecessary `Date` parsing. In most cases we can just use the ISO8601 string directly, without having to convert to `Date` and back to a string all the time.

r? @locks


locks:
Thanks, that tidies it up nicely!

@bors r+

PR #2982, opened 2020-10-30
Remove obsolete `sass` dependency

Turbo87:
We don't use SASS anymore...

Closes #2981

r? @locks


locks:
Woooo 😍 No more "recompiling" node-sass.

@bors r+

PR #2983, opened 2020-10-30
Add "compilers" category.

### LINKED ISSUE #1412: Create a new category for "language interpreters, compilers or run-times"
I am developing a Scheme interpreter and I want to publish my project on `crates.io`.  However when trying to file my project under a category none actually matches (not even closely) the actual nature of the project.

For example, from the current [category index](https://crates.io/categories) on `crates.io`:
* `command line utilities` -- although my Scheme interpreter is used as a command line tool, it's not an actual "command line utility", just as `rustc` or `python` are not;
* `development tools` -- one might say that any interpreter or language compiler is just a "development tool", however no one is referring to `rustc` or `python` as a "development tool";
* `emulators` -- one might say that an interpreter "emulates" an abstract language, however most crates in this category are actually native assembler emulators;

Therefore I think a new category should be added for proper "interpreters, compilers and run-times" for various languages people are implementing in Rust.

mental32:
Sub categories include: transpilers, interpreters, and compilers.

I decided to follow the precedence set by `"parser-implementations"` and `"parsing"` for these categories since they felt appropriately related.

Resolves #1412

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -285,27 +285,6 @@ description = """
 Compilers, interpreters, transpilers implemented for particular programming languages.\
 """
 
-[language-implementations.categories.transpilers]
-name = "Transpiler implementations"
-description = """
-Transpilers deal with source to source translations, such as\
-Python <-> Rust
-"""
-
-[language-implementations.categories.compilers]
-name = "Compiler implementations"
-description = """
-Compilers deal with source to machine translations, like compiling\
-Rust for x86 architectures.
-"""
-
-[language-implementations.categories.interpreters]
-name = "Interpreter implementations"
-description = """
-Interpreters deal with source to some IR translations, such as AST or\
-bytecode based interpreters like CPython or Rust's MIRI.
-"""
-
 [localization]
 name = "Localization (L10n)"
 description = """
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -64,7 +64,7 @@ the results.\
 [compilers]
 name = "Compilers"
 description = """
-Compiler implementations, for translating high-level languages into lower level ones.\
+Compiler implementations, including interpreters and transpilers.\
 """
 
 [command-line-interface]
@@ -283,13 +283,6 @@ name = "Internationalization (i18n)"
 description = """
 Crates to help develop software capable of adapting to various \
 languages and regions.\
-"""
-
-[interpreters]
-name = "Interpreters"
-description = """
-Interpreter implementations, without requiring them previously to have been compiled \
-into a machine language program.\
 """
 
 [localization]
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -279,6 +279,33 @@ Crates to help develop software capable of adapting to various \
 languages and regions.\
 """
 
+[language-implementations]
+name = "Language implementations"
+description = """
+Compilers, interpreters, transpilers implemented for particular programming languages.\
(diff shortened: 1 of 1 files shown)


src/boot/categories.toml
```diff
@@ -279,6 +279,33 @@ Crates to help develop software capable of adapting to various \
...
+Rust for x86 architectures.
+"""
+
+[language-implementations.categories.interpreters]
+name = "Interpreter implementations"
+description = """
+Interpreters deal with source to some IR translations, such as AST or\
+bytecode based interpreters like CPython or Rust's MIRI.
+"""
+
```
Turbo87, line 308:
```suggestion
```

we discussed this in one of our team meetings and since the lines between these are quite blurry we decided that we'd prefer to only have a single category without subcategories for now.

src/boot/categories.toml
```diff
@@ -279,6 +279,33 @@ Crates to help develop software capable of adapting to various \
 languages and regions.\
 """
 
+[language-implementations]
+name = "Language implementations"
```
Turbo87, line 283: I'm not sure yet if "Language implementations" is the best name for this. Some comments in the original issue proposed "Programming languages", which might be a better fit, but is also quite long. I'm wondering if there is a better alternative to these.

mental32: I can't think of a better name sadly but I'm not interested in nitpicking this so I'm ok with whatever the popular consensus is :D

Turbo87: I think what makes this not fit well is that someone looking for interpreters would look for "interpreter" and if they don't find it they might look for "compiler", but it seems unlikely for them to search for something called "language implementations".

"Transpilers" are ultimately just "Compilers" where input and output language is essentially the same, so we're down to deciding between "Compilers" and "Interpreters". As I mentioned above, someone looking for "Interpreters" might consider looking in the "Compilers" category, but someone looking for a compiler seems less likely to look in a category called "Interpreters".

Due to the logic above I think I would recommend renaming this caregory to "Compilers". It is technically not entirely correct for interpreters, but since we can't seem to be able to find a good category name that includes all of these concepts, I think this is the best we can do now.

mental32: I've removed the `language-implementations` category and have added two new categories "compilers" and "interpreters"

Turbo87: hmm, I don't think we would want to have two dedicated categories. The difference between compilers and interpreters is a somewhat grey area with JIT compilers being built into interpreters and stuff like that.

mental32: Should I revert to the `language-implementations` category and include `interpreters` and `compilers` as subcategories?

Also JITs are not normally used by themselves are are usually embedded into a larger process i.e. an interpreter runtime. So I don't think that should affect the other categories but instead it may make sense to have a JIT category of its own.

Turbo87: As I mentioned above, I think having a single category called "Compilers" which includes compilers, interpreters and transpilers would probably work best.

mental32: > I would recommend renaming this category to "Compilers".

My bad I didn't see this

Turbo87 (approved):
we'll discuss this one last time on the weekly on Friday and then we can hopefully have this merged :)

Turbo87:
@mental32 we just discussed this and will merge this in the next days. since a change like this will make the rollback of a bad deployment harder we would like to first deploy the other queued up changes before merging this. :)

PR #2988, opened 2020-10-31
Add a log message if a download is not counted

jtgeibel:
The implementation is a bit more complicated than hoped, because
lifetimes.

r? @Turbo87


Turbo87 (approved):
LGTM 👍 

I don't suppose there is an easy way to test this?

jtgeibel:
> I don't suppose there is an easy way to test this?

I was just able to test locally with `READ_ONLY_MODE=1 cargo run --bin server`. It should be possible to add an automated test as well, but I think that's probably more effort than its worth.

PR #2989, opened 2020-10-31
Update `hmac` and `env_logger`

jtgeibel:
r? @JohnTitor


JohnTitor (approved):
Should be fine, LGTM :)

PR #2990, opened 2020-10-31
Update to the latest semver

jtgeibel:
The new release of semver treats `>= 0` the same as the `*` wildcard.
It makes sense that these are equivalent, but this did break 3 tests
and will technically reject syntax in `Cargo.toml` that was previously
accepted.

r? @ghost

### DIFF
--- a/src/tests/builders/dependency.rs
+++ b/src/tests/builders/dependency.rs
@@ -15,7 +15,7 @@ impl DependencyBuilder {
             explicit_name_in_toml: None,
             name: name.to_string(),
             registry: None,
-            version_req: u::EncodableCrateVersionReq(semver::VersionReq::parse(">= 0").unwrap()),
+            version_req: u::EncodableCrateVersionReq(semver::VersionReq::parse("> 0").unwrap()),
         }
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -306,7 +306,7 @@ dependencies = [
  "rand",
  "reqwest",
  "scheduled-thread-pool",
- "semver 0.10.0",
+ "semver 0.11.0",
  "sentry",
  "serde",
  "serde_json",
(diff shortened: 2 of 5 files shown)


src/tests/builders/dependency.rs
```diff
@@ -15,7 +15,7 @@ impl DependencyBuilder {
             explicit_name_in_toml: None,
             name: name.to_string(),
             registry: None,
-            version_req: u::EncodableCrateVersionReq(semver::VersionReq::parse(">= 0").unwrap()),
+            version_req: u::EncodableCrateVersionReq(semver::VersionReq::parse("> 0").unwrap()),
```
jtgeibel, line 18: Added to the weekly meeting to discuss if disallowing `>= 0` (as we do for the equivalent `*` wildcard) is an acceptable change.

jtgeibel:
I've rebased and updated the error message. I've also run the following commands to determine crates already using a `>= 0` constraint.

```console
$ cd ~/.cargo/registry/index/github.com-1ecc6299db9ec823/
$ git grep --count '">= 0"' origin/master
origin/master:as/ki/askii:1
origin/master:co/ar/coarsetime:7
origin/master:co/un/count-min-sketch:2
origin/master:cu/rl/curl-sys:31
origin/master:cu/rs/cursive_buffered_backend:4
origin/master:do/h-/doh-proxy:8
origin/master:ju/mp/jumphash:4
origin/master:li/bg/libgit2-sys:38
origin/master:li/bs/libssh2-sys:16
origin/master:na/ts/nats:5
origin/master:pr/iv/privdrop:5
origin/master:vm/-m/vm-memory:2
$ git grep --count '">=0"' origin/master
origin/master:js/on/jsonschema:1
origin/master:vm/-m/vm-memory:5
```

I've checked the index for all of these, and `jsonschema` and `vm-memory` are the only 2 crates where the latest release uses the bad constraint. `jsonschema` has a dev-dependency on `mockito >=0` and `vm-memory` a dev-dependency on `matches >=0`.

r? @pietroalbini

dtolnay:
Is it possible that this update regressed the rendering of dependency version requirements? https://github.com/rust-lang/crates.io/issues/3047

Turbo87:
that does seem likely, yeah. certainly not an intentional change. 😞

PR #2991, opened 2020-10-31
Improve backend integration with Sentry

pietroalbini:
This PR introduces multiple improvements to our backend Sentry integration:

* The `X-Forwarded-For` header is stripped before sending an event, as that contains PII.
* Slow requests are not sent to Sentry anymore, as those alerts are not really actionable.
* The current git commit is included in every event sent to Sentry, to be able to identify releases.

r? @Turbo87

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,13 @@
 use diesel::prelude::*;
 use diesel_migrations::run_pending_migrations;
 use std::env;
+use std::error::Error;
 
 fn main() {
+    if let Err(err) = load_git_commit() {
+        println!("cargo:warning=failed to load git commit: {}", err);
+    }
+
     println!("cargo:rerun-if-env-changed=TEST_DATABASE_URL");
     println!("cargo:rerun-if-changed=.env");
     println!("cargo:rerun-if-changed=migrations/");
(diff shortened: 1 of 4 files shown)


Turbo87:
sweet! was wondering how we could get the commit info in there already :)

@bors r+

build.rs
```diff
@@ -1,8 +1,13 @@
 use diesel::prelude::*;
 use diesel_migrations::run_pending_migrations;
 use std::env;
+use std::error::Error;
 
 fn main() {
+    if let Err(err) = load_git_commit() {
+        println!("cargo:warning=failed to load git commit: {}", err);
```
jtgeibel, line 8: Note that this warning is triggered when deploying to heroku. I forget the details, but I think the git repo is not available during the build stage. We can obtain the hash at runtime on heroku like we do in the [`site_metadata` endpoint](https://github.com/rust-lang/crates.io/blob/ec91dd3a2b55d22a374338f9cb8f6937c0a620c4/src/controllers/site_metadata.rs#L9).

Of course this heroku environment variable is not available during local testing, but when testing locally with sentry reporting an "unknown" commit is probably fine.

PR #2992, opened 2020-10-31
Use `HEROKU_SLUG_COMMIT` if available

Turbo87:
This should hopefully address https://github.com/rust-lang/crates.io/pull/2991#pullrequestreview-521146500

r? @jtgeibel

### DIFF
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -39,9 +39,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                     .expect("SENTRY_ENV_API must be set when using SENTRY_DSN_API"),
             );
 
-            if let Ok(commit) = dotenv::var("HEROKU_SLUG_COMMIT") {
-                opts.release = Some(commit.into());
-            }
+            opts.release = dotenv::var("HEROKU_SLUG_COMMIT").ok().map(Into::into);
 
             sentry::init(opts)
         });
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -39,9 +39,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                     .expect("SENTRY_ENV_API must be set when using SENTRY_DSN_API"),
             );
 
-            if let Some(commit) = option_env!("CRATES_IO_GIT_COMMIT") {
-                opts.release = Some(commit.into());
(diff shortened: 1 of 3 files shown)


jtgeibel:
I think we should drop the changes to `build.rs` as well, because I don't want to see the warning on every deploy. We could drop just the warning, but I'm not sure there is much value in including the hash when doing local dev with sentry enabled. For example, if I'm testing local changes to my working directory, the build script will just report the hash of the last commit. (In the past, I've used something like `git describe --always` which will add `-dirty` to the end of the hash, but I don't think that's necessary here.)

Turbo87:
yeah, that sounds reasonable. I've added another commit that removes it again.

src/bin/server.rs
```diff
@@ -38,7 +38,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                     .map(Cow::Owned)
                     .expect("SENTRY_ENV_API must be set when using SENTRY_DSN_API"),
             );
-            opts.release = option_env!("CRATES_IO_GIT_COMMIT").map(Into::into);
+
+            if let Ok(commit) = dotenv::var("HEROKU_SLUG_COMMIT") {
+                opts.release = Some(commit.into());
+            }
```
jtgeibel, line 44: One final nit, I think this can be reduced back to it's original form with the new env var: `opts.release = option_env!("HEROKU_SLUG_COMMIT").map(Into::into);`

jtgeibel:
I think I recall reading about an option where heroku would make the repo available during the build, but a quick search doesn't find anything on that. And we could probably reproduce the `git describe --always` behavior with `git2`. Long term, if we can make those work, I'm all for switching to a single solution that works everywhere.

pietroalbini:
I think the best solution is to check whether `HEROKU_SLUG_COMMIT` is present in `build.rs`, and show the warning only if *both* the heroku var and the git repository are not present.

jtgeibel:
That's fine with me too. I just think the build script version makes the most sense if we also track dirty status during local development. Without that, I prefer a runtime version that defaults to None during local development.

src/bin/server.rs
Turbo87: yep, that seems to work. thanks :)

PR #2993, opened 2020-10-31
Use `clap` for CLI argument parsing

Turbo87:
This should make it easier to convert these binaries to clap subcommands in a follow-up PR. It also simplifies our argument parsing code and generates help messages and proper errors if the passed arguments don't match our expectations.

r? @jtgeibel

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ base64 = "0.13"
 cargo-registry-s3 = { path = "src/s3", version = "0.2.0" }
 chrono = { version = "0.4.0", features = ["serde"] }
 civet = "0.12.0-alpha.4"
+clap = "=3.0.0-beta.2"
 comrak = { version = "0.8", default-features = false }
 
 conduit = "0.9.0-alpha.3"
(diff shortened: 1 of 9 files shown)


src/bin/render-readmes.rs
```diff
@@ -18,42 +13,45 @@ use cargo_registry::{
...
 
-Options:
-    -h, --help         Show this message.
-    --page-size NUM    How many versions should be queried and processed at a time.
-    --older-than DATE  Only rerender readmes that are older than this date.
-    --crate NAME       Only rerender readmes for the specified crate.
-";
+#[derive(Clap, Debug)]
+#[clap(
+    name = "test-pagerduty",
```
jtgeibel, line 26: Should be "render-readmes".

jtgeibel (changes requested):
Awesome! Just 1 copy paste bug noted below. Feel free to r=me.

Cargo.toml
```diff
@@ -34,6 +34,7 @@ base64 = "0.13"
 cargo-registry-s3 = { path = "src/s3", version = "0.2.0" }
 chrono = { version = "0.4.0", features = ["serde"] }
 civet = "0.12.0-alpha.4"
+clap = "3.0.0-beta.2"
```
JohnTitor, line 37:
```suggestion
clap = "=3.0.0-beta.2"
```

I'd pin the version since `clap` often introduces breaking changes in beta releases (https://github.com/clap-rs/clap/blob/master/CHANGELOG.md) and we could avoid updating it by mistake.

Turbo87: Sounds reasonable. I've changed it to pinned. 👍

PR #2996, opened 2020-11-02
Use `dialoguer` crate for command line confirmations

Turbo87:
This simplifies our command line confirmation code in the admin tools by using https://docs.rs/dialoguer.

r? @jtgeibel

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,6 +286,7 @@ dependencies = [
  "cookie",
  "ctrlc",
  "derive_deref",
+ "dialoguer",
  "diesel",
  "diesel_full_text_search",
  "diesel_migrations",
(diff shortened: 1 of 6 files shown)


jtgeibel:
Looks good overall. One issue I see is that for some reason I don't get feedback when I first type `y` or `n`. It is only after I hit enter that the prompt is replaced with my answer. (That is, I type `n` and see no feedback, until I hit enter and then the auto-completed `no is shown.) Hopefully there is another option that can be configured to echo characters as they are typed.

So, this gets me thinking. A few weeks ago I tried out [`cursive`](https://crates.io/crates/cursive) for building a text user interface (TUI) for a side project. It worked really well. Once the admin binaries are consolidated, we could add a TUI providing an interactive way to run subcommands.

We've talked about having admin panels for various administrative tasks, and I think it would be way easier to build these in-process as a TUI than to add them to both the web server and frontend. There would be less boilerplate and integration testing to worry about.

Turbo87:
@jtgeibel I think the issue should be fixed now. It already showed `[N/y]` vs. `[n/Y]` before, but I've changed it to use a custom theme which makes the current selection even more visible.

jtgeibel:
Thanks, this makes the behavior of the prompt much clearer.

I'm fine with this behavior, but it is different than my initial expectations (and the previous behavior). Since @carols10cents and @pietroalbini  are most likely to use these commands, I'd like to make sure they are aware of the change as well. Anyone can r=me on this implementation if they agree on the prompt.

PR #2998, opened 2020-11-02
Added Accessibility category

simonsan:
I've added the `accessibility` category to cover crates that implement features for the convenience of disabled people.

I wanted to start a project that translates given colour codes to colours colorblind people can distinguish more easily, but I haven't found a category to fit it. What else belongs in this category could be crates that interoperate with peoples screenreaders, new screenreader implementations or other assistive technology.

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -26,6 +26,12 @@
 #   `@`, `:`, or `.`. They should be all lowercase.
 #
 
+[accessibility]
+name = "Accessibility"
+description = """
+Crates that implement features for the convenience of disabled people. \
+"""
+
 [algorithms]
 name = "Algorithms"
 description = """
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -29,7 +29,7 @@
 [accessibility]
 name = "Accessibility"
 description = """
-Crates that implement features for the convenience of disabled people. \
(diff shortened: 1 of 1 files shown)


src/boot/categories.toml
```diff
@@ -26,6 +26,12 @@
 #   `@`, `:`, or `.`. They should be all lowercase.
 #
 
+[accessibility]
+name = "Accessibility"
+description = """
+Crates that implement features for the convenience of disabled people. \
```
jtgeibel, line 32:
```suggestion
Tools for making your creations usable by as many people as possible. \
```

jtgeibel (changes requested):
Thanks for the PR @simonsan. I drew some inspiration from the [web accessibility docs](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/What_is_accessibility#So_what_is_accessibility) and propose an alternate description.

simonsan:
> 
> 
> Thanks for the PR @simonsan. I drew some inspiration from the [web accessibility docs](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/What_is_accessibility#So_what_is_accessibility) and propose an alternate description.

Thanks for the input. ;-) Changed it.

jtgeibel:
Thanks @simonsan!

@bors r+

PR #3011, opened 2020-11-09
Implement `crates-admin` binary

Turbo87:
This PR resolves https://github.com/rust-lang/crates.io/issues/1504 by adding a new `crates-admin` binary, that combines a lot of the previously separate binaries as subcommands.

I would recommend to review commit-by-commit :)

r? @jtgeibel


jtgeibel:
The `backfill-org-ids` bin was removed in #2810 and can be dropped here as well. Also we should be updating `bin_names` in `script/ci/prune-cache.sh` as we add/remove binaries. (This is mostly a hack to improve CI cache performance on Travis, and may be less necessary on GH Actions and with fewer binaries, but lets maintain it for now.)

For deployment purposes, `monitor` and `enqueue-job` are called from the scheduler on Heroku and it will be best to keep those around for a single deployment cycle. For the rest of the binaries included in this migration, I think it is fine to drop them in this PR.

Turbo87:
@jtgeibel thanks for the review. I've removed `monitor` from the `crates-admin` tool, added it to the `prune-cache.sh` and removed the now obsolete binaries. :)

jtgeibel:
I pushed a commit dropping the about attribute on `populate` (it was leftover from `monitor`).

I'm fine with migrating `enqueue-job` and `monitor` as well. We could even do so in a single PR. Just deploy first commits (up to where the bins are removed), update the scheduler, then deploy the merged commit.

Merging this PR and deploying to staging.

@bors r+

PR #3017, opened 2020-11-11
Remove TravisCI config

Turbo87:
We primarily use GitHub Actions these days and the TravisCI runs are just slowing us down unnecessarily now :)

r? @pietroalbini


.travis.yml
```diff
@@ -1,82 +0,0 @@
...
-  postgresql: "11"
-  apt:
-    packages:
-      - postgresql-11
-      - postgresql-client-11
-
-jobs:
-  fast_finish: true
-  allow_failures:
-    - rust: nightly
```
jtgeibel, line 49: As far as I'm aware, this is the only gap between our Travis and GH Actions configurations. On Actions we don't currently test on nightly because there wasn't an easy way to replicate the `allow_failures` functionality. I don't think this is a big deal, but if there's an easy way to do this in actions now that would be nice.

pietroalbini: There is no direct equivalent, but if we don't mind the red "x" we can just put the nighly build in a job not needed by `bors build finished`.

Turbo87: > if we don't mind the red "x"

That would look quite badly all over the GitHub UI. We could use `cargo test || true`, but then we would never see it if nightly actually starts failing.

I'm wondering if `cargo test || echo "::warning ::nightly build failed"` would work and show up in the UI 🤔

PR #3020, opened 2020-11-11
Parse query params with `serde` and `serde_urlencoded`

Turbo87:
see https://github.com/nox/serde_urlencoded

This slightly simplifies the query param parsing code, especially when conversions to numbers are involved. The pagination code had to be refactored a little bit for this to work, so I would suggest reviewing commit-by-commit to make the changes easier to understand. :)

r? @JohnTitor

### DIFF
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -109,16 +109,14 @@ impl<T> Paginated<T> {
     }
 
     pub(crate) fn prev_page_params(&self) -> Option<IndexMap<String, String>> {
-        if let Page::Numeric(1) | Page::Unspecified = self.options.page {
-            return None;
-        }
-
-        let mut opts = IndexMap::new();
         match self.options.page {
-            Page::Numeric(n) => opts.insert("page".into(), (n - 1).to_string()),
-            Page::Unspecified => unreachable!(),
-        };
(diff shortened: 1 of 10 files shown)


src/controllers/helpers/pagination.rs
```diff
@@ -14,19 +14,13 @@ pub(crate) enum Page {
...
-            let numeric_page = s.parse().map_err(|e| bad_request(&e))?;
-            if numeric_page < 1 {
-                return Err(bad_request(&format_args!(
-                    "page indexing starts from 1, page {} is invalid",
-                    numeric_page,
-                )));
+    fn new(page: Option<u32>) -> Page {
+        match page {
+            Some(page) => {
+                assert_ne!(page, 0);
```
jtgeibel, line 20: I haven't reviewed closely, but this caught my eye. We try to avoid potential panics in the request/response cycle. This refactor looks fine, but I suggest you add a doc comment with a `# Panics` section stating that the caller is responsible for ensuring this.

JohnTitor:
Overall looks fine to me and agree with Justin's comment as well. Sorry but I don't think I could do further reviews in the near future so I'd r? @jtgeibel

Turbo87:
closing this since it would conflict with the plan to eventually use cursor-based pagination

PR #3021, opened 2020-11-12
Apply clippy suggestions for comparison_to_empty

jtgeibel:
r? @Turbo87


Turbo87:
is there a way to turn that clippy warning into a hard check? I'm wondering why it did not fail CI for this before.

approving in the meantime since it's certainly an improvement 👍 

@bors r+

Issue #1451, opened 2018-07-03
Allow changing the case of a crate's name with a republish


carols10cents:
For the purposes of deciding whether a name is the same crate or a different crate, we consider names to be for the same crate no matter what case is used. Similarly, the URL to a crate is case-insensitive. [I've just filed a feature request with Cargo to make index lookups case-insensitive.](https://github.com/rust-lang/cargo/issues/5678)

If that Cargo feature is implemented, the only remaining reason to care about the crate casing is for aesthetics of display on crates.io. We should continue to display the case as published, but if someone changes their mind about how their crate name should be capitalized, we should allow them to change it by changing it in the crate's Cargo.toml and republishing (similarly to, say, the documentation URL currently-- that the last publish changes the crate's info in the database). 

Currently, crates.io thinks you're trying to publish a new crate and tells you the name has already been taken if you change the case, even if you're the owner of the crate.

jasonwilliams:
Is it possible for anyone to look into this?
Boa is stuck being Boa instead of boa

Rantanen:
I started implementing a solution to this, but encountered a complication with the download functionality. If this is still something crates.io would like to support, I'd love some comments on whether the solution over at #3034 is something that would be okay.

Personally I'd love this to be supported. I ended up making the same mistake last night when publishing [`Metamorfish`](https://crates.io/crates/Metamorfish) - which I could have (grudgingly) lived with, if not for [`protofish`](https://crates.io/crates/protofish) and [`automafish`](https://crates.io/crates/automafish) being published with lower case names already. Of course the irony here is that since `automafish` referred to `metamorfish` in the README.md, I wanted to squat that name before someone else got to it. ;_;

So instead of working on publishing a proper version of `metamorfish`, I ended up trying to solve this issue.. :)

jtgeibel:
I'm closing this based on https://github.com/rust-lang/crates.io/pull/3034#issuecomment-733311308. We can always reopen this issue if there is a crates.io specific approach with fewer sharp edges, but it looks like this probably requires an RFC across crates.io, cargo, and the language to resolve.

jasonwilliams:
@jtgeibel can we at least put some effort into warning users new crate casings are permanent and that cargo is case sensitive? I feel like its very easy to fall into this trap without any way back.

jtgeibel:
[...]

PR #3034, opened 2020-11-20
WIP: Support changing crate name case

Rantanen:
Is #1451 still on the table?

I made some implementation here, but the download functionality becomes somewhat problematic. Currently the crate download works by redirecting the client to AWS with path of:

>  `/crates/<crate_name>/<crate_name>-<version>.crate`.

This causes issues if the crate case is later renamed since the redirection will end up attempting to download the previously updated crate packages with the new name.

The solution here was to add `upload_name` column to the `versions` table that gets populated by the current crate name when migrating the database or publishing new versions.

When I started the implementation, I figured this would be just a simple matter of supporting the rename in the database in this specific case, however now that I encountered the issue with the file storage, I'd like to confirm whether this approach is suitable for crates.io at all.

The two concerns I have are:
- Each version needs to store the crate name separately.
- The database migration needs to update _all_ versions.

With ~50k crates according to crates.io front page and assuming around 10 versions per crate on average, that would yield an estimate of 500k versions. I'd guess that updating 500k versions during migration isn't _too_ big of an issue. If the average crate name is ~20 characters, the increase in database size would be around 10 MB (plus some database overhead), which doesn't sound too bad either.

_The current implementation attached to this PR is work in progress - I just wanted to get a PR up for discussion. The unimplemented bits are setting the `upload_name` when a new version is published and fixing the test case for the redirect._


jtgeibel:
Thanks for your interest in this @Rantanen. This is an area where we need to be very careful about our integration with `cargo` and also consider downstream consumers (like docs.rs and lib.rs). There are several places I see where we need to be careful:

* The [index repository](https://github.com/rust-lang/crates.io-index)
  * The path to the file with the crate's information likely needs to remain fixed, with the original casing, for compatibility reasons. While it might be possible to rename the file to the preferred casing, I don't see any reason to at this point unless other design considerations make it worth considering.
  * Within the file, should the `name` property for a version reflect the canonical (original) spelling or the casing that version was published with?
* The published crate asset
  * What casing is used for the file name? I haven't looked to see if cargo uses the filename as served from S3 or constructs it's own when saving it in `.cargo/registry/cache/*/`. `cargo` and other tools that download crate assets could become confused if the expected casing changes between versions.
  * For security reasons, when a crate is published we [check](https://github.com/rust-lang/crates.io/blob/d2c58d56c383814317192ff5fd3d8a4ac1085458/src/uploaders.rs#L207-L214) that the folder name within the tarball is `{crate_name}-{crate_version}`. This folder name determines where the files will be extracted under `.cargo/registry/src/*/`.
  * The name within the embedded `Cargo.toml` and `Cargo.toml.orig` files.
  * In general, I think it would be confusing if any of these strings differ in casing between these locations for a individual release. I think it would also cause issues (either with `cargo` or other tools) to have the casing change from one release to the next. Every tool working with crate assets would need to handle this correctly, when for the past 5+ years code assuming that these values would not change has been working correctly. 

Finally, if we had solid answers to these questions, we would want the implementation to have solid integration tests between cargo and the crates.io service. Unfortunately, we don't yet have the testing infrastructure in place to do the sort of end-to-end testing (i.e. invoking `cargo` from our tests) we would want for this sort of change.

I've opened rust-lang/cargo#8883 as an approach to help users avoid mistakes here.

Rantanen:
>   * The path to the file with the crate's information likely needs to remain fixed, with the original casing, for compatibility reasons. While it might be possible to rename the file to the preferred casing, I don't see any reason to at this point unless other design considerations make it worth considering.

The path is already lower cased: [`Metamorfish`](https://crates.io/crates/Metamorfish) crate is found in the index under [`me/ta/metamorfish`](https://github.com/rust-lang/crates.io-index/blob/9c605604d8f9cd1bc3f9a82ee74fc020b56946e5/me/ta/metamorfish)

I feel at least some of the concerns would be alleviated by requiring a major version bump when changing the casing. Then excluding `version = "*"` dependencies, at least Cargo wouldn't be stuck resolving a crate "Foo" and ending up with a version for one called "foo". Can't speak for lib.rs/etc.

Also just to be clear, I'm okay with closing this PR as "won't fix" as long as the original issues gets the same conclusion. Having a clear answer of "this won't happen" would be nicer than having a "this would be nice (but in reality we won't do anything to it to preserve backwards compatibility)". :)

(I'm not suggesting this should be closed, but the amount of concerns seems quite large and I'm not sure if having this possibility outweighs the risks. While I already got quite attached to my `Metamorfish`, I'm currently thinking of a new name for that crate to workaround this issue myself.)

jtgeibel:
Thanks for the correction on the file path. I should have double checked that.

The main alternative I see is that we could store a per crate display name, with a constraint that the crate and display names normalize to the same value. Then the exact crate name could be an implementation detail that is less prominent.

Right now cargo will suggest the correct crate name if the wrong casing is given. If it silently fixed the name, then we could always display the preferred version on the site, however rustc would have to handle this too. As you point out, right now if we allowed a rename, crates could only change casing on a major version bump because every usage of `Metamorfish` in a path or `use` statement would have to be changed to `metamorfish`. (Note that the same is not true for `-` vs `_` because the language only allows the latter.)

So it looks like this would probably require an RFC covering crates.io, cargo, and the language to really get it right. For that reason, I agree it is best to close #1451 as well. We can always reopen it if someone finds an approach with fewer sharp edges.

Razican:
> So it looks like this would probably require an RFC covering crates.io, cargo, and the language to really get it right. For that reason, I agree it is best to close #1451 as well. We can always reopen it if someone finds an approach with fewer sharp edges.

Hi @jtgeibel, is there a tracking issue / RFC in order to implement this plan?

Issue #3042, opened 2020-11-24
Pluralization of crate counts in categories and keywords is broken


balsoft:
**Describe the bug**

Visiting https://crates.io/categories, I see

Algorithms 927 crates

API bindings 1,382 crate

**To Reproduce**
1. Go to https://crates.io/categories or https://crates.io/keywords/
2. Examine the pluralization of the word "crate" in crate counts of categories or keywords; all numbers which are between 1000 and 1999 inclusive come with a singular "crate"

**Expected behavior**

I'm not a native English speaker, but I do believe that 1,382 should come with a plural "crates" and not singular "crate".

**Screenshots**

![A screenshot demonstrating the different pluralization](https://user-images.githubusercontent.com/18467667/100156757-85e36880-2eba-11eb-871f-56bc9b422da9.png)

**Desktop (please complete the following information):**

 - OS: NixOS GNU/Linux, nixpkgs-unstable
 - Browser: Firefox 82.0.2 (64-bit)

**Additional context**

I believe the problem comes from the fact that crates.io [passes](https://github.com/rust-lang/crates.io/blob/1839720df35c0aae2cbe0a2d0b07e5e786b499ec/app/templates/categories.hbs#L30) `"1,382", "crate"` to https://github.com/emberjs/ember-inflector 's `pluralize(count, word)` function, which [internally](https://github.com/emberjs/ember-inflector/blob/785c69e3e527bcef5e4b50710e77df7aff11b974/addon/lib/system/inflector.js#L227) uses `parseFloat(count) !== 1` to determine cases where plural should be used, and `parseFloat("1,382")` happens to be precisely `1`, thus leading to `pluralize` returning singular.

The "act 2" ~~is~~ was in the issue name because of #513, which is a very similar issue, and I believe the fix for it introduced _this_ issue.

kirelagin:
For context, this was introduced in c5a39cdd29c79a5d5e969c3d7d1ab9c748bd8655.

The problem is that `format-num` turns the number into a string (with a comma for numbers > 999). On the receiving end of the `pluralize` function, this string is parsed back into a number using `parseFloat`, which trips on the comma:

```
> parseFloat("1,234")
1
```

cc @carols10cents

balsoft:
The easiest fix should be along the lines of

```
          {{ format-num category.crates_cnt }} {{ pluralize category.crates_cnt "crate" { withoutCount: true } }}
```

balsoft:
[...]

PR #3043, opened 2020-11-24
Fix broken pluralizations

Turbo87:
Passing the formatted number into the `pluralize` helper doesn't work quite as expected, and in these cases we already know the pluralization, so using the `pluralize` helper is somewhat pointless anyway.

Resolves https://github.com/rust-lang/crates.io/issues/3042

r? @jtgeibel

### DIFF
--- a/app/templates/categories.hbs
+++ b/app/templates/categories.hbs
@@ -27,7 +27,7 @@
           {{category.category}}
         </LinkTo>
         <span local-class="crate-count" data-test-crate-count>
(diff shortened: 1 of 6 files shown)


balsoft:
Thank you for quick response!

I think it would also be nice to add a test or two [here](https://github.com/rust-lang/crates.io/commit/ab508f2ae96e4b2510e11fc3ab4a091874150f50#diff-c0d27e93bcd684be3d080615a454f227f29f46f922bc0cc5f02532affc121010R14) so that the issue won't be reintroduced.

Issue #2984, opened 2020-10-30
SyntaxError: expected expression, got '?'


89z:
If I visit this page:

https://crates.io

I get this:

~~~
SyntaxError: expected expression, got '?'
~~~

and the page completely fails to load. Looks like the error comes from here:

https://crates.io/assets/vendor-88810d9ca9f516e55d105cd64a11cacc.js

and the prettified code:

~~~js
const t = Ember.Object.create({
   message: e.message,
   type: e.type || "info",
   autoClear: e.autoClear ?? n.autoClear ?? !1,
   clearDuration: e.clearDuration ?? n.clearDuration ?? 3200,
   onClick: e.onClick,
   htmlContent: e.htmlContent || !1,
   cssClasses: e.cssClasses
});
~~~

The issue is `??`:

https://developer.mozilla.org/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator#Implementation_Progress

That syntax has not been standardized, and does not work with my browser
(Waterfox 2020.10).

a8nJ:
Here is another approach, only an extra 30 characters:

~~~
const r = 'autoClear', s = 'clearDuration', t = Ember.Object.create({
   message: e.message,
   type: e.type || "info",
   [r]: !(r in e) ? !(r in n) ? !1 : n[r] : e[r],
   [s]: !(s in e) ? !(s in n) ? 3200 : n[s] : e[s],
   onClick: e.onClick,
   htmlContent: e.htmlContent || !1,
   cssClasses: e.cssClasses
});
~~~

a8nJ:
looks like this broke it https://github.com/rust-lang/crates.io/commit/f05fe126aa1ab61d0c1da5d8322d4092eb1a9fe5

a8nJ:
https://github.com/rust-lang/crates.io/pull/2926

kentfredric:
Just gotta say, its *really* terrible getting entirely blank pages due to bugs in poor backcompat, where you have to use a different browser to even see what the "contact support" options are.

Can yall do server-side rendering or something at some point? 

- Total unusability for a given is very bad ( this is where you are )
- Partial breakage would be less bad, but still undesirable ( if this can be achieved, please, ... )
- Lack of functional breakage but only cosmetic damage is tolerable and a laudable goal.

I understand doing compat is hard, but christ, "totally blank pages" is the worst failure mode I can imagine. Especially for a site that is primarily a form of documentation.

kentfredric:
[...]

PR #3044, opened 2020-11-24
Show error message when JS code fails to load

Turbo87:
This should resolve https://github.com/rust-lang/crates.io/issues/2984, by adding a short error message when something fails to load before the Ember app has booted up:

Bildschirmfoto 2020-11-25 um 00 31 31

This does use an inline script though, so we might have to adjust the CSP for this 😞 

r? @jtgeibel

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -28,6 +28,7 @@
 
     {{content-for 'body'}}
 
+    <script>window.onerror=function(){document.body.innerHTML='<p style="width: 70%;background: var(--main-bg);padding: 10px;">Sorry, it looks like we were not able to load the page. Please make sure your network connection works and you are using an up-to-date browser. If the issue persists, please visit our <a href="https://github.com/rust-lang/crates.io/issues/new/choose">issue tracker</a> to report the problem.</p>'}</script>
     <script src="{{rootURL}}assets/vendor.js"></script>
     <script src="{{rootURL}}assets/cargo.js"></script>
 
--- a/app/index.html
+++ b/app/index.html
@@ -28,6 +28,7 @@
 
     {{content-for 'body'}}
 
+    <!-- if you change the following inline script make sure to change the CSP settings of nginx! -->
     <script>window.onerror=function(){document.body.innerHTML='<p style="width: 70%;background: var(--main-bg);padding: 10px;">Sorry, it looks like we were not able to load the page. Please make sure your network connection works and you are using an up-to-date browser. If the issue persists, please visit our <a href="https://github.com/rust-lang/crates.io/issues/new/choose">issue tracker</a> to report the problem.</p>'}</script>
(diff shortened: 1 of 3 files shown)


Turbo87:
I've added a hash of the inline script to the CSP config for nginx, deployed it to staging, and it appears to work :)

app/index.html
```diff
@@ -28,6 +28,7 @@
 
     {{content-for 'body'}}
 
+    <script>window.onerror=function(){document.body.innerHTML='<p style="width: 70%;background: var(--main-bg);padding: 10px;">Sorry, it looks like we were not able to load the page. Please make sure your network connection works and you are using an up-to-date browser. If the issue persists, please visit our <a href="https://github.com/rust-lang/crates.io/issues/new/choose">issue tracker</a> to report the problem.</p>'}</script>
```
jtgeibel, line 31: I think we should add a comment above this line, stating that any modifications to the inline script require an update to the CSP hash.

jtgeibel (changes requested):
This works for me locally in Firefox and Chrome when I block the `vendor-*.js` resource, but this does not work (i.e. blank page only) if I block `cargo-*.js`.

Additionally, in Firefox only, I see the error `None of the “sha512” hashes in the integrity attribute match the content of the subresource.` in my console. This is confusing because I don't have CSP enabled in my local environment, and the error seems to be referring to [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity). Otherwise the behavior between Chrome and Firefox appears the same, so this may be unrelated.

I don't understand why this appears to rely on cargo.js loading.

My test setup is to run `ember build --prod` followed by `cargo run --bin server` and then accessing via `http://localhost:8888`. This is similar to our production enviornment, sans nginx configuration (like CORS).

Turbo87:
I've managed to reproduce your issue and I think I understand what's going on:

At least in Chrome, it looks like blocking a URL through the devtools does not cause the `onerror` handler to be called, instead the request is just silently ignored. The reason why blocking the `vendor.js` file "worked" is that the `cargo.js` file depends on code in the `vendor.js` file, and since that code wasn't there it caused an "Uncaught ReferenceError: define is not defined" error visible on the devtools console.

In other words: blocking the URL through the devtools does not seem to create a realistic failure scenario. The thing we want to test here is having a syntax error in the JS file, so what I did to test this is prefix the content of one of the JS files with e.g. an `@` sign so that the browser can't parse it as valid JS anymore. This worked for both `cargo.js` and `vendor.js`, so I assume that this is still working as intended.

app/index.html
Turbo87: added :)

jtgeibel:
Well it sounds like my test steps were fatally flawed then. Thanks for investigating further and adding the comment.

@bors r+

PR #3048, opened 2020-11-27
Add an option for disabling the requirement to verify email before publishing

JWorthe:
Needing to verify your email before publishing can be a pain while running crates.io locally during development.

This PR adds a new environment variable to the config flag, DISABLE_EMAIL_VERIFICATION_REQUIREMENT. If this is set, then you can publish a crate even if your email has not been verified.

I've also updated the "contributing" docs around "Publishing a crate to your local crates.io", both to mention the email verification that you need to do, and how to disable it with the new flag.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -155,7 +155,7 @@ pub(crate) fn domain_name() -> String {
 }
 
 fn require_email_verification() -> bool {
-    !dotenv::var("DISABLE_EMAIL_VERIFICATION_REQUIREMENT").is_ok()
+    dotenv::var("DISABLE_EMAIL_VERIFICATION_REQUIREMENT").is_err()
 }
 
 fn blocked_traffic() -> Vec<(String, Vec<String>)> {
--- a/src/config.rs
+++ b/src/config.rs
@@ -18,6 +18,7 @@ pub struct Config {
     pub blocked_traffic: Vec<(String, Vec<String>)>,
     pub domain_name: String,
     pub allowed_origins: Vec<String>,
+    pub require_email_verification: bool,
 }
 
 impl Default for Config {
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -150,6 +154,10 @@ pub(crate) fn domain_name() -> String {
     dotenv::var("DOMAIN_NAME").unwrap_or_else(|_| "crates.io".into())
 }
 
+fn require_email_verification() -> bool {
+    dotenv::var("DISABLE_EMAIL_VERIFICATION_REQUIREMENT").is_err()
```
tshepang, line 158: if one had `DISABLE_EMAIL_VERIFICATION_REQUIREMENT=0`, won't this give a surprising result

JWorthe: Yes, it would be surprising, but I think consistency with the other fields in the config is more important. There are a few others that follow this same pattern in this config, where it's just checking set or unset. For example, MIRROR and HEROKU.

src/controllers/krate/publish.rs
```diff
@@ -49,14 +49,25 @@ pub fn publish(req: &mut dyn RequestExt) -> EndpointResult {
...
+    let email_address = if app.config.require_email_verification {
+        let verified_email_address = user.verified_email(&conn)?;
+        verified_email_address.ok_or_else(|| {
+            cargo_err(&format!(
+                "A verified email address is required to publish crates to crates.io. \
+                 Visit https://{}/me to set and verify your email address.",
+                app.config.domain_name,
+            ))
+        })?
+    } else {
```
carols10cents, line 61: Hm, I think what I would expect is that if `DISABLE_EMAIL_VERIFICATION_REQUIREMENT` is set, then we don't need to require the user to set *any* email.. that is, this `else` block shouldn't exist.

carols10cents (changes requested):
Could we not require *any* email instead?

Issue #2054, opened 2019-12-23
Wish for local development: run 1 command to start all services instead of 3 commands


carols10cents:
I've been thinking about this and trying some things for a while but I haven't had much luck, so I'm putting this out there in case someone else would like to try.

Here's my ideal workflow for local development:

- Check out whatever branch/PR I want to work on
- Run 1 command that starts up: 
  - `cargo run --bin server`
  - `cargo run --bin background-worker`
  - `npm install && npm run start:local` (I switch branches that change our JavaScript dependencies a lot)
- Notify me somehow that everything is ready, or that something failed
  - desktop notification? hard to do cross-platform, maybe there's a crate for that. 
  - opening a browser tab? might interrupt whatever I've switched to doing instead, would probably end up with me having a million tabs open. if failing just doesn't open a browser window, though, i probably would waste time before i go check.
  - something else??
- Have all the logs in that window, but identify which line comes from which process.
- When I CTRL-C that one command, all the running processes stop.

Bonus would be running `diesel migration run` before either of the `cargo run` commands start.

I think it would be really nice to have this optionally available to people working on the codebase; it could help us with [forgetting to document how to run the codebase locally](https://github.com/rust-lang/crates.io/issues/1998) if there's a file that we add the commands to that we use regularly, rather than the contributing documentation that we let go out of date.

Things I've considered:

- Shell script, but I don't feel like I know enough to do this properly, mostly the "stop all processes when the main one is stopped" part. And the "wait for all of the commands to be 'ready'" part. Actually all the parts.
- [foreman](https://github.com/ddollar/foreman), which I'm familiar with from my Ruby days, but I don't want to add installing Ruby to our development dependencies. There are some [ports of foreman](https://github.com/ddollar/foreman#ports), even [some Rust ones](https://crates.io/search?q=foreman), but I haven't evaluated any of those.
  - Even if we do use the Ruby foreman and make this setup optional, I've had some issues. [Here's a branch where I've tried using foreman](https://github.com/integer32llc/crates.io/tree/local-procfile), I've hit these problems:
    - Running `diesel migration run &&` before one of the two `cargo run` commands means the other one might start before the migration is done and that might cause the build to fail.
[...]

PR #3053, opened 2020-11-29
dev: update docker node version, server address, and docker-compose

nilslice:
Fixes dockerized environment to run crates.io locally.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,3 +1,6 @@
+# Toggle for server configuration in development when using Docker. 
+export DEV_DOCKER=1
+
 # Location of the *postgres* database. For example, if you have created a
 # blank database locally named `cargo_registry`, this would be
 # `postgres://postgres@localhost/cargo_registry`.
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,3 @@
-# Toggle for server configuration in development when using Docker. 
-export DEV_DOCKER=1
-
 # Location of the *postgres* database. For example, if you have created a
 # blank database locally named `cargo_registry`, this would be
 # `postgres://postgres@localhost/cargo_registry`.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
       context: .
       dockerfile: backend.Dockerfile
     environment:
+      DEV_DOCKER: "true"
       DATABASE_URL: postgres://postgres:password@postgres/cargo_registry
       SESSION_KEY: badkeyabcdefghijklmnopqrstuvwxyzabcdef
       GIT_REPO_URL: file://./tmp/index-bare
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,5 @@
 version: "3"
+
 services:
   postgres:
     image: postgres:9.6
@@ -10,6 +11,7 @@ services:
       - 5432:5432
     volumes:
       - postgres-data:/var/lib/postgresql/data
+
   backend:
     build:
       context: .
@@ -34,6 +36,7 @@ services:
       - index:/app/tmp
       - cargo-cache:/usr/local/cargo/registry
       - target-cache:/app/target
+
   frontend:
     build:
       context: .
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,6 +20,7 @@ services:
       GIT_REPO_URL: file://./tmp/index-bare
       GH_CLIENT_ID: ""
       GH_CLIENT_SECRET: ""
+      WEB_ALLOWED_ORIGINS: http://localhost:8888,http://localhost:4200
     links:
       - postgres
     ports:
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -513,6 +513,9 @@ one go:
 docker-compose up -d
 ```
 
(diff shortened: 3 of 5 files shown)


docker-compose.yml
```diff
@@ -20,6 +20,7 @@ services:
       GIT_REPO_URL: file://./tmp/index-bare
       GH_CLIENT_ID: ""
       GH_CLIENT_SECRET: ""
+      WEB_ALLOWED_ORIGINS: http://localhost:8888,http://localhost:4200
```
nilslice, line 23: Fix for error: 

```
backend_1   |      Running `target/debug/server`
backend_1   | Using local uploader, crate files will be in the local_uploads directory
backend_1   | thread 'main' panicked at 'must have `WEB_ALLOWED_ORIGINS` defined', src/lib.rs:105:39
backend_1   | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

frontend.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12.9-alpine
+FROM node:14.9-alpine
```
nilslice, line 1: Fix for error: 

```
Building frontend
Step 1/6 : FROM node:12.9-alpine
12.9-alpine: Pulling from library/node
e7c96db7181b: Pull complete
68c3b461dfcb: Pull complete
523dcfd12721: Pull complete
1e07af271ebb: Pull complete
Digest: sha256:3f9f6d4cceb65e9dbc8a8f9549c66b44e07d092353c07e64fef7da51c8ed06f7
Status: Downloaded newer image for node:12.9-alpine
 ---> 39fd6822c84f
Step 2/6 : WORKDIR /app
 ---> Running in 0674a75596c6
Removing intermediate container 0674a75596c6
 ---> 02a305b803a7
Step 3/6 : COPY package.json yarn.lock /app/
 ---> 6bcaa27e891f
Step 4/6 : RUN yarn install
 ---> Running in fad1b0f3b75a
yarn install v1.17.3
[1/5] Validating package.json...
error cargo@0.0.0: The engine "node" is incompatible with this module. Expected version "^14.9.0". Got "12.9.1"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

src/bin/server.rs
```diff
@@ -103,7 +103,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 async move { Service::from_blocking(handler, addr) }
             });
 
-        let addr = ([127, 0, 0, 1], port).into();
+        let addr = ([0, 0, 0, 0], port).into();
```
nilslice, line 106: Fix for error: 

```
ERROR: Service 'frontend' failed to build : The command '/bin/sh -c yarn install' returned a non-zero code: 1frontend_1  | Error proxying to http://backend:8888
frontend_1  | connect ECONNREFUSED 172.19.0.3:8888
```

jtgeibel: This should be put behind an environment variable that is set when run in docker. Maybe `DEV_DOCKER`. That way the server is not exposed on all interfaces when doing local development outside of docker.

The rest looks good to me. Thanks @nilslice!

nilslice: that's a good idea. I can make this update and push later this evening or tomorrow.

.env.sample
```diff
@@ -1,3 +1,6 @@
+# Toggle for server configuration in development when using Docker. 
+export DEV_DOCKER=1
```
jtgeibel, line 2: I don't think this needs to be added to the sample file, as this shouldn't be enabled for native local development. Shouldn't this be added directly to the `docker-compose.yml` file instead?

docs/CONTRIBUTING.md
```diff
@@ -512,6 +512,9 @@ one go:
 docker-compose up -d
 ```
 
+> **Note:** set `DEV_DOCKER=1` in a `.env` or `docker-compose.override.yml` 
+file when working with Docker in development.
+
```
jtgeibel, line 517: By updating `docker-compose.ym` directly, the 3 changes in this file should be dropped as well.

jtgeibel (changes requested):

.env.sample
nilslice: @jtgeibel - my thinking here was that since the `.env` file is read by the dockerized environment, it is a good place to add the toggle, instead of the `docker-compose.yml`, which _could_ be used for hosting. If it is used while hosted, then we've exposed the port too widely. 

I updated my branch using the flag directly within `docker-compose.yml` at first, and I don't mind reverting it. Just figured it would be more flexible this way.

jtgeibel: @nilslice I'm not sure I follow. *If* `docker-compose.yml` is used for deployment, then wouldn't this environment variable also need to be set for that to work?

At the least, we should drop this from `.env.sample` and mention it to `.env` only under the docker section of `CONTRIBUTING.md`. That is the only scenario where we want to recommend using this variable.

Turbo87: I've taken the liberty to rebase the branch and perform the change to the `docker-compose.yml` file, since I was considering working on https://github.com/rust-lang/crates.io/issues/2054, but then found this PR here 😅 

should be good to go from my side 👍

nilslice:
Great, thanks!!

PR #3059, opened 2020-12-01
tests/util: Remove crate name check from `enqueue_publish()`

Turbo87:
We have enough tests that explicitly assert the crate name already. Removing this code path allows us to slightly simplify the test code a little more.

r? @jtgeibel


jtgeibel:
LGTM!

@bors r+

PR #3061, opened 2020-12-02
tests/util: Replace `Response::assert_status()` with `Response::status()`

Turbo87:
`Response::status()` seems easier to use to me and doesn't hide the actual assertion. It also allows us to access things like `StatusCode::is_success()` directly.

r? @jtgeibel


pietroalbini:
Where is `assert_status` defined? Can we remove it?

PR #3068, opened 2020-12-03
Remove unused `parse_link_header` dependency

Turbo87:
https://github.com/est31/cargo-udeps indicated that this dependency might be unused, and indeed I couldn't find any references to it anywhere in our code.

r? @jtgeibel

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -302,7 +302,6 @@ dependencies = [
  "log",
  "oauth2",
  "parking_lot",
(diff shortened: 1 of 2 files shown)


JohnTitor (approved):
https://github.com/rust-lang/crates.io/pull/2796 introduced this with a new binary source but #2810 only removes the source so this should be fine.
Feel free to r=me as Justin seems busy.

PR #3072, opened 2020-12-04
tests/util: Remove unused `BackgroundJob` struct

Turbo87:
This struct does not appear to be used anywhere (anymore).

r? @JohnTitor


JohnTitor (approved):
This should be reasonable and all the tests actually pass. And if we need this, then we should export it from swirl anyway.

cc @jtgeibel just in case.

PR #3077, opened 2020-12-06
Header: Use custom focus styling

Turbo87:
The current focus outline is using a rectangle, while the input is rounded, which looks very ugly. This commit uses a `box-shadow` instead of the regular `outline` to solve this issue.

Before:
Bildschirmfoto 2020-12-06 um 15 45 59

After:
Bildschirmfoto 2020-12-06 um 15 44 37
Bildschirmfoto 2020-12-06 um 15 44 55

r? @locks

### DIFF
--- a/app/components/header.module.css
+++ b/app/components/header.module.css
@@ -58,6 +58,12 @@ input.search {
     background-size: 14px 15px;
     border-radius: 15px;
     box-shadow: none;
+    transition: box-shadow 150ms;
+
+    &:focus {
+        outline: none;
+        box-shadow: 0 0 0 4px var(--yellow);
+    }
 }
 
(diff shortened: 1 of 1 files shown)


app/components/header.module.css
```diff
@@ -58,6 +58,12 @@ input.search {
     background-size: 14px 15px;
     border-radius: 15px;
     box-shadow: none;
+    transition: box-shadow 150ms;
+
+    &:focus {
+        outline: none;
```
locks, line 64: does this not mess with accessibility?

Turbo87: only if there is no visual alternative, which the shadow provides in this case

PR #3079, opened 2020-12-07
make pg info more clear

tshepang:


Turbo87:
sounds reasonable. thanks @tshepang!

@bors r+

PR #3083, opened 2020-12-10
separate doc heading from body

tshepang:
If you look at target/doc/cargo_registry/util/index.html, you will see this heading after *errors*:
> This module implements several error types and traits. The suggested usage in returned results is as follows: 

The first line is taken as a heading, so should remain separate, so that it's readable as a summary.

### DIFF
--- a/src/util/errors.rs
+++ b/src/util/errors.rs
@@ -1,5 +1,6 @@
-//! This module implements several error types and traits.  The suggested usage in returned results
-//! is as follows:
+//! This module implements several error types and traits.
+//!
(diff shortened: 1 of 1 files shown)


Turbo87:
@tshepang an you add a pull request description explaining what you're changing here and why you're changing it? 🙏

tshepang:
just did @Turbo87

Turbo87:
I see what you mean:

Bildschirmfoto 2020-12-10 um 11 12 41


makes sense, thanks! 

@bors r+

PR #3084, opened 2020-12-10
owner test: make use of helper

tshepang:


src/tests/owners.rs
```diff
@@ -93,12 +93,8 @@ fn new_crate_owner() {
     token.enqueue_publish(crate_to_publish).good();
 
     // Add the second user as an owner (with a different case to make sure that works)
-    let user2 = app.db_new_user("Bar");
-    token.add_user_owner("foo_owner", "BAR");
```
Turbo87, line 97: this appears to be testing the case where the invitation is using uppercase while the username is using a different style. if we went with this refactoring then I think we would lose that check, no?

tshepang: agreed... is it fine if I uppercase the user on the helper (and add the case comment)

Turbo87:
@tshepang I still have a hard time figuring out what the advantage of this PR is. is it fixing any real issues? the tests appear to work fine the way they are, so I'm wondering where the need to change the code came from.

tshepang:
I was reading tests and noticed that the test fn doesn't use the helper fn, so am here reducing code duplication.

Note that I am so attached, so if this doesn't add enough value, I would be ok closing it.

Turbo87:
I think I'd prefer to keep the test as is, and only use the helper for the other tests. always having the helper use uppercase feels a bit suboptimal to me and since the helper ultimately only saves us 3 lines of code I'm not sure it's worth it to deduplicate here. 

thanks for the PR though! :)

PR #3090, opened 2020-12-13
Upgrade dependency to reduce slug size

jtgeibel:
The latest fastboot release removes some extraneous files that were
inflating our slug size. This is the minimal command I found that
bumped the version:

`yarn upgrade fastboot fastboot-app-server`

r? @Turbo87


Turbo87:
sounds good 👍 

@bors r+

PR #3100, opened 2020-12-17
Update `oauth2` to 4.0.0-alpha.2

JohnTitor:
Changes between 3.0.0 and 4.0.0-alpha.2: https://github.com/ramosbugs/oauth2-rs/compare/3.0.0...4.0.0-alpha.2
Changelog is here: https://github.com/ramosbugs/oauth2-rs/releases/tag/4.0.0-alpha.1

The main purpose is to drop `failure` dependency. Also, there're some changes on our deps.

### DIFF
--- a/src/admin/test_pagerduty.rs
+++ b/src/admin/test_pagerduty.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use clap::Clap;
-use failure::_core::str::FromStr;
+use std::str::FromStr;
 
 use crate::admin::on_call;
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,7 +273,6 @@ dependencies = [
  "diesel_migrations",
  "dotenv",
  "env_logger",
- "failure",
  "flate2",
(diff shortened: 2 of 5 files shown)


Turbo87:
see https://github.com/rust-lang/crates.io/pull/2953 🙊

src/admin/test_pagerduty.rs
```diff
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use clap::Clap;
-use failure::_core::str::FromStr;
+use std::str::FromStr;
```
Turbo87, line 3: huh, looks like IntelliJ autocompleted some interesting imports when I did the refactoring 🙈 

guess these two changes would make sense independent from the `oauth2` upgrade :)

JohnTitor:
> 
> 
> see #2953 🙊

Oh, didn't notice that! I think all the changes on that crate are reasonable and anyway we could pin the version here, but yeah let's stay then.

JohnTitor:
Superseded by #3117

PR #3101, opened 2020-12-17
Use `std`'s items instead of `failure`'s

JohnTitor:
r? @Turbo87


Turbo87:
awesome, thanks!

@bors r+

PR #3111, opened 2020-12-23
Log when a bot is suspected due to large page offset

jtgeibel:
For now just add log metadata so this is easily searchable. Eventually
we can suppress the metadata if the user is logged in, or even require
an authenticated user for very large offsets.

r? @Turbo87

### DIFF
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -1,7 +1,8 @@
+use crate::controllers::prelude::*;
 use crate::models::helpers::with_count::*;
 use crate::util::errors::{bad_request, AppResult};
+
 use diesel::pg::Pg;
-use diesel::prelude::*;
 use diesel::query_builder::*;
 use diesel::query_dsl::LoadQuery;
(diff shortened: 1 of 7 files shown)


src/controllers/helpers/pagination.rs
```diff
@@ -24,6 +28,10 @@ impl Page {
                 )));
             }
 
+            if numeric_page > MAX_PAGE_BEFORE_SUSPECTED_BOT {
+                req.log_metadata("bot", "suspected");
```
Turbo87, line 32: I have to say I'm not a big fan of having to pass in the `RequestExt` here just so that we can use the metadata logging functionality. We might want to have a look into alternatives at some point that allow for proper structured logging (e.g. the `tracing` crate).

Turbo87:
We don't seem to have any additional tests that check if the metadata is correctly logged, but that might also be hard to test properly. Since the remaining tests are working fine and I trust that you tested this locally I'm fine with not having that extra test :)

@bors r+

PR #3113, opened 2020-12-23
Clean up `TopVersions` code

Turbo87:
This PR cleans up the code that figures out the `highest` and `newest` versions of a crate in several ways. I hope that the commit list gives enough context on the individual changes.

Primarily, this decouples the code from the `Version` struct, and it moves the `0.0.0` defaulting to the `encodable()` method of the `Krate` model, instead of performing that in the low-level code directly.

Like https://github.com/rust-lang/crates.io/pull/3067, this PR is also in preparation for resolving https://github.com/rust-lang/crates.io/issues/2710.

r? @jtgeibel


jtgeibel:
Looks awesome!

@bors r+

PR #3115, opened 2020-12-23
Replace `log` and `env_logger` with `tracing` crates

Turbo87:
This brings us closer to being able to use structured logging and properly instrumenting some of functions. The change is mostly a drop-in replacement as described by https://docs.rs/tracing/0.1.22/tracing/#for-log-users.

The only difference is the slightly changed log output format:

```
# before
[2020-12-23T21:51:50Z INFO  cargo_registry::app] test23

# after
Dec 23 22:53:48.847  INFO cargo_registry::app: test23
```

This could be massaged by using a custom `FormatEvent` implementation though, if we really need this to be formatted in a specific way. Otherwise there is also a JSON-based output format which might be beneficial in production for real structured logging.

r? @pietroalbini

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -299,7 +299,6 @@ dependencies = [
  "lazy_static",
  "lettre",
  "license-exprs",
- "log",
  "oauth2",
  "parking_lot",
  "rand",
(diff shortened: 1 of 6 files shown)


pietroalbini:
I think you also need to update the test suite, after https://github.com/rust-lang/crates.io/pull/3130 landed.

Turbo87:
yep, you're totally right 👍 

should be fixed now :)

PR #3117, opened 2020-12-24
Bump to `tokio 1.0` and `hyper 0.14`

jtgeibel:
To avoid pulling in 2 versions of `tokio` at once:

* [x] Await `hyper-tls 0.5`
* [x] Await `reqwest 0.11`
* [x] Await an `oauth2` release depending on `reqwest 0.11` - ramosbugs/oauth2-rs#118
* [x] Await a `sentry` release depending on `reqwest 0.11` - getsentry/sentry-rust#304

r? @ghost


Turbo87:
@jtgeibel I've updated `sentry` to v0.22 which includes the necessary subdependency updates. I guess that means we could merge this PR now?

jtgeibel:
> I've updated `sentry` to v0.22

Awesome, that was the last thing I was waiting for!

r? @Turbo87

PR #3120, opened 2020-12-25
models: Add missing `pub` declarations

Turbo87:
Without these declarations it is quite hard to use these model structs outside of their files, and since the structs themselves are `pub` it makes sense for their fields to also be public.

r? @pietroalbini

### DIFF
--- a/src/models/dependency.rs
+++ b/src/models/dependency.rs
@@ -31,12 +31,12 @@ pub struct Dependency {
 #[derive(Debug, QueryableByName)]
 pub struct ReverseDependency {
     #[diesel(embed)]
-    dependency: Dependency,
+    pub dependency: Dependency,
     #[sql_type = "::diesel::sql_types::Integer"]
-    crate_downloads: i32,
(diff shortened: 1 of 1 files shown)


pietroalbini:
The intent of the fields not being `pub` is to force new initializations to go through the `new` method, which also does some validation. Why do you need the fields to be public?

Turbo87:
I primarily need it on the `ReverseDependency` to replace the `encodable()` method with a `From` implementation. Without the fields being `pub` it is not possible to access the fields from the `views` module.

I added the other structs because I thought it would make sense for all of the structs in the `models` module to have `pub` fields, but maybe you're right. Do you want me to remove the other structs and only keep the rev dep commit?

pietroalbini:
Ok, so, we definitely don't want fields to be `pub` for `NewVersion` (or any `New*`). For the others, I'd say let's add `pub` only when it's needed.

Turbo87:
okay, I've removed the other three commits for now :)

PR #3130, opened 2020-12-27
tests: Initialize env_logger in `TestApp::init()`

Turbo87:
This allows us to use e.g. `debug!()` in test utilities and run `cargo test` with `RUST_LOG=debug`.

This PR also adds `debug!()` calls to our HTTP proxy replay code, to make it easier to see what requests have been intercepted by the proxy.

r? @pietroalbini


pietroalbini:
@bors r+

Thanks!

PR #3144, opened 2020-12-30
notifications: Increase `clearDuration` to 10 sec

Turbo87:
Having the notifications on the screen for only 3 sec by default is far too fast for most cases. Having it on the screen for too long is far better than the opposite. :)

r? @pichfl


pichfl:
I think 10s is acceptable, but I do recommend looking at individual notifications. It could make sense to never auto clear errors for example.

@bors r+

PR #3146, opened 2020-12-30
Remove outdated information about GitHub permission error

JohnTitor:
https://crates.io/login is no longer valid (found in https://github.com/rust-lang/cargo/pull/9000#discussion_r547384911) and it's just enough to recommend re-authentication now, I think.


Turbo87:
good catch!

@bors r+

PR #3163, opened 2021-01-05
EncodableCrate: Add `max_stable_version` field

Turbo87:
This PR resolves the API part of https://github.com/rust-lang/crates.io/issues/654 by adding a `max_stable_version` field for the `krate` resource. Since naming is hard, it is called `highest_stable` internally on the `TopVersions` struct. Contrary to `newest_version` and `max_version` this field will be `null` (not `"0.0.0"`) if all available versions have been yanked.

r? @jtgeibel

### DIFF
--- a/src/models/version.rs
+++ b/src/models/version.rs
@@ -43,8 +43,7 @@ pub struct NewVersion {
 pub struct TopVersions {
     /// The "highest" version in terms of semver
     pub highest: Option<semver::Version>,
-    /// The "highest" non-prerelease version, or, if only
-    /// prereleases exist, the "highest" version
+    /// The "highest" non-prerelease version
     pub highest_stable: Option<semver::Version>,
     /// The "newest" version in terms of publishing date
     pub newest: Option<semver::Version>,
@@ -70,8 +69,7 @@ impl TopVersions {
             .into_iter()
             .map(|(_, v)| v)
             .filter(|v| !v.is_prerelease())
-            .max()
-            .or_else(|| highest.clone());
+            .max();
 
         Self {
             newest,
@@ -302,7 +300,7 @@ mod tests {
             TopVersions::from_date_version_pairs(versions),
             TopVersions {
                 highest: Some(version("1.0.0-beta.5")),
-                highest_stable: Some(version("1.0.0-beta.5")),
+                highest_stable: None,
                 newest: Some(version("1.0.0-beta.5")),
             }
         );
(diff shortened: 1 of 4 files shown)


src/models/version.rs
```diff
@@ -43,6 +43,9 @@ pub struct NewVersion {
 pub struct TopVersions {
     /// The "highest" version in terms of semver
     pub highest: Option<semver::Version>,
+    /// The "highest" non-prerelease version, or, if only
+    /// prereleases exist, the "highest" version
```
jtgeibel, line 47: I was assuming the semantics would be that if there are only prerelease versions (or all stable versions are yanked), then this would be `None`.

Turbo87: as I said, naming is hard :D

this matches the business logic that we are currently doing on the frontend when visiting a crate and figuring out which version to display. it is essentially the "most relevant version".

does that make sense?

jtgeibel: Should we keep that bit of business logic in the frontend? Using `None`/`null` here to mean "no stable releases" (either never, or all since yanked) seems a bit more useful than providing a fallback value in the backend.

We don't need that distinction in the frontend currently, but it seems nice to provide it by the backend anyway.

Turbo87: the problem is that we can't perform that logic everywhere on the frontend because e.g. for search results it would cause N+1 queries since the results only include the crate, but not the versions. for the crate details page this is acceptable but not for crate lists. I'd prefer to use the same logic everywhere to stay consistent.

we could also rename it to default_version or something like that if it helps?

jtgeibel: I don't follow why this would be N+1 on the frontend. The search API returns values for `max_version`, `newest_version`, and now `max_stable_version`. Can't the frontend fallback to `max_version` if there is no `max_stable_version`? I don't think search results would require additional queries.

Turbo87: yeah, I guess with using both fields this could work. my sidegoal was to eventually get rid of the other two fields, but I don't have strong feelings on that.

I'll change it to remove the fallback.

jtgeibel: > my sidegoal was to eventually get rid of the other two fields, but I don't have strong feelings on that.

We should keep them in case there are third party consumers. (Long term, we should at least start planning a v2 API and tracking our known v1 warts.)

> I'll change it to remove the fallback.

Thanks! Would you update the doc comment as well?

Turbo87: > Would you update the doc comment as well?

whoops, sorry about that. done :)

jtgeibel:
Thanks!

@bors r+

PR #3178, opened 2021-01-13
Version|Crate: Replace `encodable()` methods

Turbo87:
This PR is similar to #3168 and brings us one step closer to removing the `models -> views` dependency by inverting the relationship for the `Version` model and `EncodableVersion` view.

r? @pietroalbini


Turbo87:
I've added a corresponding change for `Crate` and `EncodableCrate` too now

PR #3193, opened 2021-01-19
db: Simplify `connect_now()` implementation

Turbo87:
`connect_now()` appears to share a lot of code with the `connection_url()` function, so it should be safe to use one inside of the other

r? @pietroalbini

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -56,11 +56,8 @@ impl Deref for DieselPooledConn<'_> {
 }
 
 pub fn connect_now() -> ConnectionResult<PgConnection> {
-    let mut url = Url::parse(&crate::env("DATABASE_URL")).expect("Invalid database URL");
-    if dotenv::var("HEROKU").is_ok() && !url.query_pairs().any(|(k, _)| k == "sslmode") {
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -56,11 +56,9 @@ impl Deref for DieselPooledConn<'_> {
 }
 
 pub fn connect_now() -> ConnectionResult<PgConnection> {
-    let mut url = Url::parse(&crate::env("DATABASE_URL")).expect("Invalid database URL");
-    if dotenv::var("HEROKU").is_ok() && !url.query_pairs().any(|(k, _)| k == "sslmode") {
-        url.query_pairs_mut().append_pair("sslmode", "require");
-    }
-    PgConnection::establish(&url.to_string())
+    let url = crate::env("DATABASE_URL");
+    let url = connection_url(&url);
```
pietroalbini, line 60:
```suggestion
    let url = connection_url(&crate::env("DATABASE_URL"));
```

There is no need to have separate variables for this.

Turbo87: good point, should be fixed now :)

PR #3211, opened 2021-01-28
Show "Try on Rust Playground" button for crates that are available there

Turbo87:
Bildschirmfoto 2021-01-28 um 18 06 47

r? @pichfl

### DIFF
--- a/app/components/crate-sidebar.hbs
+++ b/app/components/crate-sidebar.hbs
@@ -119,4 +119,22 @@
       {{/if}}
     {{/unless}}
   </div>
+
+  {{#if this.playgroundLink}}
+    <div>
+      <a
+        href={{this.playgroundLink}}
+        target="_blank"
+        rel="noopener noreferrer"
+        local-class="playground-button"
+        data-test-playground-button
+      >
+        Try on Rust Playground
+      </a>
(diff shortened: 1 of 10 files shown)


carols10cents:
@shepmaster We'd love your feedback on this feature and the best way to implement it to make it easy for people to try available crates out on the playground but not add playground support burden (likely in the form of "can I have x crate on the playground too?")

We're currently thinking it would make more sense to show this button only for those crates that are available on the playground by virtue of being in the top 100 crates, not because they're a dependency of the top 100 (I will open an issue on the playground repo to discuss the possibility of having that available from the playground as an API).

We'd show the button that says "Try this crate on the playground" along with some text that says "The top 100 crates are available in the online Rust playground" or similar. What do you think?

carols10cents:
Playground issue opened: https://github.com/integer32llc/rust-playground/issues/676

Turbo87:
> We'd show the button that says "Try this crate on the playground" along with some text that says "The top 100 crates are available in the online Rust playground" or similar.

I've added the help text and a bunch of tests for the functionality. This should be ready to go now :)

Turbo87:
fixed the broken test that was caused by 303e515 😅 

@bors r=pichfl

PR #3216, opened 2021-01-30
Move docs.rs link code from controller to `version` model

Turbo87:
If we want to reuse this in other places then this code should live on the model, not a specific controller.

r? @pichfl


locks:
Looks good!

PR #3217, opened 2021-01-30
tests: Move `req()` function to `util` module

Turbo87:
The `req()` function does not seem to be used anywhere else, so it's probably easiest to keep it in the same module as where it is used.

r? @pietroalbini


locks:
I think this makes sense!

@bors r+

PR #3218, opened 2021-01-30
Extract `GitHubClient` struct

Turbo87:
This PR aims to simplify our GitHub API request code a little bit by extracting a `GitHubClient` struct, and adding several request methods on it. It also allows to set a custom `base_url`, which will be useful in the future when used in combination with https://docs.rs/mockito.

r? @pietroalbini


locks:
Reviewed commit by commit and everything checks out!

@bors r+

PR #3220, opened 2021-01-31
Extract `RenderedHtml` component

Turbo87:
This PR extracts a new `RenderedHtml` component, which is used to render the README content on the crate details pages. To make the component easier to handle and align on the page, the styling was slightly adjusted so that the component does not have any outer margin by default.

r? @locks


locks:
Removing the outer margin seems like an improvement.

@bors r+

PR #3221, opened 2021-01-31
Docker: Improve `postgres` setup

Turbo87:
This PR allows us to use the docker-compose setup to run `cargo test` with the Dockerized Postgres server. It also adjusts the docker-compose setup a little bit to not expose the `postgres` container on *all* host interfaces, but instead only makes it available on 127.0.0.1.

To run just the database container you can use:

```
docker-compose up -d postgres
```

(`-d` runs it in daemon mode)

/cc @plaets

r? @pietroalbini

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: password
     ports:
-      - 5432:5432
+      - 127.0.0.1:5432:5432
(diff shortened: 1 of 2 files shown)


plaets:
Why was I mentioned here? Not that I have a problem with that, I'm just not sure if I wasn't supposed to do something 😅

Turbo87:
@plaets only for awareness because you were working on the other Docker topics :)

PR #3222, opened 2021-01-31
Replace Moment.js with date-fns

Turbo87:
see https://momentjs.com/docs/#/-project-status/

### Before

- cargo.js: 337.66 KB (46.46 KB gzipped)
- vendor.js: 994.86 KB (267.9 KB gzipped)

### After

- cargo.js: 333.25 KB (46.21 KB gzipped)
- vendor.js: 928.56 KB (251.39 KB gzipped)

r? @locks


Turbo87:
whoops... I forgot that Chart.js is using Moment.js by default 😅

PR #3223, opened 2021-01-31
Start background-worker from docker-compose.yml

plaets:
Pasting the detailed list of my changes on top of @Turbo87's branch:
* change GIT_REPO_URL to an absolute path to the repo, otherwise the background worker fails with 
  `thread 'main' panicked at 'Failed to clone index: Error { code: -1, klass: 2, message: "failed to resolve path \'file://./tmp/index-bare\': No such file or directory" }', src/bin/background-worker.rs:38:46`
* add a new volume for the `local_uploads` directory, otherwise
  backend won't be able to read readmes rendered by the background worker.
  I'm not aware of any other directories that should be shared between these two services.
* remove the `depends_on` condition for the worker - this syntax is not
  supported anymore according to the official docs
  https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on
  If it's necessary that the background worker starts after the backend, a startup
  script for the worker that polls the backend if it's ready will probably be needed
  https://docs.docker.com/compose/startup-order/
* update `docker-compose.yml` version to `3.4`, `start_period` is
  not supported in earlier versions
  https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,19 +35,11 @@ services:
       - 5432:5432
     volumes:
       - postgres-data:/var/lib/postgresql/data
-    healthcheck:
-      test: pg_isready -U postgres
-      interval: 5s
-      start_period: 30s
 
   backend:
     <<: *backend
     ports:
       - 8888:8888
-    healthcheck:
-      test: curl http://backend:8888/api/v1/site_metadata
-      interval: 5s
-      start_period: 300s
     depends_on:
       - postgres
 
@@ -69,10 +61,6 @@ services:
     volumes:
       # Mount the app/ directory so live reload works
       - ./app:/app/app:ro
-    healthcheck:
-      test: wget http://frontend:4200/ -O /dev/null
-      interval: 5s
-      start_period: 120s
 
 volumes:
   postgres-data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,10 @@ services:
       - 5432:5432
     volumes:
       - postgres-data:/var/lib/postgresql/data
+    healthcheck:
+      test: pg_isready -U postgres
+      interval: 5s
+      start_period: 30s
 
   backend:
     build:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -49,9 +69,14 @@ services:
     volumes:
       # Mount the app/ directory so live reload works
       - ./app:/app/app:ro
+    healthcheck:
+      test: wget http://frontend:4200/ -O /dev/null
+      interval: 5s
+      start_period: 120s
```
Turbo87, line 75: I'm now wondering if we actually need these healthchecks for the `frontend` and `backend` containers. I guess the `worker` container only really needs to the `postgres` container too. 🤔

Turbo87: the reason why I'm asking is that the `curl http://backend:8888/api/v1/site_metadata` above makes the console quite noisy due to the repeated HTTP requests that are coming in from the health check.

plaets: > makes the console quite noisy

Yeah, I did notice that. I'm actually not sure if any of the healthchecks are needed anymore, from what I understand depends-on does not wait for the service to be healthy, only for the container to be started. And also that docker uses healthcheks only to report the status of the service, but I believe it takes no action when a service stops being "healthy" (unless docker swarm is used, I think?)

https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on
> There are several things to be aware of when using depends_on:
> depends_on does not wait for db and redis to be “ready” before starting web - only until they have been started. If you need to wait for a service to be ready, see Controlling startup order for more on this problem and strategies for solving it.

https://docs.docker.com/engine/reference/builder/#healthcheck
> When the health status of a container changes, a health_status event is generated with the new status.

Not exactly an authority in this case, posting it for more context
https://stackoverflow.com/questions/47088261/restarting-an-unhealthy-docker-container-based-on-healthcheck

Turbo87: yeah, that makes sense. if it works fine  without the healthchecks then I'd say let's remove them. I initially added them because I clearly didn't know what I was doing :D

plaets: I get it, I also don't understand why did they (docker people) leave healthchecks in this half-useless state (or maybe I'm completely misunderstanding something), I remember having a similar issue in a different project I worked on.

It seems to work fine without healthchecks, I pushed the latest commit.

Turbo87 (approved):

Turbo87:
LGTM! nice work :)

@bors r+

PR #3224, opened 2021-01-31
Catch exceptions thrown when an anchor in a readme is invalid

plaets:
Implements a fix to #3108


Turbo87 (approved):
looks great, thanks! :)

PR #3234, opened 2021-02-02
Move download data processing code into `DownloadGraph` component

Turbo87:
This code is only needed inside the `DownloadGraph` component, so that's where the code should live. This also gives us the opportunity to simplify the code a little bit, since the Chart.js library needs to data in a different layout than the previous implementation that was based on Google Charts. Visually the results should still be identical to the previous state though :)

r? @locks


locks:
Thanks for the tidying up!

@bors r+

PR #3235, opened 2021-02-02
Fix crates-admin render-readmes for local installations

### LINKED ISSUE #3194: Crates-admin render-readmes fails with "relative URL without a base" on local uploader
**Describe the bug**
Crates-admin fails to execute `render-readmes` when crates.io is configured with a local uploader. 

**To Reproduce**
1. Start crates.io from default docker-compose (with github credentials configured)
2. Publish any crate with a README 
3. Run `cargo run --bin crates-admin render-readmes`, preferably from the backend container

**Expected behavior**
`crates-admin` renders all readmes, exits with a success.

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.10

**Additional context**
I believe this is caused by `crate_location` in `src/uploaders.rs`, depending on which type of uploader is configured, it either returns an absolute url (S3) or a relative url ("without a base", Local). `get_readme` from `src/admin/render_readmes.rs` passes whatever `crate_location` returns to an http client, which then fails with `relative URL without a base`. 

~~I made changes that seem to fix the problem, at least on my machine, [here](https://github.com/plaets/crates.io). I also had to set the user agent, otherwise, I get an error from the api.~~
Update: turns out my fix is actually useless, it causes problems when frontend tries to download readmes (cors because of a  redirect from :4200 to :8888).
[This](https://github.com/plaets/crates.io/tree/issue-3194) should be a little bit better.

https://github.com/rust-lang/crates.io/blob/8c68a622e8837473f1b10487be023c951a0fb4e4/src/uploaders.rs#L39-L55
https://github.com/rust-lang/crates.io/blob/8c68a622e8837473f1b10487be023c951a0fb4e4/src/admin/render_readmes.rs#L153-L172

plaets:
Fixes #3194

### DIFF
--- a/src/admin/render_readmes.rs
+++ b/src/admin/render_readmes.rs
@@ -3,6 +3,7 @@ use crate::{
     models::Version,
     render::readme_to_html,
     schema::{crates, readme_renderings, versions},
+    uploaders::Uploader,
     Config,
 };
 use std::{io::Read, path::Path, sync::Arc, thread};
@@ -15,6 +16,7 @@ use reqwest::{blocking::Client, header};
 use tar::{self, Archive};
 
 const CACHE_CONTROL_README: &str = "public,max-age=604800";
+const USER_AGENT: &str = "crates-admin";
 
 #[derive(Clap, Debug)]
 #[clap(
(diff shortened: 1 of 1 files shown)


src/admin/render_readmes.rs
```diff
@@ -160,7 +162,14 @@ fn get_readme(
         .uploader
         .crate_location(krate_name, &version.num.to_string());
 
-    let response = match client.get(&location).send() {
+    let location = match config.uploader {
+        Uploader::S3 { .. } => location,
+        Uploader::Local => format!("http://localhost:8888/{}", location),
+    };
```
Turbo87, line 168: shouldn't that be taken care of inside the `crate_location()` method?

plaets: I actually tried this approach at first, but it broke downloading readmes from the frontend. Frontend would get a redirect to the backend (:8888) and that caused a cors block, the branch with that version is [here](https://github.com/plaets/crates.io)

Turbo87:
I looked into it a little more this morning and while I'm not 100% happy with the solution I haven't been able to come up with anything better yet, so I guess this is good enough for now. 

Ideally we would just read the file directly for the local uploader case instead of requesting it via HTTP, but that appears to make the type system a little unhappy and I'm not sure how to get around that in a clean way.

@bors r+

Turbo87:
@bors r-

let me rebase this first :D

PR #3237, opened 2021-02-04
styles: Use numeric darkness naming and HSL for color values

Turbo87:
This PR refactors our CSS color variables to use numeric darkness values in their names (e.g. `grey900` and `yellow500`), instead of "dark, darker, darkest, ...". It also switches those color values over to use HSL instead of RGB to make it easier to find good light/dark variants of the colors in the future. The actual colors did not change though, they were only converted to the other representation :)

r? @locks

### DIFF
--- a/app/components/header.module.css
+++ b/app/components/header.module.css
@@ -62,7 +62,7 @@ input.search {
 
     &:focus {
         outline: none;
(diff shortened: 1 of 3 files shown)


locks:
Let's do this!

@bors r+

PR #3241, opened 2021-02-05
CI: Disable `yarn test` command for `embroider` job

Turbo87:
We are running into a similar issue to what is described at https://github.com/embroider-build/embroider/issues/677. We currently allow the CI step to fail, but since it waits until it times out it keeps the PRs in "running" state for an unnecessarily long time.

This PR is changing the CI to only build, but not run the tests, with embroider to avoid the above issue. Once https://github.com/embroider-build/embroider/issues/677 is addressed we can revert this change again.

r? @locks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,9 @@ jobs:
           node-version: '14.x'
 
       - run: yarn install
-      - run: yarn test
+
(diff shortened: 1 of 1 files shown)


Turbo87:
I'll take the liberty to merge this as is, since it's kinda blocking some of the other PRs and we're not actually using this CI job for anything right now anyway.

@bors r+

PR #3246, opened 2021-02-07
Sort versions in download graph using semverSort

### LINKED ISSUE #3245: Versions are sorted alphabetically instead of numerically
**Describe the bug**

1.10.0 is sorted after 1.9.0

**To Reproduce**
1. Go to https://crates.io/crates/iced-x86
2. Scroll down to the download stats
3. 1.10.0 is sorted after 1.9.0

**Expected behavior**

1.10.0 should be sorted before 1.9.0

plaets:
Fixes #3245

### DIFF
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -5,6 +5,7 @@ import Component from '@glimmer/component';
 
 import subDays from 'date-fns/subDays';
 import window from 'ember-window-mock';
+import semverSort from 'semver/functions/sort';
 
 // Colors by http://colorbrewer2.org/#type=diverging&scheme=RdBu&n=10
 const COLORS = ['#67001f', '#b2182b', '#d6604d', '#f4a582', '#92c5de', '#4393c3', '#2166ac', '#053061'];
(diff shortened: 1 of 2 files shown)


app/components/download-graph.js
```diff
@@ -5,6 +5,7 @@ import Component from '@glimmer/component';
 
 import subDays from 'date-fns/subDays';
 import window from 'ember-window-mock';
+import semverSort from 'semver/functions/sort';
```
Turbo87, line 8: if we import it here we should explicitly add it as a dependency to the `package.json` file :)

Turbo87 (changes requested):

app/components/download-graph.js
plaets: You are right, I didn't notice that this is included only in devDependencies. Can I just copy and paste it to dependencies?

Turbo87: yeah, just move it to the `dependencies` section

Turbo87 (approved):
LGTM! 👍 

@bors r+

Issue #2520, opened 2020-05-20
Add icons for documentation, repo, dependant crates link, etc..


codec-abc:
Hi,

Catching myself looking for the documentation and Github link for a crate, I realized I often struggle to find these links (usually searching on the right column before realizing they are not here). This is probably because I don’t do enough Rust bust still I think adding some icons to the links for a crate under its name would improve readability because icons are more easily catch by the eye in a glimpse than text:

So instead of something like:
![image](https://user-images.githubusercontent.com/1208646/82453848-58ec6e00-9ab1-11ea-834f-0ceb1b836134.png)

We would have something like:
![image](https://user-images.githubusercontent.com/1208646/82454699-5dfded00-9ab2-11ea-9558-7ff04ae0e2ef.png)

What do you think?

pichfl:
@Turbo87 summoned me here.

The problem here is not the missing icons, but the lack of visual and calculated contrast of these links. 

#00ac5b on #edebdd [does not provide any reasonable contrast](https://webaim.org/resources/contrastchecker/?fcolor=00AC5B&bcolor=EDEBDD), which makes it very hard to spot them. Reducing the font-size to 80% of the default doesn't help either.

I'm not against introducing icons, I just want to make sure they match the overall theme if we introduce new ones. 

In the mean time, removing the font-size: 80% and maybe even replacing the darker header background with white will improve contrast a lot and therefore also visibility. Moving the links down by 1rem (reducing the bottom spacing of the header by the same amount will also help to eliminate the problem without introducing new elements.

codec-abc:
I agree that contrast and font size is not optimal right now. And I also understand that you don't want to introduce icons. Yet, I have the feeling that icons would help visualy parse the page.

Maybe this can be done in 2 steps where the first one is dealing with the low hanging fruit (fixing font size, contrast) and once done we can see about icons.

PR #3250, opened 2021-02-07
CrateHeader: Move quick links into the sidebar

Turbo87:
I guess a picture is worth a thousand words:

localhost_4200_crates_rand

This PR moves the quick links from the crate details header to the sidebar.

It also changes their appears a little bit to make it easier to see what link you're clicking on. Instead of just showing e.g. the label "Homepage" we will now also show the actual URL to make it easier for our users to avoid malicious URLs.

To make the URLs easier to parse we also add icons for the two more common link types: docs.rs and github.com. We don't plan on adding icons for every single hosting provider available, but for the most mainstream ones it seems useful. In other words: PRs to add e.g. an icon for GitLab would probably be accepted, but a dedicated icon for your own domain that nobody else uses is very unlikely to be accepted.

This is opened in draft mode for now since @pietroalbini requested to have a look before merging it, but a review (and potentially approval) would still be good. :)

Resolves https://github.com/rust-lang/crates.io/issues/2520


public/assets/link.svg
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
pichfl, line 1: I'd recommend running `[svgo](https://github.com/svg/svgo) --disable=removeViewBox public/assets/link.svg` on these files to remove unnecessary cruft.

Turbo87: that is being done automatically by ember-svg-jar :)

PR #3260, opened 2021-02-09
models/version: Shorten `releaseTrack` for high major versions

Turbo87:
Example: https://crates.io/crates/rustc-ap-rustc_arena/versions

Having `706.x` as the release track barely fits in the release track indicator, so let's shorten this value for major versions above 100.

### DIFF
--- a/app/models/version.js
+++ b/app/models/version.js
@@ -66,7 +66,7 @@ export default class Version extends Model {
     }
 
     let { semver } = this;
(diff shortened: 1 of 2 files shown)


pichfl:
This is an acceptable stop-gap measure but not a holistic solution to the problem. We might have to revisit in the future but it should be fine for now.

@bors r+

PR #3266, opened 2021-02-09
Implement and use `/crates/:crate_id/range/:range` route

Turbo87:
... that redirects the user to the highest version that satisfies the specified semver range. I guess the added test cases should explain the purpose quite well :)

### DIFF
--- a/app/components/dependency-list/row.hbs
+++ b/app/components/dependency-list/row.hbs
@@ -12,8 +12,8 @@
 
   <span>
     <LinkTo
-      @route="crate"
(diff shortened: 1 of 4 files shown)


Turbo87:
btw I'm still wondering if `range` is the right route name. the `semver` npm dependency calls it "ranges", but the `semver` crates calls it a `VersionReq`. we could also consider naming the route `semver` instead or something like that. feel free to suggest better alternatives :)

pichfl:
I think `range` should be acceptable for now.

@bors r+

PR #3270, opened 2021-02-10
Update some crates to reduce deps

JohnTitor:
This updates only the crates that reduce/replace their dependencies. Each commit has the `cargo update` log.

## Removed
- bytes v0.5.6
- generator v0.6.23
- loom v0.3.6
- scoped-tls v1.0.0
- maybe-uninit v2.0.0

## Replaced
- smallvec v0.6.13 -> tinyvec v1.1.1

r? @jtgeibel

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -941,19 +941,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "generator"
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):
LGTM 👍

PR #3277, opened 2021-02-10
routes/application: Add `routeName` tag to all issues

Turbo87:
This will hopefully make it easier to identify which routes have the most issues 😅

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -2,11 +2,20 @@ import { action } from '@ember/object';
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 
+import * as Sentry from '@sentry/browser';
+
 export default class ApplicationRoute extends Route {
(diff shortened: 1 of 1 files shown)


app/routes/application.js
```diff
@@ -2,11 +2,20 @@ import { action } from '@ember/object';
...
 
+import * as Sentry from '@sentry/browser';
+
 export default class ApplicationRoute extends Route {
   @service progress;
+  @service router;
   @service session;
 
   beforeModel() {
+    this.router.on('routeDidChange', () => {
```
locks, line 13: Don't you want to put this in the router itself?

Turbo87: hmm, I thought extending the router was discouraged these days 🤔

locks: Hhmm, yeah let's leave it like this.

PR #3289, opened 2021-02-13
CI: Bump cargo binaries cache version

Turbo87:
This should hopefully allow cargo-tarpaulin to be cached correctly


Turbo87:
The "Install cargo-tarpaulin" step now took 0 sec. after rebasing this branch, so I guess this was successful :)

PR #3292, opened 2021-02-13
Address clippy lints currently in beta

jtgeibel:
r? @Turbo87


jtgeibel:
It doesn't seem like the new `cargo install cargo-tarpaulin` step is being cached correctly here. I think this should already be cached from the build on master.

https://github.com/rust-lang/crates.io/runs/1893654950#step:17:1

jtgeibel:
Disregard my last comment. I just now saw https://github.com/rust-lang/crates.io/pull/3289.

Turbo87 (approved):

PR #3296, opened 2021-02-14
DownloadGraph: Ignore SemVer parsing issues when sorting

### LINKED ISSUE #3295: TypeError: Invalid major version
Sentry Issue: [CRATES-IO-FRONTEND-P2](https://sentry.io/organizations/rust-lang/issues/2212224874/?referrer=github_integration)

```
TypeError: Invalid major version
  at new c (addon-tree-output/~ember-css-modules/ext.js:2:1)
  at e.exports (addon-tree-output/~ember-css-modules/ext.js:2:1)
  at None (addon-tree-output/~ember-css-modules/ext.js:2:1)
  at Array.sort (<anonymous>)
  at default (addon-tree-output/~ember-css-modules/ext.js:2:1)
...
(19 additional frame(s) were not displayed)
```

plaets:
Fixes #3295

### DIFF
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -106,7 +106,14 @@ export default class DownloadGraph extends Component {
       }
     });
 
-    let versionsList = semverSort([...versions]);
+    let versionsList = [...versions];
+    try {
+      semverSort(versionsList);
+    } catch {
+      // Catches exceptions thrown when a version number is invalid
+      // see issue #3295
+    }
+
     if (extra.length !== 0) {
       versionsList.unshift('Other');
     }


app/components/download-graph.js
```diff
@@ -106,7 +106,14 @@ export default class DownloadGraph extends Component {
       }
     });
 
-    let versionsList = semverSort([...versions]);
+    let versionsList = [...versions];
+    try {
+      semverSort(versionsList);
```
Turbo87, line 111: does this sort the array in-place?

plaets: I think it does, unless I'm misunderstanding something
https://github.com/rust-lang/crates.io/blob/master/app/components/download-graph.js#L8
https://github.com/npm/node-semver/blob/master/functions/sort.js#L2
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort
> The sort() method sorts the elements of an array in place

plaets: I also checked how the code behaves when I comment out semverSort and it seems to confirm that

Turbo87: okay, great, thanks :)

Turbo87 (approved):

PR #3299, opened 2021-02-14
Various CI improvements

jtgeibel:
This series makes several improvements to the CI workflow, in particular for caching. It is recommended to review as individual commits.

The biggest change is that the cargo registry directories are cached under a common key that is not rustc version specific. The first job in the matrix to store the cache will win. With the exception of things that are `cargo install` on a per channel basis, the cargo caches should be the same between versions.


.github/workflows/ci.yml
```diff
@@ -117,42 +117,24 @@ jobs:
       #
       # The other binaries (like `rustc`) will be overwritten by rustup.
       #
-      # Current size as of 2019-12-23: ~6 MB
+      # Current size as of 2021-02-15: ~15 MB
       - name: Cache cargo binaries
-        uses: actions/cache@v2
+        uses: actions/cache@main
```
Turbo87, line 122: what's the benefit of running `main`? it looks like the README of https://github.com/actions/cache recommends using `v2`.

.github/workflows/ci.yml
```diff
@@ -174,17 +156,14 @@ jobs:
...
 
-      # Current size as of 2019-12-23: ~336 MB
+      # Current size as of 2021-02-12: ~325 MB
       - name: Cache cargo build
-        uses: actions/cache@v2
+        uses: actions/cache@main
         with:
           path: target
-          key: ${{ runner.os }}-cargo-build-target-${{ steps.rustc.outputs.version }}-${{ hashFiles('**/Cargo.lock') }}
+          key: v2-${{ runner.os }}-cargo-build-target-${{ steps.rustc.outputs.version }}-${{ hashFiles('**/Cargo.lock') }}
```
Turbo87, line 164: do we still need this cache bump?

.github/workflows/ci.yml
jtgeibel: `v2.1.4` has more verbose output, which I think will be helpful while we evaluate cache performance. For some reason, the `v2` tag still points to the same commit as `v2.1.3` from early November. We could use `v2.1.4` instead of `main`, but in either case I agree we'll want to move back to `v2` eventually.

.github/workflows/ci.yml
jtgeibel: I'm not sure if we need to or not. I think the bins from `cargo install` are the only cache key that hasn't changed, and while we probably don't have to change this key it will help ensure that we're evaluating the performance from a clean slate.

.github/workflows/ci.yml
Turbo87: huh, interesting. I guess I'd prefer to use `v2.1.4` then instead of the moving `main` target.

.github/workflows/ci.yml
Turbo87: okay, sounds reasonable 👍

.github/workflows/ci.yml
jtgeibel: Okay, I've added a commit switching to `v2.1.4`.

PR #3304, opened 2021-02-16
Improve invalid semver string handling

### LINKED ISSUE #3294: TypeError: Cannot read property 'prerelease' of null
Sentry Issue: [CRATES-IO-FRONTEND-P0](https://sentry.io/organizations/rust-lang/issues/2212203651/?referrer=github_integration)

```
TypeError: Cannot read property 'prerelease' of null
  at isPrerelease (cargo/models/version.js:155:1)
  at obj (cargo/models/version.js:166:1)
  at _getProp (@ember/-internals/metal/index.js:1375:1)
  at compute (@glimmer/reference.js:213:1)
  at callback (@glimmer/reference.js:161:1)
...
(45 additional frame(s) were not displayed)
```

plaets:
* Add a tooltip message and a new release track indicator (`?`) to components/version-list/row
* Handle null returned from `semverParse` in models/version
* Add `loose` option to `semverParse` (https://github.com/npm/node-semver#functions)

I placed the checks so that "invalid version" has higher priority than "first version" but lower than "yanked".
I also modified css of the version list so that long versions do not overflow from tooltips. 
Before
image
After
image

 

Fixes #3294

### DIFF
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -108,7 +108,7 @@ export default class DownloadGraph extends Component {
 
     let versionsList = [...versions];
     try {
-      semverSort(versionsList);
+      semverSort(versionsList, { loose: true });
     } catch {
       // Catches exceptions thrown when a version number is invalid
       // see issue #3295
--- a/app/components/version-list/row.hbs
+++ b/app/components/version-list/row.hbs
@@ -12,6 +12,8 @@
     <div local-class="release-track">
       {{#if @version.yanked}}
         {{svg-jar "trash"}}
+      {{else if @version.invalidSemver}}
+        ?
       {{else if @version.isFirst}}
         {{svg-jar "star"}}
       {{else}}
--- a/app/components/version-list/row.hbs
+++ b/app/components/version-list/row.hbs
@@ -9,7 +9,7 @@
   ...attributes
 >
   <div local-class="version">
-    <div local-class="release-track">
+    <div local-class="release-track" data-test-release-track>
       {{#if @version.yanked}}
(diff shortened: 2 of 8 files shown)


Turbo87:
looks great. could you check if it's possible to add a few tests for these cases? 🙏

plaets:
I added some tests, but honestly, I'm not very familiar with ember so I'm not entirely happy with the results. If you have any comments, anything I should or should not test, please let me know.
Main issues:
* I couldn't figure out how to test the tooltip text, I tried [this](https://github.com/sir-dunxalot/ember-tooltips#test-helper-option-targetselector) with `[data-test-release-track-title]` as the selector, but it just fails to find the element. 
* I did not add the `test-max-version-example-crate` case to [happy path](https://github.com/plaets/crates.io/blob/c0e3a96e7393f4cb8dad4c009d945b2d9c27b5dd/tests/models/version-test.js#L101), [mirage/serializers/crate](https://github.com/plaets/crates.io/blob/c0e3a96e7393f4cb8dad4c009d945b2d9c27b5dd/mirage/serializers/crate.js#L59) fails to sort arrays with invalid versions which causes "Ember Data Request GET /api/v1/crates/crate-0 returned a 500 Payload" and I'm not sure how to solve that. 
* Is `components/version-list-row-test.js` a good name/path? `ember g component-test version-list/row` would generate the test files in a completely new path and I wasn't sure which one to pick.

And sorry for force-push spam, I forgot to remove/check some stuff before pushing.

Turbo87:
looks great, thanks!

@bors r+

Issue #2630, opened 2020-07-12
security: Recovering from a GitHub account compromise


jsha:
Right now, if an attacker compromises a crate author's GitHub account, they can use it to log into crates.io and get a session cookie. That cookie is signed by an HMAC key owned by crates.io, and contains the user ID. During authentication, crates.io verifies the HMAC on that cookie and then trusts the user ID inside. The problem is that, once the crate author regains control of their GitHub account, the attacker's session cookie is still valid, and as far as I can tell there is no way to revoke it. There are a few possible solutions:

 - Instead of a signed cookie, use a `sessions` table that has unique tokens in it. This has the added advantage that the sessions table can include the IP address and User-Agent from which the session was initiated, so the user can review extant sessions and revoke them if necessary.
 - Expire the signed cookie after a period of time (https://github.com/conduit-rust/conduit-cookie/issues/15). This is useful for defense-in-depth, but any reasonable expiration period would be too long to work for this purpose. When you want to revoke an attacker's access, you want to revoke it right away, not in a few days.
 - Add a generation number in the users table, and give the user an option to "sign out all other sessions," which would increment the generation number. The generation number would also be included in the signed cookie and required to match the current values in the table.

I think (1) is the best option but also the most work.

Another, closely related issue: If a user revokes the crates.io app on their GitHub account (for instance, as part of the cleanup after a compromise), it would be nice to temporarily disable all cookie sessions and tokens on their crates.io account until they reauthorize crates.io.

(Related: #815)

Turbo87:
@jsha FYI we shortly discussed this during our weekly team meeting and we're considering to implement the `session` table suggestion. would you be available for reviews once we have something that works?

jsha:
Excellent! Yes, I'd definitely be happy to help with reviews.

jsha:
Another item in this area: It might be good to re-verify the GitHub OAuth token at publish time. If the OAuth token has been revoked, the publish should fail.

Shnatsel:
There technically is a way to recover: the HMAC key can be changed by crates.io. But that invalidates all the cookies for all users, and can only be performed by the crates.io team.

adsnaider:
Hello! Jumping in as this issue is important to folks at Google (myself included). I've gone ahead and created a pull request (#4690) that addresses the issue.

> Disclaimer: I used #3307 to help me get started, so thank you very much @Turbo87 for the help :)

[...]

PR #3307, opened 2021-02-16
WIP: Implement `sessions` database table

Turbo87:
As discussed in the team meeting last week, this PR implements a `sessions` table in the database to address https://github.com/rust-lang/crates.io/issues/2630.

This is still work-in-progress, but feel free to comment if you have any concerns.

/cc @jsha @jtgeibel


migrations/2021-02-14-201950_session-table/up.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE sessions
+(
+    id SERIAL
+        CONSTRAINT sessions_pk
+            PRIMARY KEY,
+    user_id INTEGER NOT NULL
+        CONSTRAINT sessions_users_id_fk
+            REFERENCES users
+            ON UPDATE CASCADE ON DELETE CASCADE,
```
jsha, line 9: You'll probably want to delete stale sessions eventually. In a big database, deleting old rows can be relatively expensive. For this, you might want to use partitioning (https://www.postgresql.org/docs/10/ddl-partitioning.html), which lets you drop old, stale partitions cheaply while still treating the partitions as a single table. However, to use partitioning you need to have no foreign keys on the partitioned table. Suggestion: Make user_id not a foreign key. We don't expect that user_ids change, so we don't need `ON UPDATE CASCADE`. And if the users entry is deleted, we can simply treat the session as invalid. So we don't need `ON DELETE CASCADE`.

src/controllers/user/session.rs
```diff
@@ -143,6 +168,28 @@ fn save_user_to_database(
...
+
+        // remove the `cargo_auth` cookie
+        req.cookies_mut().remove(auth_cookie("", secure));
+
+        // try to revoke the session in the database, but explicitly
+        // ignore failures
+        let _result: Result<_, Box<dyn AppError>> = req
+            .db_conn()
+            .map_err(Into::into)
+            .and_then(|conn| Session::revoke_by_token(&conn, &token).map_err(Into::into));
```
jsha, line 190: Perhaps log or increment a stat in this case? Or have a set of expected errors (like timeout) and pass through unexpected errors (like permission denied or malformed SQL).

src/models/session.rs
```diff
@@ -0,0 +1,290 @@
...
+            diesel::update(sessions)
+                .set((
+                    sessions::last_used_at.eq(diesel::dsl::now),
+                    sessions::last_ip_address.eq(IpNetwork::from(ip_address)),
+                    sessions::last_user_agent.eq(user_agent),
+                ))
+                .get_result(conn)
+                .optional()
+        })
+        .or_else(|_| sessions.first(conn).optional())
```
jsha, line 74: It's probably a good idea to do the update only if:

 - The IP address has changed, or
 - The User-Agent has changed, or
 - The last_used_at has increased by more than 15 minutes.
 
 This will save you from having to do a DB write on every cookie-authenticated request.

jsha (changes requested):
This looks awesome! Thanks for writing it. One optimization: If the user's cookie asserts both a session ID and a token, you can avoid having an index on the hashed_token field. Indexes can take up a surprising amount of memory.

Turbo87:
closing this for now due to all the conflicts and the foundation not having given us an answer yet wether we can do this or not... 🤷‍♂️

PR #3317, opened 2021-02-20
Fix dashboard 'Load More' margin

wyhaya:
### Before
before

### After
after


Turbo87:
LGTM, thanks :)

@bors r+

PR #3320, opened 2021-02-22
Add SecureToken to enforce good token hygiene

pietroalbini:
This PR adds the `SecureToken` struct, which relies on the type system to enforce good practices for generating and verifying tokens:

* `SecureToken` forces adding an unique prefix in front of every kind of generated token. This allows token scanning services to properly identify tokens generated by crates.io, and allows crates.io to detect which kind of token it's about to revoke.
* `SecureToken` only stores and exposes the SHA256 of already generated tokens, to prevent accidental plaintext leakages. The plaintext token value is only exposes for newly generated tokens.

r? @Turbo87


src/util.rs
```diff
@@ -13,6 +12,7 @@ mod io_util;
 mod request_helpers;
 mod request_proxy;
 pub mod rfc3339;
+pub(crate) mod token;
```
Turbo87, line 15: what's the reason for `pub(crate)`? wouldn't that make it hard to use from the integration test code?

pietroalbini: If tests actually end up needing access to `SecureToken` we could make the relevant things public. In general I prefer to keep as much as possible `pub(crate)` instead of `pub`, to enable unused warnings.

src/util/token.rs
```diff
@@ -0,0 +1,133 @@
...
+    const CHARS: &[u8] = b"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+
+    OsRng
+        .sample_iter(Uniform::from(0..CHARS.len()))
+        .map(|idx| CHARS[idx] as char)
+        .take(len)
+        .collect()
+}
+
+macro_rules! secure_token_kind {
```
Turbo87, line 69: I find this a bit hard to read. Did you consider using something like `strum` instead of this custom macro? If we'll still go with the custom macro I think it'd be good to add a few more doc comments on it at least :)

src/models/token.rs
```diff
@@ -98,14 +89,6 @@ mod tests {
     use crate::views::EncodableApiTokenWithToken;
     use chrono::NaiveDate;
 
-    #[test]
-    fn test_token_constants() {
-        // Changing this by itself will implicitly revoke all existing tokens.
-        // If this test needs to be change, make sure you're handling tokens
-        // with the old prefix or that you wanted to revoke them.
-        assert_eq!("cio", TOKEN_PREFIX);
```
Turbo87, line 106: maybe we should move this test to `util/token.rs`?

src/util/token.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_generated_and_parse() {
+        const KIND: SecureTokenKind = SecureTokenKind::API;
```
Turbo87, line 115: this might be a stupid question, but why is this a `const` (vs `let`)?

src/util/token.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        let token = SecureToken::generate(KIND);
+        assert!(token.plaintext().starts_with(KIND.prefix()));
+        assert_eq!(
+            token.sha256(),
```
Turbo87, line 120: interesting. so due to the `Deref` the `NewSecureToken` struct sort of inherits all of the methods from `SecureToken`? didn't know that was possible without any extra `*` or other cryptic characters 😅

src/models/token.rs
```diff
@@ -29,41 +26,35 @@ pub struct ApiToken {
...
                 api_tokens::user_id.eq(user_id),
                 api_tokens::name.eq(name),
-                api_tokens::token.eq(digest(&plaintext, "sha256")),
+                api_tokens::token.eq(token.sha256()),
             ))
             .get_result(conn)?;
 
-        Ok(CreatedApiToken { plaintext, model })
+        Ok(CreatedApiToken {
+            plaintext: token.plaintext().into(),
```
Turbo87, line 43: not terribly important but I'm wondering if it would be possible to avoid the string clone here. if I understand correctly the `token` is not used beyond this point so it should be possible to consume it to get the original plaintext `String`, right?

src/util.rs
Turbo87: > to enable unused warnings.

hmm, I could have sworn that I saw such warnings for `pub` code too when I was working on the session table stuff. but I'm okay with keeping it as `pub(crate)` for now until we actually need it.

src/util/token.rs
pietroalbini: Mostly personal preference.

src/models/token.rs
pietroalbini: Yes, but we would need to move it out of the token. I also don't think this is really important.

src/util/token.rs
pietroalbini: Yes, that's pretty nice when wrapping other structs. It's the same reason why `String` has all the `&str` methods.

src/util/token.rs
pietroalbini: I don't think it's worth bringing another dependency in just for this. I'll try to add some comments to the macrol.

src/util.rs
Turbo87: I see, thanks :)

src/util/token.rs
Turbo87: I guess this is also fine to retrofit. I'm approving the PR in the meantime :)

PR #3328, opened 2021-02-25
Simplify `authenticate_user()` function

Turbo87:
The function was a bit hard to read and difficult to extend (see #3307). This PR simplifies it a little bit by returning early if a suitable authentication method was found.

r? @jtgeibel


JohnTitor (approved):
LGTM 👍

PR #3329, opened 2021-02-25
Import `htmlSafe` from `@ember/template`

Turbo87:
This should resolve a few of the deprecation warnings that got introduced with Ember 3.25


locks (changes requested):
There's actually a bug and no way to get rid of this deprecation until a patch version of 3.25 with the deprecation disabled lands.

Turbo87:
@locks sure, but I guess adjusting the import paths is still helpful, right?

pichfl:
Out with the old, in with the new

@bors r+

PR #3333, opened 2021-02-25
Display crate description in the `CrateHeader` component

Turbo87:
Bildschirmfoto 2021-02-25 um 18 52 09

This makes it easier to figure out on first glance what a crate does, without having to wait for and read through the README text.


Turbo87:
Bildschirmfoto 2021-02-25 um 22 43 13

I've slightly adjusted the case where the crate does not have a README to avoid displaying the crate description twice, and I've also added a spinner for the README loading task.

PR #3338, opened 2021-02-26
Prep for semver bump and document current blocker 

jtgeibel:
This partially reverts the revert (#3051) of #2990. While we can't bump the version yet, it is worth recovering these changes and documenting the remaining work.

r? @Turbo87

### DIFF
--- a/src/admin/test_new_semver_release.rs
+++ b/src/admin/test_new_semver_release.rs
@@ -1,0 +1,71 @@
+use crate::{
+    db,
+    schema::{dependencies, versions},
+};
+
+use clap::Clap;
+use diesel::prelude::*;
+
+#[derive(Clap, Copy, Clone, Debug)]
+#[clap(
+    name = "test-new-semver-release",
+    about = "Scan the database for all versions and version requirements to \
+        ensure they can be parsed by the latest `semver` crate."
+)]
+pub struct Opts {}
+
+pub fn run(_opts: Opts) {
+    use self::dependencies::dsl::*;
+    use self::versions::dsl::*;
+
+    let url = db::connection_url(&dotenv::var("READ_ONLY_REPLICA_URL").unwrap());
+    let conn = PgConnection::establish(&url).unwrap();
+    conn.transaction::<_, diesel::result::Error, _>(|| {
+        let vers = versions.select(num).load(&conn)?;
+        test_versions(&vers);
+
+        let deps = dependencies.select(req).load(&conn)?;
+        test_dependency_predicates(&deps);
+
+        Ok(())
+    })
+    .unwrap();
+
+    println!("Test finished.");
+}
+
+fn test_versions(versions: &[String]) {
+    for version in versions {
+        if let Err(e) = semver::Version::parse(version) {
+            println!("Could not parse `{}` as a semver::Version: {}", version, e);
+        }
+    }
+
+    for version in versions {
+        if let Err(e) = semver::Version::parse(version) {
+            println!("Could not parse `{}` as a semver::Version: {}", version, e);
+        }
+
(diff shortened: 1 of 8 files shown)


Turbo87:
looks like CI is unhappy 😢

jtgeibel:
Forgetting to `cargo fmt` strikes again. Should work now.

Turbo87:
you linked https://github.com/rust-lang/crates.io/pull/3094 in the description, but I'm assuming that was a typo, right? 🤔

Turbo87:
I believe https://github.com/steveklabnik/semver/issues/217 might still be blocking this

jtgeibel:
@Turbo87 you're right, I meant to link to #3049. And you're right, I think the linked issue is blocking this.

jtgeibel:
@Turbo87 I've rolled back the version bump and added a comment pointing to the blocker. I've updated the PR title and description to reflect the new scope.

Turbo87 (approved):
would it make sense for us to write a failing test for the blocker so that we don't update accidentally?

feel free to r=me :)

jtgeibel:
@Turbo87 I've added a test that covers some scenarios where I observed breakage. I'll probably move this test upstream eventually, but it is worth including here for now.

I've also added a command to the `crates-admin` binary, that will test existing versions and dependency requirements in the database against our current version and the latest `semver` release.

src/admin/test_new_semver_release.rs
```diff
@@ -0,0 +1,71 @@
...
+    use self::dependencies::dsl::*;
+    use self::versions::dsl::*;
+
+    let url = db::connection_url(&dotenv::var("READ_ONLY_REPLICA_URL").unwrap());
+    let conn = PgConnection::establish(&url).unwrap();
+    conn.transaction::<_, diesel::result::Error, _>(|| {
+        let vers = versions.select(num).load(&conn)?;
+        test_versions(&vers);
+
+        let deps = dependencies.select(req).load(&conn)?;
```
jtgeibel, line 27: I'm not entirely sure about loading all dependencies in a single query, but this is against the replica database and I expect memory usage to be somewhere around 100 MB so I think this is okay for our current scale.

src/admin/test_new_semver_release.rs
```diff
@@ -0,0 +1,71 @@
...
+    println!("Test finished.");
+}
+
+fn test_versions(versions: &[String]) {
+    for version in versions {
+        if let Err(e) = semver::Version::parse(version) {
+            println!("Could not parse `{}` as a semver::Version: {}", version, e);
+        }
+    }
+
```
Turbo87, line 43:
```suggestion
```

this looks like a duplicate to me, no?

src/admin/test_new_semver_release.rs
```diff
@@ -0,0 +1,71 @@
...
+                version, e
+            );
+        }
+    }
+}
+
+fn test_dependency_predicates(versions: &[String]) {
+    for version in versions {
+        if let Err(e) = semver::VersionReq::parse(version) {
+            println!("Could not parse `{}` as a semver::Version: {}", version, e);
```
Turbo87, line 61:
```suggestion
            println!("Could not parse `{}` as a semver::VersionReq: {}", version, e);
```

src/admin/test_new_semver_release.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        if let Err(e) = semver_next::VersionReq::parse(version) {
+            println!(
+                "Could not parse `{}` as a semver_next::Version: {}",
```
Turbo87, line 66:
```suggestion
                "Could not parse `{}` as a semver_next::VersionReq: {}",
```

PR #3347, opened 2021-02-27
Move keyword list from `CrateSidebar` to `CrateHeader`

Turbo87:
This moves the list of crate keywords right below the crate description in the header to make them easier to discover and navigate :)

Bildschirmfoto 2021-02-27 um 11 57 56

### DIFF
--- a/app/components/crate-header.hbs
+++ b/app/components/crate-header.hbs
@@ -12,10 +12,23 @@
       <FollowButton @crate={{@crate}}/>
     {{/if}}
   </div>
+
   {{#if @crate.description}}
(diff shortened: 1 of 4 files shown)


pichfl:
Not sure about this one, I was rather happy to see a slimmer header 🤔

Turbo87:
I've found it quite useful to see on first glance what a crate is about. It also helps a bit with finding related crates. I've been running this locally for a few days now and I'm quite happy with the result.

pichfl:
Good point 👍 

@bors r+

PR #3366, opened 2021-03-05
CrateSidebar: Move Playground explanation to a tooltip

Turbo87:
### Before
Bildschirmfoto 2021-03-05 um 22 56 58

### After
Bildschirmfoto 2021-03-05 um 22 53 25

This move a bit of noise from the sidebar into a tooltip that is shown when hovering over the button.

### DIFF
--- a/app/components/crate-sidebar.hbs
+++ b/app/components/crate-sidebar.hbs
@@ -139,11 +139,17 @@
         data-test-playground-button
       >
         Try on Rust Playground
+
+        {{#if this.canHover}}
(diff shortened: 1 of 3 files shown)


pichfl:
Would it be possible to do this on devices with mouse input only? Touch devices and tooltips do not mix well.

Turbo87:
@pichfl hmm, that's a fair point. any idea on how to implement that though? especially considering the fastboot requirement this would seem quite hard to do 🤔

pichfl:
@Turbo87 You can use [`@media (pointer: coarse) {}`](https://developer.mozilla.org/de/docs/Web/CSS/@media/pointer) to write CSS that's only visible on touch devices.

PR #3369, opened 2021-03-07
Use singular "feature" for versions with one feature

camelid:
I tested this change locally and confirmed that it works.


Turbo87:
looks good, thanks!

Can you check if it's possible to add a test for this? :)

camelid:
> Can you check if it's possible to add a test for this? :)

I have very little knowledge of Ember, so if you could tell me where to put the test and link me to a similar test that I could draw from, that would be very helpful :)

plaets:
Sorry for answering for Trubo, I had a similar problem when I wrote tests for VersionList::Row so I thought I could share what I learned. 

I placed the tests in [tests/components/version-list-row-test.js](https://github.com/rust-lang/crates.io/blob/master/tests/components/version-list-row-test.js), since it's the same component you probably should add your tests there too. I think you can just copy the whole `handle non-standard semver strings` fixture. You will need to add a `features` key to the objects that are passed to `this.server.create('version', {...})` and then change the `assert.dom(...)` lines, [qunit-dom docs are here](https://github.com/simplabs/qunit-dom/blob/master/API.md), `hasText` should be good enough. It could be useful to add an empty `data-test-something` attribute to the `.num-features` span so that querying it with `assert.dom` is easier.

Just in case: it's probably a good idea to use something like `npx ember exam --module 'Component | VersionList::Row'` (assuming you implement your tests in the file I linked) to run only that particular test. If you also pass `--server`, it will open the test in a browser, really useful for debugging.

camelid:
Thank you for the help @plaets! I tried running the tests locally but it unfortunately didn't work. I can't tell what's wrong because there's a deprecation warning printed hundreds of times that filled up my terminal 😆

Hopefully I'll be able to see if the test works based on CI.

tests/components/version-list-row-test.js
```diff
@@ -44,4 +44,46 @@ module('Component | VersionList::Row', function (hooks) {
...
+    this.firstVersion = versions[0];
+    this.secondVersion = versions[1];
+    this.thirdVersion = versions[2];
+
+    await render(hbs`<VersionList::Row @version={{this.firstVersion}} />`);
+    assert.dom('[local-class="num-features"]').hasText('0 Features');
+
+    await render(hbs`<VersionList::Row @version={{this.secondVersion}} />`);
+    assert.dom('[local-class="num-features"]').hasText('1 Feature');
+    !assert.dom('[local-class="num-features"]').hasText('Features');
```
camelid, line 84: Is there an existing command like `notHasText` or is negating `hasText` the right way to do it?

camelid: Oops, I guess I was thinking in a purely-functional way :)

camelid:
The test failed with:

```
not ok 161 Chrome 88.0 - [114 ms] - Exam Partition 1 - Component | VersionList::Row: pluralize "feature" only when appropriate
    ---
        stack: >
                at DOMAssertions.findTargetElement (http://localhost:7357/assets/test-support.js:36301:14)
                at DOMAssertions.hasText (http://localhost:7357/assets/test-support.js:35752:26)
                at Object.<anonymous> (http://localhost:7357/assets/tests.js:2675:50)
        message: >
            Element [local-class="num-features"] should exist
        negative: >
            false
        browser log: |
    ...
```

I'm wondering if that's because `local-class` means that the class won't show up in the rendered DOM or something. I'll try adding a `data-test-...` attribute as suggested.

camelid:
Hmm, it's still failing. Any idea why?

```
not ok 161 Chrome 88.0 - [118 ms] - Exam Partition 1 - Component | VersionList::Row: pluralize "feature" only when appropriate
    ---
        stack: >
                at DOMAssertions.findTargetElement (http://localhost:7357/assets/test-support.js:36301:14)
                at DOMAssertions.hasText (http://localhost:7357/assets/test-support.js:35752:26)
                at Object.<anonymous> (http://localhost:7357/assets/tests.js:2675:46)
        message: >
            Element [data-test-feature-list] should exist
        negative: >
            false
        browser log: |
    ...
```

plaets:
`num-features` is hidden when a version has no features

tests/components/version-list-row-test.js
```diff
@@ -44,4 +44,46 @@ module('Component | VersionList::Row', function (hooks) {
...
+
+    let store = this.owner.lookup('service:store');
+    let crateRecord = await store.findRecord('crate', crate.name);
+    let versions = (await crateRecord.versions).toArray();
+    this.firstVersion = versions[0];
+    this.secondVersion = versions[1];
+    this.thirdVersion = versions[2];
+
+    await render(hbs`<VersionList::Row @version={{this.firstVersion}} />`);
+    assert.dom('[data-test-feature-list]').hasText('0 Features');
```
Turbo87, line 80:
```suggestion
    assert.dom('[data-test-feature-list]').doesNotExist();
```

tests/components/version-list-row-test.js
```diff
@@ -44,4 +44,46 @@ module('Component | VersionList::Row', function (hooks) {
...
+
+    await render(hbs`<VersionList::Row @version={{this.secondVersion}} />`);
+    assert.dom('[data-test-feature-list]').hasText('1 Feature');
+    assert.dom('[data-test-feature-list]').doesNotIncludeText('Features');
```
Turbo87, line 84:
```suggestion
```

`hasText()` compares via `==` and since we know that it is `1 Feature` we can be sure that it can't include `Features` at the same time 😉

Turbo87 (commented):
thanks for the tests! a few small adjustments and then this should be ready to go :)

camelid:
Rebased because of #3390.

camelid:
Squashed.

Turbo87:
thank you! :)

@bors r+

PR #3371, opened 2021-03-07
Extract font-family CSS variables

Turbo87:
This PR primarily unifies the font-family that is used for monospace display.

### DIFF
--- a/app/styles/application.module.css
+++ b/app/styles/application.module.css
@@ -11,6 +11,7 @@
     --footer-bg-color: var(--green800);
 
     --font-family: "Fira Sans", sans-serif;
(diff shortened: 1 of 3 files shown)


app/styles/application.module.css
```diff
@@ -10,7 +10,8 @@
     --header-bg-color: var(--green800);
     --footer-bg-color: var(--green800);
 
-    --font-family: "Fira Sans", sans-serif;
+    --font-primary: "Fira Sans", sans-serif;
```
pichfl, line 13: If we have `--font-monospace` I would prefer naming this one `--font-sans`

PR #3373, opened 2021-03-07
Move router error handling into new middleware layer

jtgeibel:
This moves handling of routing errors into middleware and adds a test for the conversion.

Additionally we previously called `router.recognize()` instead of `router.call()`. By switching we can let the `conduit-router` logic handle adding `Params` to the request extensions.

A potential allocation is removed from json error handling, although the compiler may be able to eliminate that allocation already.

r? @Turbo87

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -191,7 +191,7 @@ mod tests {
     };
     use crate::util::EndpointResult;
 
-    use conduit::StatusCode;
+    use conduit::{Body, StatusCode};
     use conduit_test::MockRequest;
     use diesel::result::Error as DieselError;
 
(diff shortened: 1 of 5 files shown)


src/router.rs
```diff
@@ -256,17 +242,4 @@ mod tests {
...
-        let route_builder = RouteBuilder::new();
-        let mut req = MockRequest::new(::conduit::Method::GET, "/");
-
-        let (parts, body) = R404(route_builder).call(&mut req).unwrap().into_parts();
-        assert_eq!(parts.status, StatusCode::NOT_FOUND);
-        assert!(matches!(
-            body,
-            Body::Owned(vec) if vec == br#"{"errors":[{"detail":"Not Found"}]}"#
-        ));
-    }
```
Turbo87: would it make sense to convert that to a more high-level API test?

Turbo87 (approved):
LGTM

one small comment, but feel free to r=me

src/router.rs
jtgeibel: This particular test moved to the middleware. We could do something under `tests/all.rs` as well. The main difference would be that it would load other middleware layers as well.

PR #3391, opened 2021-03-10
Refactor some test code

jtgeibel:
It is best to review per-commit, as some logic is split into sub-modules under `util` which obscures the smaller diffs within. The individual commits should all be straightforward.

In particular, many `T: serde::Deserialize` bounds are removed from various test helper methods, which allowed some remaining on-off tests to switch over to the test helpers.

r? @Turbo87


Turbo87:
yes, please!! 👍 

@bors r+

Issue #3305, opened 2021-02-16
Display no data instead of zero on the downloads graph when the version didn't exist


kpreid:
**Is your feature request related to a problem? Please describe.**

On the "Downloads over the last 90 days" graph for a crate, points of zero downloads appear even if the version in question did not exist on that day. This means that every crate which has been updated in the past 90 days has one or more non-informative rows of zero-valued points at the bottom of the chart.

![image](https://user-images.githubusercontent.com/779501/108124349-efa78480-705b-11eb-86c8-81e1cf111198.png)

**Describe the solution you'd like**

I think the graph would be prettier, more informative, and more readable if these data points were omitted entirely, so that each version's line begins at the day it was published. (Note that this is not the same as hiding all points which have a zero value.)

In addition to removing clutter, this would also result in the graph revealing the release dates of each new version, without needing any additional symbols.

**Additional context**

To illustrate the potential value further, here is another screenshot from a more active crate. Removing the burgundy line of version `0.7.0-alpha.3`, in the span before the version existed (Feb 3), would make it easier to view and understand the activity of the preceding version `0.7.0-alpha.2`, and even compare it to its successor.

![image](https://user-images.githubusercontent.com/779501/108125826-e28b9500-705d-11eb-90fb-0bf5ea6f903c.png)

Turbo87:
sounds like a great idea to me. do you want to try implementing it? :)

PR #3393, opened 2021-03-11
DownloadGraph: Hide chart points from before a version was released

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/3305

Example for the `rand` crate:

Bildschirmfoto 2021-03-10 um 15 32 34


Turbo87:
and don't be scared by the long diff. the majority is a testing fixture... 😅

app/components/download-graph.js
```diff
@@ -67,81 +75,109 @@ export default class DownloadGraph extends Component {
...
-      versions.add(version_num);
+    let version_num = version.num;
 
-      let key = d.date;
-      if (dates[key]) {
-        let prev = dates[key].cnt[version_num] || 0;
-        dates[key].cnt[version_num] = prev + d.downloads;
-      }
-    });
+    versions.set(version_num, version);
```
pichfl, line 105: Is there a reason this uses underscored variables over camelCase like most of the other code here?

Turbo87: only legacy reasons 😉

Turbo87:
Fixed CI by using https://github.com/ozkxr/match-json for the download data assertion to make it possible to use predicate functions when comparing JSON objects. :)

PR #3395, opened 2021-03-11
CrateSidebar::Link: Fix missing line breaks in long URLs

Turbo87:
Some crates use long URLs for their websites and this currently makes the sidebar look ugly, because the links don't break properly. This can be reproduced on e.g. https://crates.io/crates/criterion

This PR fixes it by 1) adjusting the CSS of the links to allow for more line breaks, and 2) inserting zero-width spaces around `/` and `.` characters in the URL to allow for more natural line breaks.

### Before

Bildschirmfoto 2021-03-11 um 18 00 59

### After

Bildschirmfoto 2021-03-11 um 18 00 36

### DIFF
--- a/app/components/crate-sidebar/link.hbs
+++ b/app/components/crate-sidebar/link.hbs
@@ -9,7 +9,7 @@
       {{svg-jar "link" local-class="icon" data-test-icon="link"}}
     {{/if}}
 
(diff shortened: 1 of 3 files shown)


Turbo87:
closing in favor of https://github.com/rust-lang/crates.io/pull/3396

PR #3397, opened 2021-03-12
Log how much each step takes on slow downloads

pietroalbini:
We're getting increasingly frequent pages about quick bursts of really slow download requests. Unfortunately the amount of logging and monitoring we currently have available is not enough to properly pin down what's causing this.

This PR adds instrumentation to the download endpoint to log how much each potentially slow step (getting a db connection, getting the version info and updating the download count) takes, and the infrastructure to support such instrumentation.

To avoid outputting too much stuff into the log lines, the detailed timings are only shown on slow requests.

r? @jtgeibel

### DIFF
--- a/src/controllers.rs
+++ b/src/controllers.rs
@@ -19,6 +19,7 @@ mod prelude {
 
     pub use crate::db::RequestTransaction;
     pub use crate::middleware::app::RequestApp;
+    pub use crate::middleware::log_request::TimingRecorder;
     pub use crate::util::errors::{cargo_err, AppError, AppResult, ChainError}; // TODO: Remove cargo_err from here
     pub use crate::util::{AppResponse, EndpointResult};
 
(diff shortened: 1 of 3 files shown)


pietroalbini:
Note that I do not have a local instance of crates.io up, so I've not been able to test if this actually work. Still, the code compile and should theoretically work. I'd love if someone could try it locally before deploying it.

jtgeibel:
I've tested locally by sleeping for 1.5 seconds in the "update_count" step.

[...] `SLOW REQUEST timing_update_count="1.510534517s" timing_get_version="2.976465ms" timing_get_conn="559.789µs"`.

We could reduce the precision of the ms and s timings, but this looks great to me!

@bors r+

pietroalbini:
Actually, let's reduce the precision.

@bors r- retry

pietroalbini:
Reduced the precision to two decimal digits.

r? @jtgeibel

jtgeibel:
Thanks!

@bors r+

PR #3403, opened 2021-03-12
GET /me/crate_owner_invitations: Add `users` list

Turbo87:
This will allow us to display avatars on the frontend for the `inviter` part of the invitations.

### DIFF
--- a/src/controllers/crate_owner_invitation.rs
+++ b/src/controllers/crate_owner_invitation.rs
@@ -41,7 +41,7 @@ pub fn list(req: &mut dyn RequestExt) -> EndpointResult {
                 .unwrap_or_default();
 
             let crate_name = invitation.crate_name(conn);
-            EncodableCrateOwnerInvitation::from_basic(invitation, inviter_name, crate_name)
+            EncodableCrateOwnerInvitation::from(invitation, inviter_name, crate_name)
         })
         .collect();
 
--- a/src/controllers/crate_owner_invitation.rs
+++ b/src/controllers/crate_owner_invitation.rs
@@ -4,8 +4,7 @@ use crate::models::{CrateOwner, CrateOwnerInvitation, OwnerKind, User};
 use crate::schema::{crate_owner_invitations, crate_owners, users};
 use crate::views::{EncodableCrateOwnerInvitation, EncodablePublicUser, InvitationResponse};
 use diesel::dsl::any;
-use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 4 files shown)


src/views.rs
```diff
@@ -84,10 +85,16 @@ pub struct EncodableCrateOwnerInvitation {
...
 
 impl EncodableCrateOwnerInvitation {
-    pub fn from(invitation: CrateOwnerInvitation, conn: &PgConnection) -> Self {
+    pub fn from(
+        invitation: CrateOwnerInvitation,
+        inviter_name: String,
+        crate_name: String,
+    ) -> Self {
         Self {
-            invited_by_username: invitation.invited_by_username(conn),
```
pietroalbini, line 89: If this was the only use of `invited_by_username` could we remove that method?

src/controllers/crate_owner_invitation.rs
```diff
@@ -1,28 +1,64 @@
...
+    // Turn `CrateOwnerInvitation` list into `EncodableCrateOwnerInvitation` list
     let crate_owner_invitations = crate_owner_invitations
         .into_iter()
-        .map(|i| EncodableCrateOwnerInvitation::from(i, conn))
+        .map(|invitation| {
+            let inviter_id = invitation.invited_by_user_id;
+            let inviter_name = users
+                .iter()
+                .find(|user| user.id == inviter_id)
+                .map(|user| user.gh_login.clone())
```
pietroalbini, line 40: This could result in bad performance if there are a lot of users and a lot of invitations. Could we convert the `Vec<User>` above into a `HashMap<i32, User>` to get fast lookups?

src/controllers/crate_owner_invitation.rs
```diff
@@ -1,28 +1,64 @@
...
+                .unwrap_or_default();
+
+            let crate_name = invitation.crate_name(conn);
```
pietroalbini, line 43: This is doing N+1 queries to fetch all the crate names. I know it already did that before this PR, but since we already fixed the N+1 for the users could you do the same for crates?

Once that's done we *should* be able to remove that method.

pietroalbini (changes requested):

Turbo87:
@pietroalbini thanks for the review. your comments should be addressed now.

pietroalbini (approved):

pietroalbini:
Thanks for the PR!

@bors r+

PR #3406, opened 2021-03-13
Remove some FIXME/TODOs

JohnTitor:


src/background_jobs.rs
```diff
@@ -28,8 +28,6 @@ pub struct Environment {
     http_client: AssertUnwindSafe<Client>,
 }
 
-// FIXME: AssertUnwindSafe should be `Clone`, this can be replaced with
-// `#[derive(Clone)]` if that is fixed in the standard lib
```
JohnTitor, line 32: Implementing `Clone` doesn't seem straightforward: https://github.com/rust-lang/rust/pull/81258
I think we could remove this for now.

Turbo87 (approved):

PR #3407, opened 2021-03-13
Display username instead if they don't have the "name" field

JohnTitor:
Let's take https://crates.io/crates/libc/owners as an example. It currently looks like:

image

The problem is that the name field doesn't display anything if the account name is empty (e.g. https://github.com/rust-lang-owner).
This PR fixes that behavior to show their username (`user.login`) instead.
This also fixes the alt attribute for the avatar image by just removing `null` (it still doesn't ideal but should be better).


Turbo87:
Good idea, thanks!

@bors r+

PR #3408, opened 2021-03-13
Extract VisibilityConfig from gen_scripts module

dtolnay:
Context: I will be sending a PR to topologically sort the tables inserted into the db-dump.tar.gz by [`DumpTarball::create`](https://github.com/rust-lang/crates.io/blob/c32fef25c58e53b007dd5795e39fae48800d9bfe/src/tasks/dump_db.rs#L144). That requires access to the same `dependencies` DAG described by dump-db.toml, but outside of the context of gen_scripts.

Rationale: stream processing of db-dump.tar.gz; pipelining the extraction and import.

### DIFF
--- a/src/tasks/dump_db/configuration.rs
+++ b/src/tasks/dump_db/configuration.rs
@@ -1,0 +1,108 @@
+use std::collections::{BTreeMap, VecDeque};
+
+/// An enum indicating whether a column is included in the database dumps.
+/// Public columns are included, private are not.
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub(super) enum ColumnVisibility {
+    Private,
+    Public,
+}
+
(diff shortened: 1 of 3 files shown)


src/tasks/dump_db/configuration.rs
```diff
@@ -0,0 +1,110 @@
+use std::collections::{BTreeMap, VecDeque};
+
+impl VisibilityConfig {
+    pub(super) fn get() -> Self {
+        toml::from_str(include_str!("dump-db.toml")).unwrap()
+    }
+}
```
JohnTitor, line 7: Could you move this to the below impl?

dtolnay: Good call, done.

JohnTitor (approved):

JohnTitor:
Thanks!
r? @JohnTitor @bors r+

dtolnay:
Follow-up PR to sort the tables in the db-dump tarball: #3409.

PR #3413, opened 2021-03-15
Reduce database writes for the download endpoint

pietroalbini:
crates.io receives a lot of download requests, and we're reaching a point where executing a write query to PostgreSQL for each request is causing operational issues. Over the past few weeks we've been getting regular pages due to download requests being slow to execute or outright fail for a couple of seconds, due to the database being overwhelmed.

This PR avoids that problem by collecting download counts in memory and periodically persisting them into the database, hopefully reducing the write load coming to our primary database thanks to the new `downloads_counter` module.

The implementation is designed to reduce the amount of locking needed as much as possible to prevent requests from being unnecessarily delayed. Instead of storing the counters in a `RwLock<HashMap<i32, AtomicUsize>>`, which would require locking the whole map when a new version needs to be tracked, this uses the [`dashmap`](https://crates.io/crates/dashmap) crate.

`DashMap` offers an API similar to the standard library's `HashMap`, but stores the items in `num_cpus()*4` individually locked `RwLock<HashMap<K, V>>` (called "shards") based on the key's hash. Splitting the items in shards allows to reduce how much of the `DashMap` is locked at the same time, with concurrent download requests locking each other only inside a single shard.

Persisting download counts also takes advantage of sharding: to avoid "stopping the world" when the background thread starts persisting the counters only one shard is persisted at the time. The current configuration persists all shards over the span of one minute, and that's configurable at runtime with the `DOWNLOADS_PERSIST_INTERVAL_MS` environment variable.

To avoid losing downloads the code also persists all the shards at the same time when the server gracefully shuts down. There is still the possibility of losing download counts if the server is killed without a chance of gracefully shutting down (for example due to platform issues), but the possibility of losing less a minute of download counts is far less important than reducing database load.

r? @jtgeibel

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ conduit-static = "0.9.0-alpha.3"
 
 cookie = { version = "0.15", features = ["secure"] }
 ctrlc = { version = "3.0", features = ["termination"] }
+dashmap = { version = "4.0.2", features = ["raw-api"] }
 derive_deref = "1.1.1"
 dialoguer = "0.7.1"
 diesel = { version = "1.4.0", features = ["postgres", "serde_json", "chrono", "r2d2"] }
--- a/src/controllers/version/downloads.rs
+++ b/src/controllers/version/downloads.rs
@@ -20,19 +20,32 @@ pub fn download(req: &mut dyn RequestExt) -> EndpointResult {
     let crate_name = &req.params()["crate_id"];
     let version = &req.params()["version"];
 
-    let (crate_name, was_counted) = increment_download_counts(req, recorder, crate_name, version)?;
+    let (version_id, crate_name): (_, String) = {
+        use self::versions::dsl::*;
+
+        let conn = recorder.record("get_conn", || req.db_conn())?;
+
+        // Returns the crate name as stored in the database, or an error if we could
+        // not load the version ID from the database.
+        recorder.record("get_version", || {
+            versions
+                .inner_join(crates::table)
+                .select((id, crates::name))
+                .filter(Crate::with_name(crate_name))
+                .filter(num.eq(version))
+                .first(&*conn)
+        })?
+    };
+
+    // The increment does not happen instantly, but it's deferred to be executed in a batch
+    // along with other downloads. See crate::downloads_counter for the implementation.
+    req.app().downloads_counter.increment(version_id);
 
     let redirect_url = req
         .app()
         .config
         .uploader
         .crate_location(&crate_name, version);
-
-    // Adding log metadata requires &mut access, so we have to defer this step until
-    // after the (immutable) query parameters are no longer used.
-    if !was_counted {
-        req.log_metadata("uncounted_dl", "true");
-    }
 
     if req.wants_json() {
         #[derive(Serialize)]
@@ -43,42 +56,6 @@ pub fn download(req: &mut dyn RequestExt) -> EndpointResult {
     } else {
         Ok(req.redirect(redirect_url))
     }
-}
-
-/// Increment the download counts for a given crate version.
-///
-/// Returns the crate name as stored in the database, or an error if we could
-/// not load the version ID from the database.
-///
-/// This ignores any errors that occur updating the download count. Failure is
-/// expected if the application is in read only mode, or for API-only mirrors.
-/// Even if failure occurs for unexpected reasons, we would rather have `cargo
-/// build` succeed and not count the download than break people's builds.
-fn increment_download_counts(
-    req: &dyn RequestExt,
-    recorder: TimingRecorder,
-    crate_name: &str,
-    version: &str,
-) -> AppResult<(String, bool)> {
-    use self::versions::dsl::*;
-
-    let conn = recorder.record("get_conn", || req.db_conn())?;
-
-    let (version_id, crate_name) = recorder.record("get_version", || {
-        versions
-            .inner_join(crates::table)
-            .select((id, crates::name))
-            .filter(Crate::with_name(crate_name))
-            .filter(num.eq(version))
-            .first(&*conn)
-    })?;
-
-    // Wrap in a transaction so we don't poison the outer transaction if this
-    // fails
-    let res = recorder.record("update_count", || {
-        conn.transaction(|| VersionDownload::create_or_increment(version_id, &conn))
-    });
-    Ok((crate_name, res.is_ok()))
 }
 
 /// Handles the `GET /crates/:crate_id/:version/downloads` route.
--- a/src/downloads_counter.rs
+++ b/src/downloads_counter.rs
@@ -1,0 +1,136 @@
+use crate::App;
+use anyhow::Error;
+use dashmap::{DashMap, SharedValue};
+use diesel::{pg::upsert::excluded, prelude::*};
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicI64, AtomicUsize, Ordering};
+
+/// crates.io receives a lot of download requests, and we can't execute a write query to the
+/// database during each connection for performance reasons. To reduce the write load, this struct
+/// collects the pending updates from the current process and writes in batch.
+///
+/// To avoid locking the whole data structure behind a RwLock, which could potentially delay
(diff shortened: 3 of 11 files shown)


pietroalbini:
Opened the PR early to allow the review to start ahead of time, but this PR is lacking tests (which we should add before deploying the PR).

Cargo.toml
```diff
@@ -86,6 +86,8 @@ tracing = "0.1"
 tracing-subscriber = "0.2"
 url = "2.1"
 
+dashmap = { version = "4.0.2", features = ["raw-api"] }
```
Turbo87, line 89: can you add it to the alphabetically ordered list above? 😉

src/downloads_counter.rs
```diff
@@ -0,0 +1,149 @@
...
+            counted_versions += stats.counted_versions;
+            pending_downloads = stats.pending_downloads;
+        }
+
+        println!(
+            "download_counter all_shards counted_versions={} counted_downloads={} pending_downloads={}",
+            counted_versions,
+            counted_downloads,
+            pending_downloads,
+        );
```
Turbo87, line 76: out of curiosity, is there a reason why we're using `println!()` everywhere instead of the `log` or `tracing` macros?

src/downloads_counter.rs
```diff
@@ -0,0 +1,149 @@
...
+///
+/// The disadvantage of this approach is that download counts are stored in memory until they're
+/// persisted, so it's possible to lose some of them if the process exits ungracefully. While
+/// that's far from ideal, the advantage of batching database updates far outweights potentially
+/// losing some download counts.
+#[derive(Debug)]
+pub struct DownloadsCounter {
+    inner: DashMap<i32, AtomicUsize>,
+    shard_idx: AtomicUsize,
+    pending_count: AtomicUsize,
```
Turbo87, line 29: it tooks me a while to figure out what the purpose of these two fields is. could you add a short doc comment that explains their purpose? 🙏

src/downloads_counter.rs
```diff
@@ -0,0 +1,149 @@
...
+/// DashMaps: to avoid locking all the download requests at the same time each iteration only
+/// persists a single shard at the time.
+///
+/// The disadvantage of this approach is that download counts are stored in memory until they're
+/// persisted, so it's possible to lose some of them if the process exits ungracefully. While
+/// that's far from ideal, the advantage of batching database updates far outweights potentially
+/// losing some download counts.
+#[derive(Debug)]
+pub struct DownloadsCounter {
+    inner: DashMap<i32, AtomicUsize>,
```
Turbo87, line 27: Is the `AtomicUsize` here required? Maybe I'm misunderstanding the `DashMap`. It sounded to me like the `DashMap` was already handling all of the thread-sync problems for us.

pietroalbini: Yes, it's required.

`DashMap` does not solve all the concurrency problems we could have. What it does is reduce the amount of locked items we have when acquiring a write lock by turning `RwLock<HashMap<K, V>>` (what we'd have without the crate) into a `Vec<RwLock<HashMap<K, V>>>`, with each vector item being a shard.

If we didn't use an atomic here we'd be forced to acquire a write lock over the shard whenever we needed to update a download count. Thanks to the atomic we can instead acquire a read lock to update a download count, having to acquire a write lock only when inserting a missing version into the map.

Turbo87: I see, thanks :)

src/controllers/version/downloads.rs
```diff
@@ -45,42 +58,6 @@ pub fn download(req: &mut dyn RequestExt) -> EndpointResult {
...
-/// Increment the download counts for a given crate version.
-///
-/// Returns the crate name as stored in the database, or an error if we could
-/// not load the version ID from the database.
-///
-/// This ignores any errors that occur updating the download count. Failure is
-/// expected if the application is in read only mode, or for API-only mirrors.
-/// Even if failure occurs for unexpected reasons, we would rather have `cargo
-/// build` succeed and not count the download than break people's builds.
-fn increment_download_counts(
```
jtgeibel, line 57: We may want to temporarily leave the current implementation available, behind an environment variable. That way, if we run into anything unexpected in production, we can easily switch back.

(We can always rollback a deploy, so this is more about making sure `master` is always deployable. If we do run into an issue, we can always add this back behind a flag, so I don't see this as a blocking concern.)

Cargo.toml
pietroalbini: Done!

src/downloads_counter.rs
pietroalbini: Done!

src/downloads_counter.rs
pietroalbini: I don't think there is really a reason, but I don't see value in switching to another library right now.

src/downloads_counter.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub(crate) fn increment(&self, version_id: i32) {
+        self.pending_count.fetch_add(1, Ordering::SeqCst);
+
+        if let Some(counter) = self.inner.get(&version_id) {
+            counter.value().fetch_add(1, Ordering::SeqCst);
+        } else {
+            self.inner
+                .entry(version_id)
+                .and_modify(|counter| {
+                    counter.fetch_add(1, Ordering::SeqCst);
```
jtgeibel, line 54: So if I understand correctly, this is very similar to the `if let ... inner.get()` above. However, `get()` returns an immutable reference, while `and_modify()` passes a `&mut` to the closure (which is more strict than we really need).

The "redundant" `fetch_add()` is needed here, in case we race with another thread which inserts the version between our calls to `get()` and `entry()`. And we use the `get()` branch above for the happy path where we don't need to lock the shard.

jtgeibel (commented):
Looks great to me overall

src/controllers/version/downloads.rs
pietroalbini: Yeah, for now I'd merge it to master and rollback the deploy + revert the PR in case something goes wrong.

src/controllers/version/downloads.rs
```diff
@@ -20,20 +20,33 @@ pub fn download(req: &mut dyn RequestExt) -> EndpointResult {
     let crate_name = &req.params()["crate_id"];
     let version = &req.params()["version"];
 
-    let (crate_name, was_counted) = increment_download_counts(req, recorder, crate_name, version)?;
+    let (version_id, crate_name): (_, String) = {
+        use self::versions::dsl::*;
+
+        let conn = recorder.record("get_conn", || req.db_conn())?;
```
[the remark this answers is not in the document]
pietroalbini: Yes, let's keep it on the primary replica for now. Another worry for moving this over to the read replica is that we already route the rest of the "heavy" traffic (like the crates endpoints) to it, and we could end up sending too much to it while keeping the primary idle.

src/downloads_counter.rs
pietroalbini: Yes, your understanding is correct! I'll add comments to write it down in the source code.

pietroalbini: Added some comments on the `increment` function.

pietroalbini:
The PR has been deployed to staging.crates.io and it seems to correctly persist download counts!

There are two things that are left to do:

* Add automated tests for `DownloadsCounter` itself: it's now generally tested by the download test, but it'd be better to add in-depth tests for it as it's going to be one of the system in the critical path for download requests.
* Ensure pending downloads are persisted when the application is rebooted: this works locally and the code should be implemented correctly, but a brief test on staging shows that it might not work on Heroku.

As we're getting a lot of pages about this I don't think it's critical we address those two issues before deploying to production, but we should make sure we land followup PRs for it.

jtgeibel:
Looks great to me! On my local setup, this improved the endpoint from 876 rps to 14,068 rps! (Though I don't have many versions, so this is mostly just testing atomics, and not any sort of contention on the `DashMap`.)

@bors r+

pietroalbini:
Opened #3415 and #3416 to handle the outstanding work on this.

PR #3405, opened 2021-03-12
Add crate descriptions to dependency lists

Turbo87:
This PR adds the crate descriptions to the "Dependencies" tab on the crate details page. Since the descriptions are async loaded by Ember Data I've also added a loading state, based on the new `Placeholder` component.

Bildschirmfoto 2021-03-12 um 19 31 58

PR #3426, opened 2021-03-18
krate.search: Add support for `?ids[]=foo&ids[]=bar` queries

Turbo87:
... to reduce unnecessary N+1 requests

This can be used in combination with e.g. #3405 to speed up the loading of crate descriptions.

### DIFF
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -168,6 +168,14 @@ pub fn search(req: &mut dyn RequestExt) -> EndpointResult {
                     .filter(follows::user_id.eq(user_id)),
             ),
         );
+    } else if params.get("ids[]").is_some() {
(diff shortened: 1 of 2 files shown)


jtgeibel:
We may want to put a limit on the number of crates. We limit pagination to 100 per page, but could consider a higher limit here. (This codebase has 60 direct dependencies, so 100+ is reasonable. Would it be easy to split >100 dependencies into multiple requests on the frontend?)

Turbo87:
Since this is using the same code paths as the regular search the default pagination options apply AFAICT.

splitting the request into multiple chunks on the frontend is not a problem.

jtgeibel:
Yeah, you're right. It works fine as-is.

@bors r+

Turbo87:
for future reference, the reason for this specific design was: https://github.com/emberjs/data/blob/v4.12.8/packages/adapter/src/rest.ts#L709

PR #3431, opened 2021-03-19
Add labels to forms missing them

lf-:


lf-:
I am utterly bewildered as to this CI failure. It seems to be failing because there are no errors? Can someone more familiar with the infrastructure help me with this?

(side note: I saw the warning about multiple landmark elements but was not sure how to fix it, and it was there before, so I'm leaving it to future work)

```
  -:-  error  Pending module (`app/components/settings/api-tokens`) passes all rules. Please remove `app/components/settings/api-tokens` from pending list.  invalid-pending-module

```

Turbo87:
@lf- I think you need to remove the files from https://github.com/rust-lang/crates.io/blob/master/.template-lintrc.js#L6

lf-:
> @lf- I think you need to remove the files from https://github.com/rust-lang/crates.io/blob/master/.template-lintrc.js#L6

Ahhh yup, that does it. Thanks!

Turbo87:
LGTM! thanks :)

I took the liberty to squash the commits together and rebase them. 

@bors r+

Issue #2204, opened 2020-02-20
Tracking issue for using hyper web server in production


jtgeibel:
There are several open issues that need to be addressed before we can enable hyper in production:

* [x] The default number of threads used in development mode needs to be increased to make the site usable locally (I tried 4 and it seemed to  work ok)
* [x] Using `thread::sleep` in a controller function appeared to be blocking other requests, and this wasn't the expected behavior when using `tokio::task::spawn_blocking`
* [x] Work in progress: Release some improvements to the `conduit` crates that allow for a better implementation in `conduit-hyper` when serving static files. conduit-rust/conduit@0deff5b4a3ba044258cc42c29f40cbb4186096b9
* [x] Fix sub-categories. `:` comes through as `%3A`, resulting in a 404. #2533
* [x] Switch to `USE_HYPER=1` on production!

Cleanup:

* [x] Switch to hyper by default, so that local development aligns with production.
* [ ] Eventually remove dependency on `civet`.

## Old

* ~I want to add a callback that gives the application a chance to handle a request async if the count of background threads is exceeded.  This would potentially let us handle a download redirect instead of rejecting those requests due to exceeding our thread count capacity.~ Update: Because the download endpoint needs at least a read-only database connection, this isn't a short-term goal.

jtgeibel:
@carols10cents I've just opened #2212 to address the first task.

For the 2nd task I've added a test upstream in jtgeibel/conduit-hyper@f7f53b8 to ensure this works and doesn't regress.  I was able to replicate your results locally when adding a delay to `/me` and `/me/updates`.  By then loading the `/dashboard` frontend URL I saw that the requests were issued serially such that a blank page is shown for 20 seconds until the user data is loaded from `/me` and then the dashboard data is requested causing another delay.

By trying the same experiment with `/me` and `/summary` and then loading `/`, the two requests are sent in parallel and they both complete after just a single duration.

Turbo87:
@jtgeibel @carols10cents to we have a timeline for the last part of this issue ("Eventually remove dependency on civet.")?

I assume that this would potentially also improve our compile times a little bit :)

Turbo87:
and in case it wasn't obvious: I'm happy to open a PR for this once we think it's time to say goodbye to `civet` 😉

jtgeibel:
> I assume that this would potentially also improve our compile times a little bit :)

[...]

PR #3436, opened 2021-03-19
Remove `civet` server option

Turbo87:
as discussed in our weekly team meeting today :)

Resolves https://github.com/rust-lang/crates.io/issues/2204


jtgeibel:
I added a commit dropping the `ctrlc` crate, which is unused now!

@bors r+

PR #3440, opened 2021-03-20
Update debug middleware

jtgeibel:
The first commit drops the ability to enable this middleware on production. With current traffic levels this isn't very practical (as concurrent requests can interleave their output), and we don't want to log headers that may contain authorization cookies or tokens.

The second commit silences the debug information by default during local development. It can be re-enabled by setting the `RUST_LOG` environment variable appropriately.

r? @Turbo87


Turbo87:
@bors r+

seems like a good idea :)

any thoughts on making the environment check a compile-time flag?

Issue #240, opened 2015-12-17
Error on network failure is incomprehensible


kjgorman:
Basically if you load the crates.io page but then there is some network interruption after that point, new ajax requests will begin to fail (fair enough).

When this happens an 'oops something went wrong' page is presented, including a stack trace. I guess for an open source project like this it is probably not an issue to include debug information to public users, but the error message seems like it could be more helpful.

This came up for me when the wifi at JFK started throttling after the 30 free minutes had been used up. Obviously just refreshing the page bounced me to the "buy more wifi" splash, but it seems like it wouldn't be impossible to present a more pleasant error when there's no network connectivity.

This is the error I received at the time:
![screen shot 2015-12-16 at 7 57 54 pm](https://cloud.githubusercontent.com/assets/1994486/11858498/3690466e-a430-11e5-90d8-58c3f7f973ca.png)

You can replicate (with a less insane stack trace) by using chrome network throttling to turn off network access once you've loaded the page.

tshepang:
I saw this today, also due to bad network.

jugglerchris:
You get something very similar going to a non-existant crate, for example https://crates.io/crates/adsfadf , but it does also say "Create adsfadf does not exist" and "e is undefined".  Not sure if that's a separate issue!

dtolnay:
I filed #844 for the non-existent crate issue. I think that one is a recent regression.

PR #3442, opened 2021-03-21
Show generic flash message if crate data fails to load 

Turbo87:
Rethrowing the error means that it ends up creating noise on Sentry even for regular network errors.

Resolves https://github.com/rust-lang/crates.io/issues/240


durka:
This undid the fix for #141 implemented in #421 and it's a very poor UX to redirect to the home page instead of showing a 404. The 404 is semantically correct and more importantly, leaves the URL alone for the user to correct a typo. Can we revert or reconsider the change?

PR #3444, opened 2021-03-21
Extract `sentry` service

Turbo87:
This makes it easier to mock the Sentry code during tests to ensure that whether we send certain types of errors to Sentry or not.


pichfl:
I wish this was provided by the sentry ember sdk out of the box.

@bors r+

PR #3446, opened 2021-03-21
Run daily database maintenance tasks

jtgeibel:
Run `VACUUM version_downloads;` daily. We could alternatively tweak the
auto-vacuum parameters on this table, but long term we should archive
daily download stats outside of the database and it will be easier to
pick the parameters then, once we have only 90 days of history rather
than growing daily without bound.

r? @Turbo87

### DIFF
--- a/src/bin/enqueue-job.rs
+++ b/src/bin/enqueue-job.rs
@@ -35,6 +35,7 @@ fn main() -> Result<()> {
                 .unwrap_or_else(|| String::from("db-dump.tar.gz"));
             Ok(tasks::dump_db(database_url, target_name).enqueue(&conn)?)
         }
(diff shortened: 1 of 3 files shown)


Turbo87 (approved):
LGTM 👍 

I assume we have something in the Heroku config that triggers this every day?

jtgeibel:
Yeah, I'll add an entry to the scheduler after deploying. I'll probably add it to staging too, for consistency, even though it really isn't needed there.

jtgeibel:
Merging to include in a deploy today.

@bors r=Turbo87

jtgeibel:
Scheduled to run daily at 1AM UTC in both staging and production.

PR #3464, opened 2021-03-25
Fixes #3448 Handle cargo-style semver in `/range`

### LINKED ISSUE #3448: Error: No matching version of crate 'byteorder' found for: >=1.3.0, <1.4.0
**Describe the bug**

https://crates.io/crates/bincode/1.3.2/dependencies shows a dependency to the `byteorder` crate with the semver range `>=1.3.0, <1.4.0`. When clicking on the link (https://crates.io/crates/byteorder/range/%3E=1.3.0,%20%3C1.4.0) an error notification pops up:

> No matching version of crate 'byteorder' found for: >=1.3.0, <1.4.0

but https://crates.io/crates/byteorder/versions shows several matching versions.

**Expected behavior**

I expect to end up at https://crates.io/crates/byteorder/1.3.4 instead of seeing the error popup.

**Additional context**

This is most likely caused by a difference between the Rust and JS semver implementations around dependency requirements/ranges. We might have to slightly adjust the `req` fields from the API to be able to use them with the `semver` JS package.

https://github.com/renovatebot/renovate/blob/ed683a777ece2e0ea51cc967a55f2d7e3158d1e7/lib/versioning/cargo/index.ts#L26-L30 could be a potential solution.

plaets:
Fixes #3448

It took this long because I started overthinking and went back and forth on some decisions. In the end, I decided to just replace all ',' with spaces... It should be enough for ranges that are valid for cargo (as of 0.53.0). I'm assuming that I don't need to handle the implicit caret (like renovate does) because it's inserted automatically somewhere in the backend.   
I also wanted to note that `/range/` will happily accept  ranges that are not supported by cargo (`||`, `-`, ...). Code in this commit does not handle these cases in any way. But then, cargo is currently using semver v0.10.0, meanwhile v0.11.0 implements `||`... 

For future reference:
* [Npm semver docs, section 'Advanced Range Syntax'](https://www.npmjs.com/package/semver)
* [Semver ranges in cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio)
* [Semver parsing in cargo starts here](https://github.com/rust-lang/cargo/tree/525c5b6239eba82a03ab560402cfb48bbc2b7794/src/cargo/core/dependency.rs#L101)

### DIFF
--- a/app/routes/crate/range.js
+++ b/app/routes/crate/range.js
@@ -3,6 +3,10 @@ import { inject as service } from '@ember/service';
 
 import maxSatisfying from 'semver/ranges/max-satisfying';
 
+function cargoRangeToNpm(range) {
+  return range.replace(',', ' ');
+}
+
 export default class VersionRoute extends Route {
   @service notifications;
   @service router;
@@ -14,8 +18,9 @@ export default class VersionRoute extends Route {
     let allVersionNums = versions.map(it => it.num);
     let unyankedVersionNums = versions.filter(it => !it.yanked).map(it => it.num);
 
+    let npmRange = cargoRangeToNpm(range);
     // find a version that matches the specified range
-    let versionNum = maxSatisfying(unyankedVersionNums, range) ?? maxSatisfying(allVersionNums, range);
+    let versionNum = maxSatisfying(unyankedVersionNums, npmRange) ?? maxSatisfying(allVersionNums, npmRange);
     if (!versionNum) {
       this.notifications.error(`No matching version of crate '${crate.name}' found for: ${range}`);
       this.replaceWith('crate.index');
--- a/tests/routes/crate/range-test.js
+++ b/tests/routes/crate/range-test.js
@@ -34,6 +34,20 @@ module('Route | crate.range', function (hooks) {
     assert.dom('[data-test-notification-message]').doesNotExist();
   });
 
+  test('happy path with cargo style and', async function (assert) {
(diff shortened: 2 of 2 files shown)


app/routes/crate/range.js
```diff
@@ -14,8 +18,9 @@ export default class VersionRoute extends Route {
     let allVersionNums = versions.map(it => it.num);
     let unyankedVersionNums = versions.filter(it => !it.yanked).map(it => it.num);
 
+    let npm_range = cargoRangeToNpm(range);
```
Turbo87, line 21:
```suggestion
    let npmRange = cargoRangeToNpm(range);
```

app/routes/crate/range.js
```diff
@@ -14,8 +18,9 @@ export default class VersionRoute extends Route {
...
     // find a version that matches the specified range
-    let versionNum = maxSatisfying(unyankedVersionNums, range) ?? maxSatisfying(allVersionNums, range);
+    let versionNum = maxSatisfying(unyankedVersionNums, npm_range) ?? maxSatisfying(allVersionNums, npm_range);
```
Turbo87, line 23:
```suggestion
    let versionNum = maxSatisfying(unyankedVersionNums, npmRange) ?? maxSatisfying(allVersionNums, npmRange);
```

Turbo87 (approved):
just a very minor nitpick, but other than that this looks great, thanks! :)

PR #3470, opened 2021-03-27
Add API Token Auth restrictions on GET APIs

jakeswenson:
This PR adds explicit denial of API token authentication on most GET APIs. 
Currently, CloudFront was never passing along the `Authorization` header for `GET` requests. 
(See rust-lang/simpleinfra#43 for more details.)

In addressing that issue it was pointed out that there may be GET APIs that the team might not want to commit to allowing API tokens access to.

This PR attempts to limit the exposed APIs to:
- `/api/v1/crates?following=1` (the crates "search" API, which is the only way to get the list of crates a user is following.)
- `/api/v1/me` 
  > Note: There were already [tests verifying token access to this API][me-tests], but with the CloudFront issue this API was not actually accessible.

[me-tests]: https://github.com/rust-lang/crates.io/blob/03f7b273eb7a024de8282ac2297daaa47df33fa6/src/tests/token.rs#L270

### DIFF
--- a/src/controllers/crate_owner_invitation.rs
+++ b/src/controllers/crate_owner_invitation.rs
@@ -9,7 +9,7 @@ use std::collections::HashMap;
 /// Handles the `GET /me/crate_owner_invitations` route.
 pub fn list(req: &mut dyn RequestExt) -> EndpointResult {
     // Ensure that the user is authenticated
-    let user = req.authenticate()?.user();
+    let user = req.authenticate()?.forbid_api_token_auth()?.user();
 
     // Load all pending invitations for the user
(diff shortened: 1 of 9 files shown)


jakeswenson:
r? @jtgeibel @pietroalbini 

Also, I would like to call out that since the tests do not match the structure of the code is was painful to find existing tests and where to add new tests to.

pietroalbini:
We discussed this in the weekly meeting, and agreed to allow authenticated requests on the two endpoints you allowed in this PR! Someone will review the actual implementation soon :)

pietroalbini:
Thanks for the PR!

@bors r+

PR #3483, opened 2021-04-02
Add support for testing emails

pietroalbini:
Before this PR there wasn't really a practical way to test what the content of an email is, which makes it harder for example to test that the token included in the ownership confirmation email works with the relevant APIs.

The main thing this PR does is refactoring the email handling code, moving from standalone functions to an `Emails` struct (an instance of it is stored in the `App`). This allows to configure which email backend is used at startup instead of automatically detecting it every time we happen to send an email address. It will also allow to persist other email-related information when we'll need to.

Thanks to the new email backends support this PR also adds a "memory" backend, used during tests. Instead of sending the email via SMTP or storing it into a file, the "memory" backend stores it in a `Mutex<Vec<_>>` which can be inspected by any test.

Finally, to test the new email testing capabilities this PR adds a test to ensure a user can accept an ownership invitation with the token included in the email.

r? @jtgeibel or @Turbo87

### DIFF
--- a/src/email.rs
+++ b/src/email.rs
@@ -1,4 +1,5 @@
-use std::path::Path;
+use std::path::PathBuf;
+use std::sync::Mutex;
 
 use crate::util::errors::{server_error, AppResult};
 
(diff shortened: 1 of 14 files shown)


src/models/krate.rs
```diff
@@ -347,11 +346,14 @@ impl Crate {
...
-                            &self.name.as_str(),
-                            &ownership_invitation.token.as_str(),
+                        // Swallow any error. Whether or not the email is sent, the invitation
+                        // entry will be created in the database and the user will see the
+                        // invitation when they visit https://crates.io/me/pending-invites/.
+                        let _ = app.emails.send_owner_invite(
+                            &email,
+                            &req_user.gh_login,
+                            &self.name,
+                            &ownership_invitation.token,
```
Turbo87, line 356: would it make sense for us to at least log such errors?

src/models/user.rs
```diff
@@ -97,7 +99,8 @@ impl<'a> NewUser<'a> {
                     .optional()?;
 
                 if let Some(token) = token {
-                    crate::email::send_user_confirm_email(user_email, &user.gh_login, &token);
+                    // Swallows any error. Some users might insert an invalid email address here.
+                    let _ = emails.send_user_confirm(user_email, &user.gh_login, &token);
```
Turbo87, line 103: from a layering perspective it feels a bit wrong to me to have email sending code in the database model struct, but I guess refactoring this is out of scope for this PR

src/email.rs
```diff
@@ -8,126 +9,175 @@ use lettre::transport::smtp::SmtpTransport;
...
+    Smtp {
+        server: String,
+        login: String,
+        password: String,
+    },
+    /// Backend used locally during development, will store the emails in the provided directory.
+    FileSystem { path: PathBuf },
+    /// Backend used during tests, will keep messages in memory to allow tests to retrieve them.
+    Memory { mails: Mutex<Vec<StoredEmail>> },
+}
```
Turbo87, line 154: what we're building here is essentially an interface with three backing implementations. I'm curious why you chose to implement that with an `enum`. would something like `Box<dyn EmailTrait>` work too?

Turbo87 (approved):
LGTM 👍 

just a few questions, but none of them are blocking

src/email.rs
pietroalbini: With few variants I personally prefer `enum`s rather than a `Box<dyn T>`, but either could work just fine.

Issue #3476, opened 2021-03-31
Different font for inline code and code blocks


tspiteri:
When viewing READMEs on crates.io, I see code in blocks in Consolas, but I see code inline in the text in my default browser monospace font. I think it would make sense to use the same font for code in the same page.

PR #3487, opened 2021-04-04
RenderedHtml: Use `--font-monospace` for code samples

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/3476

### DIFF
--- a/app/components/rendered-html.module.css
+++ b/app/components/rendered-html.module.css
@@ -25,6 +25,10 @@
         }
     }
 
+    code {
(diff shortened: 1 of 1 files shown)


pichfl:
Not sure if it's appropriate to ask for this here, but I do like the way GitHub and other platforms handle this, adding a slightly darker background and a little padding to the `<code>` inline elements. Would this be possible here? Otherwise this is fine for me.

Turbo87:
it's certainly possible, but out of scope for this PR :)

PR #3493, opened 2021-04-06
Add an expiration to owner invitations

### LINKED ISSUE #2869: Expiring of old invitations
It would be very useful if ownership invitations were to expire after some amount of time to limit issues around having unaccepted invitations sitting around long term (target account takeover, auditability, etc.).

pietroalbini:
This PR fixes #2869 by implementing a 30 days expiration for all ownership invitations.

Once an invite gets past its expiration date it will disappear from the [pending invites page](https://crates.io/me/pending-invites), and trying to accept it with the email link will display a message saying the invite expired and that the user needs to request a new one from the crate owner. Technically it's possible to decline an invite after it's expired through the API, but declining just deletes the row from the database so it's fine.

The new expiration rules will apply to existing invites too, and invites created more than 30 days before the deploy will instantly be treated as expired. The expiration check is entirely done on the application side, and the database schema is not changed: rolling back the PR will revert to the previous behavior of invites never expiring for invites generated both before and after the deploy.

Note that trying to send another invite to the same user before the expiration date will *not* extend the expiration date.

There are full tests for the whole functionality, and some refactors are included as part of this PR. Thanks @Turbo87 for authoring the first three commits handling a frontend bug when accepting expired invites via email!

r? @jtgeibel 
This PR is best reviewed commit by commit.

### DIFF
--- a/src/views.rs
+++ b/src/views.rs
@@ -2,6 +2,7 @@ use chrono::NaiveDateTime;
 use std::collections::HashMap;
 use url::Url;
 
+use crate::config::Config;
 use crate::github;
 use crate::models::{
     Badge, Category, Crate, CrateOwnerInvitation, CreatedApiToken, Dependency, DependencyKind,
--- a/src/views.rs
+++ b/src/views.rs
@@ -2,7 +2,6 @@ use chrono::NaiveDateTime;
 use std::collections::HashMap;
 use url::Url;
 
-use crate::config::Config;
 use crate::github;
 use crate::models::{
     Badge, Category, Crate, CrateOwnerInvitation, CreatedApiToken, Dependency, DependencyKind,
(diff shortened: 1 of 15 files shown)


src/views.rs
```diff
@@ -82,13 +83,16 @@ pub struct EncodableCrateOwnerInvitation {
...
+    #[serde(with = "rfc3339")]
+    pub expires_at: NaiveDateTime,
 }
 
 impl EncodableCrateOwnerInvitation {
     pub fn from(
         invitation: CrateOwnerInvitation,
         inviter_name: String,
         crate_name: String,
+        config: &Config,
```
Turbo87, line 95: do we need anything else from the `Config` in here? I'm wondering if just passing `expires_at: NaiveDateTime` would be sufficient.

pietroalbini: Addressed!

jtgeibel:
Looks great!

@bors r+

PR #3499, opened 2021-04-08
Make it possible to open the search using S

r00ster91:
Now it's more flexible and works even if you (possibly accidentally) have Caps Lock enabled. This already works for the Rust docs too so I thought it's more consistent this way.


Turbo87 (approved):
thanks! 👍

PR #3505, opened 2021-04-09
Show features of dependencies

bikeshedder:
I keep getting asked why my crate depends on `tokio` besides it just needing a small subset of features: https://github.com/bikeshedder/deadpool/issues/88

This PR adds the list of features used from each dependency to the dependencies tab:

image

### DIFF
--- a/app/components/dependency-list/row.hbs
+++ b/app/components/dependency-list/row.hbs
@@ -35,6 +35,17 @@
       </span>
     </div>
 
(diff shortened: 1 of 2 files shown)


Turbo87:
Thanks for the initial work @bikeshedder! I've slightly improved the styling of the feature listing, and I think this is good to go now. :shipit: 

Bildschirmfoto 2021-08-31 um 23 12 13

Bildschirmfoto 2021-08-31 um 23 12 31

PR #3507, opened 2021-04-09
Add `webp` to possible media urls

### LINKED ISSUE #3500: Add `.webp` as a supported media URL
**Is your feature request related to a problem? Please describe.**
Wanted to add an example of animated webp file in a crate README, but could not because `.webp` is not part of supported media urls. 

Had to convert to a gif. See https://crates.io/crates/webp-animation

**Describe the solution you'd like**
Add webp to `fn is_media_url` supported array at https://github.com/rust-lang/crates.io/blob/7844bf4c543859f351fd7d82d5a971c49e80ccec/src/render.rs#L174

**Describe alternatives you've considered**
Converted to the gif

**Additional context**
Here's an earlier version of the published crate (with webp file in README), which is nonfunctional: https://crates.io/crates/webp-animation/0.1.0

[WebP](https://en.wikipedia.org/wiki/WebP) is a newer more compact image format. For browser support, see https://www.caniuse.com/webp (Safari was not supporting webp for a long time earlier, but they have nowadays support as well from version 14. See https://stackoverflow.com/a/62594498)

blaind:
As per request, fixes #3500

(not tested locally)

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -171,7 +171,7 @@ fn is_media_url(url: &str) -> MediaUrl {
                     is_media: true,
                     add_sanitize_query: true,
                 },
-                "png" | "jpg" | "jpeg" | "gif" | "mp4" | "webm" | "ogg" => MediaUrl {
+                "png" | "jpg" | "jpeg" | "gif" | "mp4" | "webm" | "ogg" | "webp" => MediaUrl {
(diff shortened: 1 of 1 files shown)


blaind:
No problem, thank you for merging!

PR #3510, opened 2021-04-10
Allow using SecureToken in Diesel queries

pietroalbini:
This commit implements the FromSql and ToSql traits for SecureToken, allowing it to be used as a model field and in queries. This commit also hides the inner SHA256 value to prevent the rest of the application from accidentally leaking it.

The motivation of this commit is to enforce even more good practices around tokens through the type system.


Turbo87:
LGTM 👍 

@bors r+

PR #3514, opened 2021-04-12
Log a flag for downloads with non-canonical crate names

jtgeibel:
We could instead include the corrected name, but that doesn't seem very
useful and this is already something I search the logs for occasionally.

This change will allow us to collect some data to see if it would be
practical to blindly redirect the request if there is an issue obtaining
a database connection from the pool.

r? @pietroalbini

### DIFF
--- a/src/controllers/version/downloads.rs
+++ b/src/controllers/version/downloads.rs
@@ -20,7 +20,7 @@ pub fn download(req: &mut dyn RequestExt) -> EndpointResult {
     let crate_name = &req.params()["crate_id"];
     let version = &req.params()["version"];
 
-    let (version_id, crate_name): (_, String) = {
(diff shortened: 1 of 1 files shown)


src/controllers/version/downloads.rs
```diff
@@ -45,7 +45,11 @@ pub fn download(req: &mut dyn RequestExt) -> EndpointResult {
         .app()
         .config
         .uploader
-        .crate_location(&crate_name, version);
+        .crate_location(&canonical_crate_name, version);
+
+    if &canonical_crate_name != crate_name {
+        req.log_metadata("bot", "suspected");
```
pietroalbini, line 51: Can we log something different to be able to track this individually?

PR #3518, opened 2021-04-12
Forbid token auth on `/me` endpoint

jtgeibel:
This prohibits access to the `/api/v1/me` endpoint via an API token. I'd like to merge and deploy this before moving forward on rust-lang/simpleinfra#43, which would expose this endpoint. (Previous lockdown work was done in #3470.)

The second commit ensure that the user is only authenticated on the search endpoint when querying for followed crates.

r? @pietroalbini


Turbo87:
Are we sure that cargo is not using `/me`?

PR #3531, opened 2021-04-17
Add support for Prometheus metrics

pietroalbini:
This PR adds support for collecting Prometheus metrics in crates.io, and adds some basic metrics to make sure everything works correctly. We'll want to add way more metrics after the PR is merged.

Prometheus is the monitoring service used for most of the Rust infrastructure, and the infra team maintains an instance of it. Prometheus periodically scrapes an endpoint of the application and ingests all the metrics contained in it. It then provides a query interface to visualize metrics over time, and supports generating alerts based on the current metrics.

An example query to detect when there are more than 100 crates published in an hour would be:

```
increase(cratesio_service_versions_total[1h])
```

## Service-level metrics

Service-level metrics are available at `/api/private/metrics/service` and include all the things that are related to the service as a whole, instead of metrics specific to the crates.io backend instance. Example of these metrics could be the number of published crates ever, or how many jobs are present in the background queue.

These metrics will be scraped by hitting `https://crates.io`, so it's likely that two requests will be served by different backend servers due to the load balancer. This means service-level metrics must never include any information specific to a single instance. 

Current output of that endpoint:

```bash
# HELP cratesio_service_crates_total  Number of crates ever published
# TYPE cratesio_service_crates_total gauge
cratesio_service_crates_total 2
# HELP cratesio_service_versions_total  Number of versions ever published
# TYPE cratesio_service_versions_total gauge
cratesio_service_versions_total 4
```

## Instance-level metrics

> **Note:** we can't collect instance-level metrics right now due to the Heroku load balancer. The infra team is looking into implementing a workaround, see [this document](https://paper.dropbox.com/doc/crates.io-monitoring--BJC8_vas_Jkav2St9QrFf__aAg-JWf5AxfJ1Nbc3lLuNcaTy).

Instance-level metrics are available at `/api/private/metrics/instance` and include all the things that are specific to a single crates.io backend instance. Example of these metrics could be the state of the database pool, how many requests were processed, what the response time was, or how many downlaods are not counted yet.

These metrics will be scraped by hitting each individual backend at the same time, and Prometheus will then aggregate the results to offer a complete picture in the dashboard. Metrics that represent the state of the whole system shouldn't be implemented as instance-level metrics, as otherwise the resulting metric will be aggregated from each backend.

Current output of that endpoint:

```bash
# HELP cratesio_instance_database_idle_conns  Number of idle database connections in the pool
# TYPE cratesio_instance_database_idle_conns gauge
cratesio_instance_database_idle_conns{pool="follower"} 3
cratesio_instance_database_idle_conns{pool="primary"} 3
# HELP cratesio_instance_database_used_conns  Number of used database connections in the pool
# TYPE cratesio_instance_database_used_conns gauge
cratesio_instance_database_used_conns{pool="follower"} 0
cratesio_instance_database_used_conns{pool="primary"} 0
# HELP cratesio_instance_requests_in_flight  Number of requests currently being processed
# TYPE cratesio_instance_requests_in_flight gauge
cratesio_instance_requests_in_flight 1
# HELP cratesio_instance_requests_total  Number of requests processed by this instance
# TYPE cratesio_instance_requests_total counter
cratesio_instance_requests_total 8
```

## Authentication

To prevent third parties from scraping our metrics (which in the future could contain sensitive data), all the metrics endpoints are protected with HTTP authentication. Requests without an `Authorization` header matching the contents of the `METRICS_AUTHORIZATION_TOKEN` environment variable will be rejected. If the environment variable is missing metrics collection will be disabled, to prevent accidental leaks.

r? @jtgeibel

### DIFF
--- a/src/metrics/instance.rs
+++ b/src/metrics/instance.rs
@@ -1,0 +1,64 @@
+//! This module defines all the instance-level metrics of crates.io.
+//!
+//! Instance-level metrics are collected separately for each instance of the crates.io application,
+//! and are then aggregated at the Prometheus level. They're not suited for service-level metrics
+//! (like "how many users are there").
+//!
+//! There are two ways to update instance-level metrics:
+//!
+//! * Continuously as things happen in the instance: every time something worth recording happens
+//!   the application updates the value of the metrics, accessing the metric through
+//!   `req.app().instance_metrics.$metric_name`.
+//!
+//! * When metrics are scraped by Prometheus: every `N` seconds Prometheus sends a request to the
+//!   instance asking what the value of the metrics are, and you can update metrics when that
+//!   happens by calculating them in the `gather` method.
+//!
+//! As a rule of thumb, if the metric requires a database query to be updated it's probably a
+//! service-level metric, and you should add it to `src/metrics/service.rs` instead.
+
+use crate::util::errors::AppResult;
+use crate::{app::App, db::DieselPool};
+use prometheus::{proto::MetricFamily, IntCounter, IntGauge, IntGaugeVec};
+
+metrics! {
+    pub struct InstanceMetrics {
+        /// Number of idle database connections in the pool
+        database_idle_conns: IntGaugeVec["pool"],
+        /// Number of used database connections in the pool
+        database_used_conns: IntGaugeVec["pool"],
+
+        /// Number of requests processed by this instance
+        pub requests_total: IntCounter,
+        /// Number of requests currently being processed
+        pub requests_in_flight: IntGauge,
+    }
+
+    // All instance metrics will be prefixed with this namespace.
+    namespace: "cratesio_instance",
+}
+
+impl InstanceMetrics {
+    pub(crate) fn gather(&self, app: &App) -> AppResult<Vec<MetricFamily>> {
+        // Database pool stats
+        self.refresh_pool_stats("primary", &app.primary_database)?;
+        if let Some(follower) = &app.read_only_replica_database {
+            self.refresh_pool_stats("follower", follower)?;
+        }
+
+        Ok(self.registry.gather())
+    }
+
+    fn refresh_pool_stats(&self, name: &str, pool: &DieselPool) -> AppResult<()> {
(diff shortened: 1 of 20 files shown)


src/metrics/instance.rs
```diff
@@ -0,0 +1,62 @@
...
+//! and are then aggregated at the Prometheus level. They're not suited for service-level metrics
+//! (like "how many users are there").
+//!
+//! There are two ways to update instance-level metrics:
+//!
+//! * Continuously as things happen in the instance: every time something worth recording happens
+//!   the application updates the value of the metrics, accessing the metric through
+//!   `req.app().instance_metrics.$metric_name`.
+//!
+//! * When metrics are scraped by Prometheus: every tot seconds Prometheus sends a request to the
```
jtgeibel, line 13: I'm not sure what "tot seconds" is in reference to.

src/middleware.rs
```diff
@@ -68,6 +67,11 @@ pub fn build_middleware(app: Arc<App>, endpoints: RouteBuilder) -> MiddlewareBui
     m.add(AppMiddleware::new(app));
     m.add(KnownErrorToJson);
 
+    if env != Env::Test {
+        // This is added *after* AppMiddleware to make sure the app is available.
+        m.add(log_request::LogRequests::default());
```
jtgeibel, line 72: Moving this line will change what we log. In particular, I'm worried that it will log requests before normalizing well-known errors into JSON. It would also log paths after normalization, instead of the original client request.

jtgeibel (commented):
Looks great overall. Just a few comments.

src/metrics/instance.rs
pietroalbini: It just means "every `N` seconds". Did I accidentally use an Italian-ism here? :sweat_smile:

pietroalbini:

pietroalbini:
Uh, I forgot to mention that I added the `cratesio_instance_requests_in_flight` as well.

jtgeibel:
LGTM!

@bors r+

PR #3544, opened 2021-04-23
More aggressively rate limit reserving new package names

jtgeibel:
Larger burst limits can be applied on a per-user basis if they contact
us ahead of time.

r? @pietroalbini


pietroalbini:
I would move the burst to an environment variable so we can tweak it without a new deploy. Maybe set it in `Config`?

jtgeibel:
Both values are now configurable.

PR #3549, opened 2021-04-24
Remove version_authors and clean up old data

### LINKED ISSUE #83227: Tracking Issue for RFC 3052: make the authors field optional
This is a tracking issue for the RFC "Make the authors field optional" (rust-lang/rfcs#3052).

### About tracking issues

Tracking issues are used to record the overall progress of implementation. They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions. A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature. Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.

### Steps

- [ ] Implement the RFC
  - [x] cargo: stop including authorship information in `cargo new`/`cargo init` and in templates
  - [x] crates.io: start accepting crates without authorship information
  - [ ] crates.io: stop displaying and serving authorship information
  - [x] docs.rs: stop displaying and serving authorship information

### Implementation history

* rust-lang/crates.io#3429 - start accepting crates without authorship information
* rust-lang/crates.io#3132 - Remove "Authors" section from crate details page
* rust-lang/cargo#9282 - Stop including authors field in manifests made by cargo new
* rust-lang/docs.rs#1322 - stop displaying and serving authorship information
* rust-lang/docs.rs#1323 -  stop collecting author data and delete the old data

0xPoe:
close https://github.com/rust-lang/rust/issues/83227


hi-rustin:
r? @pietroalbini

Can you help with the patch file issue? I'm actually tempted to delete the patch file as it doesn't seem to be helping us?

Because since we don't update it every time, it must be expired? Or do I still have a misunderstanding of how patch file works?

pietroalbini:
Pushed a commit on your branch fixing the patch!

> Because since we don't update it every time, it must be expired? Or do I still have a misunderstanding of how patch file works?

The patching tool is quite smart, and in most cases it will adapt to other `schema.rs` changes without the need to tweak the patch. Let's take for example the chunk of the patch that was giving you errors:

```diff
@@ -1016,6 +1031,7 @@
 joinable!(publish_limit_buckets -> users (user_id));
 joinable!(publish_rate_overrides -> users (user_id));
 joinable!(readme_renderings -> versions (version_id));
+joinable!(recent_crate_downloads -> crates (crate_id));
 joinable!(version_authors -> versions (version_id));
 joinable!(version_downloads -> versions (version_id));
 joinable!(version_owner_actions -> api_tokens (api_token_id));
```

The patching tool will apply this patch (adding the highlighted line) *relative* to the three lines above and below it. As long as those three lines above and below are not changed the tool will be able to apply the patch even if the rest of the file completely changes.

The problem here is that one of the lines below *did* change, as you removed the `version_authors` table. By doing so you removed the hit the patching tool needed to figure out where to apply the change.

To fix this issue, in my commit I had to remove the `version_authors` line from the patch. That was not enough though, as now there were only two lines below instead of three. After adding the following line in `schema.rs` the problem was fixed:

```diff
@@ -1016,6 +1031,7 @@
 joinable!(publish_limit_buckets -> users (user_id));
 joinable!(publish_rate_overrides -> users (user_id));
 joinable!(readme_renderings -> versions (version_id));
+joinable!(recent_crate_downloads -> crates (crate_id));
 joinable!(version_downloads -> versions (version_id));
 joinable!(version_owner_actions -> api_tokens (api_token_id));
 joinable!(version_owner_actions -> users (user_id));
```

If you have more questions around this just ask!

migrations/2021-04-24-115720_remove_version_authors/down.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE version_authors(
+    id SERIAL  NOT NULL CONSTRAINT version_authors_pkey PRIMARY KEY,
+    version_id INTEGER NOT NULL CONSTRAINT fk_version_authors_version_id REFERENCES versions ON DELETE CASCADED,
```
pietroalbini, line 3: There is an error in this query:

```suggestion
    version_id INTEGER NOT NULL CONSTRAINT fk_version_authors_version_id REFERENCES versions ON DELETE CASCADE,
```

emilyalbini (commented):
Leaving the rest of the review back to Carol. r? @carols10cents

hi-rustin:
@pietroalbini It's working well. Thanks!

pietroalbini:
r? @pietroalbini 

Thank you so much!
@bors r+

bclairw-commits (commented):
This idea builds from backend so we can squeeze in the rest as I move forward with that

PR #3555, opened 2021-04-25
Add `DB_OFFLINE` as a failsafe during outages

jtgeibel:
If set to `leader` then the follower credentials are used in read-only
mode for all connections. The background-worker will log a message every
minute as a reminder to scale down its instance until the leader is
restored. To operate the server in this mode, a value must be provided
for `READ_ONLY_REPLICA_URL`.

If set to `follower` then no follower connection pool is initialized, as
if `READ_ONLY_REPLICA_URL` was not set.

r? @pietroalbini

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -103,11 +103,9 @@ impl App {
             _ => 30,
         };
 
-        // Determine if the primary pool is also read-only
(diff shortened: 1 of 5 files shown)


pietroalbini:
@bors r+

Thanks!

Issue #3541, opened 2021-04-22
Ensure downloads work without a database connection


pietroalbini:
We've had a couple database-related outages in the past few weeks, all of which impacted the download endpoint of crates.io. That endpoint is the most important one in the application, as any error returned by it results in end-user or CI builds failing. To improve crates.io's resilience we need to make sure the endpoint works even without a database connection.

The two uses of a connection during a download are counting the downloads and canonicalizing the crate name, to ensure the user is redirect to the CDN path with the right capitalization and hyphenation. Cargo always provides the canonical name to the API though, so *only during outages* we can do an unconditional redirects without affecting Cargo downloads.

Third-party clients that don't send the canonical name might get redirected to an invalid CDN path, but 0.00039% of the download requests we served in the past 48 hours would receive a wrong response (and none of those are Cargo builds). That's better than having downloads be down for everyone until the outage is fixed.

Needed steps:

* [x] Move stuff that requires a database connection during startup to the release phase #3556 #3562
* [x] Change the downloads endpoint to do an unconditional redirect when the database is broken #3564
* [ ] Change the startup not to require a database

PR #3556, opened 2021-04-26
Create `crates-admin migrate` to sync categories and migrate the db

pietroalbini:
The new command will be used in production to run migrations instead of installing the Diesel CLI and running migrations through it. In addition to that, synchronizing categories has been moved to the `crates-admin migrate` command instead of being executed every time the application starts.

The main advantage of this is, booting the server will not *require* a fully working database connection anymore. A connection is still needed, but this removes all the blockers for removing that limit.

r? @jtgeibel 
Part of #3541

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,4 +1,3 @@
 https://github.com/emk/heroku-buildpack-rust#cfa0f06
 https://github.com/heroku/heroku-buildpack-nodejs#v176
 https://github.com/heroku/heroku-buildpack-nginx#53b03b0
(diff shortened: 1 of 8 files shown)


Turbo87:
wouldn't it be better to make this a subcommand of the crates-admin binary to avoid the additional linking step?

pietroalbini:
Updated the PR to use a `crates-admin` subcommand.

Turbo87 (approved):
LGTM, feel free to r=me

jtgeibel:
LGTM too!

@bors r=Turbo87

PR #3560, opened 2021-04-26
Avoid cloning `Config` struct

jtgeibel:
This struct contains a lot of owned data and cloning the entire struct
is unnecessary.

r? @Turbo87


Turbo87 (approved):
very good idea! 👍

jtgeibel:
I've pushed a new commit which should fix the build error. I missed a line while trying to slip a diff into smaller PRs.

PR #3561, opened 2021-04-26
Upstream current production environment variables

jtgeibel:
This updates some defaults to match what we currently run in production.

I don't expect we'll merge this as-is, so draft for now.

r? @ghost

### DIFF
--- a/src/bin/background-worker.rs
+++ b/src/bin/background-worker.rs
@@ -38,7 +38,7 @@ fn main() {
     let db_url = db::connection_url(&config.db_primary_config.url);
 
     let job_start_timeout = dotenv::var("BACKGROUND_JOB_TIMEOUT")
-        .unwrap_or_else(|_| "30".into())
+        .unwrap_or_else(|_| "300".into())
         .parse()
         .expect("Invalid value for `BACKGROUND_JOB_TIMEOUT`");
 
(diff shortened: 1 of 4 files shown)


src/bin/background-worker.rs
```diff
@@ -38,7 +38,7 @@ fn main() {
     let db_url = db::connection_url(&config.db_primary_config.url);
 
     let job_start_timeout = dotenv::var("BACKGROUND_JOB_TIMEOUT")
-        .unwrap_or_else(|_| "30".into())
+        .unwrap_or_else(|_| "300".into())
```
jtgeibel, line 41: I think the reason this is currently so long is that it can take the background_worker several minutes to boot while it clones the index. We don't want to page someone just because a job is stuck in the queue while the worker reboots.

src/bin/monitor.rs
```diff
@@ -39,7 +39,7 @@ fn check_failing_background_jobs(conn: &PgConnection) -> Result<()> {
     // Max job execution time in minutes
     let max_job_time = dotenv::var("MAX_JOB_TIME")
         .map(|s| s.parse::<i32>().unwrap())
-        .unwrap_or(15);
+        .unwrap_or(60);
```
jtgeibel, line 42: This check does not cover the `update_downloads` job, which is allowed to run for much longer. We should consider reducing this back to 15.

pietroalbini:
Hmm, I'm not sure about this PR, and actually I was thinking of opening a PR that does the opposite :sweat_smile: 

We can discuss it more during this week's Zoom call, but I think we should stop setting the defaults based on which environment we're in. If we have different values in productions we should just have those different values set through the production vars.

PR #3562, opened 2021-04-26
Ensure the release succeeds when the primary database is read-only

jtgeibel:
If the primary is in read-only mode then the leader database in
`DATABASE_URL` may be unavailable. To avoid blocking configuration
changes, the command exits immediately claiming success.

r? @pietroalbini

### DIFF
--- a/src/admin/migrate.rs
+++ b/src/admin/migrate.rs
@@ -4,11 +4,33 @@ static CATEGORIES_TOML: &'static str = include_str!("../boot/categories.toml");
 diesel_migrations::embed_migrations!("./migrations");
 
 #[derive(clap::Clap, Debug, Copy, Clone)]
-#[clap(name = "migrate", about = "Migrate the database.")]
+#[clap(
+    name = "migrate",
+    about = "Verify config, migrate the database, and other release tasks."
+)]
 pub struct Opts;
 
 pub fn run(_opts: Opts) -> Result<(), Error> {
+    let config = crate::Config::default();
+
+    // TODO: Refactor logic so that we can also check things from App::new() here.
+    // If the app will panic due to bad configuration, it is better to error in the release phase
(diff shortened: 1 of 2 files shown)


src/admin/migrate.rs
```diff
@@ -8,7 +8,22 @@ diesel_migrations::embed_migrations!("./migrations");
...
 pub fn run(_opts: Opts) -> Result<(), Error> {
+    let config = crate::Config::default();
+
+    if config.db_primary_config.read_only_mode {
+        // The service is undergoing maintenance or mitigating an outage.
+        // Exit early to ensure we don't accidentally block the release phase.
+        //
+        // Heroku will update configuration variables even if the release command fails.
+        // This will prevent deploys while in this mode, even if there are no new migrations.
+        // https://devcenter.heroku.com/articles/release-phase#release-command-failure
```
jtgeibel, line 19: I tried this on a personal dyno and the web dyno did *not* restart when changing the environment if the release phase fails. Are the docs wrong? Maybe we need to always return success here?

pietroalbini: My interpretation of that page is that the *release* will fail (so the app won't be restarted), but the configuration changes you made in the UI will not be reverted to the original values.

pietroalbini:
Hmm, I would probably change this to straight up skip everything if we're in read-only mode.

jtgeibel:
> My interpretation of that page is that the *release* will fail (so the app won't be restarted), but the configuration changes you made in the UI will not be reverted to the original values.

Yes, that seems like the correct interpretation. I've updated the PR to return success in this situation, and have added some TODO notes regarding future improvements.

pietroalbini:
@bors r+

Thanks! Could you open issues to track the TODOs you added?

PR #3566, opened 2021-04-28
Run the Rust buildpack last

jtgeibel:
We currently use heroku-buildpack-multi which is no longer actively
maintained. Heroku supports multiple buildpacks natively and it makes
sense to follow their recommendation even if our config is a bit
different. (Heroku is currently detecting the framework as
`nginx-buildpack`.)

https://devcenter.heroku.com/articles/using-multiple-buildpacks-for-an-app#adding-a-buildpack

> The buildpack for the primary language of your app should always be
> the last buildpack in the list. This ensures that defaults for that
> primary language are applied instead of those for another language,
> and allows Heroku to correctly detect the primary language of your app.

r? @pietroalbini

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,3 +1,3 @@
-https://github.com/emk/heroku-buildpack-rust#cfa0f06
 https://github.com/heroku/heroku-buildpack-nodejs#v176
 https://github.com/heroku/heroku-buildpack-nginx#53b03b0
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):
I assumed you tried this or on staging already, right? :)

pietroalbini:
I don't think this change matters much, as Rust doesn't really have native support inside Heroku. Still, this change shouldn't hurt.

PR #3572, opened 2021-04-29
Allow the application to start without a database connection

### LINKED ISSUE #3541: Ensure downloads work without a database connection

pietroalbini:
This PR implements the last bit of #3541, making sure the server binary can start without a database connection.

To avoid serving the initial requests without a connection, the server will now wait for 5 seconds at startup to initialize the database connections, and if it can't do that in that timeframe it will treat the database as unhealthy and continue starting up the server.

This is critical to our continued availability, so to make sure this doesn't regress I also implemented a test that fails if a database connection becomes a requirement to start the server again. The test is implemented by starting the server binary with the right environment variables and doing proper HTTP requests to it.

Fixes #3541 
r? @jtgeibel

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -121,7 +121,7 @@ impl App {
                 .connection_customizer(Box::new(primary_db_connection_config))
                 .thread_pool(thread_pool.clone());
 
-            DieselPool::new(&config.db_primary_config.url, primary_db_config)
+            DieselPool::new(&config.db_primary_config.url, primary_db_config).unwrap()
         };
 
         let replica_database = if let Some(url) = config.db_replica_config.as_ref().map(|c| &c.url)
@@ -141,7 +141,7 @@ impl App {
                     .connection_customizer(Box::new(replica_db_connection_config))
                     .thread_pool(thread_pool);
 
-                Some(DieselPool::new(&url, replica_db_config))
+                Some(DieselPool::new(&url, replica_db_config).unwrap())
             }
         } else {
             None
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -97,6 +97,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let mut sig_int = rt.block_on(async { signal(SignalKind::interrupt()) })?;
     let mut sig_term = rt.block_on(async { signal(SignalKind::terminate()) })?;
 
+    // When the user configures PORT=0 the operative system will allocate a random unused port.
+    // This fetches that random port and uses it to display the "listening on port" message later.
+    let actual_port = server.local_addr().port();
+
     let server = server.with_graceful_shutdown(async move {
         // Wait for either signal
         futures_util::select! {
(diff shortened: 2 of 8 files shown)


pietroalbini:
Need to investigate the tests failing later, but the PR is still ready for review. My guess is that CI doesn't have the right environment variables configured.

jtgeibel:
Looks great to me! The tests passed for me locally, but if I comment out either `GH_CLIENT_ID` or `GH_CLIENT_SECRET` in my `.env` then the server fails to start. Interestingly, these can be set to an empty string and the tests will pass.

Eventually we could probably hook GitHub and S3 requests into our existing HTTP recording code, but for now its fine if we just document that these tests should not hit endpoints that could trigger external HTTP requests.

Turbo87:
> Eventually we could probably hook GitHub and S3 requests into our existing HTTP recording code

I think GitHub and S3 are exactly what is already hooked up, or am I missing something? 🤔

jtgeibel:
> I think GitHub and S3 are exactly what is already hooked up, or am I missing something?

The 2 new `ServerBin` tests actually boot the server, configured via its environment variables. Our existing tests link to the library crate, and configure an `App` instances for each test. In particular, those tests configure the S3 uploader (`api_protocol`) and GitHub API (`gh_base_url`) to use plaintext HTTP to a per-test proxy. (As far as I can tell, we don't plaintext proxy GitHub OAuth requests in our current tests, because the actual navigation to GitHub happens client side in the frontend.)

My understanding is that for now, if we tried to hit say the crate publish endpoint in a `ServerBin` test, then the server binary would try to make HTTPS requests to S3 and GitHub using whatever credentials it has available (possibly picking them up from a local `.env` file). We can't leak test requests out into the wild, but I also don't want to build a configuration switch into release binaries that effectively switches all our outgoing requests to plaintext.

We should probably add an environment variable that disables all outgoing requests, returning an error instead. But I don't think it is necessary to block this PR on that.

pietroalbini:
Resolved the conflict and (hopefully!) fixed CI.

jtgeibel:
LGTM!! @bors r+

PR #3583, opened 2021-05-04
Update comrak to 0.10.1

kivikakk:
An XSS vulnerability has been reported in comrak < 0.10.1 (including 0.10.1-rc.1); see https://github.com/kivikakk/comrak/issues/187.

crates.io is **unaffected** due to our use of `ammonia` (which is best practice :+1:), but it's probably not ideal to be using a version with a known bug either way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -402,9 +402,9 @@ dependencies = [
 
 [[package]]
 name = "comrak"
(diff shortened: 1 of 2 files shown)


Turbo87 (approved):
thank you! 👍

PR #3592, opened 2021-05-06
Add a background job for squashing the index

jtgeibel:
This adds a background job that squashes the index into a single commit.
The current plan is to manually enqueue this job on a 6 week schedule,
roughly aligning with new `rustc` releases. Before deploying this, will
need to make sure that the SSH key is allowed to do a force push to the
protected master branch.

This job is derived from a [script] that was periodically run by the
cargo team. Relative to the original script, the push of the snapshot
branch is no longer forced. The job will fail if run more than once on
the same day. (If the first attempt fails before pushing a new root
commit upstream, then retries should succeed as long as the snapshot
can be fast-forwarded.)

[script]: https://github.com/rust-lang/crates-io-cargo-teams/issues/47#issuecomment-506236036

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -1,10 +1,11 @@
 use std::collections::HashMap;
-use std::fs::{self, OpenOptions};
+use std::fs::{self, File, OpenOptions};
+use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 
 use chrono::Utc;
 use swirl::PerformError;
-use tempfile::{Builder, TempDir};
+use tempfile::TempDir;
 use url::Url;
 
 use crate::background_jobs::Environment;
@@ -148,7 +149,7 @@ pub struct Repository {
 
 impl Repository {
     pub fn open(repository_config: &RepositoryConfig) -> Result<Self, PerformError> {
-        let checkout_path = Builder::new().prefix("git").tempdir()?;
+        let checkout_path = tempfile::Builder::new().prefix("git").tempdir()?;
 
         let repository = git2::build::RepoBuilder::new()
             .fetch_options(Self::fetch_options(&repository_config.credentials))
@@ -383,7 +384,7 @@ pub fn squash_index(env: &Environment) -> Result<(), PerformError> {
     println!("Squashing the index into a single commit.");
 
     let now = Utc::now().format("%Y-%m-%d");
-    let head = repo.head_oid()?;
+    let original_head = repo.head_oid()?;
     let msg = format!("Collapse index into one commit\n\n\
 
         Previous HEAD was {}, now on the `snapshot-{}` branch\n\n\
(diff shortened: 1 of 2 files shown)


pietroalbini:
From a glance the code looks correct, but I'm really worried about force pushing without lease. I'd rather prefer to shell out to git just for this operation rather than doing a force push without lease.

jtgeibel:
@pietroalbini I've updated the background job to now shell out to git for the `--force-with-lease` option.

Password authentication is not supported as we do not use that in production or staging. The SSH key is written to a temporary file which is deleted after the push operation completes. Finally, the SSH host key is automatically accepted as long as no conflicting key already exists.

I've tested this on staging. The production deploy key will need to enable forced pushes to `master` before this can be used on production.

src/git.rs
```diff
@@ -357,3 +376,69 @@ pub fn yank(
...
+
+    let key = match &repo.credentials {
+        Credentials::Ssh { key } => key,
+        Credentials::Http { .. } => {
+            return Err(String::from("squash_index: Password auth not supported").into())
+        }
+        _ => return Err(String::from("squash_index: Could not determine credentials").into()),
+    };
+
+    let temp_key_file = tempfile::Builder::new().tempfile()?;
```
pietroalbini, line 412: On Linux (and the production environment) this will create the file in `/tmp`, but it'd be better to create it in `/dev/shm`: files stored in `/dev/shm` are kept in memory and never persisted on disk. On other OSes we can use the default.

src/git.rs
```diff
@@ -357,3 +376,69 @@ pub fn yank(
...
+            "HEAD:refs/heads/master",
+            &format!("--force-with-lease=refs/heads/master:{}", original_head),
+        ])
+        .output()?;
+
+    if !output.status.success() {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        let message = format!("Running git command failed with: {}", stderr);
+        return Err(message.into());
+    }
```
pietroalbini, line 439: I'm wondering... should we delete the `snapshot-` branch we just created if this fails?

pietroalbini (changes requested):

src/git.rs
jtgeibel: Agree, done in first new commit.

src/git.rs
```diff
@@ -357,3 +376,69 @@ pub fn yank(
...
+
+        Previous HEAD was {}, now on the `snapshot-{}` branch\n\n\
+
+        More information about this change can be found [online] and on [this issue].\n\n\
+
+        [online]: https://internals.rust-lang.org/t/cargos-crate-index-upcoming-squash-into-one-commit/8440\n\
+        [this issue]: https://github.com/rust-lang/crates-io-cargo-teams/issues/47", original_head, now);
+
+    // Create a snapshot branch of current `HEAD`.
+    repo.push(&format!("HEAD:refs/heads/snapshot-{}", now))?;
```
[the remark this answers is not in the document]
jtgeibel: Addressed in 2nd commit.

src/git.rs
jtgeibel: I've combined both push operations, adding `--atomic` to the command line args.

jtgeibel:
I've addressed the comments and tested the new version on staging.

src/git.rs
pietroalbini: Even better! And TIL about `--atomic`.

pietroalbini (approved):

pietroalbini:
Thanks! In the future we should add tests for the whole git module, but let's not block this on that implementation as it's not going to be trivial (as we'll need to mock a git server for tests).

@bors r+

PR #3596, opened 2021-05-07
Run `cargo update` and bump to latest stable on production

jtgeibel:
This addresses a new deprecation in `Url`.

r? @Turbo87


Turbo87:
@jtgeibel https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html 

we might want to bump again before deploying to prod

PR #3600, opened 2021-05-08
Fix missing line breaks for crate name and refine sidebar layout

Unbinilium:
Notified that *crates.io* seems not optimized for small screen devices while I was finding crates on my phone, some long crate names made the website in a mess due to missing line breaks, in my opinion a simple workaround would apply to it by adding line breaks.

Test on my laptop with Safari 14.1:

- Add overflow wrap to the names of crates

  

- Add word-break to the header name and no-readme texts in detailed crate pages

  

- Add word-break to texts inside the sidebar copy button 

  

- Change the flex layout wrap to column when window width is less than `890px` to avoid warp overflow in row

  

This would not be the best workaround, it only provides a better experience while surfing the *crates.io* on the phone, on small screen devices or on a split window view.

### DIFF
--- a/app/components/crate-header.module.css
+++ b/app/components/crate-header.module.css
@@ -17,6 +17,10 @@
         padding: 0;
     }
 
+    h1 {
+        word-break: break-word;
+    }
+
     h2 {
         color: var(--main-color-light);
         margin-left: 10px;
--- a/app/components/crate-row.module.css
+++ b/app/components/crate-row.module.css
@@ -22,6 +22,7 @@
     font-weight: bold;
     text-decoration: none;
     font-size: 120%;
+    overflow-wrap: break-word;
 }
 
 .version {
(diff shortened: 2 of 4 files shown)


Unbinilium:
Fetched upstream commits to test CI build check again

Turbo87:
@Unbinilium thanks for working on this, and sorry for the review delay

one small suggestion for future PRs: if you make four separate changes in the same PR (and in this case even the same commit), they block each other from getting merged if there is a disagreement on one of them, and it also makes it harder on the reviewer to review them because it can be difficult to figure out what code change belongs to which conceptual change. in other words: it would be easier on the reviewer if PRs like this were split up into several commits, or even PRs. if you want to read up some more on this I recommend having a look at https://google.github.io/eng-practices/review/developer/small-cls.html :)

@bors r+

PR #3609, opened 2021-05-11
Reject requests with large page offset

jtgeibel:
`WEB_PAGE_OFFSET_UA_BLOCKLIST` is a comma seperated list of user-agents
that will be blocked if `WEB_MAX_ALLOWED_PAGE_OFFSET` is exceeded.

`WEB_MAX_ALLOWED_PAGE_OFFSET` defaults to 200.
`WEB_PAGE_OFFSET_UA_BLOCKLIST` defaults to the empty list. Because
empty user-agents are already blocked for non-download endpoints, the
check is disabled if this variable is empty or undefined.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,5 +1,5 @@
 use crate::publish_rate_limit::PublishRateLimit;
-use crate::{env, uploaders::Uploader, Env, Replica};
+use crate::{env, env_optional, uploaders::Uploader, Env, Replica};
 
 #[derive(Debug)]
 pub struct Config {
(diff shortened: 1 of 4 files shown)


pietroalbini:
Hmm, my worry with this PR is that we would also reject clients complying with our crawlers policy.

jtgeibel:
Sorry for the delay on this. I've added a 2nd environment variable to block specific user-agents only. I've dropped the frontend commit, and have updated the PR description.

pietroalbini (approved):

pietroalbini:
@bors r+

Thanks!

PR #3612, opened 2021-05-11
Drop crates.textsearchable_index_col from the export

jtgeibel:
This column is postgres specific and is normally populated via a
trigger. The trigger is now enabled during the import so that the column
can be dropped from the export.

This addresses part of what was raised in bullet point 2 of #2078. The large readme column remains because there could be people using that data, but the text search column is redundant.

r? @smarnach 
cc @kornelski

### DIFF
--- a/src/tasks/dump_db/dump-import.sql.hbs
+++ b/src/tasks/dump_db/dump-import.sql.hbs
@@ -16,6 +16,9 @@ BEGIN;
     TRUNCATE "{{this.name}}" RESTART IDENTITY CASCADE;
 {{~/each}}
 
+    -- Enable this trigger so that `crates.textsearchable_index_col` can be excluded from the export
+    ALTER TABLE "crates" ENABLE TRIGGER "trigger_crates_tsvector_update";
+
     -- Import the CSV data.
 {{~#each tables}}
     \copy "{{this.name}}" ({{this.columns}}) FROM 'data/{{this.name}}.csv' WITH CSV HEADER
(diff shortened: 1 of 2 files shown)


src/tasks/dump_db/dump-import.sql.hbs
```diff
@@ -16,6 +16,9 @@ BEGIN;
     TRUNCATE "{{this.name}}" RESTART IDENTITY CASCADE;
 {{~/each}}
 
+    -- Enable this trigger so that `crates.textsearchable_index_col` can be excluded from the export
+    ALTER TABLE crates ENABLE TRIGGER trigger_crates_tsvector_update;
```
pietroalbini, line 20: Doesn't change anything, mostly for consistency :D

```suggestion
    ALTER TABLE "crates" ENABLE TRIGGER "trigger_crates_tsvector_update";
```

pietroalbini (approved):
The code changes look good!

I tested locally what the impact of this would be for how long it takes to import a database dump, since the trigger would have to be executed every crate we insert. To test that I downloaded today's dump, removed `version_downloads` as that takes ages and is not relevant and I made a copy of the dump with the changes of this PR (enabling the trigger and removing the column from the dump):

* Status quo: 42.3 seconds (176 MB `crates.csv`)
* With this PR: 56.7 seconds (95 MB `crates.csv`)

So, with this importing the dump is 34% slower. We're removing 46% of the data in `crates.csv` though, which will definitely help download size and everyone who parses the `csv`s manually instead of importing them in PostgreSQL.

Personally I'd say the tradeoff is worth it.

jtgeibel:
Thanks for the analysis of the space/time tradeoff. I've pushed a new commit with your consistency suggestion.

rupert1975:
Hello

PR #3638, opened 2021-05-21
Disable page offset blocking if UA blocklist is unset

jtgeibel:
This fixes a bug where if `WEB_PAGE_OFFSET_UA_BLOCKLIST` was not set,
then all user-agents were blocked for large page offsets. This was
because all string contain the empty string.

Now, no blocking occurs if the variable is unset. If set to the empty
string, or the list contains an empty string, then all user-agents will
be blocked.

r? @pietroalbini

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -67,7 +67,9 @@ impl Default for Config {
     /// - `WEB_MAX_ALLOWED_PAGE_OFFSET`: Page offsets larger than this value are rejected. Defaults
     ///   to 200.
     /// - `WEB_PAGE_OFFSET_UA_BLOCKLIST`: A comma seperated list of user-agent substrings that will
-    ///   be blocked if `WEB_MAX_ALLOWED_PAGE_OFFSET` is exceeded.
+    ///   be blocked if `WEB_MAX_ALLOWED_PAGE_OFFSET` is exceeded. Including an empty string in the
+    ///   list will block *all* user-agents exceeding the offset. If not set or empty, no blocking
+    ///   will occur.
     ///
     /// # Panics
     ///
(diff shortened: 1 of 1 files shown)


pietroalbini:
While the implementation looks fine I think we can make it simpler.

I see no need to convert `Vec<_>` into `Option<Vec<_>>`, we can just add a `!page_offset_ua_blocklist.is_empty()` in the `if` clause to avoid filtering when no filter is set. We can also exclude empty filters with `.filter(|s| !s.is_empty())` in `config.rs`.

src/config.rs
```diff
@@ -67,7 +67,9 @@ impl Default for Config {
     /// - `WEB_MAX_ALLOWED_PAGE_OFFSET`: Page offsets larger than this value are rejected. Defaults
     ///   to 200.
     /// - `WEB_PAGE_OFFSET_UA_BLOCKLIST`: A comma seperated list of user-agent substrings that will
-    ///   be blocked if `WEB_MAX_ALLOWED_PAGE_OFFSET` is exceeded.
+    ///   be blocked if `WEB_MAX_ALLOWED_PAGE_OFFSET` is exceeded. Setting to the empty string (or
+    ///   includuing an empty string in the list) will block *all* user-agents exceeding the
```
JohnTitor, line 71: nit:
```suggestion
    ///   including an empty string in the list) will block *all* user-agents exceeding the
```

jtgeibel:
I've updated the implementation to disable blocking if the variable is not set, or is set to the empty string. If the environment variable is non-empty and contains an empty value in the list, then all user-agents will be blocked. I left this as an option in case we ever need to temporarily enforce a broad block.

PR #3648, opened 2021-05-25
Implement seek-based pagination for filterless sortless crates endpoint

pietroalbini:
This PR implements seek-based pagination for parts of the `/api/v1/crates` endpoint, and adds the infrastructure to parse `seek=` parameters in paginated APIs. The reason for this PR is that we have multiple crawlers scraping the whole `/api/v1/crates` endpoint every day, and offset-based pagination is not feasible after a high number of pages.

To keep the implementation simple seek-based pagination is only supported for requests matching these constraints:

* **No filters:** when the request is filtered the total results count is not the same as the total number of crates, so the current implementation to gather the number of results (`SELECT COUNT(*) FROM crates;`) does not work. This can be lifted in the future by refactoring how we gather the total number of results.
* **No sorting other than alphanumeric:** this is just not implemented yet, we'd need to change the condition used to seek to the correct position. Note that for anything that's not alphanumeric we'll need to sort by `($param, id)`, not just `$param`, to ensure the results are consistently ordered (there could be multiple crates with the same download count).
* **No search queries:** search queries complicate the SQL query quite a bit, introducing relevance sorting and hardcoding the exact match as the first result. Not worth implementing for the time being IMO.

When the endpoint is called with a request that supports seeking, no `meta.prev_page` will be provided and `meta.next_page` will include the `seek` parameter instead of the `page` parameter. `meta.total` will continue to be provided as usual.

The value of `seek` is a base64-encoded JSON value. The reason base64 is used is to signal to our API consumers the seek value is an implementation detail and they shouldn't manually create one. This is what other big providers do (for example GitHub). The reason to encode the value in JSON is to support more complex seek keys in the future: right now we're just encoding a bare number, and the JSON representation is the same as the ASCII representation.

This PR is fully backward compatible: if clients just rely on `meta.next_page` they will transparently start using seek-based pagination, but offset-based pagination still works. When the `page=` attribute is provided offset-based pagination will be forced, and `meta.next_page` will include `page` instead of `seek`. This way, clients manually creating `?page=` URLs will not break.

One unfortunate aspect of this PR is that we still have to provide `meta.total`, as that's required by the Cargo registries API. Getting the total count requires a full scan over the table, which right now in production is done using an index-only scan in ~20ms.

The PR is best reviewed commit-by-commit.
r? @jtgeibel

### DIFF
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -61,7 +61,23 @@ pub(crate) struct PaginationOptions {
 }
 
 impl PaginationOptions {
-    pub(crate) fn new(req: &mut dyn RequestExt) -> AppResult<Self> {
+    pub(crate) fn builder() -> PaginationOptionsBuilder {
+        PaginationOptionsBuilder
+    }
+
+    pub(crate) fn offset(&self) -> Option<u32> {
+        if let Page::Numeric(p) = self.page {
+            Some((p - 1) * self.per_page)
+        } else {
+            None
+        }
+    }
+}
+
+pub(crate) struct PaginationOptionsBuilder;
+
+impl PaginationOptionsBuilder {
+    pub(crate) fn gather(self, req: &mut dyn RequestExt) -> AppResult<PaginationOptions> {
         const DEFAULT_PER_PAGE: u32 = 10;
         const MAX_PER_PAGE: u32 = 100;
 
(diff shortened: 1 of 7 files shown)


src/controllers/helpers/pagination.rs
```diff
@@ -215,59 +216,88 @@ where
...
-            .gather(&mut req)
-            .unwrap_err()
-            .response()
-            .unwrap();
-        assert_eq!(per_page_error.status(), StatusCode::BAD_REQUEST);
+        assert_eq!(StatusCode::BAD_REQUEST, response.status());
+
+        // Conduit doesn't expose a method to get the contents of a body, so we need to extract it
+        // from the underlying body variants. Please replace this with a proper method if conduit
+        // ever adds support for that.
```
jtgeibel, line 384: It's not very discoverable, but you can `use conduit_test::ResponseExt;` and then call `request.into_cow()` which returns a `Cow<[u8]>`.

jtgeibel (commented):
:partying_face: This looks great! Just a few minor comments, then feel free to r=me.

src/controllers/helpers/pagination.rs
```diff
@@ -215,59 +216,88 @@ where
...
+        assert_error("per_page=not_a_number", "invalid digit found in string");
+        assert_error("per_page=1.0", "invalid digit found in string");
+        assert_error("per_page=101", "cannot request more than 100 items");
 
-        let mut req = mock("page=1.0");
-        let page_error = Page::new(&mut req).unwrap_err().response().unwrap();
+        let pagination = PaginationOptions::builder()
+            .gather(&mut mock("page=5"))
+            .unwrap();
+        assert_eq!(Page::Numeric(5), pagination.page);
```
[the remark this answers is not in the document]
pietroalbini: Copy-paste :sweat_smile:

src/controllers/helpers/pagination.rs
pietroalbini: TIL!

PR #3650, opened 2021-05-26
models::Version: Change `num` field to a plain `String`

Turbo87:
This PR prepares for `semver` v1.x no longer having a `diesel` feature, which would make it much more complicated for us to directly deserialize the SQL column into a `semver::Version` field. We can deserialize it to a `String` instead, and only parse it into a `semver::Version` on demand.

### DIFF
--- a/src/controllers/krate/downloads.rs
+++ b/src/controllers/krate/downloads.rs
@@ -23,7 +23,7 @@ pub fn downloads(req: &mut dyn RequestExt) -> EndpointResult {
     let krate: Crate = Crate::by_name(crate_name).first(&*conn)?;
 
     let mut versions: Vec<Version> = krate.all_versions().load(&*conn)?;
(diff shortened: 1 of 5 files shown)


jtgeibel:
LGTM! It seems there are [4 existing versions](https://github.com/dtolnay/semver/pull/239#issuecomment-848449862) that cannot be parsed in semver v1.x due to a leading zero in the pre-release portion. We'll want to do a bit more testing on those when bumping semver, but this seems like the most reasonable approach.

I've tested a few things locally by adding a few version numbers of `-1` to the database, which fails to parse with the current semver. When sorting in descending order, it look like any un-parsable versions will appear at the end (in JSON, the frontend does its own sorting and seems to maybe fallback to version id). Filtering bad semvers in `TopVersions::from_date_version_pairs` just causes the version to be blank on the search results in the worst case (currently only affecting `hxgm30-client` with only a single invalid version).

@bors r+

PR #3653, opened 2021-05-27
Update `semver` to v1.x

Turbo87:
This PR updates the `semver` crate to the new v1.0.0 release, which is basically a complete rewrite. The `diesel` feature no longer exists, but due to #3650 we don't need it anymore.

Closes https://github.com/rust-lang/crates.io/pull/3338


locks (approved):

jtgeibel:
I've tested some edge cases locally with an import of the database dump and LGTM! @bors r+

PR #3658, opened 2021-05-27
ember-test-selectors: Disable obsolete/deprecated `Component.reopen()` call

Turbo87:
see https://github.com/simplabs/ember-test-selectors/pull/721


pichfl (approved):
Yes please!

PR #3684, opened 2021-06-07
Add more instance-level metrics

pietroalbini:
Now that we have support for collecting metrics this PR adds the following ones:

* `response_times`: how long preparing responses take, grouped by endpoint. Note that this does not create a metric series for each visited URL: for example all the downloads will be counted as `/crates/:crate_id/:version/download`.
* `responses_by_status_code_total` how many responses we sent, grouped by HTTP status code.
* `downloads_not_counted_total` how many download counts aren't persisted yet.

This also had to implement histograms support, needed by `response_times`.


jtgeibel:
LGTM! @bors r+

PR #3685, opened 2021-06-08
Upgrade git2 to fix compliation on nightly

jtgeibel:
The compilation error is a known regression related to a soundness fix.
The latest git2 avoids the error. See also
https://github.com/rust-lang/rust/issues/85574#issuecomment-847227106

r? @Turbo87


Turbo87 (approved):
LGTM

feel free to r=me once CI has passed 👍

PR #3699, opened 2021-06-13
Fix missing word-break and add theme-color meta for Safari 15.0

Unbinilium:
Fix missing line breaks in crates "Versions", "Dependencies" and "Dependents" page, because some long crate names made the website in a mess on small screen devices, personally that add some 'word-break' would be a simple workaround, related to #3600, and this was a further fix.

Another subjective refine for upcoming Safari 15.x is add 'theme-color' meta to the `index.html` header to let the address bar color of Safari get matched with the header color of *crates.io*. Personally I choose the same color as the website header, its equals to `--green800`.

Test on my laptop with Safari 15.0:

- Add a `word-break` to class `page-description` in "Version" page

Screen Shot 2021-06-13 at 19 56 20

- Add a `word-break` to class `no-deps` in "Dependencies" page while there is no other dependencies

Screen Shot 2021-06-13 at 19 56 54

- Add a `word-break` to class `results-count` in "Dependents" page

Screen Shot 2021-06-13 at 19 57 22

- Add `theme-color` meta to the `index.html` header for upcoming Safari 15.x (on right side, the color is changeable using "Toggle Design" button in development mode), compared with the original website (on left side)

https://user-images.githubusercontent.com/15633984/121806498-a3272b00-cc82-11eb-9e09-6f0fcfcb045e.mov

Thanks for reviewing!

### DIFF
--- a/app/components/header.module.css
+++ b/app/components/header.module.css
@@ -54,7 +54,7 @@ input.search {
     background-color: white;
     background-image: url('/assets/search.png');
     background-repeat: no-repeat;
-    background-position: 6px 6px;
+    background-position: 8px center;
     background-size: 14px 15px;
     border-radius: 15px;
     box-shadow: none;
--- a/app/components/results-count.module.css
+++ b/app/components/results-count.module.css
@@ -1,5 +1,7 @@
 .results-count {
     composes: small from '../styles/shared/typography.module.css';
+
+    word-break: break-all;
 }
 
 .highlight {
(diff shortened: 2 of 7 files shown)


Unbinilium:
Sorry I didn't double check my code for possible issues, causing to have the additional commits to fix. Now it would be work as excepted.

Turbo87:
sorry, but this is getting a little complicated to review. can you split this PR so that each PR only changes one thing? the theme color thing seems unrelated to the work-break changes, right?

Unbinilium:
> sorry, but this is getting a little complicated to review. can you split this PR so that each PR only changes one thing? the theme color thing seems unrelated to the work-break changes, right?

Ok, I'll do it later.

PR #3703, opened 2021-06-14
Fix search icon position to vertical center

Unbinilium:
Fix search icon position to vertical center, compared in attached image:

Screen Shot 2021-06-14 at 21 38 57

*split PR in #3699

### DIFF
--- a/app/components/header.module.css
+++ b/app/components/header.module.css
@@ -54,7 +54,7 @@ input.search {
     background-color: white;
     background-image: url('/assets/search.png');
     background-repeat: no-repeat;
-    background-position: 6px 6px;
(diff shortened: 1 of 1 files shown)


Turbo87:
@Unbinilium I've just tried this out locally and in my browser the icon is now no longer centered (and Percy seems to agree) 😅 

If we want to fix this properly I think we will need something very roughly like:

```hbs
<div class="looks-like-an-input">
  <svg class="prefix-icon">
  <input class="the-actual-input">
</div>
```

and then we can use e.g. CSS flexbox to vertically center the prefix icon properly.

Turbo87:
closing due to inactivity

PR #3704, opened 2021-06-14
Fix missing line break

Unbinilium:
Fix missing line breaks in crates "Versions", "Dependencies" and "Dependents" page, because some long crate names made the website in a mess on small screen devices, personally that add some 'word-break' would be a simple workaround ('padding' is also needed), related to #3600, and this was a further fix.

Test on my laptop with Safari 15.0:

- Add a `word-break` to class `page-description` in "Version" page

121806342-0795ba80-cc82-11eb-9739-562461f97a35

- Add a `word-break` to class `no-deps` in "Dependencies" page while there is no other dependencies

121806351-17150380-cc82-11eb-99d6-35c2a3819771

- Add a `word-break` to class `results-count` in "Dependents" page

121806356-1bd9b780-cc82-11eb-8822-112b2dfe5664

Thanks for reviewing, split PR of #3699

### DIFF
--- a/app/components/results-count.module.css
+++ b/app/components/results-count.module.css
@@ -1,5 +1,7 @@
 .results-count {
     composes: small from '../styles/shared/typography.module.css';
+    word-break: break-all;
+    padding-right: 20px;
 }
 
 .highlight {
--- a/app/styles/crate/version-dependencies.module.css
+++ b/app/styles/crate/version-dependencies.module.css
@@ -9,3 +9,7 @@
         }
     }
 }
+
+.no-deps {
+     word-break: break-all;
+ }
--- a/app/styles/crate/versions.module.css
+++ b/app/styles/crate/versions.module.css
@@ -11,6 +11,8 @@
 
 .page-description {
     composes: small from '../shared/typography.module.css';
+    word-break: break-all;
+    padding-right: 20px;
 
     @media only screen and (max-width: 550px) {
         display: block;


Turbo87:
thanks! :)

@bors r+

Turbo87:
Bildschirmfoto 2021-06-14 um 18 14 25

actually, this looks a bit broken to me 🤔 

@bors r-

Unbinilium:
> Bildschirmfoto 2021-06-14 um 18 14 25
> 
> actually, this looks a bit broken to me 🤔
> 
> @bors r-

Thanks for point out the issue, I'll test on other devices later.

Unbinilium:
Screen Shot 2021-06-15 at 05 12 54

This would be fine, but there're still some layout issue on other pages. It may not easy to discover and fix that at once, because there're many code repetitively and without a specialized class name for styling like this:

https://github.com/rust-lang/crates.io/blob/33aa325d7b489e140bd8c34745398efd801106f1/app/templates/user.hbs#L18-L28

A workaround in this way would take some time.

app/styles/search.module.css
```diff
@@ -3,6 +3,16 @@
...
+
+    .search-sort {
+        display: flex;
+        flex-direction: row;
+        align-items: center;
+
+        span {
+            padding-inline-end: 6px;
+        }
+    }
```
Turbo87, line 15: can you explain how the flexbox here is related to the missing line breaks?

PR #3715, opened 2021-06-17
Address clippy findings

jtgeibel:
This addresses warnings identified by clippy. Some of these warnings are currently generated on stable, but a missing environment variable caused those warnings to be logged but not to fail the build. The next to last commit fixes this.

This final commit pins to 1.52.1 for linting, until a compiler ICE in 1.53.0 is addressed. I've just [reported] this ICE against the pre-release, but it may be too late to address in 1.53.0.

r? @Turbo87 

[reported]: https://internals.rust-lang.org/t/1-53-0-prerelease-testing/14884/2?u=jtgeibel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,8 +102,9 @@ jobs:
       - name: Install Rust
         run: |
           rustup set profile minimal
-          rustup update stable
-          rustup default stable
+          # Pin to older version until a clippy regression is fixed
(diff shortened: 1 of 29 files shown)


Turbo87:
looks great, thanks! :)

@bors r+

Turbo87:
also, thanks for the detailed commit list. makes it much easier to review :)

PR #3717, opened 2021-06-17
Cleanup unneeded lints and drop unused Debug impls

jtgeibel:
The lib crate is only used by other in-tree crates. Therefore if Debug
or Copy impls are ever needed, they can be easily added.

82 unused Debug impls are then dropped. This probably has a slight but
negligible improvement to compile/link times.

r? @Turbo87

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -15,8 +15,6 @@ use scheduled_thread_pool::ScheduledThreadPool;
 
 /// The `App` struct holds the main components of the application like
 /// the database connection pool and configurations
-// The db, oauth, and git2 types don't implement debug.
-#[allow(missing_debug_implementations)]
 pub struct App {
     /// The primary database connection pool
     pub primary_database: DieselPool,
--- a/src/background_jobs.rs
+++ b/src/background_jobs.rs
@@ -20,7 +20,6 @@ impl swirl::db::DieselPool for DieselPool {
     }
 }
 
-#[allow(missing_debug_implementations)]
 pub struct Environment {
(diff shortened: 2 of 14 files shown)


Turbo87:
hmmm..... I like the first two commits of the PR, but I'm not sure about the third one that is dropping all of the existing `Debug` traits. They might not be used right now, but they make it slightly easier to debug while developing. Unless the build time improvement is significant I think I'd prefer to keep them 🤔 

@pietroalbini any thoughts on this?

jtgeibel:
@Turbo87 yeah I was a bit split on the final commit myself. I think it might be worth dropping the Debug impls and adding them back as they prove useful for debugging. I suspect a lot of these were added just to satisfy the lint. For example, I doubt the Debug impls on view structs is very useful, since the purpose of these is to implement Serialize for generating response data (which is already easily inspectable).

However, I definitely have no concern with dropping the 3rd commit either. I don't think it would be worth the effort to go through the list of structs to analyze them ahead of time, and would gladly drop the commit from this series if that is preferred.

pietroalbini:
Forced `Debug` is annoying when there are structs with fields that don't support `Debug` themselves, so I agree with removing the lint. I'd also check for manual impls of `Debug` that could be removed, I think I added a few of them myself.

Turbo87:
I've rebased and dropped the third commit for now (still available as https://github.com/jtgeibel/crates.io/commit/2c80e6b932ed78ed69d99e65a98280cbb1fb0067) :)

@bors r+

Issue #3532, opened 2021-04-18
crates.io denies underscore-prefixed dependency in manifest


jmagnuson:
**Describe the bug**
Running `cargo publish`, I get the following error:
```
Compiling linemux v0.2.0 (/home/jon/src/jmagnuson/linemux/target/package/linemux-0.2.0)
    Finished dev [unoptimized + debuginfo] target(s) in 13.08s
   Uploading linemux v0.2.0 (/home/jon/src/jmagnuson/linemux)
error: api errors (status 200 OK): invalid upload request: invalid value: string "_tokio", expected a valid crate name to start with a letter, contain only letters, numbers, hyphens, or underscores and have at most 64 characters at line 1 column 223
```

For context, I am renaming the `tokio` dependency to `_tokio` to allow for a `tokio` feature which can enable multiple crates/sub-features:
```toml
[features]
default = ["tokio"]
tokio = ["_tokio"]

[dependencies]
_tokio = { package = "tokio", version = "1.0", features = ["fs", "io-util", "sync", "time"], optional = true }
```

I saw that `sqlx` did [something similar](https://github.com/launchbadge/sqlx/blob/e7664d8d192218096b374004b23857c04130f442/sqlx-core/Cargo.toml#L45) (which I shamelessly copied) for runtime variance, but now I am seeing that this change hasn't been published yet, so I wonder if they have discovered this limitation yet themselves.

**To Reproduce**
1. Checkout [linemux 0.2.0](https://github.com/jmagnuson/linemux/tree/0.2.0)
2. run `cargo publish --dry-run`, exits successfully
3. (Assuming permissions to publish to crates.io) run `cargo publish`, fails to publish

**Expected behavior**
Either crates.io accepts `_tokio`, or an error is thrown during the client-side dry-run.

**Additional context**
Looks like there have been some previous tickets surrounding feature names (#1395, #3089), but I didn't find one matching this particular assertion.

PR #3724, opened 2021-06-19
Allow underscore-prefixed renamed dependency in manifest

hi-rustin:
close https://github.com/rust-lang/crates.io/issues/3532


Turbo87:
👋 @hi-rustin 

the issue here is slightly more complicated. crate name generally are not allowed to start with an underscore, so adjusting `valid_ident()` to allow it could create some problems for us when people suddenly start publishing crates with a leading underscore.

https://github.com/rust-lang/crates.io/pull/1487 implemented support for "renamed dependencies", but it looks like we did not test for the "`name` has a leading underscore" case. I think the first step in fixing this issue is to create a new test, roughly similar to the existing `new_with_renamed_dependency()` test, but with a leading underscore for the `name`. That should hopefully reproduce the issue that @jmagnuson is seeing in #3532.

hi-rustin:
> I think the first step in fixing this issue is to create a new test, roughly similar to the existing `new_with_renamed_dependency()` test, but with a leading underscore for the `name`.

Added.

hi-rustin:
The test case will fail when there is no such underlined change.

```log
failed to decode: Error("missing field `crate`", line: 1, column: 249)
thread 'krate::publish::new_with_underscore_name_dependency' panicked at 'failed to decode: Error("missing field `crate`", line: 1, column: 249)', src/tests/util/response.rs:111:19
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/panicking.rs:435:5
   2: all::util::response::json
             at ./src/tests/util/response.rs:111:19
   3: all::util::response::Response<T>::good
             at ./src/tests/util/response.rs:26:9
   4: all::krate::publish::new_with_underscore_name_dependency
             at ./src/tests/krate/publish.rs:187:5
   5: all::krate::publish::new_with_underscore_name_dependency::{{closure}}
             at ./src/tests/krate/publish.rs:174:1
   6: core::ops::function::FnOnce::call_once
             at /Users/rustin/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

error: test failed, to rerun pass '-p cargo-registry --test all'

Process finished with exit code 101
```

hi-rustin:
I did some local debugging and integration testing to make sure the logic was being tested.

src/tests/krate/publish.rs
```diff
@@ -202,6 +228,34 @@ fn new_krate_with_dependency() {
...
+        CrateBuilder::new("_foo-dep", user.as_model().id).expect_build(conn);
+    });
+
+    let dependency = DependencyBuilder::new("_foo-dep").version_req("1.0.0");
+
+    let crate_to_publish = PublishBuilder::new("new_dep")
+        .version("1.0.0")
+        .dependency(dependency);
+
+    token.enqueue_publish(crate_to_publish).good();
```
Turbo87, line 247: `_foo-dep` is not a legal crate name, so this call should actually fail, because in this case we were not renaming the dependency

hi-rustin: Oh, I see what you mean! _ can only be used in renaming.

Turbo87: yeah, exactly

Turbo87: I'm not 100% sure, but I think `explicit_name_in_toml: Option<EncodableCrateName>` is the issue, and maybe it should be `explicit_name_in_toml: Option<String>` instead, or at least some type that does allow underscore prefixes.

hi-rustin: Yes, you are right! I have added it. I'll sort out the code and submit it tomorrow!

Turbo87 (approved):
looks good to me now, but I would like to get a second pair of eyes on this before we merge it.

@pietroalbini @jtgeibel @JohnTitor

pietroalbini (changes requested):
The code changes themselves look great, but I'd prefer for the internal API to be harder to misuse and misunderstand:

* The boolean `Crate::valid_name("foo", true)` or `Crate::valid_name("bar" false)` is a bit confusing if one doesn't know about this PR. I'd prefer to keep `Crate::valid_name` the same as before and adding a `Crate::valid_dependency_name`.
* `EncodableExplicitName` is also a bit confusing, it'd be better to call it `EncodableDependencyName`.

hi-rustin:
> The code changes themselves look great, but I'd prefer for the internal API to be harder to misuse and misunderstand:
> 
>     * The boolean `Crate::valid_name("foo", true)` or `Crate::valid_name("bar" false)` is a bit confusing if one doesn't know about this PR. I'd prefer to keep `Crate::valid_name` the same as before and adding a `Crate::valid_dependency_name`.
> 
>     * `EncodableExplicitName` is also a bit confusing, it'd be better to call it `EncodableDependencyName`.

Addressed! Thanks for your review.

pietroalbini (approved):
Thanks!

jmagnuson:
Thank you, @hi-rustin, and everyone who helped get this merged!

PR #3730, opened 2021-06-21
crate/settings: Replace custom `setupController()` implementation with `model()` hook

Turbo87:
... and load the necessary relationships in the `model()` hook too.

This ensures that we have all the necessary data to render the route, before we tell Ember.js that we're ready.


Turbo87:
nevermind, this might be a bad idea if we want to load other things on the page in the future too...

PR #3747, opened 2021-06-24
Add optional expiration for rate limit overrides

pietroalbini:
Most of the cases when we increase the new crates publish rate limit is when a big monorepo needs to be published for the first time, resulting in tens of crates being published at the same time. It's extremely rare that the rate limit needs to be increased forever.

To avoid overrides being active forever, which could let to abuse if the user accounts were compromised, this adds an optional expiration.


Turbo87 (approved):
LGTM feel free to r=me

Issue #2868, opened 2020-10-04
No way to view status or cancel outstanding owner invitations


Nemo157:
After inviting someone to be an owner of a crate there is no way to see the status of this invitation, other than remembering you sent the invitation and querying owners to see whether they have accepted it. Also, until the invitation is accepted there is no way to cancel/revoke the permission, so if you at some point after sending you realise the invited user is untrustworthy, there is no way to stop them from accepting the invite and removing your ownership permission.

Shnatsel:
Related: invitations never expire (#2869)

pietroalbini:
Opened https://github.com/rust-lang/crates.io/pull/3763 to provide the required data to the frontend. Once that's merged, the frontend needs to display the data and the backend needs to authorize crate owners to decline invitations on their crate.

Note that to prevent abuse the endpoint doesn't publicly expose invitations. An invitation is going to be accessible through the API only if you're authenticated and you're the invited person or an owner of the affected crate.

Shnatsel:
A one use case such privacy prevents is detailed listing of all owners by [`cargo supply-chain`](https://github.com/rust-secure-code/cargo-supply-chain), but this is probably a reasonable trade-off.

cuviper:
When it comes to supply-chain, you also don't know who the current owners may invite tomorrow. I think a pending invitation is not so different, just a step closer, as the invited person is not an owner _yet_.

PR #3763, opened 2021-06-29
Add the `/api/private/crate-owner-invitations` endpoint 

pietroalbini:
The endpoint provides a listing of all the invitations sent to the current user or all the invitations to a crate the user owns. Unauthenticated users or unrelated users won't be able to see others' invitations to prevent abuses.

The two ways to query the endpoint are:

    GET /api/private/crate-owner-invitations?crate_name={name}
    GET /api/private/crate-owner-invitations?invitee_id={uid}

The endpoint is paginated using only seek-based pagination, and the next page field is provided when more results are available.

Once the frontend switches to use the new endpoint we can remove safely remove the old "v1" endpoint, as that's only used for the frontend. Because of this, the "v1" endpoint internally uses the same logic as the new one and converts the data to the old schema.

Part of #2868 
r? @Turbo87

### DIFF
--- a/src/controllers/crate_owner_invitation.rs
+++ b/src/controllers/crate_owner_invitation.rs
@@ -2,7 +2,7 @@ use super::frontend_prelude::*;
 
 use crate::models::{CrateOwnerInvitation, User};
 use crate::schema::{crate_owner_invitations, crates, users};
-use crate::views::{EncodableCrateOwnerInvitation, EncodablePublicUser, InvitationResponse};
+use crate::views::{EncodableCrateOwnerInvitationV1, EncodablePublicUser, InvitationResponse};
 use diesel::dsl::any;
 use std::collections::HashMap;
 
@@ -62,7 +62,7 @@ pub fn list(req: &mut dyn RequestExt) -> EndpointResult {
                 .unwrap_or_else(|| String::from("(unknown crate name)"));
 
             let expires_at = invitation.expires_at(config);
-            EncodableCrateOwnerInvitation::from(invitation, inviter_name, crate_name, expires_at)
+            EncodableCrateOwnerInvitationV1::from(invitation, inviter_name, crate_name, expires_at)
         })
         .collect();
 
@@ -74,7 +74,7 @@ pub fn list(req: &mut dyn RequestExt) -> EndpointResult {
 
     #[derive(Serialize)]
     struct R {
-        crate_owner_invitations: Vec<EncodableCrateOwnerInvitation>,
+        crate_owner_invitations: Vec<EncodableCrateOwnerInvitationV1>,
         users: Vec<EncodablePublicUser>,
     }
     Ok(req.json(&R {
--- a/src/controllers/crate_owner_invitation.rs
+++ b/src/controllers/crate_owner_invitation.rs
@@ -6,7 +6,7 @@ use crate::views::{EncodableCrateOwnerInvitation, EncodablePublicUser, Invitatio
 use diesel::dsl::any;
 use std::collections::HashMap;
 
-/// Handles the `GET /me/crate_owner_invitations` route.
+/// Handles the `GET /api/v1/me/crate_owner_invitations` route.
 pub fn list(req: &mut dyn RequestExt) -> EndpointResult {
     // Ensure that the user is authenticated
     let user = req.authenticate()?.forbid_api_token_auth()?.user();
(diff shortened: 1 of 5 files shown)


bjorn3:
> The two ways to query the endpoint are:
> 
> ```
> GET /api/private/crate-owner-invitations?crate_name={name}
> GET /api/private/crate-owner-invitations?invitee_id={uid}
> ```

So there is no way to list all outstanding invitations you created? Only for specific crates or invitees?

pietroalbini:
Not in the current implementation. This API is going to be used only by the crates.io frontend, and I'm not aware of plans to display all the invitations sent by the current user. Adding a new list filter is trivial though, and if @Turbo87 wants to implement showing all the invitations sent by the current user in the frontend I'm happy to add the extra filter.

src/controllers/crate_owner_invitation.rs
```diff
@@ -83,6 +55,200 @@ pub fn list(req: &mut dyn RequestExt) -> EndpointResult {
...
+
+fn prepare_list(
+    req: &mut dyn RequestExt,
+    auth: AuthenticatedUser,
+    filter: ListFilter,
+) -> AppResult<PrivateListResponse> {
+    let pagination: PaginationOptions = PaginationOptions::builder()
+        .enable_pages(false)
+        .enable_seek(true)
+        .gather(req)?;
```
Turbo87, line 87: if this function is unconditionally using seek pagination then isn't this a breaking change for the existing endpoint? 🤔

src/controllers/crate_owner_invitation.rs
```diff
@@ -83,6 +55,200 @@ pub fn list(req: &mut dyn RequestExt) -> EndpointResult {
...
+                let krate: Crate = Crate::by_name(&crate_name).first(&*conn)?;
+                let owners = krate.owners(&*conn)?;
+                if user.rights(req.app(), &owners)? != Rights::Full {
+                    return Err(forbidden());
+                }
+
+                // Cache the crate name to avoid querying it from the database again
+                crate_names.insert(krate.id, krate.name.clone());
+
+                Box::new(crate_owner_invitations::crate_id.eq(krate.id))
```
Turbo87, line 110: is there a reason why we need to box these? wouldn't it be easier to do something like in the `search` controller, where we prepare a generic `query` and then do `query = query.filter(...)` in the condition branches?

src/controllers/crate_owner_invitation.rs
pietroalbini: Yes, but as I mentioned in the PR body I'd like for the existing endpoint to be removed once we switch the frontend to using the new endpoint. The existing endpoint also rejects token requests, so the only consumer of it is the frontend.

There also shouldn't be any user visibly affected by this change: when writing this only 14 users have pending unexpired invitations, and the most invites an user has is 11:

```
crates-io::DATABASE=> select count(*) from crate_owner_invitations where created_at > current_timestamp - interval '30 days' group by invited_user_id order by count(*) desc;
 count 
-------
    11
     4
     2
     1
     1
     1
     1
     1
     1
     1
     1
     1
     1
     1
(14 rows)
```

src/controllers/crate_owner_invitation.rs
pietroalbini: The search controller [also uses boxes](https://github.com/rust-lang/crates.io/blob/3a01128b314fa67063f22857967ba58ce0b5cf88/src/controllers/krate/search.rs#L58).

Turbo87:
sorry, forgot that this was still waiting on me 😞 

looks great though, so: @bors r+

Turbo87:
rebased to get rid of the merge failure :)

src/router.rs
```diff
@@ -127,6 +127,12 @@ pub fn build_router(app: &App) -> RouteBuilder {
     // Metrics
     router.get("/api/private/metrics/:kind", C(metrics::prometheus));
 
+    // Crate ownership invitations management in the frontend
+    router.get(
+        "/api/private/crate-owner-invitations",
```
Turbo87, line 132: hmm, probably should've noticed this earlier: we've been using snake_case for all the URLs so far, instead of kebab-case

pietroalbini: Ugh, could you send a PR renaming the endpoints?

Turbo87: yep, will do :)

PR #3770, opened 2021-07-03
Improve metrics accuracy

pietroalbini:
This should hopefully improve the accuracy of our download metrics and should stop reporting most of the downloads as taking more than 2ms (most download requests finish processing in around 1ms).


Turbo87 (approved):
feel free to r=me

PR #3775, opened 2021-07-06
Add more histograms

pietroalbini:
This PR adds more histograms to the metrics we collect, exposing information that could be useful when we get paged. Two metrics were added:

* `cratesio_instance_database_time_to_obtain_connection`: how long it takes to obtain a database connection
* `cratesio_instance_downloads_select_query_execution_time`: how long the `SELECT` in the download endpoint takes

A side effect of this is that `TimingRecorder` is not needed anymore: we used it to measure those two things, appending them in the logs for slow requests, but now that we have metrics there is no need to have it. If we need to measure more timings in the future we can just add more histograms. This PR thus removes `TimingRecorder` from the codebase.

r? @jtgeibel or @Turbo87
This can be reviewed commit-by-commit.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -62,6 +62,9 @@ impl App {
     pub fn new(config: config::Server, http_client: Option<Client>) -> App {
         use oauth2::{AuthUrl, ClientId, ClientSecret, TokenUrl};
 
+        let instance_metrics =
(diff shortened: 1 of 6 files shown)


Turbo87:
LGTM 

@bors r+

Issue #3780, opened 2021-07-11
Move new versions rate limit from nginx to the application


pietroalbini:
There are two kinds of rate limits when publishing crates: a rate limit for brand new crates [implemented in the application](https://github.com/rust-lang/crates.io/blob/d7dad029e7dd582cc26fb3daae50091af574cdbd/src/publish_rate_limit.rs) and a rate limit for all published configured in nginx ([1](https://github.com/rust-lang/crates.io/blob/d7dad029e7dd582cc26fb3daae50091af574cdbd/config/nginx.conf.erb#L179), [2](https://github.com/rust-lang/crates.io/blob/d7dad029e7dd582cc26fb3daae50091af574cdbd/config/nginx.conf.erb#L265-L270)). The nginx rate limit, while simpler in its implementation, has multiple problems:

* The rate limit state is stored in every dyno, which results in users actually having at most 2 times the rate limit depending on where the Heroku Router send their requests.
* There is no way to manually increase the limit for a subset of the users.
* The rate limit also applies to new crates, even though those have their separate rate limit.
* The rate limit is per IP address, not per user.

We should move new versions to use the same rate limiting code as new crates, but with different thresholds, and then we should remove the nginx rate limits.

Turbo87:
Resolved by https://github.com/rust-lang/crates.io/pull/6875

PR #3790, opened 2021-07-18
Move new versions rate limits to the application

pietroalbini:
This PR refactors the rate limiting code to support more kinds of rate limits, and moves the new versions rate limit from nginx to the application. See https://github.com/rust-lang/crates.io/issues/3780 for the rationale of this change.

* The rate limiting for new publishes of existing crates is now tracked per-user instead of per-IP, with the ability to add overrides similar to the rate limiting for new crates. The limit is still the usual 1 version/min with a burst of 30, but in practice this **could lower** the rate limiting, as before the limit was between 30 and 60 depending how lucky users were with the load balancing.

* `PublishRateLimit` is now called `RateLimiter`, and for every operation it requires a `LimitedAction`. This allows different rate limits to exist for `LimitedAction::PublishNew` and `LimitedAction::PublishExisting`, and adding new kinds of rate limits only requires adding a new enum variant.

* The environment variables to override the default rate limits changed to `RATE_LIMITING_{KIND}_RATE_SECONDS` and `RATE_LIMITING_{KIND}_BURST`.

There are two things that will be done in followup PRs:

* Remove the default for the `action` column.
* Rename the `publish_limit_buckets` to `rate_limit_buckets` and `publish_rate_overrides` to `rate_limit_overrides`


pietroalbini:
Sorry for the big PR, I tried making it easy to review the PR commit-by-commit.

jtgeibel:
Looks great!

@bors r+

Issue #3129, opened 2020-12-26
Please highlight the performed form of support on crates.io


epage:
**Is your feature request related to a problem? Please describe.**
I was trying to get support with a potential problem with crates.io.  I went to the website and the only relevant link was "Send us an email".  The response I got back was instead to come here.

**Describe the solution you'd like**
A "Support" link that takes someone to a page that provides the email link, security link, and a link here (plus discord or IRL if relevant) with descriptions of when to use each.

**Describe alternatives you've considered**
Link directly to here from the main page.  I figured there are enough channels, that enumerating them would be better and it would provide context that support, and not just bugs, go here (if I understood my email correctly).

PR #3791, opened 2021-07-19
Add a link to report an Issue

hi-rustin:
close https://github.com/rust-lang/crates.io/issues/3129

### DIFF
--- a/app/components/footer.hbs
+++ b/app/components/footer.hbs
@@ -6,9 +6,9 @@
     <span local-class="sep">|</span>
     <a href='mailto:help@crates.io'>Send us an email</a>
     <span local-class="sep">|</span>
-    <a href='https://www.rust-lang.org/policies/security'>Report a security issue</a>
+    <a href='https://www.rust-lang.org/policies/security'>Report a general issue</a>
     <span local-class="sep">|</span>
-    <a href='https://github.com/rust-lang/crates.io/issues/new/choose'>Report an issue</a>
+    <a href='https://github.com/rust-lang/crates.io/issues/new/choose'>Report a security issue</a>
     <span local-class="sep">|</span>
     <a href='https://github.com/rust-lang/crates.io'>Fork on GitHub</a>
     <span local-class="sep">|</span>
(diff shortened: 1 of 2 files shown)


hi-rustin:
image

I'm not sure we should force a line break?

app/components/footer.hbs
```diff
@@ -8,10 +8,12 @@
     <span local-class="sep">|</span>
     <a href='https://www.rust-lang.org/policies/security'>Report a security issue</a>
     <span local-class="sep">|</span>
+    <a href='https://github.com/rust-lang/crates.io/issues/new/choose'>Report an issue</a>
```
JohnTitor, line 11: I couldn't attend the meeting so I have no context, but I think it's clearer to say "Report a general issue", considering we have "Report a security issue". And I'd prefer to place the security link next to the general link.

hi-rustin: Make sense. Addressed. Thanks for your review!

JohnTitor (approved):
The change itself looks good, though I'm going to defer the final review to another reviewer as I'm not sure if we need a style change.

hi-rustin:
PC:
image

mobile:
image

hi-rustin:
r? @jtgeibel

Could you please take a look? Thanks!

hi-rustin:
@Turbo87 

Could you please take a look? Thanks!

Turbo87 (approved):
probably not the final state of the footer, but I guess this is a good intermediate step :)

epage:
As a user, "report an issue" sounds definitive, that I know there is an issue.  I don't remember what issue led to #3129 but I had written "I was trying to get support with a potential problem with crates.io".  The new link would not have helped me in this case.  I probably would still click "send an email" because (1) it sounds closest to a means of support and (2) there is no clarification on when emailing should be used.

hi-rustin:
> As a user, "report an issue" sounds definitive, that I know there is an issue. I don't remember what issue led to #3129 but I had written "I was trying to get support with a potential problem with crates.io". The new link would not have helped me in this case. I probably would still click "send an email" because (1) it sounds closest to a means of support and (2) there is no clarification on when emailing should be used.

The link I added here is `report a general issue
`, maybe we do need a link that explains how to use them. This is not the final footer state, maybe we can consider it in a future design?

Also, I added it just thinking we could jump to here via a direct link. It's also what you mentioned in the possible alternative solutions.

PR #3793, opened 2021-07-21
Avoid logging request_id for successful download redirects

jtgeibel:
The request ID is provided to the user in a few error cases to help us
locate the log line. However, the ID is not provided to the user for
successful requests in general, and in particular for download requests
that make up the vast majority of our traffic.

Removing this from the logs will save 48 bytes in log traffic per
download, and should reduce our log output by about 15%.

r? @pietroalbini

### DIFF
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -206,7 +206,19 @@ impl Display for RequestLine<'_> {
         line.add_field("at", at)?;
         line.add_field("method", self.req.method())?;
         line.add_quoted_field("path", FullPath(self.req))?;
-        line.add_field("request_id", request_header(self.req, "x-request-id"))?;
+
+        // The request_id is not logged for successful download requests
+        if !(self.req.path().ends_with("/download")
+            && self
+                .res
+                .as_ref()
+                .ok()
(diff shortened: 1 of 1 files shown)


src/middleware/log_request.rs
```diff
@@ -206,7 +206,19 @@ impl Display for RequestLine<'_> {
         line.add_field("at", at)?;
         line.add_field("method", self.req.method())?;
         line.add_quoted_field("path", FullPath(self.req))?;
-        line.add_field("request_id", request_header(self.req, "x-request-id"))?;
+
+        // The request_id is not logged for successful download requests
+        if !(self.req.path().ends_with("/download")
```
pietroalbini, line 211: Hmm, instead of special-casing here, could we set a boolean to true in the download endpoint to say "make this log line shorter"?

pietroalbini (approved):
Looks good. Left a comment but if you want to land this and followup with the fixed for my comment later r=me.

jtgeibel:
> Hmm, instead of special-casing here, could we set a boolean to true in the download endpoint to say "make this log line shorter"?

That seems like a good idea, but for now I'd like to get this merged and deployed.

@bors r=pietroalbini

PR #3829, opened 2021-08-15
Track used database conns with an histogram

pietroalbini:
While looking at the metrics, me and @jtgeibel noticed that the way we tracked used database connections before (sampling the current state every time metrics are scraped) is not really accurate, as the connections could be all used up until the scrape happens. This PR adds another metric recording the used database connections with an histogram.

This PR also refactors how histograms are defined in the source code, to account for different histograms needing different buckets.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -13,6 +13,7 @@ use crate::middleware::app::RequestApp;
 pub enum DieselPool {
     Pool {
         pool: r2d2::Pool<ConnectionManager<PgConnection>>,
+        used_conns_metric: Histogram,
         time_to_obtain_connection_metric: Histogram,
     },
     Test(Arc<ReentrantMutex<PgConnection>>),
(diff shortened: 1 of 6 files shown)


src/db.rs
```diff
@@ -62,8 +65,13 @@ impl DieselPool {
         match self {
             DieselPool::Pool {
                 pool,
+                used_conns_metric,
                 time_to_obtain_connection_metric,
             } => time_to_obtain_connection_metric.observe_closure_duration(|| {
+                // Record the number of used connections before obtaining the current one.
+                let state = pool.state();
```
Turbo87, line 72: I assume accessing `pool.state()` is an inexpensive operation?

Turbo87 (approved):
looks good from what I can tell. at least I didn't see any obvious flaws :D

feel free to r=me

src/db.rs
jtgeibel: This just requires obtaining a mutex lock, so shouldn't be expensive.

jtgeibel:
LGTM too!

@bors r=Turbo87

src/db.rs
Turbo87: > requires obtaining a mutex lock

sounds pretty expensive to me 😅

PR #3830, opened 2021-08-15
Fix squash message warning.

ehuss:
https://github.com/rust-lang/rust/pull/87671 added a warning when backslash continuations eat multiple lines. This uncovered that the message for squashing the index was missing some blank lines. This adds some newlines to silence the warning and fix the message formatting.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -398,11 +398,8 @@ pub fn squash_index(env: &Environment) -> Result<(), PerformError> {
     let now = Utc::now().format("%Y-%m-%d");
     let original_head = repo.head_oid()?.to_string();
     let msg = format!("Collapse index into one commit\n\n\
-
         Previous HEAD was {}, now on the `snapshot-{}` branch\n\n\
-
         More information about this change can be found [online] and on [this issue].\n\n\
-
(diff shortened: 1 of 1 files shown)


src/git.rs
```diff
@@ -398,11 +398,11 @@ pub fn squash_index(env: &Environment) -> Result<(), PerformError> {
...
     let original_head = repo.head_oid()?.to_string();
     let msg = format!("Collapse index into one commit\n\n\
-
+        \n\
         Previous HEAD was {}, now on the `snapshot-{}` branch\n\n\
-
+        \n\
         More information about this change can be found [online] and on [this issue].\n\n\
-
+        \n\
```
JohnTitor, line 405: We could just remove these lines instead of inserting `\n`, couldn't we? This adds one more blank line so the total number will be 2. I think 1 is fine here.

ehuss: Oh, oops!  I didn't notice there were already two newlines.  I have pushed an update that just removes the blank lines. It's maybe not as clear how it is formatted, though I'm not sure it really matters.

JohnTitor (approved):
We _might_ want to format like

```rust
"line1\n\
\n\
line2\n\
..."
```

instead, but I don't feel like it's a big deal.

@jtgeibel I'm going to r+ but feel free to tweak it like above if you'd like.

PR #3835, opened 2021-08-18
Add deployment instructions

Turbo87:
... and a small `create-diff-link.sh` script to create a GitHub Diff link.

I'm reasonably confident about most of the content, except the rollbacks, since I've not done that so far. Please let me know if something needs to be changed/adjusted.

PR #3837, opened 2021-08-18
Add `create-diff-link.sh` script 

Turbo87:
Replacement for https://github.com/rust-lang/crates.io/pull/3835 and symbiotic with https://github.com/rust-lang/crates-io-ops-guide/pull/4 :)

### DIFF
--- a/script/create-diff-link.sh
+++ b/script/create-diff-link.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "fetching updates from production app…"
+git fetch heroku-prod
+
+prod_sha=`git rev-parse --short heroku-prod/master`
+master_sha=`git rev-parse --short master`
+
+echo ""
+echo "production app is at: ${prod_sha}"
+echo "local \`master\` branch is at: ${master_sha}"
+echo ""
+echo "https://github.com/rust-lang/crates.io/compare/${prod_sha}...${master_sha}"


docs/DEPLOY.md
```diff
@@ -0,0 +1,3 @@
+# Deployment Instructions
+
+see <https://ops-guide.crates.io/>
```
pietroalbini, line 3: I'm not sure this file is really needed.

Turbo87: I don't know. I think the link to that ops guide should be *somewhere* in the repo, but I'm open to alternative locations.

hi-rustin: It seems that this document requires authorization to access? Maybe we need to specify that only team members can access it?

script/create-diff-link.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "fetching updates from production app…"
+git fetch heroku-prod
+
+prod_sha=`git rev-parse --short heroku-prod/master`
```
jtgeibel, line 8: One other thing to note is that if code is pushed via non-git means, then the fetched branch doesn't reflect the actual commit that is deployed. I've seen this before on staging with deploys done by the ops bot. On production that shouldn't be an issue, but on the rare occasion that we rollback a deploy, the remote repo state and deployed state will diverge.

Also, I wonder if we should parse `heroku-prod/HEAD`, as at some point we may switch to a `main` branch which Heroku now [supports](https://help.heroku.com/O0EXQZTA/how-do-i-switch-branches-from-master-to-main). 

Alternatively, we could extract the deployed commit from the server by fetching `https://crates.io/api/v1/site_metadata`. This should always reflect the latest deployed commit hash.

docs/DEPLOY.md
Turbo87: @hi-rustin yep, that's why I put "private" in the PR title 😅

script/create-diff-link.sh
Turbo87: > I've seen this before on staging with deploys done by the ops bot

does that bot not use git to deploy?

> Also, I wonder if we should parse `heroku-prod/HEAD`

let me try if that works, but generally sounds reasonable

> Alternatively, we could extract the deployed commit from the server by fetching https://crates.io/api/v1/site_metadata. This should always reflect the latest deployed commit hash.

seems like that would require something like https://stedolan.github.io/jq/ and I was hoping to avoid external dependencies for this script to keep it simple :)

Turbo87:
```
$ git rev-parse --short heroku-prod/HEAD
fatal: Needed a single revision

$ git rev-parse --short heroku-prod/master
86ded9db0
```

unfortunately it doesn't :-/

docs/DEPLOY.md
hi-rustin: Oh, sorry! I didn't notice the `private`.

script/create-diff-link.sh
jtgeibel: > does that bot not use git to deploy?

I'm not too familiar with that codebase, but this looks like the [relevant logic](https://github.com/rust-lang/crates-io-ops-bot/blob/40ff04ed705ecd3ee674f880ebb59b686365e1f3/src/commands/heroku.rs#L400-L410). I believe it makes an API request to Heroku which then fetches from GitHub. From what I recall, that method doesn't update the master branch of Heroku's version of the repo.

> let me try if that works, but generally sounds reasonable
> unfortunately it doesn't :-/

It looks like `git remote set-head heroku-prod -a` would lookup the remote `HEAD` and cache it as `heroku-prod/HEAD`, however I think that is a one-time operation and would need to either be run every time by the script, or every user would have to remember to run it if we change to a different `HEAD` branch. Therefore, it seems fine to me to just reference `master` directly and update the script if we ever rename the branch to `main`.

> seems like that would require something like https://stedolan.github.io/jq/ and I was hoping to avoid external dependencies for this script to keep it simple :)

I agree that it doesn't seem worth adding a dev dependency for this. Also, given that there are only a few of us who will run this script, I think we can remember to tweak the generated URL as necessary if we're end up rolling back a deployment.

This looks good to me, but I'll defer to @pietroalbini for the final r+.

Turbo87: > From what I recall, that method doesn't update the master branch of Heroku's version of the repo.

I see, thanks :)

docs/DEPLOY.md
Turbo87: removed the file

Issue #3838, opened 2021-08-18
crates.io fails to load on Edge for Android


Turbo87:
**Describe the bug**

![image](https://user-images.githubusercontent.com/141300/129955277-83609a6c-b163-4f83-bf88-0fd8e50f6f78.png)

see https://twitter.com/zkat__/status/1425654525730373637

**To Reproduce**
1. Install "Edge" on an Android device
2. Visit crates.io

**Expected behavior**

It should show the regular content

**Smartphone**
 - Device: OnePlus 8
 - OS: Android 11
 - Browser: Edge 46.06.4.5161

Turbo87:
I've debugged this using the instructions at https://docs.microsoft.com/en/microsoft-edge/devtools-guide-chromium/remote-debugging/ resulting in a console error:

> Uncaught SyntaxError: Unexpected token '?'

with the stacktrace leading to:

<img width="448" alt="Bildschirmfoto 2021-08-18 um 20 53 53" src="https://user-images.githubusercontent.com/141300/129955723-429143d4-3b00-44ad-837e-0e6a0b580428.png">

It looks like Edge for Android is currently using an old JS engine, which does not support the (Nullish Coalescing Operator)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator] yet, and our [targets.js](config/targets.js) file allows the operator to pass through without transpilation.

Turbo87:
somewhat related and making the fix for this unnecessarily complicated:

- https://github.com/Fyrd/caniuse/issues/5473
- https://github.com/Fyrd/caniuse/issues/5492
- https://github.com/browserslist/browserslist/issues/493

Turbo87:
I can also confirm what is stated in https://github.com/Fyrd/caniuse/issues/5473#issuecomment-850627321. With "Edge Beta for Android" the app loads fine.

PR #3839, opened 2021-08-18
Fix "Edge for Android" compatibility

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/3838 by explicitly adding "Chrome 77" to our browser targets list for now.

/cc @zkat


Turbo87:
I've deployed it to our staging environment and can confirm that it works as intended after this PR :)

PR #3842, opened 2021-08-19
Replace `RustConfig` file with `rust-toolchain`

Turbo87:
see https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

Our heroku buildpack supports that these days (see https://github.com/emk/heroku-buildpack-rust/tree/cfa0f06214#specifying-which-version-of-rust-to-use) and it ensures that local development is happening on the same Rust version that we run in production.


hi-rustin (commented):
I see that it says `rust-toolchain` in the documentation, not `rust-toolchain.toml`. toml format was only supported by rustup on Feb 22. I'm not sure if heroku supports it?

hi-rustin:
It seems that it does not detect the toml file. 

https://github.com/emk/heroku-buildpack-rust/blob/cfa0f0621411f0dbb46df9d4740e119b7a459ccc/bin/detect#L3

https://github.com/emk/heroku-buildpack-rust/blob/cfa0f0621411f0dbb46df9d4740e119b7a459ccc/bin/compile#L39

Turbo87:
you're absolutely right. I've changed it to be a `rust-toolchain` (non-TOML) file instead.

hi-rustin (approved):
LGTM 👍

Turbo87:
I've deployed this to our staging environment and can confirm that `rust-toolchain` (vs. `rust-toolchain.toml`) works as intended :)

jtgeibel:
The opposite of this was originally done in #1647, because Travis CI was picking up the configuration and running all of our builds against the same version. This issue does not appear to affect our GH Actions configuration.

Since I develop and test with `nightly` locally, I can set a [directory override](https://rust-lang.github.io/rustup/overrides.html#directory-overrides) which seems to work as expected. So overall this seems fine to me.

Turbo87:
> The opposite of this was originally done in #1647, because Travis CI was picking up the configuration and running all of our builds against the same version. This issue does not appear to affect our GH Actions configuration.

You're right, that still appears to be an issue, even with GH Actions. The `rustup update` call installs the right version, but then later the `rustc -V` reveals that it runs with 1.53.

I've swapped `rustup default` for `rustup override set` in a second commit now to hopefully fix that issue.

Turbo87:
> I've swapped rustup default for rustup override set in a second commit now to hopefully fix that issue.

yep, that does seem to work :)

pietroalbini:
Do we actually want everyone to pin 1.53.0 in their local development environment?

Turbo87:
@pietroalbini IMHO it makes sense to pin to what we use in production. are there any reasons against doing that?

pietroalbini:
I'd lean towards not adding the restriction if there isn't a reason for doing so: the version used in production is already used in CI (or if it isn't that would be the only place to make this change in). There is already one team member who doesn't use that version (Justin), and I occasionally use different versions like pre-releases as my system default to ensure they work correctly.

Turbo87:
> There is already one team member who doesn't use that version (Justin), and I occasionally use different versions like pre-releases as my system default to ensure they work correctly.

that seems easy to do by using `rustup override set <VERSION>`

> the version used in production is already used in CI (or if it isn't that would be the only place to make this change in)

we're actually not doing that. we're running `stable` and `beta` on CI, but production is running on the pinned version in `RustConfig`, though IMHO we should be running CI on the same version as on production.

> I'd lean towards not adding the restriction if there isn't a reason for doing so

The reason would be that new developers with old Rust versions on their system would automatically use the correct version that we run in production, instead of manually having to track what is in `RustConfig`. And similarly, it ensures that we don't use any features of newer Rust versions that our version in production does not support yet.

I'm not sure where the reluctance to developing on the same version as is running in production comes from. It seems strange to me to not have that by default. 🤔

Turbo87:
closing as discussed in our team meeting yesterday... 😢

PR #3848, opened 2021-08-19
Cache the version_id for download requests

jtgeibel:
This adds a caching layer in the app backend to reduce the number of
database queries. While this query is extremely fast, a massive number
of concurrent download requests can still result in a request backlog.

In this implementation cached values are never dropped. This means
that memory usage will grow over time, however the memory usage is
expected to be well within our instance size and the application is
automatically rebooted at least once every 24 hours. Additionally, on
the rare occasion that a crate version is deleted, the cache will
continue to serve redirects and attempt to record download counts for
that version.

If a version is deleted and continues to see download requests, it is
expected that the entire shard of downloads containing that version_id
will fail to persist. Even if we expired cached entries after a certain
time period, there would still be a window where a deleted version
could cause download counts for other versions to not be persisted. It
is possible we could make the download counter logic robust against this
scenario, but given that deleting versions is rare we can mitigate this
by restarting the app after deleting crates or versions.

r? @pietroalbini

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -36,7 +36,7 @@ pub struct App {
     ///
     /// This is used by the download endpoint to reduce the number of database queries. The
     /// `version_id` is only cached under the canonical spelling of the crate name.
-    pub(crate) version_id_cacher: DashMap<String, i32>,
+    pub(crate) version_id_cacher: DashMap<(String, String), i32>,
 
     /// Count downloads and periodically persist them in the database
     pub downloads_counter: DownloadsCounter,
--- a/src/app.rs
+++ b/src/app.rs
@@ -8,6 +8,7 @@ use crate::downloads_counter::DownloadsCounter;
 use crate::email::Emails;
 use crate::github::GitHubClient;
 use crate::metrics::{InstanceMetrics, ServiceMetrics};
+use dashmap::DashMap;
 use diesel::r2d2;
 use oauth2::basic::BasicClient;
 use reqwest::blocking::Client;
@@ -30,6 +31,12 @@ pub struct App {
 
     /// The server configuration
     pub config: config::Server,
+
+    /// Cache the `version_id` of a `canonical_crate_name:semver` pair
+    ///
+    /// This is used by the download endpoint to reduce the number of database queries. The
+    /// `version_id` is only cached under the canonical spelling of the crate name.
+    pub(crate) version_id_cacher: DashMap<String, i32>,
 
     /// Count downloads and periodically persist them in the database
     pub downloads_counter: DownloadsCounter,
@@ -154,6 +161,7 @@ impl App {
             github,
             github_oauth,
             config,
+            version_id_cacher: DashMap::new(),
             downloads_counter: DownloadsCounter::new(),
             emails: Emails::from_environment(),
             service_metrics: ServiceMetrics::new().expect("could not initialize service metrics"),
--- a/src/controllers/version/downloads.rs
+++ b/src/controllers/version/downloads.rs
@@ -9,6 +9,7 @@ use crate::models::{Crate, VersionDownload};
 use crate::schema::*;
 use crate::views::EncodableVersionDownload;
 use chrono::{Duration, NaiveDate, Utc};
+use dashmap::mapref::entry::Entry;
 
 /// Handles the `GET /crates/:crate_id/:version/download` route.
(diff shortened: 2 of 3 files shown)


src/app.rs
```diff
@@ -32,6 +34,12 @@ pub struct App {
     /// The server configuration
     pub config: config::Server,
 
+    /// Cache the `version_id` of a `canonical_crate_name:semver` pair
+    ///
+    /// This is used by the download endpoint to reduce the number of database queries. The
+    /// `version_id` is only cached under the canonical spelling of the crate name.
+    pub(crate) version_id_cacher: DashMap<String, i32>,
```
pietroalbini, line 41: We can support serving responses for non-canonical spellings with little effort:

```rust
struct CachedVersion {
    canonical_crate_name: String,
    version_id: i32,
}

version_id_cacher: DashMap<String, CachedVersion>
```

Then the cache key could be normalized with `crate_name.lower().replace('_', '-')`, while still having access to the canonical crate name to perform the redirect.

src/controllers/version/downloads.rs
```diff
@@ -19,59 +21,82 @@ pub fn download(req: &mut dyn RequestExt) -> EndpointResult {
...
-                app.instance_metrics
-                    .downloads_non_canonical_crate_name_total
-                    .inc();
-                log_metadata = Some(("bot", "dl"));
-            }
-            crate_name = canonical_crate_name;
+
+    match app
+        .version_id_cacher
+        .entry(format!("{}:{}", crate_name, version))
```
pietroalbini, line 27: `DashMap::entry` acquires a *write* lock, even if the entry is present in the map. This would cause every request to do a write lock on the assigned shard, which we should try to avoid. We should do something similar to [DownloadsCounter](https://github.com/rust-lang/crates.io/blob/4513c507c60bb8ec03fc5d859ceff56264cb2a94/src/downloads_counter.rs#L48-L65), where we first call `.get` to try and use the read lock and only if it fails call `.entry`.

pietroalbini, line 27: We're currently not allowing `:` in crate names so this is fine, but I'm worried we might allow it in the future (if `foo::bar` is chosen as the syntax for some kind of namespacing), and if we forget to update here people could poison the cache. It'd be safer to switch to a separator like the byte zero which we know will never be used.

pietroalbini (changes requested):

src/controllers/version/downloads.rs
jtgeibel: Excellent point, I'll switch to `.get`.

jtgeibel: Sounds good to me.

src/app.rs
jtgeibel: I was originally storing the canonical crate name as well, but ended up dropping that to minimize memory usage, especially since there are so few non-canonical requests.

I'm not actually that worried about memory usage, so this approach does sound reasonable to me as well. Maybe it's worth looking at memory usage in production first, and then adding this if we're sure we have plenty of room to spare.

pietroalbini:
I did some experiments and this should use at most 10-20 MB with the current set of crates, so this should be fine for the foreseeable future for memory usage.

Regarding cache expiration, I'd prefer to actually have some cache expiration built in with a relatively short TTL (1-5 minutes?), to limit the effect of deleted crates. We could just add a `fetched_at` timestamp in the cache value, and if it's older than the TTL then ignore the cached value and fetch from the database again. This shouldn't make the code too more complex and shouldn't use too much memory either.

There should also eventually be some automatic cleanup in place to ensure the map doesn't grow too much, but that can be done in a later PR. Deleting a random shard whenever we reach more than N items in the cache should be enough, without adding LRU logic.

It would also be interesting to have a "cache hit" metric, especially if we implement some TTL.

Turbo87:
I just discovered https://github.com/moka-rs/moka, which sounds like it might be useful for this use case. Would it make sense to use that instead of building our own cache?

Turbo87:
@jtgeibel FYI I've rebased the branch and extracted the test code refactoring into a second (or rather: first) commit

src/controllers/version/downloads.rs
Turbo87: resolved this by using `(String, String)` as the cache key instead

PR #3857, opened 2021-08-22
Revert "Track used database conns with an histogram"

Turbo87:
Reverts rust-lang/crates.io#3829

This has been causing issues in production: 

image


pietroalbini:
Note that the graph is not accurate, the download requests didn't take 5 seconds. Still, this increased the response times, so we should revert it nevertheless.

@bors r+

Issue #3484, opened 2021-04-02
Relative URLs in subdirectory readmes


clehner:
During readme markdown rendering, links with relative URLs in a crate's readme are resolved relative to the repository root, even if the crate and its readme file are in a subdirectory of the repository. This breaks consistency with how relative URLs are resolved in the source site - relative to the file in the subdirectory.

An example of a repository with multiple crates in subdirectories (using absolute URLs in the readmes):
https://github.com/RustCrypto/elliptic-curves/

An example of a crate published from a subdirectory using relative URLs in the readme which are rendered broken:
https://crates.io/crates/ssi-contexts/0.0.2
while in the source code they are working:
https://github.com/spruceid/ssi/blob/ssi-contexts%2F/v0.0.2/contexts/README.md

It would be nice if relative URLs could be used in subdirectory readmes. Possible solutions:
1. Resolve relative to the `homepage` property rather than the `repository` property.
2. Add a `readme_url` metadata field, to use as the base URL for resolution.
3. Add a `repository_path` metadata field that would be the path of the crate in the repository - where the readme is assumed to reside - for use in evaluating relative URLs in the readme.

nipunn1313:
@rustbot claim

I started looking at this. Unfortunately, as @clehner determined, it seems that crates.io doesn't have enough information at time of render to determine the subdirectory that the README is in.

relevant code:
https://github.com/rust-lang/crates.io/blob/d04dc228a7caec656e72ca5c56a11ba128fa04e1/src/controllers/krate/publish.rs#L189

and relevant code where "readme_file" is determined in cargo
https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/registry.rs#L284

> Resolve relative to the homepage property rather than the repository property.

Not sure this helps here - as the homepage isn't necessarily on github/etc showing code.

> Add a readme_url metadata field, to use as the base URL for resolution.

This definitely doesn't seem like something we'd want to expose as something to fill in the Cargo.toml. Seems that cargo should be able to infer it. Perhaps it could be inferred by cargo and added here. Still seems like if it's possible to do this with less complexity, we should see if we can.
https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/registry.rs#L284

> Add a repository_path metadata field that would be the path of the crate in the repository - where the readme is assumed to reside - for use in evaluating relative URLs in the readme.

Again - something we could have cargo infer and send on the network to crates.io


Proposal:

[...]

PR #3861, opened 2021-08-24
Leverage readme_path to support relative paths from readme's not at root

nipunn1313:
Added unit tests to confirm things work.
Manually tested end to end and confirmed that the rendered
page has a properly working relative link as long as the Cargo.toml
is in the root of the repo.

We'll need to update cargo as mentioned in #3484 in order to
support Cargo.toml that aren't in the root of the repo

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -21,7 +21,7 @@ impl<'a> MarkdownRenderer<'a> {
     ///
     /// Per `readme_to_html`, `base_url` is the base URL prepended to any
     /// relative links in the input document.  See that function for more detail.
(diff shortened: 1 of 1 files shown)


nipunn1313:
Here's me doing the best I can to show a screenshot of my manual test.
README is not in root (it's in `docs/README.md`). It has a relative link to `./Otherfile.md` (also in docs)
Prior to this diff, the rendered MD->html would have a bad link. Now it looks good!
image

nipunn1313:
(rebased on master - to handle merge conflicts)

Turbo87 (approved):
LGTM, thanks! 👍

PR #3862, opened 2021-08-24
render: add HTML image tests with align support

MicroJoe:
I was not sure that the `align` properly was supported for images, so I extended the `render.rs` tests to convince me that it indeed worked as intended.

Below is the commit message:

```
render: add HTML image tests with align support

The current align test was only used for text, but we want to be sure
that images can be aligned too. This is why a new test is added to
verify that the align attribute is also kept for images.

Since the align attribute requires RAW HTML for the <img> tag, the
relative_links test has also been updated to check for this usage.
```

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -391,6 +391,7 @@ mod tests {
         let absolute = "[hi](/hi)";
         let relative = "[there](there)";
         let image = "![alt](img.png)";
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):
LGTM thanks :)

PR #3864, opened 2021-08-24
Minor change to postgres instructions

nipunn1313:
I found from following the instructions to the tee, that I needed
postgres to be under my own user (instead of the postgres username).

These new instructions would work for me (on mac). If postgres works differently
on different platforms, perhaps more nuanced instructions would help?


Turbo87:
> These new instructions would work for me (on mac). If postgres works differently
> on different platforms, perhaps more nuanced instructions would help?

hmm, I'm using https://postgresapp.com/ on Mac and the `postgres` user seems to work fine for me, so it definitely depends on how you've installed/set up your postgres database locally. An alternative is using the Docker environment, but if I remember correctly it also needed some adjustments 😅

nipunn1313:
fair! Wouldn't want to break some more common setup's instructions either. Lets close this one then.

PR #3849, opened 2021-08-19
Bump backend dependencies and to latest stable rustc

jtgeibel:
r? @Turbo87

PR #3865, opened 2021-08-24
Clippy fixes

nipunn1313:
This fix came up when running clippy with nightly. (I was using default nightly toolchain on my machine).

I noticed CI runs with stable clippy - and I added a commit to switch back to continually using stable clippy, as the comment seemed to imply that's what you wanted. I'd understand if you'd want to keep it pinned - let me know and I'll revert the change to ci/


nipunn1313:
Spotted #3849 - so I'll remove that from this diff to keep things separate.

Turbo87:
thanks! :)

@bors r+

PR #3867, opened 2021-08-25
Add support for new feature syntax (RFC 3143)

ehuss:
This adds support for the new feature syntax described in [RFC 3143](https://rust-lang.github.io/rfcs/3143-cargo-weak-namespaced-features.html).  There are two new feature values that are allowed in the `[features]` table: weak dependencies (`foo?/feat-name`) and namespaced dependencies (`dep:bar`).  When these style of features are detected, they are placed in the index in a separate field called `features2` to prevent older versions of cargo from breaking on them. Additionally, it sets a new version field called `v` to indicate that this new field is present, which prevents versions starting at 1.51 from selecting those versions.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -98,9 +98,13 @@ pub struct Crate {
     pub deps: Vec<Dependency>,
     pub cksum: String,
     pub features: HashMap<String, Vec<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub features2: Option<HashMap<String, Vec<String>>>,
     pub yanked: Option<bool>,
     #[serde(default)]
     pub links: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub v: Option<u32>,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
--- a/src/git.rs
+++ b/src/git.rs
@@ -98,11 +98,39 @@ pub struct Crate {
     pub deps: Vec<Dependency>,
     pub cksum: String,
     pub features: HashMap<String, Vec<String>>,
+    /// This field contains features with new, extended syntax. Specifically,
+    /// namespaced features (`dep:`) and weak dependencies (`pkg?/feat`).
+    ///
+    /// It is only populated if a feature uses the new syntax. Cargo merges it
+    /// on top of the `features` field when reading the entries.
+    ///
+    /// This is separated from `features` because versions older than 1.19
+    /// will fail to load due to not being able to parse the new syntax, even
+    /// with a `Cargo.lock` file.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub features2: Option<HashMap<String, Vec<String>>>,
     pub yanked: Option<bool>,
     #[serde(default)]
     pub links: Option<String>,
+    /// The schema version for this entry.
+    ///
+    /// If this is None, it defaults to version 1. Entries with unknown
+    /// versions are ignored by cargo starting with 1.51.
+    ///
+    /// Version `2` format adds the `features2` field.
+    ///
+    /// This provides a method to safely introduce changes to index entries
+    /// and allow older versions of cargo to ignore newer entries it doesn't
+    /// understand. This is honored as of 1.51, so unfortunately older
+    /// versions will ignore it, and potentially misinterpret version 2 and
+    /// newer entries.
+    ///
+    /// The intent is that versions older than 1.51 will work with a
+    /// pre-existing `Cargo.lock`, but they may not correctly process `cargo
+    /// update` or build a lock from scratch. In that case, cargo may
+    /// incorrectly select a new package that uses a new index format. A
+    /// workaround is to downgrade any packages that are incompatible with the
+    /// `--precise` flag of `cargo update`.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub v: Option<u32>,
 }
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -297,10 +297,8 @@ impl Crate {
 
     /// Validates a whole feature string, `features = ["THIS", "ALL/THIS"]`.
     pub fn valid_feature(name: &str) -> bool {
-        match name.find('/') {
-            Some(pos) => {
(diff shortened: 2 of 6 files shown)


src/git.rs
```diff
@@ -62,9 +62,41 @@ pub struct Crate {
...
+    /// This provides a method to safely introduce changes to index entries
+    /// and allow older versions of cargo to ignore newer entries it doesn't
+    /// understand. This is honored as of 1.51, so unfortunately older
+    /// versions will ignore it, and potentially misinterpret version 2 and
+    /// newer entries.
+    ///
+    /// The intent is that versions older than 1.51 will work with a
+    /// pre-existing `Cargo.lock`, but they may not correctly process `cargo
+    /// update` or build a lock from scratch. In that case, cargo may
+    /// incorrectly select a new package that uses a new index format. A
```
jtgeibel, line 95: I'm not sure if there is a precedent for this (having editions apply beyond `rustc`), but I wonder if upon publish we should require the crate to be `edition = "2021"` if `features2` is used.

With this requirement we would ensure that if older versions of cargo do select a crate with features it doesn't understand, then the compilation would bail out early due to the new edition, rather than potentially completing successfully but with an unexpected selection of features.

ehuss: Hm, that's an interesting idea.  However, I'm not sure if it will help in enough situations to justify adding the restriction.  If the package required the feature to be set, it would fail during resolution before the package gets downloaded.  If somehow it makes it through resolution, I think at least versions older than 1.27 will fail to parse the manifest.  We are also planning to suggest only using this new syntax in a major version bump which should further limit the impact.  I would kinda prefer to avoid adding more restrictions if possible.  Do you feel strongly about trying to limit the confusion here?

src/models/krate.rs
```diff
@@ -501,6 +504,11 @@ mod tests {
         assert!(Crate::valid_feature("c++20"));
         assert!(Crate::valid_feature("krate/c++20"));
         assert!(!Crate::valid_feature("c++20/wow"));
+        assert!(Crate::valid_feature("foo?/bar"));
+        assert!(Crate::valid_feature("dep:foo"));
+        assert!(!Crate::valid_feature("dep:foo?/bar"));
```
pksunkara, line 509: Why is this not valid? `dep:foo/bar` should also be valid IMO

ehuss: We decided not to allow combining the the two syntaxes to simplify it.  There is more information at https://github.com/rust-lang/cargo/pull/9574 about that change.

djc:
Gentle ping? I think this should be labeled as waiting on review again?

src/models/krate.rs
```diff
@@ -297,12 +297,15 @@ impl Crate {
...
-            && prefix_part.map_or(true, Crate::valid_feature_prefix)
+        match name.find('/') {
+            Some(pos) => {
+                let (dep, dep_feat) = name.split_at(pos);
+                let dep_feat = &dep_feat[1..];
+                let dep = dep.strip_suffix('?').unwrap_or(dep);
+                Crate::valid_feature_prefix(dep) && Crate::valid_feature_name(dep_feat)
+            }
+            None => Crate::valid_feature_name(name.strip_prefix("dep:").unwrap_or(name)),
+        }
```
jhpratt, line 308:
```suggestion
        match name.split_once('/') {
            Some((dep, dep_feat)) => {
                let dep = dep.strip_suffix('?').unwrap_or(dep);
                Crate::valid_feature_prefix(dep) && Crate::valid_feature_name(dep_feat)
            }
            None => Crate::valid_feature_name(name.strip_prefix("dep:").unwrap_or(name)),
        }
```

jhpratt (commented):
I can't attest to the validity of the changes as a whole; I'm not familiar with the internals. The parsing of the feature seems correct. Just one minor nit on style.

Turbo87:
FWIW we discussed this on Discord a few days ago: https://discord.com/channels/442252698964721669/448525639469891595/920823792631152691

the tl;dr is that we show features on the UI these days and we need to verify that this still works correctly with this new syntax

ehuss:
> the tl;dr is that we show `features` on the UI these days and we need to verify that this still works correctly with this new syntax

I'm not sure how you'd like to verify that.  I could add something that did a database query in the `features_version_2` test if you'd like.

The feature values themselves don't seem to be displayed in the UI:

image

I can confirm from the API that it is there:

```
curl http://localhost:4200/api/v1/crates/foo | jq .versions[0].features
{
  "f1": [
    "dep:bar"
  ],
  "f2": [
    "bar?/bar-feat"
  ]
}
```

And via psql:

```
cargo_registry=# select * from versions;
 id | crate_id |  num  |         updated_at         |         created_at         | downloads |                   features                   | yanked | license | crate_size | published_by
----+----------+-------+----------------------------+----------------------------+-----------+----------------------------------------------+--------+---------+------------+--------------
  1 |        1 | 0.1.0 | 2021-12-22 23:49:25.411573 | 2021-12-22 23:49:25.411573 |         0 | {}                                           | f      | MIT     |        702 |            1
  2 |        2 | 0.1.0 | 2021-12-22 23:51:35.513816 | 2021-12-22 23:51:35.513816 |         0 | {"bar-feat": []}                             | f      | MIT     |        719 |            1
  4 |        1 | 0.2.0 | 2021-12-22 23:56:47.374767 | 2021-12-22 23:56:47.374767 |         0 | {"f1": ["dep:bar"], "f2": ["bar?/bar-feat"]} | f      | MIT     |        772 |            1
(3 rows)
```

Turbo87 (approved):
> The feature values themselves don't seem to be displayed in the UI

you're absolutely right. now that I finally had time to take a detailed look at the RFC and PR changes, I agree that the frontend shouldn't be affected by this. since the split between `features` and `features2` only happens for the index, but is otherwise kept in its merged form I don't see any issues with it.

nice work! and sorry for the huge review delay... 😞 

@pietroalbini @jtgeibel unless you have any objections I plan to merge this on 2021-12-30, or earlier if you both signal your approval until then :)

Turbo87:
rebased, merged and deployed. thanks for working on this @ehuss!

PR #3878, opened 2021-08-31
Replace `ember-prism` with `highlight.js`

Turbo87:
[highlight.js](https://highlightjs.org) is easier to integrate with the app (via ember-auto-import) and doesn't require us to use any global variables or an ember-addon at all.

The remaining two `<pre>` usages got a `.terminal` CSS class in this commit to avoid styling conflicts.


hi-rustin (approved):
Tested locally and looks great!

PR #3880, opened 2021-08-31
render: fix center image test

MicroJoe:
The image test is actually not representative and will not center the image. This fix helps users discover how to center images and proves that they will be centered on crates.io.

Note: I am the original author of this testcase from PR #3862

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -534,11 +534,11 @@ mod tests {
     #[test]
     fn image_alignment() {
         let text =
-            "<img src=\"https://img.shields.io/crates/v/clap.svg\" alt=\"\" align=\"center\">\n";
+            "<p align=\"center\"><img src=\"https://img.shields.io/crates/v/clap.svg\" alt=\"\"></p>\n";
         let result = markdown_to_html(text, None);
         assert_eq!(
             result,
-            "<img src=\"https://img.shields.io/crates/v/clap.svg\" alt=\"\" align=\"center\">\n"
+            "<p align=\"center\"><img src=\"https://img.shields.io/crates/v/clap.svg\" alt=\"\"></p>\n"
         );
     }
 }


src/render.rs
```diff
@@ -534,11 +534,11 @@ mod tests {
     #[test]
     fn image_alignment() {
         let text =
-            "<img src=\"https://img.shields.io/crates/v/clap.svg\" alt=\"\" align=\"center\">\n";
+            "<p align=\"center\"><img src=\"https://img.shields.io/crates/v/clap.svg\" alt=\"\"></p>\n";
         let result = markdown_to_html(text, None);
         assert_eq!(
             result,
-            "<img src=\"https://img.shields.io/crates/v/clap.svg\" alt=\"\" align=\"center\">\n"
+            "<p align=\"center\"><img src=\"https://img.shields.io/crates/v/clap.svg\" alt=\"\"></p>\n"
```
Turbo87, line 541: aren't these cases already sufficiently tested by the `text_alignment` test above?

MicroJoe: The `text_alignment` test guarantees that the `align` attribute can be set for `<h1>` and `<h5>` HTML tags. By being explicit in this other test, we make sure that centering images will always be supported by catching any possible future regression. Testing is cheap, finding regressions is hard.

Turbo87: shouldn't it be renamed to `paragraph_alignment` then? I guess the `<p>` doesn't care whether a `` is inside or not 😅

MicroJoe: You are right that the `<p>` does not care currently. But maybe in a future release the `<p>` will care because of implementation changes. The usecase I want to cover is centering images in `README.md` on crates.io, so I make sure that this particular usecase is tested and will continue to work in the future.

It also as a side effect documents how to center images on crates.io for users, which can be searched for with the keywords "center" and "image" in the source code of the project.

PR #3884, opened 2021-09-03
Add `FORCE_UNCONDITIONAL_REDIRECTS`

pietroalbini:
This PR adds a new environment variable, `FORCE_UNCONDITIONAL_REDIRECTS`, which forces unconditional redirects even with a fully healthy database. The environment variable is meant to be used by an on-call operator to reduce the database load when the primary database is overloaded, and should only be set during incidents.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -27,6 +27,7 @@ pub struct Server {
     pub metrics_authorization_token: Option<String>,
     pub use_test_database_pool: bool,
     pub instance_metrics_log_every_seconds: Option<u64>,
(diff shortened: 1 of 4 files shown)


pietroalbini:
I recommend using [this link](https://github.com/rust-lang/crates.io/pull/3884/files?w=1) for the review, as that hides whitespace-only changes (the bulk of the changes, as I had to change the indentation of most of the download endpoint).

Turbo87 (approved):
LGTM 👍 

feel free to r=me

Issue #3882, opened 2021-09-02
`to_semver_no_prerelease` doesn't handle build metadata with hyphens correctly


ghost:
**Describe the bug**
I noticed while messing around with the database that the `to_semver_no_prerelease` function doesn't account for build metadata containing hyphens, e.g. 0.4.45+curl-7.78.0 is not a prerelease version however `to_semver_no_prerelease('0.4.45+curl-7.78.0')` is `NULL`. I verified this with the regex from https://semver.org/ - see https://regex101.com/r/vEYRnS/1.

**To Reproduce**
1. `SELECT to_semver_no_prerelease('0.4.45+curl-7.78.0')` returns `NULL`;

**Expected behavior**
The above query should return `(0,4,45)`.

**Additional context**
This may result in reverse dependencies query returning inaccurate results.

Issue #3480, opened 2021-04-01
Ops: Prepare for postgres upgrade


jtgeibel:
I see no urgency for this migration, but we should start planning for an eventual upgrade. We are currently on PG11 and PG13 is available on Heroku for production use.

Heroku's [`pg:upgrade`](https://devcenter.heroku.com/articles/upgrading-heroku-postgres-databases#upgrading-with-pg-upgrade) serves as a guide. We will want to document our own detailed steps and test them out on staging. In particular, we may want to temporarily setup a real leader/follower configuration on staging. (Currently, both connection pools point to the same database on staging.)

We will also need to make sure the new download tracking logic works well in read-only mode (dropping download counts for the duration of the migration).

To minimize the time we operate in read-only mode, we will probably want to land #3479 first.

Turbo87:
https://github.com/rust-lang/crates.io/pull/3886#issuecomment-913432943 might be a good reason to do this in the near-term future

Turbo87:
we discussed this topic in our weekly team meeting today:

- it would be good to resolve https://github.com/rust-lang/crates.io/issues/3479 before the upgrade, but it is not strictly required. resolving it after the upgrade means that the upgrade might take a little bit longer since more data needs to be migrated.
- it should be possible to create a temporary follower database and upgrade that to get a feel for how long the migration would take, and thus the amount of time we would have to spend in read-only mode.
- @pietroalbini will look into this on the weekend and write down some instructions
- the actual upgrade should happen during a Zoom call with @jtgeibel @pietroalbini and @Turbo87 to immediately address any issues that come up

pietroalbini:
So, I created a throwaway follower and performed the upgrade on it, and it took between 7 and 8 minutes for the upgrade to finish. This would allow us to have a 10 minute read-only window, which is not bad.

The main worry I have is about the read-only follower used for the more heavy queries: when the database is upgraded the old followers are detached, so they can't be used with the new primary. This would mean we'd need to first upgrade the primary from PostgreSQL 11 to PostgreSQL 13, and *then* create a new read-only follower.

[...]

PR #3886, opened 2021-09-03
SQL: Fix metadata handling in `to_semver_no_prerelease()` function

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/3882

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,8 +90,10 @@ tracing-subscriber = "0.2"
 url = "2.1"
 
 [dev-dependencies]
+bigdecimal = { version = ">= 0.0.10, < 0.2.0" }
 claim = "0.5"
 conduit-test = "0.9.0-alpha.4"
+diesel = { version = "1.4.0", features = ["numeric"] }
 hyper-tls = "0.5"
 lazy_static = "1.0"
 tokio = "1.5.0"
--- a/migrations/2021-09-03-132838_fix_to_semver_no_prerelease/up.sql
+++ b/migrations/2021-09-03-132838_fix_to_semver_no_prerelease/up.sql
@@ -1,0 +1,19 @@
+create or replace function to_semver_no_prerelease(text) returns semver_triple
+    immutable
+    language sql
+as
+$$
+SELECT (
+        -- 2) then, we extract the major, minor and patch numbers
+        --    (dropping the prerelease part to avoid number conversion errors)
+        split_part(version_without_metadata, '.', 1)::numeric,
+        split_part(version_without_metadata, '.', 2)::numeric,
+        split_part(split_part(version_without_metadata, '-', 1), '.', 3)::numeric
+           )::semver_triple
+FROM (
+    -- 1) first, we split off the release metadata, if it exists
+    SELECT split_part($1, '+', 1) as version_without_metadata
+    ) as vwm
+-- 3) finally, we only return the result if there is no prerelease part
+WHERE strpos(version_without_metadata, '-') = 0
+$$;
--- a/src/tests/sql/custom_functions.rs
+++ b/src/tests/sql/custom_functions.rs
@@ -32,4 +32,7 @@ fn to_semver_no_prerelease_works() {
     test(&conn, "1.2.4", Some((1, 2, 4)));
     test(&conn, "1.2.4+metadata", Some((1, 2, 4)));
     test(&conn, "1.2.4-beta.3", None);
+
+    // see https://github.com/rust-lang/crates.io/issues/3882
+    test(&conn, "0.4.45+curl-7.78.0", Some((0, 4, 45)));
 }
--- a/src/tests/sql/custom_functions.rs
+++ b/src/tests/sql/custom_functions.rs
@@ -35,4 +35,5 @@ fn to_semver_no_prerelease_works() {
 
     // see https://github.com/rust-lang/crates.io/issues/3882
     test(&conn, "0.4.45+curl-7.78.0", Some((0, 4, 45)));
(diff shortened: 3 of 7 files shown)


Cargo.toml
```diff
@@ -88,8 +88,10 @@ tracing-subscriber = "0.2"
 url = "2.1"
 
 [dev-dependencies]
+bigdecimal = { version = ">= 0.0.10, < 0.2.0" }
```
Turbo87, line 91: unfortunately `diesel` does not reexport the version that they depend on, so we have to add this here ourselves

Cargo.toml
```diff
@@ -88,8 +88,10 @@ tracing-subscriber = "0.2"
...
 claim = "0.5"
 conduit-test = "0.9.0-alpha.4"
+diesel = { version = "1.4.0", features = ["numeric"] }
```
Turbo87, line 94: the `numeric` feature is necessary because the `semver_triple` returned by `to_semver_no_prerelease()` is based on the `Numeric` SQL type.

migrations/2021-09-03-132838_fix_to_semver_no_prerelease/up.sql
```diff
@@ -0,0 +1,19 @@
...
+    immutable
+    language sql
+as
+$$
+SELECT (
+        -- 2) then, we extract the major, minor and patch numbers
+        --    (dropping the prerelease part to avoid number conversion errors)
+        split_part(version_without_metadata, '.', 1)::numeric,
+        split_part(version_without_metadata, '.', 2)::numeric,
+        split_part(split_part(version_without_metadata, '-', 1), '.', 3)::numeric
```
mikerite, line 11: The split is not necessary since the `WHERE` clause ensures that there is no '-'. Also the order in the comments is wrong because `WHERE` clauses are evaluated because `SELECT`s.

src/tests/sql/custom_functions.rs
```diff
@@ -0,0 +1,38 @@
...
+            .unwrap();
+
+        let expected = expected.map(|it| (it.0.into(), it.1.into(), it.2.into()));
+        assert_eq!(result, expected);
+    }
+
+    test(&conn, "0.0.0", Some((0, 0, 0)));
+    test(&conn, "1.2.4", Some((1, 2, 4)));
+    test(&conn, "1.2.4+metadata", Some((1, 2, 4)));
+    test(&conn, "1.2.4-beta.3", None);
```
mikerite, line 34: Also an example where '-' is before '+' e.g. `0.1.4-preview+4.3.2` .

migrations/2021-09-03-132838_fix_to_semver_no_prerelease/up.sql
Turbo87: if you try running the test suite without it you will see that it fails. it looks like the `split_part()` is evaluated even if there is a `-` in the version, and that will then cause a number conversion error because for e.g. `1.2.3-beta.4` it can't convert `3-beta.4` into a `Numeric`. at least that's what I observed when I tested this locally.

src/tests/sql/custom_functions.rs
Turbo87: added 👍

pietroalbini:
While the changes here look correct, my main worry is the performance of the already slow reverse dependencies endpoint. Trying to run the old and the new function in a local database dump shows a 16% slowdown:

```
cratesio-prod=# explain (analyze, buffers) select num from versions;
                                                   QUERY PLAN                                                   
----------------------------------------------------------------------------------------------------------------
 Seq Scan on versions  (cost=0.00..9824.83 rows=367883 width=6) (actual time=0.053..51.207 rows=367883 loops=1)
   Buffers: shared hit=128 read=6018
 Planning time: 0.434 ms
 Execution time: 67.512 ms
(4 rows)

cratesio-prod=# explain (analyze, buffers) select to_semver_no_prerelease(num) from versions;
                                                     QUERY PLAN                                                      
---------------------------------------------------------------------------------------------------------------------
 Seq Scan on versions  (cost=0.00..101795.58 rows=367883 width=32) (actual time=0.241..1983.644 rows=367883 loops=1)
   Buffers: shared hit=166 read=5986
 Planning time: 0.419 ms
 Execution time: 2000.023 ms
(4 rows)

cratesio-prod=# explain (analyze, buffers) select to_semver_no_prerelease_new(num) from versions;
                                                     QUERY PLAN                                                      
---------------------------------------------------------------------------------------------------------------------
 Seq Scan on versions  (cost=0.00..101795.58 rows=367883 width=32) (actual time=0.247..2382.449 rows=367883 loops=1)
   Buffers: shared hit=192 read=5954
 Planning time: 0.198 ms
 Execution time: 2399.200 ms
(4 rows)

cratesio-prod=# 
```

I'd like to hear @jtgeibel's opinion here too.

Turbo87:
we could consider using https://www.postgresql.org/docs/current/ddl-generated-columns.html to have a generated (and stored) `version_no_prerelease` column in the `versions` table that the rev deps query can use. that way we only run the function when we insert a new version, but not all the time in the rev deps query.

the only downside of this is that generated columns were only introduced in Postgres 12 and apparently we're still using v11 in production.

I think that a 16% slowdown for a correctness bug is worth it though, and with the suggestion above we should have a viable way forward to make it much faster than before too.

Turbo87:
in our weekly team meeting today we decided to delay this bugfix until the performance concerns are addressed, which likely requires us to resolve https://github.com/rust-lang/crates.io/issues/3480 first.

Turbo87:
closing in favor of https://github.com/rust-lang/crates.io/pull/4022

Issue #1386, opened 2018-05-04
Failing to acquire a database connection should return 503 not 500


sgrif:
Our database server went down for maintenance briefly today, which resulted in 100% of requests returning a 500 status code. While most of these requests shouldn't have completed successfully, we should at least give the semantically correct error code.

PR #3888, opened 2021-09-03
Cast `PoolError::UnhealthyPool` into `503 Service Unavailable` error

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/1386

I'm not sure how to write a proper test for this unfortunately, but I've manually verified that this works as intended.

### DIFF
--- a/src/tests/unhealthy_database.rs
+++ b/src/tests/unhealthy_database.rs
@@ -2,6 +2,7 @@ use crate::{
     builders::CrateBuilder,
     util::{MockAnonymousUser, RequestHelper, TestApp},
 };
+use http::StatusCode;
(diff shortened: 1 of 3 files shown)


pietroalbini:
You can add a test like https://github.com/rust-lang/crates.io/blob/c00bc1ca73cac7fda0d58156e7ba3e0169e59fb1/src/tests/unhealthy_database.rs, enabling the real database during initialization and using chaosproxy to break the networking between the application and the server.

Turbo87:
@pietroalbini thanks for the hint. I've added a test 👍

pietroalbini (approved):

PR #3889, opened 2021-09-03
Use `json!` macro to simplify JSON serialization code

Turbo87:
see https://docs.serde.rs/serde_json/macro.json.html

I guess `+132 −349` speaks for itself 😅

### DIFF
--- a/src/controllers/keyword.rs
+++ b/src/controllers/keyword.rs
@@ -24,22 +24,15 @@ pub fn index(req: &mut dyn RequestExt) -> EndpointResult {
     let conn = req.db_read_only()?;
     let data: Paginated<Keyword> = query.load(&*conn)?;
     let total = data.total();
-    let kws = data.into_iter().map(Keyword::into).collect::<Vec<_>>();
+    let kws = data
+        .into_iter()
+        .map(Keyword::into)
+        .collect::<Vec<EncodableKeyword>>();
 
-    #[derive(Serialize)]
-    struct R {
-        keywords: Vec<EncodableKeyword>,
-        meta: Meta,
-    }
-    #[derive(Serialize)]
-    struct Meta {
-        total: Option<i64>,
-    }
-
-    Ok(req.json(&R {
-        keywords: kws,
-        meta: Meta { total: Some(total) },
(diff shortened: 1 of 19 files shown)


hi-rustin (commented):
It looks as simple as returning json in JavaScript!

src/controllers/keyword.rs
```diff
@@ -24,22 +24,15 @@ pub fn index(req: &mut dyn RequestExt) -> EndpointResult {
...
-    let kws = data.into_iter().map(Keyword::into).collect::<Vec<_>>();
-
-    #[derive(Serialize)]
-    struct R {
-        keywords: Vec<EncodableKeyword>,
-        meta: Meta,
-    }
-    #[derive(Serialize)]
-    struct Meta {
-        total: Option<i64>,
```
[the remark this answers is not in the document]
Turbo87: There was only a single codepath that filled the `Meta` struct, and that codepath always used `Some(...)`. `Some(...)` just serializes to the value within it, so it seemed a bit pointless to keep it as an `Option` if `None` is never used anywhere.

In other words: whether we use `Some(total)` or `total` results in the same response, so I preferred the simple variant.

src/controllers/krate/search.rs
```diff
@@ -338,28 +338,16 @@ pub fn search(req: &mut dyn RequestExt) -> EndpointResult {
...
-    Ok(req.json(&R {
-        crates,
-        meta: Meta {
-            total: Some(total),
-            next_page,
-            prev_page,
+    Ok(req.json(&json!({
+        "crates": crates,
+        "meta": {
+            "total": Some(total),
```
hi-rustin, line 346: @Turbo87 So, do we need to simplify here? It also looks like it will always be Some().

Turbo87: good catch! :)

we don't necessarily *need* to simplify it, but I guess we should. I've rebased and fixed it in the process.

hi-rustin (approved):
LGTM 👍

hi-rustin:
If there are no other objections, please feel free to r=me.

PR #3892, opened 2021-09-06
Add `BLOCKED_ROUTES` environment variable

pietroalbini:
This PR adds the `BLOCKED_ROUTES` environment variable, which allows to block individual routes through the environment. This is going to be useful during incident response if we need to block an individual route (for example if the route is degrading the performance of the database), as it would remove the need to deploy an hotfix to disable that route.

r? @jtgeibel

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,6 +6,7 @@ mod database_pools;
 
 pub use self::base::Base;
 pub use self::database_pools::DatabasePools;
+use std::collections::HashSet;
 
 pub struct Server {
     pub base: Base,
@@ -28,6 +29,7 @@ pub struct Server {
     pub use_test_database_pool: bool,
     pub instance_metrics_log_every_seconds: Option<u64>,
     pub force_unconditional_redirects: bool,
+    pub blocked_routes: HashSet<String>,
 }
 
 impl Default for Server {
@@ -58,6 +60,8 @@ impl Default for Server {
     ///   If the environment variable is not present instance metrics are not logged.
     /// - `FORCE_UNCONDITIONAL_REDIRECTS`: Whether to force unconditional redirects in the download
     ///   endpoint even with a healthy database pool.
+    /// - `BLOCKED_ROUTES`: A comma separated list of HTTP routes that are manually blocked by an
+    ///   operator.
     ///
     /// # Panics
     ///
@@ -100,6 +104,9 @@ impl Default for Server {
             use_test_database_pool: false,
             instance_metrics_log_every_seconds: env_optional("INSTANCE_METRICS_LOG_EVERY_SECONDS"),
             force_unconditional_redirects: dotenv::var("FORCE_UNCONDITIONAL_REDIRECTS").is_ok(),
+            blocked_routes: env_optional("BLOCKED_ROUTES")
+                .map(|routes: String| routes.split(',').map(|s| s.into()).collect())
+                .unwrap_or_else(HashSet::new),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/router.rs
```diff
@@ -150,6 +151,21 @@ struct C(pub fn(&mut dyn RequestExt) -> EndpointResult);
...
+            if req.app().config.blocked_routes.contains(pattern.pattern()) {
+                let body = "This route is temporarily blocked. See https://status.crates.io";
+
+                return Response::builder()
+                    .status(StatusCode::SERVICE_UNAVAILABLE)
+                    .header(header::CONTENT_LENGTH, body.len())
+                    .body(Body::from_vec(body.as_bytes().to_vec()))
+                    .map_err(box_error);
+            }
+        }
```
Turbo87, line 167: might make sense to check if we have any blocked routes before doing the pattern lookup?

```suggestion
        // Allow blocking individual routes by their pattern through the `BLOCKED_ROUTES`
        // environment variable. This is not in a middleware because we need access to
        // `RoutePattern` before executing the response handler.
        let blocked_routes = req.app().config.blocked_routes;
        if !blocked_routes.is_empty() {
            if let Some(pattern) = req.extensions().find::<RoutePattern>() {
                if blocked_routes.contains(pattern.pattern()) {
                    let body = "This route is temporarily blocked. See https://status.crates.io";

                    return Response::builder()
                        .status(StatusCode::SERVICE_UNAVAILABLE)
                        .header(header::CONTENT_LENGTH, body.len())
                        .body(Body::from_vec(body.as_bytes().to_vec()))
                        .map_err(box_error);
                }
            }
        }
```

src/config.rs
```diff
@@ -58,6 +60,8 @@ impl Default for Server {
     ///   If the environment variable is not present instance metrics are not logged.
     /// - `FORCE_UNCONDITIONAL_REDIRECTS`: Whether to force unconditional redirects in the download
     ///   endpoint even with a healthy database pool.
+    /// - `BLOCKED_ROUTES`: A comma separated list of HTTP routes that are manually blocked by an
+    ///   operator.
```
Turbo87, line 64:
```suggestion
    /// - `BLOCKED_ROUTES`: A comma separated list of HTTP route patterns that are manually blocked by an
    ///   operator (e.g. `/crates/:crate_id/:version/download`).
```

src/router.rs
```diff
@@ -150,6 +151,21 @@ struct C(pub fn(&mut dyn RequestExt) -> EndpointResult);
...
+        // environment variable. This is not in a middleware because we need access to
+        // `RoutePattern` before executing the response handler.
+        if let Some(pattern) = req.extensions().find::<RoutePattern>() {
+            if req.app().config.blocked_routes.contains(pattern.pattern()) {
+                let body = "This route is temporarily blocked. See https://status.crates.io";
+
+                return Response::builder()
+                    .status(StatusCode::SERVICE_UNAVAILABLE)
+                    .header(header::CONTENT_LENGTH, body.len())
+                    .body(Body::from_vec(body.as_bytes().to_vec()))
```
jtgeibel, line 164: Should we use the standard json error builder instead of outputting raw text? I think frontend endpoints should show a better error message (at least, for the endpoints where we attempt to extract that).

src/router.rs
pietroalbini: Getting the pattern is a hashmap lookup, I don't think it'll have any practical performance implication.

src/router.rs
pietroalbini: Good point, fixed!

pietroalbini:
Addressed review feedback.

src/router.rs
Turbo87: okay, just wanted to check, since I wasn't sure how expensive those calls were :)

Turbo87 (approved):

PR #3893, opened 2021-09-06
Cargo: Use v2 resolver

Turbo87:
see https://doc.rust-lang.org/cargo/reference/features.html#feature-resolver-version-2

It doesn't make a difference for us right now, but I guess we should move to the new resolver instead of staying behind on the old one for no reason :)


hi-rustin (commented):
Do we need to modify the s3's Cargo.toml?

Turbo87:
@hi-rustin good point. I've added it there too.

hi-rustin (approved):

PR #3894, opened 2021-09-06
Use a common prefix for logged errors

jtgeibel:
The `at=error ` prefix is used for Heroku error messages and using a common prefix aids in metrics collection and is easily searchable. This is the first step in monitoring logged errors discussed further in rust-lang/crates-io-heroku-metrics#3.

r? @pietroalbini


src/git.rs
```diff
@@ -272,7 +275,12 @@ impl Repository {
         let head = self.head_oid()?;
 
         if head != original_head {
-            println!("Resetting index from {} to {}", original_head, head);
+            // This isn't strictly an error, but treating it as one for monitoring as it indicates
+            // that an error has occurred while processing the previous index operation.
+            eprintln!(
+                "at=error mod=git error=\"Resetting index from {} to {}\"",
+                original_head, head
+            );
```
Turbo87, line 283: I'm wondering if we should use `error!(...)` instead, and use a custom `tracing` formatter to produce this output. `tracing` seems to explicitly be made for this kind of structured logging.

Turbo87:
we switched over to using `tracing` since this PR was opened and we tried out https://github.com/EmbarkStudios/tracing-logfmt, but since we're currently not ingesting the logs in a machine readable way there was no benefit compared to the default tracing formatter.

given all that, I guess we can close this PR? let me know if you would like to reopen and still land this :)

PR #3895, opened 2021-09-06
errors: Derive `Error` (and `Display`) for `AppErrToStdErr`

Turbo87:
Instead of implementing `Error` and `Display` ourselves we can rely on https://github.com/dtolnay/thiserror to derive them for us.


hi-rustin (approved):
LGTM

PR #3896, opened 2021-09-07
Update documentation on krate::downloads controller

dswij:
Fix some typo and inaccurate documentation where version specific download counts should be in `version::downloads` instead of `krate::downloads` (which is the current file)


Turbo87:
good catch, thanks @dswij! 👍 

@bors r+

PR #3898, opened 2021-09-07
Split `LogRequests` middleware into multiple middlewares

Turbo87:
The `LogRequests` middleware is currently performing multiple tasks:

- Save `req.path()` before the `NormalizePath` middleware potentially mutates the request path
- Calculate response timing based on the `x-request-start` HTTP header from Heroku (or the `StartInstant` extension as a fallback)
- Log requests to stdout
- Report errors to Sentry

The Sentry error reporting was originally built into the `LogRequests` middleware because it also needed access to the original path and the response timing.

This PR proposes to split the `LogRequests` middleware, so that each middlewares is performing only one of the functions respectively:

- The "original path" code is moved into the `NormalizePath` middleware that is responsible for mutating the path in the first place
- The request timing code is moved into a new `RequestTiming` middleware, which saves the response time in the request extension map
- The Sentry error reporting code is moved into a new `SentryMiddleware`

### DIFF
--- a/src/middleware.rs
+++ b/src/middleware.rs
@@ -10,6 +10,7 @@ use self::ember_html::EmberHtml;
 use self::head::Head;
 use self::known_error_to_json::KnownErrorToJson;
 use self::log_connection_pool_status::LogConnectionPoolStatus;
+use self::request_timing::RequestTiming;
 use self::static_or_continue::StaticOrContinue;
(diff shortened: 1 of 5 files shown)


Turbo87:
closing in favor of https://github.com/rust-lang/crates.io/pull/4137

PR #3900, opened 2021-09-08
errors: Remove `ChainError` impl for `Option`

Turbo87:
For `Option` there is nothing to chain onto, so we might as well use the built-in `ok_or_else()` method to convert the `Option` to a `Result`.


Turbo87:
closing in favor of https://github.com/rust-lang/crates.io/pull/3901

PR #3902, opened 2021-09-08
Match monospace font with main font

arzg:
Since Fira Sans is used across crates.io, it makes sense to use Fira Mono for monospaced text to increase consistency. After Fira Mono there is now `ui-monospace` in the font stack, which renders as SF Mono on Safari without having to manually install SF Mono. `ui-monospace` [is part of](https://www.w3.org/TR/css-fonts-4/#ui-monospace-def) the CSS Fonts Module Level 4 draft.


Turbo87:
LGTM! 👍 

@bors r+

PR #3911, opened 2021-09-10
renovatebot: Enable `rust` manager, but only with dashboard approval

Turbo87:
After this PR all pending Rust dependency updates should show up on https://github.com/rust-lang/crates.io/issues/2689. They won't be opened automatically as PRs, but they should have a checkbox allowing us to half-manually create them.

This should give us a better overview of what dependency updates are pending, without hammering CI too much.


locks:
@bors r+

Uh, interesting UI from renovatebot!

PR #3914, opened 2021-09-11
Yet more database metrics

pietroalbini:
This PR adds #3829 again with a couple improvements:

* Instead of calling `pool.state()` every time a connection is acquired, which we believe caused the incident that led to the revert, this uses r2d2's [event handling](https://docs.rs/r2d2/0.8.9/r2d2/event/trait.HandleEvent.html) to keep track of the state outside of r2d2's mutex. The overhead of keeping track of the state should be minimal.

* Since this added r2d2 event handling anyway, this also adds a new metric to track how many times obtaining a connection timed out, which will be yet another datapoint operators will have at their disposal to debug incidents.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -128,6 +128,9 @@ impl App {
                 instance_metrics
                     .database_used_conns_histogram
                     .with_label_values(&["primary"]),
+                instance_metrics
+                    .database_timeouts_when_obtaining_connections
+                    .with_label_values(&["primary"]),
             )
             .unwrap()
         };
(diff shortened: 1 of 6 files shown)


Turbo87:
@pietroalbini would you still like to land this? would it be helpful for the infra and/or on-call teams?

pietroalbini:
I won't have time myself, but yes, we should do this.

Turbo87:
closing this since we've moved from r2d2 to https://crates.io/crates/deadpool and that requires substantial changes to the PR

PR #3915, opened 2021-09-12
Don't save `cargo login $token` command to shell history

Shnatsel:
Most unix shells do not save commands starting with a space to history.

Prefixing the suggested `cargo login $token` command with a space makes stealing crates.io credentials a little bit harder.

### DIFF
--- a/app/components/settings/api-tokens.hbs
+++ b/app/components/settings/api-tokens.hbs
@@ -97,7 +97,7 @@
           its value again. For use on the command line you can save it to <code>~/.cargo/credentials</code>
           with:
 
-          <pre class="terminal">cargo login {{token.token}}</pre>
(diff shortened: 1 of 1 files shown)


app/components/settings/api-tokens.hbs
```diff
@@ -97,9 +97,9 @@
           its value again. For use on the command line you can save it to <code>~/.cargo/credentials</code>
           with:
 
-          <pre class="terminal">cargo login {{token.token}}</pre>
+          <pre class="terminal"> cargo login {{token.token}}</pre>
```
Turbo87, line 100: I'm wondering if we should suggest this command at all. Maybe we should only print the token itself and suggest `cargo login` usage without the token argument in the surrounding text. 🤔

locks: I'm on board with that idea!

PR #3918, opened 2021-09-12
renovatebot: Remove `assignees` from pull requests

Turbo87:
`reviewers` is enough, we don't need `assignees` too


locks:
Agrreeedd!
@bors r+

PR #3920, opened 2021-09-12
Extract Sentry initialization code into dedicated function

Turbo87:
This will allow us to more easily enable Sentry error reporting for the other binaries too.


JohnTitor (approved):
👍

PR #3922, opened 2021-09-13
Split DB_POOL_SIZE & DB_MIN_IDLE environment variables

justahero:
This PR splits the existing environment variables `DB_POOL_SIZE` & `DB_MIN_IDLE` into separate variables to provide different connection settings to the primary & replica databases. The main reason for this change is to be able to configure different values for both database pools.

The new env variables are:

* `DB_PRIMARY_POOL_SIZE` & `DB_REPLICA_POOL_SIZE` to configure the number of connections in the primary & replica pool
* `DB_PRIMARY_MIN_IDLE` & `DB_REPLICA_MIN_IDLE` to set the number of idle connections each pool at least maintains
* update descriptions of env vars.

**Note** the descriptions of the `*_POOL_SIZE` variables in `app.json` are most likely off & need correct wording.

### DIFF
--- a/src/config/database_pools.rs
+++ b/src/config/database_pools.rs
@@ -66,8 +66,13 @@ impl DatabasePools {
             _ => Self::DEFAULT_POOL_SIZE,
         };
 
-        let min_idle = match dotenv::var("DB_MIN_IDLE") {
-            Ok(num) => Some(num.parse().expect("couldn't parse DB_MIN_IDLE")),
+        let primary_min_idle = match dotenv::var("DB_PRIMARY_MIN_IDLE") {
+            Ok(num) => Some(num.parse().expect("couldn't parse DB_PRIMARY_MIN_IDLE")),
+            _ => None,
+        };
+
+        let replica_min_idle = match dotenv::var("DB_REPLICA_MIN_IDLE") {
+            Ok(num) => Some(num.parse().expect("couldn't parse DB_REPLICA_MIN_IDLE")),
             _ => None,
         };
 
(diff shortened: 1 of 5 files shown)


src/config/database_pools.rs
```diff
@@ -20,15 +24,30 @@ pub struct DatabasePools {
...
+    pub fn primary_max_pool_size(&self) -> u32 {
+        self.primary.pool_size
+    }
+
+    pub fn replica_max_pool_size(&self) -> u32 {
+        self.replica
+            .as_ref()
+            .map(|config| config.pool_size)
+            .unwrap_or(Self::DEFAULT_POOL_SIZE)
+    }
```
pietroalbini, line 45: This function doesn't seem to be used anywhere.

src/config/database_pools.rs
```diff
@@ -20,15 +24,30 @@ pub struct DatabasePools {
...
 }
 
 impl DatabasePools {
     pub fn are_all_read_only(&self) -> bool {
         self.primary.read_only_mode
     }
+
+    pub fn primary_max_pool_size(&self) -> u32 {
+        self.primary.pool_size
+    }
```
pietroalbini, line 38: I'm not sure if it's worth having a function just for this, the rest of the code could access the pool size directly.

src/config/database_pools.rs
justahero: Good catch!

src/config/database_pools.rs
justahero: Removed this to be consistent.

pietroalbini:
Thanks!

@bors r+

PR #3926, opened 2021-09-16
Extract `cio_markdown` subpackage

Turbo87:
This allows us to isolate our markdown rendering code in a dedicated package, which could potentially have a positive effect on the compile times of the full project. It also means we can iterate on this part of the code independent from the other parts (see #3861 or #3297), and with a well-defined API interface.


pietroalbini:
Could we move it outside of `src`? It feels really confusing otherwise to have a subdirectory be a crate.

Turbo87:
@pietroalbini yeah, I think that makes sense. I've put it there because the `cargo_registry_s3` is also inside of `src`, but I would definitely prefer it outside of it.

PR #3935, opened 2021-09-17
Extract `sql` module with all custom `sql_function!()` calls

Turbo87:
This seems a bit more organized compared to having them scattered all over the place.


JohnTitor (approved):
Nice improvement 👍

PR #3951, opened 2021-09-23
Use `minijinja` instead of `handlebars`

Turbo87:
`minijinja` is explicitly built to be a light-weight templating system with minimal dependencies. Since we don't use a lot of the complexity that a full templating system like `handlebars` brings, we can reduce our number of transitive dependencies by replacing `handlebars` with `minijinja`. Note that it currently does not reduce the number of dependencies significantly, because other dependencies are still depending on these other transitive dependencies... 🙈

see https://github.com/mitsuhiko/minijinja


JohnTitor (approved):
👍 on my end but I'm not too familiar with the dump-db script. The tests should ensure it works fine but just in case, I'd like to see Justin's review. If they cannot respond for a while, feel free to r=me.

locks (approved):

Turbo87:
@JohnTitor thanks! looks like Justin is currently quite busy with his new job. since we have tests for this functionality I feel quite confident that this should not break anything :)

@bors r=JohnTitor

PR #3954, opened 2021-09-24
Remove `ember-cli-sri` addon temporarily

Turbo87:
https://github.com/ember-cli/broccoli-asset-rev/issues/29 can in some cases cause SRI failures. specifically, when file A depends on file B depends on file C, and file C changes, then the fingerprinting in file A and file B might not change, even though the injected filename of file C has changed, which leads to a fingerprint collision, which results in a user-visible SRI failure.

Disabling SRI is certainly not perfect, but it's better than having a broken site every now and then without any way to resolve it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
     "ember-cli-meta-tags": "6.1.2",
     "ember-cli-mirage": "2.2.0",
     "ember-cli-notifications": "7.0.0",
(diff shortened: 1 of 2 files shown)


locks:
I agree that it's better to remove it until there's a better resolution.

@bors r+

Issue #3930, opened 2021-09-16
Crate file download response headers indicate tar but it's actually a tar.gz


mstange:
**Describe the bug**

The HTTP response headers of a crate download, for example https://crates.io/api/v1/crates/tokio/1.6.1/download , do not indicate that the crate is gzipped. They only contain `content-type: application/x-tar`.

**To Reproduce**

Execute the following on the web console of your favorite browser:

```js
await fetch("https://crates.io/api/v1/crates/tokio/1.6.1/download", { mode: "cors" }).then(async r => [r.headers.get("Content-Type"), [...new Uint8Array(await r.arrayBuffer()).slice(0, 3)].map(x => x.toString(16).padStart(2, "0")).join(" ")])
```

This prints `[ "application/x-tar", "1f 8b 08" ]`. The 1f 8b 08 are the magic bytes of gzip.

**Expected behavior**
The Content-Type should probably be `application/gzip`. Alternatively, the response could be delivered with `Content-Encoding: gzip` so that the browser does the decompression and then passes the uncompressed tar bytes to JS.
But that might cause problems for other tools which expect to have to do the decompression themselves (rather than it being handled by their network layer).

**Additional context**
This is not an important issue, I only noticed it because there are some places on the internet that say that .crate files are .tar files and my tar parser wasn't working.

Turbo87:
I guess this would need to be changed in https://github.com/rust-lang/crates.io/blob/274f601cb5c4c8132c2250c6252818138255bab9/src/uploaders.rs#L155, but we should also ensure that older versions of `cargo` can handle this change correctly. Afterwards we could potentially iterate over all crate files on S3 and change the content type, if S3 supports such an operation.

@ehuss can you comment on this?

ehuss:
Sorry, I don't have any particular insight on this or what will break if it is changed.  It looks like gzip encoding was explicitly removed in https://github.com/rust-lang/crates.io/commit/53cbb26d022a88f7f311740f35ec5c6d4a5a9ab0.

arlosi:
Cargo does not inspect http headers when downloading crates.

Here's the relevant section of Cargo: https://github.com/rust-lang/cargo/blob/d14c59dfe922d7b117f952bf2f210ef503968a31/src/cargo/core/package.rs#L756

It would need to call `handle.header_function(...)` to in order to look at the headers.

Turbo87:
[...]

PR #3955, opened 2021-09-24
Change crate file `Content-Type` to `application/gzip`

Turbo87:
Using `application/x-tar` is incorrect, because these files are not plain tarballs, but gzipped tarballs instead.

Resolves the first part of https://github.com/rust-lang/crates.io/issues/3930.

Existing crate files would still need to be changed using something like https://stackoverflow.com/a/46932483/1478093

### DIFF
--- a/src/tests/http-data/krate_publish_features_version_2
+++ b/src/tests/http-data/krate_publish_features_version_2
@@ -22,7 +22,7 @@
         ],
         [
           "content-type",
-          "application/x-tar"
+          "application/gzip"
         ],
         [
           "authorization",
--- a/src/tests/http-data/krate_publish_good_badges
+++ b/src/tests/http-data/krate_publish_good_badges
@@ -26,7 +26,7 @@
         ],
         [
           "content-type",
-          "application/x-tar"
+          "application/gzip"
         ],
         [
           "accept-encoding",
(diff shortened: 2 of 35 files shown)


pinkforest:
May I suggest to use the old Content-type for when .crate is requested and the new Content-Type when someone requests .tar.gz?

The .crate suffix is quite confusing and it might help gate deprecation for some people

You could even at some stage do a http redirect from .crate to .tar.gz when it gets deprecated

Turbo87:
@pinkforest we are intentionally using `.crate`, because the fact that this is a gzipped tarball is actually considered an implementation detail and not necessarily something stable that should be relied upon. it has never been changed so far though, and it's unlikely to be changed in the near future, so it's probably fine for current tooling to assume this. nevertheless, I don't think we will change the file extension for these reasons.

thinking some more about what I said, I'm wondering if `application/gzip` actually makes sense if the "gzipped tarball" part is not actually considered to be an implementation detail. I'm now wondering if the generic `application/octet-stream` makes more sense.

pinkforest:
GNU tar is "smart" enough to figure out whether it it's .gz and uses gzip automatically even without the -z flag

However some (now) esoteric historic commercial unix tar will complain :)
```

$ tar -zcvf foo.tar.gz foo
foo/
$ gzip -d foo.tar.gz 
$ tar -xvf foo.tar
foo/
$ tar -cvf foo.tar.gz foo
foo/
$ gzip -d foo.tar.gz 
gzip: foo.tar.gz: not in gzip format
$ tar -xvf foo.tar
foo/
```

I would just leave it with that current Content-Type :)

PR #3956, opened 2021-09-25
Use `HeaderValue::from_static()` instead of `::from(&str)`

Turbo87:
This should be slightly more efficient 😄


hi-rustin (approved):
Looks good!

PR #3958, opened 2021-09-26
Rename `cio-markdown` package to `cargo-registry-markdown`

Turbo87:
This is consistent with `cargo-registry` and `cargo-registry-s3`.

We could also consider renaming `cargo-registry-s3` to `cio-s3` to make the name shorter, but as long as the main package is called `cargo-registry` I guess we should us that as the prefix for the other packages too.


JohnTitor (approved):
👍

PR #3967, opened 2021-09-30
admin/render_readmes.rs to correctly identify the readme path

nipunn1313:
Previously, it would try to get a field called `readme_file`
which never existed in the Cargo.toml manifest format. This
fixes it.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field

I believe the confusion arose because the `cargo package` command
actually converts the readme contents -> readme field and the
Cargo.toml readme field -> `readme_file` field when making the
request to crates.io.

However, in the admin render_readme backfill, it is opening the
original Cargo.toml which has a `readme` field for the path.

I manually tested with a local backend/frontend and confirmed that
readme paths were being re-rendered correctly.

Manual test screenshots:

Before this change - after doing a rerender
image

After this change - after doing a rerender
image

Repro steps
Set up a local dev environment via the [docs/CONTRIBUTING.md](https://github.com/rust-lang/crates.io/blob/master/docs/CONTRIBUTING.md) instructions.
Publish a package (I used one called `innercrate`) (using above instructions) with nonroot README
```
➜  innercrate git:(master) ✗ cat docs/README.md
Check out [other file](./Otherfile.md)
➜  innercrate git:(master) ✗ cat docs/Otherfile.md
hello world
➜  innercrate git:(master) ✗ grep readme Cargo.toml
readme = "docs/README.md"
```

Re-render readmes
`cargo run --bin crates-admin -- render-readmes --crate innercrate`
Look at http://localhost:4200/crates/innercrate

### DIFF
--- a/src/admin/render_readmes.rs
+++ b/src/admin/render_readmes.rs
@@ -213,31 +213,24 @@ fn get_readme(
     };
 
     let rendered = {
-        let path = format!(
-            "{}-{}/{}",
-            krate_name, version.num, manifest.package.readme?
-        );
+        let readme_path = manifest.package.readme.as_ref()?;
+        let path = format!("{}-{}/{}", krate_name, version.num, readme_path);
         let contents = find_file_by_path(&mut entries, Path::new(&path), version, krate_name);
         text_to_html(
             &contents,
-            manifest
-                .package
-                .readme_file
-                .as_ref()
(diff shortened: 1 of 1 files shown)


nipunn1313:
Related to #3861

Turbo87 (approved):
sounds reasonable to me, thanks!

PR #3969, opened 2021-10-01
Add tests for admin/render_readmes

nipunn1313:
Depends on #3967
(Because of https://github.com/github/feedback/discussions/4477#discussioncomment-1293255 - github ends up rendering both diffs together. Go to the commits tab and just look at the most recent commit to review).

Refactors slightly so we can add tests to the untested risky portion of this code.
I already found a couple of bugs which I will fix in follow up PRs with tests.

### DIFF
--- a/src/admin/render_readmes.rs
+++ b/src/admin/render_readmes.rs
@@ -192,14 +192,18 @@ fn get_readme(
     }
 
     let reader = GzDecoder::new(response);
-    let mut archive = Archive::new(reader);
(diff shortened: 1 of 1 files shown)


Turbo87:
@nipunn1313 I took the liberty of rebasing your branch and splitting it up into a few more individual commits. I hope thats okay with you :)

nipunn1313:
Perfect. The split up commits make a lot of sense to me!

PR #3970, opened 2021-10-01
Support admin render_readme with Cargo.toml without optional readme field

nipunn1313:
Depends on #3969

(Because of github/feedback#4477 (comment) - github ends up rendering both diffs together. Go to the commits tab and just look at the most recent commit to review).

https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field

### DIFF
--- a/src/admin/render_readmes.rs
+++ b/src/admin/render_readmes.rs
@@ -210,12 +210,16 @@ fn render_pkg_readme<R: Read>(mut archive: Archive<R>, pkg_name: &str) -> Option
     };
 
     let rendered = {
-        let readme_path = manifest.package.readme.as_ref()?;
+        let readme_path = manifest
+            .package
+            .readme
+            .clone()
+            .unwrap_or_else(|| "README.md".into());
         let path = format!("{}/{}", pkg_name, readme_path);
         let contents = find_file_by_path(&mut entries, Path::new(&path), pkg_name)?;
         text_to_html(
             &contents,
-            readme_path,
+            &readme_path,
             manifest.package.repository.as_deref(),
         )
     };
@@ -302,6 +306,23 @@ readme = "README.md"
         );
         let serialized_archive = pkg.into_inner().unwrap();
         assert!(render_pkg_readme(tar::Archive::new(&*serialized_archive), "foo-0.0.1").is_none());
+    }
+
+    #[test]
+    fn test_render_pkg_implicit_readme() {
+        let mut pkg = tar::Builder::new(vec![]);
+        add_file(
+            &mut pkg,
+            "foo-0.0.1/Cargo.toml",
+            br#"
+[package]
+"#,
+        );
+        add_file(&mut pkg, "foo-0.0.1/README.md", b"readme");
+        let serialized_archive = pkg.into_inner().unwrap();
+        let result =
+            render_pkg_readme(tar::Archive::new(&*serialized_archive), "foo-0.0.1").unwrap();
+        assert!(result.contains("readme"))
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/admin/render_readmes.rs
```diff
@@ -203,18 +203,22 @@ fn render_pkg_readme<R: Read>(mut archive: Archive<R>, pkg_name: &str) -> Option
...
             .unwrap_or_else(|_| panic!("[{}] Syntax error in manifest file", pkg_name))
     };
 
     let rendered = {
-        let readme_path = manifest.package.readme.as_ref()?;
+        let readme_path = manifest
+            .package
+            .readme
+            .clone()
+            .unwrap_or("README.md".into());
```
Turbo87, line 216: the docs mention:

> If no value is specified for this field, and a file named README.md, README.txt or README exists

Seems like only checking for `README.md` might not be sufficient? 🤔

Turbo87:
Now that I read the docs on this a little more closely I'm wondering how we are handling `readme = false`, because I guess it wouldn't be able deserialize to `Option<String>`, right? do we just panic in that case? 😱

Turbo87:
I've rebased the branch btw :)

nipunn1313:
yep you're right. There's more to the spec that's not supported. Let me see what I can do.
At the least - this is an improvement (supports more cases than before).

src/admin/render_readmes.rs
nipunn1313: In order to handle this - I will first need to fix a bug where the script won't work if the archive entries are in a different order (such that it can look for multiple different paths in the archive). I think that work is sufficiently different, that I'll make it a separate PR.

I'd advocate for landing this as-is (as a step improvement), followed by the separate PRs for tar ordering and full support for different README types.

nipunn1313:
#3971 adds support for out-of-order tar files
That unblocks #3972 which adds support for README.txt and README as well as `readme = false`
Together they unblock #3973 - which supports relative links in READMEs from non-repo-root packages. Technically it only logically depends on #3971 - but for ease of code rebase I stacked them in this order.

src/admin/render_readmes.rs
```diff
@@ -290,6 +294,37 @@ readme = "README.md"
...
+            "foo-0.0.1/Cargo.toml",
+            br#"
+[package]
+"#,
+        );
+        add_file(&mut pkg, "foo-0.0.1/README.md", b"readme");
+        let serialized_archive = pkg.into_inner().unwrap();
+        let result =
+            render_pkg_readme(tar::Archive::new(&*serialized_archive), "foo-0.0.1").unwrap();
+        assert!(result.contains("readme"))
```
Turbo87, line 325: any reason for using `contains` here instead of `assert_eq`?

nipunn1313: For the shorter ones like this, I could do assert_eq as it comes out to
```<p>readme</p>\n```

For the longer ones, they're really verbose and full of stuff not relevant to the test itself, so I was using the `result.contains` pattern instead.
```
<p>docs/readme <a href=\"https://github.com/foo/foo/blob/HEAD/docs/./Other.md\" rel=\"nofollow noopener noreferrer\">link</a></p>\n
```

What do you think? Switch them over?

Turbo87: I would have still used `assert_eq`, but I guess it's fine like this too :)

Turbo87 (approved):

Turbo87:
probably caused by https://www.githubstatus.com/incidents/bdbzpz7qxmbx

@bors retry

PR #3971, opened 2021-10-01
admin::render_readmes: Support out of order tar files

nipunn1313:
Depends on #3970

(Because of github/feedback#4477 - github ends up rendering both diffs together. Go to the commits tab and just look at the most recent commit to review).


Turbo87:
hmm, I do admit that I'm a bit worried about pulling all of the contents into memory. we do have a few crates that are several dozens of megabytes gzipped, so uncompressing all of their contents and keeping them in memory might be problematic.

nipunn1313:
One saving grace is that it's only pulling them in from memory one at a time. Also it's only doing it on the machines running the admin scripts. Several dozens of megabytes in memory seems fine?

Mitigating ideas include
- Streaming the entire crate off of crates.io multiple times to get the various files (seems inelegant). One pass to get the Cargo.toml and another pass to get the README.
- heuristically stripping out parts that we're definitely not going to read. Since we're looking for README files, we can filter by extension to eliminate large files that aren't going to be read.

If I were deploying this, I'd probably try to avoid adding such heuristics/complexity until they are proven necessary. I've often found that unnecessary complexity like that can be a cure that's worse than the disease.

Since it's an admin functionality, we could try it out and fix based on the issues that come up. I don't have access to run the admin functionality (of course) - but if you assigned me a task with the failing crates, I could look at them and add heuristics around the specific issues that come up. My suspicion is that nothing would come up - as it seems unlikely that we'd have gigabyte sized crates (is there a way to check easily ahead of time?)

PR #3973, opened 2021-10-01
Support relative URLs from repo-subdirectory packages

### LINKED ISSUE #3484: Relative URLs in subdirectory readmes

nipunn1313:
Extract `.cargo_vcs_info.json` from the tarball and use it to determine the path in vcs.

Support for this was added to cargo with
https://github.com/rust-lang/cargo/pull/9866

Fixes #3484 
(note that support for this in admin::render_readmes not yet exists - would come with #4095)

Depends on #4097 (Because of github/feedback#4477 - github ends up rendering the diffs together. Go to the commits tab and just look at the most recent commits to review).

### DIFF
--- a/cargo-registry-markdown/lib.rs
+++ b/cargo-registry-markdown/lib.rs
@@ -245,18 +245,23 @@ static MARKDOWN_EXTENSIONS: [&str; 7] =
 /// use cargo_registry_markdown::text_to_html;
 ///
 /// let text = "[Rust](https://rust-lang.org/) is an awesome *systems programming* language!";
-/// let rendered = text_to_html(text, "README.md", None);
+/// let rendered = text_to_html(text, "README.md", None, None);
 /// assert_eq!(rendered, "<p><a href=\"https://rust-lang.org/\" rel=\"nofollow noopener noreferrer\">Rust</a> is an awesome <em>systems programming</em> language!</p>\n");
 /// ```
-pub fn text_to_html(text: &str, path: &str, base_url: Option<&str>) -> String {
-    let path = Path::new(path);
-    let base_dir = path.parent().and_then(|p| p.to_str()).unwrap_or("");
+pub fn text_to_html(
+    text: &str,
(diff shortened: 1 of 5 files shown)


nipunn1313:
Confirmed things worked with a repo called `testcrate` and a pkg in a subdir called `innercrate`

image

Reproed with publishing innercrate via
```
~/src/cargo/target/debug/cargo publish --index file:///Users/nipunn/src/crates.io/tmp/index-bare --token=$TOKEN
```

and double checked the backfill as well with 
```
cargo run --bin crates-admin -- render-readmes --crate innercrate
```

Turbo87 (approved):
@nipunn1313 sorry for letting this wait for so long. I misunderstood that this was ready for review. 🙈 

looks very good though. nice work! ❤️

Turbo87:
I've rebased and resolved the conflict, otherwise this is identical to the original commits :)

@bors r+

nipunn1313:
Hooray! Thanks @Turbo87

PR #3980, opened 2021-10-05
CI: Replace custom caching with `rust-cache` action

Turbo87:
see https://github.com/Swatinem/rust-cache


Turbo87:
looks like the cache sizes and CI times are comparable, but with less custom and verbose code :)

JohnTitor (approved):
Okay, makes sense to me now, thanks for clarifying!

PR #3992, opened 2021-10-07
database/versions: Add generated `version_no_prerelease` column

Turbo87:
This is an attempt to speed up the reverse dependencies query, as hinted at in https://github.com/rust-lang/crates.io/pull/3886#issuecomment-913432943.

Instead of running `to_semver_no_prerelease(num)` in every query, we use a [generated column](https://www.postgresql.org/docs/current/ddl-generated-columns.html) to speed up the query.

In my local tests with the imported database dump the query times were previously around 1700ms for the `rand` crate, and after this change they consistently dropped down to around 900ms.

### DIFF
--- a/migrations/2021-10-07-220812_add_version_no_prerelease/down.sql
+++ b/migrations/2021-10-07-220812_add_version_no_prerelease/down.sql
@@ -1,0 +1,2 @@
+ALTER TABLE versions
+    DROP COLUMN version_no_prerelease;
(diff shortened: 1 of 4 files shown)


Turbo87:
rebased on top of #3993 to fix the `versions` table trigger issue that was causing the tests to fail.

Turbo87:
closing in favor of

- #4022

PR #3999, opened 2021-10-11
Implement `version_id` cache for `download` endpoint using `moka` crate

Turbo87:
This PR builds on top of https://github.com/rust-lang/crates.io/pull/3848, but uses https://github.com/moka-rs/moka instead of https://github.com/xacrimon/dashmap for caching the `version_id` in the `download` API endpoint. I've implemented it using environment variables and default values that will allow us to fine tune the values in production, if necessary.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -36,7 +36,7 @@ pub struct App {
     ///
     /// This is used by the download endpoint to reduce the number of database queries. The
     /// `version_id` is only cached under the canonical spelling of the crate name.
-    pub(crate) version_id_cacher: DashMap<String, i32>,
+    pub(crate) version_id_cacher: DashMap<(String, String), i32>,
 
     /// Count downloads and periodically persist them in the database
     pub downloads_counter: DownloadsCounter,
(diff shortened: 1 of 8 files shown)


src/app.rs
```diff
@@ -148,12 +158,27 @@ impl App {
...
 
+        let version_id_cache_size = dotenv::var("VERSION_ID_CACHE_SIZE")
+            .ok()
+            .map(|s| s.parse().unwrap())
+            .unwrap_or(DEFAULT_VERSION_ID_CACHE_SIZE);
+
+        let version_id_cache_ttl = dotenv::var("VERSION_ID_CACHE_TTL")
+            .ok()
+            .map(|s| s.parse().unwrap())
+            .unwrap_or(DEFAULT_VERSION_ID_CACHE_TTL);
```
pietroalbini, line 169: Can we move this to `Config`?

pietroalbini:
Generally yes, but I don't have time for a full review right now.

pietroalbini (approved):

PR #4022, opened 2021-10-17
database/versions: Add new `semver_no_prerelease` column

Turbo87:
This PR is an alternative to:
- https://github.com/rust-lang/crates.io/pull/3992 and
- https://github.com/rust-lang/crates.io/pull/4016

Instead of using our custom `semver_triple` database type (which caused issues with `diesel`) we will use an array (`NUMERIC[3]`). This provides the same sorting and filtering functionality, but is easier to deal with in `diesel`.

This PR only adds the new column and adjusts the code to fill it for new crate versions, but it does not backfill it for old versions. Neither does it adjust the reverse dependencies SQL query to use the new column, as that requires the old versions to be backfilled first. This will be changed in a future PR.

A SQL script to backfill the column for old versions is:

```sql
-- add temporary function that returns an array instead of `semver_triple`

CREATE FUNCTION to_semver_no_prerelease_array(TEXT) RETURNS NUMERIC[3]
    IMMUTABLE
    LANGUAGE sql
AS
$$
SELECT ARRAY [
           split_part($1, '.', 1)::numeric,
           split_part($1, '.', 2)::numeric,
           split_part(split_part($1, '+', 1), '.', 3)::numeric
           ]
WHERE strpos($1, '-') = 0
$$;

-- fill the new column with data incrementally to avoid table locking for
-- extended time periods.

UPDATE versions
SET semver_no_prerelease = to_semver_no_prerelease_array(num)
WHERE semver_no_prerelease IS NULL
  AND created_at < '2016-01-01';

UPDATE versions
SET semver_no_prerelease = to_semver_no_prerelease_array(num)
WHERE semver_no_prerelease IS NULL
  AND created_at < '2016-02-01';

UPDATE versions
SET semver_no_prerelease = to_semver_no_prerelease_array(num)
WHERE semver_no_prerelease IS NULL
  AND created_at < '2016-03-01';

-- ...

UPDATE versions
SET semver_no_prerelease = to_semver_no_prerelease_array(num)
WHERE semver_no_prerelease IS NULL
  AND created_at < '2021-12-01';

-- drop temporary function again

DROP FUNCTION to_semver_no_prerelease_array(TEXT);
```


Turbo87:
closing in favor of https://github.com/rust-lang/crates.io/pull/4106

PR #4030, opened 2021-10-19
admin::on_call: Work around potential Rust v1.57.0-beta.1 bug

Turbo87:
The fields in the `InvalidEvent` struct are used by the `Debug` implementation, which is used by the `anyhow!()` macro call in the `Event::send()` implementation.

The Rust v1.57.0-beta.1 compiler does not appear to agree on that though and marks these fields as unused. This PR adds `#[allow(dead_code)]` to work around the issue for now and fix our CI builds to unblock other updates.

### DIFF
--- a/src/admin/on_call.rs
+++ b/src/admin/on_call.rs
@@ -62,6 +62,7 @@ struct FullEvent {
 }
 
 #[derive(serde::Deserialize, Debug)]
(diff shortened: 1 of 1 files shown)


Turbo87:
for reference, this got reported at https://github.com/rust-lang/rust/issues/90046 and it looks like there was also a previous issue:

- https://github.com/rust-lang/rust/issues/88900

PR #4108, opened 2021-10-29
Extract `worker` module

Turbo87:
This PR renames the `tasks` module to `worker`, to make it clearer that these things are running on the background worker. It also moves the other swirl tasks outside of the `tasks` module into the new `worker` module.


JohnTitor (approved):
Looks reasonable 👍

PR #4115, opened 2021-11-02
Sentry: Improve metadata collection

Turbo87:
The `request.id` tag and the "custom metadata" are currently only assigned to the Sentry scope in the `after()` method of the `LogRequest` middleware. This means that for any manual `sentry::capture_exception()` call these fields will not be set correctly.

This PR moves the `configure_scope()` calls into the `before()` method and `add_custom_metadata()` function instead. This allows us to send the fields to Sentry for error reports as soon as they are available.


JohnTitor (commented):
Makes sense to me.

JohnTitor (approved):

PR #4116, opened 2021-11-02
middleware::log_request: Split request timing code into dedicated middleware

Turbo87:
This is generic and not strictly related to the `LogRequest` middleware. The `UpdateMetrics` middleware should probably also use this in the future.

### DIFF
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -7,8 +7,8 @@ use crate::util::request_header;
 use conduit::{header, RequestExt, StatusCode};
 use conduit_cookie::RequestSession;
 
+use crate::middleware::response_timing::ResponseTime;
 use std::fmt::{self, Display, Formatter};
-use std::time::{SystemTime, UNIX_EPOCH};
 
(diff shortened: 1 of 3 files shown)


src/middleware/log_request.rs
```diff
@@ -186,8 +152,10 @@ impl Display for RequestLine<'_> {
             line.add_quoted_field("error", err)?;
         }
 
-        if self.response_time > SLOW_REQUEST_THRESHOLD_MS {
-            line.add_marker("SLOW REQUEST")?;
+        if let Some(response_time) = response_time {
+            if response_time.as_millis() > SLOW_REQUEST_THRESHOLD_MS {
+                line.add_marker("SLOW REQUEST")?;
+            }
```
JohnTitor, line 158: We could de-duplicate this `if let` by merging it into the above `if let` or binding `response_time`.

Turbo87: the problem with merging them is that it changes the field order

JohnTitor: Ah, I see, then it should be fine. Thanks for clarifying!

JohnTitor (approved):

PR #4119, opened 2021-11-02
admin::render_readmes: Use `Result` return types

Turbo87:
... instead of dropping the errors via `Option` and using `panic!()` everywhere.

/cc @nipunn1313


JohnTitor (approved):
Nice improvement :)

nipunn1313:
Yes this looks great! Thank you @Turbo87 !

PR #4120, opened 2021-11-02
worker::git: Move database update out of `yank()` job

Turbo87:
When publishing, we also save the new version (and crate) directly in the database, so we should be able to do the same when yanking/unyanking, and avoid potentially returning incorrect data right after the request and before the background job has finished.

### DIFF
--- a/src/controllers/version/yank.rs
+++ b/src/controllers/version/yank.rs
@@ -42,6 +42,12 @@ fn modify_yank(req: &mut dyn RequestExt, yanked: bool) -> EndpointResult {
     if user.rights(req.app(), &owners)? < Rights::Publish {
         return Err(cargo_err("must already be an owner to yank or unyank"));
     }
+
+    if version.yanked == yanked {
+        // The crate is alread in the state requested, nothing to do
+        return ok_true();
+    }
+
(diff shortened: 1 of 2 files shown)


src/controllers/version/yank.rs
```diff
@@ -42,6 +43,16 @@ fn modify_yank(req: &mut dyn RequestExt, yanked: bool) -> EndpointResult {
     if user.rights(req.app(), &owners)? < Rights::Publish {
         return Err(cargo_err("must already be an owner to yank or unyank"));
     }
+
+    if version.yanked == yanked {
+        // The crate is alread in the state requested, nothing to do
```
JohnTitor, line 48: Pre-existing, but:

```suggestion
        // The crate is already in the state requested, nothing to do
```

JohnTitor (approved):
LGTM, but I'd like to hear @pietroalbini's thought as they're more familiar with it than me.

Turbo87:
@JohnTitor I've discussed the PR yesterday in the team meeting with @jtgeibel and he agreed with the implementation. I've opened #4172 as a follow-up item.

@bors r+

PR #4124, opened 2021-11-03
middleware::log_request: Move `OriginalPath` code into `NormalizePath` middleware

Turbo87:
The `NormalizePath` middleware is the one that mutates the `path`, so it should also be responsible for saving the original one, if necessary.

This PR slightly changes the logic in that `OriginalPath` is only added to the request extensions if the path has been changed. The `NormalizePath` middleware now defaults to `request.path()`, if `OriginalPath` is not found in the request extensions.

### DIFF
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -169,8 +169,9 @@ impl<'a> Display for FullPath<'a> {
         let request = self.0;
 
         let original_path = request.extensions().find::<OriginalPath>();
-        // Unwrap shouldn't panic as no other code has access to the private struct to remove it
-        let path = original_path.map(|p| p.0.as_str()).unwrap();
+        let path = original_path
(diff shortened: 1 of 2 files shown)


src/middleware/log_request.rs
```diff
@@ -166,13 +162,17 @@ struct FullPath<'a>(&'a dyn RequestExt);
 
 impl<'a> Display for FullPath<'a> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let request = self.0;
+
+        let original_path = request.extensions().find::<OriginalPath>();
         // Unwrap shouldn't panic as no other code has access to the private struct to remove it
```
JohnTitor, line 168: This comment is now outdated and could be removed.

JohnTitor (approved):
Makes sense to me, feel free to r=me once the comment below is addressed.

PR #4126, opened 2021-11-04
Replace `jemallocator` with `tikv-jemallocator`

Turbo87:
`jemallocator` appears to be unmaintained and [`tikv-jemallocator`](https://github.com/tikv/jemallocator) is a friendly fork that has taken over maintenance.


JohnTitor (approved):

JohnTitor:
Good call! @bors r+

PR #4130, opened 2021-11-04
services/playground: Extract loadCrates() method

Turbo87:
This avoids duplicate requests to the playground API, caused by the `preloadPlaygroundCratesTask` in the `application` route requesting the data after it has already been loaded by the `CrateSidebar` component too.


locks:
Nicely spotted.
@bors r+

PR #4137, opened 2021-11-06
middleware::LogRequests: Extract Sentry setup into dedicated middleware struct

Turbo87:
The `LogRequests` should be responsible for logging the requests, but nothing more. This PR extracts the custom Sentry code from it and moves it into a dedicated middleware, that wraps the `sentry-conduit` middleware.


JohnTitor (approved):
Makes sense to me!

Issue #4083, opened 2021-10-25
Pending Owner Invites page in crates.io website doesn't have pagination feature


kt3k:
**Describe the bug**
Pending Owner Invites page in crates.io website doesn't have pagination feature. If you have more than 12 invites, the page only shows the first 11, and there's no indication of the rest of invites existing.

**To Reproduce**
1. Receive more than 12 invites
2. Go to `/me/pending-invites` page

**Expected behavior**
Show the `next` button somewhere in the page to navigate to the next page of pending invites.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Version: 94

EstebanBorai:
I will take a look on this!

PR #4145, opened 2021-11-07
mirage: Implement `GET /api/private/crate_owner_invitations` route handler

Turbo87:
This will be needed to address https://github.com/rust-lang/crates.io/issues/4083


locks (approved):
@bors r+

PR #4148, opened 2021-11-08
Pin Node.js version

Turbo87:
We currently use one Node.js version in production, another on CI and yet another one for our Docker setup. This PR pins them all to the same release (latest v14.x) and configures renovatebot to automatically keep the versions updated.


locks:
Pretty good 💪 Shouldn't CI pick up the version from `package.json`?

Turbo87:
> Shouldn't CI pick up the version from `package.json`?

not that I'm aware of 🤔

locks:
I may be confusing it with the volta GitHub Action 🤔

PR #4152, opened 2021-11-09
CrateSidebar: Show detailed list for small number of owners

Turbo87:
This PR extracts a `OwnersList` component and then adjusts it to show a more detailed list if the number of owners is five or less.

### Before:

Bildschirmfoto 2021-11-09 um 20 24 00

### After: 

Bildschirmfoto 2021-11-09 um 20 23 41


locks:
Love the change!

@bors r+

PR #4156, opened 2021-11-10
Show 404 pages for unknown crates/versions

Turbo87:
Instead of redirecting back to a known URL and showing a notification message, we now stay on the requested URL and show an error page instead.

Bildschirmfoto 2021-11-10 um 14 41 29

(though with "Crate/Version not found" instead)

The implementation is a little questionable, but due to the way the `error` substates in Ember.js work I've found no other way to implement this.

This PR is based on #4154.


locks:
:shipit: 
@bors r+

PR #4162, opened 2021-11-11
OwnersList: Fall back to `login` if `name` is not filled

### LINKED ISSUE #4161: Owner list is missing names
**Describe the bug**
Not all users have a `name`. 

Bildschirmfoto 2021-11-11 um 20 33 06

**To Reproduce**
1. Go to https://crates.io/crates/civet
2. See missing owner name

**Expected behavior**
The code should fall back to the `login` if `name` is not filled.

Turbo87:
Fixes #4161

### DIFF
--- a/app/components/owners-list.hbs
+++ b/app/components/owners-list.hbs
@@ -11,8 +11,8 @@
         local-class="link"
         data-test-owner-link={{owner.login}}
       >
-        <UserAvatar @user={{owner}} @size="medium-small" local-class="avatar" />
-        <span local-class="name">{{or owner.display_name owner.name}}</span>
+        <UserAvatar @user={{owner}} @size="medium-small" local-class="avatar" aria-hidden="true" />
+        <span local-class="name {{unless this.showDetailedList "hidden-name"}}">{{or owner.display_name owner.name}}</span>
       </LinkTo>
     </li>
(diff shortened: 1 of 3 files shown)


locks:
Should we add a test to avoid regressions? 👀

Turbo87:
@locks yep, good idea. I've added a few tests for the component and I've also included a small a11y improvement.

app/components/owners-list.hbs
```diff
@@ -11,8 +11,8 @@
         local-class="link"
         data-test-owner-link={{owner.login}}
       >
-        <UserAvatar @user={{owner}} @size="medium-small" local-class="avatar" />
-        <span local-class="name">{{or owner.display_name owner.name}}</span>
+        <UserAvatar @user={{owner}} @size="medium-small" local-class="avatar" aria-hidden="true" />
+        <span local-class="name {{unless this.showDetailedList "hidden-name"}}">{{or owner.display_name owner.name owner.login}}</span>
```
locks, line 15: I feel like we could move the `or` into a getter in the model, but that's just a nit.

Turbo87: yeah, I had a similar though, but I think I'd rather extract a `get display_name()` on the `User` model instead

PR #4165, opened 2021-11-11
bin/server: Simplify assignments via `match`

Turbo87:
This PR replaces two conditional assignments with `match` assignments to make the code a little less verbose. Semantically these instructions should be identical.


locks:
Nice changes 👍 
@bors r+

PR #4166, opened 2021-11-11
bin/server: Print HTTP server URL on startup

Turbo87:
This makes it easier to click on the URL instead of having to copy-paste the port.

I'm aware that the code comment says "Do not change this line!", but I've adjusted the corresponding test code too... 😉


locks:
Sounds like a small and useful quality of life improvement!

@bors r+

PR #4173, opened 2021-11-13
TestApp: Simplify `crates_from_index_head()` method

Turbo87:
We can calculate the `path` automatically from the `crate_name`. There is no need for us to manually specify it in the tests.


JohnTitor (approved):

JohnTitor:
Looks good! @bors r+

Issue #4172, opened 2021-11-13
Implement `update_index` background worker task


Turbo87:
We currently have an `add_crate` background worker task, that gets called when a new crate (or version) is published. This task appends the version metadata to the file in the package index repository which corresponds to the crate.

We currently have another background worker task, `yank`, which gets called when a version is yanked (or unyanked). This tasks changes the `yanked` value in the version metadata in the package index repository.

Our background worker implementation (https://github.com/sgrif/swirl/) mostly runs tasks in order, but, due to built-in retry logic, that is not guaranteed. This means we could end up in a situation where the tasks are run in an order like: unyank, yank, yank, publish, unyank.

Instead of telling the background worker explicitly what action should be performed, it would be better to only tell it to update the package index repository according to the current state in the database.

We should implement a new `update_index(crate_name, version_num)` background worker task, that looks at the state of the index and of the database and then updates the index according to the current state of the database.

In a second step that background worker tasks should ideally then also handle the cases where versions (or even crates) are completely removed from the database due to e.g. GDPR requests.

Turbo87:
going to close this as good enough, now that https://github.com/rust-lang/crates.io/pull/4175 is implemented.

we currently don't hold all the data from the index in the database, so an easy sync with just providing crate name and version number is unfortunately not possible until we mirror all the index data in the database too.

PR #4175, opened 2021-11-14
worker::git: Read `yanked` status from database 

Turbo87:
This PR implements one part of https://github.com/rust-lang/crates.io/issues/4172. By no longer passing `yanked` into the `yank` (now `sync_yanked`) background worker task, and instead reading `yanked` from the database, we can avoid issues from tasks running out of order.


JohnTitor (approved):

JohnTitor:
Looks fine 👍 @bors r+

PR #4293, opened 2021-12-18
Replace `license-exprs` crate with `spdx`

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/2595

tl;dr you can now use parenthesis (e.g. `X AND (Y OR Z)`) in license expressions. the custom `/` operator is still supported for now, though not directly compatible with the use of parenthesis.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,7 +271,6 @@ dependencies = [
  "indexmap",
  "lazy_static",
  "lettre",
- "license-exprs",
  "minijinja",
  "moka",
  "oauth2",
@@ -286,6 +285,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2 0.10.0",
+ "spdx",
  "swirl",
  "tar",
  "tempfile",
@@ -1463,12 +1463,6 @@ dependencies = [
  "pkg-config",
  "vcpkg",
 ]
-
-[[package]]
-name = "license-exprs"
-version = "1.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a577c3a5f3982766dd577c4440b86c53c6ffd42007e4c1d83d2da3ce9f99cc3d"
 
 [[package]]
 name = "lock_api"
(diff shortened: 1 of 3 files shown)


ehuss:
Just a few questions and observations:

* Do you know which version of the SPDX list this currently uses?  That is something that we document.
* Did you consider using the Lax mode?  It looks like this will cause a fair number of packages to start to fail to publish. I compared the parsing [here](https://gist.github.com/ehuss/68ddbcbc89b187a903b8c12f86f67e21). That includes spdx versus spdx Lax, versus license-exprs.

Turbo87:
> * Do you know which version of the SPDX list this currently uses?  That is something that we document.

`spdx@0.7` is using SPDX 3.14 at the moment, which according to https://github.com/rust-lang/crates.io/issues/2595#issuecomment-995877817 should be compatible with the 3.11 that `license-exprs` currently uses.

> * Did you consider using the Lax mode?  It looks like this will cause a fair number of packages to start to fail to publish.

As far as I understand the lax mode was created to be able to handle all currently existing `license` entries in the crates.io database. But that includes entries that are already no longer considered valid by crates.io, which is why I'm a little hesitant to use the complete Lax mode.

I have however opened https://github.com/EmbarkStudios/spdx/pull/50 two days ago, which makes the lax mode a little more configurable. That would allow us to e.g. enable only the support for the custom slash operator, but keep support for other lax mode feature disabled. I'm fine with waiting until that change has shipped, if you prefer.

ehuss:
I did another scan where I checked when the last time the crate was updated: https://gist.github.com/ehuss/ce7a72f4eecb2a7ff1d668a936f5587f

All of the errors in the past year have been due to postfix plus on gpl.  If the number was small, I was going to suggest just sending PRs to the affected crates and ignoring the Lax mode stuff.  However, there is about a 100 so that's probably too many to do manually.

If your PR gets accepted within some reasonable timeframe, I think it would be good to wait.

Turbo87:
> If your PR gets accepted within some reasonable timeframe, I think it would be good to wait.

It was just merged and released. I've rebased and adjusted this PR :)

ehuss (approved):
I have not reviewed the `spdx` crate, but otherwise this seems reasonable.

PR #4302, opened 2021-12-20
renovate: Pin dependency versions

Turbo87:
see https://docs.renovatebot.com/dependency-pinning/


Turbo87:
@bors r+

Merging this myself as I currently appear to be the only one performing dependency updates and the reasoning in the linked page seems good to me.

PR #4316, opened 2021-12-22
Improve database fallbacks

Turbo87:
As suggested on Discord (see https://discord.com/channels/442252698964721669/835156566746595386/923210089085665381), this PR:

- implements a primary database fallback if the replica is unavailable
- renames the existing connection pool methods to `db_read()` and `db_write()`
- implements an additional `db_read_prefer_primary()` method
- uses that new method for endpoints that prefer the primary database, but don't need to write anything to the database

This should improve our resilience a little bit in case of database issues.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -150,7 +150,7 @@ pub trait RequestTransaction {
 
     /// Obtain a readonly database connection from the replica pool
     ///
-    /// If there is no replica pool, the primary pool is used instead.
+    /// If the replica pool is disabled or unavailable, the primary pool is used instead.
     fn db_read_only(&self) -> Result<DieselPooledConn<'_>, PoolError>;
 }
 
@@ -160,8 +160,18 @@ impl<T: RequestExt + ?Sized> RequestTransaction for T {
     }
 
     fn db_read_only(&self) -> Result<DieselPooledConn<'_>, PoolError> {
-        match &self.app().read_only_replica_database {
-            Some(pool) => pool.get(),
+        let read_only_pool = self.app().read_only_replica_database.as_ref();
+        match read_only_pool.map(|pool| pool.get()) {
+            // Replica is available
+            Some(Ok(connection)) => Ok(connection),
+
+            // Replica is not available, but primary might be available
+            Some(Err(PoolError::UnhealthyPool)) => self.app().primary_database.get(),
+
+            // Replica failed
+            Some(Err(error)) => Err(error),
+
+            // Replica is disabled, but primary might be available
             None => self.app().primary_database.get(),
         }
     }
--- a/src/db.rs
+++ b/src/db.rs
@@ -152,6 +152,11 @@ pub trait RequestTransaction {
     ///
     /// If the replica pool is disabled or unavailable, the primary pool is used instead.
     fn db_read(&self) -> Result<DieselPooledConn<'_>, PoolError>;
+
+    /// Obtain a readonly database connection from the primary pool
+    ///
+    /// If the primary pool is unavailable, the replica pool is used instead, if not disabled.
+    fn db_read_prefer_primary(&self) -> Result<DieselPooledConn<'_>, PoolError>;
 }
 
 impl<T: RequestExt + ?Sized> RequestTransaction for T {
@@ -173,6 +178,25 @@ impl<T: RequestExt + ?Sized> RequestTransaction for T {
 
             // Replica is disabled, but primary might be available
             None => self.app().primary_database.get(),
+        }
+    }
+
+    fn db_read_prefer_primary(&self) -> Result<DieselPooledConn<'_>, PoolError> {
+        match (
+            self.app().primary_database.get(),
+            &self.app().read_only_replica_database,
+        ) {
+            // Primary is available
+            (Ok(connection), _) => Ok(connection),
+
+            // Primary is not available, but replica might be available
+            (Err(PoolError::UnhealthyPool), Some(read_only_pool)) => read_only_pool.get(),
+
+            // Primary failed and replica is disabled
+            (Err(error), None) => Err(error),
+
+            // Primary failed
+            (Err(error), _) => Err(error),
         }
     }
 }
--- a/src/db.rs
+++ b/src/db.rs
@@ -146,7 +146,7 @@ pub fn connection_url(url: &str) -> String {
 
 pub trait RequestTransaction {
     /// Obtain a read/write database connection from the primary pool
-    fn db_conn(&self) -> Result<DieselPooledConn<'_>, PoolError>;
+    fn db_write(&self) -> Result<DieselPooledConn<'_>, PoolError>;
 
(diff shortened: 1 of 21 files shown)


src/db.rs
```diff
@@ -146,25 +146,65 @@ pub fn connection_url(url: &str) -> String {
 
 pub trait RequestTransaction {
     /// Obtain a read/write database connection from the primary pool
-    fn db_conn(&self) -> Result<DieselPooledConn<'_>, PoolError>;
+    fn db_write(&self) -> Result<DieselPooledConn<'_>, PoolError>;
 
     /// Obtain a readonly database connection from the replica pool
     ///
     /// If there is no replica pool, the primary pool is used instead.
```
pietroalbini, line 153: This needs to be updated to mention the primary pool is used even if the replica pool is unavailable.

src/db.rs
```diff
@@ -146,25 +146,65 @@ pub fn connection_url(url: &str) -> String {
...
+        match read_only_pool.map(|pool| pool.get()) {
+            // Replica is available
+            Some(Ok(connection)) => Ok(connection),
+
+            // Replica failed, but primary might be available
+            Some(Err(error)) => {
+                warn!(
+                    %error,
+                    "Failed to retrieve read-only replica database connection. Falling back to primary database."
+                );
```
pietroalbini, line 178: We can't log a warning every time we failover to the other database pool, otherwise our log traffic will double whenever we have a big outage like yesterday's (eventually dropping logs). I agree we need to have visibility when this happens, but we should do so with a metric we increment every time we need to do a failover.

src/db.rs
```diff
@@ -146,25 +146,65 @@ pub fn connection_url(url: &str) -> String {
...
-        match &self.app().read_only_replica_database {
-            Some(pool) => pool.get(),
+    fn db_read(&self) -> Result<DieselPooledConn<'_>, PoolError> {
+        let read_only_pool = self.app().read_only_replica_database.as_ref();
+        match read_only_pool.map(|pool| pool.get()) {
+            // Replica is available
+            Some(Ok(connection)) => Ok(connection),
+
+            // Replica failed, but primary might be available
+            Some(Err(error)) => {
```
pietroalbini, line 174: I'm tempted to say we should failover only when `PoolError::UnhealthyPool` is returned (the amount of connections in the database is 0) instead of any error, otherwise we risk losing visibility into "smaller" bursts of errors.

src/db.rs
```diff
@@ -146,25 +146,65 @@ pub fn connection_url(url: &str) -> String {
...
+            (Err(error), None) => Err(error),
+
+            // Primary failed, but replica might be available
+            (Err(error), Some(read_only_pool)) => {
+                warn!(
+                    %error,
+                    "Failed to retrieve primary database connection. Falling back to read-only replica database."
+                );
+                read_only_pool.get()
+            }
```
pietroalbini, line 205: Same comments as the above method.

pietroalbini (changes requested):

Turbo87:
comments should all be addressed. I've removed the warning for now. we can still add the metrics thing you mentioned later.

pietroalbini:
Could you add a metric as part of this PR? Otherwise we have no visibility when this happens.

src/db.rs
```diff
@@ -146,25 +146,59 @@ pub fn connection_url(url: &str) -> String {
...
+            Some(Err(error)) => Err(error),
+
+            // Replica is disabled, but primary might be available
             None => self.app().primary_database.get(),
         }
     }
+
+    fn db_read_prefer_primary(&self) -> Result<DieselPooledConn<'_>, PoolError> {
+        match (
+            self.app().primary_database.get(),
```
pietroalbini, line 186: Also, just noticed that this returns a read/write connection, not a read-only one. We should make sure connections returned by this method are read-only even when using the primary database, otherwise we might start relying on those being read/write accidentally, causing bugs during failover.

src/db.rs
```diff
@@ -146,25 +146,59 @@ pub fn connection_url(url: &str) -> String {
...
-        match &self.app().read_only_replica_database {
-            Some(pool) => pool.get(),
+    fn db_read(&self) -> Result<DieselPooledConn<'_>, PoolError> {
+        let read_only_pool = self.app().read_only_replica_database.as_ref();
+        match read_only_pool.map(|pool| pool.get()) {
+            // Replica is available
+            Some(Ok(connection)) => Ok(connection),
+
+            // Replica is not available, but primary might be available
+            Some(Err(PoolError::UnhealthyPool)) => self.app().primary_database.get(),
```
pietroalbini, line 174: This should also return a read-only connection.

pietroalbini (changes requested):

src/db.rs
Turbo87: how?

pietroalbini: What we could do is store a `ReadOnlyConnection` unit struct in the [connection extensions](https://docs.rs/r2d2/latest/r2d2/struct.PooledConnection.html#method.extensions_mut), indicating whether that connection is currently read-only (unit struct in the extensions) or not. Then, whenever we get a connection from the database we check whether the presence of the struct is correct (present for a RO conn, missing for a RW conn), and otherwise issue the SQL query to switch the connection mode.

pietroalbini: Other notes that come through mind:

* We need to make sure this doesn't get into an inconsistent state, so we should also have a `LastSwitchFailed` or something to indicate whether the previous attempt to switch between RO and RW failed, indicating it should be performed unconditionally in the next attempt.
* This could replace the "read-only" section in the pool.
* Do we *need* to cache whether the connection is RW or RO, or can we just send the query to switch the connection mode every time we checkout the connection from the pool? That could be way simpler and less failure prone, but I don't know what the performance impact of it would be off my head.

Turbo87: while I generally like the idea of having a type safe way to distinguishing between read-only and writeable database connections, this request is increasing the complexity of this PR by multiple orders of magnitude. I guess ideally this sort of thing would be supported by Diesel itself, instead of us having to monkey patch something like this on top.

I'd appreciate if we could merge this as is for now and then work on further improvements afterwards, instead of blocking this on achieving absolute perfection first. https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good 😉

src/db.rs
Turbo87: ✅

src/db.rs
pietroalbini: While in general I agree, for something this crucial to our reliability I'd like to get it right before we merge it.

pietroalbini:
Thinking more about it, this should also have a chaosproxy test to make sure the fallback actually happens as we want.

Turbo87:
feel free to continue this PR, if you want. my goal was to incrementally improve the codebase, but all these additional requirements just to get a small improvement merged make it impossibly hard to contribute.

PR #4360, opened 2021-12-24
Remove "Filter by letter" feature

Turbo87:
With over 70k crates it has become quite useless to filter the crates by their starting letter, so we might as well remove this feature to declutter the user interface a little bit.


Turbo87:
merging as there are apparently no objections

@bors r+

PR #4382, opened 2021-12-28
Add support for `keyword:cli` and `category:no-std` search filters

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/491

These filters can be used standalone, or combined with regular search terms. The syntax is inspired by https://api-docs.npms.io

<table>
<tr>
<td>Bildschirmfoto 2021-12-29 um 00 16 37
<td>Bildschirmfoto 2021-12-29 um 00 18 00

### DIFF
--- a/app/controllers/search.js
+++ b/app/controllers/search.js
@@ -7,7 +7,7 @@ import { restartableTask } from 'ember-concurrency';
 import { bool, reads } from 'macro-decorators';
 
 import { pagination } from '../utils/pagination';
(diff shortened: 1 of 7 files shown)


pietroalbini:
There should probably be a warning if someone adds multiple categories, other than that r=me.

Turbo87:
> There should probably be a warning if someone adds multiple categories

✅  done

@bors r=pietroalbini

PR #4397, opened 2022-01-02
Remove obsolete `/authorize/github` route

Turbo87:
This route has been replaced with the `github-redirect.html` file instead and is no longer needed.

Todo:
- [x] check if CSP configuration works with the inline JS in `github-redirect.html`


Turbo87:
we have switched production over to the raw HTML page today. our plan is to monitor the situation for about a month and if there are no issues we will merge this PR and remove the obsolete route.

Turbo87:
looks like no issues were reported since then, so I'll go ahead and merge this PR now.

@bors r+

PR #4400, opened 2022-01-02
Adding categories related to Aerospace

### LINKED ISSUE #4105: Adding more categories
Hello :wave:
**Aerospace categories**
I would like to ask about adding more categories regarding `aerospace` in general (from protocol implementations, sdks, apis, simulations, etc).
This issue is more of a discussion on what categories could be added.

**Why?**
I am aware that there are a few categories like `simulations`, `science` & `science::robotics`, etc, however, I would like to have categories specifically for aerospace in order to collect and showcase them in https://areweinspaceyet.org/ automatically.

I'm planning a new version of the website which won't require manually entering crates and instead will build the list of crates for a specific categories and maybe keywords.

**Other options:**
It is possible to use a specific keyword e.g. `#aerorust` that will mark a crate which wants to end up in this list, however, it will still require some categories to be whitelisted and prevent misplaced crates.

I believe that the categories in crates.io could be used by other communities, lists (awesome-lists) and websites like:
- https://www.arewewebyet.org/
- https://areweasyncyet.rs/
- http://www.arewelearningyet.com/
- etc.

elpiel:
Resolves #4105

I've also added the category `Software-defined radio`.
Here are 2 crates that are in this category:

- https://github.com/rsadsb/adsb_deku
- https://github.com/FutureSDR/FutureSDR

I can remove the `SDR` from this PR if it's necessary to discuss it first at a meeting @Turbo87.

CC @wcampbell0x2a @bastibl @MoralCode

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -32,6 +32,47 @@ description = """
 Tools for making your creations usable by as many people as possible. \
 """
 
+[aerospace]
+name = "Aerospace"
+description = """
+Crates for aeronautics (within the atmosphere) and astronautics \
+(in outer space) implications.\
+"""
+
+[aerospace.categories.drones]
+name = "Drones"
+description = """
+Crates related to Multicopters, Fixed wing, VTOL (Vertical Takeoff and Landing) \
+and Airships/Balloons.\
+"""
+
+[aerospace.categories.protocols]
+name = "Aerospace protocols"
+description = """
+Crates of protocol implementations for aerospace implications.\
+"""
+
+[aerospace.categories.simulation]
+name = "Aerospace simulations"
+description = """
+Crates related to any kind of simulations used in aerospace - fluids, \
+aerodynamics, etc.\
+"""
+
+[aerospace.categories.space-protocols]
+name = "Space protocols"
+description = """
+Protocol implementations for implicatoins in space like CCSDS.\
+"""
+
+[aerospace.categories.unmanned-aerial-vehicles]
+name = "Unmanned aerial vehicles"
+description = """
+Crates related to Unmanned aerial vehicles like Multicopters, Fixed wing, \
+VTOL (Vertical Takeoff and Landing), Airships/Balloons, Rovers, Boats, \
+Submersibles.\
+"""
+
 [algorithms]
 name = "Algorithms"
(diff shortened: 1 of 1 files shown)


Turbo87:
> I can remove the `SDR` from this PR if it's necessary to discuss it first at a meeting @Turbo87.

yeah, let's move that out of the PR to unblock the others. I'm not sure yet about a top-level SDR category or whether we should rather move it into "Science" for example. It seems like a very specific one and I'm wondering if there is a way to generalize it a bit more.

elpiel:
> > I can remove the `SDR` from this PR if it's necessary to discuss it first at a meeting @Turbo87.
> 
> yeah, let's move that out of the PR to unblock the others. I'm not sure yet about a top-level SDR category or whether we should rather move it into "Science" for example. It seems like a very specific one and I'm wondering if there is a way to generalize it a bit more.

Ok, it's removed. Thank you!

Turbo87:
Thanks! :)

@bors r+

elpiel:
> Thanks! :)
> 
> @bors r+

No, thank you!
Happy 2022 :yum:

Turbo87:
https://crates.io/categories/aerospace

have fun 😉

MoralCode:
> I'm not sure yet about a top-level SDR category or whether we should rather move it into "Science" for example. It seems like a very specific one and I'm wondering if there is a way to generalize it a bit more.

is there already a `Radio` or `RF` or `Electromagnetic Spectrum` type categories? it seems like those might be some appropriate categories to put SDR underneath (in order of increasing generality)

Turbo87:
@MoralCode the current set of categories is defined at https://github.com/elpiel/crates.io/blob/master/src/boot/categories.toml

MoralCode:
ah yeah I saw that, what I meant was those could be some possible names for generalizations of an SDR category if they don't already exist

elpiel:
> ah yeah I saw that, what I meant was those could be some possible names for generalizations of an SDR category if they don't already exist

Maybe start a new issues to add more categories if you have an idea how they can look?
I'm also not entirely sure about the separation of the 2 categories - Unnamed aerial vehicles & drones so I'm opened for ideas.

PR #4450, opened 2022-01-16
use new cargo fmt check option

calebcartwright:
As of v1.58, cargo fmt now supports the --check flag directly. Trying to get this utilized in relevant places across the r-l repos both because it's more succinct and so more people will see/become aware they no longer have to invoke the check the old way


Turbo87:
cool, thanks! :)

@bors r+

PR #4460, opened 2022-01-19
Use Volta to pin Node.js and yarn versions

Turbo87:
see https://volta.sh


locks:
Volta rocks 👍 

@bors r+

PR #4465, opened 2022-01-20
Use embroider build system by default

Turbo87:
embroider is at v1.0.0 now and has been building a correctly working app for a while now, so it should be viable for us to upgrade now.

see https://github.com/embroider-build/embroider/


Turbo87:
This was extensively tested on the staging environment with no obvious issues.

@bors r+

PR #4496, opened 2022-01-25
EncodableCrate: Simplify `from_minimal()` method

Turbo87:
There is only one user of this method and it is not supplying badges or using the `exact_match` field, so it should be safe to remove them


Turbo87:
lol, looks like I'm wrong. the function is apparently used in more places than my IDE claimed 😅

Issue #4506, opened 2022-01-27
Keyword navigation does not fetch crates data


dgellow:
### Current Behavior

Hi there,

I found out that navigating using hashed keywords (i.e click on a `#router` link) from the crate header does not actually fetch crates data and only display the current crate. It correctly navigates to /keywordi/:keyword but does not fetch data from the server. Additionally, once on /keywordi/:keyword, refreshing the page does yield the expected content (i.e a list of crates matching the keyword).

### Expected Behavior

I expected navigating to /keywordi/:keyword by clicking on a hashed keyword link to be equivalent to manually browsing /keywordi/:keyword. Crates data should be loaded from the server and listed.

### Steps To Reproduce

1. Go to any crate that specifies keywords (here I took a random crate from the homepage, https://crates.io/crates/radixtree)
2. Click on a hashed keyword

![image](https://user-images.githubusercontent.com/2451004/151339516-f44a5cc0-3602-4d91-8275-d6a79cf0637e.png)

3. The current URL should now be https://crates.io/keywords/:keyword. Notice how (1) the number of listed crates is incorrect (1-NaN), (2) only the crate we came from is listed, (3) no request has been triggered to the server.

![image](https://user-images.githubusercontent.com/2451004/151340047-bbe0097e-847d-4749-9328-1e0c4c841016.png)

4. Refresh the current page. Notice how the displayed number of crates is correct and crates are actually listed.

![image](https://user-images.githubusercontent.com/2451004/151340115-265226f8-15f6-4c30-95e5-3fff6162c113.png)

### Environment

- Browser: Latest Edge (v97)
- OS: Windows 11

### Anything else?

I looked in the code to try to fix the issue myself. Unfortunately I do not know ember and could not understand how data are loaded when `<LinkTo @route="keyword" @model={{keyword}}>...</LinkTo>` is used.

Edit: I forgot to mention that the issue can be reproduced locally with `yarn serve`.

Turbo87:
ouch! thanks for reporting. I'll have a look :)

Turbo87:
the fix is deployed to production now :)

PR #4507, opened 2022-01-27
Fix `keyword` route links

Turbo87:
This apparently broke in 8f62c4c 😱

Fixes https://github.com/rust-lang/crates.io/issues/4506


dgellow (approved):
Thanks @Turbo87, that looks good to me 👍

PR #4551, opened 2022-02-11
Fix clippy warnings

teozkr:
Split out from #4548


Turbo87:
I'm a little confused about this, because we're running clippy in CI (see https://github.com/rust-lang/crates.io/blob/master/.github/workflows/ci.yml#L66) and it does not appear to error on these issues. any clue why they show up for you but not in CI?

Turbo87:
I see, thanks

@bors r+

PR #4552, opened 2022-02-11
Explicitly specify the branch name in init-local-index.sh

teozkr:
Otherwise the script breaks if `init.defaultBranch` is not set to `master` (with `main` becoming a common alternative these days).

Split out from #4548.


Turbo87:
sweet, thank you 👍 

@bors r+

PR #4627, opened 2022-03-08
Log sent emails and error out if non-SMTP is used in production

### LINKED ISSUE #4596: In production, panic if email environment variables aren't set
We had some reports that email verification emails weren't arriving. Our best suspicion of the cause is that somehow an environment variable wasn't set (there weren't any config changes logged on the server, so we're not sure how that could have happened).

We should always have email configured in production, so this code: https://github.com/rust-lang/crates.io/blob/dd9893b1291f5d294b4e38a08196a9d0c0f02683/src/email.rs#L17-L36

should never fall back to `EmailBackend::FileSystem` in production, it should panic. The `Env` enum can be used for this: https://github.com/rust-lang/crates.io/blob/dd9893b1291f5d294b4e38a08196a9d0c0f02683/src/lib.rs#L60-L71

cc @rust-lang/crates-io-on-call, this would be super helpful for production support.

### LINKED ISSUE #4597: Add more logging around email sending
Recently, we got some reports of email verification emails not arriving. We looked in mailgun for error reports, but there weren't any logs associated with the email addresses in mailgun's logs, so it seemed like the emails didn't even get to mailgun.

We didn't have any errors in our logs, so we didn't really have much more visibility into any potential problems.

To help debug future problems around email sending, it would be nice if we had logging of:

- Successful email send request, including the mailgun ID
- Failure of email send request, including any error messages

Email sending is around here: https://github.com/rust-lang/crates.io/blob/dd9893b1291f5d294b4e38a08196a9d0c0f02683/src/email.rs#L104-L115

I'm not sure what lettre exposes or what mailgun responds with.

This is related to https://github.com/rust-lang/crates.io/issues/4596, in that in production the email backend should always be `Smtp`, so perhaps the `Env` enum can be used in the sending code as well to ensure this.

cc @rust-lang/crates-io-on-call, this would be super helpful for production support.

pietroalbini:
This PR does three things:

* Refactors custom request metadata to use a thread local instead of a request extension. This was needed to avoid passing `&mut dyn RequestExt` to the email methods (as that's not really clean and caused borrow checker issues).
* Logs the email backend, the message ID (if successful) and the error (if failed) in each request that sends emails.
* Panics if non-SMTP backends are used in production.

Fixes #4596
Fixes #4597


pietroalbini:
Note that I don't know whether Mailgun will override the message ID with its own or not. I tested the change locally with a different mail server I have access to, but I don't know how Mailgun behaves in regards to this.

src/email.rs
```diff
@@ -106,23 +122,42 @@ or go to https://{domain}/me/pending-invites to manage all of your crate ownersh
...
+                            .authentication(vec![Mechanism::Plain])
+                            .build()
+                            .send(&email)
+                    })
+                    .map_err(|e| {
+                        add_custom_metadata("email_error", e);
+                        server_error("Failed to send the email")
+                    })?;
+
+                add_custom_metadata("email_id", message_id);
```
Turbo87, line 146: is there a reason for adding the `email_id` only after the email error was handled? wouldn't it be better to set it as soon as we generated it?

Turbo87 (approved):
feel free to r=me

src/email.rs
pietroalbini: There is no use for the `Message-ID` if Mailgun returned an error: in that case, the mail wouldn't have been sent and wouldn't show up in the Mailgun console, so having the ID offers no practical benefit.

PR #4645, opened 2022-03-17
Set `TCP_USER_TIMEOUT` to 15 seconds for database connections

pietroalbini:
In the December 22, 2021 crates.io outage, we experienced full packet loss between the database server and the application server(s). While the crates.io application supports running without a database during outages, those mitigations kicked in only after a bit more than 15 minutes of the packet loss starting.

The default parameters of the Linux network stack result in a TCP connection being marked as broken after roughly 15 minutes of no acknowledgements being received, which is what I believe happened during that outage.  Broken database mitigations kicking in after 15 minutes is too long for crates.io, as we ideally want those mitigations to kick in as soon as possible (ensuring crates.io users can continue downloading crates).

This PR tells libpq (the underlying PostgreSQL client used by diesel) to set the Linux network stack timeout to a configurable value, which we set as **15 seconds**. With this change, if an outage like the one on Dec 22 2021 one happens again, crates.io will be fully unavailable only for 15 seconds rather than 15 minutes.

Note that the "15 seconds" value is a hunch I had, I'm open on feedback on different values we could use.

This PR intentionally doesn't have any tests: while I have a local commit changing ChaosProxy to simulate packet loss, that's done by shelling out to `iptables`, which is both linux-only and root-only. I am not aware of any cross-platform, unprivileged tool we could use to simulate this in our test suite unfortunately.

Huge thanks to @weiznich in https://github.com/diesel-rs/diesel/issues/3016 for helping debug this issue and for trying possible fixes on the diesel side.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -9,6 +9,7 @@ use thiserror::Error;
 use url::Url;
 
 use crate::middleware::app::RequestApp;
+use crate::{config, Env};
 
 #[derive(Clone)]
 pub enum DieselPool {
@@ -22,10 +23,11 @@ pub enum DieselPool {
 impl DieselPool {
     pub(crate) fn new(
         url: &str,
-        config: r2d2::Builder<ConnectionManager<PgConnection>>,
+        config: &config::Server,
+        r2d2_config: r2d2::Builder<ConnectionManager<PgConnection>>,
         time_to_obtain_connection_metric: Histogram,
     ) -> Result<DieselPool, PoolError> {
-        let manager = ConnectionManager::new(connection_url(url));
+        let manager = ConnectionManager::new(connection_url(config, url));
 
         // For crates.io we want the behavior of creating a database pool to be slightly different
         // than the defaults of R2D2: the library's build() method assumes its consumers always
@@ -39,7 +41,7 @@ impl DieselPool {
         // establish any connection continue booting up the application. The database pool will
         // automatically be marked as unhealthy and the rest of the application will adapt.
         let pool = DieselPool::Pool {
-            pool: config.build_unchecked(manager),
+            pool: r2d2_config.build_unchecked(manager),
             time_to_obtain_connection_metric,
         };
         match pool.wait_until_healthy(Duration::from_secs(5)) {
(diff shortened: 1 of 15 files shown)


weiznich:
>  With this change, if an outage like the one on Dec 22 2021 one happens again, crates.io will be fully unavailable only for 15 seconds rather than 15 minutes.

Just to make restate the obvious point: `TCP_USER_TIMEOUT` refers to the timeout for a individual network request. That does not necessarily mean that after exactly 15 seconds everything will be fine. That depends on more factors. As crates.io use a connection pool the worst case scenario would be something like:
An operation tries to checkout a connection from the pool. -> The pool runs in the 15s timeout for the first connection, so it tries the next one and so on -> After the pool runs out of connections/time it will raise an error, but not earlier. That could be easily more than 15 seconds depending on the pool configuration.  That means `TCP_USER_TIMEOUT` is just only one part that can help preventing such issues again.

pietroalbini:
Thanks for your comment weiznich! That is indeed true in theory. In practice, with the traffic patterns we get in crates.io, all connections in the pool will be requested roughly at the same time (we don't get a steady stream of requests, we get bursts of 100+ parallel requests whenever someone runs `cargo build`), causing all of them to time out in parallel.

src/db.rs
```diff
@@ -131,16 +133,19 @@ impl Deref for DieselPooledConn<'_> {
...
     PgConnection::establish(&url)
 }
 
-pub fn connection_url(url: &str) -> String {
+pub fn connection_url(config: &config::Server, url: &str) -> String {
     let mut url = Url::parse(url).expect("Invalid database URL");
-    if dotenv::var("HEROKU").is_ok() && !url.query_pairs().any(|(k, _)| k == "sslmode") {
+
+    // Enforce secure connections in production.
+    if config.base.env == Env::Production && !url.query_pairs().any(|(k, _)| k == "sslmode") {
```
Turbo87: seems a bit unfortunate for the database code to have to be aware of all the other server config values. what do you think about decoupling this a bit and adding something like `sslMode` to the database config struct instead?

pietroalbini: I don't see much of a problem personally of passing the whole configuration, but I can refactor things to pass only the database configuration.

pietroalbini: Pushed a new commit to request only the database pools config instead of the whole server config. Let me know if this is what you intended here :slightly_smiling_face:

Turbo87 (approved):
LGTM thanks! :)

feel free to r=me

PR #4649, opened 2022-03-18
Add an environment variable to let us exclude crate IDs from the recent downloads list

carols10cents:
I haven't figured out a great way to add a test for this because I need to create a crate and get a crate ID and then change the app's config, and the test app doesn't easily have this... so I thought I'd throw up this change for discussion and see if anyone had better ideas.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -25,6 +25,7 @@ pub struct Server {
     pub blocked_traffic: Vec<(String, Vec<String>)>,
     pub max_allowed_page_offset: u32,
     pub page_offset_ua_blocklist: Vec<String>,
+    pub excluded_crate_ids: Vec<i32>,
     pub domain_name: String,
     pub allowed_origins: Vec<String>,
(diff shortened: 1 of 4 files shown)


Turbo87:
would it be possible to filter by crate name instead? or is the name not available in that table?

carols10cents:
The query does join to the `crates` table, so the crate name is available

Turbo87:
in that case I think you can start a test app with a random crate name in the new config value and then create that crate after the app has started, right?

Turbo87:
given that the particular crate now has almost 1/4 of the downloads of `bitflags` in total, do we also want to exclude it from the "Most downloaded" list?

carols10cents:
Good calls. Updated!

Turbo87:
looks great, thank you! 

one potential follow up: we allow the users to view the list of crates sorted by downloads too, but we don't exclude the crates there. I'm not sure how or if we should deal with that too.

merging anyways since this is a good first step :)

@bors r+

PR #4661, opened 2022-03-22
Upload index metadata to `index/` when publishing new crates

arlosi:
Cargo can access http-based registries via https://github.com/rust-lang/cargo/pull/10470.

This change causes crates.io to publish any changed metadata files to `index/` on S3 in addition to the git-based index. The S3 bucket is configured by new environment variables `S3_INDEX_*`.

A new admin tool for bulk-uploading crates is also added.

### DIFF
--- a/src/admin/upload_index.rs
+++ b/src/admin/upload_index.rs
@@ -1,0 +1,60 @@
+use std::time::{Duration, Instant};
+
+use crate::admin::dialoguer;
+use cargo_registry_index::{Repository, RepositoryConfig};
+use reqwest::blocking::Client;
+
+use crate::config;
+
+#[derive(clap::Parser, Debug)]
+#[clap(
+    name = "upload-index",
+    about = "Upload index from git to S3 (http-based index)"
+)]
+pub struct Opts {
+    /// Incremental commit. Any changed files made after this commit will be uploaded.
+    incremental_commit: Option<String>,
+}
+
+pub fn run(opts: Opts) -> anyhow::Result<()> {
+    let config = config::Base::from_environment();
+    let uploader = config.uploader();
+    let client = Client::new();
+
+    println!("fetching git repo");
+    let config = RepositoryConfig::from_environment();
+    let repo = Repository::open(&config)?;
+    repo.reset_head()?;
+    println!("HEAD is at {}", repo.head_oid()?);
+
+    let files = repo.get_files_modified_since(opts.incremental_commit.as_deref())?;
+    println!("found {} files to upload", files.len());
+    if !dialoguer::confirm("continue with upload?") {
+        return Ok(());
+    }
+
+    let mut progress_update_time = Instant::now();
+    for (i, file) in files.iter().enumerate() {
+        let crate_name = file.file_name().unwrap().to_str().unwrap();
+        let path = repo.index_file(crate_name);
+        if !path.exists() {
+            println!("skipping file `{}`", crate_name);
+            continue;
+        }
+        let contents = std::fs::read_to_string(&path)?;
+        uploader.upload_index(&client, crate_name, contents)?;
+
+        // Print a progress update every 10 seconds.
+        let now = Instant::now();
+        if now - progress_update_time > Duration::from_secs(10) {
+            progress_update_time = now;
+            println!("uploading {}/{}", i, files.len());
+        }
+    }
+
+    println!(
+        "uploading completed; use `upload-index {}` for an incremental run",
+        repo.head_oid()?
+    );
+    Ok(())
+}
--- a/src/config/base.rs
+++ b/src/config/base.rs
@@ -74,7 +74,7 @@ impl Base {
 
     pub fn test() -> Self {
         let uploader = Uploader::S3 {
-            bucket: s3::Bucket::new(
+            bucket: Box::new(s3::Bucket::new(
                 String::from("alexcrichton-test"),
                 None,
                 dotenv::var("S3_ACCESS_KEY").unwrap_or_default(),
@@ -82,7 +82,16 @@ impl Base {
                 // When testing we route all API traffic over HTTP so we can
                 // sniff/record it, but everywhere else we use https
                 "http",
-            ),
+            )),
+            index_bucket: Some(Box::new(s3::Bucket::new(
+                String::from("alexcrichton-test"),
+                None,
+                dotenv::var("S3_INDEX_ACCESS_KEY").unwrap_or_default(),
+                dotenv::var("S3_INDEX_SECRET_KEY").unwrap_or_default(),
+                // When testing we route all API traffic over HTTP so we can
+                // sniff/record it, but everywhere else we use https
+                "http",
+            ))),
             cdn: None,
         };
         Self {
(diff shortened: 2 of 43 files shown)


src/uploaders.rs
```diff
@@ -82,6 +83,17 @@ impl Uploader {
         format!("readmes/{name}/{name}-{version}.html")
     }
 
+    /// Returns the internal path of an uploaded crate's index file.
+    fn index_path(name: &str) -> String {
```
jtgeibel, line 87: This should be replaced with a call to `relative_index_file` in `lib.rs`.

src/worker/git.rs
```diff
@@ -17,12 +17,21 @@ pub fn add_crate(env: &Environment, krate: Crate) -> Result<(), PerformError> {
...
+        .open(&dst)?;
     serde_json::to_writer(&mut file, &krate)?;
     file.write_all(b"\n")?;
 
-    let message: String = format!("Updating crate `{}#{}`", krate.name, krate.vers);
+    file.rewind()?;
+    let mut contents = String::new();
+    file.read_to_string(&mut contents)?;
+    env.uploader
+        .upload_index(env.http_client(), &krate.name, contents)?;
```
jtgeibel, line 32: Here the HTTP index is being updated before the index changes are pushed to GitHub. At least for now, I would like to keep the git index repository as the authoritative index, so we should make sure that the git push succeeds before updating the HTTP index. I think the best solution would be to enqueue another background job that copies the file from the index over the S3. On the rare occasion that S3 goes down between the original publish and this task running, then it will be retried until successful and will be visible with our existing monitoring.

jtgeibel (changes requested):
Thanks for the PR @arlosi!

There are a few operational issues that we'll need to address before this can be merged.

* We need to invalidate the CloudFront cache for a file after modifying it on S3. Even with a short cache-control interval files will enter the cache at different times and clients could end up seeing an inconsistent index state (such as a crate with a dependency that does not yet appear to be published). We could then probably increase the max-age allowed to be cached by CF.
* These files should probably be stored in their own S3 bucket. We should also serve them from something like `index.crates.io` instead of `static.crates.io`. I'll discuss with infra at our next meeting.
* We need a way to populate the full repository, as these changes will only upload an index file upon a publish. A command could be added to `src/admin` that publishes all of the crate files. There will probably be some complexity around getting this right, since if there is a gap between this task and enabling it in production then updates will be missed, but if there is overlap then updates could be overwritten with older data.
* Similarly, we should add logic to the `delete_crate` admin task to delete the index file from S3 and invalidate the CF cache. This happens pretty rarely and we don't automatically update the git index yet, but it would be helpful update the HTTP index automatically so that we don't forget.

jtgeibel:
> S3 has a atomic upload a file if not exist command.

That sounds like a great solution!

> One overarching thought, the cargo side of this is not ready for stabilization. [...] As long as we have a plan in place, and each step gets us closer to it, we can take it one step at a time.

Thanks for the additional context. I agree that at this point we just need to agree on a plan for long term operations, and can begin iterating towards that. 

> invalidations only happened so quickly. If max-age is shorter then the invalidation roll out time, then there is no point.

Okay, it looks like "[o]bject invalidations typically take from [60 to 300] seconds to complete." The PR currently sets the max-age to 600 seconds. For projects that release multiple crates they expect each crate to be published within seconds (to publish crates which depend on it), not minutes, so I wonder if we would need to set a very short max-age and only get the benefits of caching for very popular crates. I'd love to have some estimates of expected traffic and the associated costs of various options to base these design decisions on.

> I don't think inconsistencies are going to be a big problem, but happy to talk it over.

My main concern here is also related to new publishes (which will probably be solved by however we solve the above). If I publish new major versions of a batch of crates that depend on each other, then currently there would be a period of up to 10 minutes where some users may have a broken build if they attempt to upgrade. In contrast, the whole git index is updated atomically and we can ensure that for every dependency in the index cargo will find at least one crate version that satisfies it. (Currently I think the only exceptions are if a crate is deleted, or if versions are yanked without a compatible replacement.)

[60 to 300]: https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-serving-outdated-content-s3/

Eh2406:
> S3 has a atomic upload a file if not exist command.

Well it has a "atomic copy if not exist" command. [`x-amz-copy-source-if-unmodified-since` on CopyObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html). For some reason that header is not available on upload.

> The PR currently sets the max-age to 600 seconds.

That is sort of in the awkward middle. If we want to ensure updates take less than say 30 seconds, then a short max-age is the only way to do that. If we are comfortable with updates occasionally taking over 300 seconds, then an infinitely long max-age and invalidations are the best way to do that. I'm not sure what's best, but my understanding is that we can change this pretty easily at anytime.

> I'd love to have some estimates of expected traffic and the associated costs of various options to base these design decisions on.

At the moment the traffic should be approximately 0, has the feature is not even available on nightly. As we stabilize it it should eventually grow too match current traffic for cloning the index. I don't know if the numbers I got from @pietroalbini can be shared publicly. I will start a DM on zulip if that works for you.

> If I publish new major versions of a batch of crates that depend on each other, then currently there would be a period of up to 10 minutes where some users may have a broken build if they attempt to upgrade.

Yes, all the oddities occur when trying to build a package within one max-age of a publish.
In general the resolver will pick an older version if it matches the requirements, or give an error message about the dependency not being available. Resolver error messages are not the best, sadly. But they should at least list witch package is out of date. Perhaps an HTTP header can tell Cargo the max-age and it can get added to the error message. "but no versions of `foo` matched `bars` requirement `^1.0.0` ... Note: versions published in the past 300 seconds my not yet be available." or something.

Even with gits nice atomic properties, even if a version does match every requirement that does not mean that every requirement can be satisfied. Every version that matches could conflict with some other requirement, or the lockfile.

pietroalbini:
Created buckets and CDNs both for staging and production, the existing credentials have access to them:

* Staging: `staging-crates-io-index` / https://index.staging.crates.io
* Production: `crates-io-index` / https://index.crates.io

src/uploaders.rs
arlosi: Done. I had to add a new function, since we don't actually want a `Path` here (we need a string).

arlosi:
> we should add logic to the delete_crate admin task to delete the index file from S3 and invalidate the CF cache. 

The `delete_crate` admin task doesn't seem to modify the git index (only the database). Maybe I'm missing something?

Eh2406:
You are not missing anything. To quote:

> Similarly, we should add logic to the delete_crate admin task to delete the index file from S3 and invalidate the CF cache. This happens pretty rarely and we don't automatically update the git index yet, but it would be helpful update the HTTP index automatically so that we don't forget.

arlosi:
Oops, I misread the request. The `delete_crate` admin task now removes the http index metadata as well.

arlosi:
Is someone available to review this? Let me know if there are changes needed.

src/uploaders.rs
```diff
@@ -90,17 +100,28 @@ impl Uploader {
...
+                ref bucket,
+                ref index_bucket,
+                ..
+            } => {
+                let bucket = match upload_bucket {
+                    UploadBucket::Index => index_bucket,
+                    UploadBucket::Default => bucket,
+                };
+                let content_length = content.seek(SeekFrom::End(0))?;
+                content.seek(SeekFrom::Start(0))?;
```
Turbo87, line 123: what's the reason for the `content_length` change here? on first glance it looks like a slightly unrelated change, which we'd ideally merge and deploy separately to lower the risks. does that make sense?

src/admin/render_readmes.rs
```diff
@@ -111,25 +110,10 @@ pub fn run(opts: Opts) -> anyhow::Result<()> {
...
                     .uploader()
-                    .upload(
-                        &client,
-                        &readme_path,
-                        content,
-                        content_length,
-                        "text/html",
-                        extra_headers,
-                    )
+                    .upload_readme(&client, &krate_name, &version.num, readme)
```
Turbo87, line 116: `upload_readme()` appears to use a different `Cache-Control` header, if I understand this correctly. admittedly, when we upload first via `upload_readme()` with an immutable cache policy then rerendering and uploading with a different header appears a bit pointless, but it still looks like a potentially significant change. do you think you can extract that refactoring into a dedicated pull request so that we can merge and deploy it separately? :)

src/config/base.rs
```diff
@@ -97,26 +106,40 @@ impl Base {
...
+            bucket: Box::new(s3::Bucket::new(
                 env("S3_BUCKET"),
                 dotenv::var("S3_REGION").ok(),
                 env("S3_ACCESS_KEY"),
                 env("S3_SECRET_KEY"),
                 "https",
-            ),
+            )),
+            index_bucket: Box::new(s3::Bucket::new(
+                env("S3_INDEX_BUCKET"),
```
Turbo87, line 117: quoting from the env sample file:

> You can leave these commented out if you're not publishing to s3 from your crates.io instance.

but if I understand the `env()` function correctly it panics if the environment variable is not set. can we make `index_bucket` an `Option<_>` for now, so that we can easily disable it in production if necessary?

src/admin/upload_index.rs
```diff
@@ -0,0 +1,50 @@
...
+    for file in files {
+        let crate_name = file.file_name().unwrap().to_str().unwrap();
+        let path = repo.index_file(crate_name);
+        if !path.exists() {
+            println!("skipping file `{}`", crate_name);
+            continue;
+        }
+        let contents = std::fs::read_to_string(&path)?;
+        uploader.upload_index(&client, crate_name, contents)?;
+    }
```
Turbo87, line 43: do you have any estimates how long this operation might take? would it make sense to add some sort of progress logging to the loop?

src/uploaders.rs
arlosi: I can move it into a separate change, but it will have to go in before this one to keep clippy happy.

src/admin/upload_index.rs
arlosi: It needs to upload ~83000 small files to S3. I would guess around an hour, but of course it depends on the internet connection.

I'll add progress logging.

src/uploaders.rs
Turbo87: > clippy said there were too many arguments

I see, thanks :)

Turbo87 (approved):
I'm approving and merging this with the understanding that this is an experimental feature without any stability guarantees for now. 

As @jtgeibel pointed out in https://github.com/rust-lang/crates.io/pull/4661#pullrequestreview-923848551, there are still a few issues to solve for this implementation, but we agree that we can solve these in an incremental way while the feature is still considered experimental.

PR #4720, opened 2022-04-11
Add `rs` alias for `rust` syntax highlighting

nihaals:
As GitHub supports `rs`, many projects seem to use `rs` without realising crates.io doesn't support it, and may either never notice or notice after publishing a version.

[Rough search for examples](https://sourcegraph.com/search?q=context%3Aglobal+repo%3Acontains.file%28Cargo%5C.toml%29+file%3A%5EREADME+%60%60%60rs&patternType=literal).

A point against this PR is that it might also make sense to intentionally have as few aliases as possible to encourage consistency, so supporting `rs` might add more of a split between the two.


Turbo87:
Sounds reasonable if GitHub supports it too. Thanks!

@bors r+

PR #4779, opened 2022-05-04
Improve database fallbacks

justahero:
This PR is based on @Turbo87 PR #4316 to improve database fallbacks.

The original PR (#4316) contains all db related changes, while this PR adds a set of tests using the ChaosProxy to check database switch overs. Thanks to @pietroalbini for the fix in the ChaosProxy to make the tests work as intended.

The changes are
* implements a primary database fallback if the replica is unavailable
* renames the existing connection pool methods to `db_read()` and `db_write()`
* implements an additional `db_read_prefer_primary()` method
* uses that new method for endpoints that prefer the primary database, but don't need to write anything to the database
* creates primary & replica connections in `ChaosProxy`
* fixes closing connection in `ChaosProxy`
* adds tests to check database switch over behavior

Thank you so much for doing the work @Turbo87. :+1:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,7 +5,7 @@ mod base;
 mod database_pools;
 
 pub use self::base::Base;
-pub use self::database_pools::DatabasePools;
+pub use self::database_pools::{DatabasePools, DbPoolConfig};
 use std::collections::HashSet;
 use std::time::Duration;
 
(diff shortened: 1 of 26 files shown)


Turbo87 (approved):
thank you! LGTM on first glance.

I'll keep this open until tomorrow in case @jtgeibel has any objections.

Turbo87:
@justahero FYI we've deployed this to prod now

PR #4806, opened 2022-05-11
Use env var to support blocking IP ranges

justahero:
This PR adds support for the new `WEB_PAGE_OFFSET_CIDR_BLOCKLIST` environment variable that holds a list of IPv4 based CIDR blocks. The use case of this env var is to block IPs coming from the same network or IP range. It should help to mitigate scenarios where a high number of distributed requests from multiple clients put a high load onto the application (e.g. bot network) which happen to come from the same IP range. The new block list is meant to complement the existing User Agent & IP block lists.

**Note** the functionality is similar to the `WEB_PAGE_OFFSET_UA_BLOCKLIST` env var which blocks requests by matching user agents. It also only applies to the same (expensive) API endpoints that fetch data using pagination. The common experience using the web site should not be affected. The list is specified as a comma separated list of IPv4 based CIDR blocks (e.g. `"192.16.0.0/16"`).

Currently the CIDR block list only supports IPv4 based entries, each block requires at least a host prefix of 16 bits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ cargo-registry-index = { path = "cargo-registry-index" }
 cargo-registry-markdown = { path = "cargo-registry-markdown" }
 cargo-registry-s3 = { path = "cargo-registry-s3" }
 chrono = { version = "=0.4.19", features = ["serde"] }
+cidr-utils = "=0.5.6"
 clap = { version = "=3.1.18", features = ["derive", "unicode"] }
 
 conduit = "=0.10.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,6 @@ cargo-registry-index = { path = "cargo-registry-index" }
 cargo-registry-markdown = { path = "cargo-registry-markdown" }
 cargo-registry-s3 = { path = "cargo-registry-s3" }
 chrono = { version = "=0.4.19", features = ["serde"] }
-cidr-utils = "=0.5.6"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -35,6 +35,7 @@ cargo-registry-index = { path = "cargo-registry-index" }
 cargo-registry-markdown = { path = "cargo-registry-markdown" }
 cargo-registry-s3 = { path = "cargo-registry-s3" }
 chrono = { version = "=0.4.19", features = ["serde"] }
+cidr-utils = "=0.5.6"
```
Turbo87, line 38: would it be possible to use https://github.com/achanda/ipnetwork instead? in that case it would be compatible with `diesel` (see https://github.com/diesel-rs/diesel/blob/master/diesel/Cargo.toml#L27) if we ever needed to store these in the database.

Turbo87 (commented):
is there a technical reason why you only implemented it for IPv4?

Cargo.toml
justahero: Replaced [cidr-utils](https://crates.io/crates/cidr-utils) with the [ipnetwork](https://github.com/achanda/ipnetwork) crate.

Turbo87:
yeah, I think it would probably even simplify the code if used generic CIDRs. unless there are any major downsides that I'm missing I think it'd be best to slightly tweak it to allow both :)

Turbo87 (approved):
LGTM, thank you! :)

PR #4807, opened 2022-05-12
Add `Virtualization` category to crates.io

eomain:
This PR proposes the addition of a virtualization category on crates.io.
Currently, there does not seem to be a category dedicated to code concerning virtualization
technology such as containerization, VMs, resource management, etc.
Some crates that would usually fall within this category have opted for others like emulation,
which while somewhat related, can be distinct from virtualization.
Any thoughts on this?

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -530,7 +530,8 @@ potentially adapting the display to various languages and regions.\
 [virtualization]
 name = "Virtualization"
 description = """
-For the creation and management of virtual environments.\
+For the creation and management of virtual environments and resources
+of any form including containerization systems.\
 """
 
 [visualization]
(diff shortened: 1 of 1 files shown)


jtgeibel:
Thanks for the PR @eomain. This sounds like a good addition to me. In the PR you also mention containerization solutions, so I wonder if it makes sense to tweak the description a bit to explicitly mention this. Though I'm also fine merging this as-is.

eomain:
@jtgeibel I did consider mentioning this in the description but thought it may have been too specific seeing as virtualization can be quite broad. Although it may be good to have as an
example, so I am okay with tweaking. Or maybe containerization can be a subcategory.

JohnTitor:
I think it's fine to tweak the description for now, when containerization-related crates become popular in the "virtualization" category and need a sub-category, we could consider it.

JohnTitor (approved):

Turbo87 (approved):
sounds good to me. thanks @eomain for this suggestion, and thank you everyone else for the quick review :)

PR #4812, opened 2022-05-17
Refactor `upload` to avoid passing content_length

arlosi:
Use `seek()` on the `content` passed to `upload` rather than separately passing in a `content_length`.

r? @Turbo87


Turbo87 (approved):
It looks like all of our uploader usages already have the payload in memory anyway and we're not streaming it directly to S3, so I guess this change will not cause any performance issues.

thanks :)

PR #4825, opened 2022-05-20
Simplify CIDR parsing code

Turbo87:
This PR simplifies the CIDR parsing code a little bit by passing the CIDR strings one-by-one to the `parse_cidr_block()` function, instead of passing in the whole `Vec`. It also refactors `parse_cidr_block()` to return a `Result` instead of panicking directly.

/cc @justahero


justahero (approved):
LGTM :+1:

PR #4826, opened 2022-05-20
Drop index/ prefix when uploading crate metadata.

arlosi:
The metadata is already uploaded into a dedicated bucket, so the additional index/ prefix is unnecessary.


Turbo87 (approved):
👍

PR #4830, opened 2022-05-22
Allow organization owners to add teams from their org to crates

pietroalbini:
Right now, to be able to add a GitHub team to the list of owners of a crate, you have to both be an existing owner of the crate and to be a member of the team you want to add. The second protection is important to prevent people from inviting random teams to their crates, as we don't have an invite system for teams.

Still, that restriction gets annoying when the user adding a team to a crate is an owner of the GitHub organization the team is defined in. The user still *can* add the team, but they first need to add themselves to the team in question (and they can do so, since they're owners of the organization), which is an unnecessary annoyance.

This PR was prompted by having to add a team in the `rust-lang` organization to a crate, when the existing owner is @rust-lang-owner. In the future infra will likely want to automate adding and removing teams from crates (based on the team repo), and this limitation would prevent implementing that.

Note that this PR does **not** allow organization owners to publish crates owned by one of their org's teams, as that change might require more scrutiny.

---

In addition to implementing the change, I refactored how GitHub teams test work by removing the need to use the recording proxy to assert the right requests were made. Instead, I put the GitHub client behind a trait and mocked that with sample data for tests. This should also allow us to remove https://github.com/crates-test-org, @crates-tester-1 and @crates-tester-2 from the real GitHub.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -26,6 +26,12 @@ pub trait GitHubClient: Send + Sync {
         username: &str,
         auth: &AccessToken,
     ) -> AppResult<GitHubTeamMembership>;
+    fn org_membership(
+        &self,
+        org_id: i32,
+        username: &str,
+        auth: &AccessToken,
+    ) -> AppResult<GitHubOrgMembership>;
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


JohnTitor (approved):
The enhancement itself makes sense to me and refactored tests look awesome!

Turbo87:
@pietroalbini can you move the test refactoring to a dedicated PR? that would make it quite a bit easier to review those changes independently :)

pietroalbini:
Extracted the refactor in https://github.com/rust-lang/crates.io/pull/4855.

pietroalbini:
Rebased.

Turbo87 (approved):
looks reasonable to me. thanks :)

PR #4836, opened 2022-05-23
admin/upload_index: Use `indicatif` for progress reporting

Turbo87:
This PR uses https://github.com/console-rs/indicatif to display a progress bar while backfilling the HTTP-based index, with the main advantage of having a somewhat reasonable ETA display :)

/cc @arlosi

### DIFF
--- a/src/admin/upload_index.rs
+++ b/src/admin/upload_index.rs
@@ -1,7 +1,6 @@
-use std::time::{Duration, Instant};
-
 use crate::admin::dialoguer;
 use cargo_registry_index::{Repository, RepositoryConfig};
+use indicatif::{ProgressBar, ProgressIterator, ProgressStyle};
 use reqwest::blocking::Client;
 
 use crate::config;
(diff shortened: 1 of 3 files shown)


arlosi:
Just a consideration: I'm not sure where the admin tasks run, but I'd expect it would need to support clearing the current line for a progress bar to work properly.

If we really get impatient, we could upload in parallel and probably get at least a 10x speedup 😀.

src/admin/upload_index.rs
```diff
@@ -33,23 +32,18 @@ pub fn run(opts: Opts) -> anyhow::Result<()> {
...
-    let mut progress_update_time = Instant::now();
-    for (i, file) in files.iter().enumerate() {
+    let pb = ProgressBar::new(files.len() as u64);
+    pb.set_style(ProgressStyle::default_bar().template("{bar:60} ({pos}/{len}, ETA {eta})"));
+
+    for file in files.iter().progress_with(pb) {
         let crate_name = file.file_name().unwrap().to_str().unwrap();
         let path = repo.index_file(crate_name);
         if !path.exists() {
-            println!("skipping file `{}`", crate_name);
```
arlosi, line 41: Why remove this?

Turbo87: it looks like when we call `println!()` while a progressbar is active it does not print the message, but instead freezes the progressbar and continues it on the next line. I guess it doesn't play well with the clearing of the current line.

JohnTitor (approved):
The motivation looks reasonable to me!

Turbo87:
> I'm not sure where the admin tasks run, but I'd expect it would need to support clearing the current line for a progress bar to work properly.

I tried it out on Heroku and it seems to work as expected :)

PR #4857, opened 2022-06-02
claim: Use git dependency

Turbo87:
This should resolve a version conflict with our `dashmap` dependency and the transitive `autocfg` dependency.

see https://github.com/svartalf/rust-claim/pull/10 and https://github.com/rust-lang/crates.io/pull/4763


jerusdp:
Thanks

PR #4861, opened 2022-06-02
router: Remove `/api/v1/` subrouter

Turbo87:
Instead of having a subrouter, we can just add the routes directly to the primary router and simplify the code a little bit.

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -11,112 +11,134 @@ use crate::util::EndpointResult;
 use crate::{App, Env};
 
 pub fn build_router(app: &App) -> RouteBuilder {
-    let mut api_router = RouteBuilder::new();
+    let mut router = RouteBuilder::new();
 
     // Route used by both `cargo search` and the frontend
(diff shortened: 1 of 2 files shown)


JohnTitor:
I'm a bit concerned that this would introduce some complexity, for example, it'd be easier to misspell/misconfigure a path (at least for me). I'd like to understand the purpose of this PR correctly, could you elaborate a bit more?

Turbo87:
> I'd like to understand the purpose of this PR correctly, could you elaborate a bit more?

The primary issue is that the `path` mutation is causing Papertrail and Sentry to show incorrect/incomplete request paths.  We currently only have `/api/v1` but if we had another prefix with similar paths then they would show up with the same paths. The same also applies to our per-route request blocking, which currently needs to skip the `/api/v1` prefix to work properly.

JohnTitor:
That totally makes sense to me, thanks for clarifying!

JohnTitor (approved):

PR #4879, opened 2022-06-08
Add a -y option for scripting of delete crate/delete version admin utils

carols10cents:
If `-y` is specified, don't ask for interactive confirmation.

There are a bunch of crates that were published in violation of our policies, and I want to run the admin delete-crate utility as part of a script and not have to manually hit `y` all the time :)


Turbo87 (approved):
👍

PR #4892, opened 2022-06-15
Port to diesel 2.0.3

jtgeibel:
With the latest upstream diesel changes in the 2.0.3 release, our final failing test is now passing!

#### Original PR Text

This is a nearly complete port to diesel 2, with the following exceptions:

* It relies on git branches of `diesel_full_text_search` (an upstream PR) and `swirl` (a development branch [here](https://github.com/jtgeibel/swirl/commit/cca24585daa89c856f1e4523c07e45b7a1890c9c)).
* Due to the upstream change from `&PgConnection` to `&mut PgConnection`, it seems to no longer be possible to run the test suite with a single connection, as required by our background job tests. The current design uses a re-entrant mutex , but this no longer works and unless I'm missing something, there is (for good reason) no safe API to hand out multiple `&mut` to the same location even if we're on the same thread. Maybe we can rework the `swirl` backend to add support for our test harness.
* A temporary workaround to the previous issue is in [jtgeibel/crates.io@diesel2-crimes](https://github.com/jtgeibel/crates.io/commit/a69501a045b9c289495934bd4f83b6844164d8dd) which switches to a regular mutex which fails upon the second attempt to take the lock. This allows the branch to build and most tests pass. There are currently 60 test failures, though most of these are due this hack.

Any suggestions on how to resolve this are greatly appreciated!


weiznich:
To just leave a comment here: Feel free to reach out to me for anything diesel related. I'm happy to provide any additional information that would be required for this update (or any other feature).
Additionally you might want to wait till the final 2.0 release as the RC contains an issue around r2d2 pools: https://github.com/diesel-rs/diesel/issues/3156

That would also allow us to release an updated `diesel_full_text_search` at that point.

weiznich:
@jtgeibel Is there anything I can help with here?

jtgeibel:
Thanks @weiznich. I've finally been able to make some good progress with updating the background job logic to reuse an existing database connection in [a85f62c](https://github.com/rust-lang/crates.io/pull/4892/commits/a85f62c2329182749c0b195e7e28fabe7e52ac43). I'm a little unsure about the `AssertUnwindSafe<&mut PgConnection>` and if this could possibly leave the connection in a bad state during an unwind. Though I've tweaked an existing test to now panic and it seems to work as expected. Your review of that commit would be much appreciated.

Other than that, I think this is finally getting close to moving out of draft. There are just a few remaining test failures that I haven't had a chance to look into yet. And we'll need a better long-term fix for [cb19137](https://github.com/rust-lang/crates.io/pull/4892/commits/cb19137e7724e598350485cb6065661820f366e6). (cc @Turbo87 the new token scopes / auth is causing a re-entrant access to the test connection in the publish endpoint. For now I just drop the connection before auth and obtain a new one after.)

weiznich:
:+1: I will have a look at https://github.com/rust-lang/crates.io/pull/4892/commits/a85f62c2329182749c0b195e7e28fabe7e52ac43 next week

weiznich:
I left a comment on the implications of `AssertUnwindSafe` on the corresponding commit: https://github.com/rust-lang/crates.io/commit/a85f62c2329182749c0b195e7e28fabe7e52ac43#r95034771

jtgeibel:
Thanks @weiznich! I've decided to break off this part of the change into #5837 and have incorporated your suggestions (backported to diesel 1.0) in [2e3637d](https://github.com/rust-lang/crates.io/pull/5837/commits/2e3637d6aee0eead13324331b4e0d0a06747561e).

jtgeibel:
I've rebased and squashed this PR on top of latest master. There are 5 failing `unhealthy_database` tests. I haven't looked into those yet. I also plan to look at improving the connection rollback logic now that diesel2 has built-in safeguards. Otherwise, this is ready for review.

weiznich:
> I also plan to look at improving the connection rollback logic now that diesel2 has built-in safeguards. 

The rollback code is [here](https://github.com/diesel-rs/diesel/blob/49b39fc9f92018b41b48bba2b9401dad953899fe/diesel/src/connection/transaction_manager.rs#L332). The main change there is that we now handle cases where we cannot rollback to safe points and have a much stricter tracking of the transaction depth. In addition we've introduced additional [checks](https://github.com/diesel-rs/diesel/blob/49b39fc9f92018b41b48bba2b9401dad953899fe/diesel/src/connection/transaction_manager.rs#L83) to see whether a transaction manager contains open transactions or not. These information is used by the [`r2d2`](https://github.com/diesel-rs/diesel/blob/49b39fc9f92018b41b48bba2b9401dad953899fe/diesel/src/pg/connection/mod.rs#L283) implementation to mark any connection with open transactions as broken. That results in the pool discarding these connections and opening a new one instead. [In addition any connection returned during a panic is marked as broken](https://github.com/diesel-rs/diesel/blob/49b39fc9f92018b41b48bba2b9401dad953899fe/diesel/src/r2d2.rs#L198)

jtgeibel:
I've pushed a commit simplifying the background worker rollback logic and another commit with a few cleanups I found while reviewing the diff. There is just one failing test.

It seems there is a subtle difference in how transaction rollback works when the commit fails due to a `ReadOnlyTransaction` error. The `or_else()` fallback path in [this logic](https://github.com/jtgeibel/crates.io/blob/9869942d97bb3fe36f08e1317205c32291c2a74a/src/models/token.rs#L81-L86) now also fails, as shown by inserting a few `dbg!`s:

```
[src/models/token.rs:83] update(tokens).set(last_used_at.eq(now.nullable())).get_result(conn) = Err(
    DatabaseError(
        ReadOnlyTransaction,
        "cannot execute UPDATE in a read-only transaction",
    ),
)
[src/models/token.rs:87] tokens.first(conn) = Err(
    DatabaseError(
        Unknown,
        "current transaction is aborted, commands ignored until end of transaction block",
    ),
)
```

This appears to only be an issue with the test, as when I ran in readonly mode locally cargo gave me the following expected error when I attempted to yank a crate:

```
Caused by:
  the remote server responded with an error (status 503 Service Unavailable): Crates.io is currently in read-only mode for maintenance. Please try again later.
```

In production, this code runs outside of any transactions, so the transaction used to attempt the update is a real transaction, however when testing we are already in a transaction, so the update is attempted within a SAVEPOINT. In the test we also [create our own SAVEPOINT](https://github.com/jtgeibel/crates.io/blob/e673609e2ac2049951756b59daae62c33a254357/src/tests/read_only_mode.rs#L62-L63) which diesel isn't aware of.. 

I did some testing in `psql` and I was only able to replicate the `current transaction is aborted, commands ignored until end of transaction block` error if I attempted the SELECT within the (failing) savepoint. After a `ROLLBACK TO SAVEPOINT` the SELECT returned results as expected. So I'm thinking maybe diesel isn't rolling back the savepoint in this scenario, but I'm not sure exactly what is going on.

Ten0:
https://github.com/diesel-rs/diesel/issues/3470#issuecomment-1382182182

jtgeibel:
With the latest upstream diesel changes in the 2.0.3 release, our final failing test is now passing! Moving out of the draft state. :tada:

jtgeibel:
Last week I made another pass over this PR and realized that at some point I downgraded back to deisel_cli 1 so this was missing changes to the `schema.rs` file. That is now fixed and I've regenerated the `schema.patch` file so that it applies cleanly.

Included in the `schema.patch` file, we now make a few additional tweaks to the generated schema. `Array<Nullable<_>>` is changed to `Array<Nullable<_>>` in `api_tokens.crate_scopes`, `api_tokens.endpoint_scopes`, `dependencies.features`. I've confirmed there there are no existing null values in any of these arrays in the production database.

I've rebased this on master so that I can deploy this on staging for some additional testing. We should be able to merge this in a day or two.

jtgeibel:
Merged and deployed!

PR #4921, opened 2022-06-27
version::downloads: Simplify `add_custom_metadata()` usage

Turbo87:
I assume that this might have been necessary when we still had `&mut Request` in the parameters of the function, but since that is no longer the case we don't need the complicated logic here.


JohnTitor (approved):
Looks fine!

PR #4954, opened 2022-07-05
Follow button should be pointer

pickfire:
image

The button is currently not pointer.

### DIFF
--- a/app/components/follow-button.module.css
+++ b/app/components/follow-button.module.css
@@ -4,4 +4,5 @@
     height: 48px;
     width: 150px;
     justify-content: center;
+    cursor: pointer;
 }


JohnTitor (approved):
Makes sense to me.

Turbo87:
https://www.w3.org/TR/CSS21/ui.html#propdef-cursor says "pointer: The cursor is a pointer that indicates a link." but this is a button, not a link. 

I'm not strictly against making all buttons have the `pointer` cursor, but then we should probably do it for **all** buttons and not just the follow button :)

pickfire:
> I'm not strictly against making all buttons have the pointer cursor, but then we should probably do it for all buttons and not just the follow button :)

I don't see other buttons have this issue yet.

pickfire:
Closing this and create a new pull request.

PR #4996, opened 2022-07-20
Link SPDX license exceptions

dtolnay:
For example the <kbd>LLVM-exception</kbd> in the license of https://crates.io/crates/cranelift/0.85.1.


Turbo87 (approved):
😊👍

just FYI: I'm currently on vacation so deploying this might take a bit

PR #5022, opened 2022-07-30
Ensure the HTTP index is updated after a yank/unyank

jtgeibel:
The intent of this PR is to apply the fix in the final commit to enqueue a sync of the HTTP index whenever a yank/unyank is performed. This mirrors the way that a sync is enqueued by the publish background job after a new version is published. Once this is deployed, we will need to run a full index sync to correct any existing divergence between the official git index and the experimental HTTP index.

The other commits only affect the test harness and http-data files. A new `RECORD=force cargo test` mode allows for regenerating these data files. In this mode all responses are simulated as a status=200 success with an empty body and no response headers. Apparently we have no existing tests to cover failed S3 requests. The old recording mode where the request is actually proxied to the destination is still available via `RECORD=passthrough`, though we can probably remove this mode as I don't think any of the current team has ever used it.

I could decouple these changes for individual review, but it was very useful to build on top of the test harness improvements to update the data files and I don't think it adds much review overhead as only the final commit changes application behavior.

### DIFF
--- a/src/tests/http-data/krate_yanking_publish_after_yank_max_version
+++ b/src/tests/http-data/krate_yanking_publish_after_yank_max_version
@@ -61,6 +61,36 @@
   },
   {
     "request": {
+      "uri": "http://alexcrichton-test.s3.amazonaws.com/fy/k_/fyk_max",
+      "method": "PUT",
+      "headers": [
+        [
+          "accept",
+          "*/*"
+        ],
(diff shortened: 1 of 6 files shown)


Turbo87:
@jtgeibel I'll split this PR up to make the review a bit more manageable. hope that's okay with you :)

PR #5032, opened 2022-08-02
Replace is_alphanumeric + is_ascii with is_ascii_alphanumeric

dtolnay:
The use of `is_alphanumeric` for keyword validation is from 2014 (ee1c49be24d47c3d6a6a4d11acc7069976872892, aebaec8f3fe8514b350d48bef6461ed41380ffb2) which is way before `is_ascii_alphanumeric` became available (2018, Rust 1.24).


Turbo87 (approved):
👍

Issue #4891, opened 2022-06-14
Encoded URLs with a + sign fail


AngelloMaggio:
Some clients, proxies, and/or repository managers may encode URLs to the upstream (for instance JFrog Artifactory). Due to blocks on crates.io some of these requests fail if there is a + sign in the URL

e.g. If we go to this URL it works fine:

https://static.crates.io/crates/libgit2-sys/libgit2-sys-0.12.25+1.3.0.crate

But if we go to the encoded URL it gives us a 403 Access Denied 
https://static.crates.io/crates/libgit2-sys/libgit2-sys-0.12.25%2B1.3.0.crate

This has also exposed some more strange behavior.
If we encode the + sign as a space (%20 instead of %2B), it works.
https://static.crates.io/crates/libgit2-sys/libgit2-sys-0.12.25%201.3.0.crate

Or even if I put a space instead in the URL (though that may be my browser encoding for me)
"https://static.crates.io/crates/libgit2-sys/libgit2-sys-0.12.25 1.3.0.crate" 

Since URL encoding is fairly standard and other encoding works, I believe this is a bug

carols10cents:
`static.crates.io` goes to our s3 bucket. [The docs for s3 object names](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html) say:

> The following characters in a key name might require additional code handling and likely need to be URL encoded or referenced as HEX. 
> 
> ...
> - Plus ("+") 

So it looks like we should be URL encoding `+`. 

To fix this, we'll need to research whether we need to upload *both* URL encoded and unencoded for compatibility reasons or not...

MaggioAngello:
Any update on this one? It's failing rust builds for us and wanted to see if there is any light at the end of the tunnel. Thank you so much! 
@carols10cents @Turbo87

carols10cents:
Hi, yes, we talked about this in our most recent meeting but due to vacations and other projects, no one has the bandwidth to work on this immediately. If you have time to help, researching whether cargo can handle URLs using spaces returned from crates.io's `/download` redirects with or without encoding would be helpful. Thanks!

ambassadorfunk:
@carols10cents - Are there any further updates on this issue and the timing of when the code fix will be merged? Thanks!

carols10cents:
No, it needs to be reviewed and tested and no one has had time to do that yet.

Turbo87:
Progress Update:

[...]

PR #5034, opened 2022-08-03
DO NOT MERGE YET encode `+` as `%2B` in S3 paths

carols10cents:
🚨 🚨 🚨 🚨  We need to do migration of some files on S3 before this can be merged!!! 🚨 🚨 🚨 🚨 

Fixes https://github.com/rust-lang/crates.io/issues/4891, I think.

To test this:

- First, on the master branch with S3 configured, publish a crate with some build metadata (that is, containing a `+`)
- Observe the behavior reported in #4891, that the S3 URL is treating the `+` more like a space: requests to the S3 URL containing `%2B` (encoded `+`) will return 403 when it should return 200.
- Manually do the mass migration I think we need to script to change all `+` into `%2B`, by downloading and re-uploading the S3 file using `%2B` instead of `+` (or using the `aws` CLI). Don't use the S3 "copy" web interface, that seems to double encode.
- Observe requests to the S3 URL containing `%2B` now return 200.
- Switch to this PR's branch.
- Create a new local project that depends on the migrated crate containing build metadata on the registry now running this branch, to simulate us having deployed it. Verify Cargo can find and download that crate version.
- Remove the old S3 file containing `+` and confirm Cargo can still find and download that crate version (clear Cargo's download cache first to be sure), to simulate cleaning up after this branch is deployed
- Publish a new version containing build metadata and observe that when this branch does the uploading, the `+` gets encoded as `%2B` in the S3 path
- Verify Cargo can find and download that crate version too.

### DIFF
--- a/src/tests/http-data/krate_publish_new_krate_version_with_plus
+++ b/src/tests/http-data/krate_publish_new_krate_version_with_plus
@@ -1,0 +1,136 @@
+[
+  {
+    "request": {
+      "uri": "http://alexcrichton-test.s3.amazonaws.com/crates/foo_plus/foo_plus-1.0.0+build-metadata.crate",
+      "method": "PUT",
+      "headers": [
+        [
+          "host",
+          "alexcrichton-test.s3.amazonaws.com"
+        ],
+        [
+          "accept-encoding",
+          "gzip"
+        ],
+        [
+          "authorization",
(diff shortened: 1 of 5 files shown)


Turbo87:
this has been superseded by https://github.com/rust-lang/crates.io/pull/6649 and some other upcoming PRs :)

PR #2510, opened 2020-05-12
Permit '+' in feature name, as in "c++20"

dtolnay:
Up front: I am familiar with the several discussions linked by https://github.com/rust-lang/crates-io-cargo-teams/issues/41#issuecomment-494168728, and the conversation beginning at https://github.com/rust-lang/crates.io/pull/1331#issuecomment-379608097. This PR is not motivated by attempting to match whatever behavior Cargo currently has. Instead, it's a small thing I think we can decide now whether to allow. But it's necessary to say no corresponding Cargo change is required to accommodate this crates.io change.

This PR updates feature validation during publish to accept e.g. "c++20" and "dependency/c++20". We continue to not accept "c++20/feature" as the prefix before the slash would normally refer to a crate name of a dependency and a '+' would not be allowed in those.

I am interested in using such feature names in https://github.com/dtolnay/cxx.

In a Cargo.toml plusses appear as:

```toml
[features]
default = ["c++20"]
"c++20" = ["my-dependency/c++20"]
```

To give some slight further justification for why this particular ascii character above other possible characters we might allow: `+` is pretty common in OS package names in a way that no other currently disallowed character is. Some examples pulled arbitrarily from `apt-cache pkgnames | rg '\+'`:

- https://packages.debian.org/buster/dvd+rw-tools
- https://packages.debian.org/buster/elpa-ghub+
- https://packages.debian.org/buster/libarpack++2-dev
- https://packages.debian.org/buster/libdwarf++0
- https://packages.debian.org/buster/libelf++0
- https://packages.debian.org/buster/magics++
- https://packages.debian.org/buster/memtest86+
- https://packages.debian.org/buster/minisat+
- https://packages.debian.org/buster/paw++
- https://packages.debian.org/buster/swish++
- https://packages.debian.org/buster/vera++
- https://packages.debian.org/buster/voro++

The actual names of the projects contain `+`; various ones in the descriptions in the above links are styled as ARPACK++, Memtest86+, Magics++, Paw++, MinSat+, SWISH++, Vera++, Voro++. When binding to something like this behind a feature, using `+` in the feature name is the most intuitive.

PR #5046, opened 2022-08-03
Allow '+' in keywords

dtolnay:
Prior art: '+' in feature names &mdash; https://github.com/rust-lang/crates.io/pull/2510. See that PR for the rationales in favor of that.

Specifically I would be interested in using "c++" as a tag on some of my FFI crates. Alphanumericizing it to "cpp" or "cxx" is not so good because those are each the name of a specific crate for a specific approach to C++ interop, and a crate that's generally suitable for C++ interop wouldn't want its keywords to imply it only works in concert with one of those two other crates.

Cargo has supported '+' in keywords since Rust 1.1.

### DIFF
--- a/src/models/keyword.rs
+++ b/src/models/keyword.rs
@@ -56,7 +56,7 @@ impl Keyword {
             Some(c) => c,
         };
         first.is_ascii_alphanumeric()
-            && chars.all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-')
+            && chars.all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-' || c == '+')
     }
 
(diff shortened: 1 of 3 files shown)


Turbo87 (commented):
seems reasonable on first glance, though I'd like to get a second opinion on this first.

@jtgeibel @carols10cents any thoughts?

JohnTitor (commented):
I don't think I could attend the next meeting so going to put +1 here.

Turbo87 (approved):
I've manually tested the effects of this change on the UI and everything appears to work well :)

Turbo87:
@dtolnay sorry that it took so long. it should be live in production now :)

dtolnay:
Thank you! Seems to work: https://crates.io/keywords/c++

PR #5066, opened 2022-08-10
Generate index metadata files from the database

arlosi:
When a crate is modified, rather than editing the existing index file in git, this change re-generates the entire file from the database. This makes the jobs idempotent and prevents the index from getting out of sync with the database.

The `delete_version` and `delete_crate` admin tasks now also modify the index.

Test file changes are caused because the tests were inserting versions into the DB without adding them to the index.

### DIFF
--- a/src/admin/delete_crate.rs
+++ b/src/admin/delete_crate.rs
@@ -1,3 +1,4 @@
+use crate::background_jobs::Job;
 use crate::{admin::dialoguer, config, db, models::Crate, schema::crates};
 
 use diesel::prelude::*;
@@ -54,5 +55,12 @@ fn delete(opts: Opts, conn: &mut PgConnection) {
         panic!("aborting transaction");
     }
 
-    uploader.delete_index(&client, &krate.name).unwrap();
+    if dotenv::var("FEATURE_INDEX_SYNC").is_ok() {
+        Job::sync_to_git_index(&krate.name).enqueue(conn).unwrap();
+        Job::sync_to_sparse_index(&krate.name)
+            .enqueue(conn)
+            .unwrap();
+    } else {
+        uploader.delete_index(&client, &krate.name).unwrap();
+    }
 }
--- a/src/admin/delete_crate.rs
+++ b/src/admin/delete_crate.rs
@@ -56,10 +56,7 @@ fn delete(opts: Opts, conn: &mut PgConnection) {
     }
 
     if dotenv::var("FEATURE_INDEX_SYNC").is_ok() {
-        Job::sync_to_git_index(&krate.name).enqueue(conn).unwrap();
-        Job::sync_to_sparse_index(&krate.name)
-            .enqueue(conn)
-            .unwrap();
+        Job::enqueue_sync_to_index(&krate.name, conn).unwrap();
     } else {
         uploader.delete_index(&client, &krate.name).unwrap();
     }
--- a/src/worker/git.rs
+++ b/src/worker/git.rs
@@ -1,14 +1,15 @@
 use crate::background_jobs::{
     Environment, IndexAddCrateJob, IndexSyncToHttpJob, IndexUpdateYankedJob, Job, NormalizeIndexJob,
 };
+use crate::models;
 use crate::schema;
 use crate::swirl::PerformError;
 use anyhow::Context;
 use cargo_registry_index::{Crate, Repository};
 use chrono::Utc;
 use diesel::prelude::*;
-use std::fs::{self, OpenOptions};
-use std::io::{BufRead, BufReader, ErrorKind};
+use std::fs::{self, File, OpenOptions};
+use std::io::{BufRead, BufReader, ErrorKind, Write};
 use std::process::Command;
 
 #[instrument(skip_all, fields(krate.name = ?krate.name, krate.vers = ?krate.vers))]
@@ -72,6 +73,98 @@ pub fn perform_index_sync_to_http(
 
 pub fn update_crate_index(crate_name: String) -> Job {
     Job::IndexSyncToHttp(IndexSyncToHttpJob { crate_name })
+}
+
+/// Regenerates or removes an index file for a single crate
+#[instrument(skip_all, fields(krate.name = ?krate))]
+pub fn sync_to_git_index(
+    env: &Environment,
+    conn: &mut PgConnection,
+    krate: &str,
+) -> Result<(), PerformError> {
+    info!("Syncing to git index");
+
+    let new = get_index_data(krate, conn).context("Failed to get index data")?;
+
+    let repo = env.lock_index()?;
+    let dst = repo.index_file(krate);
+
+    // Read the previous crate contents
+    let old = match fs::read_to_string(&dst) {
+        Ok(content) => Some(content),
+        Err(error) if error.kind() == ErrorKind::NotFound => None,
+        Err(error) => return Err(error.into()),
+    };
+
+    match (old, new) {
+        (None, Some(new)) => {
+            fs::create_dir_all(dst.parent().unwrap())?;
+            let mut file = File::create(&dst)?;
+            file.write_all(new.as_bytes())?;
+            repo.commit_and_push(&format!("Creating crate `{}`", krate), &dst)?;
+        }
+        (Some(old), Some(new)) if old != new => {
+            let mut file = File::create(&dst)?;
+            file.write_all(new.as_bytes())?;
+            repo.commit_and_push(&format!("Updating crate `{}`", krate), &dst)?;
+        }
+        (Some(_old), None) => {
+            fs::remove_file(&dst)?;
+            repo.commit_and_push(&format!("Deleting crate `{}`", krate), &dst)?;
+        }
+        _ => debug!("Skipping sync because index is up-to-date"),
+    }
+
+    Ok(())
+}
+
+/// Regenerates or removes an index file for a single crate
+#[instrument(skip_all, fields(krate.name = ?krate))]
+pub fn sync_to_sparse_index(
+    env: &Environment,
+    conn: &mut PgConnection,
+    krate: &str,
+) -> Result<(), PerformError> {
+    info!("Syncing to sparse index");
+
+    let content = get_index_data(krate, conn).context("Failed to get index data")?;
+
+    env.uploader
+        .sync_index(env.http_client(), krate, content)
+        .context("Failed to sync index data")?;
+
+    if let Some(cloudfront) = env.cloudfront() {
+        let path = Repository::relative_index_file_for_url(krate);
+
+        info!(%path, "Invalidating index file on CloudFront");
+        cloudfront
+            .invalidate(env.http_client(), &path)
+            .context("Failed to invalidate CloudFront")?;
+    }
+
+    Ok(())
+}
+
+#[instrument(skip_all, fields(krate.name = ?name))]
+pub fn get_index_data(name: &str, conn: &mut PgConnection) -> anyhow::Result<Option<String>> {
+    debug!("Looking up crate by name");
+    let Some(krate): Option<models::Crate> = models::Crate::by_exact_name(name).first(conn).optional()? else {
+        return Ok(None);
+    };
+
+    debug!("Gathering remaining index data");
+    let crates = krate
+        .index_metadata(conn)
+        .context("Failed to gather index metadata")?;
+
+    debug!("Serializing index data");
+    let mut bytes = Vec::new();
+    cargo_registry_index::write_crates(&crates, &mut bytes)
+        .context("Failed to serialize index metadata")?;
+
+    let str = String::from_utf8(bytes).context("Failed to decode index metadata as utf8")?;
+
+    Ok(Some(str))
 }
 
 /// Yanks or unyanks a crate version. This requires finding the index
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -513,6 +513,7 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "base64 0.13.1",
+ "claims",
  "dotenv",
  "git2",
  "serde",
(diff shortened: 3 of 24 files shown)


Eh2406:
There are many reasons why this is a huge step forward, thank you. (This by itself fixes a recurring problem where people download the dump of the database and are surprised that they don't have access to data that only exists in the index.)
I cannot speak to the merge-ability of this PR as I am not a maintainer of crates.io.

I do have a concern about changes to the index files, if it's more appropriate in the follow up PR then move it when it is opened. The background context: The index is a shared database, it is read by all versions of Cargo. we have made a number of improvements to make Cargo resilient to changes it does not know how to understand. But the old versions are still in use and still brittle. Specifically old enough versions (pre 1.19) will decide an entire package is unavailable if any of the versions of that package cannot be parsed. This leads to a kind of breakage that so far we have found unacceptable. Old  Cargo build the project many years ago, generating a lock file, and this build succeeds. In the intervening years one of this projects dependencies publishes a new version that uses a new syntax in the index file. Old  Cargo using the same old lockfile now errors because it cannot parse a version it wasn't using. Se `features2` for the lengths we have gone to to avoid this breakage.

I don't think sorting various fields will cause a problem. Adding `"links": null`, will hopefully not cause a problem but should probably be checked, or we can use `#[serde(skip_serializing_if = "Option::is_none")]` to only write it where it has a value. `"kind": null` and `"features": [""]` are almost certainly OK, but also should be investigated. Just in case.

arlosi:
I re-generated the entire index using only the database and found the differences between the database and git index (excluding the normalization & sorting mentioned in the description). 22 crates had differences.

https://github.com/arlosi/crates.io-index/commit/c3c7a8c72a

* Several differences on `yanked` status
* A few duplicate entries removed
* Mixup between `_` and `-` on `glib` crate?
* A few changes in `deps`

arlosi:
@Turbo87 I broke this PR into multiple commits to make it easier to split out as needed.

src/admin/import_cksum.rs
```diff
@@ -0,0 +1,110 @@
...
+
+fn import_data(conn: &PgConnection, krate: cargo_registry_index::Crate) -> QueryResult<()> {
+    let (version_id, checksum): (i32, Option<String>) = versions::table
+        .inner_join(crates::table)
+        .filter(crates::name.eq(&krate.name))
+        .filter(versions::num.eq(&krate.vers))
+        .select((versions::id, versions::checksum))
+        .first(conn)?;
+
+    if checksum.is_none() {
```
arlosi, line 73: If the change to add `checksum` is deployed before the change to add `links`, then we'll need to do something different here. Otherwise it may miss importing `links` for crates that already have checksums.

This check assumes that `checksum.is_none()` means that we haven't done the import on the crate yet.

Turbo87: yeah, I noticed that. the checksum field has been in prod for a few days already. I guess we can probably clone the index and just go through the files one-by-one to do a full sync with the database. not sure how long this will take but I assume it would faster than the full S3 index sync.

Turbo87:
@arlosi btw once we've backfilled the index field in the database I guess it would also be good to expose the new fields in the APi too :)

jtgeibel (commented):
Looks good to me. Just 2 minor comments, one of which you've already split off into another PR (so feel free to tweak it here, or just ignore).

arlosi:
@Eh2406 Is there a package & recommended pre-1.19 Cargo that I should test this with?

I have `https://github.com/arlosi/crates.io-index` set up as a database-regenerated index from last-night's DB dump, so it should be straightforward to validate with old cargo and a source-replacement of crates.io

arlosi:
Crates versions published before Feb 2018 omit the `links` field, while crates published after have `links: null`, so Cargo clearly supports both forms.

When there are no links, should we normalize to `links: null`, or omit the field? cc @Eh2406 

The change is trivial: (adding `#[serde(skip_serializing_if = "Option::is_none")]`)

Eh2406:
I'm convinced that it won't cause a problem. Thanks for the clear proof.
If it does not matter, we may as well ship less data. Lets omit.

Turbo87:
`git-import` ran through without any errors this time. database should now be up-to-date unless we missed something in the import script :)

Eh2406:
> Is there a package & recommended pre-1.19 Cargo that I should test this with?

Once for each significant kind of change, find a version that has that change and build a test crate that has a `=x.y.z` dependency on it and see if an old cargo can generate a lock file. If, as we suspect, it can generate the lock file the same on the traditional index and your generated index then we are golden.

> I have https://github.com/arlosi/crates.io-index set up as a database-regenerated index from last-night's DB dump, so it should be straightforward to validate with old cargo and a source-replacement of crates.io

That is an excellent resource! One complication is I don't remember when source replacement was added to cargo. I think before then we had an environment variable that could override the crates.io url. But I don't remember what it was. Nor do I remember the early history of what changes were made to index parsing to guess which versions of cargo are best for testing.

arlosi:
I was able to use the `CARGO_REGISTRY_INDEX` environment variable to test with Cargo 1.17.0 (the oldest version that would run on Windows), and the lockfiles were generated the same for the examples I tested.

Eh2406:
Sounds good to me!

arlosi:
The first time a crate file is re-generated (due to yank/unyank/publish), there will likely be a large diff in the git index due to the normalization described in the description.

If there are tools/sites that work purely on `git diff` to determine which versions have changed, they may see that every version of a given crate has changed due to the normalization. This will be a one-time event for each crate. I don't know of any tools that work this way, but I thought I should mention it.

Eh2406:
I know of a number of tools that use https://github.com/Byron/crates-index-diff-rs we should make sure it handles this gracefully.

jonhoo:
I'd like to propose that `crates-index-diff` be updated _first_, and given a little while to propagate before this lands. Otherwise we risk someone (_cough cough_) being trapped in a position where they need a new version of `crates-index-diff` to import new versions from the changed index, but they can't use the new version of `crates-index-diff` because they'd need to import it first.

Byron:
Thanks for reeling me in! While I am working on https://github.com/Byron/crates-index-diff-rs/pull/17 I looked here to see if there is a test already to assert that the data itself is indeed unchanged after the normalization. So I was wondering if a test is possible that builds an in-memory state of all data in memory prior to the normalization, and asserts that it is equal to the one after normalization, like 

```Rust
assert_eq!(state, normalize(&state), "the meaning of state didn't change")`
assert_ne!(state.to_string(), normalize(&state).to_string(), "only its representation does")`
```

Is this possible, or am I missing something and it's not? I am asking because this is my intuition for making diffing robust in `crates-index-diff`. Thank you.

arlosi:
There are a few crates that do have semantic changes: https://github.com/rust-lang/crates.io/pull/5066#issuecomment-1215243839

Byron:
> There are a few crates that do have semantic changes: [#5066 (comment)](https://github.com/rust-lang/crates.io/pull/5066#issuecomment-1215243839)

Thanks for the heads-up! I extracted a [static snapshot with complete history](https://github.com/Byron/crates-index-diff-rs/blob/7e85688272bc02e6e9ba923bfc72e219ee228a7a/tests/fixtures/make-index-parts.sh#L12) of the `gi/` subdirectory of the https://github.com/arlosi/crates.io-index fork of the index and will see if some of these crates are there. Otherwise it appears correct to expect the amount of actual changes in there [to be zero](https://github.com/Byron/crates-index-diff-rs/blob/aeb6f45b9866e1c15862f336b6fe49bb3cf2dc2c/tests/index/changes_from_objects.rs#L62). And I hope to know soon, it's great fun.

Byron:
`crates-index-diff` in its [latest version](https://github.com/Byron/crates-index-diff-rs/releases/tag/v11.0.0) is now able to [handle normalizationi](https://github.com/Byron/crates-index-diff-rs/blob/51559b87a7c6b28b626e74b64445335d9a9f813a/tests/index/changes_between_commits.rs#L71-L76) as shown [in this fork of crates.io](https://github.com/arlosi/crates.io-index) correctly.

A PR to integrate it into docs.rs [has been created](https://github.com/rust-lang/docs.rs/pull/1807) as well.

arlosi:
Questions for the crates.io team:
* Do we need a larger announcement of this change?
* How long do we need to wait before deploying?

jtgeibel:
> Do we need a larger announcement of this change?

I think a post on internals would make sense, or possibly a blog post.

> How long do we need to wait before deploying?

This probably depends somewhat on the feedback that we get from the community, as there may be other tools that make assumptions about how the index repo evolves over time. Probably a few weeks after announcing, unless people identify use cases that will take more time to adapt.

A few additional points that the team discussed:

* We should review https://github.com/arlosi/crates.io-index/commit/70b776361ad9890f413473759db7fcba9f69ef54 and land those changes ahead of time.
* Do we want to roll this out progressively as crates are published or do a bulk update of the registry? I'm personally (and I believe the team in general) leaning towards doing a bulk update so that there is a pair of commits with the full diff but no semantic difference. This would probably be easier to independently verify, as otherwise the migration period is unbound. It would probably make sense to do the migration followed immediately by a squash of the index. However it's possible this would be more confusing to some consumers.
* We would like to deploy this to staging. It would be nice if there was a environment flag to turn this on and off, but that's probably too annoying to incorporate now. We can probably do test merges into a `staging` branch to temporarily manage the 2 deployments.
* We're really excited by the progress already made on this and it is already a huge improvement having this information available in the database! Thank you, and sorry for the delay in getting back to you.

Eh2406:
We are already going to be removing `"links": null,`! It makes sense to remove other fields that have been added since Cargo 1.19 when they're equal to their default value.

arlosi:
`"target:"` is [not documented as an optional field](https://doc.rust-lang.org/cargo/reference/registries.html#index-format), which means we'd need to update the registry specification before removing it. Cargo works with it removed, so maybe it's worth updating the spec?

I don't see any fields besides `"links"` that we can currently remove according to the specification.

> [let's do this as a] bulk update so that there is a pair of commits with the full diff but no semantic difference

If we want to do a single bulk update, do we need to put the site in read-only mode while the index is being re-generated? How much read-only time is acceptable? We'll also need an admin command to re-generate the entire index. I have the beginnings of such a command that I used while testing.

Turbo87:
> If we want to do a single bulk update, do we need to put the site in read-only mode while the index is being re-generated?

I don't see a way to avoid that. Since we already do similar things for database updates, a one time maintenance window for something like that would seem reasonable.

> How much read-only time is acceptable?

do you know how much time it would roughly take?

arlosi:
> do you know how much time it would roughly take?

On my machine with a local postgresql instance it took about 15 minutes.

Turbo87:
if we only want to reorder existing records we don't need postgres though, right?

arlosi:
> if we only want to reorder existing records we don't need postgres though, right?

Good point. Yes we can do that.

Turbo87:
As mentioned on Zulip, I've taken a look at the proposed changes and implemented some improvements on top:

- the `index_metadata()` method now returns an unserialized `Vec<Crate>`, which makes it easier to test via `insta`
- the `cargo-registry-index` crate contains the necessary code to serialize a `Vec<Crate>` to an index file
- the `index_metadata()` method sorts by `created_at` (for production) before falling back to semver sorting (for tests, because all versions in a test db transaction will unfortunately end up with the same `created_at`)
- the `index_metadata()` method will fetch all dependencies of all relevant versions in a single database query instead of running one query per version
- the `sync` background job has been split up into one job per index (git and sparse). without this, we could end up in a situation where the sparse index update fails, causing the job to be retried, but the git index check will see that both new and old file content already match and skip the sparse index update too.
- I've added a `FEATURE_INDEX_SYNC` environment variable to turn the new behavior on and off, which should allow us to merge this but only enable it on staging for now
- the above required adding new background jobs instead of changing the existing ones

@arlosi I hope you're okay with these changes. Would be great to get a review from you before I merge this :)

src/admin/delete_crate.rs
```diff
@@ -54,5 +55,12 @@ fn delete(opts: Opts, conn: &mut PgConnection) {
         panic!("aborting transaction");
     }
 
-    uploader.delete_index(&client, &krate.name).unwrap();
+    if dotenv::var("FEATURE_INDEX_SYNC").is_ok() {
+        Job::sync_to_git_index(&krate.name).enqueue(conn).unwrap();
```
arlosi, line 59: Is it worth having a function that enqueues both the sync jobs?

arlosi (commented):
Your changes look good to me. We'll want to do a follow up PR to clean up the old code path once it's been running successfully on prod for a while.

When we enable this on prod:
* This will normalize any crates that aren't already normalized if they're changed (new version, yank/unyank).
  * Currently this is around 9000 crates
  * I think this should be fine. It will spread the normalization out over time and avoid any large normalization commit. The number of denormalized crates will shrink over time and we can do a bulk normalization to get the remainder in a few months.
* It will resolve the dependency index/db differences mentioned in the Zulip thread if they're changed. 
  * We should merge my commit that resolves these before enabling this on prod

src/worker/git.rs
```diff
@@ -74,6 +75,98 @@ pub fn update_crate_index(crate_name: String) -> Job {
...
+        (None, Some(new)) => {
+            fs::create_dir_all(dst.parent().unwrap())?;
+            let mut file = File::create(&dst)?;
+            file.write_all(new.as_bytes())?;
+            repo.commit_and_push(&format!("Creating crate `{}`", krate), &dst)?;
+        }
+        (Some(old), Some(new)) if old != new => {
+            let mut file = File::create(&dst)?;
+            file.write_all(new.as_bytes())?;
+            repo.commit_and_push(&format!("Updating crate `{}`", krate), &dst)?;
```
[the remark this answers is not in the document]
Turbo87: > This will produce less specific git commit messages than we currently use.

Yep, but I figured since we can't be sure if we only push a single change or not we may as well use a more generic message.

src/admin/delete_crate.rs
Turbo87: I guess that would make sense :D

Turbo87: done :)

PR #5077, opened 2022-08-14
database: Add `checksum` field

Turbo87:
This can eventually be used to generate the index from the database.

This PR was extracted from https://github.com/rust-lang/crates.io/pull/5066 to make it individually reviewable and mergeable :)


arlosi (approved):
The other two database changes (`links` and `explicit_name`) are also required to be able to generate the index from the database, but it's fine if we want to do them as separate PRs & DB migrations.

Turbo87:
yeah, I'd prefer to ship these changes in chunks to reduce the surface area of what might be able to break unexpectedly :)

PR #5089, opened 2022-08-17
index: Use/Allow alphabetic ordering of features and dependencies

Turbo87:
This PR was extracted from https://github.com/rust-lang/crates.io/pull/5066 to make it independently reviewable and mergeable.

The main purpose of these changes is to match the order that the entries in the index are currently serialized in. This will eventually allow us to recreate the index from a single source of truth (the database) with a minimal diff.

### DIFF
--- a/cargo-registry-index/lib.rs
+++ b/cargo-registry-index/lib.rs
@@ -139,7 +139,7 @@ pub struct Crate {
     pub v: Option<u32>,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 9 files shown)


arlosi:
There are a few more places that should probably be updated to use `BTreeMap` for consistency.

See this commit for suggested changes: https://github.com/rust-lang/crates.io/pull/5066/commits/a7da4a14ee2129f8e8a5ef7ef805950dd732d10e

arlosi (approved):

PR #5111, opened 2022-08-22
database: Add `links` column to `versions` table

arlosi:
Another PR split out from #5066 -- adding the `links` column. This is the final database schema change required to have all data from git in the database.

r? @Turbo87


Turbo87 (approved):
awesome, thanks :)

PR #5118, opened 2022-08-24
ESLint: Add missing `@babel/plugin-proposal-decorators` dependency

Turbo87:
This is referenced by the ESLint config and can sometimes cause issues if not explicitly specified. This is probably a side effect of us switching to pnpm, which is a bit more strict in terms of peer dependencies.


JohnTitor (approved):
Sounds reasonable.

PR #5121, opened 2022-08-25
Address some clippy warnings

jtgeibel:
The 2nd commit address a warning from the `beta` channel, and the 1st commit is from `nightly`. Both cleanups are reasonable.


arlosi (approved):

Turbo87 (approved):
looks like the failing Percy CI job is an issue on their side. seems highly unlikely that changing the Rust code would cause any visual changes on the frontend test suite 😅

PR #5127, opened 2022-08-25
Autofill search box when browsing categories

seam345:
This will fill the search box with `category:<category> ` on all
category pages helping the user discover how to filter by category.

Also means when the user appends to the search field it will be searching 
the current category that is selected, instead of previously all the crates.

### DIFF
--- a/app/routes/category.js
+++ b/app/routes/category.js
@@ -5,9 +5,11 @@ import { inject as service } from '@ember/service';
 export default class CategoryRoute extends Route {
   @service router;
   @service store;
+  @service header;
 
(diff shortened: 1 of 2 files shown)


seam345:
Due to running this in docker I was unable to run the tests 🤞 they all pass if not I'll mark as draft and get to fixing :)

seam345:
Some snapshots of what it looks like 
image
image

Turbo87 (approved):
I've rebased, added a small test for the new behavior and successfully tried it out locally. thanks for your contribution :)

PR #5133, opened 2022-08-28
tests: Use `insta` to assert on full version API responses

Turbo87:
see https://insta.rs

This allows us to more easily assert on the exact structure of our API responses and give us a straight-forward way to keep the assertions up-to-date. 

Currently, if we add a new field to our API responses, the tests will still pass, because the tests are using the same structs as the production code. Even worse, if we remove a field the same is true, even though this should be considered a breaking change. With snapshot tests it will be easy to detect these sort of things because the test files will need to be updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -251,6 +251,7 @@ dependencies = [
  "hyper-tls",
  "indexmap",
  "indicatif",
+ "insta",
  "ipnetwork",
  "lettre",
  "minijinja",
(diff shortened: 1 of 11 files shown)


Turbo87:
@JohnTitor can I interpret your 👍 as an approval? 😅

JohnTitor:
Nah, I just read the PR description and would like to show "I like this idea!", actually haven't reviewed it yet. Sorry for the confusion!

JohnTitor (approved):
The change looks great to me! I guess it's helpful to document how to update snapshot results (IIUC `cargo insta review` does it, right?) but it'd be a bonus point.

Turbo87:
> I guess it's helpful to document how to update snapshot results (IIUC `cargo insta review` does it, right?) but it'd be a bonus point.

yeah, that's correct. It seemed to me that https://insta.rs/docs/quickstart/ already explained it quite well, so I didn't want to copy-paste it into this repo again :)

PR #5135, opened 2022-08-28
versions: Adjust `checksum` column to be `NOT NULL`

Turbo87:
Now that we have run the import script, all of our `versions` in the database have a non-null `checksum` field. This PR adjusts the column to be a `NOT NULL` column to avoid having to unnecessarily deal with `Option<_>` for this field.


arlosi (approved):
LGTM

PR #5175, opened 2022-09-01
CrateHeader: Add `v` prefix to version number

Turbo87:
The tiniest possible change, but this matches how we display them in the search results and makes the page a little more consistent.

Bildschirmfoto 2022-09-01 um 20 08 35


JohnTitor (approved):
Makes sense to me. Some places still don't have `v`, but almost have some context (e.g. under the version column) so I guess it's okay. A candidate would be the download graph but it's not a strong opinion. LGTM with the test fixed.

PR #5177, opened 2022-09-01
Social media tags

elpiel:
Related to #5176 I've decided to open a WIP PR for adding some initial tags and to drive the discussion on how to implement this feature.

### DIFF
--- a/app/templates/head.hbs
+++ b/app/templates/head.hbs
@@ -1,1 +1,15 @@
 <meta name="description" content={{this.model.description}}>
+
+<meta property="og:title" content="{{this.model.name}}">
+<meta property="og:type" content="website" />
+<meta property="og:image" content="https://crates.io/assets/Cargo-Logo-Small.png">
+<meta property="og:url" content="https://crates.io/crates/{{this.model.name}}">
+<meta name="twitter:card" content="summary_large_image">
+
+<!--  Non-Essential, But Recommended -->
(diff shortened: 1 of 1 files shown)


Turbo87:
the main issue with this is that we don't use server-side rendering of the HTML. for Google indexing this is fine, but pages like Twitter don't execute the JS rendering code, so just adding the tags is unfortunately not enough for them to actually be useful in the current setup. there have been some attempts to use server-side rendering with the JS code, but last time it was tried I think it caused excessive load on our servers so it was disabled again.

elpiel:
If I understand correctly, this, with the current setup, is challenging to do?!
Should we leave the discussion to see what could be other solutions for this? E.g. Some sort of proxy for when twitter requests the webpage

Turbo87:
> If I understand correctly, this, with the current setup, is challenging to do?!

unfortunately, yes. unless we want to pay a lot more for more powerful servers it will most likely not happen anytime soon.

> E.g. Some sort of proxy for when twitter requests the webpage

yeah, some sort of redirect for well known user-agents might be an option. I think we're serving our frontend behind AWS CloudFront though, so that might have to be taken into account in any solution proposal.

> Should we leave the discussion to see what could be other solutions for this?

Sure, yeah, the discussion can stay open as a long-term feature request, but I would propose to close the PR, since this is unlikely to be solve in the near-term future.

PR #5185, opened 2022-09-02
Add pagination to the "Pending Owner Invites" page

EstebanBorai:
Provides the missing pagination for the Pending Owner Invites page
on `/me/pending-invites`.

Refer: https://github.com/rust-lang/crates.io/issues/4083

---

## Demo

https://user-images.githubusercontent.com/34756077/188549186-9b5224de-4368-46de-bb85-3834d1142079.mov


mirage/serializers/crate-owner-invitation.js
```diff
@@ -24,7 +24,9 @@ export default BaseSerializer.extend({
     delete hash.id;
     delete hash.token;
 
-    hash.crate_id = Number(hash.crate_id);
+    // TODO: Check this further, `crate_id` are strings in the current fixtures
+    // when attempting to parse into number we will get `NaN`s here.
+    // hash.crate_id = Number(hash.crate_id);
```
EstebanBorai, line 29: This is causing some trouble when retrieving crates by its ID, given that crates IDs are actually strings with the name of such crate.

EstebanBorai:
I'm taking this back to draft as I still have to check for backend pagination support

src/controllers/crate_owner_invitation.rs
```diff
@@ -139,7 +139,16 @@ fn prepare_list(
...
+        Page::Numeric(_) => query
+            .pages_pagination(pagination)
+            .filter(
+                crate_owner_invitations::crate_id.gt(seek_key.0).or(
+                    crate_owner_invitations::crate_id
+                        .eq(seek_key.0)
+                        .and(crate_owner_invitations::invited_user_id.gt(seek_key.1)),
+                ),
+            )
+            .load(&*conn)?,
```
Turbo87, line 151: I think page-based pagination has been explicitly not implemented here for performance reasons. We would like to move all of our APIs to seek-based pagination in the future for these reasons.

mirage/fixtures/crate-owner-invitations.js
```diff
@@ -0,0 +1,102 @@
...
+  },
+  {
+    inviteeId: 1,
+    inviterId: 2,
+    crateId: 'rusted_cypher',
+  },
+  {
+    inviteeId: 1,
+    inviterId: 2,
+    crateId: 'nanomsg',
```
Turbo87, line 25: could you try to use mirage factories instead? we're trying to move away from hard-coded fixtures :)

src/controllers/crate_owner_invitation.rs
EstebanBorai: Got it! Do we have support for seek-based pagination already in some place?

EstebanBorai: @Turbo87 I can see Seek pagination is already supported for owner invites for the backend here: https://github.com/rust-lang/crates.io/blob/04d4350b3c2b561a39e14abdec7c2d1e727eaf2c/src/controllers/crate_owner_invitation.rs#L130

But I still have to check if it works for the pagination implementation

EstebanBorai:
Hi @Turbo87!

Im working on providing support for the seek pagination format on the pagination component.
And I notice we have a fragment already wrote for the `next_page` which comes handy for the `next` button.
But when it comes for page numbers, I'm thinking of possible approaches.

As I see in this fragment
https://github.com/rust-lang/crates.io/blob/ef0502d8f194b300d4a76880229c4a160f4ad1b8/src/controllers/crate_owner_invitation.rs#L150

As we must know the ids of the last entry in the current page in order to fetch the next page (or a specific one), I was thinking about retrieving next sets as part of the meta included in the response. For instance if you fetch page 1 and establishing that we will always have a range of 10 pages per view. I would retrieve a `pages` property with the corresponding ID tuples for the next 10 pages.

What do you think? Perhaps we already have a definition for this?

Thanks in advance!

EstebanBorai:
Moved into just next page support due to the behavior of the seek pagination.
The following demo shows how seek pagination is used to load the next page. In a next iteration it would be great (IMHO) to be able to load next invites in the same page, similar to an _infinity scroll_ behavior.

### Demo

https://user-images.githubusercontent.com/34756077/190247325-2f960938-a244-40fb-9d34-f7e5c4eb4bb1.mov

EstebanBorai:
Also noticed the button on gray was a bit "misleading" based on the color of remaining anchors around the site.
So I fixed it using the same green as we always do.

Screen Shot 2022-09-14 at 4 40 25 PM

src/controllers/crate_owner_invitation.rs
```diff
@@ -147,15 +150,12 @@ fn prepare_list(
...
-                    last.invited_user_id,
-                ))?,
-            );
-            Some(req.query_with_params(params))
+            let seek_key = crate::controllers::helpers::pagination::encode_seek((
+                last.crate_id,
+                last.invited_user_id,
+            ))?;
+
+            Some(seek_key)
```
EstebanBorai, line 158: I changed this a bit to receive the `next_page` value inside the `meta` object which is the `seek` value encoded directly instead of using the prebuilt query parameter. This way I just take the value of the seek encoded cursor directly.

Turbo87: unfortunately this is a breaking change to our existing API responses and we need to be quite careful with those. ideally we will implement this without any changes to the backend code.

EstebanBorai: @Turbo87 of course! Let me take this back to the previous state and make sure the client side works as expected.

app/utils/seek-pagination.js
```diff
@@ -0,0 +1,12 @@
...
+export function pagination() {
+  return macro(function () {
+    const { nextPage } = this;
+    const nextPageParams = new URLSearchParams(nextPage);
+
+    return {
+      nextPage: nextPageParams.get('seek'),
+    };
+  });
+}
```
EstebanBorai, line 12: Parses the `next_page` value from the response `meta` field which comes a URL Query Params, and retrieves the value from `seek` (if available) otherwise returns `undefined`.

src/controllers/crate_owner_invitation.rs
```diff
@@ -47,6 +49,7 @@ pub fn list(req: &mut dyn RequestExt) -> EndpointResult {
     Ok(req.json(&json!({
         "crate_owner_invitations": crate_owner_invitations,
         "users": users,
+        "meta": meta,
```
EstebanBorai, line 52: Adds `meta` to the response similar to how its done for `crates` endpoint, this way we can retrieve the seek hash for the next page.

Turbo87: sorry that I just noticed this now, but https://github.com/rust-lang/crates.io/pull/3763 introduced a new endpoint for the invitations (`GET /api/private/crate_owner_invitations`) that already supports seek-based pagination. I guess it would be best if we first switched to this new endpoint in the frontend in a dedicated MR and then implemented pagination on top of it. let me know if you want to tackle this :)

EstebanBorai: Sounds good! Lets do that!
So we close this PR?

Turbo87: whatever you prefer. feel free to keep it open and then rebase it once the other stuff is done :)

EstebanBorai:
Closing and moving into another pr with the new seek based pagination...

PR #5200, opened 2022-09-07
Add CloudFront invalidation for index files

### LINKED ISSUE #4913: Send invalidations for index.crates.io when updating crates
The HTTP index (index.crates.io) currently has a TTL of 10 minutes and does not send invalidations when publishing new crates. Users may need to wait up to 10 minutes after publishing a crate for it to be visible in the HTTP index.

If we sent invalidations to CloudFront when a crate was updated, we could increase the TTL (for better caching), and users would see updates sooner.

arlosi:
Send invalidation request to CloudFront when a crate is modified.

Does nothing if `CLOUDFRONT_DISTRIBUTION` isn't set.

Fixes #4913

### DIFF
--- a/src/worker/cloudfront.rs
+++ b/src/worker/cloudfront.rs
@@ -1,0 +1,75 @@
+use std::time::SystemTime;
+
+use aws_sigv4::{
+    http_request::{self, SignableRequest, SigningSettings},
+    SigningParams,
+};
+use reqwest::blocking::Client;
+
+#[derive(Clone)]
+pub struct CloudFront {
+    distribution_id: String,
+    access_key: String,
+    secret_key: String,
+}
+
+impl CloudFront {
+    pub fn from_environment() -> Option<Self> {
+        let distribution_id = dotenv::var("CLOUDFRONT_DISTRIBUTION").ok()?;
+        let access_key =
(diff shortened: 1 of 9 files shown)


src/worker/cloudfront.rs
```diff
@@ -0,0 +1,74 @@
...
+        Some(Self {
+            distribution_id,
+            access_key,
+            secret_key,
+        })
+    }
+
+    /// Invalidate a file on CloudFront
+    ///
+    /// `path` is the path to the file to invalidate, such as `config.json`, or `re/ge/regex`
```
jtgeibel, line 32: We may want to add a `let path = path.strip_prefix('/').unwrap_or(path);` to handle the case where an initial `/` is included.

jtgeibel (approved):
Looks great to me, thanks! One small suggestion.

arlosi:
@jtgeibel I've updated the PR to include your suggestion. Can this be merged?

jtgeibel:
Sorry for the delay. I've just merged and will deploy to staging. I've pinged the [infra](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/crates.2Eio.20cloudfront.20environment.20variables/near/302372874) team about getting the environment variables set.

PR #5216, opened 2022-09-13
Add FAQ on how to leave crate open for adoption

EstebanBorai:
Provides a section in the policies page to communicate to users how they could leave a crate
open for transferring. The idea was first shared by @carols10cents.

### DIFF
--- a/app/templates/policies.hbs
+++ b/app/templates/policies.hbs
@@ -45,14 +45,14 @@
 <h3 id='delete-crate'><a href='#delete-crate'>How can I delete a crate I own from the registry?</a></h3>
 
 <p>
-  You can't delete crates from the registry but you could leave it open for
-  ownership transfering.
+  You can't delete crates from the registry, but you can leave it open for
+  transferring ownership to others.
 </p>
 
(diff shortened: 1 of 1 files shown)


app/templates/policies.hbs
```diff
@@ -42,6 +42,24 @@
...
 
+<h3 id='delete-crate'><a href='#delete-crate'>How can I delete a crate I own from the registry?</a></h3>
+
+<p>
+  You can't delete crates from the registry but you could leave it open for
+  ownership transfering.
+</p>
+
+<p>
+  To do this, you must yank all of the versions in the crate
```
Turbo87, line 53: I'm not sure if yanking is strictly necessary. IMHO there is nothing wrong with keeping the old versions around in case other apps or libraries depend on them.

EstebanBorai: Makes sense to me! I'm cc'ing @carols10cents because we discussed this on chat before.

Turbo87: I took the liberty of updating the PR :)

PR #5267, opened 2022-09-28
Fix typo in footer

HigherOrderLogic:


Turbo87:
the PR description is a bit sparse on why you think this is a typo. could you explain why you would like to change this? 😅

HigherOrderLogic:
> the PR description is a bit sparse on why you think this is a typo. could you explain why you would like to change this? 😅

Since other field all have first letter upper-cased, except for `rust-lang.com`, which is an URL. And beside, `the crates.io team` is a name so I think this is a typo.

Turbo87 (approved):
sounds like a reasonable explanation, thanks :)

PR #5294, opened 2022-10-07
Introduce daily limit of published versions per crate

### LINKED ISSUE #5291: Limit number of new versions per one crate to 10 per day
Between UTC midnights, only allow up to 10 publishes of new versions of any one particular crate.

nappa85:
Introduces a new env var MAX_NEW_VERSIONS_DAILY ~~with default value 10~~

Resolves #5291

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -25,6 +25,7 @@ pub struct Server {
     pub max_upload_size: u64,
     pub max_unpack_size: u64,
     pub publish_rate_limit: PublishRateLimit,
+    pub new_version_rate_limit: Option<u32>,
     pub blocked_traffic: Vec<(String, Vec<String>)>,
     pub max_allowed_page_offset: u32,
     pub page_offset_ua_blocklist: Vec<String>,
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -145,6 +148,8 @@ impl Default for Server {
             version_id_cache_ttl: Duration::from_secs(
                 env_optional("VERSION_ID_CACHE_TTL").unwrap_or(DEFAULT_VERSION_ID_CACHE_TTL),
             ),
+            max_new_versions_daily: env_optional("MAX_NEW_VERSIONS_DAILY")
+                .unwrap_or(DEFAULT_MAX_NEW_VERSIONS_DAILY),
```
Turbo87, line 152: I think I would prefer it if we kept `max_new_versions_daily` an `Option` without the default value. That way we can easily enable/disable this when we deploy the code.

```suggestion
            max_new_versions_daily: env_optional("MAX_NEW_VERSIONS_DAILY"),
```

src/models/version.rs
```diff
@@ -172,6 +177,10 @@ impl NewVersion {
                 )));
             }
 
+            if let Some(limit) = daily_limit {
+                self.rate_limit(conn, limit)?;
+            }
```
Turbo87, line 182: I assume you've put the rate limiting here to ensure that it's running in the same database transaction? Since we have the `publish_rate_limit` already in the publish controller, I think that might be a more straight forward place for it. That way we also don't need to adjust all callers of the `save()` method to add this additional parameter.

src/config.rs
nappa85: Yes, it's not a problem

src/models/version.rs
nappa85: I was indeed trying to separate the concepts, since I saw the existing rate limiter uses a bucket system, and here I'm simply querying on a time range.
I can move the method inside the PublishRateLimit struct without problems

Turbo87: > I can move the method inside the PublishRateLimit struct without problems

not necessarily inside the struct, but I think it makes more sense to do the time range query in the controller, instead of in the model method :)

PR #5301, opened 2022-10-08
Allow crate version deletion in the first 24 hours from creation

### LINKED ISSUE #5293: Allow deletion of crates within 24 hours of publishing
NPM has this policy to better handle accidental publishing.

nappa85:
Adds route DELETE /api/v1/crates/:crate_id

Resolves #5293

### DIFF
--- a/src/controllers/version.rs
+++ b/src/controllers/version.rs
@@ -1,3 +1,4 @@
+pub mod delete;
 pub mod deprecated;
 pub mod downloads;
 pub mod metadata;
--- a/src/controllers/version/delete.rs
+++ b/src/controllers/version/delete.rs
@@ -1,0 +1,68 @@
+//! Functionality related to deleting a crate.
+
+use chrono::{Duration, Utc};
+use diesel::dsl::count_star;
+
+use crate::controllers::cargo_prelude::*;
+
+use crate::schema::*;
+use crate::util::errors::internal;
+use crate::views::{EncodableCrate, GoodCrate, PublishWarnings};
+
+use super::version_and_crate;
+
+/// Handles the `DELETE /crates/:crate_id/:version` route.
+///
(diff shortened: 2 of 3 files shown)


Turbo87:
@nappa85 just for awareness: https://github.com/rust-lang/crates.io/issues/5293#issuecomment-1272406916

tl;dr I'm not sure whether we should allow crate deletion, or if it makes more sense for us to allow version deletion, and then if there is no version left the crate automatically gets deleted too.

nappa85:
> @nappa85 just for awareness: [#5293 (comment)](https://github.com/rust-lang/crates.io/issues/5293#issuecomment-1272406916)
> 
> tl;dr I'm not sure whether we should allow crate deletion, or if it makes more sense for us to allow version deletion, and then if there is no version left the crate automatically gets deleted too.

Updated, let me know if there is anything else

Turbo87:
I think there is some work left on correctly syncing both the git and the http index after the deletion. But before we continue, I talked to a few people about this and since it is a quite significant change we might actually require to go through the RFC process for it since other people are currently relying on the fact that crates don't just disappear like that again.

Turbo87:
and just to be clear, I'm very thankful for the quick implementation! I just want to make sure we're not we're not surprising too many people with this change :)

nappa85:
> and just to be clear, I'm very thankful for the quick implementation! I just want to make sure we're not we're not surprising too many people with this change :)

Understandable, don't worry

nappa85:
Something nasty happened during the rebase, probably better restart from scratch...

Turbo87:
FWIW there is now an RFC for this: https://github.com/rust-lang/rfcs/pull/3660

since the implementation will need a couple of changes (and rebases) I'll close this PR and will use it as inspiration when implementing the RFC spec :)

PR #5302, opened 2022-10-08
WIP: add paginated versions API

### LINKED ISSUE #5292: Add paginated versions API

nappa85:
Made some assumptions here, feel free to comment.
* didn't know if the issue was about creating a new API or editing the current one
* there seems to be multiple APIs serving versions (e.g. also /crates/:crate_id returns these infos)
* data is sorted by semver in code, this obviously isn't compatible with pagination, so I've added query sort by created_at to mitigate the problem, but I'm quite not sure about it

Still haven't checked if it need some edits frontend-side

Resolves #5292

### DIFF
--- a/src/controllers/krate/metadata.rs
+++ b/src/controllers/krate/metadata.rs
@@ -8,8 +8,8 @@ use std::cmp::Reverse;
 use std::str::FromStr;
 
 use crate::controllers::frontend_prelude::*;
+use crate::controllers::helpers::pagination::{Paginated, PaginationOptions};
 use crate::controllers::helpers::Paginate;
-use crate::controllers::helpers::pagination::{PaginationOptions, Paginated};
 
 use crate::models::{
     Category, Crate, CrateCategory, CrateKeyword, CrateVersions, Keyword, RecentCrateDownloads,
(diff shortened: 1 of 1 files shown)


Turbo87:
There is also https://github.com/theory/pg-semver, apparently, but I don't know if that works well and is supported by Heroku Postgres. Might be worth trying out though.

nappa85:
> There is also https://github.com/theory/pg-semver, apparently, but I don't know if that works well and is supported by Heroku Postgres. Might be worth trying out though.

As far as I know (source: https://help.heroku.com/7KM0BSLY/how-do-i-add-a-postgres-extension-to-my-database) it's not possible to install extensions on Heroku's Postgres, you can only use the preinstalled ones

src/controllers/helpers/pagination.rs
```diff
@@ -73,7 +73,7 @@ impl PaginationOptionsBuilder {
         self
     }
 
-    pub(crate) fn gather(self, req: &mut dyn RequestExt) -> AppResult<PaginationOptions> {
```
nappa85, line 76: @Turbo87 I don't know if I dared too much here, I haven't seen a real motivation for that mutable reference, and in facts the rest of the code is untouched

Turbo87: this seems like a useful change, but it would be best if such refactorings were kept out of this PR for a functional change. (see chapter 6 of https://mtlynch.io/code-review-love/ for a much better explanation :D)

nappa85:
Something nasty happened during the rebase, probably better restart from scratch...

Turbo87:
closing in favor of #8037

PR #5314, opened 2022-10-11
Crate with all versions yanked display reduced info

### LINKED ISSUE #1058: Crates who have yanked all their versions should only return name, owners, and that it's yanked

mario-areias:
This PR reduces the information returned by the API when crates have all versions yanked. Specifically `badges`, `documentation`, `homepage` and `repository` are set to `None` and `max_version` is set to `"0.0.0"`. These changes are applied on:
 - `GET /crates/:crate_id`  (crate page)
 - `GET /crates` (search).

In this PR the definition for a crate with all versions yanked is when `crate.versions()` returns no results. The [versions](https://github.com/rust-lang/crates.io/blob/f84eda9819d2ba2699268287402d1d08fa6f0732/src/models/krate.rs#L512) method loads all non-yanked versions. 

This definition simplifies the code a bit, specifically in `src/controllers/krate/search.rs` since the versions for the queried crates are already loaded in memory, so there is no need to run yet one more query. Semantically though, this definition is not entirely correct. `crate.versions()` could also mean that all versions have been deleted.

I don't think the difference matters much for this PR, but please do let me know if this is not a valid definition for all versions yanked crates and I will update the PR. 

Fix #1058

Search page changes:

image
Crate with 2 yanked versions and 1 non-yanked.

image
Crate with all 3 versions yanked.

--------------------------
Crate page changes:
image
Non-yanked crate.

image
Yanked crate.

### DIFF
--- a/src/controllers/krate/metadata.rs
+++ b/src/controllers/krate/metadata.rs
@@ -40,6 +40,7 @@ pub fn summary(req: &mut dyn RequestExt) -> EndpointResult {
         let krates = data.into_iter().map(|(c, _)| c).collect::<Vec<_>>();
 
         let versions: Vec<Version> = krates.versions().load(&*conn)?;
+        let all_versions_yanked = versions.is_empty();
         versions
             .grouped_by(&krates)
             .into_iter()
@@ -53,6 +54,7 @@ pub fn summary(req: &mut dyn RequestExt) -> EndpointResult {
                     None,
                     false,
                     recent_downloads,
+                    all_versions_yanked,
                 ))
             })
             .collect()
@@ -125,6 +127,8 @@ pub fn summary(req: &mut dyn RequestExt) -> EndpointResult {
 
 /// Handles the `GET /crates/:crate_id` route.
 pub fn show(req: &mut dyn RequestExt) -> EndpointResult {
+    use diesel::dsl::exists;
+
     let name = &req.params()["crate_id"];
     let include = req
         .query()
@@ -135,6 +139,14 @@ pub fn show(req: &mut dyn RequestExt) -> EndpointResult {
 
     let conn = req.db_read()?;
     let krate: Crate = Crate::by_name(name).first(&*conn)?;
+
+    let all_versions_yanked = if include.versions {
+        // this query checks if there any non-yanked versions for a crate.
+        // then inverts the result.
+        !diesel::select(exists(krate.versions())).get_result::<bool>(&*conn)?
+    } else {
+        false
+    };
 
     let versions_publishers_and_audit_actions = if include.versions {
         let mut versions_and_publishers: Vec<(Version, Option<User>)> = krate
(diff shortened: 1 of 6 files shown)


src/controllers/krate/metadata.rs
```diff
@@ -202,7 +214,7 @@ pub fn show(req: &mut dyn RequestExt) -> EndpointResult {
     } else {
         None
     };
-    let top_versions = if include.versions {
+    let top_versions = if include.versions && !all_versions_yanked {
```
mario-areias, line 217: Same as above.

src/controllers/krate/publish.rs
```diff
@@ -253,7 +253,14 @@ pub fn publish(req: &mut dyn RequestExt) -> EndpointResult {
...
 
         Ok(req.json(&GoodCrate {
-            krate: EncodableCrate::from_minimal(krate, Some(&top_versions), None, false, None),
+            krate: EncodableCrate::from_minimal(
+                krate,
+                Some(&top_versions),
+                None,
+                false,
+                None,
+                false,
```
mario-areias, line 262: Assuming there is no way to publish an yanked crate so setting up this to false.

Turbo87:
did you check what effects this has on the frontend?

mario-areias:
@Turbo87 I did a basic test in my local machine, publishing crates locally. I just added screenshots in the PR description for the changes I can see. I can confirm the API is responding correctly too. 

I don't know much about the frontend code though (specifically about badges for example), if you can point me what other tests I should do locally, I will certainly do them.

Turbo87:
FWIW the reason I haven't felt comfortable yet with merging this is that it increases the complexity of the code quite a bit, and in some cases even requires an extra database call.

I'm wondering if it would be better to have a sort of "default version" reference from the crates table to the versions table and store the corresponding data there instead. 🤔

Turbo87:
as mentioned in https://github.com/rust-lang/crates.io/issues/1058#issuecomment-1531276138, we decided to change course on how we want to deal with and display yanked crates/versions.

sorry for letting this issue be in limbo for so long :-/

PR #5315, opened 2022-10-11
Ignore "exact match" for search when sorted by downloads

### LINKED ISSUE #5247: Sorting by downloads and recent-downloads produces unexpected results
### Current Behavior

Sorting by either (all-time) downloads or by recent-downloads returns results that are not (properly) sorted.

E.g. the [search for `multiset` ordered by downloads in the API](https://crates.io/api/v1/crates?q=multiset&sort=downloads) returns the following results:

| # | id | downloads |
|-|-|-|
| 1 | multiset | 37315 |
| 2 | hashbag | 363685 |
| 3 | mset | 39412 |
| 4 | hyperloglogplus | 16332 |
| 5 | threshold | 13464 |
| 6 | winter-air | 10209 |
| 7 | utote | 2260 |
| 8 | more_collections | 2207 |
| 9 | sets_multisets | 1030 |
| 10 | winterfell | 900 |

It took me quite a while to realise that `multiset` is placed first here because it matches the search term; the *expected* order would be (row numbers retained to aid referencing):

| # | id | downloads |
|-|-|-|
| 2 | hashbag | 363685 |
| 3 | mset | 39412 |
| 1 | multiset | 37315 |
| 4 | hyperloglogplus | 16332 |
| 5 | threshold | 13464 |
| 6 | winter-air | 10209 |
| 7 | utote | 2260 |
| 8 | more_collections | 2207 |
| 9 | sets_multisets | 1030 |
| 10 | winterfell | 900 |

So with `multiset` with 37315 placed 3rd.

The same applies to `sort=recent_downloads`

### Expected Behavior

When ordering search results by all-time  or by recent downloads, *only sort by those numbers*. An exact keyword match should not matter here, that's what `relevance` sorting is for.

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

gabrielleecredera:
When searching for crates with results order specified (by all-time / recent downloads), ignore the "exact match" for the result ordering.

Resolves #5247

### DIFF
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -220,16 +220,17 @@ pub fn search(req: &mut dyn RequestExt) -> EndpointResult {
         ));
     }
 
+    // Any sort other than 'relevance' (default) would ignore exact crate name matches
     if sort == Some("downloads") {
         // Custom sorting is not supported yet with seek.
         supports_seek = false;
 
-        query = query.then_order_by(crates::downloads.desc())
+        query = query.order(crates::downloads.desc())
     } else if sort == Some("recent-downloads") {
         // Custom sorting is not supported yet with seek.
         supports_seek = false;
 
-        query = query.then_order_by(recent_crate_downloads::downloads.desc().nulls_last())
+        query = query.order(recent_crate_downloads::downloads.desc().nulls_last())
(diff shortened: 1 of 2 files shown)


Turbo87 (approved):
LGTM, thank you! :)

PR #5324, opened 2022-10-13
Remove outdated "Deploying & using a mirror" instructions

Turbo87:
This document is five years old, and while some of the information is still correct we do not support running the crates.io project as a registry mirror (anymore). We will remove this piece of documentation here to avoid confusion for any users looking at this.


JohnTitor (approved):

CapableWeb:
Would it maybe be more appropriate to point users in a different direction if it's no longer advised to run things that way?

As it stands right now, bunch of other things point to the MIRROR.md document and as this is the internet, there is bound to be lots of links that now leads you to a unhelpful 404 page instead of just a page describing the old process + message about it has been deprecated.

Turbo87:
that's a fair point. I don't know what else to recommend as alternatives though. all I know is that crates.io hasn't officially supported mirroring for multiple years at this point.

PR #5342, opened 2022-10-18
Adjust search input placeholder text

ahmedkabd:
The search section displays "Click or press 'S' to search...", which doesn't make any sense.


app/components/search-form.hbs
```diff
@@ -22,7 +22,7 @@
     local-class="input-lg"
     name="q"
     id="cargo-desktop-search"
-    placeholder="Click or press 'S' to search..."
```
Turbo87, line 25: The "Click or press S" refers to how to get the search field focussed. Since we auto focus it these days this can be a bit misleading 😅

ahmedkabd: It definitely is misleading. How about making it more clear?

PR #5347, opened 2022-10-18
Use AWS_ environment variables for authenticating to AWS services

arlosi:
Infra team has requested that we use a single environment variable set for authenticating for AWS services.

This PR standardizes on: `AWS_ACCESS_KEY` and `AWS_SECRET_KEY`.


Turbo87 (approved):
I was gonna ask if we can fall back to the old env vars for now, but then I realized that we can just have both env vars in parallel on production before we deploy 😅 

LGTM, thank you! :)

Turbo87:
for anyone following this: I've added the required env vars to the staging and production Heroku apps, so this should be safe to deploy

PR #5370, opened 2022-10-21
Use a fluid space scale

Turbo87:
Buckle up, this is a big one...! 😅 

This PR introduces several new CSS variables prefixed with `--space`, which should be used instead of hardcoding arbitrary pixel values. These values are also "fluid" in that they change their values depending on the viewport width. The concepts behind this are explained at https://utopia.fyi/.

We are using this scale: https://utopia.fyi/space/calculator?c=320,14,1.2,1140,18,1.25,5,2,&s=0.75|0.5|0.25|0.125,1.5|2|3|4|6,s-l

The pages look quite similar before and after this change, but some of them scale a little better down to mobile viewports.

### DIFF
--- a/app/components/crate-downloads-list.hbs
+++ b/app/components/crate-downloads-list.hbs
@@ -1,6 +1,6 @@
 <ul local-class="list">
   {{#each @crates as |crate|}}
-    <li local-class="row">
+    <li>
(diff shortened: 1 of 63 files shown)


Turbo87:
let's do this...! 🚀

PR #5389, opened 2022-10-28
fix: use new endpoint for crate owner invites

EstebanBorai:
Provides implementation on seek pagination for Pending Owner Invites page.

Previously on: https://github.com/rust-lang/crates.io/pull/5185


Turbo87:
I just realized that there is already an implementation for that route handler and I just forgot about it... 🤦‍♂️ 

see https://github.com/rust-lang/crates.io/blob/59e4fd4ab2be9f37c40f7beb536415dda81d2125/mirage/route-handlers/invites.js 😅 

sorry for the back and forth. I think with this out of the way the next step would be to switch the existing Ember Data code over to using the new endpoint instead.

since this PR was AFAIU only about adding the mirage endpoint I'll go ahead and close it until we have a PR that updates the Ember Data code :)

Turbo87:
sorry, I meant to close https://github.com/rust-lang/crates.io/pull/5397 instead 😅

PR #5397, opened 2022-10-31
feat: update mirage tests and route handlers for owner invites

EstebanBorai:
Uses `api/private` over `api/v1/me` for mirage pending owner invites endpoint.
And also adds the `expires_at` attribute as part of the payload received as
found in the `EncodableCrateOwnerInvitationV1` struct from the API.

### DIFF
--- a/mirage/route-handlers/me.js
+++ b/mirage/route-handlers/me.js
@@ -98,6 +98,15 @@ export function register(server) {
     return { ok: true };
   });
 
+  server.get('/api/v1/me/crate_owner_invitations', function (schema) {
+    let { user } = getSession(schema);
+    if (!user) {
+      return new Response(403, {}, { errors: [{ detail: 'must be logged in to perform that action' }] });
+    }
+
+    return schema.crateOwnerInvitations.where({ inviteeId: user.id });
+  });
+
   server.get('/api/private/crate_owner_invitations', function (schema) {
     let { user } = getSession(schema);
     if (!user) {
--- a/mirage/route-handlers/me.js
+++ b/mirage/route-handlers/me.js
@@ -98,7 +98,7 @@ export function register(server) {
     return { ok: true };
   });
 
-  server.get('/api/v1/me/crate_owner_invitations', function (schema) {
+  server.get('/api/private/crate_owner_invitations', function (schema) {
     let { user } = getSession(schema);
     if (!user) {
       return new Response(403, {}, { errors: [{ detail: 'must be logged in to perform that action' }] });
--- a/tests/mirage/me/crate-owner-invitations/list-test.js
+++ b/tests/mirage/me/crate-owner-invitations/list-test.js
@@ -5,6 +5,101 @@ import fetch from 'fetch';
 import { setupTest } from '../../../helpers';
 import setupMirage from '../../../helpers/setup-mirage';
 
+module('Mirage | GET /api/v1/me/crate_owner_invitations', function (hooks) {
+  setupTest(hooks);
+  setupMirage(hooks);
+
+  test('empty case', async function (assert) {
+    let user = this.server.create('user');
+    this.server.create('mirage-session', { user });
+
+    let response = await fetch('/api/v1/me/crate_owner_invitations');
+    assert.equal(response.status, 200);
+    assert.deepEqual(await response.json(), { crate_owner_invitations: [] });
+  });
+
+  test('returns the list of invitations for the authenticated user', async function (assert) {
+    let nanomsg = this.server.create('crate', { name: 'nanomsg' });
+    this.server.create('version', { crate: nanomsg });
+
+    let ember = this.server.create('crate', { name: 'ember-rs' });
+    this.server.create('version', { crate: ember });
+
+    let user = this.server.create('user');
+    this.server.create('mirage-session', { user });
+
+    let inviter = this.server.create('user', { name: 'janed' });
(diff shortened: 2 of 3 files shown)


mirage/route-handlers/me.js
```diff
@@ -98,7 +98,7 @@ export function register(server) {
     return { ok: true };
   });
 
-  server.get('/api/v1/me/crate_owner_invitations', function (schema) {
+  server.get('/api/private/crate_owner_invitations', function (schema) {
```
Turbo87, line 101: it looks like several tests are failing now, which makes sense, because they are currently still using the old endpoint. in this first step I would only add the new mirage endpoint, but also keep the old one at the same time, until we've migrated the frontend code to use the new endpoint.

tests/mirage/me/crate-owner-invitations/list-test.js
```diff
@@ -55,6 +55,7 @@ module('Mirage | GET /api/v1/me/crate_owner_invitations', function (hooks) {
           invited_by_username: 'janed',
           invitee_id: Number(user.id),
           inviter_id: Number(inviter.id),
+          expires_at: '2017-01-24T12:34:56Z',
```
Turbo87, line 58: while this is generally useful, it is somewhat unrelated to the new mirage endpoint AFAICT, so it might be easier to put this in a dedicated PR.

we probably also would want to add this new field to the corresponding mirage "factory", so that all endpoints will always serialize the new field.

EstebanBorai: I took this field from the struct here: https://github.com/rust-lang/crates.io/blob/8be845c314905f95fc972419c67eba57084dae2c/src/views.rs#L117

Noticed that is also returned by the controller here, https://github.com/rust-lang/crates.io/blob/3d712177dafbd972aa04177690a4aa8fcbfb53c2/src/controllers/crate_owner_invitation.rs#L42

I could introduce it when working on factories?

mirage/route-handlers/me.js
EstebanBorai: Addressed on [7ecadfd](https://github.com/rust-lang/crates.io/pull/5397/commits/7ecadfd0f90d39732bf4a13d9092552ed7fa826b)

tests/mirage/me/crate-owner-invitations/list-test.js
Turbo87: sure, but what you linked is Rust code, which we don't run at all during the frontend tests, so on the mirage mock backend side there is currently no code that crates those fields

tests/mirage/me/crate-owner-invitations/list-test.js
```diff
@@ -5,15 +5,15 @@ import fetch from 'fetch';
 import { setupTest } from '../../../helpers';
 import setupMirage from '../../../helpers/setup-mirage';
 
-module('Mirage | GET /api/v1/me/crate_owner_invitations', function (hooks) {
+module('Mirage | GET /api/private/crate_owner_invitations', function (hooks) {
```
Turbo87, line 8: we should probably also keep the tests for the old endpoint until we receive it :)

mirage/route-handlers/me.js
```diff
@@ -107,6 +107,15 @@ export function register(server) {
     return schema.crateOwnerInvitations.where({ inviteeId: user.id });
   });
 
+  server.get('/api/private/crate_owner_invitations', function (schema) {
+    let { user } = getSession(schema);
+    if (!user) {
+      return new Response(403, {}, { errors: [{ detail: 'must be logged in to perform that action' }] });
+    }
+
+    return schema.crateOwnerInvitations.where({ inviteeId: user.id });
```
Turbo87, line 116: the new endpoint actually works a little bit different. it supports (and iirc requires) query parameters to determine which invites to show.

tests/mirage/me/crate-owner-invitations/list-test.js
EstebanBorai: Yeah, my goal was to be transparent with structure in the server. Perhaps I misunderstanding this field? But I wanted to mock the responses from server to match them with what Mirage will return. I could either remove this or update Mirage factories to include this, not sure to be honest now that I'm a bit confused.

EstebanBorai: Removed the expires at field on [974476a](https://github.com/rust-lang/crates.io/pull/5397/commits/974476a2eba5cc8173cd694f71dd5cd39b4c25f1)

tests/mirage/me/crate-owner-invitations/list-test.js
EstebanBorai: Reintroduce it here: [9fde45f](https://github.com/rust-lang/crates.io/pull/5397/commits/9fde45f06005902a17e91869ede891a755b95342)

Turbo87:
I just realized that there is already an implementation for that route handler and I just forgot about it... 🤦‍♂️

see https://github.com/rust-lang/crates.io/blob/59e4fd4ab2be9f37c40f7beb536415dda81d2125/mirage/route-handlers/invites.js 😅

sorry for the back and forth. I think with this out of the way the next step would be to switch the existing Ember Data code over to using the new endpoint instead.

since this PR was AFAIU only about adding the mirage endpoint I'll go ahead and close it until we have a PR that updates the Ember Data code :)

PR #5407, opened 2022-11-03
fix: diesel cli install command

EstebanBorai:
Running: `cargo install diesel_cli --no-default-features --features postgres --version 1`
results on invalid SemVer version: `the `--version` provided, `1`, is not a valid semver version`.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -285,7 +285,7 @@ by typing `\q`) without any errors to connect to your running Postgres server.
 On all platforms, install through `cargo` by running:
 
 ```
-cargo install diesel_cli --no-default-features --features postgres --version 1
+cargo install diesel_cli --no-default-features --features postgres --version ^1
 ```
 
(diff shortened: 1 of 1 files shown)


EstebanBorai:
@Turbo87 just noticed this going through the Diesel CLI installation command while setting up my local environment
for Rust development.

This is the original log:

```bash
cargo install diesel_cli --no-default-features --features postgres --version 1
error: the `--version` provided, `1`, is not a valid semver version: cannot parse '1' as a semver

if you want to specify semver range, add an explicit qualifier, like ^1
➜  crates.io git:(feat/update-crate-owner-invitations-mirage) cargo install diesel_cli --no-default-features --features postgres --version ^1
    Updating crates.io index
  Downloaded diesel_cli v1.4.1
  Downloaded 1 crate (48.0 KB) in 1.77s
```

Turbo87:
huh, that's surprising. I thought the CLI option would work the same way as a dependency in the Cargo.toml file, but apparently it doesn't 😅

Turbo87 (approved):

Issue #5419, opened 2022-11-04
Recent download count higher than all-time


allan2:
### Current Behavior

I recently published a new crate. When I searched for it on crates.io, I saw this:
![new-crate-search-page](https://user-images.githubusercontent.com/6740989/200065211-f362bfde-0b8d-4f7c-ae2e-426e49ff343c.png)

The response body of *https://crates.io/api/v1/crates/my-crate-name* contained `"downloads": 2` and `"recent_downloads": 4`.

carols10cents:
The all-time count is processed in a background job, so it can be delayed. Not sure what to do about that.

PR #5426, opened 2022-11-06
downloads: Use `recent_downloads` as `downloads` if it is higher

Turbo87:
This resolves https://github.com/rust-lang/crates.io/issues/5419. Not a perfect solution, but in an eventually consistent system this is hard to fix properly... 🤷


Turbo87:
@carols10cents @jtgeibel any thoughts/objections on this?

PR #5451, opened 2022-11-09
Vendor a portion of the `swirl` crate

jtgeibel:
This PR vendors about 800 lines of code (of which ~150 are tests). This PR is split off from my port to diesel 2.0. In particular, the new version of diesel requires that a `&mut PgConnection` be passed around and it should be much easier to finish that port once we can modify this in-tree.

This also addresses a long standing pain point, which is that due to the use of `inventory` the background workers (and thus a lot of unnecessary dependencies) end up linked into every binary increasing build times and file sizes.

The amount of additional code for us to maintain can probably be trimmed further. For now I've simply extracted this from my `diesel2` branch and made sure it compiles and I think this is ready for initial review. Locally, 4 of the `unhealthy_database` tests are failing for me, but I'm also seeing that intermittently on master so it could be my local setup.

Reduction in release binaries:
* background-worker: 27M, unchanged
* crates-admin: 29M to 27M
* enqueue-job: 25M to 14M
* monitor: 25M to 18M
* server: 35M to 30M

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -418,10 +418,10 @@ dependencies = [
  "serde_json",
  "sha2",
  "spdx",
- "swirl",
  "tar",
  "tempfile",
  "thiserror",
+ "threadpool",
  "tikv-jemallocator",
  "tokio",
  "toml",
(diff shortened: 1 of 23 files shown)


Turbo87 (approved):
looks good to me on first glance 👍 

do we want to try this out on staging first? anything to look out for before we can deploy this?

Turbo87:
status report: Justin is planning to deploy and test this on our staging environment over the weekend :)

jtgeibel:
I've rebased this PR and have tested it extensively on staging. This should be good to go.

PR #5479, opened 2022-11-15
cloudfront: Retry requests on server or network error

Turbo87:
While testing out invalidations for the HTTP-based crate index we noticed that the CloudFront server often returns 503 errors. The official (but still experimental) AWS SDK uses a default request retry policy (max. 3 attempts, exponential backoff with 1 sec base delay) to apparently avoid these issues. This PR introduces the `retry` dependency and uses it with a similar retry policy for the CloudFront requests.

/cc @arlosi

### DIFF
--- a/src/worker/cloudfront.rs
+++ b/src/worker/cloudfront.rs
@@ -6,6 +6,8 @@ use aws_sigv4::{
     SigningParams,
 };
 use reqwest::blocking::Client;
+use retry::delay::{jitter, Exponential};
+use retry::OperationResult;
 
 #[derive(Clone)]
 pub struct CloudFront {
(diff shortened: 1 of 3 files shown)


src/worker/cloudfront.rs
```diff
@@ -49,30 +58,61 @@ impl CloudFront {
...
+                        .text()
+                        .unwrap_or_else(|_| "Failed to decode response payload".to_string());
+
+                    Err(error).context(format!("{headers}; body: {body}"))
+                }
+            };
+
+            match result {
+                Ok(_) => OperationResult::Ok(()),
+                Err(error) if status.is_server_error() => OperationResult::Retry(error),
```
arlosi, line 112: We may want to retry on more than just server errors (5xx). The [AWS SDK](https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html) says

> Transient errors (HTTP status codes 400, 408, 500, 502, 503, and 504) and throttling errors (HTTP status codes 400, 403, 429, 502, 503, and 509) can all be retried.

arlosi (approved):

src/worker/cloudfront.rs
Turbo87: hmm... it feels a bit weird to me to retry client errors, because those would indicate that something is wrong on our side 🤔 

I'd say let's try it with the current config, and hopefully the official AWS SDK will be stable soonish and then we can migrate to that.

Issue #3400, opened 2021-03-12
Automatically respond to crates.io API token leaks on GitHub (via GitHub secret scanning)


greysteil:
**Is your feature request related to a problem? Please describe.**
Token leaks [happen](https://github.com/ChALkeR/notes/blob/master/Do-not-underestimate-credentials-leaks.md). There's been previous discussion about [avoiding them in cargo](https://github.com/rust-lang/cargo/issues/2063) but I couldn't find any discussion about automatically responding to them.

I'd like crates.io to automatically revoke any credentials that are leaked into a GitHub public repo. I run the [GitHub secret scanning](https://docs.github.com/en/developers/overview/secret-scanning) team, so I can provide you with a feed of potential matches we find in public repos, if you're willing to take action on them automatically.

We have a similar setup with folks like AWS - if you leak your AWS credentials on GitHub you'll quickly get an email from AWS saying the credential has been quarantined. Under the hood, they're receiving the feed of matches from GitHub.

**Describe the solution you'd like**
crates.io would have an endpoint for receiving details of leaked secrets from GitHub, following the [specification here](https://docs.github.com/en/developers/overview/secret-scanning#create-a-secret-alert-service). crates.io would automatically revoke leaked secrets, and would email their owners to notify them that they had been revoked. The email would include details of where the secret had been exposed (as provided by GitHub).

**Describe alternatives you've considered**
crates.io could notify users of leaked secrets, but not automatically revoke them. However, this would leave leaked credentials exposed.

Turbo87:
@greysteil would https://github.com/rust-lang/crates.io/issues/2653 fit what you're requesting here?

greysteil:
If API tokens were able to self-revoke then we could hit the self-revoke endpoint with our findings, yep. It might feel a bit weird, though - since the tokens we find don't "belong" to us, it's not totally natural that we would presume we can automatically revoke them. They do "belong" to crates (together with the user), so it feels more natural that you would initiate the revoke (and GitHub is just informing you of the leak).

Turbo87:
@greysteil we've discussed this topic in our team meeting today. we generally agree that participating in the automated token scanning would be beneficial. we are however currently in the transition phase from Mozilla to the Rust Foundation (see https://foundation.rust-lang.org), and there are a few open question that the foundation would have to answer before we can move forward here.

in other words: we will have to put this issue on hold for a few weeks until the situation is a bit more structured, but we do plan to return to this :)

[...]

PR #5495, opened 2022-11-18
Add API endpoint to receive alerts of exposed API tokens from GitHub

dangardner:
Adds a new POST endpoint at `/tokens/alert/github` to receive alerts from GitHub when crates.io API tokens are exposed.

Requires adding a dependency on [ring](https://crates.io/crates/ring) for ECDSA signature validation.

Intended to resolve [#3400](https://github.com/rust-lang/crates.io/issues/3400).

### DIFF
--- a/src/email.rs
+++ b/src/email.rs
@@ -91,6 +91,33 @@ or go to https://{domain}/me/pending-invites to manage all of your crate ownersh
         self.send(email, subject, &body)
     }
 
+    /// Attempts to send an API token exposure notification email
+    pub fn send_token_exposed_notification(
+        &self,
+        email: &str,
+        url: &str,
+        reporter: &str,
+        source: &str,
+        token_name: &str,
+    ) -> AppResult<()> {
+        let subject = "Exposed API token found";
+        let mut body = format!(
+            "{reporter} has notified us that your crates.io API token {token_name}\n
+has been exposed publicly. We have revoked this token as a precaution.\n
+Please review your account at https://{domain} to confirm that no\n
+unexpected changes have been made to your settings or crates.\n
+\n
+Source type: {source}\n",
+            domain = crate::config::domain_name()
+        );
+        if url.is_empty() {
+            body.push_str("\nWe were not informed of the URL where the token was found.\n");
+        } else {
+            body.push_str(&format!("\nURL where the token was found: {url}\n"));
+        }
+        self.send(email, subject, &body)
+    }
+
     /// This is supposed to be used only during tests, to retrieve the messages stored in the
     /// "memory" backend. It's not cfg'd away because our integration tests need to access this.
     pub fn mails_in_memory(&self) -> Option<Vec<StoredEmail>> {
--- a/src/router.rs
+++ b/src/router.rs
@@ -157,6 +157,12 @@ pub fn build_router(app: &App) -> RouteBuilder {
         C(crate_owner_invitation::private_list),
     );
 
+    // Alerts from GitHub scanning for exposed API tokens
+    router.post(
+        "/api/github/secret-scanning/verify",
+        C(github::secret_scanning::verify),
+    );
+
     // Only serve the local checkout of the git index in development mode.
     // In production, for crates.io, cargo gets the index from
     // https://github.com/rust-lang/crates.io-index directly.
--- a/src/tests/util/github.rs
+++ b/src/tests/util/github.rs
@@ -1,3 +1,4 @@
+use cargo_registry::controllers::github::secret_scanning::GitHubPublicKey;
 use cargo_registry::github::{
     GitHubClient, GitHubOrgMembership, GitHubOrganization, GitHubTeam, GitHubTeamMembership,
     GithubUser,
@@ -41,6 +42,14 @@ pub(crate) const MOCK_GITHUB_DATA: MockData = MockData {
             login: "user-org-owner",
             name: "User owning the org",
             email: "owner@example.com",
+        },
+    ],
+    // Test key from https://docs.github.com/en/developers/overview/secret-scanning-partner-program#create-a-secret-alert-service
+    public_keys: &[
+        MockPublicKey {
+            key_identifier: "f9525bf080f75b3506ca1ead061add62b8633a346606dc5fe544e29231c6ee0d",
+            key: "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEsz9ugWDj5jK5ELBK42ynytbo38gP\nHzZFI03Exwz8Lh/tCfL3YxwMdLjB+bMznsanlhK0RwcGP3IDb34kQDIo3Q==\n-----END PUBLIC KEY-----",
+            is_current: true,
         },
     ],
 };
@@ -159,11 +168,16 @@ impl GitHubClient for MockGitHubClient {
             Err(not_found())
         }
     }
+
+    fn public_keys(&self, _username: &str, _password: &str) -> AppResult<Vec<GitHubPublicKey>> {
+        Ok(self.data.public_keys.iter().collect())
+    }
 }
 
 pub(crate) struct MockData {
     orgs: &'static [MockOrg],
     users: &'static [MockUser],
+    public_keys: &'static [MockPublicKey],
 }
 
 struct MockUser {
@@ -184,4 +198,30 @@ struct MockTeam {
     id: i32,
     name: &'static str,
     members: &'static [&'static str],
+}
+
+struct MockPublicKey {
+    key_identifier: &'static str,
+    key: &'static str,
+    is_current: bool,
+}
+
+impl From<&'static MockPublicKey> for GitHubPublicKey {
+    fn from(k: &'static MockPublicKey) -> Self {
+        Self {
+            key_identifier: k.key_identifier.to_string(),
+            key: k.key.to_string(),
+            is_current: k.is_current,
+        }
+    }
+}
+
+impl FromIterator<&'static MockPublicKey> for Vec<GitHubPublicKey> {
+    fn from_iter<I: IntoIterator<Item = &'static MockPublicKey>>(iter: I) -> Self {
+        let mut c = Vec::new();
+        for k in iter {
+            c.push(k.into());
+        }
+        c
+    }
 }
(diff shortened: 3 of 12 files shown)


src/controllers/token.rs
```diff
@@ -114,3 +117,133 @@ pub fn revoke_current(req: &mut dyn RequestExt) -> EndpointResult {
...
+        .ok_or_else(|| bad_request("missing HTTP header: GITHUB-PUBLIC-KEY-SIGNATURE"))?;
+    let sig = base64::decode(sig)
+        .map_err(|e| bad_request(&format!("failed to decode signature as base64: {e:?}")))?;
+
+    // Fetch list of public keys from GitHub API
+    let app = req.app();
+    let public_keys = app
+        .github
+        .public_keys(&app.config.gh_client_id, &app.config.gh_client_secret)
+        .map_err(|e| bad_request(&format!("failed to fetch GitHub public keys: {e:?}")))?;
```
Turbo87, line 141: do you think it would make sense to cache the keys to avoid hitting the GitHub API rate limit?

src/controllers/token.rs
```diff
@@ -114,3 +117,133 @@ pub fn revoke_current(req: &mut dyn RequestExt) -> EndpointResult {
...
+            &alert.url,
+            "GitHub",
+            &alert.source,
+            &token.name,
+        ),
+    }
+}
+
+/// Handles the `POST /tokens/alert/github` route.
+pub fn alert_github(req: &mut dyn RequestExt) -> EndpointResult {
```
Turbo87, line 219: I'm wondering if we should move this endpoint and the related code into a dedicated module. something like `controllers::github::secret_scanning` maybe?

src/controllers/token.rs
```diff
@@ -114,3 +117,133 @@ pub fn revoke_current(req: &mut dyn RequestExt) -> EndpointResult {
...
+/// Revokes an API token and notifies the token owner
+fn alert_revoke_token(
+    req: &dyn RequestExt,
+    alert: &GitHubSecretAlert,
+) -> Result<(), Box<dyn AppError>> {
+    let conn = req.db_write()?;
+    // not using ApiToken::find_by_api_token in order to preserve last_used_at
+    let token = api_tokens::table
+        .filter(api_tokens::token.eq(alert.token.as_bytes()))
+        .first::<ApiToken>(&*conn)?;
```
Turbo87, line 188: what happens if GitHub sends us a token that is a false-positive and not in our database? if I understand correctly it would trigger the "Error revoking API token in GitHub secret alert" error below, but from the docs it sounds like this should not be handled as an error case, but something that can just occasionally happen.

src/controllers/token.rs
```diff
@@ -114,3 +117,133 @@ pub fn revoke_current(req: &mut dyn RequestExt) -> EndpointResult {
...
+
+    diesel::update(&token)
+        .set(api_tokens::revoked.eq(true))
+        .execute(&*conn)?;
```
Turbo87, line 192: we might be able to use https://www.postgresql.org/docs/current/dml-returning.html here to reduce the number of required database queries. (disclaimer: I haven't checked if/how diesel supports that)

src/email.rs
```diff
@@ -91,6 +91,33 @@ or go to https://{domain}/me/pending-invites to manage all of your crate ownersh
...
+\n
+Source type: {source}\n",
+            domain = crate::config::domain_name()
+        );
+        if url.is_empty() {
+            body.push_str("\nWe were not informed of the URL where the token was found.\n");
+        } else {
+            body.push_str(&format!("\nURL where the token was found: {url}\n"));
+        }
+        self.send(email, subject, &body)
```
Turbo87, line 118: I was wondering whether we should synchronously send the emails within the request handler or delegate that to the background worker queue, but I guess we could try it out like this first and deal with it if we actually see any issues.

src/tests/util/github.rs
```diff
@@ -159,11 +167,25 @@ impl GitHubClient for MockGitHubClient {
...
+    fn public_keys(&self, _username: &str, _password: &str) -> AppResult<Vec<GitHubPublicKey>> {
+        Ok(self
+            .data
+            .public_keys
+            .iter()
+            .map(|k| GitHubPublicKey {
+                key_identifier: k.key_identifier.to_string(),
+                key: k.key.to_string(),
+                is_current: *k.is_current,
+            })
```
Turbo87, line 180: maybe implement `From< MockPublicKey>` for `GitHubPublicKey` below to make the conversion a bit easier :)

Turbo87 (commented):
I left a few minor comments, but this looks pretty good already. nice work!

dangardner:
Thanks for the feedback, I'll be working on that.

src/controllers/token.rs
```diff
@@ -114,3 +117,133 @@ pub fn revoke_current(req: &mut dyn RequestExt) -> EndpointResult {
...
+/// Handles the `POST /tokens/alert/github` route.
+pub fn alert_github(req: &mut dyn RequestExt) -> EndpointResult {
+    let max_size = 8192;
+    let length = req
+        .content_length()
+        .ok_or_else(|| bad_request("missing header: Content-Length"))?;
+
+    if length > max_size {
+        return Err(bad_request(&format!("max content length is: {max_size}")));
+    }
```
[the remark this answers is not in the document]
dangardner: Here's an example of what I expect an alert to look like, based on GitHub's docs:

`[{"token":"6ORQU1djLP/WGj2JS473V7+vsaTQKFJs5nw","type":"crates.io API token","url":"https://github.com/rust-lang/crates.io/pull/5495/commits/2c285a790a4b3fff8245dc5a7d3298978b23b87f","source":"GitHub commit message"}]`

This is 218 bytes. Rounding that up to 256 bytes gives us 32 alerts per request. Do you mind if we ask @greysteil to confirm if this OK from a GitHub perspective in https://github.com/rust-lang/crates.io/issues/3400?

src/controllers/token.rs
dangardner: I'm happy to add caching if you would prefer. I couldn't see any other responses from GitHub being cached, so I assumed this would be OK too.

src/controllers/token.rs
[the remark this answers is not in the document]
greysteil: Oh niiiice!! Let me get someone from the secret scanning partnerships team to answer any questions for y'all

src/controllers/token.rs
jamisonhyatt: :wave: 
We recommend making authenticated requests (e.g., with a PAT) to fetch the public keys to avoid rate limiting.

src/controllers/token.rs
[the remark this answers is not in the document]
jamisonhyatt: Our default (currently) is to report two (2) alerts at a time, unless explicitly overridden.

src/controllers/token.rs
jamisonhyatt: You have the option of providing feedback for false positives in the response: https://docs.github.com/en/developers/overview/secret-scanning-partner-program#provide-feedback-for-false-positives

src/controllers/token.rs
dangardner: We do make an authenticated request to fetch the public keys but that counts towards the 5000 requests per hour rate limit.

jamisonhyatt: I think you'd be fairly safe caching these for a very long time (like for the lifetime of the process), assuming you accounted for if we started signing with a new key, you'd re-fetch.

src/controllers/token.rs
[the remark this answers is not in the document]
Turbo87: > Our default (currently) is to report two (2) alerts at a time, unless explicitly overridden.

sweet, then I guess the 8192 byte limit should be sufficient for now :)

src/controllers/token.rs
Turbo87: @jamisonhyatt FWIW the docs are a bit unclear about whether that payload is meant to be sent as the response. all it says is:

> When you send us feedback

which I initially interpreted as us sending another request to the GitHub API and then wondered what URL to send it to 😅

src/router.rs
```diff
@@ -113,6 +113,7 @@ pub fn build_router(app: &App) -> RouteBuilder {
     router.put("/api/v1/me/tokens", C(token::new));
     router.delete("/api/v1/me/tokens/:id", C(token::revoke));
     router.delete("/api/v1/tokens/current", C(token::revoke_current));
+    router.post("/api/v1/tokens/alert/github", C(token::alert_github));
```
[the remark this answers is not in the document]
dangardner: I went with https://github.com/rust-lang/crates.io/pull/5495/commits/43f77961212edba54cb94ec4762b3846dc4a7ab4#diff-08d36d8c77bb769044ed645092d1240f6e4c15187866104347da55c3709017b7R160-R165 based on your suggestion.

src/controllers/token.rs
dangardner: I added a cache for downloaded public keys that is valid for 24 hours, which seemed like a reasonable balance to strike.

src/controllers/token.rs
dangardner: Moved as suggested.

src/controllers/token.rs
dangardner: Thanks, it wasn't immediately clear to me either. I've added the feedback in the response now.

src/controllers/token.rs
dangardner: I found Diesel's [get_result() function](https://docs.diesel.rs/2.0.x/diesel/query_dsl/trait.RunQueryDsl.html#method.get_result) and used that to revoke and return the revoked token using the same query.

src/email.rs
dangardner: Since we know that, for now at least, GitHub will only send us a maximum of two alerts at a time, I think sending the emails synchronously should be fine.

Turbo87:
rebased once more due to `base64` breaking changes and merge conflicts in `use` statements :)

Turbo87 (approved):
LGTM 👍 

there are a few little improvements left, but those can be done in a follow-up PR :)

src/controllers/token.rs
jamisonhyatt: 👋 just noticed this feedback.  I agree, we'll get those updated.

aashah:
👋 I'm on the Secret Scanning team too. Just tested out the endpoint with some real API tokens, and things look good from our side! I received the email from crates.io as well!

PR #5497, opened 2022-11-18
Social Media Optimization 📱

sanjaiyan-dev:
Adding social medial related `meta` tags.

_Sorry If I made any mistakes :(_

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -6,6 +6,11 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <title>crates.io: Rust Package Registry</title>
+    <meta name="description" content="The Rust community’s crate registry" />
+    <meta name="og:title" content="Rust Package Registry" />
+    <meta name="og:description" content="The Rust community’s crate registry" />
+    <meta name="og:image" content="https://crates.io/assets/Cargo-Logo-Small.png" />
+    <meta name="theme-color" content="#f9f7ec" />
 
     {{content-for 'head'}}
 


Turbo87:
can you explain what the result of this will be? :)

sanjaiyan-dev:
> can you explain what the result of this will be? :)

Hi @Turbo87 ,

Hi,

Actually this helps to add tailored social media previews for website. And `<meta name="theme-color" />` will add a custom app bar colour.


Reference URLs
- https://developers.facebook.com/docs/sharing/webmasters%C2%A0 

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

sanjaiyan-dev:
> can you explain what the result of this will be? :)

Hi @Turbo87 ,

Hi,

Actually this helps to add tailored social media previews for website. And `<meta name="theme-color" />` will add a custom app bar colour.


Reference URLs
- https://developers.facebook.com/docs/sharing/webmasters%C2%A0 

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

app/index.html
```diff
@@ -6,6 +6,11 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <title>crates.io: Rust Package Registry</title>
+    <meta name="description" content="The Rust community’s crate registry" />
+    <meta name="og:title" content="Rust Package Registry" />
+    <meta name="og:description" content="The Rust community’s crate registry" />
```
Turbo87, line 11: this feels a bit redundant given that the `<title>` element already contains very similar wording. isn't Facebook falling back to using that anyway?

sanjaiyan-dev: > this feels a bit redundant given that the `<title>` element already contains very similar wording. isn't Facebook falling back to using that anyway?

Yes, but for `og:image` there is no fallback (not sure) and I thought putting other open graph tags also as a best practice.

sanjaiyan-dev: > this feels a bit redundant given that the `<title>` element already contains very similar wording. isn't Facebook falling back to using that anyway?

@t

sanjaiyan-dev: > this feels a bit redundant given that the `<title>` element already contains very similar wording. isn't Facebook falling back to using that anyway?

@Turbo87 

Yes, but for `og:image` there is no fallback (not sure) and I thought putting other open graph tags also as a best practice.

sanjaiyan-dev: Sorry if I made any mistakes :(

app/index.html
```diff
@@ -6,6 +6,11 @@
...
+    <meta name="og:image" content="https://crates.io/assets/Cargo-Logo-Small.png" />
+    <meta name="theme-color" content="#f9f7ec" />
```
Turbo87, line 13: I just had a closer look at this and it looks like https://github.com/rust-lang/crates.io/blob/69f90830490b1840d86266de719af5520d6512a2/config/manifest.js is already adding the `theme-color` meta tag at build time.

app/index.html
```diff
@@ -6,6 +6,11 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <title>crates.io: Rust Package Registry</title>
+    <meta name="description" content="The Rust community’s crate registry" />
+    <meta name="og:title" content="Rust Package Registry" />
+    <meta name="og:description" content="The Rust community’s crate registry" />
+    <meta name="og:image" content="https://crates.io/assets/Cargo-Logo-Small.png" />
```
Turbo87, line 12: from reading up on the topic it looks like the recommended dimensions of such an image are at least 1200x630, but the linked image is significantly smaller and not really made for this purpose. it might be better to skip the `og:image` until we actually have an image that fits better.

app/index.html
sanjaiyan-dev: Sorry, I think a certain mobile broywser won't apply this special on iOS unless the website is PWA.

Sorry if I am wrong :(

Turbo87:
Since some of the tags are duplicates and the others currently don't add much information aside from what is already in the `<title>` I'll close this PR for now.

Ideally we would use some sort of server-side rendering to produce these tags, so that search engine crawlers and social media previews would show the correct information per crate and/or version, but that work is slightly more involved :-/

PR #5502, opened 2022-11-20
extend doc url blocks to arbitrary subdomains

jendrikw:
As discussed at https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/link.20to.20scam/near/311155912

PR #5503, opened 2022-11-20
Fix documentation URL blocking for substring domains

Turbo87:
#5502 was causing false positives due to the simplified `ends_with()` usage. This PR fixes it by checking against `.{domain}` instead, but avoids unnecessary allocations by performing the original check first.


jendrikw:
Good catch! Will the link of the underscore crate be removed retroactively?

Turbo87:
yeah, that should be fixed once we deploy these changes :)

PR #5511, opened 2022-11-22
worker::cloudfront: Increase retry attempts

Turbo87:
It looks like three retries isn't always enough, and since we run multiple background jobs in parallel retrying ten times should not be a big issue.

### DIFF
--- a/src/worker/cloudfront.rs
+++ b/src/worker/cloudfront.rs
@@ -1,4 +1,4 @@
-use std::time::SystemTime;
+use std::time::{Duration, SystemTime};
 
 use anyhow::Context;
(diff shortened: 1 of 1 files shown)


arlosi:
It looks like the last retry attempt will have a 512 second delay, which might be a bit much. Would it make sense to limit the interval to something like 30 seconds using [`with_max_interval`](https://docs.rs/backoff/latest/backoff/exponential/struct.ExponentialBackoffBuilder.html#method.with_max_interval) ?

Turbo87:
yeah, I guess that's a reasonable suggestion :)

PR #5528, opened 2022-11-23
WIP: Implement basic token scopes

Turbo87:
This PR is **work in progress**!

It will add token scope columns to the database and check these columns in the route handlers that require authentication. 

see https://github.com/rust-lang/crates.io/issues/5443

Issue #5443, opened 2022-11-08
Tracking Issue: Token Scopes


Turbo87:
This is a tracking issue for the RFC "crates.io Token Scopes" (rust-lang/rfcs#2947).

**Steps:**

Backend: 
- [x] add token scopes to the database (https://github.com/rust-lang/crates.io/pull/5562)
- [x] check token scopes for authenticated endpoints (https://github.com/rust-lang/crates.io/pull/5572)
- [x] implement API support for token scopes (https://github.com/rust-lang/crates.io/pull/5973 and https://github.com/rust-lang/crates.io/pull/6310)

Frontend: 
- [x] create a dedicated token creation route (https://github.com/rust-lang/crates.io/pull/6395)
- [x] implement support for endpoint scopes to the route (https://github.com/rust-lang/crates.io/pull/6428)
- [x] implement support for crates scopes to the route (https://github.com/rust-lang/crates.io/pull/6432)
- [x] show token scopes in the API token list (https://github.com/rust-lang/crates.io/pull/6450)

The exact details of these steps are still tbd. Feel free to discuss here or contact the @rust-lang/crates-io team if you have any questions. It might be best to discuss the plans first before working on and opening a PR 😉 

**Unresolved questions:**

- Are there more scopes that would be useful to implement from the start?
- Is the current behavior of crate scopes on endpoints that don't interact with
  crates the best, or should a token with crate scopes prevent access to
  endpoints that don't act on crates?

mbhall88:
Was there a reason for limiting the patterns allowed for token scope to, for example, `serde*`? What I mean is the wildcard is a suffix here, but `*serde` is not allowed - i.e., a wildcard prefix.

My use case is I have a workspace with binary `foo` and a library `libfoo`. I'd like to be able to create a crates API token scoped to `*foo` but his is not possible with the current scoping patterns.

(Apologies if this is not the right place to ask this question)

Turbo87:
@mbhall88 mostly for implementation simplicity. you can add multiple scopes to one token though, so this should still be relatively easy to achieve.

PR #5562, opened 2022-11-30
database/api_tokens: Add `crate_scopes` and `endpoint_scopes` columns

Turbo87:
This is the very first basic step in implementing #5443. The PR just adds the two `crate_scopes` and `endpoint_scopes` columns to the database with a `Option<Vec<String>>` on the Rust side. A follow-up PR will then migrate the Rust side to more specific types (see #5528).

This is PR is as small as possible to reduce the need for a revert with a database migration involved.


Turbo87:
checked this on our staging environment and everything still appears to work well

PR #5569, opened 2022-12-01
Implement `CrateScope` struct

Turbo87:
Similar to #5568, this PR implements a `CrateScope` struct, which we use for the `crate_scopes` column of the API tokens database table. It also includes a `matches()` method to verify whether it matches a crate name or not.

### DIFF
--- a/src/models/token/scopes.rs
+++ b/src/models/token/scopes.rs
@@ -70,6 +70,30 @@ impl TryFrom<&str> for CrateScope {
     }
 }
 
+impl TryFrom<String> for CrateScope {
+    type Error = String;
+
+    fn try_from(pattern: String) -> Result<Self, Self::Error> {
+        match CrateScope::is_valid_pattern(&pattern) {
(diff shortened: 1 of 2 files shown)


src/models/token/scopes.rs
```diff
@@ -50,3 +51,135 @@ impl FromSql<Text, Pg> for EndpointScope {
...
+        if pattern == "*" {
+            return true;
+        }
+
+        let name_without_wildcard = pattern.strip_suffix('*').unwrap_or(pattern);
+        Crate::valid_name(name_without_wildcard)
+    }
+
+    pub fn matches(&self, crate_name: &str) -> bool {
+        let canonicalize = |name: &str| name.replace('-', "_");
```
Turbo87, line 112: for further discussion: do we want this canonicalization? or do we want the users to be precise?

Turbo87:
this feature is experimental. so while there is still an open question, we can adjust that later and still merge the PR for now.

PR #5602, opened 2022-12-07
Vendor `conduit-hyper` crate

Turbo87:
see https://github.com/conduit-rust/conduit-hyper

We are the only users of this crate and we would eventually like to migrate away from using `conduit`. Vendoring the crate into the main crates.io repo will hopefully make it easier to refactor things towards a different web framework.


conduit-hyper/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "conduit-hyper"
+version = "0.4.2"
+authors = ["Justin Geibel <jtgeibel@gmail.com>"]
+license = "MIT OR Apache-2.0"
+description = "Host a conduit based web application on a hyper server"
+repository = "https://github.com/jtgeibel/conduit-hyper"
```
JohnTitor, line 7: How about updating this URL and publishing a new version just in case? It'd be helpful when a crate seeker finds the crate.

JohnTitor (approved):
The transition itself makes sense to me! Leaving one opinion wrt. the crate management.

PR #5647, opened 2022-12-14
Revert `base64` dependency upgrade

Turbo87:
This should temporarily resolve https://github.com/rust-lang/crates.io/issues/5646 until we figure out what exactly went wrong and why the background worker still seems to be running correctly despite the Sentry error reports.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -247,12 +247,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
-name = "base64"
-version = "0.20.0"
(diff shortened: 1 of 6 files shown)


paolobarbolini:
It probably has to do with base64 0.20 requiring correct padding by default when decoding

Turbo87:
yeah, that's possible. I wish their changelog was a bit more explicit about changing the default behavior of their main API interface... 😩

paolobarbolini:
yeah. BTW considering all of the changes that had to be done to move to 0.20 you too might be interested in https://github.com/marshallpierce/rust-base64/issues/205

Turbo87:
oh dear... thanks for the link 🙈

Turbo87:
> why the background worker still seems to be running correctly despite the Sentry error reports.

I figured it out... I did not notice yesterday that the Sentry issues were only for the staging environment, but in production everything was working fine. 😅

PR #5666, opened 2022-12-15
middleware/log_request: Switch to axum-based logging

Turbo87:
This PR drops the logging part of the `conduit` middleware and uses the `axum` middleware for it instead, which is now on par with the `conduit` implementation.


Turbo87:
I've enabled the `axum` logging on production and it was not causing any issues, so I guess we're ready to switch over 🎉

PR #5717, opened 2022-12-21
Bump to latest nginx buildpack

jtgeibel:
I've tested this on staging and it works as expected.

### DIFF
--- a/.buildpacks
+++ b/.buildpacks
@@ -1,3 +1,3 @@
 https://github.com/unfold/heroku-buildpack-pnpm#76cf1fc
-https://github.com/heroku/heroku-buildpack-nginx#4cd76d0
+https://github.com/heroku/heroku-buildpack-nginx#760407b
 https://github.com/emk/heroku-buildpack-rust#cfa0f06


.buildpacks
```diff
@@ -1,3 +1,3 @@
 https://github.com/unfold/heroku-buildpack-pnpm#76cf1fc
-https://github.com/heroku/heroku-buildpack-nginx#4cd76d0
+https://github.com/heroku/heroku-buildpack-nginx#v1.9
```
Turbo87, line 2:
```suggestion
https://github.com/heroku/heroku-buildpack-nginx#760407b
```

I think I'd prefer to keep pointing to a specific commit because git tags can be moved and this ensures that we will see an error if for some reason the tag changes.

Turbo87 (approved):

.buildpacks
jtgeibel: Given the repo is under the heroku org I'm hoping they won't pull such shenanigans, but it makes sense to stay consistent. I've updated the PR and will merge.

PR #5730, opened 2022-12-23
Fix default credentials location

HomyeeKing:


Turbo87:
hmm, https://doc.rust-lang.org/cargo/reference/config.html says that the path is `$CARGO_HOME/credentials.toml`. why do you want to remove the `.toml` suffix?

weihanglo:
Files without `.toml` extension exist for backward compatibility. IIRC Cargo team encourages the use of `credentials.toml`. See also <https://github.com/rust-lang/cargo/blob/74c7aab19a9b3a045e7af13319409a9de2cf4ef7/src/cargo/util/config/mod.rs#L1400-L1404>

> The purpose of this function is to aid in the transition to using
> .toml extensions on Cargo's config files, which were historically not used.
> Both 'config.toml' and 'credentials.toml' should be valid with or without extension.
> When both exist, we want to prefer the one without an extension for
> backwards compatibility, but warn the user appropriately.

HomyeeKing:
in which cargo version add `.toml` suffix mine is `argo 1.65.0 (4bc8f24d3 2022-10-20)`? 
 because I do `cargo login <token>` today then I want to check if the token was stored, but the `~/.cargo/credentials.toml` 
is not exist.
I only found  `~/.cargo/credentials`

weihanglo:
From here https://github.com/rust-lang/cargo/pull/7295 since 1.39, specifically [this comment](https://github.com/rust-lang/cargo/blob/74c7aab19a9b3a045e7af13319409a9de2cf4ef7/src/cargo/util/config/mod.rs#L2096-L2098). However, the transition period was not what I expected or not finished. Feel like Cargo should by default write to `credentials.toml`.

I'd suggest **transferring this to `rust-lang/cargo`**. Cargo then can do such enhancements:

- File a patch writing to `credentials.toml` if `credentials` doesn't exist.
- Document the fallback behavior properly.
- `rust-lang/crates.io` could cross-link to cargo doc to avoid doc duplicate.

I have no enough information of why it designed like that. Could bring it up to the next triage meeting (though most of the team are on vacation 🏝).

Turbo87:
merging this for now until cargo is updated to actually generate `credentials.toml` by default :)

PR #5821, opened 2023-01-04
conduit-test: Simplify `MockRequest` by relying on `http::Request`

Turbo87:
This simplifies the `MockRequest` struct by removing all existing fields and replacing them with a single `request: http::Request<Cursor<bytes::Bytes>>` field.

### DIFF
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -411,7 +411,7 @@ mod tests {
     }
 
     fn mock(query: &str) -> MockRequest {
-        let mut req = MockRequest::new(http::Method::GET, "");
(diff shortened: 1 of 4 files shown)


src/controllers/helpers/pagination.rs
```diff
@@ -411,7 +411,7 @@ mod tests {
     }
 
     fn mock(query: &str) -> MockRequest {
-        let mut req = MockRequest::new(http::Method::GET, "");
+        let mut req = MockRequest::new(http::Method::GET, "/");
```
Turbo87, line 414: an empty `URI` is invalid, so we had to change the default `path` value here to make the tests pass without panicking

PR #5842, opened 2023-01-05
conduit::Handler: Expect `axum::Response` return values

Turbo87:
Instead of using `http::Response<Bytes>` we can use `axum::Response` to bring us closer to being able to return `axum` responses from the crates.io request handlers.

### DIFF
--- a/conduit-axum/src/tests.rs
+++ b/conduit-axum/src/tests.rs
@@ -1,20 +1,24 @@
-use crate::{box_error, ConduitRequest, Handler, HandlerResult};
-use axum::body::Bytes;
+use crate::{ConduitRequest, Handler, HandlerResult};
+use axum::response::IntoResponse;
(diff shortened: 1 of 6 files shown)


conduit-axum/src/tests.rs
```diff
@@ -36,20 +40,7 @@ impl Handler for Panic {
...
-            .body(Bytes::from_static(b"discarded"))
-            .map_err(box_error)
-    }
-}
-
-struct InvalidStatus;
-impl Handler for InvalidStatus {
-    fn call(&self, _req: &mut ConduitRequest) -> HandlerResult {
-        Response::builder()
-            .status(1000)
```
Turbo87, line 50: this case appears to be impossible (?) to reproduce with `axum` responses, but since it's that hard we probably don't need it that much anyway.

PR #5875, opened 2023-01-06
Add `cargo add` instruction to crate-sidebar

stephank:
This adds a `cargo add` instruction to the crate sidebar, with clipboard support:

Screenshot 2023-01-06 at 20 07 01

This came about after @beeb [made a good point](https://hachyderm.io/@beeb/109641418504476058) that copying a snippet from crates.io prevents typosquatting, and not having a snippet for `cargo add` is reason to avoid it.

### DIFF
--- a/app/components/crate-sidebar.js
+++ b/app/components/crate-sidebar.js
@@ -14,6 +14,10 @@ export default class CrateSidebar extends Component {
     return homepage && (!repository || simplifyUrl(repository) !== simplifyUrl(homepage));
   }
 
+  get cargoAddCommand() {
+    return `cargo add ${this.args.crate.name}`;
+  }
+
   get tomlSnippet() {
     return `${this.args.crate.name} = "${this.args.version.num}"`;
   }
--- a/app/components/crate-sidebar.js
+++ b/app/components/crate-sidebar.js
@@ -15,7 +15,9 @@ export default class CrateSidebar extends Component {
   }
 
   get cargoAddCommand() {
-    return `cargo add ${this.args.crate.name}`;
+    return this.args.requestedVersion
+      ? `cargo add ${this.args.crate.name}@${this.args.requestedVersion}`
+      : `cargo add ${this.args.crate.name}`;
   }
 
   get tomlSnippet() {
--- a/app/components/crate-sidebar.hbs
+++ b/app/components/crate-sidebar.hbs
@@ -37,7 +37,23 @@
   <div>
     <h2 local-class="heading">Install</h2>
 
-    <p local-class="copy-help">Add the following line to your Cargo.toml file:</p>
+    <p local-class="copy-help">Run the following Cargo command in your project directory:</p>
+    {{#if (is-clipboard-supported)}}
+      <CopyButton
+        @copyText={{this.cargoAddCommand}}
+        title="Copy command to clipboard"
+        local-class="copy-button"
+      >
+        <span>{{this.cargoAddCommand}}</span>
+        {{svg-jar "copy" aria-hidden="true" local-class="copy-icon"}}
+      </CopyButton>
+    {{else}}
+      <code local-class="copy-fallback">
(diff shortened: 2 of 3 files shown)


beeb:
Awesome feature! Thanks!

app/components/crate-sidebar.js
```diff
@@ -14,6 +14,10 @@ export default class CrateSidebar extends Component {
     return homepage && (!repository || simplifyUrl(repository) !== simplifyUrl(homepage));
   }
 
+  get cargoAddCommand() {
+    return `cargo add ${this.args.crate.name}`;
```
Turbo87, line 18:
```suggestion
    return `cargo add ${this.args.crate.name}@${this.args.version.num}`;
```

any thoughts about including the version number, like we do for the `Cargo.toml` snippet?

I understand that `cargo add` will automatically use the latest version, but if you're viewing the page of a specific version then maybe it would make sense to include the version in the command?

Turbo87 (approved):
LGTM, thank you!

I'm wondering about adding the version to the command, but that probably doesn't need to block this PR. let me know if I should merge as is and then we can still improve this afterwards :)

app/components/crate-sidebar.js
stephank: Good point! I didn't consider version pages. I added a commit that adds the version only on version detail pages. The regular crate detail page still shows the versionless variant. That makes the most sense, I think?

epage: > Good point! I didn't consider version pages. I added a commit that adds the version only on version detail pages. The regular crate detail page still shows the versionless variant. That makes the most sense, I think?

This is exactly the solution I was imagining when I was considering this idea as `cargo-add` was being merged into cargo.

Thanks for taking this on! I originally held off on getting this done as I wanted more time for versions of `cargo` with `add` to proliferate but lost track of coming back to this.

Turbo87 (approved):
awesome! thanks again! :)

nrxus:
Is there a way for a crate to specify that it's meant to be a dev dependency only? I am an author for a crate that explicitly only wants to be added as a dev dependency because it modifies a lot of code at compile time to make structs mockable. While just adding the dependency alone won't hurt anything my README explicitly states to only add it as a dev dependency to make it fool proof that nothing bad can happen outside of tests. While I understand the motivation for this change I think it unintentionally implies that all crates are meant to be installed the same which is not the case which could cause issues for crates that intentionally want to only be for tests or during build.

lnicola:
@nrxus `cargo add --dev foo`

lnicola:
@nrxus Sorry for the previous `cargo add --dev` comment, I submitted it too quickly.

That said, I've found that even crates usually meant to be used as dev dependencies can be useful in other contexts, like code generator tools, perhaps. I suppose you could check for `cfg(test)` somewhere in your generated code, though that's indeed not ideal. Has this been a problem for your users in the past?

Turbo87:
@nrxus I don't think there are currently any flags implemented to indicate that. I'm open to the suggestion though. Do you have any thoughts on how we could implement it?

nrxus:
The most flexible thing would be to an extra field in `Cargo.toml` that allows you to override the install command shown in crates.io:

```toml
[package]
name = "faux"
install-command = "cargo add --dev"
```

riscygeek:
> The most flexible thing would be to an extra field in `Cargo.toml` that allows you to override the install command shown in crates.io:
> 
> ```toml
> [package]
> name = "faux"
> install-command = "cargo add --dev"
> ```

@nrxus I don't think that this is ideal, because someone malicious could do something like that:
```toml
install-command = "rm -rf *"
```

I would opt for something like this:
```toml
add-as = "dev" # or "build"
```
It's a little uglier and less flexible, but safe. Otherwise you'd have to sanitize/check the install-command for every crate.

nrxus:
That makes sense to me 👍

PR #5918, opened 2023-01-12
Share the existing background job connection with (only) some jobs

jtgeibel:
This should resolve the deadlock issues seen in the initial implementation in #5837. The first commit reverts the revert done in #5909. The second commit then adds the connection pool back to the state available within `Job::perform()`, though now the pool is an `Option<_>` and only available in production. If test coverage of these background jobs is ever desired it could potentially be implemented with the existing infrastructure used in `unhealthy_database` tests.


Turbo87 (approved):
LGTM on first glance 👍

jtgeibel:
I'm not expecting any issues, but I have time to monitor things for a while so I'm going to merge and deploy.

PR #6001, opened 2023-01-25
Clarify a11y category description

kornelski:
The accessibility category has many crates not related to assistive technologies (web crawlers, internationalization, audio parsers, date and time):

https://crates.io/categories/accessibility

I suspect that the description is too broad and not understood by people who aren't already familiar with the "accessibility" term.


Turbo87 (approved):
LGTM, thank you! 👍

PR #6032, opened 2023-02-02
Implement GitHub SSH certificate checks

Turbo87:
Previously, we were relying on libgit2 to check the hostkey fingerprint, but since our Heroku dynos do not have a `known_hosts` file, this couldn't actually have worked, which results in us not checking the fingerprints at all. 

https://github.com/rust-lang/crates.io/pull/5996 updates libgit2, which enables fingerprint checks by default. Due to the missing `known_hosts` file this broke our background worker though.

This PR enables custom fingerprint checks for `github.com` by hardcoding the fingerprints from https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints. As the code comment says, this is most likely good enough, but could eventually be replaced by a `known_hosts`-like environment variable.

### DIFF
--- a/cargo-registry-index/lib.rs
+++ b/cargo-registry-index/lib.rs
@@ -7,6 +7,8 @@ extern crate tracing;
 pub mod testing;
 
 use anyhow::{anyhow, Context};
+use git2::CertificateCheckStatus;
+use once_cell::sync::Lazy;
 use std::cmp::Ordering;
 use std::collections::BTreeMap;
 use std::io::Write;
@@ -17,6 +19,18 @@ use tempfile::TempDir;
 use url::Url;
 
 static DEFAULT_GIT_SSH_USERNAME: &str = "git";
+
+// We are hardcoding the github.com SSH key fingerprints from
+// https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints
+// here. This is considered good enough for now, but could eventually be replaced
+// by an environment variable that acts as a `known_hosts` file.
+static GITHUB_FINGERPRINTS: Lazy<[Vec<u8>; 3]> = Lazy::new(|| {
+    [
+        base64::decode(b"+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU").unwrap(),
+        base64::decode(b"p2QAMXNIC1TJYWeIOttrVc98/R1BUFWu3/LiyKgUfQM").unwrap(),
+        base64::decode(b"nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8").unwrap(),
+    ]
+});
 
 #[derive(Clone)]
 pub enum Credentials {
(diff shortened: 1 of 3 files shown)


cargo-registry-index/lib.rs
```diff
@@ -18,6 +21,24 @@ use url::Url;
...
+    [
+        "SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU"
+            .parse()
+            .unwrap(),
+        "SHA256:p2QAMXNIC1TJYWeIOttrVc98/R1BUFWu3/LiyKgUfQM"
+            .parse()
+            .unwrap(),
+        "SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8"
+            .parse()
+            .unwrap(),
```
pietroalbini, line 38: I wouldn't pull up two dependencies just for this: the part after `SHA256:` is base64-encoded (with a missing `=` for padding), we can just decode it manually and hardcode the decoded byte array.

cargo-registry-index/lib.rs
```diff
@@ -496,9 +517,37 @@ impl Repository {
...
+                return Err(error)
+            };
+
+            if GITHUB_FINGERPRINTS.iter().any(|fingerprint| hash == fingerprint.as_bytes()) {
+                Ok(CertificateCheckStatus::CertificateOk)
+            } else {
+                let mut error = git2::Error::from_str("Certificate check for github.com failed");
+                error.set_code(ErrorCode::Certificate);
+                Err(error)
+            }
```
pietroalbini, line 548: This seems... fine, even though I'd do something like:

```rust
if hostname == "github.com" {
    if let Some(sha256) = cert.as_hostkey().and_then(|key| key.hash_sha256()) {
        if GITHUB_FINGERPRINTS.iter().any(|fingerprint| hash == fingerprint.as_bytes()) {
            return Ok(CertificateCheckStatus::CertificateOk);
        }
    }
}
Ok(CertificateCheckStatus::CertificatePassthrough)
```

That way we don't override the libgit2 behavior if GitHub ever rotates keys.

pietroalbini (changes requested):

cargo-registry-index/lib.rs
Turbo87: ✅

cargo-registry-index/lib.rs
Turbo87: good point. it looks like we already were using the `base64` crate anyway, so I went with that to make it a little more straight forward to update the fingerprints if necessary.

pietroalbini (approved):

ehuss:
Can you say why this is checking the hash instead of the actual key? I suppose a sha256 collision is unlikely, but that's not the normal process for validating a key.

Also, any particular reason it can't just upload a normal known_hosts file?

ehuss:
It would probably be better to check the key. I think it could just be added to the static as a tuple, such as `(SshHostKeyType::Ed255219, base64::decode("..."))`.

Ah, I'm not too familiar with how Heroku works. I was thinking that whatever process copies/installs the crates.io binaries could copy/install arbitrary files.

PR #6037, opened 2023-02-03
Bump CI to use Ubuntu 22.04

jtgeibel:
Staging has been bumped to the heroku-22 stack. Assuming no unexpected
issues, we can bump CI to match and then upgrade production.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ env:
 jobs:
   changed-files:
     name: Changed Files
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v3.3.0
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -115,6 +115,8 @@ jobs:
       - run: rustup default ${{ env.RUST_VERSION }}
 
       - uses: Swatinem/rust-cache@v2.2.0
+        with:
+          prefix-key: "2023-02-02-reset-cache"
```
Turbo87, line 119: hmm... interesting. is this an issue we should report somewhere? cargo maybe? seems to me like an OS update should invalidate the cache automatically, no?

Turbo87 (approved):

.github/workflows/ci.yml
Turbo87: FYI I just rebased the branch on top of `master` and removed the cache buster commit since I deleted the relevant caches manually :)

PR #6043, opened 2023-02-05
background-worker: Speed up index cloning by 3x

Turbo87:
Previously we were using libgit2 to clone the index, which regularly took 2-3 minutes. This PR switches us over to using the `git` CLI, which appears to be significantly faster and clones the index in just over 30 seconds.

### DIFF
--- a/cargo-registry-index/lib.rs
+++ b/cargo-registry-index/lib.rs
@@ -539,23 +539,31 @@ impl Repository {
         let checkout_path = self.checkout_path.path();
         command.current_dir(checkout_path);
 
-        let temp_key_path = self.credentials.write_temporary_ssh_key()?;
-        command.env(
-            "GIT_SSH_COMMAND",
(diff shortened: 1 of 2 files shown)


Eh2406:
cc @Byron, sounds like a good place for gitoxide some day.

Byron:
Thanks for the heads-up! There is nothing stopping you from using `gitoxide` for this today, and depending on the hardware the worker runs on it should be a little faster, but with this repo no more than a second even with ample cores. The repository doesn't actually spend a lot of time in the delta-resolution phase which is the only phase that can be sped up meaningfully.

Since it's a background worker, one would also not be interested in progress here so using the `git` CLI seems like a good choice.

The only advantage one would have with `gitoxide` is to more easily detect network-related errors to retry, but then again, a simple retry-three-times logic might also be sufficient here.

Turbo87:
@jtgeibel FYI I've tested this out on staging and it appears to work fine :)

PR #6143, opened 2023-03-03
Updates nginx config to always send Access-Control-Allow-Origin header

### LINKED ISSUE #6142: Crates.io API does not correctly set Access-Control-Allow-Origin header when looking up crates that do not exist.
### Current Behavior

When using the crates.io API (https://crates.io/api/v1/crates/:name) with a crate name that does not exist, no ` Access-Control-Allow-Origin`  header is sent in the response.

### Expected Behavior

When using the crates.io API (https://crates.io/api/v1/crates/:name) with a crate name that does not exist, an ` Access-Control-Allow-Origin: *`  header should be sent in the response.

### Steps To Reproduce

When looking up a crate that exists, the header is set correctly:
` ` ` 
$ curl -v https://crates.io/api/v1/crates/test
...
< access-control-allow-origin: *
...
` ` ` 

When looking up a crate that does not exist, the header is missing:

` ` ` 
$ curl -v https://crates.io/api/v1/crates/test2
...
*no access-control-allow-origin header in output*
` ` ` 

### Environment

- Browser: curl
- OS: Debian 11

### Anything else?

_No response_

xfbs:
The nginx config injects the Access-Control-Allow-Origin header to allow cross-origin API requests to crates.io. This is done using the add_header directive in the configuration. This directive only injects the header for successful response codes, as per the documentation:

http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header

This means that cross-origin requests to crates.io are allowed when a crate exists (successful status code), but denied when it does not exist (404 response).

This commit fixes that behaviour and allows other sites to make use of the crates.io API, even when looking up crates that do not exist.

This fixes #6142.

Please test this change if possible. I think this change is safe, I think the missing header was an oversight. I do not know how to add a unit test for this to prevent future regressions.

### DIFF
--- a/config/nginx.conf.erb
+++ b/config/nginx.conf.erb
@@ -222,7 +222,7 @@ http {
 		add_header X-XSS-Protection "1; mode=block";
 
 		add_header Content-Security-Policy "default-src 'self'; connect-src 'self' *.ingest.sentry.io https://docs.rs https://play.rust-lang.org https://<%= s3_host(ENV) %>; script-src 'self' 'unsafe-eval' 'sha256-n1+BB7Ckjcal1Pr7QNBh/dKRTtBQsIytFodRiIosXdE='; style-src 'self' 'unsafe-inline' https://code.cdn.mozilla.net; font-src https://code.cdn.mozilla.net; img-src *; object-src 'none'";
-		add_header Access-Control-Allow-Origin "*";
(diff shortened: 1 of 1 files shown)


Turbo87:
https://github.com/rust-lang/crates.io/issues/6142#issuecomment-1456675240 verified that this works well, so I'll go ahead and merge this. thanks for the PR! :)

Issue #6135, opened 2023-03-02
`features2` in index broke cargo backward compatibility


Turbo87:
### Current Behavior

According to https://www.reddit.com/r/rust/comments/11348jp/comment/j8o4nn6/ `ripgrep` can no longer be built by older cargo versions (e.g. 1.13).

### Expected Behavior

Older cargo versions should be able to still build `ripgrep`

### Steps To Reproduce

according to the Reddit thread:

clone the `ripgrep` repo and run `cargo +1.13.0 b --verbose`

### Anything else?

We have identified that the issue is related to how we introduced `features2` to the index. `ripgrep` appears to use `clap` and in v4.0.0-rc.1 `clap` started to use the `dep:` prefix, causing the index to use `features2` field.

The way `clap` used it and crates.io split the `features` map into `features` and `features2` resulted in the following:

```json
{
  "features": {
    "unstable-doc": [
      "derive"
    ]
  },
  "features2": {
    "derive": [
      "dep:clap_derive",
      "dep:once_cell"
    ]
  }
}
```

You can see that the `unstable-doc` feature depends on the `derive` feature. But since `derive` is in `features2` the old cargo version thinks that the feature does not exist.

When splitting the feature maps we should have recursively moved the `unstable-doc` feature to `features2` too, because it depends on a feature which is itself in `features2`.

Turbo87:
After discussing this with the cargo team on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/Old.20cargoes.20failing.20to.20parse.20the.20index/near/340396777), we decided that the easiest way forward is to move all features into `feature2` if we detect that any feature uses the new feature syntax.

Turbo87:
https://github.com/rust-lang/crates.io-index/commit/e5dc6f717d801eee4df4884bc4c309dbb2765606 updated the `clap` index entry, which should finally fix the issue :)

PR #6168, opened 2023-03-09
publish: Move all `features` to `feature2` if new syntax is used

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/6135


ehuss (approved):
Seems reasonable to me.

This may change the error message users get when running `cargo update` in versions 1.19 to 1.51.  There are different scenarios:
* The update selects a version with the new syntax, but did not need any of the features using the new syntax: Previously would get an error parsing the features.
* The update selects a version with the new syntax, and it needed a feature with the new syntax: Previously would get a `failed to select a version` (after 1.26? saying it "does not have these features").

With this change, I think it will always be the second case.

PR #6169, opened 2023-03-09
Add `FixFeature2` background worker job

Turbo87:
This job "fixes" the `features2` entries by moving all `features` into `feature2` if `feature2` already has content. This should avoid a mismatch of `features` pointing to `features2` for older cargo versions.

Related:
- https://github.com/rust-lang/crates.io/issues/6135
- https://github.com/rust-lang/crates.io/pull/6168/

### DIFF
--- a/src/admin/enqueue_job.rs
+++ b/src/admin/enqueue_job.rs
@@ -23,6 +23,10 @@ pub enum Command {
         #[arg(long = "dry-run")]
         dry_run: bool,
     },
+    FixFeatures2 {
+        #[arg(long = "dry-run")]
+        dry_run: bool,
+    },
 }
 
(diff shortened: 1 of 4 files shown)


ehuss:
It looks like this would affect 2049 files in the index, which is an uncomfortably large number. It would be good to get some data on how long that will affect the index update, particularly with a cached copy that hasn't been updated in a while (like sometime between Jan 13 and Feb 1 to be after the normalization commit).

Turbo87:
https://github.com/rust-lang/crates.io/pull/9551 has fixed the original issue, but our background worker code has evolved enough to not be useful to try to rebase this PR. 

we'll keep running with the new feature splitting code for a while and may perform a batch update at a later point in time to reduce the number of git deltas needed for the update.

PR #6198, opened 2023-03-14
Update CONTRIBUTING.md to install diesel_cli 2.x

LawnGnome:
Diesel was upgraded to version 2 in #4892, but the contributing instructions still have the reader install diesel_cli 1.x, which will result in some fun errors when it attempts to regenerate `src/schema.rs` when running `diesel migration run`, as recommended later in the contributing guide.


Turbo87 (approved):
good catch, thanks :)

PR #6201, opened 2023-03-15
Update the import script to handle a couple of problematic scenarios

LawnGnome:
More detail is in the individual commits, but this addresses two issues I ran into while setting up a local crates.io environment:

* The import script hard codes the database name, instead of picking it up from the environment — this now checks the environment first before falling back to the hard coded name.
* The summary API endpoint 500s after running import because `recent_crate_downloads` isn't populated.


Turbo87 (approved):
thanks! 👍

PR #6204, opened 2023-03-16
Fix download charts after a DST transition

### LINKED ISSUE #5477: Downloads are being displayed on the wrong day following daylight saving
### Current Behavior

The following screenshot is from `quote`'s downloads (https://crates.io/crates/quote) but the same pattern is visible in the downloads graph of *all* widely used crates.

I am viewing the page in US Pacific Standard Time (UTC -8) but I don't know if that makes a difference.

Notice how there is a very obvious difference between weekday downloads (high) and weekend downloads (low) which is historically extremely well established for pretty much all widely used crates.

Notice anything strange? One of the weekends is missing a day. The missing day is November 6 which is the day that US clocks move back 1 hour for the end of daylight saving time. I highly doubt that this is a coincidence.

Prior to DST ending, all the points showing low are Saturdays and Sundays, like October 29 and October 30. After DST ending, all the points shown low are Fridays and Saturdays, like November 11 and November 12.

Based on my understanding of typical download trends and patterns on crates.io, I do not believe that this is a real phenomenon in the correct data. I believe the data is either being collected wrong or being plotted wrong.

Here is the corresponding download counts for the same crate from the crates.io database dump (https://crates.io/data-access):

```console
2022-10-29  107667 (weekend)
2022-10-30   93257 (weekend)
2022-10-31  250133
2022-11-01  264847
2022-11-02  282642
2022-11-03  285118
2022-11-04  258680
2022-11-05  118764 (weekend)
2022-11-06  109690 (weekend)
2022-11-07  281916
2022-11-08  318215
2022-11-09  295067
2022-11-10  289740
2022-11-11  237139
2022-11-12  114835 (weekend)
2022-11-13  108130 (weekend)
```

This data appears correct as far as I can tell. 2022-11-06 is low in the database dump but is incorrectly shown high in the graph, as is 2022-11-13. Meanwhile 2022-11-11 is high in the database dump but is rendered low on the graph.

### Expected Behavior

The data rendered in the graph should match the download counts in the database dump.

### Steps To Reproduce

Go to https://crates.io/crates/quote, maybe in a negative (West) UTC offset that recently ended daylight saving.

### Environment

_No response_

### Anything else?

_No response_

LawnGnome:
When falling back, a bug in date-fns[^0] results in the 23 hour day being skipped when building the `dates` array in `toChartData` if the current time is between 00:00 and 01:00 UTC. This results in that day's data essentially going missing, resulting in the subtly broken charts noted in #5477.

This commit adds a workaround adapted from a comment on date-fns/date-fns#571[^1] by @bertho-zero, which correctly adjusts the new date based on the time zone offsets, and means that `dates` is built as expected.

Fixes #5477.

### Broken

Note that November 6 is missing: there's only one "down" day that weekend:

image

### Fixed

Now note that there are two days:

image

### Or, if you like animations:

out

## Testing

I tested this by hacking the crate download backend endpoint to shift the current data to be between 2022-09-03 and 2022-12-01, no matter what the current date is, and the frontend JS to use 2022-12-01 as "now". (These dates were chosen to capture last year's "fall back" in `America/Vancouver`, which is my local time zone.)

The bug then became evident at 17:00 PST, which corresponds to 00:00 UTC.

Should you wish to reproduce this, here is the (genuinely awful) diff on top of this branch to do so:

```diff
diff --git a/app/components/download-graph.js b/app/components/download-graph.js
index 4c951f14a..893dcbe43 100644
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -97,7 +97,8 @@ export function toChartData(data) {
   let versions = new Map();
   let crate = null;
 
-  let now = new Date();
+  // let now = new Date();
+  let now = new Date('2022-12-01');
   for (let i = 0; i < 90; i++) {
     let date = subDays(now, i);
     dates[date.toISOString().slice(0, 10)] = { date, cnt: {} };
diff --git a/src/controllers/krate/downloads.rs b/src/controllers/krate/downloads.rs
index 9b22ed0ae..0f32ee2fc 100644
--- a/src/controllers/krate/downloads.rs
+++ b/src/controllers/krate/downloads.rs
@@ -3,8 +3,11 @@
 //! The endpoint for downloading a crate and exposing version specific
 //! download counts are located in `version::downloads`.
 
+use std::cell::RefCell;
 use std::cmp;
 
+use chrono::{Duration, NaiveDateTime};
+
 use crate::controllers::frontend_prelude::*;
 
 use crate::models::{Crate, CrateVersions, Version, VersionDownload};
@@ -26,14 +29,20 @@ pub async fn downloads(state: AppState, Path(crate_name): Path<String>) -> AppRe
             .sort_by_cached_key(|version| cmp::Reverse(semver::Version::parse(&version.num).ok()));
         let (latest_five, rest) = versions.split_at(cmp::min(5, versions.len()));
 
+        let start = NaiveDateTime::from_timestamp_opt(1669881600, 0).unwrap() - Duration::days(90);
+        let dater = FakeDater::new(&start);
         let downloads = VersionDownload::belonging_to(latest_five)
             .filter(version_downloads::date.gt(date(now - 90.days())))
             .order(version_downloads::date.asc())
             .load(conn)?
             .into_iter()
-            .map(VersionDownload::into)
+            .map(|mut vd: VersionDownload| -> EncodableVersionDownload {
+                vd.date = dater.next().date();
+                vd.into()
+            })
             .collect::<Vec<EncodableVersionDownload>>();
 
+        let dater = FakeDater::new(&start);
         let sum_downloads = sql::<BigInt>("SUM(version_downloads.downloads)");
         let extra: Vec<ExtraDownload> = VersionDownload::belonging_to(rest)
             .select((
@@ -43,7 +52,13 @@ pub async fn downloads(state: AppState, Path(crate_name): Path<String>) -> AppRe
             .filter(version_downloads::date.gt(date(now - 90.days())))
             .group_by(version_downloads::date)
             .order(version_downloads::date.asc())
-            .load(conn)?;
+            .load(conn)?
+            .into_iter()
+            .map(|mut xd: ExtraDownload| {
+                xd.date = dater.next().date().to_string();
+                xd
+            })
+            .collect();
 
         #[derive(Serialize, Queryable)]
         struct ExtraDownload {
@@ -60,3 +75,23 @@ pub async fn downloads(state: AppState, Path(crate_name): Path<String>) -> AppRe
     })
     .await
 }
+
+struct FakeDater {
+    start: NaiveDateTime,
+    days: RefCell<i64>,
+}
+
+impl FakeDater {
+    fn new(start: &NaiveDateTime) -> Self {
+        Self {
+            start: start.clone(),
+            days: RefCell::new(0),
+        }
+    }
+
+    fn next(&self) -> NaiveDateTime {
+        let mut days = self.days.borrow_mut();
+        *days += 1;
+        self.start.clone() + Duration::days(*days)
+    }
+}
```

[^0]: https://github.com/date-fns/date-fns/issues/571
[^1]: https://github.com/date-fns/date-fns/issues/571#issuecomment-602496322

### DIFF
--- a/app/components/download-graph.js
+++ b/app/components/download-graph.js
@@ -3,7 +3,7 @@ import { inject as service } from '@ember/service';
 import { waitForPromise } from '@ember/test-waiters';
 import Component from '@glimmer/component';
 
-import subDays from 'date-fns/subDays';
+import { addMinutes, subDays as brokenSubDays, subMinutes } from 'date-fns';
 import window from 'ember-window-mock';
 import semverSort from 'semver/functions/sort';
 
@@ -189,4 +189,19 @@ export function toChartData(data) {
 
 function midnightForDate(date) {
   return Date.parse(date.toISOString().slice(0, 10));
+}
+
+// This works around a bug in date-fn's subDays() function when crossing a DST
+// transition: https://github.com/date-fns/date-fns/issues/571
+//
+// The specific implementation is based on @bertho-zero's here:
+// https://github.com/date-fns/date-fns/issues/571#issuecomment-602496322
+function subDays(date, amount) {
+  const originalTZO = date.getTimezoneOffset();
+  const endDate = brokenSubDays(date, amount);
+  const endTZO = endDate.getTimezoneOffset();
+
+  const dstDiff = originalTZO - endTZO;
+
+  return dstDiff >= 0 ? addMinutes(endDate, dstDiff) : subMinutes(endDate, Math.abs(dstDiff));
 }


Turbo87 (commented):
wow, nice find! :D

do you know if it's possible to write a regression test for this?

LawnGnome:
> do you know if it's possible to write a regression test for this?

I played around with this a bit, but I don't see an obvious way to programmatically reproduce the bug inside the Ember test runner. The problem here is that the JS runtime has to be running in a time zone that (a) has DST transitions (so not UTC), and (b) has a backward transition at a known date. While we could make a reasonable effort to find a backward transition in the current time zone and then set up a test from there, the fact that it wouldn't work in all time zones feels like it might make it too likely to be skipped to be useful.

I also don't think there's a reliable way to specify which time zone is in use — I guess we could change `package.json` to run the `test` command with `TZ=America/Vancouver` or something, but that still feels pretty brittle.

I can put some more time into this if you want, but I don't see an easy win in terms of automated testing.

Turbo87:
sounds reasonable, thanks for checking!

PR #6316, opened 2023-04-12
PUT /me/token: Remove unused `revoked` field from response payload

Turbo87:
The `revoked` field is not used by the frontend, and the only API endpoint that returns the `revoked` field is the one that creates a brand new token, in which case it makes no sense for the token to be immediately revoked.

This is technically a breaking change, but since this endpoint is only meant to be used by the frontend when creating new tokens via the user interface the risk of breaking something should be fairly small.

### DIFF
--- a/src/models/token.rs
+++ b/src/models/token.rs
@@ -143,7 +143,6 @@ mod tests {
             id: 12345,
             name: "".to_string(),
             token: "".to_string(),
-            revoked: false,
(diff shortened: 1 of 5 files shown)


Turbo87:
The cargo team confirmed [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/unused.20API.20field.20removal/near/348761994) that they are okay with this change, so I'll go ahead and merge this :)

PR #6317, opened 2023-04-12
Sort crate dependencies before adding to index

arlosi:
We [sorted the crate dependencies](https://github.com/rust-lang/crates.io/pull/5089) when we did the normalization, but newly published crates do not get their dependencies sorted.

* Updates the `Ord` impl on `Dependency` so all fields are included to ensure a total order.
* Adds a test that verifies newly published crates get sorted (verified this failed before the fix).

r? @Turbo87 

### Related:

- https://github.com/rust-lang/crates.io/pull/5089

### DIFF
--- a/cargo-registry-index/lib.rs
+++ b/cargo-registry-index/lib.rs
@@ -175,7 +175,26 @@ impl Ord for Dependency {
         // misinterpreted. With this manual `Ord` implementation we ensure that
         // `normal` dependencies are always first when multiple with the same
         // `name` exist.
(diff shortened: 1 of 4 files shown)


Turbo87:
I'll take a close look tomorrow. Do you think it would be possible to write a regression test for this?

arlosi:
@Turbo87 Thanks. Test added.

Turbo87 (approved):
looks good, thank you!

I guess that means we might have to run another normalization pass for the crates that have been published since the last time 🙈

PR #6318, opened 2023-04-12
Make it easier to debug test failures due to incorrect http-data

arlosi:
When tests fail due to a difference in the body of an HTTP request, it shows up as an escaped string vs a Vec of bytes.

This makes the assert compare the base64 encoded strings (which is what's actually in the JSON http data file).

r? @Turbo87

### DIFF
--- a/src/tests/record.rs
+++ b/src/tests/record.rs
@@ -363,10 +363,8 @@ fn replay_http(
 
     async {
         let _ = &exchange;
-        assert_eq!(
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):
that's quite literally exactly what I had on my `dbg/base64-recorder` branch that I mentioned yesterday 😅 

I do have an idea how we can make this even better, but I'll merge this in the meantime 👍

Issue #6414, opened 2023-05-02
Show README and metadata for yanked crates/versions


Turbo87:
https://github.com/rust-lang/crates.io/issues/1058 proposed to reduce information when a crate has all of its versions yanked, but as https://github.com/rust-lang/crates.io/issues/1058#issuecomment-1531276138 mentions we decided to do something different:

Instead of showing a note that a crate/version is yanked, we should instead show the existing metadata that the backend returns but with a very visible banner (or similar) that this crate/version was yanked by one of the crate owners.

This helps to e.g. figure out who the current owner of a crate is, even when all of the versions are yanked. We currently only display a generic note, without any crate metadata.

PR #6415, opened 2023-05-02
Show README and metadata for yanked crates/versions

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/6414

### Before

Bildschirmfoto 2023-05-02 um 13 30 16

### After

Bildschirmfoto 2023-05-02 um 13 29 45

I'm opening this as a draft PR for discussion among the team. Also, tests are still missing for this change.

### DIFF
--- a/app/components/crate-header.hbs
+++ b/app/components/crate-header.hbs
@@ -8,6 +8,11 @@
         <span local-class="yanked-badge">
           {{svg-jar "trash"}}
           Yanked
+
+          <EmberTooltip>
+            This crate has been yanked, but it is still available for download for other crates that
+            may be depending on it.
+          </EmberTooltip>
         </span>
       {{/if}}
     {{/if}}
--- a/app/components/crate-header.hbs
+++ b/app/components/crate-header.hbs
@@ -5,7 +5,7 @@
       <small data-test-crate-version>v{{@version.num}}</small>
 
       {{#if @version.yanked}}
-        <span local-class="yanked-badge">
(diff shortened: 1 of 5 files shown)


app/components/crate-header.hbs
```diff
@@ -3,6 +3,13 @@
     <span data-test-crate-name>{{@crate.name}}</span>
     {{#if @version}}
       <small data-test-crate-version>v{{@version.num}}</small>
+
+      {{#if @version.yanked}}
+        <span local-class="yanked-badge">
+          {{svg-jar "trash"}}
+          Yanked
+        </span>
```
LawnGnome, line 11: Do we want to link to something — or at least have a tooltip — which explains what yanking means in the same way the previous version said "This crate has been yanked, but it is still available for download for other crates that may be depending on it"?

LawnGnome (approved):
I think this is a good change, provided we take care not to confuse anyone as to what "yanked" actually means. :+1:

app/components/crate-header.hbs
Turbo87: that sounds like a good idea, thanks! 👍

Turbo87: Bildschirmfoto 2023-05-03 um 16 27 14

Turbo87:
I've modified the existing tests for this route to cover the new behavior.

This PR is ready for review/merging now. Unless there are any objections I plan on merging this some time tomorrow.

LawnGnome (approved):

PR #6422, opened 2023-05-03
Fix typo in token.rs

ianstanton:
Fix typo in `token.rs`.


Turbo87 (approved):
good catch, thank you! :D

PR #6423, opened 2023-05-03
README: Add styling for `<kbd>` tags

ToBinio:
## Before
img

## After
image

This is a small PR adding some style for the \<kbd> tag in the the markdown display.
It was created after a short discussion #6420.


Turbo87 (approved):
looks good, thank you! :)

PR #6426, opened 2023-05-04
Footer: Remove Discord link

Turbo87:
We decided to use Zulip instead of Discord a while ago, so we should remove the Discord link at this point.

see https://github.com/rust-lang/team/pull/893


jtgeibel (approved):

jtgeibel:
We should definitely remove this link. Should we add a Zulip link instead?

PR #6433, opened 2023-05-05
Versions: Add color explanations on hover

### LINKED ISSUE #6238: Yellow text for recent publish date may be confusing
### Current Behavior

Currently, when a crate version was published recently, the text showing how long ago it was published is yellow.

Screenshot 2023-03-24 at 20 18 23

When the latest version was published less recently - even if it is the latest version - the text is the standard grey.

Screenshot 2023-03-24 at 20 08 33

I discovered this recently and thought that the yellow indicated some kind of warning (maybe the version hadn't been fully propogated to all mirrors, ... I really had no idea).

Hovering over the text doesn't show any additional information.

Screenshot 2023-03-24 at 20 08 22

I later found out that this was likely just an indicator of newness. However, for me it was a little confusing.

### Expected Behavior

It would be great if there could be some indicator why the date has a different color.

This is already done for the crate versions themselves. For example, the latest crate version is shown in green and the hover text has information in brackets which explains why.

Screenshot 2023-03-24 at 20 09 43

A pre-release version is shown in yellow and the hover text also explains why.

Screenshot 2023-03-24 at 20 09 49

I think that it would be great if the hover text for a recent date included that information. Here is an example mock up:

Screenshot 2023-03-24 at 20 14 40

Additionally, the text explaining why the text is yellow could also be in yellow, to further link the information for the user.

Screenshot 2023-03-24 at 20 14 17

Alternatively (or additionally), the recent dates could be shown in green, indicating fresh and tying in with the green for the latest release. Or perhaps the intention was to "warn" the user that this crate version is very new and therefore less tested, in which case the yellow makes sense.

### Steps To Reproduce

1. Go to crates.io main page
2. Select a crate from the Just Updated list
3. Select the Versions tab for that crate
4. Check the Published date

### Environment

On all paltforms.

### Anything else?

This is a pretty minor thing. If the team disagree with my sentiments (a few people I spoke to agreed with me, but that's hardly a consensus), then closing this issue would be totally acceptable.

ToBinio:
This PR should resolve #6238 

though implementing this suggested solution.
Screenshot 2023-05-05 153337

it also adds similar color coding to the version (Release Track) text.
Screenshot 2023-05-05 154213

### Note
I am not completely happy with the implementation of the color coding of the version text.

I wanted to use the current modifier system to prevent a gigantic mess in the HTML
the problem is that it seems like Ember does not resolve the `local-class` when using HTML injection.

so I was forced to directly style the `mark` tag


ToBinio:
I am not completely happy with the way you select which color `bracket-mark` has. But I couldn't think of any better way without something like typescript support.

Turbo87:
@ToBinio I've rebased and simplified the implementation a little bit. Since we are using CSS modules we can import the CSS module in the JS code to gain access to the class name mapping and use that (plus `htmlSafe()`) to add `<span>` tags to the tooltips.

PR #6437, opened 2023-05-05
Migrate dev server reverse proxying to custom middleware

LawnGnome:
This is required for further admin console work: the proxy middleware installed by `ember-cli` when using the `--proxy` flag proxies based on the presence of an `Accept` header containing `text/html`, which means that it's impossible to return HTML to a normal (non-XHR) browser request from a proxied backend.[^accept]

Instead, we'll now define the specific routes that will be proxied in a custom middleware installed in `server/index.js`, which better matches how this is all deployed in practice anyway. (Well, as long as it matches the nginx configuration.)

Proxying to the backend is now controlled by the optional `PROXY_BACKEND` environment variable: if present, proxying will take place; if absent, static fixtures will be loaded through `ember-cli-mirage` as before. This matches the behaviour of `--proxy` from the point of view of a crates.io contributor, even if the actual mechanism is now different.

[^accept]: https://github.com/ember-cli/ember-cli/blob/b9bec5472ed885cef8cb2994284a7c37652e9af0/lib/tasks/server/middleware/history-support/index.js#L82


Turbo87 (approved):

Turbo87:
I've rebased this to avoid the `package.json` conflict and took the liberty of converting `installBackendProxy()` to a proper `function` 😉 

LGTM though and appears to work as intended 👍

PR #6442, opened 2023-05-07
Front Page: adding "see all" to every list

ToBinio:
this PR resolves #5403

It adds a `see all` button to every list.

image

Maybe these buttons should be made smaller or removed completely and instead make the title clickable. simply because i think there now is a bit to much green (see all)


Turbo87:
> and instead make the title clickable

yeah, that sounds like a good idea to me 👍 

I'll still merge this since it seems like a good incremental improvement.

PR #6445, opened 2023-05-08
Add `rust-toolchain.toml` file

Turbo87:
This file ensures that CI and all developers, by default, use the same Rust version. This helps reduce differences between e.g. `clippy` or `rustfmt` versions.

If people prefer to use other versions (e.g. beta or nightly) they can still use `rustup override` on the crates.io project, but having a `rust-toolchain.toml` with the right version should make things a little easier for new contributors.


LawnGnome (approved):
The Renovate side of things is definitely above my knowledge level, but the toolchain file definitely works fine. :+1:

PR #6452, opened 2023-05-08
Remove confusing comment fragments about mirror mode

wangkirin:
Since crate.io has already deprecated the `mirror` mode and remove related code, these two comment fragments may cause confusion.
This PR fix it


Turbo87 (approved):
good catch, thank you! :)

PR #6456, opened 2023-05-09
Define admin users and extend `AuthCheck` to handle them

LawnGnome:
This adds a concept of admin users, who are defined by their GitHub IDs, and allows them to be defined through an environment variable, falling back to a static list of the current `crates.io` team.

`AuthCheck` now has a builder method to require that the current cookie or token belong to an admin user.

In the future, this will be extended to use Rust's team API for the fallback.

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -57,6 +57,7 @@ pub struct Server {
     pub version_id_cache_ttl: Duration,
     pub cdn_user_agent: String,
     pub balance_capacity: BalanceCapacityConfig,
+    pub admin_user_github_ids: HashSet<i32>,
 
     /// Instructs the `cargo_compat` middleware whether to adjust response
     /// status codes to `200 OK` for all endpoints that are relevant for cargo.
@@ -104,6 +105,8 @@ impl Server {
     ///   endpoint even with a healthy database pool.
     /// - `BLOCKED_ROUTES`: A comma separated list of HTTP route patterns that are manually blocked
     ///   by an operator (e.g. `/crates/:crate_id/:version/download`).
+    /// - `ADMIN_USER_GH_IDS`: A comma separated list of GitHub user IDs that will be considered
+    ///   admins.
     ///
     /// # Panics
     ///
@@ -206,6 +209,10 @@ impl Server {
             balance_capacity: BalanceCapacityConfig::from_environment()?,
             cargo_compat_status_code_config: var_parsed("CARGO_COMPAT_STATUS_CODES")?
                 .unwrap_or(StatusCodeConfig::AdjustAll),
+            admin_user_github_ids: HashSet::from_iter(list_parsed(
+                "ADMIN_USER_GH_IDS",
+                i32::from_str,
+            )?),
             serve_dist: true,
             serve_html: true,
             content_security_policy: Some(content_security_policy.parse()?),
--- a/.env.sample
+++ b/.env.sample
@@ -86,3 +86,6 @@ export SENTRY_ENV_API=local
 # export TEST_S3_INDEX_REGION=http://127.0.0.1:19000
 # export TEST_AWS_ACCESS_KEY=minio
 # export TEST_AWS_SECRET_KEY=miniominio
+
+# IDs of GitHub users that are admins on this instance, separated by commas.
+export ADMIN_USER_GH_IDS=
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -62,7 +62,10 @@ impl AuthCheck {
         conn: &mut PgConnection,
     ) -> AppResult<Authentication> {
         let auth = authenticate(request, conn)?;
+        self.check_authentication(auth)
+    }
 
+    pub fn check_authentication(&self, auth: Authentication) -> AppResult<Authentication> {
         if let Some(token) = auth.api_token() {
             if !self.allow_token {
                 let error_message =
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -1,5 +1,8 @@
+use std::collections::HashSet;
+
 use crate::controllers;
 use crate::controllers::util::RequestPartsExt;
+use crate::middleware::app::RequestApp;
 use crate::middleware::log_request::RequestLogExt;
 use crate::middleware::session::RequestSession;
 use crate::models::token::{CrateScope, EndpointScope};
@@ -16,6 +19,7 @@ pub struct AuthCheck {
     allow_token: bool,
     endpoint_scope: Option<EndpointScope>,
     crate_name: Option<String>,
+    require_admin: bool,
 }
 
 impl AuthCheck {
@@ -27,6 +31,7 @@ impl AuthCheck {
             allow_token: true,
             endpoint_scope: None,
             crate_name: None,
+            require_admin: false,
         }
     }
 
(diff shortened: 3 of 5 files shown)


src/models/helpers/admin.rs
```diff
@@ -0,0 +1,91 @@
+use std::{collections::HashSet, num::ParseIntError, str::FromStr};
+
+use lazy_static::lazy_static;
```
Turbo87, line 3: we might want to use `once_cell` instead :)

src/models/helpers/admin.rs
```diff
@@ -0,0 +1,91 @@
...
+
+lazy_static! {
+    static ref AUTHORIZED_ADMIN_USER_IDS: HashSet<i32> =
+        parse_authorized_admin_users(dotenv::var("GH_ADMIN_USER_IDS"));
+}
+
+// The defaults correspond to the current crates.io team.
+//
+// FIXME: this needs to be removed once we can detect the admins from the Rust teams API.
+const DEFAULT_ADMIN_USER_IDS: [i32; 5] = [193874, 22186, 141300, 229984, 20070360];
```
Turbo87, line 15: might be best to add the GitHub usernames in comments next to those magic numbers

Turbo87 (commented):
just a first glance review... I'll take a closer look tomorrow 😉

src/models/helpers/admin.rs
LawnGnome: Done, thanks!

src/models/helpers/admin.rs
LawnGnome: Yep, totally forgot to do that before pushing. Thanks!

src/models/helpers/admin.rs
```diff
@@ -0,0 +1,90 @@
...
+    match maybe_users {
+        Ok(users) => users
+            .split(|c: char| !(c.is_ascii_digit()))
+            .filter(|uid| !uid.is_empty())
+            .map(i32::from_str)
+            .collect::<Result<HashSet<i32>, ParseIntError>>()
+            .expect("parsing admin users from $GH_ADMIN_USER_IDS"),
+        Err(_err) => DEFAULT_ADMIN_USER_IDS.into_iter().collect(),
+    }
+}
```
Turbo87, line 26: now that I think about it some more, we might want to put this into the `config.rs` file, next to all the other env var server config code. the `config::Server` struct already has a dedicated implementation for our test suite, so the `#[cfg(test)]` below on the `check_user_id()` fn wouldn't be necessary anymore.

one downside of this is that we need access to that config struct when checking if someone is an admin user, but in `AuthCheck::check()` we have access to the `Request` struct, which iirc should give use access to the state extension and thus access to the config struct too.

src/models/user.rs
```diff
@@ -177,4 +179,88 @@ impl User {
...
     }
+
+    /// Attempt to turn this user into an AdminUser
+    pub fn admin(&self) -> AppResult<AdminUser> {
+        AdminUser::new(self)
+    }
+}
+
+#[derive(Debug)]
+pub struct AdminUser(User);
```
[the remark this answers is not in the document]
LawnGnome: 100% stole this from #5376. I'm happy to change it to just be a boolean check instead.

src/models/user.rs
```diff
@@ -6,11 +6,13 @@ use crate::app::App;
...
 
-use crate::models::{ApiToken, Crate, CrateOwner, Email, NewEmail, Owner, OwnerKind, Rights};
+use crate::models::{
+    helpers::admin, ApiToken, Crate, CrateOwner, Email, NewEmail, Owner, OwnerKind, Rights,
+};
 use crate::schema::{crate_owners, emails, users};
 
 /// The model representing a row in the `users` database table.
-#[derive(Clone, Debug, PartialEq, Eq, Queryable, Identifiable, AsChangeset)]
+#[derive(Clone, Debug, PartialEq, Eq, Queryable, Identifiable, AsChangeset, Selectable)]
```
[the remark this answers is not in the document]
LawnGnome: Remnant of an earlier version of this branch, sorry. Removed.

src/models/helpers/admin.rs
LawnGnome: Will investigate, either today if I have some free time at the conference or next week when I get back from my long weekend Ucluelet.

LawnGnome:
Updated per review. Definitely simpler this way, so thanks!

I don't _love_ the drilling of `gh_admin_user_ids` into `AuthCheck`, but here we are. We could sugar our way in with a helper on `App` just to clean up the controllers or a tuple struct wrapping the `HashSet` that isn't as wordy, if that's preferred, but I think this is just about OK.

src/controllers/crate_owner_invitation.rs
```diff
@@ -19,7 +19,7 @@ use std::collections::{HashMap, HashSet};
 pub async fn list(app: AppState, req: Parts) -> AppResult<Json<Value>> {
     conduit_compat(move || {
         let conn = &mut app.db_read()?;
-        let auth = AuthCheck::only_cookie().check(&req, conn)?;
+        let auth = AuthCheck::only_cookie().check(&req, &app.config.gh_admin_user_ids, conn)?;
```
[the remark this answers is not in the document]
LawnGnome: It wasn't, but this is a good lesson in remembering to actually write up one's problems: I couldn't figure out the magic incantation to access shared state given a `RequestPartsExt` (or related traits) last week, but then by the time I pushed this I think I was just so annoyed with the whole thing that I forgot to write it down.

This week, I found `RequestApp`.

Anyway, I'm an idiot, and the diff is now much smaller. Thank you for your patience. :see_no_evil:

src/config/server.rs
```diff
@@ -12,11 +12,19 @@ use crate::storage::StorageConfig;
...
 
 const DEFAULT_VERSION_ID_CACHE_SIZE: u64 = 10_000;
 const DEFAULT_VERSION_ID_CACHE_TTL: u64 = 5 * 60; // 5 minutes
 
+// The defaults correspond to the current crates.io team, which as at the time of writing, is the
+// GitHub user names carols10cents, jtgeibel, Turbo87, JohnTitor, LawnGnome, and mdtro.
+//
+// FIXME: this needs to be removed once we can detect the admins from the Rust teams API.
+const DEFAULT_GH_ADMIN_USER_IDS: [i32; 6] = [193874, 22186, 141300, 25030997, 229984, 20070360];
+
```
Turbo87, line 27: I'm wondering if we should really hardcode these values or just default to an empty list instead. I think I tend towards `unwrap_or_default()`.

src/config/server.rs
```diff
@@ -326,3 +342,28 @@ fn parse_ipv6_based_cidr_blocks() {
...
     );
 }
+
+fn parse_gh_admin_user_ids(users: String) -> Result<HashSet<i32>, ParseIntError> {
+    users
+        .split(|c: char| !(c.is_ascii_digit()))
+        .filter(|uid| !uid.is_empty())
+        .map(i32::from_str)
+        .collect()
+}
```
Turbo87, line 352: the docs above say "A comma separated list". why do we allow other separators? and do we need the `filter()` step?

Turbo87 (approved):

LawnGnome:
Closing in favour of #7852.

PR #6464, opened 2023-05-10
Integrate `cargo-deny`

Turbo87:
see https://github.com/EmbarkStudios/cargo-deny


Turbo87:
Note that the "Backend / cargo-deny" CI job has been skipped here since the `Cargo.lock` file was not changed :)

PR #6470, opened 2023-05-11
Replace `abbr` with `EmberTooltips`

ToBinio:
This PR removes some `abbr` tags and replaces them with `EmberTooltips` this is mainly for visual consistency and they should be functionally identical.
The same change was made for copy-button.

Before:
Screenshot 2023-05-11 202407

After:
Screenshot 2023-05-11 202301

### DIFF
--- a/app/components/crate-row.hbs
+++ b/app/components/crate-row.hbs
@@ -25,18 +25,34 @@
   <div local-class='stats'>
     <div local-class='downloads' data-test-downloads>
       {{svg-jar "download" local-class="download-icon"}}
-      <span><abbr title="Total number of downloads">All-Time:</abbr> {{ format-num @crate.downloads }}</span>
+      <span>
+        <span>
+          All-Time:
+          <EmberTooltip @text="Total number of downloads"/>
+        </span>
+        {{ format-num @crate.downloads }}
+      </span>
     </div>
     <div local-class="recent-downloads" data-test-recent-downloads>
       {{svg-jar "download" local-class="download-icon"}}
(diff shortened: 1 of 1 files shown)


app/components/copy-button.hbs
```diff
@@ -1,3 +1,4 @@
 <button type="button" ...attributes {{on "click" this.copy}}>
   {{yield}}
+  <EmberTooltip @text={{@hoverText}}/>
```
Turbo87, line 3: I don't think the tooltip belongs in the `CopyButton` component itself. this component should only provide the "click-to-copy" functionality, but not any styling. if a button wants to use a tooltip then the `{{yield}}` should make it quite easy to add without mutating the `CopyButton` component itself.

Turbo87 (commented):
> The same change was made for copy-button.

I think this is a good change for the `<abbr>` elements, but since the `CopyButton` changes are mixed into the `<abbr>` changes it unfortunately makes it hard to merge them individually. It might be better to open dedicated PRs for them, or at least package them in dedicated commits :)

ToBinio:
sounds reasonable.

btw [Turbo87](https://github.com/Turbo87) thanks for being so helpful and making it so easy to contribute

should I add another commit that reverts the copy-button changes or should I open a new PR?

Turbo87:
> should I add another commit that reverts the copy-button changes or should I open a new PR?

I personally would use `git rebase --interactive` to clean up the branch history, but you can also push another cleanup commit on top. In that case I would most likely squash-merge the PR to keep the git history somewhat clean :)

ToBinio:
I believe I got the rebase to work....

Turbo87 (approved):

PR #6473, opened 2023-05-12
index/repo: Use `git` CLI for fetching and pushing

Turbo87:
Using the `git` CLI seems significantly less complex compared to using libgit2, and, as we discovered in the past, it can also have performance advantages in some situations.


JohnTitor (approved):
LGTM for simplicity

JohnTitor:
> it can also have performance advantages in some situations.
Just curious but do we have any stats/metrics for that?

PR #6510, opened 2023-05-22
embroider: Add resolution for `@embroider/core|compat`

Turbo87:
This allows us to use `ember-cli-code-coverage` with a more recent version of embroider

Related: 
- https://github.com/kategengler/ember-cli-code-coverage/pull/382


Turbo87:
well... looks like that doesn't work 😩

Issue #1059, opened 2017-09-15
Should publishes with versions only differing in metadata be allowed?


sfackler:
For example, https://crates.io/crates/openssl-src has `110.0.0` and `110.0.0+1.1.0f`, which are semver-equivalent since build metadata doesn't affect the actual version. It seems like this probably shouldn't be allowed?

mbrubeck:
Some previous discussion in https://github.com/steveklabnik/semver/issues/107

carols10cents:
Sure sounds like we should be disallowing versions like these that only differ in build metadata....

jeikabu:
Looks like I fell victim to this.

I'd been using semver metadata only to [discover it doesn't work as expected](https://crates.io/crates/runng-sys/versions).  Rather than getting rid of metadata support it would make sense to let them be semver equivalent.  Whether some attempt is made to parse and make sense of the metadata is a convention that could be developed separately.  Until then taking the most recent would be the "least surprising".

To me the primary use case is crates that have some external, non-crate dependency and you want to mirror it's semver version.  "sys" crates fall in this category.  I have a crate that corresponds to NNG `1.1.1`.  Being able to version my crate as `1.1.1` instead of some unrelated version scheme that requires mapping between the two is the simplest.  If I need to alter crate contents in some way (e.g. packaging or documentation- some non-API-breaking change) `1.1.1+SOMETHING` would make the most sense.

As it is, if you have a crate versioned `0.1.0+XYZ` and you then publish/yank `0.1.0` [the crate is "yanked"](https://crates.io/crates/cargo_crates-io_docs-rs_test) even when there's a [semver-equivalent crate available](https://crates.io/crates/cargo_crates-io_docs-rs_test/versions), `0.1.0+3`.

sgrif:
I really do think this just boils down to a bug in semver. Build metadata is ignored for precedence, but that does not mean it's ignored for equality. To me this implies that if two versions differ only in metadata, we may select one at random to treat as the "latest version", but they are still *different* versions, and should be treated as such both by crates.io and Cargo. I laid out some further reasoning in https://github.com/steveklabnik/semver/issues/107#issuecomment-521760730

pietroalbini:
The ordering being the same is not going to be changed as far as I can see, and that makes any possible use of releases with the same version but different metadata practically impossible.

[...]

PR #6518, opened 2023-05-23
Prevent similar versions that only differ in build metadata from being published

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/1059

`cargo` does not handle these situations particularly well and the SemVer specification is a little ambiguous about how to handle build metadata.

We still have roughly 600 problematic versions in the database, so we can't introduce a unique index yet, but the query appears to be fast enough according to my local testing:

```sql
EXPLAIN SELECT * FROM versions WHERE num = '1.0.0' AND crate_id = 463;
```
```
Index Scan using unique_num on versions  (cost=0.42..8.45 rows=1 width=190)
  Index Cond: ((crate_id = 463) AND ((num)::text = '1.0.0'::text))
```

vs.

```sql
EXPLAIN SELECT * FROM versions WHERE split_part(num, '+', 1) = '1.0.0' AND crate_id = 463;
```
```
Bitmap Heap Scan on versions  (cost=4.57..79.36 rows=1 width=190)
  Recheck Cond: (crate_id = 463)
"  Filter: (split_part((num)::text, '+'::text, 1) = '1.0.0'::text)"
  ->  Bitmap Index Scan on unique_num  (cost=0.00..4.57 rows=19 width=0)
        Index Cond: (crate_id = 463)
```

The `cost` is certainly higher, but the actual wall time for these queries appears to be roughly similar in the end. Since the publish endpoint is only receiving few requests per minute this should not result in any issues AFAICT.

### DIFF
--- a/src/models/version.rs
+++ b/src/models/version.rs
@@ -7,6 +7,7 @@ use crate::util::errors::{cargo_err, AppResult};
 
 use crate::models::{Crate, Dependency, User};
 use crate::schema::*;
+use crate::sql::split_part;
 
 // Queryable has a custom implementation below
 #[derive(Clone, Identifiable, Associations, Debug, Queryable, Deserialize, Serialize)]
@@ -168,14 +169,16 @@ impl NewVersion {
         use diesel::{insert_into, select};
 
         conn.transaction(|conn| {
+            let num_no_build = strip_build_metadata(&self.num);
+
             let already_uploaded = versions
                 .filter(crate_id.eq(self.crate_id))
-                .filter(num.eq(&self.num));
+                .filter(split_part(num, "+", 1).eq(num_no_build));
+
             if select(exists(already_uploaded)).get_result(conn)? {
                 return Err(cargo_err(&format_args!(
-                    "crate version `{}` is already \
-                     uploaded",
-                    self.num
+                    "crate version `{}` is already uploaded",
+                    num_no_build
                 )));
             }
 
@@ -217,6 +220,13 @@ fn validate_license_expr(s: &str) -> AppResult<()> {
     })?;
 
     Ok(())
+}
+
+fn strip_build_metadata(version: &str) -> &str {
+    version
+        .split_once('+')
+        .map(|parts| parts.0)
+        .unwrap_or(version)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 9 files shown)


LawnGnome (approved):
The implementation looks good. :+1:

For what it's worth, playing around with this locally seemed to throw up some bigger deltas between the old query and the new one. With a cold cache, the old query executed in 12 ms for me, whereas it took 48 ms for the new query to execute. (I tested this with both the crate ID you used, and the one that has the most actual versions; the delta was similar in both cases.)

I can get the difference back below the noise floor by adding an index on the computed field, which I've pushed to 757dce55b4302cd0be83663baceb82ebfd68f5e1. My guess is that you're right that this PR isn't likely to cause problems anyway, but you could cherry pick that in if you want to add the index for more assurance.

Turbo87:
> For what it's worth, playing around with this locally seemed to throw up some bigger deltas between the old query and the new one. With a cold cache, the old query executed in 12 ms for me, whereas it took 48 ms for the new query to execute.

yep, I remember getting similar numbers here, though for the publish endpoint I guess this would probably be acceptable?

> I can get the difference back below the noise floor by adding an index on the computed field, which I've pushed to [757dce5](https://github.com/rust-lang/crates.io/commit/757dce55b4302cd0be83663baceb82ebfd68f5e1).

As I mentioned in https://github.com/rust-lang/crates.io/commit/757dce55b4302cd0be83663baceb82ebfd68f5e1#r114738341, the performance actually got worse for me locally when I added this index. If I add an index on the `crate_id` column and the computed `split_part()` result I got significantly better results.

Also, I am a bit scared about adding indices in migrations. Locally, adding the index took about three seconds, but I've been bitten by this in the past already, where in production it took a lot longer. Since migrations are currently running after shutdown of the old app version and startup of the new app version it would effectively cause multi-second downtime for us if we add the index in a migration. We can use `CREATE INDEX CONCURRENTLY` to avoid the table lock when creating the index, but that unfortunately also does not solve the migration issue. I'm happy to brainstorm on this issue separately :)

Turbo87:
Assuming that this was discussed in the team meeting last week (where I unfortunately couldn't attend) and since there are apparently no objections I will merge and deploy this change now. Since it does not involve any database migrations we can always revert this if we decide that this is the wrong step forward.

PR #6522, opened 2023-05-25
bump base64 to 0.21.1

mdtro:
Closes #6503.

Renovate's update failed. I created a new branch and updated the deprecated function calls to their new equivalents.

Let me know if you'd rather I close this and commit to Renovate's branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -650,7 +650,7 @@ dependencies = [
  "aws-sigv4",
  "axum",
  "axum-extra",
- "base64 0.13.1",
+ "base64 0.21.2",
  "chrono",
  "claims",
  "clap",
@@ -719,7 +719,7 @@ name = "crates_io_index"
 version = "0.0.0"
 dependencies = [
  "anyhow",
- "base64 0.13.1",
+ "base64 0.21.2",
  "claims",
  "dotenvy",
  "git2",
@@ -745,7 +745,7 @@ dependencies = [
 name = "crates_io_s3"
 version = "0.0.0"
 dependencies = [
- "base64 0.13.1",
+ "base64 0.21.2",
  "chrono",
  "hmac",
  "reqwest",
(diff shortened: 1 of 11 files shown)


Turbo87:
I'm still a bit worried about https://github.com/marshallpierce/rust-base64/blob/HEAD/RELEASE-NOTES.md#padding since last time we tried the update it was causing https://github.com/rust-lang/crates.io/issues/5646 🙈

cargo-registry-index/repo.rs
```diff
@@ -30,10 +31,28 @@ impl RepositoryConfig {
...
+                                    Ok(decoded_key) => decoded_key,
+                                    Err(no_pad_error) => panic!("failed to base64 decode the ssh key STANDARD_NO_PAD engine: {no_pad_error}"),
+                                }
+                        }
+                        _ => panic!(
+                            "failed to base64 decode the ssh key using STANDARD engine: {error}"
+                        ),
+                    },
+                };
+
```
mdtro, line 52: It's not pretty, but let me know if this would be sufficient. The `InvalidPadding` error in https://github.com/rust-lang/crates.io/issues/5646 will be handled here.

1. First attempt to decode it with padding expected (`general_purpose::STANDARD`)
2. If we end up with an `InvalidPadding` error...
3. Attempt to decode it with _no padding_ expected (`general_purpose::STANDARD_NO_PAD`)

There's probably a cleaner way to write these nested matches too. I'm open to feedback. :)

Turbo87: sorry, I totally missed the update on this 🙈 

I thought some more about this and figured since we control the `GIT_SSH_KEY` value ourselves it would probably be easiest to use correct padding for its value on Heroku 😅 

I've rebased your branch on top of #6503 once more and dropped the double-decode workaround again. sorry for the extra work! 🙈

PR #6524, opened 2023-05-25
db/background_jobs: Add `priority` column

Turbo87:
This PR adds a `priority` (`smallint` aka. `i16`) column to the `background_jobs` table. Which allows us to prioritize certain tasks over others.

The PR also adjusts the priorities of the sync-to-index and render-readme tasks, with the former being the most important task, and the latter being medium-high priority. This leads to a small change in how crates are processed when they are published. While previously the README was rendered "before" the index was synced, the order is now reversed, with the index sync having priority. (The quotes are intentional since multiple background jobs can effectively run in parallel)

If also adjusted our `monitor` tool to not trigger for stalled jobs with low (= below zero) priority. This should allow us to queue up e.g. one background job per existing version to backfill certain database fields (e.g. `rust_version`) without paging the on-call team.


LawnGnome (approved):
:+1:

PR #6526, opened 2023-05-25
Dashboard: remove black background from following.svg

ToBinio:
This PR removes the black background from `following.svg`.

Or am I missing some reason why this is the case?

Before:
image

After:
image


LawnGnome (approved):
(Literally) LGTM. I can't think of a reason it needs to have a dark background. :shrug:

Turbo87 (approved):

PR #6545, opened 2023-05-29
Expand error messages for invalid license information

ehuss:
This expands the error message when the `license` expression is not valid. It provides some help text if the user needs a custom license, and also points to the cargo documentation. It also includes the actual error message from the validator, which can be helpful if you have a more complex expression, and it may not be clear exactly which part of the expression is invalid.

This also clarifies the wording when `license` is not specified.

The error may look something like this:

```
error: failed to publish to registry at http://localhost:8888/

Caused by:
  the remote server responded with an error: unknown or invalid license expression; see http://opensource.org/licenses for options, and http://spdx.org/licenses/ for their identifiers
  Note: If you have a non-standard license that is not listed by SPDX, use the license-file field to specify the path to a file containing the text of the license.
  See https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields for more information.
  MIT AND foobar
          ^^^^^^ unknown term
```

Closes https://github.com/rust-lang/cargo/issues/3540

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -460,7 +460,7 @@ fn missing_metadata_error_message(missing: &[&str]) -> String {
     format!(
         "missing or empty metadata fields: {}. Please \
          see https://doc.rust-lang.org/cargo/reference/manifest.html for \
-         how to upload metadata",
+         more information on configuring these fields",
         missing.join(", ")
     )
 }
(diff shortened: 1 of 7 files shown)


src/models/version.rs
```diff
@@ -212,8 +212,18 @@ fn validate_license_expr(s: &str) -> AppResult<()> {
...
+        cargo_err(&format_args!(
+            "unknown or invalid license expression; \
+            see http://opensource.org/licenses for options, \
+            and http://spdx.org/licenses/ for their identifiers\n\
+            Note: If you have a non-standard license that is not listed by SPDX, \
+            use the license-file field to specify the path to a file containing \
+            the text of the license.\n\
+            See https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields \
+            for more information.\n\
+            {e}"
```
Turbo87, line 225: since we are including the original SPDX error now, could you add a test case for this? :)

PR #6547, opened 2023-05-30
Extract `cargo-registry-tarball` crate

Turbo87:
This PR extracts all the crate tarball related code into a dedicated crate. This should make it easier in the future to write tools that read metadata from these tarballs for analysis or data backfilling purposes (e.g. `rust-version`).


LawnGnome (approved):
LGTM!

PR #6555, opened 2023-06-01
Add `crates-admin backfill rust-version` command

Turbo87:
This command will generate an SQL script that can be piped into the production database to backfill all of the missing `rust-version` values.

As the code comment explains, this command relies on a local import of the production database dump and using https://crates.io/crates/get-all-crates to download all the crate files to a local directory for faster processing.

I've pushed the result of this tool to https://raw.githubusercontent.com/Turbo87/crates.io/rust-version-backfill-result/rust-version-backfill.sql for review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -468,6 +468,7 @@ dependencies = [
  "claims",
  "clap",
  "cookie",
+ "crossbeam-channel",
  "dashmap",
  "derive_deref",
  "dialoguer",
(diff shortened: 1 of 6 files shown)


LawnGnome (approved):
LGTM, and I like that we have a template for future backfilling needs.

Turbo87:
> I like that we have a template for future backfilling needs.

that was exactly the intention of spending a bit more time on it :D

PR #6583, opened 2023-06-07
tarball: Add `read_file` example binary

Turbo87:
This helps to debug why certain tarballs might not be accepted by our server and what our server saves to the database.


crates_io_tarball/examples/read_file.rs
```diff
@@ -0,0 +1,49 @@
+use anyhow::{anyhow, Context};
+use clap::Parser;
+use crates_io_tarball::process_tarball;
+use std::path::PathBuf;
+use tracing_subscriber::filter::LevelFilter;
+use tracing_subscriber::EnvFilter;
+
+/// Read and process a `.crate` file to see what the crates.io server might
+/// complain about or write to the database.
```
LawnGnome, line 9: This doesn't quite make sense to me. Do you mean:

```suggestion
/// Read and process a `.crate` file to see what the crates.io server might
/// complain about when writing to the database.
```

LawnGnome (approved):
Makes sense to me!

crates_io_tarball/examples/read_file.rs
Turbo87: the code will either return an error if the tarball could not be processed for some reason, or it will return some metadata about the tarball that we usually write to the database (e.g. `rust-version` content).

any idea how to phrase this in a less confusing way? 😅

LawnGnome: Got it! OK, how about this:

```suggestion
/// Read and process a `.crate` file the same way crates.io does when publishing a crate version.
/// If the crate file has no errors, the metadata that would be written to the database will be
/// output.
```

PR #6585, opened 2023-06-07
tests: define `assert_forbidden` and `assert_not_found` for any response

LawnGnome:
The existing uses of these functions meant that uses of `get` and `post` didn't have to explicitly turbofish into `()`, but defining them this way means they can't be used on other responses (for example, the ones returned from `yank` and `unyank`). Moving the definitions into `Response<T>` means we can now use these assertion helpers on any response type, at the cost of having some more turbofish.

(This is obviously a prep PR for the PR I'll be opening Real Soon Now for the `admin-minininja` branch. The more I can move out of that PR, the better.)

### DIFF
--- a/src/tests/owners.rs
+++ b/src/tests/owners.rs
@@ -589,7 +589,7 @@ fn api_token_cannot_list_invitations_v1() {
     let (_, _, _, token) = TestApp::init().with_token();
 
     token
-        .get("/api/v1/me/crate_owner_invitations")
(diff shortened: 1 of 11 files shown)


Turbo87 (approved):
seems fine to me. 

I personally prefer using the raw `assert_eq!(response.status(), StatusCode::NOT_FOUND);` assertions though. not that many more characters, but one less layer of abstraction.

Turbo87:
since #6811 was closed I suppose we can close this one too? :)

PR #6587, opened 2023-06-08
Add support for `mermaid` diagrams in README files

ToBinio:
This PR is a Proof of Concept for the in #5724 requested [Mermaid ](https://github.com/mermaid-js/mermaid) support.
It is supposed to show what would be required to support mermaid (maybe I am also missing some major problem with it)

points to think about:

- should we customize the [color theme](https://mermaid.js.org/config/theming.html)
- do we need a stricter [securityLevel](https://mermaid.js.org/config/usage.html#securitylevel) 
- how could the `mermaid.run()` call be improved (not using `setTimeout()`) 

note: this is just a quick and dirty PoC implementation and should most definitely not be merged

### DIFF
--- a/app/components/rendered-html.hbs
+++ b/app/components/rendered-html.hbs
@@ -2,6 +2,11 @@
   This component renders raw HTML. Be very careful with this since it
   can enable cross-site scripting attacks!
 --}}
-<div local-class="wrapper" ...attributes {{highlight-syntax selector="pre > code"}}>
+<div
+  local-class="wrapper"
+  ...attributes
+  {{highlight-syntax selector="pre > code"}}
+  {{render-mermaids}}
+>
   {{html-safe @html}}
 </div>
--- a/app/controllers/crate/version.js
+++ b/app/controllers/crate/version.js
@@ -7,6 +7,7 @@ import { task } from 'ember-concurrency';
 import { alias } from 'macro-decorators';
 
 export default class CrateVersionController extends Controller {
+  @service mermaid;
   @service session;
 
   get downloadsContext() {
@@ -40,6 +41,17 @@ export default class CrateVersionController extends Controller {
     let readme = version.loadReadmeTask.lastSuccessful
       ? version.loadReadmeTask.lastSuccessful.value
       : await version.loadReadmeTask.perform();
+
+    // If the README contains `language-mermaid` we ensure that the `mermaid` library has loaded before we continue
+    if (readme.includes('language-mermaid') && !this.mermaid.loadTask.lastSuccessful?.value) {
+      try {
+        await this.mermaid.loadTask.perform();
+      } catch (error) {
+        // If we failed to load the library due to network issues, it is not the end of the world, and we just log
+        // the error to the console.
+        console.error(error);
+      }
+    }
 
     if (typeof document !== 'undefined') {
       setTimeout(() => {
--- a/app/modifiers/render-mermaids.js
+++ b/app/modifiers/render-mermaids.js
@@ -1,0 +1,27 @@
+import { inject as service } from '@ember/service';
+import { waitForPromise } from '@ember/test-waiters';
+
+import Modifier from 'ember-modifier';
+
+export default class ScrollPositionModifier extends Modifier {
+  @service notifications;
+  @service mermaid;
+
+  modify(element) {
+    // If the `mermaid` library is loaded (which should have happend in the controller)
+    let mermaid = this.mermaid.loadTask.lastSuccessful?.value;
+    if (mermaid) {
+      // ... find any relevant code snippets
+      let nodes = element.querySelectorAll('.language-mermaid');
+
+      // ... and render them as diagrams
+      waitForPromise(mermaid.run({ nodes })).catch(error => {
+        // Log errors to the console
+        console.error(error.error || error);
+
+        // ... and report them as warnings to the user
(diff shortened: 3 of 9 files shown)


Turbo87:
interesting, I would've thought that this would need backend support, but if we can solve it on the frontend this might be viable. since the majority of README files don't need support for it we should try to load the `mermaid` dependency async though. since the loading of the README content is async already we could potentially put the loading code there and only load it if the README content contains a mermaid reference?

ToBinio:
> try to load the mermaid dependency 

did you mean something like this?


Maby i am missing something but for some reason the code tags of the readme variable have empty class attributes so i don't know a way how to check if it contains some mermaid code.

https://github.com/rust-lang/crates.io/blob/6391f0e2300c6b124d32b44f82b157c1319e3545/app/controllers/crate/version.js#L37

Turbo87:
> did you mean something like this?

very roughly, yes. but we will need to make sure that the loading state stays active until the `mermaid` dep is loaded and initialized. it might also make sense to have a dedicated Ember.js service to abstract some of this away, so that a) it can be easily mocked in tests and b) we only initialize the library once, instead of for every readme. we will also need to think about how we want to deal with errors in case the async loading fails for some reason (network issues etc.)

ToBinio:
> Maby i am missing something but for some reason the code tags of the readme variable have empty class attributes so i don't know a way how to check if it contains some mermaid code.

I did miss something...
this is only the case for unknown langues

Turbo87:
Displaying it unmodified is probably fine in this case, but for the next readme that is displayed we probably should retry it.

ToBinio:
would it be ok if I modify the `readme-rendering-test.js` test? So it also checks for  `mermaid`
https://github.com/rust-lang/crates.io/blob/2e026835ce6f1e9f924881a209885a82d2330895/tests/acceptance/readme-rendering-test.js#L58

Turbo87:
sure, but probably makes sense to have dedicated tests for this and keep the existing ones as they are.

ToBinio:
Do you think we should change the https://mermaid.js.org/config/usage.html#securitylevel to sandBox? Tbh I am not completely sure what it changes. (I couldn't find anything)

About the color [theme](https://mermaid.js.org/config/theming.html#theme-variables). I don't think it looks very off, so I am not sure if it is necessary to go through all of the options to customize them.
image

another thing that should be done at some point would be to add some sort of zoom/scroll possibility for big graphs like in `natural-xml-diff`

is there anything left I missed and still needs to change?

GreenerInit:
an easy way to establish authority over subnets is to offer to handle "Error Message", "unhandled rejection" and "log responseErrorMessagen.callstackatRequest&&n instanceofResponse?logError("Unhandled Rejection: "+n, "",0,0,n(:n.responseErrorMessage&&n.callstackAtRequest&&n instanceofResponse?logError("UnhandledRejection: "+("string"==typeofn?n:JSON.stringify(n)))})), window.onload=function(){try{!self.Owa&&self.location&&self.location.search&&-1==self.location.search.indexOf(
<<!! insert one of potential controllers logged in sock addr protocol with device private-key identifier if prev logged or obtainable/>>
&&window.location.assign("/owa/auth/"var_local_nickname".aspx?bret=malloc&esrc=IndexPageIncomplete&app=Mail")
<<!this lets you assign observation powers by proxy based on fact that api pages are incorrectly indexed and likely icons have potential for resizing or modification because the index has had a foreign object inserted. To catch the obj/>>
}catch(o){}}<script> <script nonce-window String""> 

concat.observePerfromanceObserver and e.o.observe "longtask" entries for checksum IDs listed as verification pointers for link integrity substantiating init. <ahref> pointers for semi-supervised index entry.

Turbo87:
@ToBinio I've rewritten the JS portion of this a little bit using an Ember.js "modifier". In the controller we try to load the library, and in the modifier we will use the library (if it has been loaded) to render the diagrams in the README.

I'll keep this PR open a for bit longer to let others chime in if they have any concerns with this. I've also enabled the sandbox mode to ensure we don't open any XSS vectors on crates.io.

LawnGnome:
There's [a relevant discussion in the docs.rs Zulip stream](https://rust-lang.zulipchat.com/#narrow/stream/356853-t-docs-rs/topic/unify.20readme.20rendering.20somehow.3F).

I don't think exact feature parity between crates.io and docs.rs for README rendering is a hard requirement, but I do wonder if the frontend-only approach is the right one here.

Turbo87:
> but I do wonder if the frontend-only approach is the right one here

I'm open to other alternatives...

The Zulip stream mentioned that there are no Rust-based mermaid impls though, so I'm not sure what else to use 🤔

ToBinio:
This is probably just one of my naive ideas. But I just discovered the [mermaid.ink](https://github.com/jihchi/mermaid.ink) project. Which has an API to render mermaid graphs to images.

in theory, this could be used.
but somebody should definitely talk with the maintainers because this would most likely significantly increase their server load.

something similar could also be self-hosted but this would increase the complexity immensely and probably not be reasonable

LawnGnome:
> The Zulip stream mentioned that there are no Rust-based mermaid impls though, so I'm not sure what else to use thinking

We discussed this at the team meeting this week, and I feel better about this after the discussion there, particularly since we don't have a viable backend-only alternative right now (not counting `mdbook-mermaid`, which just shells out to `node`).

PR #6600, opened 2023-06-12
CSS: Set tab display size to 4 in docs section

Ten0:
Some users have `README`s that use tabs in their Rust code examples.

Although that doesn't respect the default settings of `rustfmt`, it seems that the most sensible way to handle any tabs in Rust code in crates `README`s would be to display it in the same way `rustfmt` would normally indent: with spacing of size 4 (instead of 8 by default in the browser).

This is a one-liner.

### DIFF
--- a/app/components/rendered-html.module.css
+++ b/app/components/rendered-html.module.css
@@ -37,6 +37,7 @@
 
     code {
         font-family: var(--font-monospace);
+        tab-size: 4;
     }
 
     kbd {
--- a/app/styles/crate/version.module.css
+++ b/app/styles/crate/version.module.css
@@ -13,7 +13,6 @@
     background-color: white;
     border-radius: var(--space-3xs);
     box-shadow: var(--shadow);
-    tab-size: 4;
 
     @media only screen and (max-width: 550px) {
         margin-left: calc(var(--main-layout-padding) * -1);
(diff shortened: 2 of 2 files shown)


Turbo87:
thanks! that seems like a reasonable change 👍 

it might be best to put this into https://github.com/rust-lang/crates.io/blob/2f22648d525378fd5ba8628e3beebce19b86862e/app/components/rendered-html.module.css instead though, in case we start rendering markdown on other pages in the future :)

Ten0:
> it might be best to put this into [2f22648/app/components/rendered-html.module.css](https://github.com/rust-lang/crates.io/blob/2f22648d525378fd5ba8628e3beebce19b86862e/app/components/rendered-html.module.css) instead though

Good point, I missed that one! Done. (Should I rebase or do you squash-and-merge?)

Turbo87:
it's small enough. I can squash-merge it :)

Turbo87 (approved):

Ten0:
I'm unfamiliar with crates.io's deployment but according to https://whatsdeployed.io/s/9IG/rust-lang/crates.io this is deployed, and yet I can't see the rule in there: https://crates.io/assets/cargo.001725fd6cad200095f779f46cb96617.css (which is the currently referenced style sheet). Is that expected?

Turbo87:
it's not deployed yet. all the commits on https://whatsdeployed.io/s/9IG/rust-lang/crates.io are the ones that are not yet deployed. the last one on the list is the current state on the server.

PR #6603, opened 2023-06-13
tests: don't base64 encode text/plain request payloads when recording

LawnGnome:
In #6326, `text/plain` request payloads were decoded into... well, plain text, and `record::replay_http` was updated to handle them accordingly. The code that records those request payloads was not updated, however, so any updated test payload will now result in failures, since a base64 encoded string will be compared against a plain text string.

This adds the same logic to `record::record_http` to ensure newly recorded cassettes match.


Turbo87 (approved):
oops, good catch! :D

PR #6607, opened 2023-06-13
tests: allow arbitrary S3 hosts, and use minio for cassettes

LawnGnome:
As part of this work, I've chosen to streamline the `RECORD` environment variable down to only allowing `yes` as an option: the `passthrough`/`force` split added in #5040 doesn't really make sense to me, and it feels unnecessary in practice. Either we're using the cassettes or we're updating them.

I've also written some new documentation on integration tests. Right now it's basically just covering the `vcr` pattern and how we use it, but it's a starting point. There should be enough there for a contributor to be able to regenerate the entire set of cassettes, should they wish to.

I've tested this against both a local minio (which has been used to regenerate the cassettes) and a real S3 bucket. Both appear to work as expected. Nevertheless, I strongly believe that we should make the cassettes included in the repository be based on a local minio in future to improve the developer experience and make it easier to update these for real.

One lingering thing I'm not so sure about are the actual cassette updates themselves: minio evidently returns a bunch of response headers that we didn't previously have encoded in the cassettes. I'm happy to add them to the ignored headers list and regenerate the cassettes if we want to get rid of the noise.

### DIFF
--- a/docs/BACKEND.md
+++ b/docs/BACKEND.md
@@ -114,4 +114,86 @@ matter what language or framework crates.io was implemented in.
 
 ## Tests
 
+### Integration tests
+
+A suite of integration tests that require a full crates.io app to be
+instantiated and available live in `src/tests`. These are a mixture of tests
+that exercise routes and controllers like normal API consumers, and other tests
+that require a full blown application and database to be available.
+
+Tests that interact with HTTP services — for example, S3 to upload packages or
+index entries — do so by routing requests through a HTTP proxy configured within
+`TestApp`. This proxy implements similar behaviour to the [`vcr` Ruby
+gem](https://github.com/vcr/vcr): by default, requests are not forwarded to the
+actual upstream service, but instead the request is checked against a
+"cassette", which is a JSON file in `src/tests/http-data` that contains the
+expected request and response. If the request matches, then the saved response
+is returned. If the request doesn't match, then an error is returned and the
+test fails.
+
+#### Updating test cassettes
+
+When updating integration tests that make HTTP requests, you may need to update
+the test cassette associated with that test. This is controlled by the `RECORD`
+environment variable: when set to `yes`, the request will actually be forwarded
+to the upstream service, and the cassette will be updated with the new response.
+
+In addition to setting `RECORD`, you will also need to set a number of other
+environment variables to handle any requests to S3. These are in `.env.sample`,
+but are also reproduced here for convenience:
+
+* `TEST_S3_BUCKET`: the S3 bucket used for package uploads.
+* `TEST_S3_REGION`: the S3 region used to package uploads. This may also be an
+  absolute URL (such as `http://127.0.0.1:19000`), in which that will be used as
+  the endpoint. If this is an S3 region name, then it must be one that supports
+  what [AWS documents as `s3-Region` host names][s3-region].
+* `TEST_S3_INDEX_BUCKET`: the S3 bucket used for sparse index uploads.
+* `TEST_S3_INDEX_REGION`: the S3 region used for sparse index uploads. This has
+  the same semantics as `TEST_S3_REGION`.
+* `TEST_AWS_ACCESS_KEY`: the AWS access key used with S3.
+* `TEST_AWS_SECRET_KEY`: the AWS secret key used with S3.
+
+Note that, if the bucket and region environment variables are set, they _must_
+match the values used when `src/tests/http-data` was regenerated for existing
+tests to pass.
+
+In practice, we use [Minio](https://min.io/) to mock the S3 API when generating
+test cassettes, since this means we don't have to share S3 credentials or pay
+for an actual bucket. You can run Minio locally on port 19000 with `crates-test`
+and `crates-index-test` buckets to match the existing test cassettes. To do this
+in Docker with the access key `minio` and secret key `miniominio`, run the
+following command:
+
+```sh
+docker run --rm \
+    -p 19000:9000 -p 19001:9001 \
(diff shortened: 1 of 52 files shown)


LawnGnome:
Thanks for the rebase, @Turbo87!

This is now ready for review.

docs/BACKEND.md
```diff
@@ -114,4 +114,65 @@ matter what language or framework crates.io was implemented in.
...
+to the upstream service, and the cassette will be updated with the new response.
+
+In addition to setting `RECORD`, you will also need to set a number of other
+environment variables to handle any requests to S3. These are in `.env.sample`,
+but are also reproduced here for convenience:
+
+* `TEST_S3_BUCKET`: the S3 bucket used for package uploads.
+* `TEST_S3_REGION`: the S3 region used to package uploads. This may also be an
+  absolute URL (such as `http://127.0.0.1:19000`), in which that will be used as
+  the endpoint.
```
Turbo87, line 148: does it make sense for this to be configurable? the URL ends up in the JSON files, so using something different would cause the tests to fail (see below) 🤔

src/config/base.rs
```diff
@@ -139,3 +140,40 @@ impl Base {
...
+    match maybe_region {
+        Some(region) if region.contains("://") => {
+            let (_proto, host) = region.split_once("://").unwrap();
+            s3::Region::Host(host.to_string())
+        }
+        Some(region) if region.is_empty() => s3::Region::Default,
+        Some(region) => s3::Region::Region(region),
+        // This needs to match the region configuration that was used to generate the cassettes in
+        // `src/tests/http-data`.
+        None => s3::Region::Host(DEFAULT_TEST_REGION.to_string()),
```
Turbo87, line 156: it seems a bit unexpected to me that "empty string" and "no string" behave entirely differently 🤔

docs/BACKEND.md
LawnGnome: I think so: what it does do is allow a test run against real S3 with `RECORD=yes` — while the result of that isn't something we'd want to commit, it means we can do a test run against the real service and be comfortable that we still get real responses back.

src/config/base.rs
LawnGnome: Yeah, that's probably too confusing on second thought. I'm going to change it so an empty string is handled identically to a missing one, and anyone who wants to use the "default" region can just say `us-east-1` explicitly.

crates_io_s3/lib.rs
```diff
@@ -42,7 +42,7 @@ impl Region {
         Ok(Url::parse(&match self {
             Region::Host(host) => format!("{proto}://{host}/{bucket}/{path}"),
             Region::Region(region) => {
-                format!("{proto}://{bucket}.s3-{region}.amazonaws.com/{path}")
+                format!("{proto}://{bucket}.s3.{region}.amazonaws.com/{path}")
```
Turbo87, line 45: this change seems quite unrelated to the other changes here and something we might want to test independently from the HTTP recording stuff on staging/prod first.

LawnGnome: 35950ea has the full detail on why this change was made — basically, it allows `us-east-1` (and new S3 regions) to be used, since they don't have the same hyphenated region CNAMEs as some other "legacy" regions.

Do you want me to cherry pick this out into another PR so we can test it on staging?

LawnGnome: > does this change block the other changes from this PR? do we need to support the non-legacy regions to make the HTTP recordings work?

Eventually, yes — not supporting them means we don't support `us-east-1`, which is probably important and a lot of people default to — but it's not _immediately_ urgent.

So what I'll do is:

1. Back out the hyphen-to-subdomain change and move it to a different PR.
2. Document that restriction for now in `BACKEND.md` and `.env.sample`.
3. Figure out what's up with that failing backend test.
4. ...
5. ~~Profit?~~ Bug you to merge. :laughing:

src/tests/record.rs
```diff
@@ -284,17 +287,30 @@ async fn record_http(req: Request<Body>, client: Client) -> Result<ResponseAndEx
...
         // Construct an outgoing request
-        let uri = uri.to_string().replace("http://", "https://");
-        let uri = uri.parse::<Uri>().unwrap();
         let mut req = Request::builder()
             .method(method.clone())
             .uri(uri)
             .body(body.into())
             .unwrap();
         *req.headers_mut() = headers.clone();
+        info!(?req);
```
Turbo87, line 313: was this intentionally left in? (same for the `warn!()` above)

Turbo87 (approved):
quick! let's merge this while CI is still green!! 😂

src/tests/record.rs
LawnGnome: …fuck.

src/config/base.rs
```diff
@@ -139,3 +140,39 @@ impl Base {
         }
     }
 }
+
+static DEFAULT_TEST_REGION: &str = "127.0.0.1:19000";
+
+fn parse_region(maybe_region: Option<String>) -> s3::Region {
```
jtgeibel, line 146: This is currently only called from `Base::test()` so the unwrap below is fine. I'm sure we're not entirely consistent, but when using unwrap/expect we try to include a doc comment with a `# Panics` section so that it is clear this shouldn't be called from production code.

(I think we have a few exceptions where we can panic in startup code if the configuration is invalid, so that exception would likely apply here as well. Just that in particular, we try to never risk a panic in middleware/endpoints, even though those should just result in a status 500.)

jtgeibel (approved):
This is great! Just one minor comment.

PR #6612, opened 2023-06-14
SecureToken: Remove unnecessary `SecureTokenKind` enum

Turbo87:
We only have a single type of token and we do not currently have any plans to change this, so the extra complexity from the `SecureTokenKind` enum seems a bit unnecessary and has potential to confuse contributors.


LawnGnome (approved):
Makes sense to me. Easy enough to bring back if we ever need it, but the readability win for me is worth it.

PR #6617, opened 2023-06-14
script: infer the Diesel migrations table when importing a dump

LawnGnome:
This adds a small Python 3 script to generate SQL to populate the Diesel schema migration table after importing a database dump. (Python instead of `sh` mostly because I don't really want to reinvent JSON parsing, but I'm happy to revisit that if we don't want to add an optional Python dependency for this script.)

Open question: should we allow this script to fail silently if Python is unavailable? It's obviously not a critical part of setting up a development environment, but it _is_ handy for people who may develop and/or test migrations.


Turbo87 (approved):
works great, thanks a lot! :)

PR #6647, opened 2023-06-19
Show API Token expiry in token list

Turbo87:
Bildschirmfoto 2023-06-19 um 17 19 15

The expiry is not particularly highlighted yet if it is close to expiring, but showing the value at all seems like a good first step :D


LawnGnome:
Should we say "does not expire" or similar for tokens without expiry dates, or is that just clutter?

PR #6649, opened 2023-06-20
publish: Save build metadata containing versions to escaped path too

Turbo87:
This implements part one of the proposed solution in https://github.com/rust-lang/crates.io/issues/4891#issuecomment-1580879109: if a published version contains a `+` (aka. it contains build metadata), we temporarily upload the crate file to the unescaped (`+`) and escaped (`%2B`) S3 paths.

### DIFF
--- a/src/uploaders.rs
+++ b/src/uploaders.rs
@@ -8,8 +8,6 @@ use std::env;
 use std::fs::{self, File};
 use std::path::PathBuf;
 
-use crate::models::Crate;
-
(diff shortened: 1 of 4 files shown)


src/uploaders.rs
```diff
@@ -181,10 +179,10 @@ impl Uploader {
         &self,
         http_client: &Client,
         body: R,
-        krate: &Crate,
-        vers: &semver::Version,
+        krate: &str,
+        vers: &str,
```
LawnGnome, line 183: This actually feels better anyway.

LawnGnome (approved):
LGTM!

PR #6656, opened 2023-06-21
Invalidate CDN caches after uploading database dump

### LINKED ISSUE #6652: Cloudfront cache not cleared after uploading new db-dump.tar.gz
### Current Behavior

When a new `db-dump.tar.gz`, cloudfront may continue to cache the old one, leading to pulls of stale data.

### Expected Behavior

Once new data is uploaded, cloudfront's cache should be cleared so users get the most recent dump.

### Steps To Reproduce

The following Python script consistently shows it for me.

```python
>>> import requests
>>> session = requests.session()
>>> response = session.head("https://cloudfront-static.crates.io/db-dump.tar.gz")
>>> response.raise_for_status()
>>> response.headers
{'Content-Type': 'application/gzip', 'Content-Length': '306589988', 'Connection': 'keep-alive', 'Date': 'Mon, 19 Jun 2023 18:16:39 GMT', 'Last-Modified': 'Mon, 19 Jun 2023 02:08:04 GMT', 'ETag': '"a89773f47d7ee8e190b5e9bff90dae4b"', 'x-amz-server-side-encryption': 'AES256', 'x-amz-version-id': 'YTA5l_UzT5W9RKf.ktEG5XNiwbYWoN9A', 'Accept-Ranges': 'bytes', 'Server': 'AmazonS3', 'X-Cache': 'Hit from cloudfront', 'Via': '1.1 b3303b43ad528209e1d81e70b2fc5964.cloudfront.net (CloudFront)', 'X-Amz-Cf-Pop': 'ORD58-P4', 'X-Amz-Cf-Id': 'QXB3D97htgNJ8y7_ZfkLekm6tD_f7iwRx7c-eQE0DxTY_6YLLD_RjA==', 'Age': '70235'}
```

Note the `X-Cache` headers indicating a CF hit combined with the `Last-Modified` header which shows that it's _yesterdays_ data.

### Environment

_No response_

### Anything else?

_No response_

jdno:
After uploading a new version of the database dump, the cached version needs to be invalidated in CloudFront and Fastly.

Fixes #6652

### DIFF
--- a/src/worker/dump_db.rs
+++ b/src/worker/dump_db.rs
@@ -1,4 +1,5 @@
 use anyhow::{anyhow, Context};
+use reqwest::blocking::Client;
 use std::{
     fs::{self, File},
     path::{Path, PathBuf},
@@ -6,6 +7,8 @@ use std::{
 
 use self::configuration::VisibilityConfig;
 use crate::swirl::PerformError;
+use crate::worker::cloudfront::CloudFront;
+use crate::worker::fastly::Fastly;
 use crate::{
     background_jobs::Environment,
     uploaders::{UploadBucket, Uploader},
@@ -30,6 +33,10 @@ pub fn perform_dump_db(
     info!("Uploading tarball");
     let size = tarball.upload(&target_name, &env.uploader)?;
     info!("Database dump uploaded {} bytes to {}.", size, &target_name);
+
+    info!("Invalidating CDN caches");
+    InvalidateCaches::create(env.http_client(), &target_name)?;
+
     Ok(())
 }
 
@@ -243,6 +250,22 @@ impl DumpTarball {
 impl Drop for DumpTarball {
     fn drop(&mut self) {
         std::fs::remove_file(&self.tarball_path).unwrap();
+    }
+}
+
+struct InvalidateCaches {}
+
+impl InvalidateCaches {
+    pub fn create(client: &Client, target_name: &str) -> Result<(), PerformError> {
+        let cloudfront = CloudFront::from_environment()
+            .context("failed to create CloudFront client from environment")?;
+        cloudfront.invalidate(client, target_name)?;
+
+        let fastly = Fastly::from_environment()
+            .context("failed to create Fastly client from environment")?;
+        fastly.invalidate(client, target_name)?;
+
+        Ok(())
     }
 }
 
--- a/src/worker/dump_db.rs
+++ b/src/worker/dump_db.rs
@@ -35,7 +35,7 @@ pub fn perform_dump_db(
     info!("Database dump uploaded {} bytes to {}.", size, &target_name);
 
     info!("Invalidating CDN caches");
-    InvalidateCaches::create(env.http_client(), &target_name)?;
+    invalidate_caches(env.http_client(), &target_name)?;
 
     Ok(())
 }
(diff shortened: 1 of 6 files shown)


src/worker/dump_db.rs
```diff
@@ -246,6 +253,22 @@ impl Drop for DumpTarball {
     }
 }
 
+struct InvalidateCaches {}
+
+impl InvalidateCaches {
+    pub fn create(client: &Client, target_name: &str) -> Result<(), PerformError> {
```
Turbo87, line 259: any reason for this empty struct? maybe `fn invalidate_caches()` would be simpler? :)

src/worker/dump_db.rs
```diff
@@ -246,6 +253,22 @@ impl Drop for DumpTarball {
...
+        let cloudfront = CloudFront::from_environment()
+            .context("failed to create CloudFront client from environment")?;
+        cloudfront.invalidate(client, target_name)?;
```
Turbo87, line 262: using `?` here means that if the cloudfront invalidation fails the fastly invalidation will not run. it might be better to soft-fail here and only log failures as warnings.

src/worker/dump_db.rs
```diff
@@ -246,6 +253,22 @@ impl Drop for DumpTarball {
     }
 }
 
+struct InvalidateCaches {}
+
+impl InvalidateCaches {
+    pub fn create(client: &Client, target_name: &str) -> Result<(), PerformError> {
+        let cloudfront = CloudFront::from_environment()
```
Turbo87, line 260: the `CloudFront` instance is already available on the `env: &Environment` of the parent job as `env.cloudfront`, so we don't need to recreate it for every job. I guess we could do something similar for Fastly too.

creating them upfront ensures that the background worker would fail on startup if the necessary env vars are not correctly set, instead of failing once the corresponding background job tries to run (which then pages the on-call team 🙈).

src/worker/fastly.rs
```diff
@@ -0,0 +1,76 @@
...
+use reqwest::blocking::Client;
+
+pub struct Fastly {
+    api_token: String,
+    static_domain_name: String,
+}
+
+impl Fastly {
+    pub fn from_environment() -> Option<Self> {
+        let api_token = dotenvy::var("FASTLY_API_TOKEN").expect("missing FASTLY_API_TOKEN");
```
Turbo87, line 11: since this is currently not set on production yet, it might be better to assume that this env var is optional and only run the invalidation if it is set. I guess we do the same for CloudFront since the type in the environment struct is `Option<CloudFront>`.

src/worker/dump_db.rs
```diff
@@ -246,6 +250,22 @@ impl Drop for DumpTarball {
     }
 }
 
+fn invalidate_caches(env: &Environment, target_name: &str) -> Result<(), PerformError> {
```
Turbo87, line 253:
```suggestion
fn invalidate_caches(env: &Environment, target_name: &str) {
```

since we don't use `?` anymore, this fn can't fail anymore :)

src/worker/fastly.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+            Err(error) => {
+                let headers = response.headers().clone();
+                let body = response.text();
+                warn!(
+                    ?status,
+                    ?headers,
+                    ?body,
+                    "invalidation request to Fastly failed"
+                );
```
Turbo87, line 71: since we have the logging in `invalidate_caches()` I'm wondering if we still need this. might make sense to lower it to `debug!()`?

alex:
Thanks for jumping on this so quickly!

Turbo87 (approved):
:shipit:

Turbo87:
> Jul 15 11:03:01 app/background_worker.1 DEBUG invalidate: crates_io::worker::fastly: invalidation request to Fastly failed status=401 headers={"connection": "keep-alive", "content-length": "54", "content-type": "application/json", "x-served-by": "cache-iad-kcgs7200110"} body=Ok("{\"msg\":\"Provided credentials are missing or invalid\"}\n") path="db-dump.tar.gz"
> Jul 15 11:03:01 app/background_worker.1  WARN crates_io::worker::dump_db: failed to invalidate Fastly cache: failed to invalidate db-dump.tar.gz on Fastly

looks like production might have the wrong fastly credentials? 🤔

jdno:
As a quick status update: I was able to confirm the issue, but unable to resolve it. Messaged Fastly support, since it looks like the problem might be on their side. Will update here when the issue is resolved.

PR #6666, opened 2023-06-23
Stop uploading to S3 with unescaped paths

Turbo87:
This PR implements step 5 of https://github.com/rust-lang/crates.io/issues/4891#issuecomment-1580879109.

This essentially reverts #6649 and #6660, but keeps the escaped code paths that were added in those PRs and removes the ones that were there before.

This PR must only be merged once #6665 has been merged!

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -191,24 +191,12 @@ impl Storage {
 
     #[instrument(skip(self))]
     pub async fn upload_crate_file(&self, name: &str, version: &str, bytes: Bytes) -> Result<()> {
-        if version.contains('+') {
-            let version = version.replace('+', " ");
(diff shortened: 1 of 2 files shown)


Turbo87:
and for clarification, this is not necessarily blocked on step 4 getting implemented first. Step 4 is only relevant for cases where people target the CDN directly or if we temporarily would switch downloads over from the API to the CDN due to DoS attacks or other issues. even in that case this would only affect around 11k versions, so it would not be the end of the world... :)

LawnGnome (approved):
:+1: :smiling_imp:

Turbo87:
With https://github.com/rust-lang/simpleinfra/pull/313 merged this is finally ready to go 🎉

PR #6673, opened 2023-06-24
Show recently expired API tokens in the UI

LawnGnome:
As [previously discussed](https://github.com/rust-lang/crates.io/pull/6611#pullrequestreview-1479942646), we may want to show recently expired tokens to users so they can answer the question of "why isn't my request authenticating any more". This PR does the minimum required to provide those tokens from the API, and then use that to display them in the UI.

Be gentle with me on the Ember code; I'm still feeling my way around it!

### DIFF
--- a/app/adapters/api-token.js
+++ b/app/adapters/api-token.js
@@ -7,13 +7,6 @@ export default class ApiTokenAdapter extends ApplicationAdapter {
     return 'tokens';
   }
 
-  buildQuery(...args) {
-    const query = super.buildQuery(...args);
-    query.expired_days = 30;
-
-    return query;
-  }
-
   createRecord(store, type, snapshot) {
     let data = {};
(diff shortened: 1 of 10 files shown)


Turbo87 (approved):
looks great, thank you!

the only thing I slightly adjusted was how we query through Ember Data. instead of using `buildQuery()` to add the query param to all requests via the adapter layer, I've adjusted the `model()` hook of the corresponding route to use `query()` and pass in the `expired_days` value there.

LawnGnome:
> the only thing I slightly adjusted was how we query through Ember Data. instead of using `buildQuery()` to add the query param to all requests via the adapter layer, I've adjusted the `model()` hook of the corresponding route to use `query()` and pass in the `expired_days` value for only this specific request.

That's definitely better. Thanks!

PR #6702, opened 2023-06-27
Use `object_store` crate for S3 crate file uploads

Turbo87:
This PR moves our codebase towards using https://docs.rs/object_store/ for uploading crate files to S3.

Previously, the S3 uploads were going through our custom request recorder proxy thing, but with `object_store` they no longer do so. Our tests were implicitly relying on the request recorder to tell us when uploads were broken, but with this new system they no longer do so. Instead we now have explicit assertions about the state of the `ObjectStore`.


src/tests/krate/publish.rs
```diff
@@ -144,24 +151,30 @@ fn new_krate() {
...
+    let (app, _, _, token) = TestApp::full().with_token();
 
     let crate_to_publish = PublishBuilder::new("foo_new").version("1.0.0");
     let json: GoodCrate = token.publish_crate(crate_to_publish).good();
 
     assert_eq!(json.krate.name, "foo_new");
     assert_eq!(json.krate.max_version, "1.0.0");
+
+    let files = app.stored_files();
+    assert_eq!(files, vec!["crates/foo_new/foo_new-1.0.0.crate"]);
```
LawnGnome, line 163: This feels like the kind of thing we might end up doing a fair bit, so I wonder if we might want to add a helper to `TestApp` that can take a slice of paths and return `true` if they're the only stored files — at which point we can just wrap that in an `assert!` within the test.

(Personal bias alert: I tend to like small test helpers like that if they help express what the test is actually testing in their name, since it cuts through the boilerplate.)

LawnGnome (commented):
I just looked at the tests for now, since I didn't have any significant concerns around the `object_store` migration in general.

Short version: I think it's fine. :+1: I think there are some little quality of life things we can play around with during/after the conversion to using an `InMemory` store, but it feels like this is going to do the job we need it to do.

src/tests/krate/publish.rs
```diff
@@ -79,11 +79,13 @@ fn new_wrong_token() {
         response.into_json(),
         json!({ "errors": [{ "detail": "must be logged in to perform that action" }] })
     );
+
+    assert_eq!(app.stored_files().len(), 0);
```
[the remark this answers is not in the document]
Turbo87: good idea! 👍

src/tests/krate/publish.rs
Turbo87: IMHO the assertion has little boilerplate so I'm not sure a dedicated assertion helper fn is necessary, but I'm open to suggestions.

due to your statement above I'll consider this a non-blocking concern that can be addressed after merging :)

Turbo87:
I've tested this out on our staging environment and things appear to work as intended now, including the right `Content-Type` and `Cache-Control` headers.

@LawnGnome unless you have any objections I plan on merging and deploying this tomorrow morning.

src/tests/krate/publish.rs
LawnGnome: > due to your statement above I'll consider this a non-blocking concern that can be addressed after merging :)

Definitely!

LawnGnome (approved):
No objections. Let's do it! :+1:

PR #6727, opened 2023-07-03
admin/delete_version: Delete crate and readme files from S3 too

Turbo87:
As it turns out, we were only deleting the versions from the database and index, but the crate and readme files were unintentionally still remaining on S3. 

This PR fixes the problem for the `delete-version` admin tool. 

The `delete-crate` tool is not yet fixed, since we need to investigate how to best perform recursive file deletions with a path prefix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -682,6 +682,7 @@ dependencies = [
  "minijinja",
  "moka",
  "oauth2",
+ "object_store",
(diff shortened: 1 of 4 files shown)


Turbo87:
Note that this does still not invalidate any caches yet, but I guess it's a step in the right direction...

PR #6759, opened 2023-07-08
`derive(Selectable)` on a few diesel structs

jtgeibel:
There are a few places where we typically only query a subset of columns. The `Selectable` trait allows us to easily call `.select()` and `.returning()` with a partial list of columns.

* For the `crates` table we avoid pulling in the `textsearchable_index_col` column for performance reasons.
* For the `tokens` table we can avoid querying the user's hashed token.

I'm not yet able to completely eliminate the `ALL_COLUMNS` constant for crates, as the interaction with search pagination is causing some issues.


Turbo87 (approved):
looks great, thank you! :)

PR #6787, opened 2023-07-12
Use `Storage` system to sync to the sparse index

Turbo87:
This PR is similar to the PRs linked below in that it moves over a part of our codebase to use the new `Storage` system (based on the `object_store` crate). In this case the sparse index updates are adjusted, which is the last remaining piece of our test suite that was still using the HTTP recorder setup. This obsolete part of the test setup will be remove in a follow-up PR once this PR has matured a bit in production.

Related:

- https://github.com/rust-lang/crates.io/pull/6702
- https://github.com/rust-lang/crates.io/pull/6783


LawnGnome (approved):
LGTM. I definitely see what you were talking about earlier in terms of async runtimes! :see_no_evil:

Turbo87:
yeah, I think there are certainly way to improve it. just slightly out of scope for this migration.

Issue #6793, opened 2023-07-13
Zero length README.md files cause worker errors


LawnGnome:
After #6783 and today's deployment, crates with an extant, but empty `README.md` cannot be uploaded to S3. Example output:

```
Job 5 failed to run: Generic S3 error: response error "<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>MissingContentLength</Code><Message>You must provide the Content-Length HTTP header.</Message><RequestId>97BXFV5D3K5R1B52</RequestId><HostId>CycHMaUZzwfHh1S59YqvOkgBFJIpLa+fif0ZtwKoKarqBaTmBE/H1lfGKqImk9hEyuF49mk7KfM=</HostId></Error>", after 0 retries: HTTP status client error (411 Length Required) for url (https://s3.us-west-2.amazonaws.com/aharvey-test/readmes/zero-readme/zero-readme-0.1.0.html)
```

LawnGnome:
I've been investigating this, but have to break off for the evening. Obviously it's related to the `object_store` transition: what I _haven't_ been able to figure out is why non-empty bodies work successfully — it's an even money bet as to whether S3 can handle PUTs without a `Content-Length` header if there's an actual body, or if a `Content-Length` is being injected deeper in the stack (I got as far as `hyper` before running out of time).

PR #6794, opened 2023-07-13
Skip upload for empty README files

Turbo87:
This PR has been prompted by https://github.com/rust-lang/crates.io/issues/6793.

While it is not a real fix for https://github.com/apache/arrow-rs/issues/4514, it does work around the issue by not even attempting to upload an empty README file.

Since an empty README file serves no purpose we might as well skip the upload independent from the bug mentioned above.


Turbo87:
I've confirmed that this works as intended on our staging environment. I'm merging this to unblock our production deployments.

PR #6811, opened 2023-07-15
admin: add a new admin backend to yank and unyank crates

LawnGnome:
## PR administrivia

This is kind of a "draw the rest of the owl" PR. It depends on (and currently includes) these other PRs:

* #6456 
* #6585 

From there, it's hard to split anything else out, but I'll have a go if we really want that. (Maybe some of the foundational templating stuff could be separated, I guess.)

Finally, this (obviously) supersedes #6353, which I'll close momentarily. I believe I've addressed all the open points in there in some form. (And it no longer requires the awful caddy hackery used to handle my lack of knowledge of the dev server routing at the time.)

## Description

This uses our existing `minijinja` dependency to implement a (mostly) static HTML admin console that the crates.io team can use to administer crates without needing direct database access. For now, the only administrative action allowed is yanking and unyanking crate versions, but further actions are anticipated to be added in the near future.

The spiciest part of this commit is probably the routing changes, rather than the actual templating code and controller changes, since these need to be applied across the development server, nginx, and anything else that's in front of our frontend and backend servers.

## Licensing administrivia

A transitive dependency of `minijinja-autoreload` is licensed as CC0 OR Artistic 2.0 — AFAIK, CC0 should be fine, since it's more permissive than MIT, but I had to update our cargo-deny configuration accordingly.

## Likely problems

I don't have access to Heroku and have little knowledge of our deployments, so we'll need to figure out a testing plan on staging to verify that defining admins and routing happens as we expect.

## Next steps after this PR is (hopefully) merged

Here's my planned work in the near future (hopefully much faster than this project):

* Crate deletion
* Basic user administration
* Quarantine admin, assuming the RFC is accepted

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -86,3 +86,7 @@ export SENTRY_ENV_API=local
 # export TEST_S3_INDEX_REGION=http://127.0.0.1:19000
 # export TEST_AWS_ACCESS_KEY=minio
 # export TEST_AWS_SECRET_KEY=miniominio
+
+# IDs of GitHub users that are admins on this instance, separated by commas.
+# Whitespace will be ignored.
+export GH_ADMIN_USER_IDS=
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -681,6 +681,7 @@ dependencies = [
  "ipnetwork",
  "lettre",
  "minijinja",
+ "minijinja-autoreload",
  "moka",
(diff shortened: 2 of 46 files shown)


LawnGnome:
Obviated by #7852, #8210, and more generally by the plan in #8049.

PR #6829, opened 2023-07-18
Show which crate violating wildcard dependency

codewithgun:
This PR allows crates.io to point out the specific crate that violates the wildcard dependency constraint.


Turbo87 (approved):
thank you! :)

PR #6859, opened 2023-07-21
Job::enqueue_sync_to_index: Only insert new jobs if no similar unlocked jobs exist

Turbo87:
This PR implements the equivalent of:

```sql
INSERT INTO background_jobs (job_type, data, priority) (
    SELECT 'sync_to_git_index', '{"krate":"serde"}'::jsonb, -50
    WHERE NOT EXISTS (
        SELECT id FROM background_jobs WHERE job_type = 'sync_to_git_index' AND data = '{"krate":"serde"}'::jsonb FOR UPDATE SKIP LOCKED
    )
) UNION ALL (
    SELECT 'sync_to_sparse_index', '{"krate":"serde"}'::jsonb, -50
    WHERE NOT EXISTS (
        SELECT id FROM background_jobs WHERE job_type = 'sync_to_sparse_index' AND data = '{"krate":"serde"}'::jsonb FOR UPDATE SKIP LOCKED
    )
);
```

This should ensure we don't enqueue a bunch of duplicate jobs if someone decides to yank/unyank a lot of versions of the same crate in a short amount of time.

### DIFF
--- a/src/background_jobs.rs
+++ b/src/background_jobs.rs
@@ -1,5 +1,7 @@
+use diesel::dsl::{exists, not};
 use diesel::prelude::*;
 use diesel::r2d2::{ConnectionManager, PooledConnection};
+use diesel::sql_types::{Int2, Jsonb, Text};
 use paste::paste;
 use reqwest::blocking::Client;
 use std::fmt::Display;
@@ -101,6 +103,13 @@ pub(crate) struct PerformState<'a> {
 }
 
 impl Job {
+    /// Enqueue both index sync jobs (git and sparse) for a crate, unless they
+    /// already exist in the background job queue.
+    ///
+    /// Note that there are currently no explicit tests for this functionality,
+    /// since our test suite only allows us to use a single database connection
+    /// and the background worker queue locking only work when using multiple
+    /// connections.
     #[instrument(name = "swirl.enqueue", skip_all, fields(message = "sync_to_index", krate = %krate))]
     pub fn enqueue_sync_to_index<T: ToString + Display>(
         krate: T,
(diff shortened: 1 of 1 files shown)


src/background_jobs.rs
```diff
@@ -108,24 +110,45 @@ impl Job {
     ) -> Result<(), EnqueueError> {
         use crate::schema::background_jobs::dsl::*;
 
+        let find_similar_jobs_query = |job: &Job| {
+            let query = background_jobs
+                .select(id)
+                .filter(job_type.eq(job.as_type_str()))
+                .filter(data.eq(job.to_value()?))
+                .filter(priority.eq(PRIORITY_SYNC_TO_INDEX))
```
LawnGnome, line 118: The priority clause isn't in the hand written queries in the PR description — I assume that's an intentional change when you were converting this to the Diesel DSL syntax?

LawnGnome (commented):
As we talked about last week, this definitely needs some inline commentary before we merge it, but I think this is good otherwise pending tests.

src/background_jobs.rs
Turbo87: yeah, I figured if the `priority` was different there might be a reason for duplicating the job 🤷‍♂️

Turbo87:
> needs some inline commentary before we merge it

✅  let me know if this is sufficient or if you need more :)

> pending tests

I've tried to write tests for this, but due to https://github.com/rust-lang/crates.io/blob/9189d1a76196d7ec772baffee5c8500ff45593c5/src/tests/util/test_app.rs#L97-L99 this has proven quite difficult. Without a major refactoring of our test suite I'm not sure how to write proper tests for this :-/

LawnGnome (approved):
> white_check_mark let me know if this is sufficient or if you need more :)

Nope, that looks good to me. Thanks!

Given the test difficulties, I'm happy going ahead with this as is for now — maybe we just create a follow up issue noting that we'd like to have test coverage here, and maybe some bright person will jump in, or one of us will have a sudden brainwave.

Turbo87:
> maybe we just create a follow up issue noting that we'd like to have test coverage here, and maybe some bright person will jump in, or one of us will have a sudden brainwave.

I've added a doc comment to the function, including a warning about the function not having any explicit tests.

Turbo87:
I've just tested this out on staging and subsequently fixed a small calculation bug regarding the `skipped_jobs_count` value, but generally this appears to work as intended now.

PR #6868, opened 2023-07-24
settings/tokens/new: Disable autocomplete for "Name" field

Turbo87:
This should disable password managers from assuming that this input is expecting a person name. Since users shouldn't be using the same token name multiple times the disabled autocomplete shouldn't be a big issue.


Turbo87:
unfortunately, it looks like this did not do the trick. 1password is still trying to autocomplete the field... 😩

PR #6872, opened 2023-07-25
First batch of rate limiter changes

pietroalbini:
This PR contains the first batch of changes needed to implement proper rate limiting for yanking and unyanking:

* It adds a new macro to easily generate postgres enums.
* It adds an `action` column to the rate limit tables. Right now it only supports the `publish-new` action, but in a future PR it'll allow for different kinds of rate limits.
* It renames `PublishRateLimit` to `RateLimiter`, in anticipation for it being able to limit more actions.


src/util.rs
```diff
@@ -32,3 +32,38 @@ impl Maximums {
...
+        #[repr(i32)]
+        $vis enum $name {
+            $($item = $int,)*
+        }
+
+        impl diesel::deserialize::FromSql<diesel::sql_types::Integer, diesel::pg::Pg> for $name {
+            fn from_sql(bytes: diesel::pg::PgValue<'_>) -> diesel::deserialize::Result<Self> {
+                match <i32 as diesel::deserialize::FromSql<diesel::sql_types::Integer, diesel::pg::Pg>>::from_sql(bytes)? {
+                    $($int => Ok(Self::$item),)*
+                    n => Err(format!("unknown value: {}", n).into()),
```
Turbo87, line 55: can we somehow integrate `$name` in the error message to make this easier to debug?

src/util.rs
```diff
@@ -32,3 +32,38 @@ impl Maximums {
         }
     }
 }
+
+#[macro_export]
+macro_rules! pg_enum {
```
Turbo87, line 37: this seems like it would be a better fit for the `crates_io::sql` module instead of the generic `util`

src/schema.rs
```diff
@@ -650,6 +656,12 @@ diesel::table! {
         ///
         /// (Automatically generated by Diesel.)
         expires_at -> Nullable<Timestamp>,
+        /// The `action` column of the `publish_rate_overrides` table.
+        ///
+        /// Its SQL type is `Int4`.
+        ///
+        /// (Automatically generated by Diesel.)
+        action -> Int4,
```
Turbo87, line 664: make sure to adjust the ops guide accordingly once this is merged 😉

src/publish_rate_limit.rs
```diff
@@ -115,6 +105,16 @@ impl PublishRateLimit {
...
+#[derive(Queryable, Insertable, Debug, PartialEq, Clone, Copy)]
+#[diesel(table_name = publish_limit_buckets, check_for_backend(diesel::pg::Pg))]
+#[allow(dead_code)] // Most fields only read in tests
+struct Bucket {
+    user_id: i32,
+    tokens: i32,
+    last_refill: NaiveDateTime,
+    action: LimitedAction,
+}
+
```
Turbo87, line 124: what's the reason for moving the `struct`?

src/schema.rs
pietroalbini: Yep, I'll do so once the second PR has been merged.

src/util.rs
pietroalbini: Done!

pietroalbini:

Turbo87 (approved):

PR #6873, opened 2023-07-25
tarball: Allow lowercase `cargo.toml` files for now

Turbo87:
see https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/.60cargo.2Etoml.60.20on.20case.20insensitive.20filesystems/near/376986823

We have about 50 crate files published from case-insensitive file systems (most likely Windows) that are containing a lowercase `cargo.toml` instead of the regular `Cargo.toml` file.

We don't want to support this case in the future and `cargo` is planning to normalize the filename on publish, but a) the current implementation of cargo produces such artifacts in some cases and b) we have existing crate files that work with cargo on such filesystems. Thus we support this case on the server for now until the cargo normalization has been shipped. Afterwards we will not accept such crate files anymore, but we might still need to keep this code around to be able to process old crate files for backfilling/analysis purposes.

### DIFF
--- a/crates_io_tarball/src/lib.rs
+++ b/crates_io_tarball/src/lib.rs
@@ -56,6 +56,7 @@ pub fn process_tarball<R: Read>(
     let mut vcs_info = None;
 
     let manifest_path = Path::new(&pkg_name).join("Cargo.toml");
+    let manifest_path_lower = Path::new(&pkg_name).join("cargo.toml");
     let mut manifest = None;
 
     for entry in archive.entries()? {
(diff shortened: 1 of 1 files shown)


crates_io_tarball/src/lib.rs
```diff
@@ -87,7 +88,7 @@ pub fn process_tarball<R: Read>(
             let mut contents = String::new();
             entry.read_to_string(&mut contents)?;
             vcs_info = CargoVcsInfo::from_contents(&contents).ok();
-        } else if entry_path == manifest_path {
```
LawnGnome, line 90: Technically, this means that the order we load `Cargo.toml` and `cargo.toml` may be undefined if both are present in a tarball. I don't think this is worth actually doing anything about (it's like an edge case of an edge case at this point), but it does feel worth calling out.

LawnGnome (approved):

crates_io_tarball/src/lib.rs
Turbo87: I had actually thought about that, but figured it's currently not worth addressing it. I have some more plans with this package though, so in the future I might fix it :)

PR #6875, opened 2023-07-25
Add application-level rate limits for crate updates, yanking and unyanking

pietroalbini:
This PR follows up on #6872 by adding support for multiple kinds of rate limits in `RateLimiter` and adding a limit for yanking and unyanking, which would've prevented last Friday's outage.

As I was already changing this code, I also moved the rate limit for publishing existing crates from nginx into the application. This makes the limit more precise (as it's not per-server anymore, potentially allowing up to 4x the limit), and allows overriding the limit just for a subset of users.

With this PR, the rate limits are:

| Action | Burst amount | Refill time |
| --- | --- | --- |
| Publish new crates | 5 | 1 every 10 minutes |
| Publish updates to existing crates | 30 | 1 every minute |
| Yanking or unyanking | 100 | 1 every minute |

Note that the environment variables also changed:

* From `WEB_NEW_PKG_RATE_LIMIT_RATE_MINUTES` to `RATE_LIMITER_PUBLISH_NEW_RATE_SECONDS`
* From `WEB_NEW_PKG_RATE_LIMIT_BURST` to `RATE_LIMITER_PUBLISH_NEW_BURST`

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -2,7 +2,7 @@ use anyhow::{anyhow, Context};
 use ipnetwork::IpNetwork;
 use oauth2::{ClientId, ClientSecret};
 
-use crate::rate_limiter::RateLimiter;
+use crate::rate_limiter::{LimitedAction, RateLimiterConfig};
 use crate::{env, env_optional, Env};
 
 use super::base::Base;
@@ -10,7 +10,7 @@ use super::database_pools::DatabasePools;
 use crate::config::balance_capacity::BalanceCapacityConfig;
 use crate::storage::StorageConfig;
 use http::HeaderValue;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::net::IpAddr;
 use std::time::Duration;
 
@@ -30,7 +30,7 @@ pub struct Server {
     pub gh_client_secret: ClientSecret,
     pub max_upload_size: u64,
     pub max_unpack_size: u64,
-    pub rate_limiter: RateLimiter,
+    pub rate_limiter: HashMap<LimitedAction, RateLimiterConfig>,
     pub new_version_rate_limit: Option<u32>,
     pub blocked_traffic: Vec<(String, Vec<String>)>,
     pub max_allowed_page_offset: u32,
@@ -140,6 +140,24 @@ impl Default for Server {
             .map(|s| s.parse().expect("SERVER_THREADS was not a valid number"))
             .ok();
 
+        // Dynamically load the configuration for all the rate limiting actions. See
+        // `src/rate_limiter.rs` for their definition.
+        let mut rate_limiter = HashMap::new();
+        for action in LimitedAction::VARIANTS {
+            let env_var_key = action.env_var_key();
+            rate_limiter.insert(
+                *action,
+                RateLimiterConfig {
+                    rate: Duration::from_secs(
+                        env_optional(&format!("RATE_LIMITER_{env_var_key}_RATE_SECONDS"))
+                            .unwrap_or_else(|| action.default_rate_seconds()),
+                    ),
+                    burst: env_optional(&format!("RATE_LIMITER_{env_var_key}_BURST"))
+                        .unwrap_or_else(|| action.default_burst()),
+                },
+            );
+        }
+
         Server {
             db: DatabasePools::full_from_environment(&base),
             storage: StorageConfig::from_environment(),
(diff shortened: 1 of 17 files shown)


Turbo87:
I was going to rebase the PR for you, but apparently you didn't check the "Maintainers are allowed to edit this pull request." checkbox... 😅

pietroalbini:
Rebased.

src/config/server.rs
```diff
@@ -140,6 +140,24 @@ impl Default for Server {
...
+        // Dynamically load the configuration for all the rate limiting actions. See
+        // `src/rate_limiter.rs` for their definition.
+        let mut rate_limiter = HashMap::new();
+        for action in LimitedAction::VARIANTS {
+            let env_var_key = action.env_var_key();
+            rate_limiter.insert(
+                *action,
+                RateLimiterConfig {
+                    rate: Duration::from_secs(
+                        env_optional(&format!("RATE_LIMITER_{env_var_key}_RATE_SECONDS"))
```
Turbo87, line 152: it looks like previously we were using `WEB_NEW_PKG_RATE_LIMIT_RATE_MINUTES`. so I guess this is a breaking change? would've been good to mention that in the PR description 😉

src/rate_limiter.rs
```diff
@@ -12,25 +12,29 @@ use std::time::Duration;
...
         match self {
             LimitedAction::PublishNew => 5,
+            LimitedAction::PublishUpdate => 30,
         }
     }
 
     pub fn env_var_key(&self) -> &'static str {
         match self {
             LimitedAction::PublishNew => "PUBLISH_NEW",
+            LimitedAction::PublishUpdate => "PUBLISH_EXISTING",
```
Turbo87, line 40: why not `PUBLISH_UPDATE` to keep it consistent? :)

src/tests/krate/publish.rs
```diff
@@ -1062,6 +1061,67 @@ fn publish_rate_limit_doesnt_affect_existing_crates() {
...
+    let crate_to_publish = PublishBuilder::new("rate_limited1").version("1.0.1");
+    token.publish_crate(crate_to_publish).good();
+
+    let json = anon.show_crate("rate_limited1");
+    assert_eq!(json.krate.max_version, "1.0.1");
+    assert_eq!(app.stored_files().len(), 3);
+
+    // Uploading the second update to the crate is rate limited
+    let crate_to_publish = PublishBuilder::new("rate_limited1").version("1.0.2");
+    token.publish_crate(crate_to_publish).assert_rate_limited();
```
Turbo87, line 1088: could you add assertions about the returned JSON payload?

something like:

```rs
    assert_eq!(
        response.into_json(),
        json!({ "errors": [{ "detail": "foo" }] })
    );
```

src/rate_limiter.rs
```diff
@@ -1,52 +1,93 @@
...
+            LimitedAction::PublishUpdate => 60,
+            LimitedAction::YankUnyank => 60,
+        }
+    }
+
+    pub fn default_burst(&self) -> i32 {
+        match self {
+            LimitedAction::PublishNew => 5,
+            LimitedAction::PublishUpdate => 30,
+            LimitedAction::YankUnyank => 30,
```
Turbo87, line 33: we talked about this PR in the team meeting on Friday and felt that the burst limit for yanks might be a little on the low side. could you change this to 100? :)

```suggestion
            LimitedAction::YankUnyank => 100,
```

src/config/server.rs
pietroalbini: Sure, the environment variable is not in use right now though.

src/rate_limiter.rs
pietroalbini: Good call, will change it.

src/rate_limiter.rs
pietroalbini: Will update it.

src/tests/krate/publish.rs
pietroalbini: Done!

pietroalbini:

Turbo87 (approved):

Turbo87:
thanks! :)

Turbo87:
```
thread 'krate::publish::publish_new_crate_rate_limited' panicked at 'assertion failed: `(left == right)`
  left: `429`,
 right: `200`', src/tests/krate/publish.rs:1024:10
```

it looks like one of the tests is a little flaky :-/

PR #6884, opened 2023-07-26
Implement basic smoke test setup

Turbo87:
This PR adds a new `crates_io_smoke_test` project to the workspace and a "Smoke Test" GitHub Actions workflow. The latter can be manually triggered and runs the former. The new tool will query the current state of the staging environment, run `cargo new`, adjust the `version` field and then run `cargo publish` to see if publishing works correctly.

Example: https://github.com/Turbo87/crates.io/actions/runs/5670151138/job/15364394197

Note that this is just a first step. In the next step we can `cargo new` another project with a dependency on the published version to check that downloads are working as intended.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -1,0 +1,21 @@
+name: Smoke Test
+
+on:
+  workflow_dispatch:
+
+concurrency:
+  group: smoke-test
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  test:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3.5.3
+      - uses: Swatinem/rust-cache@v2.5.1
+      - run: cargo build --manifest-path crates_io_smoke_test/Cargo.toml
+      - run: cargo run --manifest-path crates_io_smoke_test/Cargo.toml --quiet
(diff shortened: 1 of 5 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,23 @@
...
+  test:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3.5.3
+      - uses: Swatinem/rust-cache@v2.5.1
+      - run: cargo build --manifest-path crates_io_smoke_test/Cargo.toml
+      - run: cargo run --manifest-path crates_io_smoke_test/Cargo.toml --quiet
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.STAGING_SMOKE_TEST_TOKEN }}
```
Turbo87, line 23: @rust-lang/infra to make this work we need to add a repository secret with a staging API token, scoped to `publish-update` and the `crates-staging-test-tb` crate.

should I generate such a token myself? do you want to do that? should I invite any users to my test crate?

crates_io_smoke_test/src/main.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    {
+        let readme_path = project_path.join("README.md");
+        info!(readme_path = %readme_path.display(), "Creating `README.md` file…");
+        let mut readme_file =
+            File::create(readme_path).context("Failed to open `README.md` file")?;
+
+        let new_content = format!(
+            "# {} v{}\n\n![](https://media1.giphy.com/media/Ju7l5y9osyymQ/200.gif)\n",
```
LawnGnome, line 123: :laughing:

LawnGnome (approved):
Looks a good starting point to me! :+1:

PR #6897, opened 2023-07-28
tarball: Fail if `Cargo.toml` manifest is missing or invalid

Turbo87:
This changes the `crates_io_tarball` package to a) require a `Cargo.toml` file to always be present in a crate file, and b) no longer silently fail if the `Cargo.toml` file can't be parsed.

This ensures that only crate files with valid manifests end up on crates.io. We previously accepted almost arbitrary tarballs, even without any manifests. Note that "valid" is still pretty vague since we currently only deserialize very few fields of the TOML file, but this will change in the near future.

Unfortunately this change also required a few changes to our test suite, which previously did not include `Cargo.toml` files in most of the tarballs sent to the server. The `Cargo.toml` that we send now is a veeeery basic manifest and is actually still lacking the required `name` and `version` fields, but, since we don't enforce this part yet, are left out for now.

### DIFF
--- a/crates_io_tarball/src/lib.rs
+++ b/crates_io_tarball/src/lib.rs
@@ -20,7 +20,7 @@ mod vcs_info;
 
 #[derive(Debug)]
 pub struct TarballInfo {
-    pub manifest: Option<Manifest>,
+    pub manifest: Manifest,
     pub vcs_info: Option<CargoVcsInfo>,
 }
 
@@ -32,6 +32,10 @@ pub enum TarballError {
     InvalidPath(String),
     #[error("unexpected symlink or hard link found: {0}")]
     UnexpectedSymlink(String),
+    #[error("Cargo.toml manifest is missing")]
+    MissingManifest,
+    #[error("Cargo.toml manifest is invalid: {0}")]
+    InvalidManifest(#[source] toml::de::Error),
     #[error(transparent)]
     IO(#[from] std::io::Error),
 }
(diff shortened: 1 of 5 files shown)


Turbo87:
Note that after this change all the mentioned crate files from https://gist.github.com/Turbo87/366fc2555db2aa5d37d71bdff369a45c (except for lowercase `cargo.toml`, see #6873) and the ones from https://gist.github.com/Turbo87/2956337ab735fa2a5adcd6fdc3d5b470 will no longer be usable with `crates_io_tarball`.

I ran the `check_all_crates` example on a recent download of https://github.com/dtolnay/get-all-crates and the above is basically the result :)

crates_io_tarball/src/lib.rs
```diff
@@ -93,10 +97,14 @@ pub fn process_tarball<R: Read>(
             // erroring if it cannot be read.
             let mut contents = String::new();
             entry.read_to_string(&mut contents)?;
-            manifest = toml::from_str(&contents).ok();
+            manifest = Some(toml::from_str(&contents).map_err(TarballError::InvalidManifest)?);
```
epage, line 100: Something we'll have to keep in mind as we add new features is to determine if crates-io's validation code needs updating.

epage: (to be clear, atm it is very minimal so its less likely)

Turbo87: yep, definitely!

PR #6900, opened 2023-07-30
Migrate from docs.rs' builds API to status API

Nemo157:
To simplify refactoring how our builds are processed I want to remove the `/builds.json` endpoint on docs.rs. As a replacement [I added](https://github.com/rust-lang/docs.rs/pull/2147) a new endpoint simply giving the status that sites like crates.io wants: are there docs for this version available to link to. This also fixes some small issues with the old endpoint (some edgecases meant the first build was successful but there were no docs, or there were docs but the first build was marked unsuccessful; if the version in the path doesn't match an exact version it is used as a semver query, but using a `=version` path like done here on the old API would hit CORS issues).

### DIFF
--- a/app/models/version.js
+++ b/app/models/version.js
@@ -133,13 +133,13 @@ export default class Version extends Model {
     }
   });
 
-  loadDocsBuildsTask = task(async () => {
-    return await ajax(`https://docs.rs/crate/${this.crateName}/${this.num}/builds.json`);
+  loadDocsStatusTask = task(async () => {
+    return await ajax(`https://docs.rs/crate/${this.crateName}/=${this.num}/status.json`);
   });
 
   get hasDocsRsLink() {
-    let docsBuilds = this.loadDocsBuildsTask.lastSuccessful?.value;
-    return docsBuilds?.[0]?.build_status === true;
+    let docsStatus = this.loadDocsStatusTask.lastSuccessful?.value;
+    return docsStatus?.doc_status === true;
   }
 
   get docsRsLink() {
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -54,7 +54,7 @@ export default class VersionRoute extends Route {
 
     let { crate, version } = model;
     if (!crate.documentation || crate.documentation.startsWith('https://docs.rs/')) {
-      version.loadDocsBuildsTask.perform().catch(error => {
+      version.loadDocsStatusTask.perform().catch(error => {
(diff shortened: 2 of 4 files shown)


app/models/version.js
```diff
@@ -133,13 +133,13 @@ export default class Version extends Model {
     }
   });
 
-  loadDocsBuildsTask = task(async () => {
-    return await ajax(`https://docs.rs/crate/${this.crateName}/${this.num}/builds.json`);
+  loadDocsStatusTask = task(async () => {
+    return await ajax(`https://docs.rs/crate/${this.crateName}/=${this.num}/status.json`);
```
Turbo87, line 137: Bildschirmfoto 2023-07-31 um 09 56 42

it looks like the URL with `=` just redirects to the same URL without the `=`. are you sure that this is necessary?

Turbo87: alright, I just had a quick look at the linked PR. looks like `=1.2.3` will return 404 if the version does not exist and redirect to `1.2.3` if the version exists. then `1.2.3` will look for `1.2.3` and if that does not exist will try `^1.2.3`, but since we know from the first call that it exists it won't interpret it as semver.

I'm not a huge fan of this API decision, but I guess we have to live with it now :D

Turbo87 (approved):

app/models/version.js
Nemo157: It only redirects if that exact version is known about, without the `=` it does a semver match and will redirect to a different version if one is available. As an example <https://docs.rs/crate/stylish-core/0.1.0/status.json> redirects to <https://docs.rs/crate/stylish-core/0.1.1/status.json> (because of reasons around circular dependencies 0.1.0 was never added to the docs.rs database).

This is existing behavior for all the endpoints, including `/builds.json`, but that did not get CORS headers applied when it returned a redirect making it still get treated as an error, e.g. on <https://crates.io/crates/stylish-core/0.1.0>

> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://docs.rs/crate/stylish-core/0.1.0/builds.json. (Reason: CORS request did not succeed). Status code: (null).

Since that is also fixed on this endpoint, it would link to the 0.1.1 docs without the `=`.

Turbo87: yeah, I understand that this is just legacy behavior we have to live with now. it's fine, was just a bit surprising to see that we need the extra ping-pong for exact version queries. maybe it would make sense for the JSON endpoints to return the content directly for `=1.2.3` queries instead of redirecting?

PR #6953, opened 2023-08-09
Revert "Update dependency node to v18.17.1"

LawnGnome:
Reverts rust-lang/crates.io#6951.

Obviously needs to wait for the buildpack to be updated. My bad.


Turbo87:
hmm, interesting. normally, this isn't a problem with Heroku, but apparently this version had issues 😅

LawnGnome:
I think they're just behind on getting the buildpack updated. Looks like it's being built now, so I'll try the version upgrade again later today or tomorrow.

PR #6958, opened 2023-08-10
Replace hardcoded credentials path with cargo docs link

federicomenaquintero:
https://crates.io/settings/tokens incorrectly says `~/.cargo/credentials`, it's missing the `.toml` in the filename.

### DIFF
--- a/app/components/settings/api-tokens.hbs
+++ b/app/components/settings/api-tokens.hbs
@@ -25,7 +25,8 @@
 
 <p local-class="explainer">
   To use an API token, run <a href="https://doc.rust-lang.org/cargo/commands/cargo-login.html"><code>cargo login</code></a>
-  on the command line and paste the key when prompted. This will save it to a local file (<code>~/.cargo/credentials</code> by default).
+  on the command line and paste the key when prompted. This will save it to a
+  <a href="https://doc.rust-lang.org/cargo/reference/config.html#credentials">local credentials file</a>.
   For CI systems you can use the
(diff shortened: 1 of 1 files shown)


Turbo87:
for context: https://github.com/rust-lang/crates.io/pull/5730

LawnGnome:
OK, so if I'm reading the history correctly:

* Cargo added support for `credentials.toml` (with the extension) in 1.39.0 (November 2019).
* Cargo has created `credentials.toml` (with the extension) by default since 1.68.0 (March this year).

I guess my question is which of `credentials` and `credentials.toml` wins — if a user has a pre-existing `credentials`, and we tell them to add things to `credentials.toml`, is that going to break things?

Feels like it might be a little soon to make this change, but I'm interested in if anyone on the @rust-lang/cargo team has an opinion here?

weihanglo:
> I guess my question is which of `credentials` and `credentials.toml` wins — if a user has a pre-existing `credentials`, and we tell them to add things to `credentials.toml`, is that going to break things?

Nothing gonna be broken. `credentials.toml` would be ignored in that case.

Maybe just say it being saved in a local credential file and link to where [The Cargo Book mentions it](https://doc.rust-lang.org/cargo/reference/config.html#credentials)?


BTW just created <https://github.com/rust-lang/cargo/pull/12479#issue-1847111018> to clarify the situation. Thanks for the summary @LawnGnome :)

Turbo87:
> Maybe just say it being saved in a local credential file and link to where [The Cargo Book mentions it](https://doc.rust-lang.org/cargo/reference/config.html#credentials)?

that seems like a good idea to me. @federicomenaquintero could you amend the pull request accordingly? :)

PR #6961, opened 2023-08-11
PublishNew was allowing 1 new crate an hour rather than every 10 minutes

carols10cents:
In https://github.com/rust-lang/crates.io/pull/6875, @pietroalbini said:

> With this PR, the rate limits are:
> 
> | Action | Burst amount | Refill time |
> | --- | --- | --- |
> | Publish new crates | 5 | 1 every 10 minutes |
> | Publish updates to existing crates | 30 | 1 every minute |
> | Yanking or unyanking | 100 | 1 every minute |

but the code didn't match that intention, which explains why we were getting so many emails asking for rate limit increases since that PR went out.

The problem was that the `default_rate_seconds` for `PublishNew` was set to `60 * 60`, aka 1 hour, when the intention was that it would be set to `10 * 60`, or 10 minutes.

I added a new test, and for `PublishNew` it fails with:

```
---- rate_limiter::tests::default_rate_limits stdout ----
thread 'rate_limiter::tests::default_rate_limits' panicked at 'assertion failed: `(left == right)`
  left: `[2023-08-11T19:15:38, 2023-08-11T19:25:38, 2023-08-11T19:35:38, 2023-08-11T19:45:38, 2023-08-11T19:55:38, 2023-08-11T20:05:38, 2023-08-11T20:15:38, 2023-08-11T20:25:38, 2023-08-11T20:35:38, 2023-08-11T20:45:38]`,
 right: `[2023-08-11T19:05:38, 2023-08-11T19:05:38, 2023-08-11T19:05:38, 2023-08-11T19:05:38, 2023-08-11T19:05:38, 2023-08-11T20:05:38, 2023-08-11T20:05:38, 2023-08-11T20:05:38, 2023-08-11T20:05:38, 2023-08-11T20:05:38]`', src/rate_limiter.rs:212:9
```

because the `last_refill` times get set to the time at which you're publishing if you're not rate limited.

The CI run for the first commit shows the failure, then the second commit shows the test passing (modulo the flaky test).

### DIFF
--- a/src/rate_limiter.rs
+++ b/src/rate_limiter.rs
@@ -177,6 +177,74 @@ mod tests {
     use crate::test_util::*;
 
     #[test]
+    fn default_rate_limits() -> QueryResult<()> {
+        let conn = &mut pg_connection();
+        let now = now();
+
+        // Set the defaults as if no env vars have been set in production
+        let mut rate_limiter = HashMap::new();
+        for action in LimitedAction::VARIANTS {
+            rate_limiter.insert(
+                *action,
+                RateLimiterConfig {
+                    rate: Duration::from_secs(action.default_rate_seconds()),
+                    burst: action.default_burst(),
+                },
+            );
+        }
+        let rate = RateLimiter::new(rate_limiter);
+
+        let user_id = new_user_bucket(conn, 5, now)?.user_id;
+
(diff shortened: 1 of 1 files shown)


pietroalbini (approved):
Whelp, good catch!

PR #6910, opened 2023-08-01
Optimize reverse dependencies query

paolobarbolini:
This optimizes the reverse dependencies query by saving the output of `to_semver_no_prerelease(num)` into a `GENERATED` column and adding indexes. On an Hetzer CAX21 server querying reverse dependencies for `serde` now takes 600ms.

<details>
<summary>Initial attempt at optimizing the query</summary>

This optimizes the reverse dependencies query by using [Common Table Expression (CTE) Materialization](https://www.postgresql.org/docs/current/queries-with.html#id-1.5.6.12.7) as a method of steering the query optimizer into the desired direction.

This makes http://localhost:8888/api/v1/crates/serde/reverse_dependencies?page=1&per_page=10 respond in 3 seconds, which is much nicer than having it reach the 10 seconds timeout as it did before.

---

While trying to make the query faster on PG 16beta2 I also created two indexes but they don't seem to be helping here so I left them out of migrations for now.
I'll save them here for later:

```sql
CREATE INDEX idx1 ON versions (id, crate_id);
CREATE INDEX idx2 ON versions (crate_id, to_semver_no_prerelease(num) DESC NULLS LAST, id) WHERE NOT yanked;
```

```sql
WITH crate_versions_crate_ids AS MATERIALIZED (
    SELECT versions.crate_id
    FROM versions
    INNER JOIN dependencies
      ON dependencies.version_id = versions.id
    WHERE dependencies.crate_id = $1
    GROUP BY versions.crate_id
), all_versions AS MATERIALIZED (
    SELECT
        id,
        crate_id
    FROM (
        SELECT
            versions.id,
            versions.crate_id,
            row_number() OVER (
                PARTITION BY crate_id
                ORDER BY to_semver_no_prerelease(num) DESC NULLS LAST
            ) rn
        FROM versions
        WHERE NOT versions.yanked
        -- This is completely redundant, but it's faster to filter the versions
        -- early even if this subselect is done via an index scan.
        AND versions.crate_id IN (
            SELECT crate_id
            FROM crate_versions_crate_ids
        )
    ) v
    WHERE rn = 1
)

-- Apply pagination to the whole thing
SELECT *, COUNT(*) OVER () as total FROM (
    -- Multple dependencies can exist, make it distinct
    SELECT DISTINCT ON (crate_downloads, crate_name)
    dependencies.*,
    crates.downloads AS crate_downloads,
    crates.name AS crate_name
    FROM dependencies
[...]

PR #6966, opened 2023-08-13
Improve correctness of the reverse dependencies query

paolobarbolini:
These commits have been branched out of #6910 after I kept finding improvements I could make which weren't performance related.

The last commit actually gives the reverse dependencies query a performance boost by allowing Postgres to write a query plan parallelizing most of the query to parallel worker processes, which it didn't before because of [parallel safety](https://www.postgresql.org/docs/current/parallel-safety.html).
I put it in because while it is a performance improvement it also seemed like a flag which should be present everytime the function implementation allows for it.
It also helps speed-up refreshing of indexes or materialized views.

### DIFF
--- a/migrations/2023-08-13-175220_to_semver_no_prerelease_parallel_safe/up.sql
+++ b/migrations/2023-08-13-175220_to_semver_no_prerelease_parallel_safe/up.sql
@@ -1,0 +1,10 @@
+DROP FUNCTION to_semver_no_prerelease(text);
+
+CREATE FUNCTION to_semver_no_prerelease(text) RETURNS semver_triple IMMUTABLE PARALLEL SAFE AS $$
+  SELECT (
+    split_part($1, '.', 1)::numeric,
+    split_part($1, '.', 2)::numeric,
+    split_part(split_part($1, '+', 1), '.', 3)::numeric
+  )::semver_triple
+  WHERE strpos($1, '-') = 0
+  $$ LANGUAGE SQL SET search_path = public;
--- a/migrations/2023-08-13-175220_to_semver_no_prerelease_parallel_safe/up.sql
+++ b/migrations/2023-08-13-175220_to_semver_no_prerelease_parallel_safe/up.sql
@@ -1,10 +1,10 @@
 DROP FUNCTION to_semver_no_prerelease(text);
 
-CREATE FUNCTION to_semver_no_prerelease(text) RETURNS semver_triple IMMUTABLE PARALLEL SAFE AS $$
+CREATE FUNCTION to_semver_no_prerelease(text) RETURNS public.semver_triple IMMUTABLE PARALLEL SAFE AS $$
   SELECT (
     split_part($1, '.', 1)::numeric,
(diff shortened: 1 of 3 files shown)


migrations/2023-08-13-175220_to_semver_no_prerelease_parallel_safe/up.sql
```diff
@@ -0,0 +1,17 @@
...
+);
+
+CREATE FUNCTION to_semver_no_prerelease(text) RETURNS semver_triple IMMUTABLE PARALLEL SAFE AS $$
+  SELECT (
+    split_part($1, '.', 1)::numeric,
+    split_part($1, '.', 2)::numeric,
+    split_part(split_part($1, '+', 1), '.', 3)::numeric
+  )::semver_triple
+  WHERE strpos($1, '-') = 0
+  $$ LANGUAGE SQL;
```
LawnGnome, line 17: This isn't strictly related to what you're doing, but while we're in here, we should also add:

```suggestion
  $$ LANGUAGE SQL SET search_path = public;
```

Right now this works by accident: the `schema.sql` generated when we dump the database each night is actually subtly wrong, but PostgreSQL doesn't check the actual return type of the function definition unless it's used somewhere, which it's not when importing the data dump. However, if and when we add a generated column (as in #6910), this will start getting used, and an error occurs when importing the schema because the return type isn't `public.semver_triple`.

migrations/2023-08-13-175220_to_semver_no_prerelease_parallel_safe/up.sql
```diff
@@ -0,0 +1,17 @@
+DROP FUNCTION to_semver_no_prerelease(text);
+DROP TYPE semver_triple;
+
+CREATE TYPE semver_triple AS (
+  major numeric,
+  minor numeric,
+  teeny numeric
+);
```
LawnGnome, line 8: Do we need to recreate the type?

LawnGnome (approved):
LGTM (other than one technically unrelated cleanup I'd like to make while we're in here). Thanks!

To state the obvious (as implied by the migration numbers): we would need to merge this before #6910.

migrations/2023-08-13-175220_to_semver_no_prerelease_parallel_safe/up.sql
paolobarbolini: No reason in particular. I copied it from a different migration that did that. Now it's fixed.

migrations/2023-08-13-175220_to_semver_no_prerelease_parallel_safe/up.sql
paolobarbolini: CI doesn't seem to like it

LawnGnome: It doesn't! Sorry about that.

I hope you don't mind — I've pushed a commit that should fix it. (Easier than trying to do it through GitHub code review suggestions.)

PR #7008, opened 2023-08-22
crates_io_tarball: error if multiple manifests are found

### LINKED ISSUE #6978: Reject crate files with multiple Cargo.toml manifests of different casing
For the time being, we'd only want to apply this to new publishes, not retroactively to crate files that have already been published.

LawnGnome:
This will handle both differences in case (ie a crate file that contains both `package-x.y.z/Cargo.toml` and `package-x.y.z/cargo.toml`, as may happen on a case-sensitive filesystem) and malformed tarballs that contain multiple copies of the same file.

Effort has been made to otherwise preserve the same semantics as `cargo` where, if there is only one manifest, only `Cargo.toml` and `cargo.toml` are accepted. (It's debatable if we should even accept `cargo.toml`, and recent versions of `cargo` won't allow crates to be published with that casing, but since there may be legacy setups out there that are still publishing crates with this variation and we've accepted them in the past, we may as well preserve that behaviour for now.)

In practice, this is what the user will see from `cargo publish` if there are multiple manifests:

image

And what they'll see if they're publishing from a case insensitive filesystem and have a weirdly cased `CaRgO.tOmL`:

image

Fixes #6978.

### DIFF
--- a/crates_io_tarball/src/lib.rs
+++ b/crates_io_tarball/src/lib.rs
@@ -92,6 +92,8 @@ pub fn process_tarball<R: Read>(
             ));
         }
 
+        // Let's go hunting for the VCS info and crate manifest. The only valid place for these is
+        // in the package root in the tarball.
         if entry_path.parent() == Some(pkg_root) {
             let entry_file = entry_path.file_name().unwrap_or_default();
             if entry_file == ".cargo_vcs_info.json" {
--- a/crates_io_tarball/src/lib.rs
+++ b/crates_io_tarball/src/lib.rs
@@ -9,8 +9,9 @@ use crate::manifest::validate_manifest;
 pub use crate::vcs_info::CargoVcsInfo;
 pub use cargo_toml::Manifest;
 use flate2::read::GzDecoder;
+use std::collections::HashMap;
 use std::io::Read;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use tracing::instrument;
 
 #[cfg(any(feature = "builder", test))]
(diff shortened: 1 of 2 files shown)


crates_io_tarball/src/lib.rs
```diff
@@ -89,25 +92,46 @@ pub fn process_tarball<R: Read>(
...
-            let mut contents = String::new();
-            entry.read_to_string(&mut contents)?;
-            vcs_info = CargoVcsInfo::from_contents(&contents).ok();
-        } else if entry_path == manifest_path || entry_path == manifest_path_lower {
-            // Try to extract and read the Cargo.toml from the tarball, silently
-            // erroring if it cannot be read.
-            let mut contents = String::new();
-            entry.read_to_string(&mut contents)?;
-            manifest = Some({
+        if entry_path.parent() == Some(pkg_root) {
```
Turbo87, line 95: this should already be covered by the `if !entry_path.starts_with(pkg_name)` check above, right? admittedly with a more questionable condition though :D

crates_io_tarball/src/lib.rs
```diff
@@ -89,25 +92,46 @@ pub fn process_tarball<R: Read>(
...
+            let entry_file = entry_path.file_name().unwrap_or_default();
+            if entry_file == ".cargo_vcs_info.json" {
```
Turbo87, line 97: this makes it apply to a `foo-1.2.3/a/b/c/.cargo_vcs_info.json` file too, while previously we were only reading the top level file

Turbo87 (approved):
LGTM 👍

PR #7048, opened 2023-08-30
Update the data access page on crates.io

LawnGnome:
The motivation for this is a couple of recent requests and discussions where it's been clear that (a) we need to document the sparse index as an option, and (b) it's unclear if the crawling limits apply to regular API usage as well.

This integrates the sparse index into the data access recommendations, and makes it clearer that the limits described in the crawling policy also apply to regular API usage.

Cosmetically, the page is now less dense, with some new verbiage particularly around index and API usage.

There should be no change in the actual expectations we have of users in this commit; this is simply trying to more accurately codify what our expectations really are.


Turbo87 (approved):
👍

Issue #6952, opened 2023-08-09
Fix flaky rate limiter test


LawnGnome:
```
thread 'krate::publish::publish_new_crate_rate_limited' panicked at 'assertion failed: `(left == right)`
  left: `429`,
 right: `200`', src/tests/krate/publish.rs:1024:10
```

it looks like one of the tests is a little flaky :-/

_Originally posted by @Turbo87 in https://github.com/rust-lang/crates.io/issues/6875#issuecomment-1659999388_

nic-hartley:
Running this test alone, it seems to always work. (Hard to verify a negative, but just anecdotally it hasn't failed on me yet.) I feel pretty confident diagnosing the issue as, predictably, the hardcoded time limit sometimes naturally expiring. This is supported by seeing things like `service=737ms` in the logs of failing runs but `service=9ms` when it succeeds. I know, I'm just as shocked as you are that the root cause of the flaky ratelimit test appears to be a timing issue. Who could have predicted this.

My pitch for fixing this is to split this test into two separate ones to cover the same functionality:

- Insert a row into `publish_rate_overrides` for a ratelimit that `expires_at` sometime far into the future (like, now + 10 years). Then hit the endpoint and assert it 429s with the correct try again time.
- Insert a row into `publish_rate_overrides` that "just expired" (e.g. now - 1 second). Then hit the endpoint and assert it 200s.

The upside is this is basically impossible to fail spuriously, since `chrono::NaiveDateTime` doesn't account for DST/leap seconds/etc. The downside is the test stops being a straightforward API hit and starts reaching into the guts of the implementation instead. Right now it's *actually getting ratelimited*, and then waiting for that ratelimit to actually expire. This change would make assumptions about functionality. That said, the odds of subtle changes that don't break the test but do break the behavior seems... unlikely? 

If that sounds good to y'all I can put in a PR.

LawnGnome:
That approach sounds reasonable to me — I'd gotten about as far as your analysis (sorry, I obviously forgot to write it up), but hadn't really figured out how to deal with it yet. I think those two tests would, as you said, basically exercise the actual behaviour we want to test in a way that should prevent regressions.

I'd be happy to review a PR from you!

nic-hartley:
Hm, I'm realizing now I misread the code the first time, I thought it referenced the same table twice but it doesn't, whoops. Correct me if I'm wrong but it looks like the code is:

[...]

PR #7051, opened 2023-08-31
Allow removing deleted organizations from crate owners

### LINKED ISSUE #1818: Can't remove deleted organizations from crate owners
Title. Also see similar? issue #1512, which the author closed.

The error is

```
error: failed to remove owners from crate

Caused by:
  api errors (status 404 Not Found): Not Found
```

A workaround exists, which is to recreate the organization and team, which lets you delete it.

*But* that may also be a potential security issue, what if someone else creates it and maliciously abuses their newly granted publish privileges? What about for organization renames?

nic-hartley:
Fixes #1818.

My intended solution here is:

- Add a parameter to decide how the team gets looked up (in the database: `true`, vs. in the GH API: `false`)
- Pass `true` when deleting owners and `false` when adding them

Unfortunately the `Crate::owner_add` and `Crate::owner_remove` methods are like four steps above the actual DB lookup, and they call into the exact same stack of methods to get an `Owner`. I could have also created a new stack of methods but this way seems like it'll require modifying less code, which is a big positive for someone who doesn't really understand how this project works.

That said, I'm not 100% confident in where I've put the branch. Right now it's right next to the error message itself but it might belong higher up, e.g. maybe in `Team::create_or_update` it should try to look up the team by name instead of passing it down to the lower-level code.

Todo:

- [x] Get `cargo test` working
- [x] Add the parameter to the call sites in `owner_add` and `owner_remove`, all the way down to `Team::create_or_update_github_team`, with a `todo!` in place of the actual implementation
- [x] Reconsider whether it should be that deep down, or whether looking up `Owner`s in the database should be higher up.
- [x] Add/modify tests as necessary to ensure you can delete non-existent org owners
- [x] Implement the `in_db` lookup branch, wherever that should terminate

### DIFF
--- a/src/controllers/krate/owners.rs
+++ b/src/controllers/krate/owners.rs
@@ -141,7 +141,7 @@ fn modify_owners(
             msgs.join(",")
         } else {
             for login in &logins {
-                krate.owner_remove(app, conn, user, login)?;
+                krate.owner_remove(conn, login)?;
             }
             if User::owning(&krate, conn)?.is_empty() {
                 return Err(cargo_err(
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -397,14 +397,8 @@ impl Crate {
         }
     }
 
-    pub fn owner_remove(
-        &self,
-        app: &App,
-        conn: &mut PgConnection,
-        req_user: &User,
-        login: &str,
-    ) -> AppResult<()> {
-        let owner = Owner::find_or_create_by_login(app, conn, req_user, login)?;
+    pub fn owner_remove(&self, conn: &mut PgConnection, login: &str) -> AppResult<()> {
+        let owner = Owner::find_by_login(conn, login)?;
 
         let target = crate_owners::table.find((self.id(), owner.id(), owner.kind()));
         diesel::update(target)
--- a/src/models/owner.rs
+++ b/src/models/owner.rs
@@ -1,8 +1,8 @@
 use diesel::pg::Pg;
 use diesel::prelude::*;
 
-use crate::app::App;
 use crate::util::errors::{cargo_err, AppResult};
+use crate::{app::App, schema::teams};
 
 use crate::models::{Crate, Team, User};
 use crate::schema::{crate_owners, users};
@@ -59,6 +59,7 @@ impl Owner {
     /// up-to-date GitHub ID. Fails out if the user isn't found in the
     /// database, the team isn't found on GitHub, or if the user isn't a member
     /// of the team on GitHub.
+    ///
     /// May be a user's GH login or a full team name. This is case
     /// sensitive.
     pub fn find_or_create_by_login(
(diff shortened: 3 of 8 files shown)


nic-hartley:
I'd appreciate some feedback. I've got a few ways I could do this and I'm not sure which is the best.

- Unconditionally look up the `login` in the database before trying to hit the remote API and return it from the DB if it exists.

  I like this from a cleanliness point of view, but we'd end up in an issue where you can add a dead org as an owner. It still wouldn't do anything, but it might be confusing if you try to create a new org under a dead one's name, successfully add an org with the same name, and still can't use it to auth. (But the only way to solve this would be something that purges dead orgs from the DB.)

- Pass a boolean down the call stack to decide whether to try looking it up in the DB (`true` for `owner_remove`, `false` for `owner_add`)

  This gets annoying to implement, and I'm not sure where in the call stack the boolean should be checked. I'm leaning towards `Team::create_or_update` but not for any concrete reasons I can explain.

- Create a new function, e.g. `Owner::lookup` (please bikeshed the name, not sure what the convention is) that *only* looks it up in the DB.

  This probably adds the most code but it separates things into two completely separate call stacks, which seems nice. I don't think we'd run into any bugs from this approach as any owner being removed (which is the only time you'd call `lookup`) would have already been added to the DB by create, but adding this sort of "stateful" assumption always worries me, especailly in codebases I'm not super familiar with.

nic-hartley:
(paging @carols10cents for no other reason than you've helped me here before and I'd appreciate your advice, if you've got a moment)

Turbo87:
I'm not that deeply familiar with the team code but using `find_or_create_by_login()` in the `owner_remove()` fn seems wrong to me. In that case we should probably be using something like `find_by_login()`, though unfortunately that does not exist yet.

I think @carols10cents's original comment explains it quite well too:

> The problem is that the current code is asking GitHub for existence of the org/team before removing it from the crate's owners, and the code should not need to check anything with GitHub in this case; it should just change the record in `crate_owners` for that team+crate to `deleted='t'`.

nic-hartley:
@Turbo87 Appreciate the feedback!

I think you're right; I'm gonna add something like `Team::find_by_login` and call that. (I'd thought it'd have to share with the other code path much more than it actually does -- I didn't realize `login` was a unique ID.) Name TBD once I poke through the codebase to see if there's an existing pattern.

nic-hartley:
I spent an embarrassingly long time trying to figure out how I broke ratelimiting (`publish_new_crate_rate_limited`) before running the tests on main and discovering it fails there, too. I assume it's something with how I've configured things on my machine. Hopefully it passes on CI!

Turbo87:
yeah, unfortunately that test is currently flaky (see https://github.com/rust-lang/crates.io/issues/6952)

just to manage expectations, due to RustConf review times might currently be a bit slower than usual. after 22 hours of travel I probably shouldn't be reviewing code anymore 😅

nic-hartley:
Oh, good, it's not just on my machine! I'm glad it passed on CI.

Don't worry about it, it's not like this is urgent. (At least, I *hope* it's not, considering the issue's from 2019, hah.)

nic-hartley:
Rebased onto the latest `main`

Turbo87 (approved):
sorry that it took so long... 🙈 

looks very good to me though. nice work! :)

nic-hartley:
@Turbo87 No worries! Thanks for the review and merge!

PR #7052, opened 2023-08-31
Replace `cargo_toml` with `cargo-manifest`

Turbo87:
The `cargo_toml` maintainer currently appears unresponsive, while several pull requests for `cargo-manifest` have already been merged and released. Since both crates offer essentially the same functionality we should probably use the one where we can help with the development and maintenance.

### DIFF
--- a/src/admin/render_readmes.rs
+++ b/src/admin/render_readmes.rs
@@ -4,15 +4,17 @@ use crate::{
     schema::{crates, readme_renderings, versions},
 };
 use anyhow::{anyhow, Context};
+use std::path::PathBuf;
 use std::{io::Read, path::Path, sync::Arc, thread};
 
 use crate::storage::Storage;
 use chrono::{TimeZone, Utc};
 use crates_io_markdown::text_to_html;
-use crates_io_tarball::Manifest;
+use crates_io_tarball::{Manifest, StringOrBool};
 use diesel::prelude::*;
 use flate2::read::GzDecoder;
 use reqwest::{blocking::Client, header};
+use std::str::FromStr;
 use tar::{self, Archive};
 
 const USER_AGENT: &str = "crates-admin";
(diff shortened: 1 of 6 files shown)


src/controllers/krate/publish.rs
```diff
@@ -195,11 +194,11 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
                     .map_err(tarball_to_app_error)?;
 
-            let rust_version = tarball_info
-                .manifest
-                .package()
-                .rust_version()
-                .map(|rv| rv.deref().to_string());
+            // `unwrap()` is safe here since `process_tarball()` validates that
+            // we only accept manifests with a `package` section and without
+            // inheritance.
```
LawnGnome, line 199: We might want to add a comment on the `process_tarball()` side as well noting that the validation behaviour is depended on elsewhere, just so future developers don't inadvertently remove it in a refactor.

src/admin/render_readmes.rs
```diff
@@ -184,26 +186,35 @@ fn render_pkg_readme<R: Read>(mut archive: Archive<R>, pkg_name: &str) -> anyhow
...
-        let readme_path = readme.as_path().unwrap_or_else(|| Path::new("README.md"));
-        let path = Path::new(pkg_name).join(readme_path);
+        let readme = manifest
+            .package
+            .as_ref()
+            .and_then(|p| p.readme.as_ref())
+            .and_then(|r| r.as_ref().as_local());
+
+        let readme_path = match readme {
+            Some(StringOrBool::Bool(bool)) if !(*bool) => return Ok("".to_string()),
```
LawnGnome, line 196: I like that the `bool` handling is more explicit now, rather than being hidden in `OptionalFile::as_path()`.

src/admin/render_readmes.rs
```diff
@@ -184,26 +186,35 @@ fn render_pkg_readme<R: Read>(mut archive: Archive<R>, pkg_name: &str) -> anyhow
...
         // Would need access to cargo_vcs_info
         let pkg_path_in_vcs = None;
 
-        text_to_html(
-            &contents,
-            &readme_path,
-            manifest.package().repository(),
-            pkg_path_in_vcs,
-        )
+        let package = manifest.package.unwrap();
```
LawnGnome, line 209: We should reproduce the comment about why this is safe here as well, since it's not guaranteed a future contributor will know the context.

LawnGnome (approved):
:+1:

Let's not deploy this on a Friday, though. :laughing:

hi-rustin (approved):
LGTM

src/controllers/krate/publish.rs
Turbo87: looks like this is already covered by the test suite :)

src/admin/render_readmes.rs
Turbo87: I thought I had implemented it in a way where `manifest.package == None` resulted in an early return from the function, but as it turns out I had changed the code to treat `readme == None` as "use the default readme path" which means there is no early return in this case anymore. I've changed the code to use the same `and_then()` chain as above instead.

PR #7071, opened 2023-09-05
Add comments for some dummy lets

hi-rustin:
Maybe it is helpful when reading the code.


hi-rustin:
@Turbo87 
Could you please take a look? Thanks! 💚 💙 💜 💛 ❤️

Or maybe we can remove this operation? It seems we don't really need it.

PR #7101, opened 2023-09-12
tests: Remove obsolete HTTP recorder

Turbo87:
With our move to using the https://crates.io/crates/object_store crate for S3 uploads we are no longer using the custom-built HTTP recorder. This PR removes move of the related code.

We are however still setting the `reqwest` client up to use a `proxy` URL to prevent us from accidentally performing real HTTP requests when running our test suite.


hi-rustin (approved):
Looks good to me! Thanks!

PR #7105, opened 2023-09-13
Fix flakey ratelimit test

### LINKED ISSUE #6952: Fix flaky rate limiter test

nic-hartley:
Closes #6952

This deletes the flakey ratelimit test and adds four new ones:

- `publish_new_crate_ratelimit_hit`: Adds a row into `publish_limit_buckets` as though the user had been ratelimited, with a `last_refill` far in the future to ensure the ratelimiting code is hit
- `publish_new_crate_ratelimit_expires`: Ditto, but `last_refill` is now juuuust far enough in the past for the ratelimit to hit
- `publish_new_crate_ratelimit_expires`: Defaults the ratelimit to one crate every day, then sets the override to allow the test user to publish two crates a day, and then test that they can publish 2 tests a day and no more.
- `publish_new_crate_expired_override_ignored`: Ditto, but the override `expires_at` is set to just moments ago, and then checks that the test user has the normal (1-crate) limit.

These tests should be infallible (well, assuming the code isn't broken, at least!) and I'm mostly only confident about that because `NaiveDateTime` has no time zones to worry about. Maybe try to avoid running it on a leap second boundary just in case.

(CC @LawnGnome, when you have a second)

### DIFF
--- a/src/tests/krate/publish.rs
+++ b/src/tests/krate/publish.rs
@@ -1,14 +1,19 @@
-use crate::builders::{CrateBuilder, DependencyBuilder, PublishBuilder};
 use crate::new_category;
 use crate::util::insta::assert_yaml_snapshot;
 use crate::util::{RequestHelper, TestApp};
-use crates_io::controllers::krate::publish::{
-    missing_metadata_error_message, MISSING_RIGHTS_ERROR_MESSAGE,
+use crate::{
+    builders::{CrateBuilder, DependencyBuilder, PublishBuilder},
+    util::TestDatabase,
 };
+use chrono::{NaiveDateTime, Utc};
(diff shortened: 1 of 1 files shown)


LawnGnome (approved):
Makes sense to me, and I can't break it locally. Thanks for doing this, and so thoroughly!

nic-hartley:
No problem, glad I could help!

PR #7113, opened 2023-09-15
controllers/krate/publish: Replace `tarball_to_app_error()` fn with `From` implementation

Turbo87:
This lets us get rid of the unnecessary `map_err()` operation since the `?` operator calls `into()` automatically :)


hi-rustin (approved):
Looks good to me. Thanks! :shipit:

PR #7120, opened 2023-09-15
tarball: Use `BTreeMap` for `manifests`

Turbo87:
This should not be a huge deal in terms of performance, but ensures that we generate consistent error messages since `BTreeMap` has a defined non-random iteration order.


hi-rustin (approved):
Looks good to me.

PR #7138, opened 2023-09-18
tests/PublishBuilder: Generate correct manifests

Turbo87:
Up until now our `PublishBuilder` only generated a very basic manifest with `name` and `version` fields by default, even if e.g. dependencies or features were added to the builder. This led to inconsistencies between the JSON metadata and the content of the `Cargo.toml` file in the generated tarball.

This PR fixes the issue by moving the manifest and tarball generation into the `build()` method and extending it to generate the necessary data in the `Cargo.toml` file using `cargo_manifest`.

Note that the `tarball()` and `files()` methods of the `PublishBuilder` had to be replaced for this to work properly. `tarball()` is replaced by direct usage of `PublishBuilder::create_publish_body()`. `files()` is replaced by `add_file()`. Additionally `custom_manifest()` and `no_manifest()` have been added to be able to influence the manifest generation for certain test cases.

This PR is a prerequisite for reading more data from the `Cargo.toml` file during publishes instead of the metadata JSON blob.

### DIFF
--- a/src/tests/builders/publish.rs
+++ b/src/tests/builders/publish.rs
@@ -1,4 +1,5 @@
 use crates_io::views::krate_publish as u;
+use hyper::body::Bytes;
 use std::collections::BTreeMap;
 
 use crates_io_tarball::TarballBuilder;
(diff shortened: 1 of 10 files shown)


src/tests/builders/publish.rs
```diff
@@ -133,33 +121,105 @@ impl PublishBuilder {
...
         let new_crate = u::EncodableCrateUpload {
             name: u::EncodableCrateName(self.krate_name.clone()),
-            vers: u::EncodableCrateVersion(self.version),
-            features: self.features,
-            deps: self.deps,
-            description: self.desc,
+            vers: u::EncodableCrateVersion(self.version.clone()),
+            features: self.features.clone(),
+            deps: self.deps.clone(),
+            description: self.desc.clone(),
```
Turbo87, line 145: I'm not particularly happy about all of the cloning here, but I haven't found a better way yet without a significant code complexity increase...

src/tests/krate/publish/basics.rs
```diff
@@ -22,7 +22,7 @@ fn new_krate() {
     assert!(crates[0].deps.is_empty());
     assert_eq!(
         crates[0].cksum,
-        "8a8d84b87f379d5e32566b14df153c0ab0e1ea87dae79a00b891bb41f93dbbf6"
+        "270bbe1624abd766746bf9938b791fadd88e7e0135339510837e11b45e167350"
```
Turbo87, line 25: This has changed because the `PublishBuilder` by default sets a description and license on the published packages. Previously those two fields haven't been written to the generated `Cargo.toml` file, but now they are, which caused the tarball checksum to change.

src/tests/routes/crates/read.rs
```diff
@@ -143,7 +137,7 @@ fn version_size() {
         .iter()
         .find(|v| v.num == "1.0.0")
         .expect("Could not find v1.0.0");
-    assert_eq!(version1.crate_size, Some(134));
+    assert_eq!(version1.crate_size, Some(158));
```
Turbo87, line 140: similar to the comment above, these numbers changed because of the `description` and `license` fields that are now added to the `Cargo.toml` file by default

PR #7144, opened 2023-09-19
Only upgrade deps once a week

hi-rustin:
We have too many robot pull requests. Sometimes we will bump the same package a couple of times in a short period of time. So I guess maybe we can batch it. Only update the dep once a week. This would help us to save some CI and review resources.

@Turbo87 What do you think? Feel free to close the PR if you think it doesn't matter 🤣

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,4 +1,7 @@
 {
+    schedule: [
+        'before 5am on Monday',
+    ],
     "extends": [
         "config:base",
         ":pinAllExceptPeerDependencies",


Turbo87:
I'd rather not throttle renovatebot like that. The individual PRs have the advantage of catching issues early and making it very visible which version introduced a problem.

We already throttle the updates for a couple of low-priority dependencies that update quite a lot: https://github.com/rust-lang/crates.io/blob/aab95692baa0dd2374a2ab5cb2cb2d89d7b2a2eb/.github/renovate.json5#L26-L27

but I don't think we need to do this for **all** dependencies.

PR #7191, opened 2023-09-27
models/krate: Split `create_or_update()` fn into `create()` and `update()`

Turbo87:
This PR attempts to simplify the `NewCrate::create_or_update()` fn, by splitting up into two more straight-forward `create()` and `update()` fns. This allows us to explicitly use only `create()` in our test suite code, and a combination of both in the publish endpoint.

### DIFF
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -105,7 +105,7 @@ impl<'a> NewCrate<'a> {
         conn.transaction(|conn| {
             // To avoid race conditions, we try to insert
             // first so we know whether to add an owner
-            if let Some(krate) = conn.transaction(|conn| self.create(conn, uploader).optional())? {
+            if let Some(krate) = self.create(conn, uploader).optional()? {
                 return Ok(krate);
             }
 
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -120,32 +120,34 @@ impl<'a> NewCrate<'a> {
         })
     }
 
-    fn save_new_crate(&self, conn: &mut PgConnection, user_id: i32) -> QueryResult<Option<Crate>> {
+    pub fn create(&self, conn: &mut PgConnection, user_id: i32) -> QueryResult<Crate> {
         use crate::schema::crates::dsl::*;
 
         conn.transaction(|conn| {
-            let maybe_inserted: Option<Crate> = diesel::insert_into(crates)
+            let krate: Crate = diesel::insert_into(crates)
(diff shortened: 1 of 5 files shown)


hi-rustin (approved):
Looks good to me. Thanks! I think it is safe to split it.

src/models/krate.rs
```diff
@@ -101,50 +101,39 @@ pub struct NewCrate<'a> {
...
 
-    fn save_new_crate(&self, conn: &mut PgConnection, user_id: i32) -> QueryResult<Option<Crate>> {
+    pub fn create(&self, conn: &mut PgConnection, user_id: i32) -> QueryResult<Crate> {
         use crate::schema::crates::dsl::*;
 
         conn.transaction(|conn| {
-            let maybe_inserted: Option<Crate> = diesel::insert_into(crates)
+            let krate: Crate = diesel::insert_into(crates)
                 .values(self)
                 .on_conflict_do_nothing()
```
LawnGnome, line 120: Do we still want `ON CONFLICT DO NOTHING` here if this is meant to be a straight insert of a new record?

Specifically, I'm wondering how this intersects with the crate owner code below now being unconditional. What happens if `create()` is called and the crate already exists?

Turbo87: - if we use `on_conflict_do_nothing()`:
  - the `get_result()` will return `Ok(Crate)` if the crate was inserted into the database or `Err(NotFound)` if it already exists
  - the `?` after `get_result()` will ensure that the crate owner is not inserted in case of `Err(NotFound)`
  - calling `.optional()` on the return value of `create()` it transforms it into `Result<Option<Crate>>`

- if we **don't** use `on_conflict_do_nothing()`:
  - the `get_result()` will return `Ok(Crate)` if the crate was inserted into the database or `Err(DatabaseError)` if it already exists
  - the error case here would be a bit harder to handle since it will be represented as a somewhat random uniqueness constraint error

> the crate owner code below now being unconditional

it is still conditional through the early return caused by the `?` operator :)

LawnGnome: Gotcha. OK, I was mistaken about what `ON CONFLICT DO NOTHING` would do from a Rust perspective there.

Thanks!

LawnGnome (approved):

src/models/krate.rs
Turbo87: no worries, I've spent a significant amount of time today figuring this out 😅

PR #7194, opened 2023-09-27
publish: Use `description`, `license` and `license_file` fields from embedded `Cargo.toml` file

Turbo87:
This PR changes our handling of the `description`, `license` and `license_file` fields so that they are read from the `Cargo.toml` file embedded in the crate tarball, instead of the metadata JSON blob.

This is another step forward towards avoiding issues like https://blog.vlt.sh/blog/the-massive-hole-in-the-npm-ecosystem in the Rust ecosystem. Note that, compared to e.g. npm, this is currently not exploitable due to the way cargo works, but it is still confusing to our users if the metadata does not match what is actually in the tarball.

This change required a small reordering of things in the publish endpoint. Specifically, the publish size limit is read from the database earlier than previously, which could potentially cause a race condition, but from what I can tell this race condition should not matter in practice (see commit message for details). Similarly, the tarball analysis step now happens before we run any `INSERT` queries on the database, since the tarball analysis result is now needed before we can insert anything. Thirdly, the description and license validation now happens after the tarball analysis for obvious reasons.

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -125,6 +125,14 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
             )));
         }
 
+        let pkg_name = format!("{}-{}", &*metadata.name, &*metadata.vers);
+        let tarball_info = process_tarball(&pkg_name, &*tarball_bytes, maximums.max_unpack_size)?;
+
+        // `unwrap()` is safe here since `process_tarball()` validates that
+        // we only accept manifests with a `package` section and without
+        // inheritance.
+        let package = tarball_info.manifest.package.unwrap();
+
         // Create a transaction on the database, if there are no errors,
         // commit the transactions to record a new or updated crate.
         conn.transaction(|conn| {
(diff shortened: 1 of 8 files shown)


hi-rustin:
/cc myself

I will take a look tomorrow.

src/controllers/krate/publish.rs
```diff
@@ -110,6 +91,51 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
+        let license = package.license.map(|it| it.as_local().unwrap());
+        let license_file = package.license_file.map(|it| it.as_local().unwrap());
+
+        // Make sure required fields are provided
+        fn empty(s: Option<&String>) -> bool {
+            s.map_or(true, String::is_empty)
+        }
+
+        // It can have up to three elements per below conditions.
+        let mut missing = Vec::with_capacity(3);
```
LawnGnome, line 127: The old code had the same behaviour, but aren't there two possible elements, not three?

LawnGnome (approved):
One trivial question, but no substantive concerns. LGTM! :+1:

src/controllers/krate/publish.rs
Turbo87: yeah, I noticed that too :D

the code will soon be replaced anyway, so I didn't feel like adjusting it :)

LawnGnome: Fine by me. We can probably afford the extra 16 bytes. :smiley:

PR #7206, opened 2023-09-29
Wire up typosquatting checks when new packages are published

LawnGnome:
This PR adds experimental support for checking the names of newly published crates for potential typosquatting. It uses essentially the same algorithm as [typogard](https://github.com/mt3443/typogard) as adapted by @dangardner for [offline crates.io checks](https://github.com/dangardner/typogard), reimplemented in Rust by me in the new [typomania crate](https://crates.io/crates/typomania).

I want to emphasise the word _experimental_ in the above paragraph: right now, the _only_ thing that will happen if a new crate might be typosquatting a popular crate is that an e-mail will be sent to a list of interested parties, which at present is defined as myself and @walterhpearce (who has kindly volunteered to help me with this), acting on behalf of the Rust Foundation Security Initiative. No private information is sent in those e-mails; it's a simple notification that crate _A_ might be squatting crates _B_ and _C_. If anyone else would like to be on that list, I'm happy to update the configuration accordingly.

Should there be _actual_ typosquatting, we'll bring it to the crates.io team via Zulip and we can handle it from there using our usual process. (Well, realistically, I'll probably take off my Rust Foundation hat, put on my crates.io hat, and handle it myself in consultation with the rest of you.)

I would like to run this experiment for about two months before taking any further action, to judge the potential impact of using these checks for anything more formal or in a way that might impact the load on the crates.io team.

To be _very_ clear, the crates.io team can say no to this. I promise I won't be mad.

Possible courses of action after that time period might include:

1. If the notifications are way too noisy to be useful, backing this PR out. (I've intentionally kept it as separate as possible — even in places where it would have been simpler to integrate more closely with the core of crates.io — to make this straightforward.)
2. If the notifications are a little bit too noisy, I may adjust the checks to try to improve the signal to noise ratio. (This may be something I do sooner if it's very obviously bad.)
3. If the notifications feel just right, then we might do one or both of:
    1. Notify the crates.io team using Zendesk or some other method, and/or
    2. If the [crate quarantine RFC](https://github.com/rust-lang/rfcs/pull/3464) is ultimately finished and accepted, then we could also feed this into a potential quarantine of new crates that are potential typosquats.

To lay my cards on the table, my hope is that this ends up with course 3(i) — notifying the crates.io team in general. Using quarantine is a possibility, but it's probably something I'd prefer to keep off unless we're actively under a spam attack.

## Technical detail

This is implemented as a new background job that is enqueued when a new crate is published. The background job keeps an in-memory cache of the most popular 3000 crates (as judged by download count) to make the checks speedy and minimise extra load on the database. (Regenerating the cache locally takes about 700 ms, so I'm not terribly concerned about this in practice.) 

The cache TTL is set to 12 hours right now; we could probably even go longer, given the top crates don't really change much.

With the top 3000 crates, the cache uses about 1.1 MiB of heap. I think we can afford that.

Because we have to send e-mails from the background job, `Emails` now has to be plumbed through the `Environment` that is exposed to background jobs. I don't think this is a problem in practice, but it is an extra change.

Finally, as this is an experiment, the configuration is hardcoded into the new worker job module. If this becomes a longer term thing, this would be split out into our normal configuration system for easier management, but right now this isolates the changes as much as possible in `worker::typosquat`.


src/email.rs
```diff
@@ -91,6 +91,32 @@ or go to https://{domain}/me/pending-invites to manage all of your crate ownersh
...
+            "New crate {crate_name} may be typosquatting one or more other crates.\n
+Visit https://{domain}/crates/{crate_name} to see the offending crate.\n
+\n
+Specific squat checks that triggered:\n
+\n
+- {squats}\n",
+            domain = crate::config::domain_name(),
+            squats = squats
+                .iter()
+                .map(|squat| format!("{squat}"))
```
Turbo87, line 112:
```suggestion
                .map(ToString::to_string)
```

I think something like that should work too, right?

Turbo87:
```suggestion
                .map(|squat| format!("- {squat}\n"))
```

This may be an alternative which is a bit more straight-forward compared to the `- {squats}\n` approach above 😅

src/worker/typosquat/config.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Valid characters in crate names.
+pub(super) static CRATE_NAME_ALPHABET: &str =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890-_";
+
+/// E-mail addresses to notify when a potential typosquat is found.
+pub(super) static NOTIFY_EMAILS: &[&str] = &[
+    "adamharvey@rustfoundation.org",
+    "walterpearce@rustfoundation.org",
+];
```
Turbo87, line 17: I think I'd prefer it if we would read these from an env var instead

src/worker/typosquat/config.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// The number of crates to consider in the "top crates" corpus.
+pub(super) static TOP_CRATES: i64 = 3000;
+
+/// This is based on a pre-existing list we've used with crates.io for "easily confused
+/// characters". (I'm not really sure that I consider all of these easily confused, but it's better
+/// than nothing.)
```
Turbo87, line 24: I assume this is based on https://github.com/dangardner/typogard/blob/5ce51b2ec45b7d4847dd944bc3048b206840f873/typogard_crates.py#L80-L123?

The "based on QWERTY keyboard locality" part of the original comment might be relevant here. I'm wondering if we should add QWERTZ and AZERTY locality too? QWERTZ should be relatively straight-forward since it's only a single character that was switched. AZERTY might increase the size of the matrix a bit more though.

src/worker/typosquat/types.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+impl From<Owner> for UniqueOwner {
+    fn from(value: Owner) -> Self {
+        Self::new(value.id(), value.kind())
+    }
+}
+
+// Implemented so we can call `contains` above with a `&str`.
+impl Borrow<str> for UniqueOwner {
```
Turbo87, line 174: out of curiosity, why `Borrow` and not `Deref`?

src/worker/typosquat/types.rs
```diff
@@ -0,0 +1,238 @@
...
+    // The one thing typogard-crates did that isn't implemented here is some basic NLP on the crate
+    // descriptions: this relies on the Python spaCy module, which we'd either have to reimplement
+    // the fuzzy matching logic from, or install a Python environment in the background-worker
+    // buildpack. Neither of those are particularly attractive options, so we're just going to
+    // accept a potentially higher level of false positives than we'd otherwise get.
+}
+
+/// A crate and its owners.
+pub(super) struct OwnedCrate {
+    krate: Crate,
```
Turbo87, line 118: do we need the full `Crate` or would it be sufficient to only store the `description`? 

and since it isn't super-obvious, what is the description used for? I understand that in the original offline implementation NLP was used on the description for some checks, but we don't use that here, right? 🤔

src/worker/typosquat/cache.rs
```diff
@@ -0,0 +1,100 @@
...
+use super::{config, types::TopCrates};
+
+/// Gets the typomania harness for the cached top crates, regenerating it if it is out of date.
+pub(super) fn get_harness(
+    conn: &mut PgConnection,
+) -> Result<Arc<Harness<TopCrates>>, PerformError> {
+    HARNESS_CACHE.get(conn)
+}
+
+static HARNESS_CACHE: Cache = Cache::new();
```
Turbo87, line 23: I'm not a huge fan of this since the global state makes integration tests leaky. Is there some way we can tie this to the running application/worker instance? Maybe put the cache on the `Environment` struct or something like that?

src/worker/typosquat.rs
```diff
@@ -0,0 +1,99 @@
...
+    let krate: Box<dyn Package> = Box::new(OwnedCrate::from_name(conn, name)?);
+    let squats = cache::get_harness(conn)?.check_package(name, krate)?;
+    if !squats.is_empty() {
+        // Well, well, well. For now, the only action we'll take is to e-mail people who hopefully
+        // care to check into things more closely.
+        info!(?squats, "Found potential typosquatting");
+
+        for email in config::NOTIFY_EMAILS.iter() {
+            if let Err(e) = emails.send_possible_typosquat_notification(email, name, &squats) {
+                error!(?e, ?email, "sending possible typosquat notification");
```
Turbo87, line 40:
```suggestion
                error!(?e, ?email, "Failed to send possible typosquat notification");
```

src/worker/typosquat/cache.rs
```diff
@@ -0,0 +1,100 @@
...
+                }
+            }
+        }
+        None
+    }
+
+    fn update(&mut self, harness: Arc<Harness<TopCrates>>) {
+        self.harness = Some(harness);
+        self.last_update = Some(Instant::now());
+    }
```
Turbo87, line 99: since Heroku restarts our dynos every 24 hours I'm wondering if we even need the TTL/update functionality. if it would simplify the code then we could consider dropping it for now.

src/email.rs
Turbo87: one more thought that just passed my mind: we are linking to `https://{domain}/crates/{crate_name}` above, but I guess it would be nice to also link to the squatted crates in this list?

Cargo.toml
```diff
@@ -88,12 +88,20 @@ tar = "=0.4.40"
...
 threadpool = "=1.8.1"
-tokio = { version = "=1.32.0", features = ["net", "signal", "io-std", "io-util", "rt-multi-thread", "macros"]}
+tokio = { version = "=1.32.0", features = [
+  "net",
+  "signal",
+  "io-std",
+  "io-util",
+  "rt-multi-thread",
+  "macros",
+] }
```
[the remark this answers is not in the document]
LawnGnome: You are correct. I need to figure out how to turn that off on a per-project basis!

src/email.rs
LawnGnome: Agreed; done.

src/tests/util/test_app.rs
```diff
@@ -257,6 +257,7 @@ impl TestAppBuilder {
                 index,
                 app.http_client().clone(),
                 None,
+                Emails::new_in_memory(),
```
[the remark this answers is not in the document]
LawnGnome: Fixed in the rebased branch I'm about to push.

src/worker/typosquat/config.rs
LawnGnome: Done; we now read from `$TYPOSQUAT_NOTIFICATION_EMAILS` (feel free to bikeshed).

src/worker/typosquat/config.rs
LawnGnome: Good catch. Yeah, adding QWERTZ and AZERTY wasn't hard, so I did that.

src/worker/typosquat/types.rs
LawnGnome: You are correct that actually keeping the description was pointless since we're not doing any fancy NLP stuff in the background job. Removed, and now we just keep the name and ownership information.

src/worker/typosquat/types.rs
LawnGnome: Honestly, because that's how I know how to handle this case (string newtype that you want to be able to use as a hashmap key, but also use `&str` to access).

What does the `Deref` version of this look like?

src/worker/typosquat/cache.rs
LawnGnome: I've moved it to `Environment`, which is kinda ugly in a different way, but does get rid of the global state.

src/worker/typosquat/cache.rs
LawnGnome: Yep, good call. I've removed all the cache TTL stuff for now.

LawnGnome:
This has been heavily rebased after the various background worker changes in the last couple of weeks — although @Turbo87 looked at this a while back, I'd appreciate a re-review (preferably from him, but I'm also grateful to anyone else who wants to cast a quick eye over this).

Tested locally; all seems to work as expected based on trying to publish a `serd` crate.

src/worker/typosquat/types.rs
Turbo87: > What does the `Deref` version of this look like?

thanks to https://crates.io/crates/derive-deref it would only be `#[derive(Deref)]`, which would derive an impl that would deref to `&String`.

src/worker/typosquat/database.rs
```diff
@@ -0,0 +1,230 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::{test_util::pg_connection, worker::typosquat::test_util::Faker};
+    use thiserror::Error;
+
+    use super::*;
+
+    #[test]
+    fn top_crates() -> Result<(), Error> {
```
Turbo87, line 170: wouldn't it be easier to use `anyhow::Result` directly instead of the custom error type? seems like we don't really care about any of the specific enum variants?

src/worker/environment.rs
```diff
@@ -17,6 +20,9 @@ pub struct Environment {
     pub storage: Arc<Storage>,
     pub connection_pool: DieselPool,
     pub emails: Emails,
+
+    /// A lazily initialised cache of the most popular crates ready to use in typosquatting checks.
+    typosquat_cache: OnceLock<Result<typosquat::Cache, typosquat::CacheError>>,
```
Turbo87, line 25: if I understand this implementation correctly, if the first attempt to fill the cache fails it will not be tried again?

I noticed a similar situation with the `repository` field of this struct lately and fixed it like this: https://github.com/rust-lang/crates.io/blob/f8fc9ae18efa59f4965ce2ce07de4250ccdb74e0/src/worker/environment.rs#L40-L54

basically, for each access of the method it checks whether the field is initialized and returns the initialized thing if so. otherwise it tries to initialize and returns an error if it fails. the next time the method is used the initialization will be retried though.

I assume this might also resolve the need for some of the `Arc` gymnastics in the error type below :)

src/worker/typosquat/mod.rs
```diff
@@ -0,0 +1,9 @@
+mod cache;
+mod config;
+mod database;
+
+#[cfg(test)]
+pub(super) mod test_util;
+
+pub use cache::{Cache, Error as CacheError};
+pub use database::Crate;
```
Turbo87, line 9: just a small nitpick: any reason why this module lives in the `worker` module? it happens to run in the background worker with the current implementation, but I assume nothing is stopping us from using it elsewhere, right?

for similar reasons I've lately moved the CloudFront and Fastly clients out of the `worker` module too :)

src/controllers/krate/publish.rs
```diff
@@ -222,9 +222,10 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
             return Err(cargo_err("expected at most 5 categories per crate"));
         }
 
-        let max_features = existing_crate
-            .and_then(|c| c.max_features.map(|mf| mf as usize))
-            .unwrap_or(app.config.max_features);
+        let max_features = match &existing_crate {
+            Some(c) => c.max_features.map(|mf| mf as usize),
+            None => None,
+        }.unwrap_or(app.config.max_features);
```
Turbo87, line 228:
```suggestion
        let max_features = existing_crate.as_ref()
            .and_then(|c| c.max_features.map(|mf| mf as usize))
            .unwrap_or(app.config.max_features);
```

`.as_ref()` seems a bit easier to resolve the problem :)

src/worker/typosquat/mod.rs
LawnGnome: > just a small nitpick: any reason why this module lives in the `worker` module?

No; it made more sense with the original, pre-rebase module layout, but it doesn't any more. I'll move it out like those modules.

src/controllers/krate/publish.rs
LawnGnome: Ugh, pick the very last thing I did on a day when I was still recovering from being sick. :see_no_evil:

src/worker/environment.rs
LawnGnome: > if I understand this implementation correctly, if the first attempt to fill the cache fails it will not be tried again?

Yep. That was actually intentional on my part — the typosquatting checks aren't mission critical, so I'd rather it fails with a warning and just disables itself until the next dyno restart rather than putting load on the database trying to rebuild the cache each time a crate is published if there's some sort of weird problem preventing the query from running properly.

I'm open minded on whether the right thing to do is to change it to be more like the repo behaviour your mentioned, or if I should just document that choice more clearly in the code. Thoughts?

src/worker/typosquat/database.rs
LawnGnome: `typomania::Result` doesn't require an error that can coerce to `anyhow::Error` (because it doesn't include `Sync` as a constraint). That's probably a mistake, but applying a little bit of `thiserror` was the quickest way around the API issue for now.

Turbo87: hmm, okay. guess we can clean this up later.

src/worker/environment.rs
Turbo87: okay, if it's intentional then that's fine. just wanted to check :)

Turbo87 (approved):
:shipit:  let's give this a try :)

src/worker/typosquat/types.rs
LawnGnome: I see the issue now — we need to be able to call `contains` with a `&str`, because that's all the `AuthorSet` and `Package` traits require, but `Deref` would require a `&String`.

I think the `Borrow` is correct here.

Turbo87: yeah, just played around with it a bit. looks like you're right 👍

PR #7209, opened 2023-09-29
publish: Read `categories` metadata from embedded `Cargo.toml` file

Turbo87:
... instead of the metadata JSON blob

see also:

- https://github.com/rust-lang/crates.io/pull/7194
- https://github.com/rust-lang/crates.io/pull/7201
- https://github.com/rust-lang/crates.io/pull/7204

PR #7214, opened 2023-09-30
publish: Read `keywords` metadata from embedded `Cargo.toml` file

Turbo87:
... instead of the metadata JSON blob

see also:

- https://github.com/rust-lang/crates.io/pull/7194
- https://github.com/rust-lang/crates.io/pull/7201
- https://github.com/rust-lang/crates.io/pull/7204
- https://github.com/rust-lang/crates.io/pull/7209

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -157,6 +157,25 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
         validate_url(documentation.as_deref(), "documentation")?;
         validate_url(repository.as_deref(), "repository")?;
 
+        let keywords = package
+            .keywords
+            .map(|it| it.as_local().unwrap())
+            .unwrap_or_default();
+
+        if keywords.len() > 5 {
+            return Err(cargo_err("expected at most 5 keywords per crate"));
+        }
+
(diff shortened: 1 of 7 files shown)


src/controllers/krate/publish.rs
```diff
@@ -157,6 +157,25 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
+        let keywords = package
+            .keywords
+            .map(|it| it.as_local().unwrap())
+            .unwrap_or_default();
+
+        if keywords.len() > 5 {
+            return Err(cargo_err("expected at most 5 keywords per crate"));
+        }
+
+        for keyword in keywords.iter() {
```
hi-rustin, line 169: Maybe we can extract it as a validation function. The publish function is a bit too long.

hi-rustin (approved):
A nit issue. Rest LGTM.

src/controllers/krate/publish.rs
Turbo87: yeah, that's the plan once I have all of the other things migrated :)

hi-rustin: Sounds good! :shipit:

PR #7218, opened 2023-10-02
Switch to binary prefixes

kana-rus:
This is from
- https://github.com/rust-lang/crates.io/discussions/7177

and changes the unit of package size to use binary prefixes as shown in test cases in `tests/unit/helpers/pretty-bytes-test.js`.

### DIFF
--- a/app/helpers/pretty-bytes.js
+++ b/app/helpers/pretty-bytes.js
@@ -5,8 +5,8 @@ import prettyBytes from 'pretty-bytes';
 /**
  * See https://github.com/rust-lang/crates.io/discussions/7177
  * 
- * Here set {minimum,maximum}FractionDigits to 1 because
- * `cargo publish` uses this format
+ * Here we set fraction digits to 1 because `cargo publish`
+ * uses this format (see https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_package.rs#L168-L171)
 */
 export default helper(([bytes], options) => prettyBytes(bytes, {
     ...options,
--- a/app/helpers/pretty-bytes.js
+++ b/app/helpers/pretty-bytes.js
@@ -6,7 +6,7 @@ import prettyBytes from 'pretty-bytes';
  * See https://github.com/rust-lang/crates.io/discussions/7177
  *
  * Here we set fraction digits to 1 because `cargo publish`
- * uses this format (see https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_package.rs#L168-L171)
+ * uses this format (see https://github.com/rust-lang/cargo/blob/0.73.1/src/cargo/ops/cargo_package.rs#L167-L170)
  */
 export default helper(([bytes], options) =>
   prettyBytes(bytes, {
--- a/app/helpers/pretty-bytes.js
+++ b/app/helpers/pretty-bytes.js
@@ -4,15 +4,10 @@ import prettyBytes from 'pretty-bytes';
 
 /**
  * See https://github.com/rust-lang/crates.io/discussions/7177
- *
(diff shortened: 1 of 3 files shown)


app/helpers/pretty-bytes.js
```diff
@@ -2,4 +2,17 @@ import { helper } from '@ember/component/helper';
 
 import prettyBytes from 'pretty-bytes';
 
-export default helper(([bytes], options) => prettyBytes(bytes, options));
+/**
+ * See https://github.com/rust-lang/crates.io/discussions/7177
+ *
+ * Here we set fraction digits to 1 because `cargo publish`
+ * uses this format (see https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_package.rs#L168-L171)
```
hi-rustin, line 9: Could you please use a tag instead of the master branch here?
Using a fixed version can make sure we can always find it.

hi-rustin (commented):
From https://percy.io/crates-io/crates.io/builds/30429495/changed/1682351587?browser=firefox&browser_ids=49&subcategories=changes_requested,unreviewed&viewLayout=overlay&viewMode=new&width=1280&widths=375,1280

We display `KIB` in the version list and `KiB` in the crate view. I suggestion to unify them to `KiB`. The `KIB` caused by this CSS: https://github.com/rust-lang/crates.io/blob/3a0631aeb86620b325b97d1ee937b91a8469e165/app/components/version-list/row.module.css#L133

Thanks!

app/helpers/pretty-bytes.js
kana-rus: I see! I'll fix them...

app/helpers/pretty-bytes.js
```diff
@@ -2,4 +2,17 @@ import { helper } from '@ember/component/helper';
...
+ *
+ * Here we set fraction digits to 1 because `cargo publish`
+ * uses this format (see https://github.com/rust-lang/cargo/blob/0.73.1/src/cargo/ops/cargo_package.rs#L167-L170)
+ */
+export default helper(([bytes], options) =>
+  prettyBytes(bytes, {
+    ...options,
+    binary: true,
+    minimumFractionDigits: 1,
+    maximumFractionDigits: 1,
```
Turbo87, line 16:
```suggestion
    binary: true,
    minimumFractionDigits: 1,
    maximumFractionDigits: 1,
    ...options,
```

this should make it possible to override these defaults

Turbo87, line 16: tbh I'm not convinced by including these fractional values by default. unless the crate is very small they basically serve no real purpose and just clutter the user interface (see percy snapshots)

kana-rus: Yes, thinking about it, it looks better for cleaner UI to not set fraction digits to 1 by default

Turbo87:
thanks :)

app/helpers/pretty-bytes.js
```diff
@@ -2,4 +2,12 @@ import { helper } from '@ember/component/helper';
...
 
-export default helper(([bytes], options) => prettyBytes(bytes, options));
+/**
+ * See https://github.com/rust-lang/crates.io/discussions/7177
+ */
+export default helper(([bytes], options) =>
+  prettyBytes(bytes, {
+    binary: true,
+    ...options,
+  }),
```
Andrew15-5, line 12: @Turbo87, what's the point of "making it possible to override these defaults", if it is never (AFAIU) overridden anywhere?

In https://github.com/rust-lang/crates.io/discussions/7177#discussioncomment-7154109 I intentionally wrote `{...options, binary: true}` so that binary prefixes will be anywhere and everywhere (that was the point from the start).

PR #7227, opened 2023-10-03
Use workspace dependencies to manage the common dependencies

hi-rustin:
Use [workspace dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) to manage the common dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,23 @@ default = ["slow-tests"]
 # to disable these tests.
 slow-tests = []
 
-[dependencies]
+[workspace.dependencies]
 anyhow = "=1.0.75"
+clap = "4.4.6"
+crates_io_index = { path = "crates_io_index" }
+reqwest = "=0.11.22"
+secrecy = "=0.8.0"
+semver = "=1.0.20"
+serde = "=1.0.189"
+serde_json = "=1.0.107"
+tempfile = "=3.8.0"
+thiserror = "=1.0.49"
+tracing = "=0.1.37"
+tracing-subscriber = "=0.3.17"
+bytes = "=1.5.0"
+
+[dependencies]
+anyhow.workspace = true
 async-trait = "=0.1.73"
 aws-sigv4 = "=0.56.1"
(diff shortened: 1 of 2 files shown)


Turbo87:
I'm personally not a huge fan of this feature. Since we use renovatebot to update the deps it's not a big deal to have the deps duplicated and this way we have everything in one place.

I'm curious what the rest of the team thinks about it though.

hi-rustin:
@Turbo87 We discussed it last night's meeting, they don't have strong opinions about using it or not. But we think this helps us reduce some duplication and organize the common dependencies in the top-level Cargo.toml. 

So could you please help merge it or do you have any concerns about it?

Turbo87:
> reduce some duplication

I agree that this reduces the duplication of version numbers for the dependencies, but at the same time the dependency names that are used by multiple packages are now duplicated in the package `Cargo.toml` files *and* the workspace.

It feels a bit weird that once a dependency happens to be used by multiple packages it now needs to be moved to the workspace and be inherited everywhere. I can see the benefit of using something like when there are multiple published libraries in a project, but that does not apply here and since we pin all of the dependencies and renovatebot takes care of updates, the added complexity from the workspace inheritance simply does not seem worth it to me.

hi-rustin:
> but that does not apply here and since we pin all of the dependencies and renovatebot takes care of updates, the added complexity from the workspace inheritance simply does not seem worth it to me.

We also discussed it last night. We want to put all crates into the workspace. There is no reason to make them as independent crates. (We also discussed maybe we can publish crates.io as a crate then we can use docs.rs to host our docs for the code) So inheriting metadata from the workspace will also help to deal with the homepage and public state, etc. So I think it is valuable to use it. There are no hurts for Roboot automatic updating and it helps us reduce a lot of duplications.

hi-rustin:
But if you have strong opinions about not wanting to use it in crates.io. I am OK to close my PR.
Thanks for your review!  💚 💙 💜 💛 ❤️

PR #7252, opened 2023-10-08
Replace `ring` dependency with `p256`

Turbo87:
The `ring` project is releasing new versions without tags on the corresponding repository or even changelog entries, so we basically have no idea of what is happening over there. `p256` is a pure-Rust alternative for the only `ring` algorithm that we're currently using and it also includes functions to parse the PEM format public keys.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,6 +356,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


hi-rustin (approved):
Looks good to me. Thanks!

PR #7253, opened 2023-10-08
Render footnotes in crates' README.md

dnaka91:
I realized that crates.io currently doesn't seem to render _Markdown Footnotes_ which are part of the GitHub Flavored Markdown spec. This is already supported on GitHub (obviously) and docs.rs renders them as well.

This conflicts a little bit with the HTML sanitiser, as it rewrites IDs but I did my best to match them properly in the URL sanitizer.

PR #7265, opened 2023-10-10
markdown: Use insta inline snapshots for tests

Turbo87:
see https://insta.rs/docs/snapshot-types/#inline-snapshots

The second commit in this PR also fixes a bug in our code block rendering tests which was discovered by the introduction of the snapshot tests. The combination of `r#"` strings and `\` led to us not actually using correct markdown syntax and the test only working by accident.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -784,6 +784,7 @@ dependencies = [
  "ammonia",
  "comrak",
  "htmlescape",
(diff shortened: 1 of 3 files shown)


Turbo87:
@dnaka91 I will rebase #7253 on top of this once it's merged :)

Issue #7269, opened 2023-10-11
Crate metadata requests take very long on crates with a lot of features


Carlosted:
### Current Behavior

When requesting metadata through [`shields.io`](https://shields.io/) for a crate I maintain, [`icondata`](https://crates.io/crates/icondata), I have to specify a specific version of the crate in order for the API request to complete. 

This crate makes heavy use of feature flags which bloats the response payload of the API. A request for a specific version is fine, with a payload size of ~1MB, but when I don't specify a version i.e., when every version's metadata is included, the payload is ~20MB, which seems to make the request timeout on `shields.io`'s side.

### Expected Behavior

I know this is partly due to how `shields.io` works, but perhaps being able to specify the version as `latest` instead of a specific semver would help. An example of what it would enable use to do is request the license of the latest crate without requesting every crate's metadata.

### Steps To Reproduce

GET request to `https://crates.io/api/v1/crates/icondata`
vs.
GET request to `https://crates.io/api/v1/crates/icondata/0.1.0`

### Environment

- Browser: Chrome version 118.0.5993.70
- OS: Mac OS (Apple silicon m1)

Turbo87:
Hi @Carlosted, on behalf of the crates.io team let me tell you that we were quite surprised to see a crate with over 23.000 features... 😅 

Let me first respond directly to the primary request in this issue: supporting something like `/:crate_name/latest`. I think this is a reasonable request, and I'm wondering if we could even generalize it to `/:crate_name/:version_req` so that something like `/foo/^1.2.3` would redirect to `/foo/1.4.0` if that was the highest matching version.

Regarding the `icondata` crate specifically, you will find that having that many features will surface a lot of other issues eventually. You can see for example that the index file for this crate (https://index.crates.io/ic/on/icondata) already takes up 11 MB, with only 9 versions published so far. If more versions of the crate are published, this file will quickly grow and `cargo` will take a loooong time resolving dependencies for any project where this crate is being used because it needs to download this big file all of the time.

We quickly discussed the situation between some of the crates.io and cargo team members today. It seems likely that we will have to introduce some kind of limit on the maximum number of features a crate can have, to keep the registry working correctly in the future.

Due to the above we would like to encourage you to think of different ways to improve the compile times without needing over 23.000 features for a single crate :)

Turbo87:
[...]

PR #7276, opened 2023-10-13
Introduce a limit on the number of features a crate can provide or enable

Turbo87:
As surfaced by https://github.com/rust-lang/crates.io/issues/7269, having crates with large numbers of features can have significant side effects.

This PR introduces a **limit of 300 features** that a crate can have, and another limit on the number of features a crate feature can enable. This still leaves us with a hypothetical maximum of 90,000 combinations, but anyone reaching this limit would most likely be a malicious actor doing this deliberately.

As with other limits on crates.io, these limits can be overridden in the database on a per-crate basis. While the [`icondata`](https://crates.io/crates/icondata) crate is an example of something we probably don't want to allow, crates like [`web-sys`](https://crates.io/crates/web-sys) and [`windows`](https://crates.io/crates/windows) have somewhat legitimate reasons for having larger numbers of features.

After merging and deploying this we may want to take a look at https://gist.github.com/Turbo87/730bb71d4bb134d708cbde7a78ecae38 and set limit overrides for existing crates with larger numbers of features. Since it's unlikely for those few crates to publish right at the same time we deploy and before we have updated their limits we most likely don't need a technological solution like a two-step deployment process for this.

Related:

- https://github.com/rust-lang/crates.io/issues/7269
- https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/5.20billion.20features


migrations/2023-10-13-080226_add-features-limit/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE crates ADD max_features SMALLINT;
```
LawnGnome, line 1: This maxes us out at 32,767 possible features for a crate version ever. I'm personally OK with this, but are we sure there isn't a case where this could bite us later?

(I mean, I guess practically we could just change it if and when we need to.)

src/controllers/krate/publish.rs
```diff
@@ -221,14 +221,34 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
+        }
+
         for (key, values) in features.iter() {
             if !Crate::valid_feature_name(key) {
                 return Err(cargo_err(&format!(
                     "\"{key}\" is an invalid feature name (feature names must contain only letters, numbers, '-', '+', or '_')"
                 )));
             }
 
+            let num_features = values.len();
```
hi-rustin, line 244: I think this only happens if we use a bunch of optional deps. Because if you have more than 200 real features then it would fail in the first check. So could you please add some comments to explain it? Please correct me if I misunderstand it.

hi-rustin (commented):
Thanks!

migrations/2023-10-13-080226_add-features-limit/up.sql
Turbo87: > are we sure there isn't a case where this could bite us later?

there are probably use cases that could use more features than that, but given the current technological constraints allowing that many features for any crate has significant consequences. before those constraints are resolved I wouldn't be comfortable allowing crates with more than 32k features, and afterwards we probably won't even need the limit anymore.

src/controllers/krate/publish.rs
```diff
@@ -221,14 +221,34 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
 
+        let max_features = existing_crate
+            .and_then(|c| c.max_features.map(|mf| mf as usize))
+            .unwrap_or(app.config.max_features);
+
         let features = tarball_info.manifest.features.unwrap_or_default();
+        let num_features = features.len();
+        if num_features > max_features {
+            return Err(cargo_err(&format!(
+                "Maximum number of features was reached (uploaded: {}, limit: {})",
```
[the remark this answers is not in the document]
Turbo87: I've been wondering about that, but then I noticed that our "your crate file is too big" error message also has no such information and people are apparently clever enough to contact our support email address. I'm open to adding it though if you think it would be useful.

src/controllers/krate/publish.rs
Turbo87: if you have a dependency on another crate that has 5000 features, then in your crate you can have a feature that enables e.g. 1000 of the features of the dependency. if you do this in an application then that's fine, but if you do it in a library and publish that to crates.io we run into similar issues to the issue report linked above.

Eh2406:
The underlying limiting resource is the size of the full git checkout, or if we are no longer worried about git than it's the size of the index file for the package. I wonder if it makes sense to have a general limit on the size of an index entry in addition to these specific known problem cases. I love that this error message tells us exactly what field was at fault. But I feel like we're playing whack a mole finding problems. If there was a "size of index entry < 500KB" (with exceptions granted) then we can be quite confident we have covered all of the problems.

Turbo87:
@Eh2406 yeah, I was considering that option, but then most people that release crates on crates.io probably have no idea of the implementation details of the crate index format. In other words: seeing such an error message is not particularly actionable unless you know these implementation details.

the other factor is that the index entries are constructed asynchronously in the background worker, so by the time we answer the request we have no idea yet how big the index entry will be unless we want to construct it twice.

src/controllers/krate/publish.rs
hi-rustin: Oh, I see. Do you mean we depend on one crate that doesn't follow this limitation(Released before this PR)? Then in our index file, we have to include all the features of that dependency.

But shouldn't we check the features field from the dependency instead of the features from this package?

Turbo87: > Do you mean we depend on one crate that doesn't follow this limitation(Released before this PR)?

not necessarily. Imagine you are writing library `x` and you depend on libraries `y` and `z`. In your `x/Cargo.toml` you could have:

```toml
[features]
all_the_things = [
  "y/feature1",
  "y/feature2",
  "y/feature3",
  # ...
  "z/feature1",
  "z/feature2",
  "z/feature3",
  # ...
]
```

so while `y` and `z` each have less than 300 features, their combination can have more than that.

src/controllers/krate/publish.rs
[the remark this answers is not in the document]
Turbo87: I thought some more about this, and I guess I agree. I've made the error message a bit more friendly and included a sentence about sending us an email to increase the limits.

PR #7280, opened 2023-10-13
Correct the package name

hi-rustin:
We discussed this during the team meeting. We can give it the right name 😆

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -10,7 +10,7 @@
     {{content-for 'head'}}
 
     <link rel="stylesheet" href="{{rootURL}}assets/vendor.css">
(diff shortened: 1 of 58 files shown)


Turbo87 (approved):
just for context, when the project was originally created the name "crates.io" was not a thing yet and it was just generically called "cargo registry". we've renamed the backend project a while ago too, so I guess keeping it consistent on the frontend makes sense. it's a little unfortunate that it breaks caching for our users due to the different filename, but that seems like a minor issue compare to some other problems we have with the frontend code 😅

hi-rustin:
Thanks for your review!  💚 💙 💜 💛 ❤️

PR #7289, opened 2023-10-14
Make membership of the workspace more explicit

hi-rustin:
Path dependencies are automatically members of the workspace and don't have to be added explicitly. This PR tries to make it more explicit to reduce the confusion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 default-run = "server"
 
 [workspace]
-members = ["crates_io_smoke_test"]
+members = ["crates_io_*"]
 
 [profile.release]
 opt-level = 2


Turbo87:
> There is no good reason we only put the smoke test as a member.

The reason we haven't added them is that path dependencies are automatically members of the workspace and don't have to be added explicitly.

As with the other PR, workspace inheritance adds another level of complexity to the manifests that is barely worth it with the little metadata that is actually duplicated between the different packages in this repository. Instead of looking for information in the `Cargo.toml` file of the package I now have the extra indirection of needing to go to the workspace manifest instead and jumping back and forth for the inherited and not-inherited metadata.

For library projects where this kind of metadata gets published and should stay in sync workspace inheritance can be quite useful, but it does not seem worth the extra complexity for an application project like crates.io.

hi-rustin:
> The reason we haven't added them is that path dependencies are automatically members of the workspace and don't have to be added explicitly.

I think making it more explicit may be better. What do you think?

hi-rustin:
> Instead of looking for information in the `Cargo.toml` file of the package I now have the extra indirection of needing to go to the workspace manifest instead and jumping back and forth for the inherited and not-inherited metadata.

Yes. Right now we have this kind of problem. Hope one day the IDE can help us with this situation. I will revert this change.

PR #7317, opened 2023-10-19
SQL: Fix metadata handling in to_semver_no_prerelease() function

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/3882

This PR is basically reviving https://github.com/rust-lang/crates.io/pull/3886, since we now have a generated and stored `semver_no_prerelease` column in the `versions` table, which should fix the performance concerns that led to #3886 being initially rejected.

### DIFF
--- a/src/sql.rs
+++ b/src/sql.rs
@@ -53,3 +53,39 @@ macro_rules! pg_enum {
 }
 
 pub(crate) use pg_enum;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::db::sql_types::semver::Triple;
+    use crate::schema::sql_types::SemverTriple;
+    use crate::test_util::pg_connection;
+    use diesel::prelude::*;
+    use diesel::select;
+
+    sql_function!(fn to_semver_no_prerelease(x: Text) -> Nullable<SemverTriple>);
+
+    #[test]
+    fn to_semver_no_prerelease_works() {
+        let mut conn = pg_connection();
+
+        #[track_caller]
+        fn test(conn: &mut PgConnection, text: &str, expected: Option<(u64, u64, u64)>) {
+            let query = select(to_semver_no_prerelease(text));
+            let result = query
+                .get_result::<Option<Triple>>(conn)
+                .unwrap()
+                .map(|triple| (triple.major, triple.minor, triple.teeny));
+
+            assert_eq!(result, expected);
+        }
+
+        test(&mut conn, "0.0.0", Some((0, 0, 0)));
+        test(&mut conn, "1.2.4", Some((1, 2, 4)));
+        test(&mut conn, "1.2.4+metadata", Some((1, 2, 4)));
+        test(&mut conn, "1.2.4-beta.3", None);
+        // TODO: see https://github.com/rust-lang/crates.io/issues/3882
+        test(&mut conn, "0.4.45+curl-7.78.0", None);
+        test(&mut conn, "0.1.4-preview+4.3.2", None);
+    }
+}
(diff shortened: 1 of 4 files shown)


Turbo87:
@paolobarbolini since @LawnGnome is on vacation and you've proven that you know a lot about Postgres, any chance you could take a look at this? :)

I'm most concerned about two things: 
- performance impact on the reverse dependencies query, which I assume should be zero since we have the stored column now
- impact of running the `create or replace function` migration. it looks like `drop` + `create` is fast, but requires us to add the generated column again afterwards, which is slow. `create or replace function` is also fast but appears to keep the generated column, though potentially with old values.

src/sql.rs
```diff
@@ -53,3 +53,39 @@ macro_rules! pg_enum {
...
+            assert_eq!(result, expected);
+        }
+
+        test(&mut conn, "0.0.0", Some((0, 0, 0)));
+        test(&mut conn, "1.2.4", Some((1, 2, 4)));
+        test(&mut conn, "1.2.4+metadata", Some((1, 2, 4)));
+        test(&mut conn, "1.2.4-beta.3", None);
+        // see https://github.com/rust-lang/crates.io/issues/3882
+        test(&mut conn, "0.4.45+curl-7.78.0", Some((0, 4, 45)));
+        test(&mut conn, "0.1.4-preview+4.3.2", None);
```
hi-rustin, line 89: What about `1.0.0-beta+exp.sha.5114f85`? It seems we can not handle it right now. But it is a valid version in [regex101.com/r/vEYRnS/1](https://regex101.com/r/vEYRnS/1).

hi-rustin (commented):
Thanks!

src/sql.rs
hi-rustin: I also tested it with semver:

```rust
#!/usr/bin/env -S cargo +nightly -Zscript

//! ```cargo
//! [dependencies]
//! semver = "1.0"
//! ```


use semver::Version;

fn main() {
    match Version::parse("1.0.0-beta+exp.sha.5114f85") {
        Ok(v) => println!("Valid version: {}", v),
        Err(e) => println!("Invalid version: {}", e),
    }
}
```

```console
cargo +nightly -Zscript version.rs
warning: `package.edition` is unspecifiead, defaulting to `2021`
   Compiling version v0.0.0 (/Volumes/t7/code/cargo-scripts)
    Finished dev [unoptimized + debuginfo] target(s) in 0.39s
     Running `/Users/hi-rustin/.cargo/target/bd/d06c14836752f5/debug/version`
Valid version: 1.0.0-beta+exp.sha.5114f85
```

hi-rustin (approved):
Looks good to me. I forgot this is called `to_semver_no_prerelease`, so I deleted my previous review comments:(

Tested locally and it worked well.

LawnGnome (approved):
Makes sense to me, and the tests cover the two variations that are important (dash before plus, and vice versa). :+1:

paolobarbolini:
> @paolobarbolini since @LawnGnome is on vacation and you've proven that you know a lot about Postgres, any chance you could take a look at this? :)
> 
> I'm most concerned about two things:
> 
>     * performance impact on the reverse dependencies query, which I assume should be zero since we have the stored column now
> 
>     * impact of running the `create or replace function` migration. it looks like `drop` + `create` is fast, but requires us to add the generated column again afterwards, which is slow. `create or replace function` is also fast but appears to keep the generated column, though potentially with old values.

I missed the notification 😅.

* No impact
* You should be able to replace the function. The GENERATED column won't be updated though. What you can do is write an UPDATE which assigns the raw version string to itself in order to force the GENERATED column to refresh. Should be even faster if you can filter it to only the rows which may actually need to be regenerated, although at your size it should be possible to YOLO it and just refresh the entire table and VACUUM.

PR #7323, opened 2023-10-20
Migrate request and body timeouts from nginx to axum

Turbo87:
We've been using the [`client_body_timeout 30` nginx option](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_timeout) to ensure timeouts between request body read operations that can be caused by maliciously slow clients.

https://docs.rs/tower-http/latest/tower_http/timeout/struct.RequestBodyTimeoutLayer.html serves a similar purpose, and while we're at it, I've also added corresponding request and response body timeouts (see http://nginx.org/en/docs/http/ngx_http_core_module.html#send_timeout). This should cover all of the timeouts that nginx and the Heroku router have provided us so far.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,7 +92,7 @@ threadpool = "=1.8.1"
 tokio = { version = "=1.33.0", features = ["net", "signal", "io-std", "io-util", "rt-multi-thread", "macros"]}
 toml = "=0.8.2"
 tower = "=0.4.13"
-tower-http = { version = "=0.4.4", features = ["fs", "catch-panic"] }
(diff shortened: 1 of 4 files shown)


Turbo87:
@rust-lang/crates-io a sanity check for these changes would be appreciated! :)

JohnTitor (approved):

Turbo87:
@syphar brought up via Zulip that since we currently don't apply a `send_timeout` in the nginx configuration we probably shouldn't introduce it as a side effect of this pull request, and I tend to agree. I've removed the `ResponseBodyTimeoutLayer` commit from the pull request for now. We can always add it later, if we think that it makes sense.

PR #7336, opened 2023-10-22
Format all markdown files with prettier

hi-rustin:
When I working on [this PR](https://github.com/rust-lang/crates.io/pull/7335), my vscode automatically changes the format of markdown files. So I think it would be better to use prettier to format all markdowns.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "lint:deps": "ember dependency-lint",
     "lint:hbs": "ember-template-lint app",
     "lint:js": "eslint . --cache",
-    "prettier": "prettier --write package.json '**/*.js'",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "lint:hbs": "ember-template-lint app",
     "lint:js": "eslint . --cache",
     "precompress": "node ./script/precompress-assets.mjs",
-    "prettier": "prettier --write package.json '**/*.js' '**/*.mjs'",
+    "prettier": "prettier --write package.json '**/*.js' '**/*.mjs' '**/*.md'",
```
Turbo87, line 25: keep in mind though that on CI we only check prettier stuff via ESLint, so there is a risk of the files diverging again in the future

PR #7345, opened 2023-10-23
CSP: Fix `Content-Security-Policy` header for `github-redirect.html` file

### LINKED ISSUE #7343: Unable to "Log in with GitHub"
### Current Behavior

Tried with Safari, Firefox and Chrome.

Once I click on "Log in with GitHub" and authorize rust-lang, the redirect leads to a blank page. Below is the error from console window:

#### Safari
```
[Error] Refused to execute a script because its hash, its nonce, or 'unsafe-inline' does not appear in the script-src directive of the Content Security Policy. (github-redirect.html, line 4)
```

#### Firefox
```
Content-Security-Policy: The page’s settings blocked the loading of a resource at inline (“script-src”). [github-redirect.html:4:10](https://crates.io/github-redirect.html?code=39278c642b776265f0ce&state=BxN0Puui71cniOBUWTNr4w)
```

#### Chrome
```
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval' 'sha256-n1+BB7Ckjcal1Pr7QNBh/dKRTtBQsIytFodRiIosXdE='". Either the 'unsafe-inline' keyword, a hash ('sha256-dbf9FMl76C7BnK1CC3eWb3pvsQAUaTYSHAlBy9tNTG0='), or a nonce ('nonce-...') is required to enable inline execution.
```

### Expected Behavior

The redirect is successful and I am able to log in successfully.

### Steps To Reproduce

0. Unauthorize rust-lang in GitHub
1. Open Safari or Firefox (above versions).
2. Visit crates.io
3. Click on "Log in with GitHub"
4. Authorize rust-lang

### Environment

- Browser: Safari 17.0 / Firefox 118.0.2 / Chrome 118.0.5993.96
- OS: macOS 14.0 (23A344)

### Anything else?

_No response_

Turbo87:
This should fix #7343

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -165,10 +165,12 @@ impl Default for Server {
 
         let storage = StorageConfig::from_environment();
 
+        // `sha256-dbf9FMl76C7BnK1CC3eWb3pvsQAUaTYSHAlBy9tNTG0=` refers to
+        // the `script` in `public/github-redirect.html`
         let content_security_policy = format!(
             "default-src 'self'; \
             connect-src 'self' *.ingest.sentry.io https://docs.rs https://play.rust-lang.org {cdn_domain}; \
-            script-src 'self' 'unsafe-eval' 'sha256-n1+BB7Ckjcal1Pr7QNBh/dKRTtBQsIytFodRiIosXdE='; \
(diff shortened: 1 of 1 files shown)


Turbo87:
merging directly to test this out on the staging system ASAP

Turbo87:
for context: I assume that #7325 broke this for some reason. it looks like with nginx serving the `github-redirect.html` file the `Content-Security-Policy` header was not present, but now that we've moved it to axum it is being applied? seems weird on first glance though since the header was added unconditionally by nginx. I'm not sure how exactly this previously was not applied, but nevertheless this fix should be correct.

PR #7359, opened 2023-10-25
Implement `X-Forwarded-For` header processing

Turbo87:
This PR essentially implements http://nginx.org/en/docs/http/ngx_http_realip_module.html inside of our axum server, as another step towards removing the nginx wrapper.

… but there is a twist. Heroku is somewhat broken. Their broken handles the header incorrectly in case the original request contains multiple `X-Forwarded-For` headers. Instead of adding the connecting IP to the last header, or merging the headers, it adds the IP to the first header 😱 

I've reported this to Heroku in the form of a support ticket, but while that is being processed we will unfortunately have to use a workaround. Luckily CloudFront merges these headers, so if we get multiple headers we can know for certain that the request did not come from CloudFront.

On the bright side, we only use the value for logging at the moment. The PR adds another field to the logs with the IP addressed that was parsed from the `X-Forwarded-For` headers. It also adds a log marker in case the IP is different from the one that nginx supplies us in the `X-Real-IP` header. This will allow us to see if there are differences to the nginx algorithm once production traffic hits this code. Spoiler alert: yes, the Heroku workaround is a difference, but for regular, non-malicious traffic this should not matter.

### DIFF
--- a/src/real_ip.rs
+++ b/src/real_ip.rs
@@ -1,0 +1,68 @@
+use http::HeaderValue;
+use std::net::IpAddr;
+use std::str::from_utf8;
+
+/// Parses the content of an `X-Forwarded-For` header into a
+/// `Vec<Result<IpAddr, &[u8]>>`.
+#[allow(dead_code)]
+fn parse_xff_header(header: &HeaderValue) -> Vec<Result<IpAddr, &[u8]>> {
+    let bytes = header.as_bytes();
+    if bytes.is_empty() {
+        return vec![];
+    }
+
+    bytes
+        .split(|&byte| byte == b',')
+        .map(|bytes| parse_ip_addr(bytes))
+        .collect()
+}
+
+fn parse_ip_addr(bytes: &[u8]) -> Result<IpAddr, &[u8]> {
+    from_utf8(bytes)
+        .map_err(|_| bytes)?
+        .trim()
+        .parse()
+        .map_err(|_| bytes)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use http::HeaderValue;
+
+    #[test]
+    fn test_parse_xff_header() {
+        #[track_caller]
+        fn test(input: &'static [u8], expectation: Vec<Result<&str, &[u8]>>) {
+            let header = HeaderValue::from_bytes(input).unwrap();
+
+            let expectation: Vec<Result<IpAddr, &[u8]>> = expectation
+                .into_iter()
+                .map(|ip| ip.map(|ip| ip.parse().unwrap()))
+                .collect();
+
+            assert_eq!(parse_xff_header(&header), expectation)
+        }
+
+        test(b"", vec![]);
+        test(b"1.2.3.4", vec![Ok("1.2.3.4")]);
+        test(
+            b"1.2.3.4, 11.22.33.44",
+            vec![Ok("1.2.3.4"), Ok("11.22.33.44")],
+        );
+        test(
+            b"oh, hi,,127.0.0.1,,,,, 12.34.56.78  ",
+            vec![
+                Err(b"oh"),
+                Err(b" hi"),
+                Err(b""),
+                Ok("127.0.0.1"),
+                Err(b""),
+                Err(b""),
+                Err(b""),
+                Err(b""),
+                Ok("12.34.56.78"),
+            ],
+        );
+    }
+}
--- a/src/real_ip.rs
+++ b/src/real_ip.rs
@@ -1,17 +1,199 @@
 use http::{HeaderMap, HeaderValue};
+use ipnetwork::IpNetwork;
+use once_cell::sync::Lazy;
+use std::iter::Iterator;
 use std::net::IpAddr;
 use std::str::from_utf8;
 
 const X_FORWARDED_FOR: &str = "X-Forwarded-For";
 
+const CLOUD_FRONT_STRS: &[&str] = &[
+    // CloudFront IP addresses from http://d7uri8nf7uskq.cloudfront.net/tools/list-cloudfront-ips
+    // Last updated: 2022-03-26
+    "3.10.17.128/25",
+    "3.11.53.0/24",
+    "3.35.130.128/25",
+    "3.101.158.0/23",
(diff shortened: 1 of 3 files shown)


src/real_ip.rs
```diff
@@ -0,0 +1,299 @@
...
+use std::iter::Iterator;
+use std::net::IpAddr;
+use std::str::from_utf8;
+
+const X_FORWARDED_FOR: &str = "X-Forwarded-For";
+
+const CLOUD_FRONT_STRS: &[&str] = &[
+    // CloudFront IP addresses from http://d7uri8nf7uskq.cloudfront.net/tools/list-cloudfront-ips
+    // Last updated: 2022-03-26
+    "3.10.17.128/25",
```
Turbo87, line 13: this list was imported from: https://github.com/rust-lang/crates.io/blob/e874a242c5cbff8470bdac62fdf6f05223acc129/config/nginx.conf.erb#L27-L159

src/real_ip.rs
```diff
@@ -0,0 +1,299 @@
...
+    "205.251.250.0/23",
+    "205.251.252.0/23",
+    "205.251.254.0/24",
+    "216.137.32.0/19",
+    "223.71.11.0/27",
+    "223.71.71.96/27",
+    "223.71.71.128/25",
+];
+
+static CLOUD_FRONT_NETWORKS: Lazy<Vec<IpNetwork>> = Lazy::new(|| {
```
walterhpearce, line 146: Using `hashbrown` or `tinyset` here on CIDR integer values  would be significant faster than an iterated string match on a `Vec`. The `cidr` crate implements these as integer types.

src/real_ip.rs
```diff
@@ -0,0 +1,299 @@
...
+    CLOUD_FRONT_STRS
+        .iter()
+        .map(|s| s.parse().unwrap())
+        .collect()
+});
+
+fn is_cloud_front_ip(ip: &IpAddr) -> bool {
+    CLOUD_FRONT_NETWORKS
```
walterhpearce, line 154: Same comment about the speed.

src/real_ip.rs
```diff
@@ -0,0 +1,299 @@
...
+        .any(|trusted_proxy| trusted_proxy.contains(*ip))
+}
+
+pub fn process_xff_headers(headers: &HeaderMap) -> Option<IpAddr> {
+    let mut xff_iter = headers.get_all(X_FORWARDED_FOR).iter();
+    let first_header = xff_iter.next()?;
+
+    let has_more_headers = xff_iter.next().is_some();
+    return if has_more_headers {
+        // This only happens for requests going directly to crates-io.herokuapp.com,
```
walterhpearce, line 165: Are there cases where direct calls to `crates-io.herokuapp.com` should even actually be supported? As a DDOS protection method, wouldn't it make sense to *only* support calls directly to our utilized CDN proxies?

walterhpearce (commented):
Mainly, two major comments:
- String comparisons on a flat vec for something per-request is very slow, it'd be beneficial to use `cidr` in an appropriate integer-sized hashmap for lookups. 
- Should we even allow direct calls via heroku, or should we just start banishing all requests not using a CDN?

src/real_ip.rs
Turbo87: I've created a benchmark for this at https://github.com/Turbo87/cidr-list-benchmark. 

These are the results on my machine:

```
cidr_check              fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ cidr_bitstring_tree  165.3 ns      │ 51.87 µs      │ 749.3 ns      │ 782.9 ns      │ 1159371 │ 1159371
│                       6.048 Mitem/s │ 19.27 Kitem/s │ 1.334 Mitem/s │ 1.277 Mitem/s │         │
├─ cidr_vec_iter        181.6 ns      │ 1.255 µs      │ 197.2 ns      │ 201.8 ns      │ 133017  │ 4256544
│                       5.506 Mitem/s │ 796.2 Kitem/s │ 5.07 Mitem/s  │ 4.954 Mitem/s │         │
├─ ip_network_table     26 ns         │ 378.5 ns      │ 30.57 ns      │ 31.38 ns      │ 123100  │ 15756800
│                       38.44 Mitem/s │ 2.641 Mitem/s │ 32.71 Mitem/s │ 31.86 Mitem/s │         │
├─ ip_network_vec_iter  115.1 ns      │ 1.399 µs      │ 124.3 ns      │ 127.8 ns      │ 193559  │ 6193888
│                       8.68 Mitem/s  │ 714.7 Kitem/s │ 8.043 Mitem/s │ 7.821 Mitem/s │         │
╰─ ipnetwork_vec_iter   72.88 ns      │ 425.7 ns      │ 76.8 ns       │ 79.4 ns       │ 139257  │ 8912448
                        13.72 Mitem/s │ 2.348 Mitem/s │ 13.01 Mitem/s │ 12.59 Mitem/s │         │
```

As the results show, we could speed it up by about 1.5x via https://crates.io/crates/ip_network_table, but given the median time of 76ns for the current implementation I'm not sure it's worth the extra dependency and build time 😅

src/real_ip.rs
Turbo87: > Are there cases where direct calls to `crates-io.herokuapp.com` should even actually be supported?

no

> As a DDOS protection method, wouldn't it make sense to _only_ support calls directly to our utilized CDN proxies?

yes

in other words: I fully agree with you, we just haven't gotten to that part yet. I don't think this is a blocker for this PR though and can be worked on in parallel :)

Turbo87:
Since this is purely additive and the performance concerns have been addressed, I'll go ahead and merge this to test it out on staging and then deploy it to production.

src/real_ip.rs
```diff
@@ -0,0 +1,299 @@
...
+        // since AWS CloudFront automatically merges these headers into one.
+        //
+        // The Heroku router has a bug where it currently (2023-10-25) appends
+        // the connecting IP to the **first** header instead of the last.
```
Turbo87, line 169: for context: https://github.com/heroku/roadmap/issues/238

PR #7361, opened 2023-10-25
Mention concerns for "the first person who asks" for ownership transfer

Marcono1234:
See https://github.com/rust-lang/crates.io/discussions/7173
And see also the [npm event-stream incident](https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident).

Any feedback, especially regarding wording, is appreciated!

(CC @carols10cents since you apparently proposed adding this section in #5216)

### DIFF
--- a/app/templates/policies.hbs
+++ b/app/templates/policies.hbs
@@ -82,6 +82,11 @@
     help@crates.io for it.
   </blockquote>
 
+  <p>
+    Keep in mind that the new owner might develop your crate in a way you never intended it, or might completely
+    repurpose your crate. Transferring the crate to a malicious user could have a significant impact for any
+    existing users of your crate.</p>
+
   <p>The crates.io team may delete crates from the registry that do not comply with the policies on this document. In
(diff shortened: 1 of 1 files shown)


Marcono1234:
In case this pull request is accepted in some form, should https://github.com/rust-lang/rfcs/pull/3463 then be updated as well?

Marcono1234:
> it seems like this paragraph applies to maintainer invitations in general, and not just to these specific situations where someone reserved a crate name and doesn't need it anymore

Do you mean the new paragraph I added. Yes, I guess you are right. But as seen with https://crates.io/crates/wgpu-noboiler, it appears users might also apply "How can I delete a crate I own from the registry?" in other situations. That was why I though this new paragraph was needed for this section the most.

> delay this PR until we've integrated that policy update, and then we can re-evaluate where this paragraph would make most sense. does that sound good to you?

Yes that is completely fine for me, thanks also for discussing this PR!

Marcono1234:
Thanks a lot @Turbo87 for updating this pull request and including the warning in the policies!

PR #7364, opened 2023-10-26
Remove `X-Real-Ip` header

Turbo87:
This PR is essentially a continuation of https://github.com/rust-lang/crates.io/pull/7302, which was unlocked by #7359. Since we process the `X-Forwarded-For` header now in the same way that nginx does (plus the Heroku workaround...), we no longer need to have nginx do the same thing. This removes one more reason for us to still use nginx on our servers.


Turbo87:
#7359 has been running in production for a day now without any issues or differences to the nginx algorithm. let's get this merged to reduce our log volume again :)

PR #7365, opened 2023-10-26
Add `USE_NGINX` env var feature flag

Turbo87:
If this environment variable is set to `n` it will disable the use of the nginx wrapper process on our Heroku servers. The servers are currently not using this environment variable and thus should keep using nginx for now:

- `dotenvy::var("USE_NGINX")` will return `Err(Error::EnvVar(VarError::NotPresent))`
- `unwrap_or_default()` will turn this `Err` into `""`
- `"" != "n"` results in `true`
- -> `let use_nginx_wrapper = true`

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -104,12 +104,16 @@ impl Default for Server {
     ///
     /// This function panics if the Server configuration is invalid.
     fn default() -> Self {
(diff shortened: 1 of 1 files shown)


Turbo87:
I've tested this on our staging server and everything appears to work as expected... :shipit:

PR #7370, opened 2023-10-27
Remove obsolete nginx wrapper

Turbo87:
https://github.com/rust-lang/crates.io/commit/2b7331015ea1daa808cfdbf94470d02e8ffd129e introduced nginx as a proxy in front of the app in 2014. Since then we have migrated to the https://github.com/tokio-rs/axum web framework, and were using nginx for less and less things.

At this point all of the functionality from our nginx config has been migrated into our axum server and there is nothing left for nginx to do.

Additionally, our use of nginx on Heroku also runs into https://github.com/heroku/heroku-buildpack-nginx/issues/31, which causes us to drop thousands of connections on every deployment.

This PR removes the nginx wrapper and buildpack from crates.io, exposing the axum server directly to the Heroku router.

Note that I've opened this as a draft pull request for now until https://github.com/rust-lang/crates.io/pull/7365 has run in production for a couple of days without issues :)


Turbo87:
This has been running in production without any issues for about a week now. I guess we can say goodbye to nginx now :)

Issue #7250, opened 2016-09-03
Confusing error for invalid feature names


Boddlnagg:
I just tried to publish a crate containing features with names that have dots in them (see e.g. [here](https://github.com/contextfree/winrt-rust/blob/3ec231209a224edc92e913cf3fab1a3aa0e7d454/Cargo.toml)).
On `cargo publish` I got the following: `api errors: invalid upload request: ApplicationError("invalid crate name specified: windows.web")`, and there are actually several issues:
1. The message confusingly complains about the _crate_ name
2. It should complain earlier, not on `publish`, when you think that everything should work
3. Is there a good reason that dots are not allowed?

By the way this also affects @retep998's 0.3 rewrite of winapi, where I originally got the idea for these features names: https://github.com/retep998/winapi-rs/blob/dev/Cargo.toml

Turbo87:
interesting, I just read through https://doc.rust-lang.org/cargo/reference/features.html#the-features-section and I wasn't aware that we had restrictions on feature names that only apply to the crates.io side. I was under the impression that the same restrictions existed on the cargo side too.

I'm open to lifting some of these restrictions, but we will first need to find out why they exist in the first place so that we don't accidentally break something. if anyone feels like working on this, you are very welcome to help figure this out. git blame will likely give a few hints :)

RE the error message: this had been fixed quite some time ago already. it now warns about the feature name not being valid. though I guess we could give some more context with a link to the docs or something like that.

I agree with @epage that clientside validation by cargo could be tricky with the different format restrictions.

hi-rustin:
I went through the git history, I think we don't have a special reason to refuse `.` in the feature name.
1. We introduced the basic check in this PR. Called  `name_to_slug` https://github.com/rust-lang/crates.io/commit/45131f36dc371b50559fa2a85df3a63da5d25b74
2. We renamed it to `valid_name`. https://github.com/rust-lang/crates.io/commit/89c407977bee6377c9b47996749a60bb268f6c79
3. We added support for `/` in the feature name. https://github.com/rust-lang/crates.io/commit/d64528f9e9218bb1152b93a9b5e88e735044762e
4. We allowed feature names to begin with numbers. https://github.com/rust-lang/crates.io/commit/4bf9efa4af7421d4b3d04553437fc35c403ff87b
5. We added support for `+` in the feature name. https://github.com/rust-lang/crates.io/commit/5f842f73c6a9f32719508db4b1c534959b02f6e8

[...]

PR #7379, opened 2023-10-29
Use the same feature and dep name validation rules from Cargo 

hi-rustin:
close https://github.com/rust-lang/crates.io/issues/7250

- This PR allowed `.` in the feature name. I added some unit tests and integration tests for it.
- This PR also updated the feature, crate, and dep name validation rules from Cargo.

The summary:

| category                    | Cargo                                                                                                   | crates.io before change                                                                                      | crates.io after change                                                                                 |
| --------------------------- | ------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- |
| Crate name                  | 1. cannot start with a number<br/>2. can only start with most letters or `_`<br/>3. can only contain numbers, `-`, `_`, or most letters. | 1. must start with alphabetic<br/>2. can only contain numbers, letters, `-`, `_`                         | 1. must start with alphabetic<br/>2. can only contain numbers, letters, `-`, `_` **(Nothing changes)**                |
| Dep name                    | 1. cannot start with a number<br/>2. can only start with most letters or `_`<br/>3. can only contain numbers, `-`, `_`, or most letters. | 1. first character must be alphabetic or `_`<br/>2. can only contain numbers, letters, `-`, `_`         | 1. cannot start with a number<br/>2. can only start with ASCII letters or `_`<br/>3. can only contain numbers, `-`, `_`, or ASCII letters.  **(Nothing changes)**       |
| Normal feature name         | 1. can only start with most letters, `_`, or `0` to `9`<br/>2. can only contain numbers, `+`, `-`, `_`, `.`, or most letters | can only contain numbers, `+`, `-`, `_`, or most letters                                       | 1. can only start with most letters, `_`, or `0` to `9`<br/>2. can only contain numbers, `+`, `-`, `_`, `.`, or most letters  |
| Optional dep name   | same as dep name                                                                                        | can only contain numbers, `+`, `-`, `_`, or most letters                                       | same as dep name                                     |
| Optional dep feature name   | dep_name/feat_name<br/>1. so the first part is the same as the dep name<br/>2. the second part is the same as the normal feature name. | can only contain numbers, `+`, `-`, `_`, or most letters                                       | dep_name/feat_name<br/>1. so the first part is the same as the dep name<br/>2. the second part is the same as the normal feature name.                                   |

### DIFF
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -220,12 +220,41 @@ impl Crate {
                 .unwrap_or(false)
     }
 
-    /// Validates the THIS parts of `features = ["THIS", "and/THIS"]`.
-    pub fn valid_feature_name(name: &str) -> bool {
-        !name.is_empty()
-            && name
-                .chars()
-                .all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-' || c == '+' || c == '.')
+    /// Validates the THIS parts of `features = ["THIS", "and/THIS", "dep:THIS", "dep?/THIS"]`.
+    /// 1. The name must be non-empty.
+    /// 2. The first character must be a Unicode XID start character, `_`, or a digit.
+    /// 3. The remaining characters must be Unicode XID characters, `_`, `+`, `-`, or `.`.
+    pub fn valid_feature_name(name: &str) -> AppResult<()> {
+        if name.is_empty() {
+            return Err(cargo_err("feature cannot be an empty"));
+        }
+        let mut chars = name.chars();
+        if let Some(ch) = chars.next() {
+            if !(unicode_xid::UnicodeXID::is_xid_start(ch) || ch == '_' || ch.is_ascii_digit()) {
+                return Err(cargo_err(&format!(
+                    "invalid character `{}` in feature `{}`, \
+                the first character must be a Unicode XID start character or digit \
+                (most letters or `_` or `0` to `9`)",
+                    ch, name,
+                )));
+            }
+        }
+        for ch in chars {
+            if !(unicode_xid::UnicodeXID::is_xid_continue(ch)
+                || ch == '+'
+                || ch == '-'
+                || ch == '.')
+            {
+                return Err(cargo_err(&format!(
+                    "invalid character `{}` in feature `{}`, \
+                characters must be Unicode XID characters, `+`, `-`, or `.` \
+                (numbers, `+`, `-`, `_`, `.`, or most letters)",
+                    ch, name,
+                )));
+            }
+        }
+
+        Ok(())
     }
 
     /// Validates the prefix in front of the slash: `features = ["THIS/feature"]`.
@@ -242,9 +271,9 @@ impl Crate {
         match name.split_once('/') {
             Some((dep, dep_feat)) => {
                 let dep = dep.strip_suffix('?').unwrap_or(dep);
-                Crate::valid_feature_prefix(dep) && Crate::valid_feature_name(dep_feat)
+                Crate::valid_feature_prefix(dep) && Crate::valid_feature_name(dep_feat).is_ok()
             }
-            None => Crate::valid_feature_name(name.strip_prefix("dep:").unwrap_or(name)),
+            None => Crate::valid_feature_name(name.strip_prefix("dep:").unwrap_or(name)).is_ok(),
         }
     }
 
@@ -518,6 +547,10 @@ mod tests {
     #[test]
     fn valid_feature_names() {
         assert!(Crate::valid_feature("foo"));
+        assert!(Crate::valid_feature("1foo"));
+        assert!(Crate::valid_feature("_foo"));
+        assert!(Crate::valid_feature("_foo-_+.1"));
+        assert!(Crate::valid_feature("_foo-_+.1"));
         assert!(!Crate::valid_feature(""));
         assert!(!Crate::valid_feature("/"));
         assert!(!Crate::valid_feature("%/%"));
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -225,7 +225,7 @@ impl Crate {
         !name.is_empty()
             && name
                 .chars()
-                .all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-' || c == '+')
+                .all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-' || c == '+' || c == '.')
     }
 
     /// Validates the prefix in front of the slash: `features = ["THIS/feature"]`.
@@ -531,5 +531,7 @@ mod tests {
         assert!(!Crate::valid_feature("dep:foo?/bar"));
         assert!(!Crate::valid_feature("foo/?bar"));
         assert!(!Crate::valid_feature("foo?bar"));
+        assert!(Crate::valid_feature("bar.web"));
+        assert!(Crate::valid_feature("foo/bar.web"));
     }
 }
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -288,15 +288,15 @@ impl Crate {
         Ok(())
     }
 
-    /// Validates a whole feature string, `features = ["THIS", "ALL/THIS"]`.
-    pub fn valid_feature(name: &str) -> bool {
+    /// Validates a whole feature string, `features = ["THIS", "and/THIS", "dep:THIS", "dep?/THIS"]`.
+    pub fn valid_feature(name: &str) -> AppResult<()> {
         match name.split_once('/') {
             Some((dep, dep_feat)) => {
                 let dep = dep.strip_suffix('?').unwrap_or(dep);
-                Crate::valid_dependency_name(dep).is_ok()
-                    && Crate::valid_feature_name(dep_feat).is_ok()
+                Crate::valid_dependency_name(dep)?;
+                Crate::valid_feature_name(dep_feat)
             }
-            None => Crate::valid_feature_name(name.strip_prefix("dep:").unwrap_or(name)).is_ok(),
+            None => Crate::valid_feature_name(name.strip_prefix("dep:").unwrap_or(name)),
         }
     }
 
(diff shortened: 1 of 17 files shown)


hi-rustin:
I also tested it locally. It worked well:
```toml
[package]
name = "hello-bisect"
version = "0.2.0"
edition = "2021"
license = "MIT"
description = "A simple hello world program for testing bisect"
# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]

[features]
"foo.bar" = []
```

```console
❯ cargo publish  --index http://localhost:8888/git/index --token xxxx 
    Updating `http://localhost:8888/git/index` index
warning: manifest has no documentation, homepage or repository.
See https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata for more info.
   Packaging hello-bisect v0.2.0 (/Volumes/t7/code/hello-bisect)
   Verifying hello-bisect v0.2.0 (/Volumes/t7/code/hello-bisect)
   Compiling hello-bisect v0.2.0 (/Volumes/t7/code/hello-bisect/target/package/hello-bisect-0.2.0)
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
    Packaged 5 files, 1.6KiB (1.0KiB compressed)
   Uploading hello-bisect v0.2.0 (/Volumes/t7/code/hello-bisect)
    Uploaded hello-bisect v0.2.0 to registry `http://localhost:8888/git/index`
note: Waiting for `hello-bisect v0.2.0` to be available at registry `http://localhost:8888/git/index`.
You may press ctrl-c to skip waiting; the crate should be available shortly.
warning: timed out waiting for `hello-bisect v0.2.0` to be available in registry `http://localhost:8888/git/index`
note: The registry may have a backlog that is delaying making the crate available. The crate should be available soon.
```

image

Turbo87:
Allowing `.` in general seems useful to me. 👍 

I'm wondering though whether we should migrate to the ruleset that cargo is using (see https://doc.rust-lang.org/cargo/reference/features.html#the-features-section):

> Feature names may include characters from the [Unicode XID standard](https://unicode.org/reports/tr31/) (which includes most letters), and additionally allows starting with `_` or digits `0` through `9`, and after the first character may also contain `-`, `+`, or `.`.

The original issue is complaining about the server having different rules than cargo, and while allowing `.` characters would solve this specific case, it might be worth addressing the underlying issue of cargo and crates.io using different validation schemes.

hi-rustin:
> I'm wondering though whether we should migrate to the ruleset that cargo is using (see [doc.rust-lang.org/cargo/reference/features.html#the-features-section](https://doc.rust-lang.org/cargo/reference/features.html#the-features-section)):

Sounds make sense. I will fix it tomorrow.

Turbo87:
> I will fix it tomorrow.

Might be worth getting opinions from the rest of the team first. I'm not sure if there is a reason why crates.io is currently more restrictive than cargo, but some of the other team members might know.

jtgeibel:
I'm not aware of any intentional restrictions put in place here. Looking through the [history outlined here](https://github.com/rust-lang/crates.io/issues/7250#issuecomment-1784011867), it looks like this originally shared an implementation with the crate name validation logic and that we've just incrementally allowed more characters as they were requested.

As I understand it, we're only easing the constraints on the `feature_name` portion of a feature string, and not the `crate_name` portion in the following examples:
* `dep:crate_name`
* `crate_name/feature_name`
* `crate_name?/feature_name`
* `feature_name`

In that case I don't see any issues with lifting these constraints on feature names to align with cargo.

I am a bit confused by some of our existing logic though:
* On publish, the crate name must be a `valid_feature_prefix` and the first character must be alphabetic ([here](https://github.com/rust-lang/crates.io/blob/ba20ff2f7f2f2acde54be434098abe1fe42f1bc5/src/models/krate.rs#L202-L209)).
* When checking the dependencies during a publish, we enforce that each dependency must be a `valid_feature_prefix` and the first character must be alphabetic or `_`. ([here](https://github.com/rust-lang/crates.io/blob/ba20ff2f7f2f2acde54be434098abe1fe42f1bc5/src/models/krate.rs#L216-L223)). It seems like we allow the `_` prefix to handle dependency renaming, though it isn't clear to me that we only enforce it in that scenario. (I also have no idea what constraints cargo puts on renamed dependencies.)
* Then when validating the feature prefix (what I call `crate_name` in the examples above) we seem to only validate that all characters `|c| c.is_ascii_alphanumeric() || c == '_' || c == '-'` ([here](https://github.com/rust-lang/crates.io/blob/ba20ff2f7f2f2acde54be434098abe1fe42f1bc5/src/models/krate.rs#L233-L240)).

Not that it would need to be done in this PR, but it seems like maybe we should be more strict when validating the `crate_name` portion, as currently the following are allowed when they seem invalid to me:
```rust
        assert!(Crate::valid_feature("0"));
        assert!(Crate::valid_feature("-"));
        assert!(Crate::valid_feature("0/bar"));
        assert!(Crate::valid_feature("-/bar"));
```

It's also not clear to me if we should allow the following always, or possibly verify that a dependency has been reanamed to `_`:
```rust
        assert!(Crate::valid_feature("_"));
        assert!(Crate::valid_feature("_/bar"));
```

hi-rustin:
From the Cargo side:
For crate name:  https://github.com/rust-lang/cargo/blob/b1a3ad24c4088d125fd478670a2e171c3951cfa6/src/cargo/util/restricted_names.rs#L45
1. Package name can not start with a number.
2. Package name can only start with most letters or `_`.
3. Package name can only contain numbers, `-`, `_`, or most letters.

For future name: https://github.com/rust-lang/cargo/blob/b1a3ad24c4088d125fd478670a2e171c3951cfa6/src/cargo/core/summary.rs#L433
1. Feature name can only start with most letters or `_` or `0` to `9`.
2. Feature name can only contain numbers, `+`, `-`, `_`, `.`, or most letters.

hi-rustin:
> ```rust
>         assert!(Crate::valid_feature("0/bar"));
>         assert!(Crate::valid_feature("-/bar"));
> ```

So I think these two cases are invalid.

hi-rustin:
The summary:

|          \        |   Cargo  | crates.io |
|  ---------------- |  ------  | ----  |
|   crate name      | 1. can not start with a number </br> 2.can only start with most letters or `_` <br/> 3. can only contain numbers, `-`, `_`, or most letters.     | 1. must start with alphabetic </br> 2. can only contain letters and `-`, `-`  |
|    dep name       |   1. can not start with a number </br> 2.can only start with most letters or `_` <br/> 3. can only contain numbers, `-`, `_`, or most letters.   | 1. first character must be alphabetic or `_` <br/> 2. can only contain letters and `-`, `-`  |
|  feature name     |  1. can only start with most letters or `_` or `0` to `9` <br/> 2. can only contain numbers, `+`, `-`, `_`, `.`, or most letters    | can only contain numbers, `+`, `-`, `_`, or most letters |

hi-rustin:
We discussed it in tonight's meeting:
1. We can try to use the same rules from Cargo.
2. We will support UnicodeXID for feature names.
3. We won't support UnicodeXID for crate names.(considering the security issues, we want to keep it only ASCII letters now)

hi-rustin:
I will summarize all the changes tomorrow.

src/models/krate.rs
```diff
@@ -192,59 +192,112 @@ impl Crate {
...
     }
 
-    pub fn valid_dependency_name(name: &str) -> bool {
-        let under_max_length = name.chars().take(MAX_NAME_LENGTH + 1).count() <= MAX_NAME_LENGTH;
-        Crate::valid_dependency_ident(name) && under_max_length
-    }
+    // Checks that the name is a valid identifier.
+    // 1. The name must be non-empty.
+    // 2. The first character must be an ASCII character or `_`.
+    // 3. The remaining characters must be ASCII alphanumerics or `-` or `_`.
```
Turbo87, line 218: I'm not sure I understand why you changed all of these implementations too? I thought the goal of this PR was to adjust the validation rules for feature names, but not for crate names 🤔

hi-rustin: I remember the last meeting, we decided not to allow unicode for crate names, but we can follow the same rules for the validations.

> Not that it would need to be done in this PR, but it seems like maybe we should be more strict when validating the `crate_name` portion, as currently the following are allowed when they seem invalid to me:

This also addressed these issues.

But if you think we should only change the feature name first. I can revert my changes for crate names.

hi-rustin: If we don't change the create name validation as well, `_a?/feature_name` still is invalid in crates.io but it works in Cargo.

Turbo87: I think changing the rules on crate names should be out of scope for this PR. That would need a bit more investigation since these are relevant for URLs, S3 paths, and other things where allowing additional characters could have unintended consequences. Let's focus this PR on just the feature names.

hi-rustin: Make sense. I will revert it. Thanks!

hi-rustin: I reverted it in https://github.com/rust-lang/crates.io/pull/7379/commits/c495ff45c399b0aa4cd1c66cb3796bb986723955

I retained the modification of the message as it provides a clearer error message to users.

Turbo87: > I retained the modification of the message as it provides a clearer error message to users.

the problem is that this makes the diff a lot harder to review because the PR is doing multiple things at once. (see https://mtlynch.io/code-review-love/#5-narrowly-scope-changes 😉)

hi-rustin: > the problem is that this makes the diff a lot harder to review because the PR is doing multiple things at once.

Sorry for the big PR. I tried to make sure one commit only had one main change. Could you please review it by commit? The reason it has a really big change is the old validations for create name and dep name also use some feature validation rules(functions). We mixed them.

hi-rustin: If you think it is still really difficult to review. I can split it into three different PRs. Sorry again for the confusing.

hi-rustin: The first one: https://github.com/rust-lang/crates.io/pull/7500

PR #7390, opened 2023-10-30
tests: Replace `FreshSchema` architecture with `TestDatabase`

Turbo87:
### Before (`FreshSchema`)

The `FreshSchema::new()` fn was creating a new Postgres schema in the test database, and then running the migrations on that new schema to create the tables, indices, routines, etc.

Every test using `FreshSchema::new()` was spending about 180ms on connecting to the database, only 1ms to create the new schema, and then about 230ms on running the migrations.

Most of our tests are currently using https://github.com/allaboutapps/integresql#approach-2a-isolation-by-transactions instead of `FreshSchema::new()`, so this change is currently only relevant for a small minority of our test suite.

### After (`TestDatabase`)

Instead of creating new schemas for each test, `TestDatabase::new()` creates completely new databases for each test. This sounds slower at first glance, and with a naive implementation it certainly is. Creating a database takes around 60-70ms, while a new schema can be created in just 1ms.

The advantage of creating a new database is that they can be created based on template databases. This PR implements it so that on first request a template database is created, the migrations applied and for any subsequent request that template database is used to create new databases. Additionally, the corresponding `TemplateDatabase` struct keeps a pool of database management connections open to avoid the delay of first having to connect to the database.

The creation of the template database takes around 500ms, but then every subsequent test database creation is only taking around 70ms, compared to the ~400ms for `FreshSchema::new()`.

This is roughly based upon https://github.com/allaboutapps/integresql#approach-3a-isolation-by-templates, and that document is giving pointers on how the performance could be improved even further.

### TL;DR

We trade a slightly higher startup cost for a significant speed-up per test.


Turbo87:
@rust-lang/crates-io a sanity check on this would be greatly appreciated :)

src/tests/util/test_database.rs
```diff
@@ -0,0 +1,158 @@
...
+        &INSTANCE
+    }
+
+    #[instrument]
+    fn new() -> Self {
+        let base_url = env("TEST_DATABASE_URL");
+        let base_url = Url::parse(&base_url).expect("failed to parse TEST_DATABASE_URL");
+
+        let prefix = base_url.path().strip_prefix('/');
+        let prefix = prefix.expect("failed to parse database name").to_string();
```
Turbo87, line 31: this is currently unconditionally using the `TEST_DATABASE_URL` as the management connection URL and also as the prefix for the generated template and test databases. I've kept it like this for now for backwards compatibility purposes. We could potentially also think about using `DATABASE_URL` as the management connection and have a separate `TEST_DATABASE_PREFIX` option, but I think it's sufficient to do that in a follow-up PR if we decide to do so.

LawnGnome: I think this is fine, but maybe we document this in `docs/CONTRIBUTING.md` as well.

src/tests/util/test_database.rs
```diff
@@ -0,0 +1,158 @@
...
+            .min_idle(Some(0))
+            .build_unchecked(ConnectionManager::new(base_url.as_ref()));
+
+        // Get a connection from the pool, and create the template database
+        let mut conn = pool.get().expect("failed to connect to the database");
+
+        let template_name = format!("{prefix}_template");
+        let _ = drop_database(&template_name, &mut conn);
+        create_template_database(&template_name, &mut conn)
+            .expect("failed to create template database");
```
LawnGnome, line 47: I'm wondering if it would be better to have a random per-test-run prefix to allow for multiple test runs in parallel? Since we are dropping the template database on `Drop` anyway, I don't think this adds much risk of polluting the user's database instance.

LawnGnome (approved):
Couple of nit/questions, but LGTM. Looking forward to speedier tests!

src/tests/util/test_database.rs
Turbo87: ✅

src/tests/util/test_database.rs
Turbo87: ✅

PR #7406, opened 2023-10-31
worker: Extract job structs

Turbo87:
Buckle up, this is a big one!

I recommend reviewing this commit-by-commit and with the "Hide whitespace" option turned on, to ignore whitespace-only indentation changes.

The goal of this PR is to unify how our background job structs look like. After this PR they all have an associated `JOB_NAME` constant and a `run()` fn, all with the same arguments. This allows us to generate the `perform()` fn with the existing `jobs!` macro.

And if you now think "this looks a lot like it should be trait definition" then you are totally right. Extracting a corresponding trait and making the runner generic over the registered jobs will come in a follow-up PR :)

As a side-effect the move to the `JOB_NAME` constant allows us to get rid of the `paste` dependency again and significantly simplify the `jobs!` macro.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -771,7 +771,6 @@ dependencies = [
  "once_cell",
  "p256",
  "parking_lot",
- "paste",
  "prometheus",
  "rand",
  "reqwest",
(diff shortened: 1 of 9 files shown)


LawnGnome (approved):
Looking forward to the trait, but this all makes sense to me as an interim step. If I ever have to understand `jobs!`, this should make it much clearer. :laughing:

PR #7407, opened 2023-10-31
dump_db: improve README based on #7030

LawnGnome:
I think @szabgab is right that we should have this included in the tarball.


Turbo87:
@LawnGnome is https://github.com/rust-lang/crates.io/pull/7307 not sufficient?

LawnGnome:
@Turbo87 I did see that PR before opening this, but I don't think so, because it's in a different place that someone who goes straight to the CSVs is unlikely to see.

PR #7422, opened 2023-11-01
Use fresh databases for all tests

Turbo87:
soooo... yeah... this is slower...

but now we can actually have multiple connections to the database during a test, including a background worker with a dedicated connection pool. in other words: the test are now a lot closer to reality than they were before, and it allows us to remove quite a bit of complexity related to our current "max one database connection per test" limit.

Admittedly the testing perf hit is quite significant, but given how much the database pooling code can be simplified it seems worth it. This also allows us to get rid of the `build.rs` script, since no code is actually needing the test database to have applied migrations now, because everything is happening lazily in the `TemplateDatabase` code.


LawnGnome (approved):
OK, this seems to do the job as expected, so :+1: for the simplification.

It's definitely quite a bit slower, though — looks like everything takes about 2½x as long now on my laptop, compared to `main`. Probably just in the category of "well, it is what it is" at this point. :shrug:

Turbo87:
we've discussed this in the crates.io team meeting yesterday again and concluded that while the performance decrease is unfortunate it unlocks quite a bit of cleanup and simplification of the background worker code and so we will go ahead and merge this.

PR #7435, opened 2023-11-02
Fix IP-based request blocking

Turbo87:
As https://github.com/rust-lang/crates.io/pull/7434 explains, IP-based request blocking is currently broken. This PR fixes the blocking based on the `BLOCKED_IPS` environment variable that we use on production.


LawnGnome (approved):
It occurs to me that we should probably document `BLOCKED_IPS` somewhere at some point along with its exact semantics (comma separated, whitespace _not_ ignored, unless we add a `str::trim` to the chain where we collect the `HashSet` from the environment variable), but I don't think that's important in this PR.

We should probably also add some tests for the blocking eventually, since this is all in-process now and we should be able to do so. Again, not urgent right now.

PR #7441, opened 2023-11-03
cloudfront: Use official AWS SDK

Turbo87:
As #7439 is causing quite some issues, I thought it would probably be easiest to use the high(ish...)-level CloudFront SDK from AWS instead. Turns out their API isn't particularly easy to use either, but I've somehow made it work.

I've just tested this on our staging server, and everything appears to work as intended.

### DIFF
--- a/src/cloudfront.rs
+++ b/src/cloudfront.rs
@@ -1,45 +1,50 @@
-use anyhow::Context;
-use aws_sigv4::http_request::{self, SignableRequest, SigningSettings};
-use aws_sigv4::SigningParams;
-use reqwest::blocking::Client;
+use aws_credential_types::Credentials;
+use aws_sdk_cloudfront::config::Region;
+use aws_sdk_cloudfront::error::SdkError;
+use aws_sdk_cloudfront::types::{InvalidationBatch, Paths};
+use aws_sdk_cloudfront::{Client, Config};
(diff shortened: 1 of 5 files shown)


src/cloudfront.rs
```diff
@@ -1,45 +1,50 @@
...
     ///
     /// `path` is the path to the file to invalidate, such as `config.json`, or `re/ge/regex`
-    #[instrument(skip(self, client))]
-    pub fn invalidate(&self, client: &Client, path: &str) -> anyhow::Result<()> {
-        let path = path.trim_start_matches('/');
-        let url = format!(
-            "https://cloudfront.amazonaws.com/2020-05-31/distribution/{}/invalidation",
-            self.distribution_id
-        );
-        trace!(?url);
```
JohnTitor, line 42: Seems some trace logging has been removed. Is it intentional?

JohnTitor (approved):
Looks reasonable with one question!

src/cloudfront.rs
Turbo87: with the SDK we don't build the URL ourselves anymore so we can't trace log it

JohnTitor: Ah I see, thanks for clarifying!

PR #7459, opened 2023-11-06
bin/background-worker: Simplify restart behavior

Turbo87:
All instances of our code hitting the `panic!()` on production so far have been because of a read-only database during database maintenance events (see Sentry).

The r2d2 connection pool can easily recover from these kinds of issues, so a restart of the dyno or re-initialization of the connection pool seems unnecessary.

This is emphasized by the fact that the startup time of the background worker in production is currently multiple minutes due to the index repository cloning. It seems bad to restart the background worker after just five unsuccessful connection attempts when it then takes multiple minutes again until the next attempt.

### DIFF
--- a/src/bin/background-worker.rs
+++ b/src/bin/background-worker.rs
@@ -85,21 +85,17 @@ fn main() -> anyhow::Result<()> {
 
     let environment = Arc::new(environment);
 
-    let build_runner = || {
-        let connection_pool = r2d2::Pool::builder()
-            .max_size(10)
-            .min_idle(Some(0))
-            .build_unchecked(ConnectionManager::new(&db_url));
+    let connection_pool = r2d2::Pool::builder()
+        .max_size(10)
+        .min_idle(Some(0))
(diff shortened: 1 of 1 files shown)


LawnGnome (approved):
I think I'm OK with this, but let's think through the possible failure mode here:

If a panic occurs that _isn't_ a read only database panic, and it's something that isn't transient, we might end up in a situation where the background worker just keeps restarting the same jobs indefinitely, right? Presumably we'd eventually detect that through the job queue size getting large enough that a page occurs?

Of course, the first step in response to a page would probably be to restart the background worker, which would not really be any different to the code that exists today, except for the part where someone might have been rolled out of bed.

So, at worst, we might be slowing down the restart of the background worker. (Of course, for a non-database error, that might not resolve the problem anyway, so a human would have been woken up regardless.)

So, yeah, I think this is fine — it won't really slow down response to other panics, and not restarting in the read-only case is definitely an improvement.

PR #7461, opened 2023-11-06
worker: Clone index repository in the background

Turbo87:
We are currently delaying the execution of **any** tasks until the index repository is cloned. This was useful while we were mostly running tasks that needed the repository anyway, but with the introduction of the sparse index, this is no longer the case.

This PR adjusts the code to lazily clone the index when requested, but then also fire off a background thread to start that operation once the background worker has started.

This should allow us to process sparse index updates while the git repository is still being cloned.

Note that due to the way the background worker is implemented this is still not entirely true. While the repository is being cloned a `Mutex` is being held. At some point all five background worker threads might be handling `sync_to_git_index` jobs with all of them waiting to lock that mutex. At this point no other jobs can run anymore since the workers are busy. This can be solved in the future by having multiple queues, with a dedicated queue for jobs that involve the git repository.

### DIFF
--- a/src/bin/background-worker.rs
+++ b/src/bin/background-worker.rs
@@ -78,6 +78,15 @@ fn main() -> anyhow::Result<()> {
 
     let environment = Arc::new(environment);
 
+    std::thread::spawn({
+        let environment = environment.clone();
+        move || {
+            if let Err(err) = environment.lock_index() {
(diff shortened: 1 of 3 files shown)


LawnGnome (approved):
LGTM. The worst case scenario you outlined isn't practically any different to what happens today anyway, so this should be fine.

PR #7475, opened 2023-11-08
fastly: Use async `reqwest` client

Turbo87:
This PR refactors our basic `Fastly` client to use async fns and an async `reqwest` HTTP client.

This will should us to eventually write async test cases. This is currently prevented by the `reqwest::blocking::Client` constructor panicking if executed from within an async context.

The background jobs are not async functions yet, but the `DumpDb` job which uses the Fastly client already has an async runtime to upload the database dump to S3 and invalidate CloudFront, so it is relatively straight-forward to integrate there.


hi-rustin (approved):
Looks good to me. Thanks!

PR #7486, opened 2023-11-09
Use `googletest` crate for assertions

Turbo87:
This allows us:

- to write tests with non-fatal assertions (e.g. when unit testing a function multiple times in the same test with different inputs)
- see better failure output including the expression that was tested and a diff between the actual and expected value
- reduce the confusion of what side of `assert_eq!()` is the actual and expected value
- implement custom matchers to improve assertion expressiveness (e.g. for `assert_that!(response.status(), is_success())` it will show the actual status code instead of a boolean value)

see https://github.com/google/googletest-rust

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -924,6 +924,7 @@ dependencies = [
  "flate2",
  "futures-channel",
  "futures-util",
+ "googletest",
  "hex",
  "http",
  "http-body",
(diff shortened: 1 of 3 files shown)


Turbo87:
I've scoped this PR down to only the "non-fatal assertions" use case for now. Other improvements will come in follow-up PRs.

PR #7506, opened 2023-11-11
Add language spec to code blocks

hi-rustin:
1. Added language spec to code blocks to have better syntax support.
2. In Markdown, the angle brackets (<>) are used to automatically convert a URL into a clickable link. So my linter added it.

[rendered](https://github.com/hi-rustin/crates.io/blob/rustin-patch-code-block/docs/CONTRIBUTING.md)

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -136,8 +136,8 @@ talk to:
 
 | Command                                      | Backend                                                           | Use case                                                |
 | -------------------------------------------- | ----------------------------------------------------------------- | ------------------------------------------------------- |
-| `pnpm start:live`                            | https://crates.io                                                 | Testing UI changes with the full live site's data       |
-| `pnpm start:staging`                         | https://staging-crates-io.herokuapp.com                           | Testing UI changes with a smaller set of realistic data |
+| `pnpm start:live`                            | <https://crates.io>                                               | Testing UI changes with the full live site's data       |
+| `pnpm start:staging`                         | <https://staging-crates-io.herokuapp.com>                         | Testing UI changes with a smaller set of realistic data |
 | `pnpm start`                                 | Static fixture test data in `mirage/fixtures`                     | Setting up particular situations, see note              |
 | `pnpm start:local`                           | Backend server running locally                                    | See the Working on the backend section for setup        |
 | `PROXY_BACKEND=https://crates.io pnpm start` | Whatever is specified in the `PROXY_BACKEND` environment variable | If your use case is not covered here                    |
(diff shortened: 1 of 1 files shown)


hi-rustin:
The comparison:

Before:
image
image

After:

image

image

docs/CONTRIBUTING.md
```diff
@@ -293,7 +293,7 @@ This will install a binary named `diesel`, so you should be able to run `diesel
 --version` to confirm successful installation.
 
 > If you're on Linux and this fails with an error that looks like `error:
-linking with `cc` failed: exit code: 1`, you're probably missing some
+linking with`cc`failed: exit code: 1`, you're probably missing some
```
Turbo87, line 296: why did you remove the spaces?

docs/CONTRIBUTING.md
```diff
@@ -194,11 +194,11 @@ methods we'd recommend for each operating system:
...
-  >     ```
-  >     error: linking with `cc` failed: exit code: 1
-  >     [lots of output]
-  >     = note: /usr/bin/ld: cannot find -l[something]
-  >     ```
+  >```console
+  >error: linking with `cc` failed: exit code: 1
+  >[lots of output]
+  >= note: /usr/bin/ld: cannot find -l[something]
+  >```
```
Turbo87, line 201: This code block is text output. My understanding is that console code blocks are specifically for commands that you enter into the console, no?

docs/CONTRIBUTING.md
hi-rustin: Good catch. I think the linter thought it was `a` and `b`, but it actually is `a and b`. I will fix it.

hi-rustin: Updated.

docs/CONTRIBUTING.md
hi-rustin: I think the console also contains what is printed out in your console(stdout + stderr). But I've updated it.

hi-rustin (commented):
🔢 Self-check

docs/CONTRIBUTING.md
```diff
@@ -136,8 +136,8 @@ talk to:
...
-| `pnpm start:staging`                         | https://staging-crates-io.herokuapp.com                           | Testing UI changes with a smaller set of realistic data |
+| `pnpm start:live`                            | <https://crates.io>                                                 | Testing UI changes with the full live site's data       |
+| `pnpm start:staging`                         | <https://staging-crates-io.herokuapp.com>                           | Testing UI changes with a smaller set of realistic data |
 | `pnpm start`                                 | Static fixture test data in `mirage/fixtures`                     | Setting up particular situations, see note              |
 | `pnpm start:local`                           | Backend server running locally                                    | See the Working on the backend section for setup        |
 | `PROXY_BACKEND=https://crates.io pnpm start` | Whatever is specified in the `PROXY_BACKEND` environment variable | If your use case is not covered here                    |
```
Marcono1234, line 143: Should the indentation here then be adjusted? The last column of these rows is shifted compared to the other rows:
Column shifted in diff

hi-rustin: Good catch! Updated.

hi-rustin:
@Turbo87 Ping~ Could you please take another look when you have time? Thanks!

docs/CONTRIBUTING.md
Turbo87: you can escape backticks by doubling the number of outer backticks.

for example: 

```
``foo`bar`baz``
```

will render as ``foo`bar`baz``

I've added a small commit on top that adjusts this :)

PR #7521, opened 2023-11-13
Add debug assert to crates-admin cli

hi-rustin:
- Added debug assert test for the admin CLI.

### DIFF
--- a/src/bin/crates-admin.rs
+++ b/src/bin/crates-admin.rs
@@ -53,3 +53,9 @@ fn main() -> anyhow::Result<()> {
         Command::EnqueueJob(command) => enqueue_job::run(command),
     }
 }
+
+#[test]
+fn verify_cli() {
+    use clap::CommandFactory;
+    Command::command().debug_assert();
+}


Cargo.toml
```diff
@@ -109,3 +109,4 @@ googletest = "=0.10.0"
 hyper-tls = "=0.5.0"
 insta = { version = "=1.34.0", features = ["json", "redactions"] }
 tokio = "=1.34.0"
+trycmd = "0.14.19"
```
Turbo87, line 112: would it be possible to use `insta` for this instead? seems unfortunate to add that many new dev dependencies just for asserting the help output of an internal CLI tool 😅

hi-rustin: > seems unfortunate to add that many new dev dependencies

Can you tell me about your concert? Are you concerned about the compilation time?

Turbo87: > But then we would need to handle how to obtain the binary and get the stdout and stderr outputs ourselves.

does `clap` have an API to get the help output as a `String` instead?

> This would help us avoid breaking anything during the big version upgrade.

what big version upgrade are we talking about? 😅 

are there breaking changes on the horizon that would cause behavior changes without compilation errors?

> Are you concerned about the compilation time?

yeah, exactly.

I'm not sure yet whether the tradeoff is worth it. If this was a public application where people expect a stable CLI interface then it's definitely worth it, but since this application is an internal tool I'm not sure we need to care that much about interface stability. One notable exception are the subcommands that are configured as cronjobs at Heroku, but those subcommands are very simple and I would be surprised if a clap upgrade broke those without causing compilation errors.

hi-rustin: > what big version upgrade are we talking about? 😅

The major version change. Sorry for my bad English. In Chinese, we call the `BIG` version upgrade.



> yeah, exactly.

It increased the building time by about 10s. Anyway, I will drop it.

hi-rustin: > does `clap` have an API to get the help output as a `String` instead?

It has a `render_help` API. It will return a StyledStr. I will try it in the future. (If you think it is worth to give a try 😅)

src/bin/crates-admin.rs
```diff
@@ -53,3 +53,9 @@ fn main() -> anyhow::Result<()> {
         Command::EnqueueJob(command) => enqueue_job::run(command),
     }
 }
+
+#[test]
+fn verify_cli() {
+    use clap::CommandFactory;
+    Command::command().debug_assert();
```
Turbo87, line 60: I was curious what `debug_assert()` is doing. It looks like it runs all of https://docs.rs/clap_builder/4.4.8/src/clap_builder/builder/debug_asserts.rs.html, but only when `debug_assertions` is enabled, which should be the case when running the test suite. In other words, `debug_assert()` apparently enables some additional checks that won't even be caught when running the binary from a release build.

tl;dr LGTM 👍

Cargo.toml
Turbo87: it looks like https://docs.rs/clap_builder/4.4.8/clap_builder/builder/struct.StyledStr.html implements `Display` and then produces a string without ANSI escape codes. seems perfect for using it with `insta::assert_display_snapshot!(...)` :)

PR #7549, opened 2023-11-17
Disable non-canonical downloads if `REJECT_NON_CANONICAL_DOWNLOADS` env var is set

Turbo87:
see https://blog.rust-lang.org/2023/10/27/crates-io-non-canonical-downloads.html

This PR implements a feature flag that can be used to disable support for non-canonical downloads in our backend API. As the blog post mentions, next Monday we plan to disable support for them and roughly a month later we will remove the supporting code paths and the custom error message.

### DIFF
--- a/src/controllers/version/downloads.rs
+++ b/src/controllers/version/downloads.rs
@@ -10,6 +10,7 @@ use crate::models::{Crate, VersionDownload};
 use crate::schema::*;
 use crate::views::EncodableVersionDownload;
 use chrono::{Duration, NaiveDate, Utc};
+use std::fmt::Display;
 use tokio::runtime::Handle;
 use tracing::Instrument;
 
@@ -85,6 +86,13 @@ pub async fn download(
                         .downloads_non_canonical_crate_name_total
                         .inc();
                     req.request_log().add("bot", "dl");
+
+                    if app.config.reject_non_canonical_downloads {
+                        return Err(Box::new(NonCanonicalDownload {
+                            requested_name: crate_name,
+                            canonical_name: canonical_crate_name,
+                        }));
+                    }
                 } else {
                     // The version_id is only cached if the provided crate name was canonical.
                     // Non-canonical requests fallback to the "slow" path with a DB query, but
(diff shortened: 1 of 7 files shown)


src/controllers/version/downloads.rs
```diff
@@ -141,6 +149,34 @@ pub async fn download(
     }
 }
 
+#[derive(Debug)]
+// #[error("see https://blog.rust-lang.org/2023/10/27/crates-io-non-canonical-downloads.html")]
```
hi-rustin, line 153: I am not sure do we need to remove this line.

hi-rustin (approved):
Looks good to me. Thanks!

LawnGnome (approved):
Implementation looks good; small bikeshed on the exact error message.

src/controllers/version/downloads.rs
```diff
@@ -141,6 +149,34 @@ pub async fn download(
...
+
+impl Display for NonCanonicalDownload {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "You requested to downloaded a version of the `{requested_name}` \
+            crate, but that crate is named `{canonical_name}`. Support for \
+            \"non-canonical\" downloads has been deprecated and disabled. \
+            See https://blog.rust-lang.org/2023/10/27/crates-io-non-canonical-downloads.html \
+            for more details.",
```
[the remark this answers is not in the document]
Turbo87: no worries, makes sense. and I guess the test should use a snapshot assertion anyway.

src/controllers/version/downloads.rs
Turbo87: oooops, yeah, that was from an earlier implementation attempt and I somehow forgot to remove it 😂

Turbo87: I originally wanted to derive `thiserror::Error` for the struct, but since there is a blanket implementation or `AppError` for `Error` things didn't quite work as intended, so I had to remove the `thiserror` derive again.

hi-rustin (approved):

PR #7571, opened 2023-11-21
typosquat: add suffix checks

LawnGnome:
This extends our new typosquatting checks (see #7206) to detect an attack vector we've seen more recently where a bad actor tries to squat an existing, popular crate by adding or removing a common suffix (such as `-rs` or `-sys`).

The suffix list in the configuration has been taken _approximately_ from the most popular suffixes in the existing set of crates, with a small amount of human judgement involved on which ones are more likely to be abused based on recent incidents.

Most of this PR is actually tests by weight.

### DIFF
--- a/src/typosquat/cache.rs
+++ b/src/typosquat/cache.rs
@@ -7,7 +7,7 @@ use typomania::{
     Harness,
 };
 
-use super::{config, database::TopCrates};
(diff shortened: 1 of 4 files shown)


Turbo87 (approved):
looks good on first glance 👍

PR #7578, opened 2023-11-21
Add blog post link to `max_features` error messages

Turbo87:
Most people contacting us about the `max_features` restriction apparently did not see or read the blog post yet that explains the introduction of the limit. Adding the URL to the blog post should give people more context before they send that email to help@crates.io.

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -231,9 +231,13 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
         if num_features > max_features {
             return Err(cargo_err(format!(
                 "crates.io only allows a maximum number of {max_features} \
-                features, but your crate is declaring {num_features} features. \
-                If you have a valid use case needing more features, please \
-                send us an email to help@crates.io to discuss the details."
+                features, but your crate is declaring {num_features} features.\n\
+                \n\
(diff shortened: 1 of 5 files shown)


src/controllers/krate/publish.rs
```diff
@@ -246,9 +250,13 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
-                    {num_features} features or dependencies. If you have a \
-                    valid use case needing to increase this limit, please send \
-                    us an email to help@crates.io to discuss the details."
+                    {num_features} features or dependencies.\n\
+                    \n\
+                    Take a look at https://blog.rust-lang.org/2023/10/26/broken-badges-and-23k-keywords.html \
+                    to understand why this restriction was introduced.\n\
+                    \n\
+                    If you have a valid use case needing to increase this limit, \
+                    please send us an email to help@crates.io to discuss the details."
```
LawnGnome, line 259: Sorry, I obviously missed this last time:

```suggestion
                    If you have a use case that requires an increase of this limit, \
                    please send us an email at help@crates.io to discuss the details."
```

Dropped "valid" because it felt redundant — presumably nobody who contacts us thinks their use case is invalid in the first place. :slightly_smiling_face: 

Snapshots will need to be updated as well.

LawnGnome (approved):
Yeah, we should definitely do this. :+1:

PR #7586, opened 2023-11-22
Rearchitect `swirl::Runner` implementation

Turbo87:
Instead of calling `run_all_pending_jobs()` in a loop, which then fills up the `threadpool` to its maximum capacity, we now call `start()`, which starts up `num_workers` background worker threads and then `wait_for_shutdown()` can be used to wait for the worker threads to finish. In normal operation these thread should not finish on their own. In testing mode, when `shutdown_when_queue_empty()` has been used, the worker threads will shutdown once they can't find any pending jobs in the queue anymore.

This change also removes the `Event` enum and the related channel code, since the new implementation does not require communication between the workers and the runner anymore.

Rearchitecting the runner like this should make it a bit easier to replace our usage of threads with tokio tasks in the future.

And sorry, I couldn't figure out how to make the diff easier to digest... 😞

### DIFF
--- a/src/worker/swirl/runner.rs
+++ b/src/worker/swirl/runner.rs
@@ -1,18 +1,15 @@
 use crate::db::{DieselPool, DieselPooledConn, PoolError};
-use crate::worker::swirl::errors::FetchError;
 use crate::worker::swirl::{storage, BackgroundJob};
 use anyhow::anyhow;
 use diesel::prelude::*;
-use diesel::result::Error::RollbackTransaction;
 use std::any::Any;
 use std::collections::HashMap;
 use std::panic::{catch_unwind, AssertUnwindSafe, PanicInfo};
-use std::sync::mpsc::{sync_channel, SyncSender};
 use std::sync::Arc;
+use std::thread::JoinHandle;
 use std::time::Duration;
-use threadpool::ThreadPool;
 
-const DEFAULT_JOB_START_TIMEOUT: Duration = Duration::from_secs(30);
+const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(1);
 
 type RunTaskFn<Context> = dyn Fn(&Context, serde_json::Value) -> anyhow::Result<()> + Send + Sync;
 
@@ -26,33 +23,38 @@ fn runnable<J: BackgroundJob>(ctx: &J::Context, payload: serde_json::Value) -> a
 /// The core runner responsible for locking and running jobs
 pub struct Runner<Context> {
     connection_pool: DieselPool,
-    thread_pool: ThreadPool,
+    num_workers: usize,
     job_registry: JobRegistry<Context>,
     context: Context,
-    job_start_timeout: Duration,
+    poll_interval: Duration,
+    shutdown_when_queue_empty: bool,
 }
 
 impl<Context: Clone + Send + 'static> Runner<Context> {
     pub fn new(connection_pool: DieselPool, context: Context) -> Self {
         Self {
             connection_pool,
-            thread_pool: ThreadPool::new(1),
+            num_workers: 1,
             job_registry: Default::default(),
             context,
-            job_start_timeout: DEFAULT_JOB_START_TIMEOUT,
+            poll_interval: DEFAULT_POLL_INTERVAL,
+            shutdown_when_queue_empty: false,
         }
     }
 
+    /// Set the number of workers to spawn.
     pub fn num_workers(mut self, num_workers: usize) -> Self {
-        self.thread_pool.set_num_threads(num_workers);
+        self.num_workers = num_workers;
         self
     }
 
-    pub fn job_start_timeout(mut self, job_start_timeout: Duration) -> Self {
-        self.job_start_timeout = job_start_timeout;
+    /// Set the interval after which each worker polls for new jobs.
+    pub fn poll_interval(mut self, poll_interval: Duration) -> Self {
+        self.poll_interval = poll_interval;
         self
     }
 
+    /// Register a new job type for this job runner.
     pub fn register_job_type<J: BackgroundJob<Context = Context>>(mut self) -> Self {
         self.job_registry
             .insert(J::JOB_NAME.to_string(), Arc::new(runnable::<J>));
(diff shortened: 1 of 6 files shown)


src/worker/swirl/runner.rs
```diff
@@ -26,155 +23,153 @@ fn runnable<J: BackgroundJob>(ctx: &J::Context, payload: serde_json::Value) -> a
...
-    /// Waits for all running jobs to complete, and returns an error if any
-    /// failed
+    /// Check if any jobs in the queue have failed.
     ///
-    /// This function is intended for use in tests. If any jobs have failed, it
-    /// will return `swirl::JobsFailed` with the number of jobs that failed.
-    ///
-    /// If any other unexpected errors occurred, such as panicked worker threads
-    /// or an error loading the job count from the database, an opaque error
-    /// will be returned.
```
LawnGnome, line 122: While this was apparently wrong about the details before, do we want to retain the explicit documentation that an error gets returned when the check fails? (I guess it's _fairly_ obvious from the function signature, but I'm always conscious of the diminished brainpower in late night coding sessions. Or, in my case, early morning ones.)

src/worker/swirl/runner.rs
```diff
@@ -26,155 +23,153 @@ fn runnable<J: BackgroundJob>(ctx: &J::Context, payload: serde_json::Value) -> a
...
-            let jobs_to_queue = if pending_messages == 0 {
-                // If we have no queued jobs talking to us, and there are no
-                // available threads, we still need to queue at least one job
-                // or we'll never receive a message
-                max(available_threads, 1)
-            } else {
-                available_threads
-            };
+    /// This returns a `RunningRunner` which can be used to wait for the workers to shutdown.
+    pub fn start(&self) -> anyhow::Result<RunningRunner> {
```
LawnGnome, line 74: [`MFW `RunningRunner.`](https://en.wikipedia.org/wiki/Neville_Neville#Name)

Bikeshed: what about `RunHandle`?

src/worker/swirl/runner.rs
```diff
@@ -26,155 +23,153 @@ fn runnable<J: BackgroundJob>(ctx: &J::Context, payload: serde_json::Value) -> a
...
+    /// Wait for all background workers to shut down.
+    pub fn wait_for_shutdown(self) {
+        self.handles.into_iter().for_each(|handle| {
+            if let Err(error) = handle.join() {
+                let error = try_to_extract_panic_info(&error);
+                warn!(%error, "Background worker thread panicked");
+            }
+        })
     }
 }
```
LawnGnome, line 130: I can definitely see how this would be adapted to a set of tasks (\*cough\* `join_all` \*cough\*). :+1:

src/worker/swirl/runner.rs
```diff
@@ -26,155 +23,153 @@ fn runnable<J: BackgroundJob>(ctx: &J::Context, payload: serde_json::Value) -> a
...
-                Ok(Some(j)) => {
-                    let _ = self.sender.send(Event::Working);
-                    j
+    /// Run background jobs forever, or until the queue is empty if `shutdown_when_queue_empty` is set.
+    pub fn run(&self) {
+        loop {
+            match self.run_next_job() {
+                Ok(Some(_)) => {}
+                Ok(None) if self.shutdown_when_queue_empty => {
+                    debug!("No pending background worker jobs found. Shutting down the worker…");
```
LawnGnome, line 147: Maybe a `trace!` instead, just to limit ambient noise? (At a skim, I don't think we currently log in this case.)

src/worker/swirl/runner.rs
```diff
@@ -26,155 +23,153 @@ fn runnable<J: BackgroundJob>(ctx: &J::Context, payload: serde_json::Value) -> a
...
-                    return Err(RollbackTransaction);
+                Err(error) => {
+                    error!(%error, "Failed to run job");
+                    std::thread::sleep(self.poll_interval);
                 }
+            }
+        }
+    }
+
+    /// Run the next job in the queue, if there is one.
```
LawnGnome, line 165: Let's document the return values here, since I spent more than one second thinking about it when reviewing the `match` above:

```suggestion
    /// Run the next job in the queue, if there is one.
    ///
    /// Returns:
    /// - `Ok(Some(job_id))` if a job was run successfully
    /// - `Ok(None)` if no jobs were waiting
    /// - `Err(...)` if there was an error retrieving the job, or the job returned an error
```

src/worker/swirl/runner.rs
```diff
@@ -26,155 +23,153 @@ fn runnable<J: BackgroundJob>(ctx: &J::Context, payload: serde_json::Value) -> a
...
+    fn run_next_job(&self) -> anyhow::Result<Option<i64>> {
+        let conn = &mut *self.connection_pool.get()?;
+
+        conn.transaction(|conn| {
+            debug!("Looking for next background worker job…");
```
LawnGnome, line 170: Again, maybe a `trace!` in practice?

src/worker/swirl/runner.rs
```diff
@@ -26,155 +23,153 @@ fn runnable<J: BackgroundJob>(ctx: &J::Context, payload: serde_json::Value) -> a
...
                 }
+            }
+        }
+    }
+
+    /// Run the next job in the queue, if there is one.
+    fn run_next_job(&self) -> anyhow::Result<Option<i64>> {
+        let conn = &mut *self.connection_pool.get()?;
+
+        conn.transaction(|conn| {
```
LawnGnome, line 169: I think hiding the transaction handling in `run_next_job` is a good call, and makes this more readable. :+1:

LawnGnome (approved):
LGTM. Nice job!

One question I do have — that I could go answer myself, but I'm going to be lazy[^lazy] about is: assuming I'm reading the code correctly, if a test app runner is started without jobs in the queue and the test queues jobs after, the effective time that the runner will wait for a job has been reduced from 5s to 1s. Are we reasonably confident that no test cases rely on that, and that they always enqueue jobs before starting the runner? Just trying to avoid any non-deterministic test failures in the future.

[^lazy]: I could pretend that it's some devious plan to make sure you'd thought about this, but really it's because you probably already know the answer and, again, I'm lazy.

src/worker/swirl/runner.rs
Turbo87: I tend to use `trace!` only for logging random, internal state variables and less for real log messages. Do you know of good guidelines on when to use what log level?

I'm think I'll keep the `debug!` for now since we only use `trace!` in one other place in the codebase.

Turbo87:
> if a test app runner is started without jobs in the queue and the test queues jobs after, the effective time that the runner will wait for a job has been reduced from 5s to 1s

that's a good catch, and I think you're right. it may even have been reduced to 0s, since the worker will quit immediately if the queue is empty.

OTOH we don't run many things in parallel in our test suite and with a linear execution model this probably shouldn't be a problem in practice. I've run the test suite a couple of times while developing this and haven't seen any such issues so far.

PR #7597, opened 2023-11-23
worker/swirl/runner: Spawn worker threads on tokio runtime

Turbo87:
Do I know what I'm doing? Barely...

Does it work? Apparently!? 😅 

This PR essentially replaces the `std::thread::spawn()` call in the `start()` fn with `tokio::task::spawn_blocking()`. This means that all jobs are now executed within the context of a tokio runtime, which allows us to get rid of the per-job runtime creation that was happening in some of the jobs. This will hopefully also allow us to make the `run()` fn of the `BackgroundJob` trait async in the future.

### DIFF
--- a/src/bin/background-worker.rs
+++ b/src/bin/background-worker.rs
@@ -13,6 +13,7 @@
 #[macro_use]
 extern crate tracing;
 
+use anyhow::Context;
 use crates_io::cloudfront::CloudFront;
 use crates_io::db::DieselPool;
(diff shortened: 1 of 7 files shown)


LawnGnome (approved):
:+1:

Obviously it'll be nice to get to a point where we can have `async fn` in all the things and not have to handle explicit runtimes and handles around, but this is a good step on the road towards that.

PR #7601, opened 2023-11-24
worker/swirl/background_job: Change `run()` fn to be async

Turbo87:
This PR makes all of our background worker jobs async functions. Most of them involve database operations though, so some usage of `spawn_blocking()` is still needed, but it should make it easier to convert more and more parts of our codebase to async/await in the future.


LawnGnome (approved):
Definitely grateful for the "hide whitespace" option on this one. :laughing: 

Looks good, and I'm pleased to get rid of a bunch of explicit `Handle`s. It might look better some day when we have `async` Diesel, but I think the benefits here are obvious even without that.

PR #7611, opened 2023-11-27
Use `async fn` and return-position `impl Trait` from std library for background jobs

hi-rustin:
Follow up https://github.com/rust-lang/crates.io/pull/7601

Because the async trait is already stabled in the std library, I think we probably can switch to using it. Then probably we can get rid of one depends.

### DIFF
--- a/crates_io_worker/src/background_job.rs
+++ b/crates_io_worker/src/background_job.rs
@@ -1,15 +1,14 @@
 use crate::errors::EnqueueError;
 use crate::schema::background_jobs;
-use async_trait::async_trait;
 use diesel::prelude::*;
 use diesel::PgConnection;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
+use std::future::Future;
 use tracing::instrument;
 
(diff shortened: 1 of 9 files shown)


hi-rustin:
Oops. I forget we need to wait for the next stable release.

Turbo87:
looks like this needs a rebase :)

crates_io_worker/src/background_job.rs
```diff
@@ -27,8 +26,8 @@ pub trait BackgroundJob: Serialize + DeserializeOwned + Send + Sync + 'static {
     /// The application data provided to this job at runtime.
     type Context: Clone + Send + 'static;
 
-    /// Execute the task. This method should define its logic
-    async fn run(&self, ctx: Self::Context) -> anyhow::Result<()>;
+    /// Execute the task. This method should define its logic.
+    fn run(&self, ctx: Self::Context) -> impl Future<Output = anyhow::Result<()>> + Send;
```
Turbo87, line 30: `async fn` is apparently not possible here yet due to the necessary `Send` bound

PR #7621, opened 2023-11-27
worker: Move `block_on()` call further outside

Turbo87:
This PR converts more of our worker code to async/await by moving the `block_on()` call further to the outside. Each new layer inside is converted to async/await accordingly.


hi-rustin (approved):
Looks good to me. thanks!

LawnGnome (approved):
I'm kinda taking some of the Sentry API details on faith, but it seems reasonable to me. :+1:

PR #7627, opened 2023-11-27
Remove unused `hyper-tls` dependency

Turbo87:
This is used internally by `reqwest`, but there appears to be no need for us to have this as an explicit dependency.


Turbo87:
git archeology reveals that this was used by the HTTP recorder proxy implementation, which we are no longer using anywhere.

hi-rustin (approved):

PR #7640, opened 2023-11-29
Add support for running multiple background worker queues

Turbo87:
We are currently running a single queue with 5 workers in production. Some of our background jobs require exclusive access to the cloned git index repository, which is handled by a `Mutex`. This means that we can run into situations where one worker is working on the repository and the other four workers are waiting for the mutex, while they could instead work on other jobs that don't require repository access.

This PR adds a `QUEUE` field to the `BackgroundJob` trait, which allows us to define on which queue a job should run. These queues can be configured via `configure_queue()` and can have different numbers of workers.

This PR then makes use of the new `QUEUE` field and adds a `repository` queue for all three jobs that interact with the git index repository.

### DIFF
--- a/crates_io_worker/src/background_job.rs
+++ b/crates_io_worker/src/background_job.rs
@@ -7,6 +7,8 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 use tracing::instrument;
 
+pub const DEFAULT_QUEUE: &str = "default";
+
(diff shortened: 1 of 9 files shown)


LawnGnome (approved):
Took me a little bit to track the new queue indirection through, but it looks like it holds together to me. :+1:

PR #7651, opened 2023-12-03
worker/worker: Convert `run()` fns to async

Turbo87:
This PR converts the main `Worker::run()` fn to async/await, and wraps most of the `run_next_job()` fn content in a `spawn_blocking()` call.

This should get us one step closer to using an async database connection pool.

### DIFF
--- a/crates_io_worker/src/util.rs
+++ b/crates_io_worker/src/util.rs
@@ -28,7 +28,7 @@ where
     F: FnOnce() -> Fut,
     Fut: Future<Output = Result<R, E>>,
 {
-    let hub = Hub::current();
+    let hub = Hub::new_from_top(Hub::current());
     let _scope_guard = hub.push_scope();
 
     let tx_ctx = sentry_core::TransactionContext::new(transaction_name, "swirl.perform");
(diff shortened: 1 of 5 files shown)


Turbo87:
> Popped scope guard out of order

hmm, looks like I have some more work to do here 🤔

crates_io_worker/src/util.rs
```diff
@@ -3,6 +3,22 @@ use sentry_core::Hub;
...
+    R: Send + 'static,
+    E: Send + From<JoinError> + 'static,
+{
+    let hub = Hub::current();
+    tokio::task::spawn_blocking(move || Hub::run(hub, f))
+        .await
+        // Convert `JoinError` to `E`
+        .map_err(Into::into)
+        // Flatten `Result<Result<_, E>, E>` to `Result<_, E>`
+        .and_then(std::convert::identity)
```
LawnGnome, line 20: I always forget `identity` exists! (My usual pattern there is `Ok(...await??)`, but I don't think that's actually any better. Just interesting seeing the different style.)

LawnGnome (approved):
Makes sense to me. :+1:

PR #7725, opened 2023-12-14
github: Extract `GitHubError` enum

Turbo87:
This reduces the need for `e.is::<NotFound>()` checks and makes the code a little less coupled to our HTTP layer and its error responses.


LawnGnome (approved):
Makes sense to me, and I definitely agree the code using it in a `match` is cleaner.

PR #7755, opened 2023-12-18
Extract `Email` trait

Turbo87:
Previously we implemented new email variants by adding another method to the `Emails` struct, which then internally called `self.send(...)`.

This PR explores a different implementation where each email variant is represented by its own struct, holding the values that it needs to generate the email body. Each of the structs implements a new `Email` trait, which has a `const SUBJECT` and `fn body()` on it.

This allows us to move the email generation code closer to where the emails are sent and should scale better with an increasing number of email variants. It also makes the email sending code in the `email` module a bit more generic and less coupled to the application logic.

### DIFF
--- a/src/controllers/user/me.rs
+++ b/src/controllers/user/me.rs
@@ -6,6 +6,7 @@ use crate::controllers::frontend_prelude::*;
 use crate::controllers::helpers::*;
 
 use crate::controllers::helpers::pagination::{Paginated, PaginationOptions};
+use crate::email::UserConfirmEmail;
 use crate::models::{
     CrateOwner, Email, Follow, NewEmail, OwnerKind, User, Version, VersionOwnerAction,
 };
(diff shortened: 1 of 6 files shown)


src/email.rs
```diff
@@ -12,10 +12,15 @@ use lettre::transport::stub::StubTransport;
...
+pub trait Email {
+    const SUBJECT: &'static str;
+    fn body(&self) -> String;
+}
+
 #[derive(Debug, Clone)]
 pub struct Emails {
     backend: EmailBackend,
-    domain: String,
+    pub domain: String,
```
Turbo87, line 23: questionable, but good enough for now, I guess :D

src/controllers/user/me.rs
```diff
@@ -157,9 +157,13 @@ pub async fn update_user(
...
-            let _ = state
-                .emails
-                .send_user_confirm(user_email, &user.gh_login, &token);
+            let email = UserConfirmEmail {
+                user_name: &user.gh_login,
+                domain: &state.emails.domain,
+                token: &token,
+            };
+
+            let _ = state.emails.send(user_email, email);
```
LawnGnome, line 166: I realise this is exactly equivalent to the old code, but while we're in here, should we `?` on the result of `state.emails.send()` instead of swallowing it?

src/email.rs
LawnGnome: I'll allow it. :laughing: 

Turning this into some sort of `EmailState` type or whatever that gets passed into `body` and can provide the domain etc would probably be a good first issue for someone.

LawnGnome (approved):
LGTM! :email:

src/email.rs
Turbo87: I like that idea! 👍

src/controllers/user/me.rs
Turbo87: one thing at a time 😉

PR #7776, opened 2023-12-20
Implement cargo compatibility status code enforcement feature flag

Turbo87:
In https://github.com/rust-lang/crates.io/pull/7715 we implemented a middleware that ensured `200 OK` response status codes for all API endpoints that are relevant for cargo (see https://doc.rust-lang.org/cargo/reference/registry-web-api.html).

The PR also mentioned:

> This will allow us to remove cargo_err() and use proper status codes internally. This is particularly useful for helper functions that are used by both cargo- and non-cargo endpoints. It also makes it easier to phase out the 200 OK requirement since the middleware could easily be disabled by a feature flag in the future.

This PR implements such a feature flag, based on the value of a `CARGO_COMPAT_STATUS_CODES` environment variable. If that env var is set to `"n"` it will disable this part of the middleware and return the real status codes instead.

Also, our integration test suite is using this behavior by default now. The middleware has dedicated unit tests which already verify that the middleware works as expected. In the integration tests we can test what would happen if we toggled the status code enforcement off eventually.

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -8,6 +8,7 @@ use crate::Env;
 use super::base::Base;
 use super::database_pools::DatabasePools;
 use crate::config::balance_capacity::BalanceCapacityConfig;
+use crate::middleware::cargo_compat::StatusCodeConfig;
 use crate::storage::StorageConfig;
 use crates_io_env_vars::{required_var, var, var_parsed};
 use http::HeaderValue;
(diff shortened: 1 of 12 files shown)


LawnGnome (approved):
LGTM other than the test failure. :see_no_evil:

Turbo87:
shouldn't have pushed in last minute fixes... 😅

src/middleware/cargo_compat.rs
```diff
@@ -1,11 +1,41 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "disabled" => Ok(Self::Disabled),
+            "adjust-all" => Ok(Self::AdjustAll),
+            "adjust-success" => Ok(Self::AdjustSuccess),
+            _ => Err(StatusCodeConfigError),
+        }
+    }
+}
```
Turbo87, line 34: I've changed the implementation from a simple `bool` to an enum. This allows us to use the original status codes in tests, 200 everywhere in production, and e.g. 200 for 2xx on staging or eventually on production too.

src/config/server.rs
```diff
@@ -221,6 +226,8 @@ impl Server {
             cdn_user_agent: var("WEB_CDN_USER_AGENT")?
                 .unwrap_or_else(|| "Amazon CloudFront".into()),
             balance_capacity: BalanceCapacityConfig::from_environment()?,
+            cargo_compat_status_code_config: var_parsed("CARGO_COMPAT_STATUS_CODES")?
+                .unwrap_or(StatusCodeConfig::AdjustAll),
```
Turbo87, line 230: if no value is set it will default to `AdjustAll` (i.e. convert everything to 200 OK)

PR #7789, opened 2023-12-21
 Minor Performance Enhancements and Fixes for Website 💻

sanjaiyan-dev:
Hi, this pull request contains minor performance improvements and fixes for the website.

1) Making the Open Graph image a full path URL (`/assets/og-image.png` to `https://crates.io/assets/og-image.png`). This change ensures that the Open Graph image works as expected, as Open Graph only supports absolute file paths.

2) Adding the `inputmode` attribute to the search input for better accessibility (Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode#search).

3) Adding the `decoding` attribute to images with async for better image rendering. This is a default option in the popular Next.js framework (Reference: https://nextjs.org/docs/app/api-reference/components/image#other-props).

4) Adding `<link rel="preload" as="script" href="..." />` for faster loading of JavaScript needed for the website (Reference: https://web.dev/articles/preload-critical-assets#preloading_javascript_files).

5) Adding `scroll-behavior: smooth` in CSS for a sleeker user experience.

Extremely sorry, if I made any mistakes :)

PR #7800, opened 2023-12-24
Enhance Image Rendering: Introduce Async Decoding Attribute

sanjaiyan-dev:
Ref-: https://github.com/rust-lang/crates.io/pull/7789

Adding the decoding attribute to images with async for better image rendering. This is a default option in the popular Next.js framework (Reference: https://nextjs.org/docs/app/api-reference/components/image#other-props).

Adding <link rel="preload" as="script" href="..." /> for faster loading of JavaScript needed for the website (Reference: https://web.dev/articles/preload-critical-assets#preloading_javascript_files).

Closes #7789

### DIFF
--- a/app/components/user-avatar.hbs
+++ b/app/components/user-avatar.hbs
@@ -4,5 +4,6 @@
   height={{this.size}}
   alt={{this.alt}}
   title={{this.title}}
+  decoding="async"
   ...attributes
 />


app/index.html
```diff
@@ -25,6 +25,9 @@
 
     <meta property="og:image" content="/assets/og-image.png">
     <meta name="twitter:card" content="summary_large_image">
+
+    <link rel="preload" as="script" href="{{rootURL}}assets/vendor.js" />
+    <link rel="preload" as="script" href="{{rootURL}}assets/crates-io.js" />
```
Turbo87, line 30: what is the performance gain for these?

I understand that they make sense at the top of the file when the file has a lot of server-rendered content, but as long as we have a fairly empty HTML file I'm wondering if they will actually have a positive effect.

sanjaiyan-dev: > what is the performance gain for these?
> 
> I understand that they make sense at the top of the file when the file has a lot of server-rendered content, but as long as we have a fairly empty HTML file I'm wondering if they will actually have a positive effect.

I'm extremely sorry if I am wrong; I thought browsers mark these scripts as high priority. Ref: https://nitropack.io/blog/post/link-rel-preload-explained

Turbo87: Bildschirmfoto 2023-12-25 um 10 17 55

(from https://www.webpagetest.org/result/231225_AiDcZ4_67C/1/details/#waterfall_view_step1)

if I interpret this waterfall diagram correctly, the JS and CSS files all start loading at the same time. since the preload hints in this PR come after the style links, they most likely wouldn't make a big difference compared to relying on the regular `script` tags for loading the files.

my understanding is that preload is only an advantage if a) the server-rendered page has a lot of HTML content in it and the script tags are at the bottom, or b) if e.g. a webfont is referenced in an CSS file but we could start loading the webfont already before loading and processing the CSS file.

sanjaiyan-dev: Hi @Turbo87, should I revert the preload-related change?

Turbo87: I've interactive rebased on top of the `main` branch and dropped the preload commit

sanjaiyan-dev: Thanks

sanjaiyan-dev: Thanks

PR #7801, opened 2023-12-24
Enhance Accessibility and User Experience: Add inputmode attribute to search input and implement smooth scroll-behavior in CSS 📜

sanjaiyan-dev:
Ref-: https://github.com/rust-lang/crates.io/pull/7789

Adding the `inputmode` attribute to the search input for better accessibility (Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode#search).

Adding scroll-behavior: smooth in CSS for a sleeker user experience.

### DIFF
--- a/app/components/search-form.hbs
+++ b/app/components/search-form.hbs
@@ -19,6 +19,7 @@
 
   <input
     type="text"
+    inputmode="search"
     local-class="input-lg"
     name="q"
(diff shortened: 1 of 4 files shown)


Turbo87 (approved):
for future reference: https://css-tricks.com/almanac/properties/s/scroll-behavior/

PR #7812, opened 2023-12-27
util/errors: Extract `CustomApiError` struct

Turbo87:
Instead of having a dedicated `struct` per HTTP status code (and then some more), we can use a single struct and save the status code inside the struct, next to the error message. This allows us to throw away a significant chunk of verbose code while the API behaves exactly the same as before. Admittedly this removes the ability for `e.is::<NotFound>()` checks, but we were not and probably shouldn't be using those anyway.

Note that a few structs were not converted to using this "generic" struct yet:
- `InsecurelyGeneratedTokenRevoked` because it is used in an `e.is::<_>()` check
- `ReadOnlyMode` because it is used in an `e.is::<_>()` check
- `TooManyRequests` because it is setting a custom header, which `CustomApiError` does not currently allow

### DIFF
--- a/src/controllers/metrics.rs
+++ b/src/controllers/metrics.rs
@@ -1,5 +1,5 @@
 use crate::controllers::frontend_prelude::*;
-use crate::util::errors::{forbidden, not_found, MetricsDisabled};
+use crate::util::errors::{custom, forbidden, not_found};
 use prometheus::TextEncoder;
 
 /// Handles the `GET /api/private/metrics/:kind` endpoint.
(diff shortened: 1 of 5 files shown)


LawnGnome (approved):
LGTM.

Losing the ability to assert on specific errors is a _little_ sad, but I assume we could add a way for tests to introspect `CustomApiError` if we really needed that for any of the errors that have been converted. It's probably not worth keeping the boilerplate that's getting removed here.

Turbo87:
> Losing the ability to assert on specific errors is a _little_ sad

if there are cases where the introspection makes sense then I'm not against having custom error structs for that, but in 99% of our code base we don't need it :)

PR #7818, opened 2023-12-28
Remove `AppError::chain()` method

Turbo87:
Chained errors were used in our codebase to print a `cause` field in the log output. `internal(error_message).chain(forbidden())` for example would print `cause=<error_message>` and use `forbidden()` as the HTTP response.

Unfortunately our `AppError` struct is incompatible with `std::error::Error` and its `source()` fn though, which makes our error chaining less useful.

This PR removes/replaces our custom error chaining to make it easier in the future to build one based on `std::error::Error` instead.

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -670,9 +670,9 @@ pub fn add_dependencies(
 impl From<TarballError> for BoxedAppError {
     fn from(error: TarballError) -> Self {
         match error {
-            TarballError::Malformed(err) => err.chain(cargo_err(
+            TarballError::Malformed(_err) => cargo_err(
                 "uploaded tarball is malformed or too large when decompressed",
-            )),
+            ),
             TarballError::InvalidPath(path) => cargo_err(format!("invalid path found: {path}")),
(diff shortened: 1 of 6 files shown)


src/controllers/krate/publish.rs
```diff
@@ -670,9 +670,9 @@ pub fn add_dependencies(
 impl From<TarballError> for BoxedAppError {
     fn from(error: TarballError) -> Self {
         match error {
-            TarballError::Malformed(err) => err.chain(cargo_err(
+            TarballError::Malformed(_err) => cargo_err(
```
LawnGnome, line 673: I'm assuming there's just no real value in the `TarballError` in practice, so it's not worth logging?

LawnGnome (approved):
One question, but this looks good otherwise.

src/controllers/krate/publish.rs
Turbo87: The main cause of this variant is when the size limit is reached, but in that case it's a client error and we don't care that much about logging it since we don't need to act on it on our side

PR #7828, opened 2023-12-28
Replace remaining `TEST_DATABASE_URL` usage with `TestDatabase`

Turbo87:
`TEST_DATABASE_URL` is only supposed to be used as a database name prefix at this point, but not as a real, existing database. This PR adjusts the remaining usages to use the `TestDatabase` struct instead.

This PR is based on https://github.com/rust-lang/crates.io/pull/7827, since `Faker` needs to be made compatible with the pooled database connection that `TestDatabase::connect()` returns.

### DIFF
--- a/src/models/category.rs
+++ b/src/models/category.rs
@@ -145,21 +145,12 @@ impl<'a> NewCategory<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::test_util::pg_connection_no_transaction;
-    use diesel::connection::SimpleConnection;
-
-    fn pg_connection() -> PgConnection {
-        let mut conn = pg_connection_no_transaction();
-        // These tests deadlock if run concurrently
-        conn.batch_execute("BEGIN; LOCK categories IN ACCESS EXCLUSIVE MODE")
-            .unwrap();
-        conn
-    }
+    use crate::test_util::test_db_connection;
 
     #[test]
     fn category_toplevel_excludes_subcategories() {
         use self::categories;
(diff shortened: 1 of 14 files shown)


src/models/category.rs
```diff
@@ -145,21 +145,12 @@ impl<'a> NewCategory<'a> {
...
 mod tests {
     use super::*;
-    use crate::test_util::pg_connection_no_transaction;
-    use diesel::connection::SimpleConnection;
-
-    fn pg_connection() -> PgConnection {
-        let mut conn = pg_connection_no_transaction();
-        // These tests deadlock if run concurrently
-        conn.batch_execute("BEGIN; LOCK categories IN ACCESS EXCLUSIVE MODE")
-            .unwrap();
```
Turbo87, line 155: this is no longer needed since we use a fresh database for each test

src/test_util.rs
```diff
@@ -1,15 +1,14 @@
...
+use diesel::r2d2::{ConnectionManager, PooledConnection};
 
-pub fn pg_connection_no_transaction() -> PgConnection {
-    let database_url = required_var("TEST_DATABASE_URL").unwrap();
-    PgConnection::establish(&database_url).unwrap()
-}
-
-pub fn pg_connection() -> PgConnection {
-    let mut conn = pg_connection_no_transaction();
-    conn.begin_test_transaction().unwrap();
```
Turbo87, line 13: this is also no longer needed due to using a fresh database for each test

LawnGnome (approved):
:+1:

PR #7852, opened 2024-01-02
Introduce `users.is_admin` column and allow admins to yank/unyank versions

Turbo87:
This PR does multiple things, but I felt that without an actual use case for admin users their introduction wouldn't make a lot of sense...

The commit messages should explain the steps, but the short version is that this PR introduces a new `is_admin` column on the `users` table, which default to `false`. We can manually grant this permission for the time being, and later introduce a background job that regularly syncs the data with the https://github.com/rust-lang/team/ repo.

In the next step, this PR allows users with the `is_admin` flag to yank/unyank crate versions that they don't own themselves. This is not granting our admins any new permissions, since such operations have previously been performed through the `crates-admin yank-version` admin tool already, but it reduces the amount of permissions that admin users need to have. Previously they needed full permissions on Heroku to run the admin tool, and now they only need the database flag. The PR also adds explicit logging of such an admin intervention.

The last part in this PR adds an `is_admin` flag to the output of `/api/v1/me` and then uses the field to show the yank/unyank buttons on the frontend when an admin views the version list of a crate they don't own themselves.

Lastly, note that this PR is implementing admin users slightly different from https://github.com/rust-lang/crates.io/pull/6456. After thinking about the various implementation options it felt like a database flag with regular team repo sync might be easier than the environment variable approach, since a team repo sync in that case would need to be performed on server startup to work properly.


src/tests/routes/crates/versions/yank_unyank.rs
```diff
@@ -324,4 +326,26 @@ mod auth {
...
+            diesel::update(admin.as_model())
+                .set(users::is_admin.eq(true))
+                .execute(conn)
+                .unwrap();
+        });
+
+        let response = admin.yank(CRATE_NAME, CRATE_VERSION);
+        assert_eq!(response.status(), StatusCode::OK);
+        assert_eq!(response.into_json(), json!({ "ok": true }));
+
```
LawnGnome, line 346: Can we assert that the yank actually happened (and, below, that the unyank happened)?

LawnGnome (approved):

src/tests/routes/crates/versions/yank_unyank.rs
Turbo87: ✅

PR #7888, opened 2024-01-06
Add a '+' to the MSRV on the crate page

Sky9x:
The MSRV indicates not a specific version requirement, but a "this version or later" requirement (hence the 'minimum supported' in 'minimum supported rust version').
This PR adds a '+' to the end of the version display to indicate this.

Spawned off of https://github.com/badges/shields/pull/9871#discussion_r1443839958

### DIFF
--- a/app/components/crate-sidebar.hbs
+++ b/app/components/crate-sidebar.hbs
@@ -21,7 +21,7 @@
       <div local-class="msrv" data-test-msrv>
         {{svg-jar "rust"}}
         <span>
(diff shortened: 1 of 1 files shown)


Turbo87:
in `Cargo.toml` if you write e.g. `serde = "1.2"` it means v1.2 or higher. the way we represent it on the webpage matches this, so I'm not sure the extra `+` is necessary.

Sky9x:
ok

PR #7889, opened 2024-01-07
controllers/crates: default to sort by recent downloads

punkeel:
"Browse All Crates" is the most prominent link in the menu bar, displayed on all pages. Before this change, this link would present the user with (mostly) irrelevant crates, as it was sorted alphabetically: crates like `a`, `a-`, `a0`, etc.

With this change, the list is now sorted to present the most downloaded crates first.

`/keywords/:term` already defaults to that same order.

### DIFF
--- a/app/controllers/crates.js
+++ b/app/controllers/crates.js
@@ -9,7 +9,7 @@ export default class CratesController extends Controller {
   queryParams = ['page', 'per_page', 'sort'];
   @tracked page = '1';
   @tracked per_page = 50;
-  @tracked sort = 'alpha';
+  @tracked sort = 'recent-downloads';
 
   @reads('model.meta.total') totalItems;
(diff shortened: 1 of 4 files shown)


Turbo87 (approved):
seems like a good idea to me, thanks! :)

Turbo87:
@punkeel it looks like a couple of tests are failing though. could you take a look?

eth3lbert:
Digging into the failure a bit further, to make the test pass, you need to:

1. Make the response reflect the sorting method: Add support for sorting by `recent-downloads` in the `mirage/route-handlers/crates.js` file.
2. Update the corresponding assert value: Since the default sorting method has now changed to `recent-downloads`, we need to update crates sort, `All-Time` and `Recent` values accordingly.

PR #7895, opened 2024-01-08
tests/util: Add `Content-Type: application/json` request headers

Turbo87:
The `axum::Json` extractor requires this header to be present and rejects requests that don't have it. This change ensures we can move our code over from custom JSON parsing to the `axum` extractor eventually.

### DIFF
--- a/src/tests/util.rs
+++ b/src/tests/util.rs
@@ -139,8 +139,15 @@ pub trait RequestHelper {
     /// Issue a PUT request
     #[track_caller]
     fn put<T>(&self, path: &str, body: impl Into<Bytes>) -> Response<T> {
+        let body = body.into();
+        let is_json = body.starts_with(b"{") && body.ends_with(b"}");
+
         let mut request = self.request_builder(Method::PUT, path);
-        *request.body_mut() = body.into();
+        *request.body_mut() = body;
+        if is_json {
+            request.header(header::CONTENT_TYPE, "application/json");
+        }
+
         self.run(request)
     }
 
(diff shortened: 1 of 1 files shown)


Turbo87:
the cargo team confirmed via [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/.60Content-Type.60.20request.20headers) that this header is being sent by cargo and has been for a long time :)

src/tests/util.rs
```diff
@@ -139,8 +139,15 @@ pub trait RequestHelper {
     /// Issue a PUT request
     #[track_caller]
     fn put<T>(&self, path: &str, body: impl Into<Bytes>) -> Response<T> {
+        let body = body.into();
+        let is_json = body.starts_with(b"{") && body.ends_with(b"}");
```
LawnGnome, line 143: I feel like this will eventually cause someone to have an annoying day writing tests if we ever add an endpoint that doesn't expect an object, but presumably that's not an issue right now.

I guess the other option would be to add a parallel `put_json` method that sets the header. I don't feel very strongly about this. :shrug:

LawnGnome (approved):

src/tests/util.rs
Turbo87: yeah, though arguably APIs shouldn't return raw arrays anyway and that extra friction might be good 😄

the only PUT that does not deal with Json currently is the publish endpoint. it felt not worth it to have a dedicated method just for that.

PR #7916, opened 2024-01-11
publish: Introduce dependency limit

Turbo87:
Allowing users to publish crates with an unlimited number of dependencies has the potential to cause various issues. The crate with the most dependencies currently has 256 dependencies, so a limit of 500 dependencies seems reasonable for now. If necessary, we can make this limit configurable with an environment variable in the future, or even per crate with another database column.

see https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/dependencies.20limit


LawnGnome (approved):
:+1:

PR #7921, opened 2024-01-11
typosquat: check for prefixes being manipulated like suffixes

LawnGnome:
In #7571, we added checks for crate names that added or removed suffixes from popular crates. This has turned out to be a useful check! (Spoiler alert for the blog post I'm publishing next week.)

@Turbo87 pointed out that this can also apply to prefixes, especially `cargo-`. This generalises the suffix check to also check prefixes, and adjusts the typomania configuration to add `cargo` to the list of interesting affixes. For now, the same set of affixes will be used for both, but depending on what we see, a future tweak would be to separate the prefix and suffix lists. Let's see how that pans out.

In terms of implementation, I briefly toyed with making this generic over the prefix/suffix combination to remove the copy/paste code, then was reminded by rust-analyzer that `std::str::pattern::Pattern` isn't stable. I'd rather duplicate 20 lines than deal with that, so here we are.

### DIFF
--- a/src/typosquat/cache.rs
+++ b/src/typosquat/cache.rs
@@ -7,7 +7,7 @@ use typomania::{
     Harness,
 };
 
-use super::{checks::Suffixes, config, database::TopCrates};
+use super::{checks::Affixes, config, database::TopCrates};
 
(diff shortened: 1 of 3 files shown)


Turbo87 (approved):
👍

PR #7941, opened 2024-01-15
Improve crates endpoint performance

eth3lbert:
This pr improve the performance by utilizing a count subquery for total retrieval.

P.S. make a separate query to fetch total should also outperforms window functions in this context.

### DIFF
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -5,6 +5,7 @@ use diesel::dsl::*;
 use diesel::sql_types::Array;
 use diesel_full_text_search::*;
 use indexmap::IndexMap;
+use once_cell::sync::OnceCell;
 
 use crate::controllers::cargo_prelude::*;
 use crate::controllers::helpers::Paginate;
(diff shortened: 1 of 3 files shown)


src/controllers/krate/search.rs
```diff
@@ -356,4 +245,174 @@ pub async fn search(app: AppState, req: Parts) -> AppResult<Json<Value>> {
...
+
+    fn authed_user_id(&self, req: &Parts, conn: &mut PgConnection) -> AppResult<&i32> {
+        self._auth_user_id.get_or_try_init(|| {
+            let user_id = AuthCheck::default().check(req, conn)?.user_id();
+            Ok(user_id)
+        })
+    }
+
+    fn supports_seek(&self) -> bool {
+        // Calculating the total number of results with filters is supported but paging is not supported yet.
```
LawnGnome, line 290: I'm correct in thinking this isn't a regression, right, because we don't support this today?

LawnGnome (approved):
I like this! I particularly like the new test assertions.

I experimented with a similar change last year, but trying to fully automatically build the count query instead of having it added separately. Turned out to be really hard (maybe actually impossible, right now?) because of what Diesel does and doesn't expose from its innards. So I like this as a pragmatic middle ground, and I like that the changes to the controller weren't as bad as I thought they'd be last year.

@Turbo87 Let's give this a quick once over on staging before we deploy, but I think this is good.

src/controllers/krate/search.rs
eth3lbert: Yes, and I also like to make a follow up PR which expands the scope of the seek-based pagination for crates endpoint.

eth3lbert: For offset-based pagination, the response will remain unchanged. The only modification pertains to seek-based pagination. Previously, we provided an approximate total count; however, with this PR, the total count will align precisely with the offset-based count.

Turbo87:
> Let's give this a quick once over on staging before we deploy, but I think this is good.

I've just deployed `main` to production, so we can now merge this and test it out on staging.

thanks a lot for the work, @eth3lbert! :)

carols10cents:
Cross-referencing here: we had to revert this PR because it broke search queries with spaces in them. New issue for re-landing: https://github.com/rust-lang/crates.io/issues/8052

PR #7943, opened 2024-01-15
Remove deprecated `/api/v1/versions` endpoints

Turbo87:
These endpoints were deprecated by https://github.com/rust-lang/crates.io/commit/1364c27d2f3f7022a6a3b20fc438f995d3ce36ee in 2017 (6ish years ago).

Over the past months we only had a single bot using this API and they were requesting `/api/v1/versions` without any query parameters and always getting back an empty array... 🤷‍♂️


JohnTitor (approved):
👍, let's say good-bye.

LawnGnome (approved):
giphy

PR #7965, opened 2024-01-18
controllers/krate/owners: Use `Json` extractor to parse request body

Turbo87:
This will allow us to eventually get rid of our custom `BytesRequest` extractor and it also gives users more helpful error messages.


LawnGnome (approved):
Makes sense to me. :+1:

PR #7982, opened 2024-01-22
Fix "Invited by" column

Turbo87:
The "Invited by" column on https://crates.io/me/pending-invites has not shown any content for the past 14 months (since https://github.com/rust-lang/crates.io/pull/5501 was merged and deployed).

When working on the migration back then, I missed that the new API endpoint does not contain the `invited_by_username` field anymore, but our frontend template was still using that field.

This PR removes the `invited_by_username` field from the corresponding Ember Data model and replaces it with the `inviter` and `invitee` relationships, which are then used to display the inviter name on https://crates.io/me/pending-invites.

Best reviewed commit-by-commit.


LawnGnome (approved):
I'm obviously blissfully ignorant of the finer implications of how Ember works internally, but makes intuitive sense to me.

Sorry, I mean, it makes `${sense}_id` to me.

Turbo87:
> I'm obviously blissfully ignorant of the finer implications of how Ember works internally

*Ember Data 😉

LawnGnome:
> *Ember Data 😉

See how blissful I am?

(The funny part is that I actually wrote Ember Data to start with, then Googled it, and convinced myself it was just part of Ember.)

PR #7983, opened 2024-01-22
Expand the scope of seek-based pagination on search

eth3lbert:
This PR add seek-based pagination support for all sorting.

Optimization opportunities remain, but not within this PR:
- Instead of simply limiting the query to `per_page`, we could check for additional records by querying with `per_page + 1`. If the actual count exceeds `per_page`, it indicates there's a next page; otherwise, the next page is None. This eliminates unnecessary requests when no further pages exist.

### DIFF
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -121,19 +121,19 @@ pub async fn search(app: AppState, req: Parts) -> AppResult<Json<Value>> {
         // to ensure predictable pagination behavior.
         if sort == Some("downloads") {
             seek = Some(Seek::Downloads);
-            query = query.order((crates::downloads.desc(), crates::id.asc()))
+            query = query.order((crates::downloads.desc(), crates::id.desc()))
         } else if sort == Some("recent-downloads") {
             seek = Some(Seek::RecentDownloads);
             query = query.order((
                 recent_crate_downloads::downloads.desc().nulls_last(),
-                crates::id.asc(),
+                crates::id.desc(),
             ))
         } else if sort == Some("recent-updates") {
             seek = Some(Seek::RecentUpdates);
-            query = query.order((crates::updated_at.desc(), crates::id.asc()));
+            query = query.order((crates::updated_at.desc(), crates::id.desc()));
         } else if sort == Some("new") {
             seek = Some(Seek::New);
-            query = query.order((crates::created_at.desc(), crates::id.asc()));
+            query = query.order((crates::created_at.desc(), crates::id.desc()));
         } else {
             seek = seek.or(Some(Seek::Name));
             // Since the name is unique value, the inherent ordering becomes naturally unique.
@@ -413,12 +413,12 @@ impl<'a> FilterParams<'a> {
             }
             SeekPayload::New(New(created_at, id)) => {
                 // Equivalent of:
-                // `WHERE (created_at = created_at' AND id > id') OR created_at < created_at'`
+                // `WHERE (created_at = created_at' AND id < id') OR created_at < created_at'`
                 vec![
                     Box::new(
                         crates::created_at
                             .eq(created_at)
-                            .and(crates::id.gt(id))
+                            .and(crates::id.lt(id))
                             .nullable(),
                     ),
                     Box::new(crates::created_at.lt(created_at).nullable()),
@@ -426,12 +426,12 @@ impl<'a> FilterParams<'a> {
             }
             SeekPayload::RecentUpdates(RecentUpdates(updated_at, id)) => {
                 // Equivalent of:
-                // `WHERE (updated_at = updated_at' AND id > id') OR updated_at < updated_at'`
+                // `WHERE (updated_at = updated_at' AND id < id') OR updated_at < updated_at'`
                 vec![
                     Box::new(
                         crates::updated_at
                             .eq(updated_at)
-                            .and(crates::id.gt(id))
+                            .and(crates::id.lt(id))
                             .nullable(),
                     ),
                     Box::new(crates::updated_at.lt(updated_at).nullable()),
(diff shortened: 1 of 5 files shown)


src/tests/routes/crates/list.rs
```diff
@@ -791,12 +791,13 @@ fn seek_based_pagination() {
         if let Some(new_url) = resp.meta.next_page {
             assert_that!(resp.crates, len(eq(1)));
             url = Some(new_url);
+            assert_eq!(resp.meta.total, 3);
         } else {
             assert_that!(resp.crates, empty());
+            assert_eq!(resp.meta.total, 0);
```
Turbo87, line 970: I don't think I understand this assertion change. shouldn't `total` still show `3`?

eth3lbert: Previously, the `total` was retrieved through separated db request. After the refactoring, we use single query to fetch data and `total`, but it now includes empty rows. This caused the `total` value becomes 0 and mirrors the behavior of offset-based pagination.

src/controllers/krate/search.rs
```diff
@@ -102,59 +95,70 @@ pub async fn search(app: AppState, req: Parts) -> AppResult<Json<Value>> {
...
-            query = query.order(crates::updated_at.desc());
+            seek = Some(Seek::RecentUpdates);
+            query = query.order((crates::updated_at.desc(), crates::id.asc()));
         } else if sort == Some("new") {
-            // Custom sorting is not supported yet with seek.
-            supports_seek = false;
-
-            query = query.order(crates::created_at.desc());
+            seek = Some(Seek::New);
+            query = query.order((crates::created_at.desc(), crates::id.asc()));
```
eth3lbert, line 137: If larger `ids` signify newer items with the same `created_at` timestamp, it might be more intuitive to switch the sorting order from ascending `crates::id.asc()` to descending `crates::id.desc()`.

Turbo87: > If larger `ids` signify newer items with the same `created_at` timestamp

yep, that's how it would be. I'm not sure if it's worth breaking the consistency. I haven't checked, but I would be surprised if we actually had crates in the database with the exact same `created_at` timestamp.

src/controllers/krate/search.rs
```diff
@@ -407,6 +397,199 @@ impl<'a> FilterParams<'a> {
...
+            crates_aliased
+                .find(id)
+                .select(crates_aliased.field(crates::name))
+                .single_value()
+        };
+        let conditions: Vec<BoxedCondition<'_>> = match *seek_payload {
+            SeekPayload::Name(Name(id)) => {
+                // Equivalent of:
+                // `WHERE name > name'`
+                vec![Box::new(crates::name.nullable().gt(crate_name_by_id(id)))]
```
Turbo87, line 415: why the `nullable()` here?

Turbo87 (approved):
looks good to me on first/second glance :)

@rust-lang/crates-io a second pair of eyes won't hurt though

src/controllers/krate/search.rs
eth3lbert: Without `nullable()`, compiler complains about a type mismatched.
```
error[E0271]: type mismatch resolving `<Grouped<Subselect<SelectStatement<FromClause<Alias<crates_aliased>>, SelectClause<AliasedField<crates_aliased, name>>, NoDistinctClause, ..., ..., ...>, ...>> as Expression>::SqlType == Text`
   --> src/controllers/krate/search.rs:415:44
    |
415 |                 vec![Box::new(crates::name.gt(crate_name_by_id(id)))]
    |                                            ^^ expected `Nullable<Text>`, found `Text`
    |
    = note: expected struct `diesel::sql_types::Nullable<diesel::sql_types::Text>`
               found struct `diesel::sql_types::Text`
    = note: required for `Grouped<Subselect<SelectStatement<FromClause<Alias<crates_aliased>>, SelectClause<AliasedField<crates_aliased, name>>, NoDistinctClause, ..., ..., ...>, ...>>` to implement `AsExpression<diesel::sql_types::Text>`
```

src/controllers/krate/search.rs
eth3lbert: If we can guarantee that timestamps will never be identical, then the auxiliary ordering column becomes unnecessary and can be removed. However, if the possibility of duplicate timestamps exists, then it's better to keep it.

---

Since we all use descending order for sorting, I think it would be more intuitive to change all the auxiliary column `id.asc()` to `id.desc()`. What do you think?

Turbo87: > I think it would be more intuitive to change all the auxiliary column `id.asc()` to `id.desc()`

yeah, that should also work :)

src/tests/routes/crates/list.rs
```diff
@@ -262,60 +306,172 @@ fn index_sorting() {
...
+        assert_eq!(json.crates[2].name, "bar_sort");
+        assert_eq!(json.crates[3].name, "other_sort");
+    }
+    let (resp, calls) = page_with_seek(&anon, "sort=recent-downloads");
+    assert_eq!(resp[0].crates[0].name, "foo_sort");
+    assert_eq!(resp[1].crates[0].name, "baz_sort");
+    assert_eq!(resp[2].crates[0].name, "bar_sort");
+    assert_eq!(resp[3].crates[0].name, "other_sort");
+    assert_eq!(resp[3].meta.total, 4);
+    assert_eq!(calls, 5);
```
eth3lbert, line 474: This code snippet can be safely removed as it's already covered by above test case.

src/controllers/krate/search.rs
eth3lbert: It is implemented in the new commit.

eth3lbert:
Rebased and resolved all merge conflicts.

Turbo87:
Rebased once more. AFAICT all of the existing functionality should still work the same as before, so I think we can try this out as an experimental feature. If it works, great, if it doesn't, we can always revert :)

PR #8002, opened 2024-01-25
Make frontend build scripts work on Windows

### LINKED ISSUE #7360: Multiple frontend build scripts don't work on Windows
### Current Behavior

Multiple frontend build scripts defined in `package.json` don't work on Windows because they use a syntax for setting an environment variable which does not work on Windows. For example `pnpm start:live` fails with:
```
'PROXY_BACKEND' is not recognized as an internal or external command,
operable program or batch file.
```

See also [related Stack Overflow question](https://stackoverflow.com/q/40806796).

### Expected Behavior

Either:
- The build scripts also work on Windows
- Or, there are additional variants which work on Windows
- Or, https://github.com/rust-lang/crates.io/blob/main/docs/CONTRIBUTING.md#setting-up-a-development-environment explicitly mentions that building on Windows is not fully supported.
Currently the "Frontend requirements" in CONTRIBUTING.md suggest that Windows is (fully?) supported.

### Steps To Reproduce

_No response_

### Environment

- OS: Windows 10
- Windows Terminal as default terminal application

### Anything else?

_No response_

cijiugechu:
This PR uses pnpm's [shell-emulator](https://pnpm.io/cli/run#shell-emulator) to execute scripts, so that syntax like `PROXY_BACKEND=https://crates.io` can run on Windows.

Resolves #7360

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+shell-emulator=true


Turbo87 (approved):
I can't test this as I'm not on Windows, but the documentation makes it sounds like this is indeed exactly what we need. Thank you! :)

Marcono1234:
Thanks! At least for me that seems to have solved the issue; I can now run `pnpm start:live` on Windows.

PR #8010, opened 2024-01-26
Implement CDN log file parsers

Turbo87:
This PR adds a new `crates_io_cdn_logs` package in the repository, with all the code necessary to parse and process log files from our two CDN systems: CloudFront and Fastly. Example log files are included in this PR.

When we receive these log files they are compressed using gzip and zstd respectively, thus this PR also contains code to decompress the content on-the-fly.

The code is written in an async/streamy kind of way to avoid buffering the whole log file in memory while processing it. On my local machine a 80 MB log file with 120k records is processed in roughly 1.6 sec.

Note that the initial design was using the `regex` and `url` crates in a couple of places, but, as the code comments hint at, they have proven significantly slower than doing the parsing by hand. Somewhat related: this PR also contains benchmark code for both file formats 😉 

Feel free to DM me if you need real-life log files to play around with.

Finally: best reviewed commit-by-commit :)


paolobarbolini:
If you ever need to use the logs for counting downloads in the database let me know. PostgreSQL is very fast and efficient at ingesting CSVs via the `COPY` command.

Turbo87:
@paolobarbolini the problem with ingesting the download counts into the database is that we don't have the `version_id` yet.

I'm currently playing with an approach like this:

Bildschirmfoto 2024-01-26 um 15 53 33

which appears to work decently well for a typical real-life log file:

```
Number of crates: 5702
Number of needed inserts: 15654
Total number of downloads: 116192
Time to parse: 1.48536s

Inserting into database…
Time to insert: 2.165449s
```

paolobarbolini:
> @paolobarbolini the problem with ingesting the download counts into the database is that we don't have the `version_id` yet.
> ```
> Number of crates: 5702
> Number of needed inserts: 15654
> Total number of downloads: 116192
> Time to parse: 1.48536s
> 
> Inserting into database…
> Time to insert: 2.165449s
> ```

That looks like a good approach. You should make sure `temp_buffers` is big enough to fix the entire temporary table, if it's reasonable to do so, in order to avoid disk writes.

Turbo87:
@paolobarbolini thanks, that's a good hint! remind me when I open the follow-up PR that implements this 😉

paolobarbolini:
I'm wandering whether async makes sense here, since I'd expect it to mostly be CPU-bound work

Turbo87:
> I'm wandering whether async makes sense here, since I'd expect it to mostly be CPU-bound work

I didn't feel comfortable loading the full 80MB log file into memory. Knowing our traffic graphs the size of these file will grow quite a bit in the future 😅

paolobarbolini:
> > I'm wandering whether async makes sense here, since I'd expect it to mostly be CPU-bound work
> 
> I didn't feel comfortable loading the full 80MB log file into memory. Knowing our traffic graphs, the size of these file will grow quite a bit in the future 😅

If the problem is with streaming the file you can get a pre-signed download URL from the AWS SDK and then use whatever HTTP client to stream the file download.

eth3lbert:
I totally agree with the two points @paolobarbolini  mentioned:
- Preprocessing the data into CSV format and using the `COPY` statement to insert it into a temporary table can significantly improve the insert speed when the data volume is large.

- The `temp_buffers` setting is also important. You can either declare a large enough size for it initially or choose a suitable batch size.

crates_io_cdn_logs/src/paths.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    let version = filename.strip_prefix(folder)?;
+    let version = version.strip_prefix('-')?;
+    let version = Version::parse(version).ok()?;
+
+    Some((folder.to_owned(), version))
+}
+
+#[cfg(test)]
+mod tests {
```
LawnGnome, line 31: I wouldn't mind seeing a couple more tests for these specific cases:

1. A crate name with one or more dashes, eg `/crates/foo-bar/foo-bar-1.0.0.crate`.
2. A version number with one or more dashes, eg `/crates/foo/foo-1.0.0-rc.1.crate`.

(I know these are handled correctly now, but just in case someone decides to "improve" this later.)

crates_io_cdn_logs/src/cloudfront.rs
```diff
@@ -0,0 +1,217 @@
...
+        if let Some(version) = line.strip_prefix(HEADER_VERSION) {
+            let version = version.trim();
+            if version != "1.0" {
+                anyhow::bail!("Unsupported version: {}", version);
+            }
+            continue;
+        }
+
+        if let Some(fields_str) = line.strip_prefix(HEADER_FIELDS) {
+            let fields = fields_str.trim().split(' ').collect::<Vec<_>>();
```
LawnGnome, line 38: It scares me a tiny bit that neither the Amazon documentation nor the (27 year old draft!) RFC it refers actually specifies the separator for field names, but obviously we know what it seems to be in practice. :grimacing:

crates_io_cdn_logs/src/fastly/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    Ok(downloads)
+}
+
+#[instrument(level = "debug", skip(line))]
+fn parse_line(line: &str) -> Option<&str> {
+    // A regex could also be used here, but the `find()` call appears to
+    // be roughly 10x faster.
+    line.find(r#"]: {"#).map(|pos| &line[pos + 3..])
```
LawnGnome, line 63: I had a little bit of trouble finding the log line format we use, which is https://docs.fastly.com/en/guides/changing-log-line-formats#classic-format — could we include that in a comment wherever you feel is appropriate?

crates_io_cdn_logs/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+pub use crate::download_map::DownloadsMap;
+use std::io::Cursor;
+use tokio::io::{AsyncBufRead, AsyncReadExt};
+
+pub async fn count_downloads<R>(mut reader: R) -> anyhow::Result<DownloadsMap>
+where
+    R: AsyncBufRead + Unpin,
+{
+    // Read the first byte to determine the file format.
+    match reader.read_u8().await? {
```
LawnGnome, line 19: I'm handling my notification backlog in chronological order, so there's probably context in a later PR or issue that explains this, but is sniffing something we actually need here? Is there a scenario where we won't know the source of a log file?

(Personal bias: I'm automatically mistrustful of anything that looks like data sniffing.)

LawnGnome (approved):
Apart from a question about the public API of this crate, no concerns. Looks good!

crates_io_cdn_logs/src/cloudfront.rs
Turbo87: "whitespace" and yet it uses spaces in the header and tabs in the payload... 🙄

crates_io_cdn_logs/src/fastly/mod.rs
Turbo87: > I had a little bit of trouble finding the log line format we use

you and me both, which is why I didn't include any links 😂

crates_io_cdn_logs/src/lib.rs
Turbo87: the alternative is looking at the folder structure on the S3 bucket and choosing the file type based on that. or the other alternative: fastly logs are using zstd and cloudfront is currently using gzip.

both of those approaches feel a bit brittle though, which is why I implemented the auto-detection.

Turbo87:
rebased and feedback addressed. thanks for the review @LawnGnome! :)

PR #8036, opened 2024-01-31
Implement `ProcessCdnLog` and `ProcessCdnLogQueue` background jobs

Turbo87:
This PR builds on top of https://github.com/rust-lang/crates.io/pull/8010, and implements two new background worker jobs.

The `ProcessCdnLog` job takes an AWS region, S3 bucket name and object path inside the bucket, connects to the bucket via the `object_store` crate, and then stream-downloads the file while running `count_downloads()` on it. If download counting is successful it prints a summary of the result to the logs.

The `ProcessCdnLogQueue` job connects to an AWS SQS queue, receives messages from it, parses them and enqueues `ProcessCdnLog` jobs based on the content of these messages.

You may be asking: "Where are the counted downloads saved to the database?!" and the answer is: nowhere. The plan is to run this read-only setup in production for a couple of days to ensure that the download counting works correctly with production data. Once we have verified that everything works as expected, we will adjust the `ProcessCdnLog` job to save the result to the database instead of or in addition to reporting it to the logs.

Best reviewed commit by commit 😉


Turbo87:
Since CodeCov was unhappy I've now doubled the size of the diff and added a couple of tests... 😅

src/worker/jobs/downloads.rs
```diff
@@ -0,0 +1,521 @@
...
+            .with_context(|| format!("Failed to parse path: {:?}", self.path))?;
+
+        let meta = store.head(&path).await;
+        let meta = meta.with_context(|| format!("Failed to request metadata for {path:?}"))?;
+
+        let reader = object_store::buffered::BufReader::new(Arc::new(store), &meta);
+        let decompressor = Decompressor::from_extension(reader, path.extension())?;
+        let reader = BufReader::new(decompressor);
+
+        let parse_start = SystemTime::now();
```
LawnGnome, line 90: This should use `Instant`, since `SystemTime` isn't monotonic.

src/worker/jobs/downloads.rs
```diff
@@ -0,0 +1,521 @@
...
+                debug!("Processing message: {message_id}");
+
+                let Some(receipt_handle) = message.receipt_handle() else {
+                    warn!("Message {message_id} has no receipt handle; skipping");
+                    continue;
+                };
+
+                debug!("Deleting message {message_id} from the CDN log queue…");
+                queue
+                    .delete_message(receipt_handle)
```
LawnGnome, line 217: Maybe a question for infra, but this queue is configured to only deliver messages exactly once, right?

(What I'm concerned about here is double counting if this job runs twice, and the second `receive_messages()` call occurs before messages being handled by the first job are deleted.)

src/config/cdn_log_queue.rs
```diff
@@ -0,0 +1,33 @@
...
+    },
+    Mock,
+}
+
+impl CdnLogQueueConfig {
+    pub fn from_env() -> anyhow::Result<Self> {
+        if let Some(queue_url) = var("CDN_LOG_QUEUE_URL")? {
+            let access_key = required_var("CDN_LOG_QUEUE_ACCESS_KEY")?;
+            let secret_key = required_var("CDN_LOG_QUEUE_SECRET_KEY")?.into();
+            let region = required_var("CDN_LOG_QUEUE_REGION")?;
```
LawnGnome, line 20: These should probably be documented in `.env.sample`.

LawnGnome (approved):
Looking forward to having this "in production" so we can check it against the existing logic!

src/worker/jobs/downloads.rs
Turbo87: > this queue is configured to only deliver messages exactly once, right?

from what I've seen so far that does not appear to be the case. I assume if it was the case then the delivered messages would be auto-deleted upon reception, right?

Turbo87: I will treat this as a non-blocking concern since the code here is experimental and read-only for now anyway. If we can change the queue to use once-only delivery then we can just delete the `delete_message()` calls here.

jdno: > Maybe a question for infra, but this queue is configured to only deliver messages exactly once, right?

This is a [standard queue](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/standard-queues.html) that only guarantees [`at least once`](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/standard-queues-at-least-once-delivery.html) delivery. If we want [`exactly once`](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues-exactly-once-processing.html), we'd need to recreate the queue as a [FIFO queue](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-fifo-queues.html).

PR #8037, opened 2024-01-31
Add basic support for seek-based pagination to versions endpoint

### LINKED ISSUE #5292: Add paginated versions API

eth3lbert:
This PR is an improved implementation of #5302.
Add seek-based pagination support to versions endpoint with following:
- move versions to its own file
- To keep backward compatibility, paginate versions only if `per_page` is provided.
- Support `sort` query parameter to choose either sorting by `date` or `semver` .

For date sorting, we can perform sorting and pagination directly in the database. However, Heroku Postgres doesn't support the semver PG extension, so sorting and pagination for semver needs to be done on the app server.

Resolves #5292

### DIFF
--- a/src/controllers/krate/versions.rs
+++ b/src/controllers/krate/versions.rs
@@ -2,7 +2,11 @@
 
 use std::cmp::Reverse;
 
+use diesel::connection::DefaultLoadingMode;
+use indexmap::IndexMap;
+
 use crate::controllers::frontend_prelude::*;
+use crate::controllers::helpers::pagination::{encode_seek, Page, PaginationOptions};
 
 use crate::models::{Crate, CrateVersions, User, Version, VersionOwnerAction};
 use crate::schema::{users, versions};
@@ -10,9 +14,11 @@ use crate::util::errors::crate_not_found;
 use crate::views::EncodableVersion;
 
 /// Handles the `GET /crates/:crate_id/versions` route.
-// FIXME: Not sure why this is necessary since /crates/:crate_id returns
-// this information already, but ember is definitely requesting it
-pub async fn versions(state: AppState, Path(crate_name): Path<String>) -> AppResult<Json<Value>> {
+pub async fn versions(
+    state: AppState,
+    Path(crate_name): Path<String>,
+    req: Parts,
+) -> AppResult<Json<Value>> {
     spawn_blocking(move || {
         let conn = &mut *state.db_read()?;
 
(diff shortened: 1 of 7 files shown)


src/controllers/krate/versions.rs
```diff
@@ -21,27 +27,217 @@ pub async fn versions(state: AppState, Path(crate_name): Path<String>) -> AppRes
...
+    Ok(PaginatedVersionsAndPublishers {
+        data,
+        meta: ResponseMeta { total, next_page },
+    })
+}
+
+// Unfortunately, Heroku Postgres has no support for the semver PG extension.
+// Therefore, we need to perform both sorting and pagination manually on the server.
+fn list_by_semver(
+    krate: &Crate,
```
Turbo87, line 122: I'm aware that this was already here before, but in a follow-up PR we could probably replace `krate: &Crate` with `crate_id: i32` and no longer load the full `Crate` in the `versions()` fn

src/controllers/krate/versions.rs
```diff
@@ -21,27 +27,217 @@ pub async fn versions(state: AppState, Path(crate_name): Path<String>) -> AppRes
...
+        sorted_versions.sort_by_cached_key(|_, (num, _)| Reverse(semver::Version::parse(num).ok()));
+
+        let mut idx = Some(0);
+        if let Page::Seek(ref seek) = options.page {
+            idx = seek
+                .decode::<i32>()
+                .map(|id| sorted_versions.get_index_of(&id))?
+                .filter(|i| i + 1 < sorted_versions.len())
+                .map(|i| i + 1);
+        }
```
Turbo87, line 148: do we need an `else` branch here? I assume for `Page::Unspecified` doing nothing here is correct, but should we handle the `Page::Numeric` case in some way? the surrounding code should make this impossible, but technically `list_by_semver()` could be called with `Page::Numeric`.

src/controllers/krate/versions.rs
```diff
@@ -21,27 +27,217 @@ pub async fn versions(state: AppState, Path(crate_name): Path<String>) -> AppRes
...
+            let ids = sorted_versions[start..end].keys().collect::<Vec<_>>();
+            for result in krate
+                .all_versions()
+                .left_outer_join(users::table)
+                .select((versions::all_columns, users::all_columns.nullable()))
+                .filter(versions::id.eq_any(ids))
+                .load_iter::<(Version, Option<User>), DefaultLoadingMode>(conn)?
+            {
+                let row = result?;
+                sorted_versions.insert(row.0.id, (row.0.num.to_owned(), Some(row)));
```
Turbo87, line 160: it might be worth adding a comment somewhere in this function (maybe on the `let sorted_versions`) explaining the purpose of that map and why we're reusing the `sorted_versions` map here. the naive implementation would collect the results in a `Vec` and sort that again, but I assume the reason for reusing `sorted_versions` is to not run `Version::parse()` again?

src/controllers/krate/versions.rs
```diff
@@ -21,27 +27,217 @@ pub async fn versions(state: AppState, Path(crate_name): Path<String>) -> AppRes
...
+            }
+            (
+                sorted_versions
+                    .values()
+                    .flat_map(|(_, v)| v)
```
Turbo87, line 165: why `flat_map()` instead of `filter_map()`? am I missing something?

src/controllers/krate/versions.rs
```diff
@@ -21,27 +27,217 @@ pub async fn versions(state: AppState, Path(crate_name): Path<String>) -> AppRes
...
+
+mod seek {
+    use chrono::naive::serde::ts_microseconds;
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Deserialize, Serialize)]
+    pub(super) struct Date(
+        #[serde(with = "ts_microseconds")] pub(super) chrono::NaiveDateTime,
+        pub(super) i32,
+    );
```
Turbo87, line 207: I think I'd feel more comfortable with named fields for this, unless there is a strong reason why we can't use that.

src/controllers/krate/versions.rs
```diff
@@ -21,27 +27,217 @@ pub async fn versions(state: AppState, Path(crate_name): Path<String>) -> AppRes
...
+        next_page = next_seek_params(&data, options, |last| {
+            seek::Date(last.0.created_at, last.0.id)
+        })?
+        .map(|p| req.query_with_params(p));
+    };
+
+    // Since the total count is retrieved through an additional query, to maintain consistency
+    // with other pagination methods, we only make a count query while data is not empty.
+    let total = if !data.is_empty() {
+        versions::table.count().get_result(conn)?
```
Turbo87, line 108: I think there is a bug here, since this will return the number of all versions in the database, not just the ones belonging to this crate. I assume this would have to be something like `krate.all_versions().count()`.

Turbo87 (commented):
one small bug, and a couple of comments, but overall this looks good to me. nice work! 👍

Turbo87:
and in case you're wondering, I've rebased the branch on the current `main` :)

src/controllers/krate/versions.rs
eth3lbert: Great catch! You're right. Although we've disabled the option in the caller, the function itself doesn't technically prevent calling with the page-enabled option. However, we currently don't offer offset-base pagination capability and don't plan to in the future due to performance concerns. Perhaps we could add a NOTE comment clarifying this and simply return an not supported error. What do you think?

src/controllers/krate/versions.rs
eth3lbert: Yes, kind of. We can either fetch all joined data at once and perform sorting and pagination on it, or fetch only the data needed for sorting and pagination, then make another query for the sorted and paginated response.

Due to the fact that the versions would only increase in the future and sorting and pagination need to happen on the app server, I think the second approach, which minimizes payload and memory usage, would be better. In this case, we can utilize the sorted map and enrich it later without sorting twice.

src/controllers/krate/versions.rs
eth3lbert: Whoops, I just misused it for no reason. Thanks for catching that.

src/controllers/krate/versions.rs
```diff
@@ -21,27 +27,217 @@ pub async fn versions(state: AppState, Path(crate_name): Path<String>) -> AppRes
...
+                .select((versions::all_columns, users::all_columns.nullable()))
+                .filter(versions::id.eq_any(ids))
+                .load_iter::<(Version, Option<User>), DefaultLoadingMode>(conn)?
+            {
+                let row = result?;
+                sorted_versions.insert(row.0.id, (row.0.num.to_owned(), Some(row)));
+            }
+            (
+                sorted_versions
+                    .values()
```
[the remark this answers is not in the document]
eth3lbert: Thanks, the suggested changes for this and the previous one are indeed superior to mine.

src/controllers/krate/versions.rs
eth3lbert: The reason for this is to align the implementation of seek pagination with other routes. Currently, we encode seek as a `Vec`, which doesn't include field names. While we could still customize the `Serializer/Deserializer` for it, I think it would be better to implement this alignment in a separate PR.

Also, #7983 has already been merged, so I would add another fixup commit to use the `seek!` macro to further align the implementation.

src/controllers/krate/versions.rs
eth3lbert: Good catch! I should also make it fail in the test suite.

src/controllers/krate/versions.rs
Turbo87: I'd say `assert!()` + doc comment clarifying that the fn will panic in this case should probably be sufficient

src/controllers/krate/versions.rs
Turbo87: hmm okay, sounds like there are valid reasons. might be good to add write them down in a code comment though :)

eth3lbert:
feedback addressed. thanks for the review :)

diff [3f1a04e90057ae7f3714bcf3e6bd165223f00710..93e73621ccaac6b05afecc8d7488858aa84ae0b9](https://github.com/rust-lang/crates.io/compare/3f1a04e90057ae7f3714bcf3e6bd165223f00710..93e73621ccaac6b05afecc8d7488858aa84ae0b9)

Turbo87 (approved):
alright, let's give this a try! the test changes are all additive so I'm assuming the existing functionality won't break :)

PR #8053, opened 2024-02-02
Re-introducing improve crates endpoint performance (#7941)

### LINKED ISSUE #8052: #7941 broke searches with spaces in them
See also: https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/can't.20search.20with.20a.20space

@eth3lbert, could you have a look, please?

eth3lbert:
This PR re-introduce #7941 and resolves #8052 by reverting the use of the `to_tsquery_with_search_config` for `TsQuery` part back to the original code.

### DIFF
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -350,6 +350,8 @@ where
         out.push_sql(") t LIMIT ");
         out.push_bind_param::<BigInt, _>(&self.options.per_page)?;
         if let Some(offset) = self.options.offset() {
+            // Injection safety: `offset()` returns `Option<i64>`, so this interpolation is constrained to known
+            // valid values and this is not vulnerable to user injection attacks.
             out.push_sql(format!(" OFFSET {offset}").as_str());
         }
         Ok(())
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -146,6 +146,18 @@ pub(crate) trait Paginate: Sized {
             options,
         }
     }
+
+    fn pages_pagination_with_count_query<C>(
+        self,
+        options: PaginationOptions,
+        count_query: C,
+    ) -> PaginatedQueryWithCountSubq<Self, C> {
+        PaginatedQueryWithCountSubq {
+            query: self,
+            count_query,
+            options,
+        }
+    }
 }
 
 impl<T> Paginate for T {}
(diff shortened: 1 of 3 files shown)


src/controllers/helpers/pagination.rs
```diff
@@ -303,6 +315,61 @@ pub(crate) fn decode_seek<D: for<'a> Deserialize<'a>>(seek: &str) -> anyhow::Res
...
+{
+    fn walk_ast<'b>(&'b self, mut out: AstPass<'_, 'b, Pg>) -> QueryResult<()> {
+        out.push_sql("SELECT *, (");
+        self.count_query.walk_ast(out.reborrow())?;
+        out.push_sql(") FROM (");
+        self.query.walk_ast(out.reborrow())?;
+        out.push_sql(") t LIMIT ");
+        out.push_bind_param::<BigInt, _>(&self.options.per_page)?;
+        if let Some(offset) = self.options.offset() {
+            out.push_sql(format!(" OFFSET {offset}").as_str());
```
carols10cents, line 353: Thank you for addressing my concern in the issue about injection-- this line in particular, I had to look at surrounding code to convince myself that this wasn't a possible user injection location because `offset()` returns `Option<i64>` (as opposed to a user-provided string without any processing on our end, which could be Quite Bad Indeed).

Could you please add a comment to this line, much like the sort of comment that's good practice to leave when you write `unsafe` Rust code, that documents why this is safe? Something like:

```suggestion
            // Injection safety: `offset()` returns `Option<i64>`, so this interpolation is constrained to known
            // valid values and this is not vulnerable to user injection attacks.
            out.push_sql(format!(" OFFSET {offset}").as_str());
```

just to encode this assumption a bit more for the future? It would help me sleep better at night 😁 Thank you!

eth3lbert: I appreciate your suggestion, and you're absolutely right. Even I had to review the `push_sql` part multiple times to avoid unintended consequences.

We could even consider adding an `offset` field and using `push_bind_value` instead, which might reduce potential risks.

Turbo87:
```suggestion
            out.push_sql(" OFFSET ");
            out.push_bind_param::<BigInt, _>(&offset)?;
```

is there a reason why we're not using something like this?

looks like we're doing the same for the `per_page` value above, so I'm wondering if there is a reason for not doing the same here too :)

eth3lbert: My current implementation follows the existing pattern for consistency. However, if we want to apply the same approach to `offset` as with `per_page`, storing its value would be necessary to avoid potential lifetime issues.

Turbo87: > storing its value would be necessary to avoid potential lifetime issues

I see. I guess the `format!()` resolves that by serializing it to a `String`. nevermind then :)

PR #8056, opened 2024-02-02
Reduce versions page load time

eth3lbert:
This PR significantly improve the versions page's loading time by:
- Instead of calculating `isFirst` and `isHighestOfReleaseTrack` on the fly for every version, pre-compute `firstVersionId` and `releaseTrackSet` and just do comparisons during rendering for a speed boost.
- Wrap `releaseTrackTitle` inside `EmberTooltip` component rather than pass as `@text` attribute.

I'm not an Ember.js expert, so there might be alternative ways to structure this code.
Please feel free to review and provide feedback.


app/components/version-list/row.hbs
```diff
@@ -20,7 +20,9 @@
         {{@version.releaseTrack}}
       {{/if}}
 
-      <EmberTooltip @text={{this.releaseTrackTitle}} @side="right" data-test-release-track-title/>
+      <EmberTooltip @side="right" data-test-release-track-title>
+        {{this.releaseTrackTitle}}
+      </EmberTooltip>
```
Turbo87, line 25: I assume this is faster because `this.releaseTrackTitle` is now only evaluated lazily when the tooltip is rendered, instead of being evaluated eagerly because it was passed as an argument to the component?

app/models/crate.js
```diff
@@ -41,6 +42,41 @@ export default class Crate extends Model {
...
+  }
+
+  @cached get firstVersionId() {
+    return this.versionIdsByDate.at(-1);
+  }
+
+  @cached get versionsObj() {
+    let versions = this.versions.toArray() ?? [];
+    return Object.fromEntries(versions.map(v => [v.id, v]));
+  }
```
Turbo87, line 62: are these recomputing correctly when the content of `this.versions` changes?

app/models/version.js
```diff
@@ -49,7 +49,10 @@ export default class Version extends Model {
   }
 
   @cached get isFirst() {
-    let { versions } = this.crate;
+    let { versions, firstVersionId } = this.crate;
+    if (firstVersionId !== undefined) {
+      return this.id === firstVersionId;
+    }
```
Turbo87, line 55: is the `else` branch possible? as in, is it possible for `firstVersionId` to be `null`? I assume it's possible if `this.versions` in the crate is empty, but I guess in that was we wouldn't be looking at a `Version` model for the crate?

app/models/version.js
```diff
@@ -85,7 +88,11 @@ export default class Version extends Model {
     }
 
     let { crate, semver, releaseTrack } = this;
-    let { versions } = crate;
+    let { versions, releaseTrackSet } = crate;
+    if (releaseTrackSet !== undefined) {
```
Turbo87, line 92: unless I'm misreading the above code, `releaseTrackSet` should never be `undefined` and will always at least be an empty `Set`, right?

Turbo87 (commented):
nice work! 👍 

for future PRs: my rule of thumb is, if there is a list of unrelated changes in the PR description it's probably best to make it multiple PRs.

e.g. I would merge the tooltip change already, but since it's in the same commit and PR, it's blocked until the other comments are resolved :D

app/components/version-list/row.hbs
eth3lbert: You're absolutely right, that's exactly what this code snippet does. No worries, I usually try to follow that approach as well. In this case, I could split it into two commits for better clarity.

app/models/version.js
eth3lbert: To be honest, my lack of familiarity with Ember.js means this approach might be more prone to falling back to the `else` branch if something unexpected happens. Ideally, with everything working as intended, we wouldn't need to use the `else` branch at all.

app/models/version.js
eth3lbert: Similar to the previous reason, this additional check can be seen as a defensive programming measure. If all goes well, it will indeed behave as you described.

app/models/crate.js
eth3lbert: Yes, I assumed so. Would you like me to add a test for this functionality?

Turbo87: depends on how much effort that requires. maybe try to write a test that visits the version page, goes to another page, pushes another version into the Ember Data store, visits the version page again and asserts that the new version is now also correctly displayed?

app/models/version.js
Turbo87: I think I'd just make it `return this.id === this.crate?.firstVersionId`. Worst thing that could happen then is that `this.crate` or `this.crate.firstVersionId` are indeed undefined for some weird reason and in that case `isFirst` would be `false` 🤷‍♂️

app/models/version.js
Turbo87: same here, I think. `return this.crate.releaseTrackSet.has(this.id)` should probably be sufficient. worst case, it will return `false` and not display the right styles, but it won't break the app.

app/models/crate.js
eth3lbert: Instead of revisiting the page, I used `findRecord` with `reload` param to ensure fresh data. I'm not sure if this is sufficient, though.

Turbo87: yeah, that should also be sufficient 👍

eth3lbert:
feedback addressed. thanks for the review :)

tests/models/version-test.js
```diff
@@ -193,6 +193,38 @@ module('Model | Version', function (hooks) {
...
+      crateRecord = await this.store.findRecord('crate', crate.name, { reload: true });
+      versions = (await crateRecord.versions).slice();
+
+      assert.deepEqual(
+        versions.map(it => ({ num: it.num, isHighestOfReleaseTrack: it.isHighestOfReleaseTrack })),
+        [
+          { num: '0.4.0', isHighestOfReleaseTrack: false },
+          { num: '0.4.1', isHighestOfReleaseTrack: false },
+          { num: '0.4.2', isHighestOfReleaseTrack: true },
+          { num: '0.4.3', isHighestOfReleaseTrack: true },
```
Turbo87, line 224:
```suggestion
          { num: '0.4.2', isHighestOfReleaseTrack: true },
          { num: '0.4.3', isHighestOfReleaseTrack: true },
```

I assume 0.4.2 is highest because 0.4.3 is yanked, but why does 0.4.3 also have `isHighestOfReleaseTrack: true`?

Turbo87: did the previous implementation work the same way? I have a feeling that this was unintentional previously... 😂

eth3lbert: Ha! I implemented it this way solely to pass the previous tests, although I admit feeling a bit strange with it in the past.

If we would also like to ignore yanked crates and only has one `highestOfReleaseTrack` for each `releaseTrack`, we could simplify the implementation to
```js
  @cached get releaseTrackSet() {
    let map = new Map();
    let { versionsObj: versions, versionIdsBySemver } = this;
    for (let id of versionIdsBySemver) {
      let { releaseTrack, isPrerelease, yanked } = versions[id];
      if (releaseTrack && !isPrerelease && !yanked && !map.has(releaseTrack)) {
        map.set(releaseTrack, id);
      }
    }
    return new Set(map.values());
  }
```

Turbo87: yeah, I'd say let's go for the simpler implementation 👍

Turbo87 (approved):
very much appreciated. thanks a lot for these improvements! :)

PR #8081, opened 2024-02-07
Use `processed_log_files` table to deduplicate log file processing

Turbo87:
As [discussed on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/download.20counting.20via.20CDN.20logs/near/419800774), we unfortunately can't use an SQS queue with exactly-once delivery behavior. This means we have to ensure ourselves that we don't process log files more than once.

This PR adds a new `processed_log_files` database table and adjusts the `ProcessCdnLog` background worker job in two ways:

1. on job startup it checks whether the file was already processed and skips the rest of the job if that is the case.

2. after processing the log file it inserts a row into the new table to record that the file was processed. this is happening in the same transaction as the rest of the database writes to ensure that the download counting write fails if the `processed_log_files` insertion fails.


LawnGnome (approved):
The only concern I had was the transaction thing you'd already pointed out. Otherwise LGTM.

Turbo87:
@LawnGnome thanks for the review

there are a couple of changes I want to make once I rebase this on top of #8096:

- the table should get a `time` column so that we can delete entries that are e.g. older than 2 weeks.
- I'm thinking about dropping the `region` and `bucket` columns. They will likely be the same for all rows anyway, and if they aren't then it seems unlikely for the path to be the same. To take this further, if we replicate the CDN logs to a second bucket for whatever reason and that is piped into the queue we might get events for both files, with the same path, but on different buckets/regions. In that case it would actually make sense to only count one of the files.

Turbo87:
rebased and adjusted according to my comments in https://github.com/rust-lang/crates.io/pull/8081#issuecomment-1935545826

LawnGnome (approved):

PR #8087, opened 2024-02-08
worker/jobs/downloads/queue: Improve `message_id` logging

Turbo87:
Instead of passing the `message_id` reference everywhere we can refactor the code a bit to add the message ID to a `tracing` span, which automatically adds it to all log messages within the scope of the span.

### DIFF
--- a/src/worker/jobs/downloads/queue/job.rs
+++ b/src/worker/jobs/downloads/queue/job.rs
@@ -97,31 +97,31 @@ async fn run(
 }
 
 /// Processes a single message from the CDN log queue.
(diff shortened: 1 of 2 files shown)


src/worker/jobs/downloads/queue/job.rs
```diff
@@ -89,55 +89,61 @@ async fn run(
...
-                })?;
+            process_message(message, queue, connection_pool).await?;
         }
     }
 
     Ok(())
 }
 
 /// Processes a single message from the CDN log queue.
+#[instrument(skip_all, fields(cdn_log_queue.message.id = %message.message_id().unwrap_or("<unknown>")))]
```
Turbo87, line 100: this is the relevant line 😉

PR #8096, opened 2024-02-08
Implement writing CDN log counted downloads to the database

Turbo87:
This PR resolves the `TODO` comment in the `ProcessCdnLog` background job by implementing the database writing part of the job.

The writing behavior is enabled by a new `CDN_LOG_COUNTING_ENABLED` feature flag. If this flag is set, regular download counting on the download endpoint is disabled and the writing behavior of background job is enabled.

Writing the downloads to the database is a three step process. First, a new temporary database table will be created. Then we insert the content of the `DownloadMap` into this temporary table. Finally, we use the temporary table to insert data into the `version_downloads` table. This multi-step process is needed because we only have the name and version of the downloaded crate, but not the corresponding `version_id` and looking them up one-by-one would be quite slow. The third step uses a couple of joins to figure out the right `version_id` from the crate name and version columns from that temporary table instead.

Note the this does not implement the deduplication of log files yet. #8081 has been put into draft mode for now, since its implementation will be significantly easier to test once the database writing in this PR is implemented.

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -36,6 +36,13 @@ pub struct Server {
     pub max_blocking_threads: Option<usize>,
     pub db: DatabasePools,
     pub storage: StorageConfig,
+    /// If `true`, disables download counting on the download API endpoint, and
+    /// enables writing download counts from the CDN logs to the database.
+    ///
+    /// If `false`, downloads will be counted on the download API endpoint and
+    /// the download counts from the CDN logs will only be reported in the
+    /// logs.
+    pub cdn_log_counting_enabled: bool,
     pub cdn_log_storage: CdnLogStorageConfig,
     pub cdn_log_queue: CdnLogQueueConfig,
     pub session_key: cookie::Key,
@@ -176,6 +183,7 @@ impl Server {
         Ok(Server {
             db: DatabasePools::full_from_environment(&base)?,
             storage,
+            cdn_log_counting_enabled: var("CDN_LOG_COUNTING_ENABLED")?.is_some(),
             cdn_log_storage: CdnLogStorageConfig::from_env()?,
             cdn_log_queue: CdnLogQueueConfig::from_env()?,
             base,
--- a/src/controllers/version/downloads.rs
+++ b/src/controllers/version/downloads.rs
@@ -33,9 +33,11 @@ pub async fn download(
     if let Some(version_id) = cache_result {
         app.instance_metrics.version_id_cache_hits.inc();
 
-        // The increment does not happen instantly, but it's deferred to be executed in a batch
-        // along with other downloads. See crate::downloads_counter for the implementation.
-        app.downloads_counter.increment(version_id);
+        if !app.config.cdn_log_counting_enabled {
+            // The increment does not happen instantly, but it's deferred to be executed in a batch
+            // along with other downloads. See crate::downloads_counter for the implementation.
+            app.downloads_counter.increment(version_id);
+        }
     } else {
         app.instance_metrics.version_id_cache_misses.inc();
 
@@ -67,9 +69,11 @@ pub async fn download(
                         get_version_id(&crate_name, &version, &mut conn)
                     })?;
 
-                    // The increment does not happen instantly, but it's deferred to be executed in a batch
-                    // along with other downloads. See crate::downloads_counter for the implementation.
-                    app.downloads_counter.increment(version_id);
+                    if !app.config.cdn_log_counting_enabled {
+                        // The increment does not happen instantly, but it's deferred to be executed in a batch
+                        // along with other downloads. See crate::downloads_counter for the implementation.
+                        app.downloads_counter.increment(version_id);
+                    }
 
                     Ok(Some(version_id))
                 } else {
--- a/src/tests/routes/crates/downloads.rs
+++ b/src/tests/routes/crates/downloads.rs
@@ -82,6 +82,28 @@ fn test_download() {
 }
 
 #[test]
+fn test_download_with_counting_via_cdn() {
+    let (app, anon, user) = TestApp::init()
+        .with_config(|config| config.cdn_log_counting_enabled = true)
+        .with_user();
+
+    app.db(|conn| {
+        CrateBuilder::new("foo", user.as_model().id)
+            .version(VersionBuilder::new("1.0.0"))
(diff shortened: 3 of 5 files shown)


LawnGnome (approved):
:+1:

One nagging thought at the back of my mind is if we should come with some way to log when a name/version pair _aren't_ found in the database, since that should be rare (only when a crate version is deleted), but that's currently hidden from us by the use of `INNER JOIN` in the save query. It probably isn't worth the hassle of running another query to figure that out, but I wonder if it's a useful signal that something's wrong if a bunch of crate versions aren't found.

eth3lbert:
Is there a chance that the single path is being picked by multiple workers simultaneously? Currently, #8081 only fails at the end when it tries to set processed. To prevent this, I suggest leveraging PostgreSQL's `advisory lock` at the beginning of the process to ensure only one worker processes the same log file at a time.

Turbo87:
> Is there a chance that the single path is being picked by multiple workers simultaneously?

the background worker is written in a way where rows in the `background_jobs` table are locked as long as the job is being run, which should prevent jobs from running in parallel on multiple workers.

the situation in which this could happen is if the `process_cdn_log_queue` job takes a long time to finish, and the cron job that schedules that job enqueues it again while the first is still running. in that case the second job could potentially see the same messages from the queue before they are deleted by either job and enqueue them as another `process_cdn_log` job. in that case we would have two `process_cdn_log` jobs with the same path in the background worker queue. the chances of this happening are non-zero, but seem pretty small to me though.

> Currently, #8081 only fails at the end when it tries to set processed.

My plan is to put the `INSERT INTO processed_log_files` into the same transaction as the other database writing code in this job. My understanding is that if two jobs for the same path run in parallel, they might count the downloads in parallel, but the database writing of the slower job would be suspended until the faster job has committed its transaction, at which point the slower job would fail with a uniqueness error.

If a third job is later scheduled for the same path the initial check on the `processed_log_files` table would prevent it from even starting to count the downloads to prevent any unnecessary work.

> we should come with some way to log when a name/version pair _aren't_ found in the database

yeah, I had similar thoughts. we could potentially add a nullable `version_id` column to the temporary database, have a `UPDATE` query fill that through the inner joins (?), and then user the `version_id` column for the inserts into to `version_downloads` table. Afterwards we could query the temporary table for `version_id IS NULL` and report it back. I have no clue yet how much extra work that would be for the database though.

eth3lbert:
> yeah, I had similar thoughts. we could potentially add a nullable `version_id` column to the temporary database, have a `UPDATE` query fill that through the inner joins (?), and then user the `version_id` column for the inserts into to `version_downloads` table. Afterwards we could query the temporary table for `version_id IS NULL` and report it back. I have no clue yet how much extra work that would be for the database though.

Could try something like following:
```sql
WITH joined_data AS (
    SELECT versions.id, temp_downloads.*
    FROM temp_downloads
    LEFT JOIN crates ON crates.name = temp_downloads.name
    LEFT JOIN versions ON versions.num = temp_downloads.version AND versions.crate_id = crates.id
), inserted AS (
    INSERT INTO version_downloads (version_id, date, downloads)
    SELECT joined_data.id, joined_data.date, joined_data.downloads
    FROM joined_data
    WHERE joined_data.id IS NOT NULL
    ORDER BY (joined_data.id, joined_data.date)
    ON CONFLICT (version_id, date)
    DO UPDATE SET downloads = version_downloads.downloads + EXCLUDED.downloads
    RETURNING version_downloads.version_id
)
SELECT joined_data.name, joined_data.version, joined_data.date
FROM joined_data
LEFT JOIN ins ON joined_data.id = inserted.version_id
WHERE joined_data.id IS NULL
;
```

Turbo87:
- is there a reason for the `ORDER BY (joined_data.id, joined_data.date)` part?
- do we need the `LEFT JOIN inserted ON joined_data.id = inserted.version_id`? does it skip the `INSERT INTO` query without using the result in the main query?

eth3lbert:
> * is there a reason for the `ORDER BY (joined_data.id, joined_data.date)` part?

Whoops! I just updated the SQL based on my own testing 😅. It's safe to remove that part now. 

>  * do we need the `LEFT JOIN inserted ON joined_data.id = inserted.version_id`? does it skip the `INSERT INTO` query without using the result in the main query?

The following example is taken from official docs:
```sql
WITH t AS (
    UPDATE products SET price = price * 1.05
    RETURNING *
)
SELECT * FROM products;
```
the outer `SELECT` would return the original prices before the action of the `UPDATE`, while in

```sql
WITH t AS (
    UPDATE products SET price = price * 1.05
    RETURNING *
)
SELECT * FROM t;
```
the outer `SELECT` would return the updated data.

Ref: https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-MODIFYING

Turbo87:
> the outer `SELECT` would return the original prices before the action of the `UPDATE`

makes sense, but since our outer query is not returning the `version_downloads.downloads` column I guess there is no difference?

I've just tried the above query on a production dataset. The internal caching in the database makes the performance a bit hard to compare, but with a warm cache it looks like the proposed query is roughly similarly fast as the one in the current iteration of the PR.

I've also verified that commenting out the two lines mentioned in https://github.com/rust-lang/crates.io/pull/8096#issuecomment-1936055398 gives the same results.

I was initially a bit worried because the query resulted in 500+ rows of name/version combinations that did not exist in my local database clone, but after importing the most recent database dump it was down to a single row. That single row was about the `note` crate, which was a name-squatted crate that we removed a while ago but that was still present in the old CDN log file I was using.

tl;dr the query proposed above seems to be working great and I'll adjust the PR to use it instead :)

eth3lbert:
> makes sense, but since our outer query is not returning the `version_downloads.downloads` column I guess there is no difference?

"Data-modifying statements in WITH are executed exactly once, and always to completion, independently of whether the primary query reads all (or indeed any) of their output. Notice that this is different from the rule for SELECT in WITH: as stated in the previous section, execution of a SELECT is carried only as far as the primary query demands its output."

Based on what I read, I agree!

> tl;dr the query proposed above seems to be working great and I'll adjust the PR to use it instead :)

I'm glad it helped :)

LawnGnome (approved):
That query looks good to me as well. Thanks, @eth3lbert!

Issue #2055, opened 2019-12-23
Redundant database indexes


smarnach:
I believe that these database indexes are redundant (mostly with some unique constraints for primary keys):
```
api_tokens_token_idx
index_crates_categories_crate_id
index_crates_keywords_crate_id
index_follows_user_id
index_keywords_keyword
```
This could be verified by people with production database access by comparing the number of index scans for them on two consecutive days. If the indexes are not scanned in a full day, we can safely remove them.

If they _are_ scanned, it doesn't necessarily follow that they are still needed, though. If an index is redundant with some unique constraint, Postgres might still use the index rather than the unique constraint in some cases.

The list is probably incomplete, and may include indexes that are still needed.

carols10cents:
> If they are scanned, it doesn't necessarily follow that they are still needed, though. If an index is redundant with some unique constraint, Postgres might still use the index rather than the unique constraint in some cases.

lol yep, that's what it looks like for `api_tokens_token_idx`, it's being used instead of `api_tokens_token_key`:

```
   relid   | indexrelid | schemaname |          relname           |               indexrelname                |  idx_scan  | idx_tup_read | idx_tup_fetch
-----------+------------+------------+----------------------------+-------------------------------------------+------------+--------------+---------------
   9164079 |    9164097 | public     | api_tokens                 | api_tokens_token_idx                      |       6232 |         6476 |          6116
   9164079 |    9164090 | public     | api_tokens                 | api_tokens_token_key                      |          0 |            0 |             0
```

It looks like all of the indexes on `crates_categories` are being used? Which index looked like it was redundant with `index_crates_categories_crate_id`?

```
select * from pg_stat_user_indexes where relname = 'crates_categories';
 relid | indexrelid | schemaname |      relname      |            indexrelname             | idx_scan | idx_tup_read | idx_tup_fetch
-------+------------+------------+-------------------+-------------------------------------+----------+--------------+---------------
 25523 |      28185 | public     | crates_categories | index_crates_categories_crate_id    |  3923405 |      4479874 |       4474966
 25523 |      28184 | public     | crates_categories | index_crates_categories_category_id |    16306 |      9102625 |             0
 25523 |    2931343 | public     | crates_categories | crates_categories_pkey              |  4383396 |      2266337 |       2264086
```

Similarly with `index_crates_keywords_crate_id`:

```
select * from pg_stat_user_indexes where relname = 'crates_keywords';
[...]

PR #8112, opened 2024-02-12
database: Remove unused `index_follows_user_id` index

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/2055

/cc @eth3lbert @paolobarbolini


eth3lbert:
LGTM.

Digging a little deeper, the index should only be used on two routes: `/crates?following` and `/me/updates`. However, the `pg_stat_user_indexes` report indicates that the index is currently unused. Therefore, I think it's safe to remove it, for now. If we encounter slow queries later, we can still reintroduce it.

PR #8127, opened 2024-02-15
downloads: Avoid `spawn_blocking()` and database query if `cdn_log_counting_enabled` is set

Turbo87:
Previously, we were just not calling `downloads_counter.increment()` if the feature flag is enabled. This commit changes the code to not perform the `version_id` database query, not get a database connection from the pool, and not get a blocking thread from tokio via `spawn_blocking()`. This should speed up the download endpoint, and make it resilient against database issues.

Best reviewed in ignore-whitespace mode 😉

### DIFF
--- a/src/controllers/version/downloads.rs
+++ b/src/controllers/version/downloads.rs
@@ -22,97 +22,95 @@ pub async fn download(
 ) -> AppResult<Response> {
     let wants_json = req.wants_json();
 
-    let cache_key = (crate_name.to_string(), version.to_string());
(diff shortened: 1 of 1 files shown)


hi-rustin (approved):
If I understand correctly. We only have one purpose of version_id_cacher which is to speed up the download endpoint. so if we don't count it in the API then it makes sense to skip the cache.

Looks good to me. Thanks! 👍

LawnGnome (approved):

PR #8131, opened 2024-02-15
worker/jobs/downloads/update_metadata: Implement optional batch processing

Turbo87:
The previous implementation to update the `versions` and `crates` table was quite slow and creating a lot of bloat in the database, since it updated the `crates` and `metadata` tables for every single version that was downloaded.

This commit adds a second implementation, guarded by the `BATCH_UPDATE_DOWNLOADS` feature flag, which updates the tables in batches with a static query that does not need to transfer data back and forth between the background worker and the database.

For more context, see https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/error.20in.20crates.2Eio.20summary.

### DIFF
--- a/src/worker/jobs/downloads/update_metadata.rs
+++ b/src/worker/jobs/downloads/update_metadata.rs
@@ -5,6 +5,7 @@ use crate::worker::Environment;
 use crates_io_worker::BackgroundJob;
 use diesel::prelude::*;
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 
 #[derive(Serialize, Deserialize)]
 pub struct UpdateDownloads;
@@ -17,24 +18,52 @@ impl BackgroundJob for UpdateDownloads {
     async fn run(&self, env: Self::Context) -> anyhow::Result<()> {
         spawn_blocking(move || {
             let mut conn = env.connection_pool.get()?;
-            update(&mut conn)?;
+            update(&mut conn, env.config.batch_update_downloads)?;
             Ok(())
         })
         .await
     }
 }
 
-fn update(conn: &mut PgConnection) -> QueryResult<()> {
+fn update(conn: &mut PgConnection, use_batch_update: bool) -> QueryResult<()> {
     use diesel::dsl::now;
     use diesel::select;
 
-    let rows = version_downloads::table
-        .filter(version_downloads::processed.eq(false))
-        .filter(version_downloads::downloads.ne(version_downloads::counted))
-        .load(conn)?;
+    if use_batch_update {
+        info!("Updating versions…");
 
-    info!(rows = rows.len(), "Updating versions");
-    collect(conn, &rows)?;
+        // After 45 minutes, we stop the batch updating process to avoid
+        // triggering the long-running job alert. 15 minutes later, the job
+        // will be started again by our cron service anyway.
+        const TIME_LIMIT: Duration = Duration::from_secs(45 * 60);
+
+        // We update the `downloads` columns in batches to a) avoid the
+        // back and forth between the application and the database and b) avoid
+        // holding locks on any of the involved tables for too long.
+        const BATCH_SIZE: i64 = 5_000;
+
+        let start_time = Instant::now();
+        loop {
+            let update_count = batch_update(BATCH_SIZE, conn)?;
+            info!("Updated {update_count} versions");
+            if update_count < BATCH_SIZE {
+                break;
+            }
+
+            if start_time.elapsed() > TIME_LIMIT {
+                info!("Time limit reached, stopping batch update");
+                break;
+            }
+        }
+    } else {
+        let rows = version_downloads::table
+            .filter(version_downloads::processed.eq(false))
+            .filter(version_downloads::downloads.ne(version_downloads::counted))
+            .load(conn)?;
+
+        info!(rows = rows.len(), "Updating versions");
+        collect(conn, &rows)?;
+    }
     info!("Finished updating versions");
 
     // Anything older than 24 hours ago will be frozen and will not be queried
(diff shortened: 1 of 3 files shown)


src/worker/jobs/downloads/update_metadata.rs
```diff
@@ -17,24 +18,52 @@ impl BackgroundJob for UpdateDownloads {
...
-
-    info!(rows = rows.len(), "Updating versions");
-    collect(conn, &rows)?;
+    if use_batch_update {
+        info!("Updating versions…");
+
+        // After 45 minutes, we stop the batch updating process to avoid
+        // triggering the long-running job alert. 15 minutes later, the job
+        // will be started again by our cron service anyway.
+        const TIME_LIMIT: Duration = Duration::from_secs(45 * 60);
```
LawnGnome, line 38: Do we actually want to do this? If this job is unable to keep up with the uncounted rows, is there an alert that would fire?

src/worker/jobs/downloads/update_metadata.rs
```diff
@@ -17,24 +18,52 @@ impl BackgroundJob for UpdateDownloads {
...
+
+        // We update the `downloads` columns in batches to a) avoid the
+        // back and forth between the application and the database and b) avoid
+        // holding locks on any of the involved tables for too long.
+        const BATCH_SIZE: i64 = 5_000;
```
LawnGnome, line 43: Empirically, my testing of this suggests that there aren't any locks held during the big update query. Assuming that's generally true, would we want to batch at all?

src/worker/jobs/downloads/update_metadata.rs
```diff
@@ -54,6 +83,83 @@ fn update(conn: &mut PgConnection) -> QueryResult<()> {
...
+    let query = format!(
+        r#"
+            WITH downloads_batch AS (
+                -- Select a batch of downloads (incl. `crate_id`) that have not been
+                -- counted yet.
+                SELECT crate_id, version_id, date, version_downloads.downloads - counted AS downloads
+                FROM version_downloads
+                INNER JOIN versions ON versions.id = version_id
+                WHERE NOT processed AND version_downloads.downloads != counted
+                LIMIT {batch_size}
```
LawnGnome, line 112: _eye twitches very slightly at string interpolation in a query, even though it's obviously safe here_

LawnGnome (approved):
I'm not sure about the behaviour around stopping the job if it runs for a long time, but otherwise LGTM.

Since I know we want to experiment with this and it's behind a feature flag, I'm happy to treat the feedback in here as non-blocking for now and we can iterate based on real world experience.

src/worker/jobs/downloads/update_metadata.rs
Turbo87: I tried binding it, but for whatever reason it resulted in an SQL syntax error that I couldn't figure out.

eth3lbert:
It's also worth noting that to mitigate performance degradation for other queries, particularly in non-batch scenarios, we could introduce a delay between iterations or after a specific number of iterations. This would help prevent excessive resource consumption by a single query.

src/worker/jobs/downloads/update_metadata.rs
eth3lbert: You could change the placeholder from `?` to `$1` then it should be fine.

eth3lbert:
> I'm not sure about the behaviour around stopping the job if it runs for a long time, but otherwise LGTM.

It would be great if we could use an explicit transaction to be more transparent.

src/worker/jobs/downloads/update_metadata.rs
Turbo87: indeed, that worked. thanks! :)

src/worker/jobs/downloads/update_metadata.rs
Turbo87: The way this worked previously is that on job startup it queried the full list of rows that need updates and then worked through them until all have been processed.

With this new query we are constantly querying new data too, which could lead to a situation where the loop never quits. From my estimations we probably should be done with this job in 3-5 min. with the new implementation, but just to be save I've added this limit.

I'd feel a bit uneasy if we used an infinite loop that could actually run forever, but I'm open to other suggestions :)

src/worker/jobs/downloads/update_metadata.rs
Turbo87: I think I prefer the batching so that we get regular status updates in the logs. If we happen to have 1M rows to update and then the job just doesn't say anything for 2 hours that would seem bad to me.

Turbo87:
> I'm not sure about the behaviour around stopping the job if it runs for a long time, but otherwise LGTM.

this should be similar to the previous behavior. the materialized view might not be updated and the `processed` columns might not get set, but according to what I've read the CTE query should be atomic like the transaction we used before.

src/worker/jobs/downloads/update_metadata.rs
LawnGnome: That's a good point, thanks. I think I still had the old behaviour of calculating the update set in advance in my head when I wrote that.

That said, I still think we'll have to figure out some sort of alerting strategy here in the longer term to avoid the problem where the job just falls further and further behind. I'm not sure exactly what that looks like before I'd had coffee at 7:39 am.

Turbo87: just as a data point: the runs so far have taken roughly 5-10 minutes, though sometimes interrupted by the deadlock issues I had mentioned

LawnGnome: That's good, then. Eventually this will probably be an issue, but presumably something else will break first.

PR #8145, opened 2024-02-18
worker/jobs/downloads/process_log: Remove unused `RETURNING` expression

Turbo87:
We're not using the results of `inserted` anywhere, so there should be no need for us to include a `RETURNING` expression in the subquery.

/cc @eth3lbert


eth3lbert:
LGTM 👍

PR #8146, opened 2024-02-18
Use a consistent update order for the `version_downloads` table

Turbo87:
Ensuring a consistent update order should reduce the risk of deadlocks between the two background jobs.

### DIFF
--- a/src/worker/jobs/downloads/process_log.rs
+++ b/src/worker/jobs/downloads/process_log.rs
@@ -330,6 +330,7 @@ fn save_to_version_downloads(conn: &mut PgConnection) -> QueryResult<Vec<NameAnd
                 SELECT joined_data.id, joined_data.date, joined_data.downloads
                 FROM joined_data
                 WHERE joined_data.id IS NOT NULL
+                ORDER BY joined_data.id, joined_data.date
                 ON CONFLICT (version_id, date)
                 DO UPDATE SET downloads = version_downloads.downloads + EXCLUDED.downloads
                 RETURNING version_downloads.version_id
--- a/src/worker/jobs/downloads/update_metadata.rs
+++ b/src/worker/jobs/downloads/update_metadata.rs
@@ -145,12 +145,19 @@ fn batch_update(batch_size: i64, conn: &mut PgConnection) -> QueryResult<i64> {
                     FROM downloads_batch
                 ) sum
                 WHERE sum.downloads > 0
+            ), sorted_downloads_batch AS (
+                -- Sort the `downloads_batch` CTE by `version_id` and `date` to
+                -- ensure that the `version_downloads` table is updated in a
+                -- consistent order to avoid deadlocks.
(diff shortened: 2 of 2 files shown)


src/worker/jobs/downloads/update_metadata.rs
```diff
@@ -145,12 +145,19 @@ fn batch_update(batch_size: i64, conn: &mut PgConnection) -> QueryResult<i64> {
                     FROM downloads_batch
                 ) sum
                 WHERE sum.downloads > 0
+            ), sorted_downloads_batch AS (
+                -- Sort the `downloads_batch` CTE by `version_id` and `date` to
+                -- ensure that the `version_downloads` table is updated in a
+                -- consistent order to avoid deadlocks.
+                SELECT version_id, date, downloads FROM downloads_batch
+                ORDER BY version_id, date
```
eth3lbert, line 153: We might need a `SELECT ORDER BY FOR UPDATE` to guarantee lock order.
Something like:
```sql
SELECT downloads_batch.*
FROM version_downloads
JOIN downloads_batch using (version_id, date)
ORDER BY version_id, date
FOR UPDATE
```

src/worker/jobs/downloads/process_log.rs
eth3lbert, line 1: Same reason for process_log we want to add another cte.
```sql
WITH joined_data AS (
    SELECT versions.id, temp_downloads.*
    FROM temp_downloads
    LEFT JOIN crates ON crates.name = temp_downloads.name
    LEFT JOIN versions ON versions.num = temp_downloads.version AND versions.crate_id = crates.id
), sorted_joined_data (
    SELECT joined_data.*
    FROM version_downloads
    JOIN joined_data ON (version_downloads.version_id, version_downloads.date) = (joined_data.id, joined_data.date)
    ORDER BY version_id, date
    FOR UPDATE
) , inserted AS (
    INSERT INTO version_downloads (version_id, date, downloads)
    SELECT sorted_joined_data.id, sorted_joined_data.date, sorted_joined_data.downloads
    FROM sorted_joined_data
    ON CONFLICT (version_id, date)
    DO UPDATE SET downloads = version_downloads.downloads + EXCLUDED.downloads
)
SELECT joined_data.name, joined_data.version
FROM joined_data
WHERE joined_data.id IS NULL;
```

eth3lbert: We could also change the `id` to `version_id` which would make `JOIN ... USING` clause possible.

Turbo87: does that still work if the join does not find a match row?

Turbo87: I suspect that you're right and that we'll need something like it, but I'm curious if the current state with the `INSERT ... SELECT ... ORDER BY` works well enough :D

If we're still seeing deadlocks afterwards we can put the above into a follow-up PR :)

PR #8155, opened 2024-02-19
database: Drop `badges` table

Turbo87:
We haven't returned the data from the API for two years and haven't shown the badges on the frontend for four years, so it's time to get rid of the unused data in the database.

Related:

- https://github.com/rust-lang/crates.io/issues/2436
- https://github.com/rust-lang/crates.io/pull/2440
- https://github.com/rust-lang/crates.io/pull/5071


Turbo87:
for future reference: the patch file was regenerated using the instructions in https://github.com/rust-lang/crates.io/pull/6531#issuecomment-1952289014

PR #8182, opened 2024-02-23
Add async database connection pools based on `deadpool`

Turbo87:
We're currently using https://docs.rs/r2d2/ for our database connection pools, but unfortunately that is not supported by [diesel_async](https://github.com/weiznich/diesel_async) (see https://github.com/weiznich/diesel_async?tab=readme-ov-file#built-in-connection-pooling-support).

This PR introduces additional database connection pools based on `deadpool` next to the existing r2d2 pools and migrates one API endpoint and one background job to use these async connection pools instead.

All of the configuration options of the sync connection pools are still supported, with one exception: `min_size()`. `deadpool` does not shrink the pool size automatically and once `max_size` is reached these connections will be kept forever. This does not seem like a huge deal for our use case though, and if we discover that we need it we can build it ourselves using regular `Pool::retain()` calls.

Note that this does not migrate to diesel_async yet, but it meant as an intermediate step to ease the migration to that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -990,6 +990,8 @@ dependencies = [
  "crates_io_test_db",
  "crates_io_worker",
  "dashmap",
+ "deadpool",
+ "deadpool-diesel",
  "derive_builder 0.20.0",
  "derive_deref",
  "dialoguer",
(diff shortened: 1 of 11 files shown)


LawnGnome (approved):
LGTM. None of my comments are blockers.

One overarching slight worry I have is around our database instance itself: I presume we know we have the overhead for doubling the number of connections because we used to have twice as many `web` processes, but I would like us to be sure that we know what our limits are (if any) before we release this to production.

Turbo87:
> I presume we know we have the overhead for doubling the number of connections because we used to have twice as many `web` processes

yep, this shouldn't be a problem. in addition to that, the new async pools use dedicated environment variables for the pool size, so we can tune both pool sizes individually on production.

```
$ SHOW max_connections;
500
```

PR #8194, opened 2024-02-26
Fix the issue of relevant search not working with spaces in them when using seek pagination

eth3lbert:
Revert `to_tsquery_with_search_config` usage for `TsQuery` to the original implementation to address the issue. This is the same process as in PR #8053.


Turbo87 (approved):
thank you! :)

PR #8200, opened 2024-02-27
Remove `BATCH_UPDATE_DOWNLOADS` feature flag

Turbo87:
We've very successfully used this on production for almost two weeks now. There does not appear to be any reason for switching it back off, so this commit is removing the feature flag and permanently enables the batch processing.


LawnGnome (approved):
kill-it-with-fire-fire

Issue #8049, opened 2024-02-01
Admin action plan


LawnGnome:
_Previously, on crates.io gaining support for admin actions without needing to run `crates-admin`..._

Adding support for admin actions in UI has been identified as a priority for a while. Last year, I opened #6811 (among other PRs) to start fleshing that out as a separate admin console using server side rendering. I didn't particularly like the code, it was warty, and (as it turned out) other things came up that demanded a bit more of my attention like artifact signing and malware detection, so it hasn't progressed since then.

More recently, @Turbo87 merged #7852, which provides minimal support for the concept of crates.io admins, and uses it to allow them to yank and unyank crates.

This is a good first step, and I definitely prefer the idea of incorporating admin functionality more directly into crates.io rather than building a new frontend, but is incomplete — there are other actions that we also need to be able to perform, it's probably a little too easy for us to accidentally perform an action right now[^reversibility], and the logging is arguably too ephemeral at present.

Here's what I'd like to do:

1. Only allow admin actions when an admin explicitly opts in. Practically, this would involve something like GitHub's sudo mode (so you only get privileges for a limited period in one session), and we may want to consider requiring crates.io-specific 2FA as part of flipping that switch on.
2. Migrate logging of admin actions into the database, rather than logging them into the normal logging machinery, so that we can retain them indefinitely, _and_
3. Require an admin action to include an explanation for why it's being taken (in most cases, presumably just a link to a Zulip thread or Zendesk ticket).
4. Finally, expand the list of admin actions in the UI to include:
   1. User locking and unlocking
   2. Mass crate yanking and unyanking
   3. Crate and crate version deletion[^deletion]
   4. _your idea here_

Feedback encouraged, @rust-lang/crates-io!

[^deletion]: This will probably require further thought on implementation, since it's _not_ trivially reversible. @walterhpearce suggested putting it on a timer of a few minutes to allow easy cancellation _before_ the action is taken.
[^reversibility]: Of course, yank and unyank actions are trivially reversible, so it's not _really_ that bad in practice when that's all we support. A misclick can be fixed with another click.

hi-rustin:
> 4\. UI to include

Do you mean we need to implement it in our current frontend crates.io UI instead of an admin console UI?



> 2\. Mass crate yanking and unyanking

I think we should create a separate page on crates.io if we want to add a base to our current frontend. Is that correct?

hi-rustin:
[...]

PR #8210, opened 2024-02-29
Add sudo mode for admins

LawnGnome:
This PR adds a concept of "sudo mode" for admins logged into crates.io. Actions that require admin privileges will be disabled by default unless the admin explicitly turns on admin actions from the user menu, at which point they will be given privileges for six hours or until they disable admin actions again from the user menu. Sudo mode is per-user, per-browser; it is _not_ synced across multiple browser sessions, even for the same user.

These cases look like this:

image

image

In addition to this, a new `PrivilegedAction` component has been added that can gracefully handle rendering controls that can be used either by a regular user or an admin in sudo mode, including showing the control in a disabled mode when an admin is logged in but they have not enabled sudo mode. The disabled mode looks like this:

image

(I don't know why the tooltip wraps, but I don't really care much, honestly.)

Finally, the existing yank/unyank functionality has been wrapped in `PrivilegedAction`, both because it needs to be anyway, and because it demonstrates how this is used in practice.

This is likely best reviewed commit by commit. (In theory, each of these could be a PR unto themselves, but I don't see much point splitting this out without the motivating example of how it interacts with the yank button.)

Open questions:

* Is six hours the appropriate amount of time for sudo mode to be enabled by default? (In theory, we could actually allow the admin to specify how long they want, but that seems unnecessarily complex.)
* Is it OK that sudo mode is per-browser?
* Do we want a visible indicator next to the user avatar that sudo mode is enabled? I've used a :mage: emoji as a placeholder, but if we want this, we probably want an actual icon of some kind.
* Should this just be called "sudo mode" in the UI as well?
* How many lints have I accidentally broken because I can't figure out how to configure my local environment in the same way as CI expects? :grimacing: 

This implements step 1 of the plan laid out in #8049.

### DIFF
--- a/app/components/header.hbs
+++ b/app/components/header.hbs
@@ -21,7 +21,7 @@
       {{#if this.session.currentUser}}
         <Dropdown data-test-user-menu as |dd|>
           <dd.Trigger local-class="dropdown-button" data-test-toggle>
-            <UserAvatar @user={{this.session.currentUser}} @size="small" local-class="avatar" data-test-avatar />
+            <UserAvatar @user={{this.session.currentUser}} @size="small" @sudo={{this.session.isSudoEnabled}} local-class="avatar" data-test-avatar />
             {{ this.session.currentUser.name }}
           </dd.Trigger>
 
--- a/app/components/header.hbs
+++ b/app/components/header.hbs
@@ -21,7 +21,10 @@
       {{#if this.session.currentUser}}
         <Dropdown data-test-user-menu as |dd|>
           <dd.Trigger local-class="dropdown-button" data-test-toggle>
-            <UserAvatar @user={{this.session.currentUser}} @size="small" @sudo={{this.session.isSudoEnabled}} local-class="avatar" data-test-avatar />
+            {{#if this.session.isSudoEnabled}}
+              <div local-class="wizard-hat">🧙</div>
+            {{/if}}
+            <UserAvatar @user={{this.session.currentUser}} @size="small" local-class="avatar" data-test-avatar />
             {{ this.session.currentUser.name }}
           </dd.Trigger>
 
--- a/app/components/header.hbs
+++ b/app/components/header.hbs
@@ -25,11 +25,25 @@
             {{ this.session.currentUser.name }}
           </dd.Trigger>
 
-          <dd.Menu local-class="current-user-links" as |menu|>
-            <menu.Item><LinkTo @route="dashboard">Dashboard</LinkTo></menu.Item>
-            <menu.Item><LinkTo @route="settings" data-test-settings>Account Settings</LinkTo></menu.Item>
-            <menu.Item><LinkTo @route="me.pending-invites">Owner Invites</LinkTo></menu.Item>
-            <menu.Item local-class="menu-item-with-separator">
+          <dd.Menu local-class='current-user-links' as |menu|>
+            <menu.Item><LinkTo @route='dashboard'>Dashboard</LinkTo></menu.Item>
+            <menu.Item><LinkTo @route='settings' data-test-settings>Account Settings</LinkTo></menu.Item>
+            <menu.Item><LinkTo @route='me.pending-invites'>Owner Invites</LinkTo></menu.Item>
+            {{#if this.session.isAdmin}}
+              <menu.Item local-class='sudo'>
+                {{#if this.session.isSudoEnabled}}
+                  <button local-class='sudo-menu-item' type='button' {{on 'click' this.disableSudo}}>
+                    Disable admin actions
+                    <div local-class='expires-in'>expires at {{date-format this.session.sudoEnabledUntil 'HH:mm'}}</div>
+                  </button>
+                {{else}}
+                  <button local-class='sudo-menu-item' type='button' {{on 'click' this.enableSudo}}>
+                    Enable admin actions
+                  </button>
+                {{/if}}
+              </menu.Item>
+            {{/if}}
+            <menu.Item local-class='menu-item-with-separator'>
               <button
                 type="button"
                 disabled={{this.session.logoutTask.isRunning}}
--- a/app/components/header.js
+++ b/app/components/header.js
@@ -1,6 +1,19 @@
+import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import Component from '@glimmer/component';
 
 export default class Header extends Component {
+  /** @type {import("../services/session").default} */
   @service session;
+
+  @action
+  enableSudo() {
+    // FIXME: hard coded six hour duration.
+    this.session.setSudo(6 * 60 * 60 * 1000);
+  }
+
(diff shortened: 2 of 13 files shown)


app/components/header.js
```diff
@@ -1,6 +1,19 @@
...
 import Component from '@glimmer/component';
 
 export default class Header extends Component {
+  /** @type {import("../services/session").default} */
   @service session;
+
+  @action
+  enableSudo() {
+    // FIXME: hard coded six hour duration.
+    this.session.setSudo(6 * 60 * 60 * 1000);
```
Turbo87, line 12: maybe extract this to a named constant to avoid the random magic number :)

app/components/header.module.css
```diff
@@ -184,3 +185,13 @@
         margin-right: var(--space-2xs);
     }
 }
+
+.sudo > button {
+    flex-direction: column !important;
```
Turbo87, line 190: I assume there is a reason for the `!important` here? no way to avoid it?

app/components/header.module.css
```diff
@@ -184,3 +185,13 @@
         margin-right: var(--space-2xs);
     }
 }
+
+.sudo > button {
```
Turbo87, line 189:
```suggestion
.sudo-menu-item {
```

or am I missing something? 😅

app/components/user-avatar.hbs
```diff
@@ -6,4 +6,7 @@
   title={{this.title}}
   decoding="async"
   ...attributes
-/>
\ No newline at end of file
+/>
+{{#if this.sudo}}
```
Turbo87, line 10:
```suggestion
{{#if @sudo}}
```

then we don't need the `get sudo()` at all

app/components/user-avatar.hbs
```diff
@@ -6,4 +6,7 @@
...
+  🧙
+{{/if}}
```
Turbo87, line 12: ... but I'm actually not convinced that the `UserAvatar` component is the right place for this. I think I'd rather prefer this in the `Header` component since the `UserAvatar` component is used in a couple of other places where the sudo icon does not make much sense.

app/services/session.js
```diff
@@ -30,6 +40,36 @@ export default class SessionService extends Service {
     }
   }
 
+  get isAdmin() {
+    return this.currentUser?.is_admin === true;
+  }
+
+  get isSudoEnabled() {
+    return this.currentUser?.is_admin === true && this.sudoEnabledUntil !== null && this.sudoEnabledUntil >= Date.now();
```
Turbo87, line 48:
```suggestion
    return this.isAdmin && this.sudoEnabledUntil !== null && this.sudoEnabledUntil >= Date.now();
```

app/components/privileged-action.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * @param {Element} element
+   * @return {void}
+   */
+  disableComponents(element) {
+    [...element.querySelectorAll('input, select, textarea, button')].forEach(control => {
+      control.setAttribute('disabled', 'disabled');
+    });
```
Turbo87, line 53: I'm a bit worried about this part since it breaks the declarative nature of the template system. Did you test this in combination with a button that has `disabled` dynamically specified in the template too? It would make it a bit hard to understand what definition is "winning".

app/components/header.hbs
```diff
@@ -21,15 +21,29 @@
...
-            <menu.Item><LinkTo @route="settings" data-test-settings>Account Settings</LinkTo></menu.Item>
-            <menu.Item><LinkTo @route="me.pending-invites">Owner Invites</LinkTo></menu.Item>
-            <menu.Item local-class="menu-item-with-separator">
+          <dd.Menu local-class='current-user-links' as |menu|>
+            <menu.Item><LinkTo @route='dashboard'>Dashboard</LinkTo></menu.Item>
+            <menu.Item><LinkTo @route='settings' data-test-settings>Account Settings</LinkTo></menu.Item>
+            <menu.Item><LinkTo @route='me.pending-invites'>Owner Invites</LinkTo></menu.Item>
+            {{#if this.session.isAdmin}}
+              <menu.Item local-class='sudo'>
+                {{#if this.session.isSudoEnabled}}
```
Turbo87, line 34: Does this update correctly when the sudo period is over? Since `isSudoEnabled` uses `Date.now()` instead of a "tracked" thing we might need something to tell the template to rerender/reevaluate every X seconds.

I think the way I would architect it to avoid the issue is by using http://ember-concurrency.com: we would have a restartable `sudoEnabledTask` on the `session` service with just `await rawTimeout(6 hours)` in it, and then we can use the derived property `this.sudoEnabledTask.isRunning` to check if sudo mode is enabled or not. that property should automatically cause the template to update once the task has finished (i.e. after the 6 hour timeout). to manually disable the sudo mode we can use `this.sudoEnabledTask.cancelAll()`.

app/components/header.module.css
LawnGnome: Only the 4123 refactors that occurred before I pushed the branch.

app/components/user-avatar.hbs
LawnGnome: Fair enough. Moved.

app/components/header.hbs
LawnGnome: > Does this update correctly when the sudo period is over?

No, and I meant to put that in the PR description and clearly forgot when a giant golden retriever started pointing out that he was late for his game of fetch last night.

I like the task approach, so I've implemented that.

app/components/privileged-action.js
LawnGnome: If I understand how render modifiers work correctly (I probably don't), then `did-insert` fires after the Ember templating has executed, and therefore (at most) we'd be re-setting `disabled` on the element.

Empirically, hacking together a custom component with an element with a dynamic `disabled` attribute in the template and putting it inside a `PrivilegedAction` seems to work how I'd expect: if the placeholder state is active, then the element is disabled regardless of its own attribute.

I can't figure out a good way to express this in a unit test, but I'm open to suggestions.

LawnGnome:
I've pushed new commits in response to the review. (Thanks!)

To simplify re-review, I haven't rebased them into the previous set of commits, but (if approved) I'll reorder and squash them down as appropriate before merging.

app/components/privileged-action.js
Turbo87: I did a little bit more research and it looks like another alternative could be https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#disabled. It should allow us to replace the imperative code above with declarative template code.

Turbo87 (approved):
One thing we should add for this change is an acceptance/application test that asserts that the full flow with the sudo mode is working correctly. This can be added in a follow-up PR though and this seems useful enough that I'll merge it without the test already :)

PR #8225, opened 2024-03-03
Refactor query to use crate_id only instead of Crate entity

eth3lbert:
This PR eliminates the need to use the entire `Crate` entity when only the `crate_id` is required.


Turbo87 (approved):
👍

PR #8232, opened 2024-03-04
Extract `crate_downloads` table

Turbo87:
This PR is the first in a sequence of multiple PRs that move the `crates.downloads` column to a dedicated database table. 

Whenever we update the `crates.downloads` column a full new row is written by Postgres, which causes a lot of churn for the columns in this table that don't actually change that much. We've been seeing quite a few performance issues lately that are supposedly (partially) caused by this behavior.

This PR adds a database migration which creates a new `crate_downloads` table, with a commented out query in the migration script to fill the table with the current values. The migration also adds an insert trigger on the `crates` table to ensure a corresponding `crate_downloads` row is automatically created for all newly inserted crates. Deletion of crates is automatically handled by `on delete cascade`.

The PR also adjusts the `update_downloads` background job to update both `crates.downloads` and `crate_downloads.downloads` for now. We can remove the `crates.downloads` updates once we switch the backend endpoint to read from the `crate_downloads` table instead.


LawnGnome (approved):

src/worker/jobs/downloads/update_metadata.rs
eth3lbert, line 1: Given that the `crates::downloads` column is not being dropped in this PR, should we still assert its correctness?

eth3lbert, line 1: Should we expand the test data for crates and their versions to ensure we are updating the correct one?

eth3lbert (approved):
LGTM! My comments are only suggestions, not blockers.

src/worker/jobs/downloads/update_metadata.rs
Turbo87: good point, I've adjusted the commit to only add new assertion instead of replacing the existing ones

Turbo87: I guess that would generally make sense. Since that was already a problem with the existing code I don't want to conflate that with the purpose of this PR though :)

PR #8233, opened 2024-03-04
Remove `crates.downloads` column

Turbo87:
This PR builds on top of https://github.com/rust-lang/crates.io/pull/8232. It switches the backend code to use the `crate_downloads` table for all downloads queries, and it subsequently drops the obsolete `crates.downloads` column.

This should be deployed separately from https://github.com/rust-lang/crates.io/pull/8232!


eth3lbert:
I haven't started reviewing this PR yet. However, I noticed that index-related migration seems to be missing. If this is intentional, I think it would be helpful to include a comment in the migration script explaining the rationale behind this decision.

Turbo87:
> I noticed that index-related migration seems to be missing. If this is intentional

good catch, that was indeed not intentional.

eth3lbert:
Yes, and we should also consider adding an index on `crate_downloads`  because there are sortings by downloads.

Turbo87:
@eth3lbert I've added the `create index` calls to the `down.sql` of the second migration

I think this is ready to go now 🤔

eth3lbert (approved):
🎉 LGTM 👍

PR #8237, opened 2024-03-04
Add sudo mode acceptance tests, per #8210 feedback

LawnGnome:
Whichever stage of sudo grief we're at, we're now at acceptance.

I don't think there's anything surprising here. I used the yank button as a proxy for the general `PrivilegedAction` functionality rather than creating a new test-only component, since it's already on `main` and is unlikely to change significantly. Unit tests are covering all the actually interesting bits anyway.

The one thing that _I_ was surprised by is that the `qunit-dom` helpers were capable of clicking a button in a disabled `fieldset`, even though that's not possible from the browser UI, which means I can't write a sensible "failed to yank" style test. (I tried a few different ways just to be sure.)

### DIFF
--- a/tests/acceptance/sudo-test.js
+++ b/tests/acceptance/sudo-test.js
@@ -1,0 +1,100 @@
+import { click, find, waitFor } from '@ember/test-helpers';
+import { module, test } from 'qunit';
+
+import format from 'date-fns/format';
+
+import { setupApplicationTest } from 'crates-io/tests/helpers';
+
+import { visit } from '../helpers/visit-ignoring-abort';
+
+module('Acceptance | sudo', function (hooks) {
+  setupApplicationTest(hooks);
+
+  const prepare = context => {
+    const user = context.server.create('user', {
+      login: 'johnnydee',
+      name: 'John Doe',
+      email: 'john@doe.com',
+      avatar: 'https://avatars2.githubusercontent.com/u/1234567?v=4',
+      isAdmin: true,
+    });
+
+    const crate = context.server.create('crate', {
+      name: 'foo',
+      newest_version: '0.1.0',
+    });
+
+    const version = context.server.create('version', {
+      crate,
+      num: '0.1.0',
+    });
+
+    context.authenticateAs(user);
+    return { user, crate, version };
+  };
+
+  test('admin user is not initially in sudo mode', async function (assert) {
+    prepare(this);
+
+    await visit('/crates/foo/versions');
+
+    // Test the various header elements.
+    assert.dom('[data-test-wizard-hat]').doesNotExist();
+    assert.dom('[data-test-disable-admin-actions]').doesNotExist();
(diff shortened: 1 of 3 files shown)


Turbo87:
> The one thing that _I_ was surprised by is that the `qunit-dom` helpers were capable of clicking a button in a disabled `fieldset`

The `click()` fn comes from https://github.com/emberjs/ember-test-helpers, not `qunit-dom`, which only deals with DOM state assertions. 

https://github.com/emberjs/ember-test-helpers/blob/master/addon/addon-test-support/%40ember/test-helpers/dom/click.ts is the implementation of `click()`, which emulates the native events that the browser would fire when a users clicks a button. But admittedly https://github.com/emberjs/ember-test-helpers/blob/9cec68dc6aa9c0a7a449eb89797eb81299fa727f/addon/addon-test-support/%40ember/test-helpers/dom/click.ts#L116-L118 only checks if the element itself is disabled, but does not care about parent fieldsets or forms.

tests/acceptance/sudo-test.js
```diff
@@ -0,0 +1,100 @@
...
+import format from 'date-fns/format';
+
+import { setupApplicationTest } from 'crates-io/tests/helpers';
+
+import { visit } from '../helpers/visit-ignoring-abort';
+
+module('Acceptance | sudo', function (hooks) {
+  setupApplicationTest(hooks);
+
+  const prepare = context => {
```
Turbo87, line 13:
```suggestion
  function prepare(context) {
```

or is there a reason why this needs to be an arrow fn? :)

tests/acceptance/sudo-test.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+    assert.strictEqual(seen, 1);
+
+    // Test that the fieldset is not present.
+    assert.dom('[data-test-placeholder-fieldset]').doesNotExist();
+    assert.dom('[data-test-version-yank-button="0.1.0"]').exists();
+  });
+
+  test('admin can yank a crate in sudo mode', async function (assert) {
+    this.server.loadFixtures();
```
Turbo87, line 85: do we need the fixtures when we use the `prepare()` fn?

Turbo87 (commented):
LGTM 👍 

we might want an additional test though that checks that the sudo mode controls are invisible when the user is not an admin

tests/acceptance/sudo-test.js
LawnGnome: The reason is about ten years of deeply ingrained house styles at multiple companies. :see_no_evil: 

I'll change it.

tests/acceptance/sudo-test.js
LawnGnome: No, that was a refactoring mistake that left it in.

LawnGnome:
> we might want an additional test though that checks that the sudo mode controls are invisible when the user is not an admin

Good call; added.

PR #8244, opened 2024-03-05
controllers/krate/search: Read downloads from `crate_downloads` table

Turbo87:
This PR builds on top of https://github.com/rust-lang/crates.io/pull/8232 and switches our `GET /api/v1/crates` endpoint over to using the `crate_downloads` table instead of the `crates.downloads` column.

Performance tests on the production replica database indicated that this should not significantly affect the performance of the endpoint (see https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/.60crate_downloads.60.20table).

This should allow us to eventually remove the `crates.downloads` column, reducing the number of writes to the `crates` table.

### DIFF
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -219,11 +219,13 @@ pub async fn search(app: AppState, req: Parts) -> AppResult<Json<Value>> {
             .zip(recent_downloads)
             .map(
                 |(((max_version, krate), perfect_match), recent_downloads)| {
+                    let downloads = krate.downloads as i64;
                     EncodableCrate::from_minimal(
                         krate,
                         Some(&max_version),
                         Some(vec![]),
                         perfect_match,
+                        downloads,
                         Some(recent_downloads),
                     )
                 },
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -72,6 +72,7 @@ pub async fn search(app: AppState, req: Parts) -> AppResult<Json<Value>> {
         let selection = (
             ALL_COLUMNS,
             false.into_sql::<Bool>(),
+            crate_downloads::downloads,
             recent_crate_downloads::downloads.nullable(),
             0_f32.into_sql::<Float>(),
         );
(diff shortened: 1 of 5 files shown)


eth3lbert (approved):
LGTM 👍

src/controllers/krate/search.rs
```diff
@@ -551,29 +557,30 @@ mod seek {
...
             Query(bool, i32)
             Relevance(bool, f32, i32)
         }
     }
 
     impl Seek {
-        pub(crate) fn to_payload(&self, record: &(Crate, bool, Option<i64>, f32)) -> SeekPayload {
+        pub(crate) fn to_payload(
+            &self,
+            record: &(Crate, bool, i64, Option<i64>, f32),
```
LawnGnome, line 569: From a pure maintainability perspective I don't really love the 5-tuple. It was sort of almost OK when all the fields had significantly distinct types, but now there's an `i64` and an `Option<i64>` it's starting to feel a little muddy.

I'm not sure this _needs_ to change immediately — presumably we'd have to introduce a new `struct` that implements `From<(Crate, bool, i64, Option<i64>, f32)>` or similar, and that might not be a pure win — but I'm worried that someone (OK, me) is eventually going to forget which numbered field is which.

LawnGnome (approved):
Non-blocking readability concern, but let's ship it and see what happens.

src/controllers/krate/search.rs
Turbo87: yeah, I agree about that. I mentioned something similar in https://github.com/rust-lang/crates.io/pull/8037#discussion_r1499030402.

@eth3lbert is this something you could look into? :)

eth3lbert: Indeed, I think I can modify the macro to add support for braced structs. Then, I would like to know if we want to continue supporting tuple structs.

Turbo87: > I would like to know if we want to continue supporting tuple structs.

I think it would be fine to move everything over to structs with named fields and remove support for tuples

PR #8246, opened 2024-03-05
Added encoding declaration

thesayfulla:
You can read more about [here.](https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations)

### DIFF
--- a/script/infer-database-dump-version.py
+++ b/script/infer-database-dump-version.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# -*- mode:python; coding:utf-8; -*-
 
 import argparse
 import json


Turbo87:
> If no encoding declaration is found, the default encoding is UTF-8.

if the default is already utf-8 then what is the advantage of explicitly specifying it?

thesayfulla:
> > If no encoding declaration is found, the default encoding is UTF-8.
> 
> if the default is already utf-8 then what is the advantage of explicitly specifying it?

Reason is clarity, self-documentation, and future-proofing like this [script.](https://github.com/AlmaLinux/albs-sign-file/blob/main/db_manage.py)

Turbo87:
unless I'm missing something there is only ASCII in this file anyway, so even if Python decided to change the default it seems quite unlikely to break.

I'm not a big fan of unnecessary visual clutter, so unless there is a good technical reason for this I'm not sure we should add this.

LawnGnome:
Thanks, but given that this script would undoubtedly be broken by more than a default encoding change in a future Python major version that changed it, I don't think this is necessary.

thesayfulla:
Thank you for your time and feedback.

PR #8250, opened 2024-03-06
controllers/user/other: Read downloads from `crate_downloads` table

Turbo87:
This is roughly similar to #8244 and switches the `/api/v1/users/:user_id/stats` endpoint over to using the `crate_downloads` database table instead of the `crates.downloads` column.

### DIFF
--- a/src/controllers/user/other.rs
+++ b/src/controllers/user/other.rs
@@ -1,7 +1,8 @@
 use crate::controllers::frontend_prelude::*;
+use bigdecimal::{BigDecimal, ToPrimitive};
 
 use crate::models::{CrateOwner, OwnerKind, User};
-use crate::schema::{crate_owners, crates, users};
+use crate::schema::{crate_downloads, crate_owners, crates, users};
 use crate::sql::lower;
 use crate::views::EncodablePublicUser;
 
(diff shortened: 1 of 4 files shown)


src/controllers/user/other.rs
```diff
@@ -29,11 +30,13 @@ pub async fn stats(state: AppState, Path(user_id): Path<i32>) -> AppResult<Json<
...
+        let data = CrateOwner::by_owner_kind(OwnerKind::User)
             .inner_join(crates::table)
+            .inner_join(crate_downloads::table.on(crates::id.eq(crate_downloads::crate_id)))
             .filter(crate_owners::owner_id.eq(user_id))
-            .select(sum(crates::downloads))
-            .first::<Option<i64>>(conn)?
+            .select(sum(crate_downloads::downloads))
+            .first::<Option<BigDecimal>>(conn)?
+            .map(|d| d.to_u64().unwrap_or(u64::MAX))
             .unwrap_or(0);
```
eth3lbert, line 40: Just leaving information here. Due to the fact that the `sum` function over `bigint` columns in PostgreSQL returns `numeric` (`sum ( bigint ) → numeric`), I believe this is the reason we use `BigDecimal` here instead of `i64`.

eth3lbert (approved):

src/controllers/user/other.rs
Turbo87: yeah, exactly. Diesel has the same behavior built into the typesystem, which is why I had to change it like this

PR #8252, opened 2024-03-06
models/krate_reverse_dependencies.sql: Read downloads from `crate_downloads` table

Turbo87:
This is roughly similar to #8244 and switches the reverse dependencies endpoint over to using the `crate_downloads` database table instead of the `crates.downloads` column.


src/models/krate_reverse_dependencies.sql
eth3lbert, line 1: The query should work without issues, but since we recently changed the schema, it has the potential for further optimization.  I think it would be better to introduce this optimization in a separate PR.

eth3lbert (approved):

PR #8260, opened 2024-03-06
Convert seek to named fields

eth3lbert:
As mention in commits https://github.com/rust-lang/crates.io/pull/8244#discussion_r1513641615 and https://github.com/rust-lang/crates.io/pull/8037#discussion_r1499030402.

### DIFF
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -404,7 +404,7 @@ impl<T, C> PaginatedQueryWithCountSubq<T, C> {
 macro_rules! seek {
     // Field struct
     (@variant_struct $vis:vis $variant:ident {
-        $($(#[$field_meta:meta])? $field:ident: $ty:ty),*
+        $($(#[$field_meta:meta])? $field:ident: $ty:ty),* $(,)?
     }) => {
         paste::item! {
             #[derive(Debug, Default, Deserialize, PartialEq)]
--- a/src/controllers/helpers/pagination.rs
+++ b/src/controllers/helpers/pagination.rs
@@ -608,13 +608,22 @@ mod tests {
 
     mod seek {
         use chrono::naive::serde::ts_microseconds;
-        seek! {
+        seek!(
             pub(super) enum Seek {
-                Id{id: i32},
-                New{#[serde(with="ts_microseconds")] dt: chrono::NaiveDateTime, id: i32},
-                RecentDownloads{ downloads: Option<i64>, id: i32 },
+                Id {
+                    id: i32,
+                },
+                New {
+                    #[serde(with = "ts_microseconds")]
+                    dt: chrono::NaiveDateTime,
+                    id: i32,
+                },
+                RecentDownloads {
+                    downloads: Option<i64>,
+                    id: i32,
+                },
             }
-        }
+        );
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


src/controllers/helpers/pagination.rs
```diff
@@ -580,15 +610,16 @@ mod tests {
         use chrono::naive::serde::ts_microseconds;
         seek! {
             pub(super) enum Seek {
-                Id(i32)
-                New(#[serde(with="ts_microseconds")] chrono::NaiveDateTime, i32)
-                RecentDownloads(Option<i64>, i32)
+                Id{id: i32}
+                New{#[serde(with="ts_microseconds")] dt: chrono::NaiveDateTime, id: i32}
+                RecentDownloads{ downloads: Option<i64>, id: i32 }
```
LawnGnome, line 615: This could use some manual formatting, since `rustfmt` obviously isn't getting in here.

src/controllers/krate/search.rs
```diff
@@ -553,13 +560,13 @@ mod seek {
...
-            Downloads(i64, i32)
-            Query(bool, i32)
-            Relevance(bool, f32, i32)
+            Name{ id: i32 }
+            New{#[serde(with="ts_microseconds")] created_at: chrono::NaiveDateTime, id: i32}
+            RecentUpdates{#[serde(with="ts_microseconds")] updated_at: chrono::NaiveDateTime, id: i32}
+            RecentDownloads{recent_downloads: Option<i64>, id: i32}
+            Downloads{downloads: i64, id: i32}
+            Query{exact_match: bool, id: i32}
+            Relevance{exact_match: bool, rank: f32, id: i32}
```
LawnGnome, line 569: As above, this would be improved with some additional formatting.

src/controllers/krate/search.rs
```diff
@@ -568,20 +575,34 @@ mod seek {
...
+                    id,
+                    updated_at,
+                    created_at,
+                    ..
+                },
+                exact_match,
+                downloads,
+                recent_downloads,
+                rank,
+            ) = *record;
```
LawnGnome, line 589: Some day I hope to be brave enough to write a `let` destructure like this. :laughing:

src/controllers/krate/versions.rs
```diff
@@ -236,16 +236,17 @@ mod seek {
     // doesn't include field names.
     seek! {
         pub enum Seek {
-            Semver(i32)
-            Date(#[serde(with="ts_microseconds")] chrono::NaiveDateTime, i32)
+            Semver{id: i32}
+            Date{#[serde(with="ts_microseconds")] created_at: chrono::NaiveDateTime, id: i32}
```
LawnGnome, line 240: Last macro that could use some extra spaces. :space_invader:

LawnGnome (commented):
I'm definitely not smart enough to review the new implementation of the `seek` macro, but it seems to work just fine, and I definitely prefer the new code that uses it.

:+1: from me as long as we get the formatting fixed up in the actual uses of `seek!`. Nice job!

src/controllers/helpers/pagination.rs
Turbo87: I've managed to figure out how to `rustfmt` can be used for this:

1. we need to have valid Rust code inside the macro invocation, but the implementation was missing trailing commas for the enum variants
2. it needs to be invoked as `seek!(...);` instead of `seek! { ... }` and apparently that enables rustfmt to format the invocation content (see https://users.rust-lang.org/t/rustfmt-skips-macro-arguments/74807/4)

I've added two commits to the PR for the above.

eth3lbert: Thank you for your help. I appreciate it.

PR #8278, opened 2024-03-11
Add `clean_processed_log_files` background job

Turbo87:
Our `processed_log_files` is already at 35k+ rows and growing steadily. We don't want it to grow infinitely though, so this commit implements a new background job that can be run regularly to remove obsolete rows from the table. The cut off date is currently hardcoded to one week, which should be sufficient to prevent duplicate SQS events and keep the table at a reasonable size.

### DIFF
--- a/src/admin/enqueue_job.rs
+++ b/src/admin/enqueue_job.rs
@@ -15,6 +15,7 @@ use secrecy::{ExposeSecret, SecretString};
 )]
 pub enum Command {
     UpdateDownloads,
+    CleanProcessedLogFiles,
     DumpDb {
         #[arg(env = "READ_ONLY_REPLICA_URL")]
         database_url: SecretString,
@@ -58,6 +59,9 @@ pub fn run(command: Command) -> Result<()> {
             } else {
                 jobs::UpdateDownloads.enqueue(conn)?;
             }
+        }
+        Command::CleanProcessedLogFiles => {
+            jobs::CleanProcessedLogFiles.enqueue(conn)?;
         }
         Command::DumpDb {
             database_url,
--- a/src/worker/jobs/downloads/clean_processed_log_files.rs
+++ b/src/worker/jobs/downloads/clean_processed_log_files.rs
@@ -1,0 +1,112 @@
+use crate::schema::processed_log_files;
+use crate::worker::Environment;
+use anyhow::anyhow;
+use crates_io_worker::BackgroundJob;
+use diesel::prelude::*;
+use std::sync::Arc;
+
+/// This job is responsible for cleaning up old entries in the
+/// `processed_log_files` table.
+///
+/// Rows older than one week will be deleted.
+#[derive(Serialize, Deserialize)]
+pub struct CleanProcessedLogFiles;
+
+impl BackgroundJob for CleanProcessedLogFiles {
+    const JOB_NAME: &'static str = "clean_processed_log_files";
+    const QUEUE: &'static str = "downloads";
+
+    type Context = Arc<Environment>;
+
+    async fn run(&self, env: Self::Context) -> anyhow::Result<()> {
+        let conn = env.deadpool.get().await?;
+        conn.interact(run)
+            .await
+            .map_err(|err| anyhow!(err.to_string()))??;
+
+        Ok(())
+    }
+}
+
+fn run(conn: &mut PgConnection) -> QueryResult<()> {
+    let filter = processed_log_files::time.lt(cut_off_date());
+    diesel::delete(processed_log_files::table.filter(filter)).execute(conn)?;
+
+    Ok(())
+}
+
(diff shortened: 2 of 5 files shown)


src/worker/jobs/downloads/clean_log_dedupe_table.rs
```diff
@@ -0,0 +1,109 @@
...
+use crates_io_worker::BackgroundJob;
+use diesel::prelude::*;
+use std::sync::Arc;
+
+/// This job is responsible for cleaning up old entries in the
+/// `processed_log_files` table.
+///
+/// Rows older than one week will be deleted.
+#[derive(Serialize, Deserialize)]
+pub struct CleanLogDedupeTable;
```
LawnGnome, line 13: Naming nit: "dedupe" can also be a (very informal) verb and word order is obviously a nightmare in English, so maybe this would be better if it was named closer to the table, eg `CleanProcessedLogFilesTable` or similar. (We probably don't actually need all those words.)

Or, put another way: it wasn't immediately obvious to me which verb was the actual verb in that identifier.

src/worker/jobs/downloads/clean_log_dedupe_table.rs
```diff
@@ -0,0 +1,109 @@
...
+        conn.interact(run)
+            .await
+            .map_err(|err| anyhow!(err.to_string()))??;
+
+        Ok(())
+    }
+}
+
+fn run(conn: &mut PgConnection) -> QueryResult<()> {
+    let cut_off_date = chrono::Utc::now() - chrono::Duration::try_weeks(1).unwrap();
```
LawnGnome, line 32: `try_weeks` :grimacing:

LawnGnome (approved):
_oh no, there's a bikeshed_

src/worker/jobs/downloads/clean_log_dedupe_table.rs
Turbo87: don't get me started........

eth3lbert: We can also achieve this using diesel:
```rust
use diesel::dsl::{now, IntervalDsl};

let cut_off_date = now.into_sql::<diesel::sql_types::Timestamptz>() - 1.weeks();
```
similar to how it's done in the https://github.com/rust-lang/crates.io/blob/a114476740872411f07e2be746a63fbfe391e8aa/src/controllers/krate/publish.rs#L429-L437

src/worker/jobs/downloads/clean_log_dedupe_table.rs
eth3lbert: +1 to renaming "dedup". It's a bit unclear here. "Outdated," "old," or "processed" might be better.

src/worker/jobs/downloads/clean_log_dedupe_table.rs
```diff
@@ -0,0 +1,109 @@
...
+    use chrono::{DateTime, Utc};
+    use insta::assert_debug_snapshot;
+
+    #[test]
+    fn test_cleanup() {
+        let (_test_db, conn) = &mut test_db_connection();
+
+        let now = chrono::Utc::now();
+        let hours = chrono::Duration::try_hours(1).unwrap();
+
```
eth3lbert, line 53:
```suggestion
        let weeks = chrono::Duration::try_weeks(1).unwrap();

        insert(
            conn,
            vec![
                ("very-old-file", now - weeks * 4),
                ("old-file", now - weeks - hours),
                ("newish-file", now - weeks + hours),
                ("brand-new-file", now),
                ("future-file", now + weeks),
            ],
        );
```
I would suggest performing the arithmetic based on our cutoff value. This would be clearer and easier to follow.

eth3lbert (approved):

src/worker/jobs/downloads/clean_log_dedupe_table.rs
```diff
@@ -0,0 +1,109 @@
...
+        insert(
+            conn,
+            vec![
+                ("very-old-file", now - hours * 30 * 24),
+                ("old-file", now - hours * (7 * 24 + 1)),
+                ("newish-file", now - hours * (6 * 24 + 23)),
+                ("brand-new-file", now),
+                ("future-file", now + hours * 7 * 24),
+            ],
+        );
```
eth3lbert, line 63:
```suggestion
        let weeks = chrono::Duration::try_weeks(1).unwrap();
        insert(
            conn,
            vec![
                ("very-old-file", now - weeks * 4),
                ("old-file", now - weeks - hours),
                ("newish-file", now - weeks + hours),
                ("brand-new-file", now),
                ("future-file", now + weeks),
            ],
        );
```

I would suggest performing the arithmetic based on our cutoff value. This would be clearer and easier to follow.

Issue #8154, opened 2024-02-19
API token expiry warning emails


Turbo87:
Some time last year we introduced optional expiration periods for API tokens. One caveat of the existing system is that users don't get notified when their tokens expire and they have to proactively monitor this if they use expiration.

In our team meeting on Friday we discussed how we could improve the situation and one idea was to automatically send out warning emails when an API token expires.

As https://github.com/rust-lang/crates.io/issues/6664#issuecomment-1952208242 states, such emails are seen as a requirement before we can change the default expiration setting on the API token creation page.

A couple of open questions:

- When should we send these emails? At the time when the token expires? A week before? A day before?
- Should we offer a way to create a new token based on the settings from an existing token?
- Should we use a feature flag while testing this out on the staging environment?

In terms of implementation:

- We will probably need to track for what tokens we have sent out warnings already. One way to do this would be to add another column to the `api_tokens` table.
- A new background job could then be implemented that scans the `api_tokens` table for tokens that have expired (or are going to expire) and where a warning has not been sent yet, and then sends out such a warning email.

hi-rustin:
> * Should we offer a way to create a new token based on the settings from an existing token?

I haven't seen this feature on other products and platforms. Do you have any examples we can refer to?

Are you working on this? If not, do you mind me to take it and help implement it?

mdtro:
I'll follow up with the infra team on Zulip in regards to hardening our DMARC and SPF configurations in preparation for sending these emails. 🙂

hi-rustin:
> I haven't seen this feature on other products and platforms. Do you have any examples we can refer to?

GitHub has this feature, for instance:
<img width="1100" alt="image" src="https://github.com/rust-lang/crates.io/assets/29879298/a71512ea-88e4-4def-a406-14c735aa8f7b">
The UI looks like this:

<img width="860" alt="image" src="https://github.com/rust-lang/crates.io/assets/29879298/72f9213a-243e-425e-b5a4-ee284ff68e66">

Turbo87:
Since the API token expiry notification emails have been implemented I will close this issue now. The token copy functionality has its own issue now: https://github.com/rust-lang/crates.io/issues/8717 :)

Issue #6664, opened 2023-06-23
The "New API token" page defaults to "no expiration"


fasterthanlime:
### Current Behavior

The "New API token" page defaults to "no expiration"

<img width="429" alt="image" src="https://github.com/rust-lang/crates.io/assets/7998310/56ea76ef-87de-4842-b9c2-d9c439a7657a">

### Expected Behavior

I think it should default to, like, "90 days". Or even "365 days" if you want to be conservative.

Defaults matter here. If I'm going to be irresponsible with a never-expiring token, make me do two additional clicks.

### Steps To Reproduce

1. Go to https://crates.io/settings/tokens/new
2. Notice it defaults to no expiration

### Environment

- Browser: Microsoft Edge, for the street cred
- OS: Windows 11, to have sound

### Anything else?

Thanks for your work on this! The new form is definitely a step in the right direction.

Turbo87:
We've chosen this default for now since it matches the previous behavior. Once people are a bit more familiar we will most likely switch it over to one of the presets instead.

zahash:
can i work on this issue? i can change it to 30 days.

Turbo87:
we have not decided yet at what point we want to change the default value. until then it's probably not worth looking into :)

Turbo87:
we talked a bit about this issue in our team meeting last friday. the conclusion was that we would like to have an implementation of token expiry warning emails first before we switch the default expiry time. the reasoning behind this is the situation where a Rust newcomer creates a default token on crates.io to put it into cargo and then 30 days later is surprised that nothing works anymore. we feel like this situation can be mitigated to some degree by having such expiry emails.

PR #8290, opened 2024-03-12
Send API token expiry notification emails

hi-rustin:
ref https://github.com/rust-lang/crates.io/issues/8154

## Description
This PR adds a new background job to send the expiry notification for tokens that will expire within 3 days.

As https://github.com/rust-lang/crates.io/issues/6664#issuecomment-1952208242 states, such emails are seen as a requirement before we can change the default expiration setting on the API token creation page.

Currently, GitHub only sends notifications twice - one after 6 days and another after 12 hours, and after it has expired.

Perhaps we should consider adopting a similar approach, although it may be too aggressive.
So I just picked 3 days as the default value. It should be easy to extend the current design to send multiple emails by reusing 'expiry_notification_at' multiple times.

## Explanation of Changes
1. It added a new field `expiry_notification_at` to the `api_tokens` table.
2. It added a new background job called `SendTokenExpiryNotifications ` that will scan all the expiry tokens within 3 days and send an email to the users.
3. It added an integration test to check we sent the right email.


hi-rustin (commented):

hi-rustin:
Tested locally:
1. Start the crates.io in the docker
```console
crates.io on  rustin-patch-expiry_notification_job [$!] via 🐳 orbstack is 📦 v0.0.0 via  v20.12.0 via 🦀 v1.77.0 
❯ docker ps
CONTAINER ID   IMAGE               COMMAND                  CREATED          STATUS          PORTS                                       NAMES
ea2f555e8f9f   cratesio-frontend   "pnpm start"             40 minutes ago   Up 15 minutes   0.0.0.0:4200->4200/tcp, :::4200->4200/tcp   cratesio-frontend-1
a3f3d61c8730   c944ce643ba1        "cargo run --bin bac…"   40 minutes ago   Up 2 minutes                                                cratesio-worker-1
d0dc3fabda26   cratesio-backend    "/app/docker_entrypo…"   40 minutes ago   Up 14 minutes   0.0.0.0:8888->8888/tcp, :::8888->8888/tcp   cratesio-backend-1
13a8ba48f41e   postgres:13         "docker-entrypoint.s…"   40 minutes ago   Up 40 minutes   127.0.0.1:5432->5432/tcp                    cratesio-postgres-1
```
2. Create a new token which will expire within one day
image

3. Trigger a background job: ` DATABASE_URL=postgresql://postgres:password@localhost:5432/cargo_registry ./target/debug/crates-admin enqueue-job check_about_to_expire_token`
4. Check the email from the worker's tmp dir
```console
Message-ID: <MJOpaJaGzSXbORzJjtAbjXqAcc0STbPw@crates.io>
To: rustin.liu@gmail.com
From: noreply@crates.io
Subject: Your token is about to expire
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: quoted-printable
Date: Fri, 05 Apr 2024 08:26:31 +0000

Hi @hi-rustin,

We noticed your token "test_token" will expire in about 3 days.

If this token is still needed, visit https://crates.io/settings/tokens/new =
to generate a new one.

Thanks,
The crates.io team
```
5. trigger it again: `DATABASE_URL=postgresql://postgres:password@localhost:5432/cargo_registry ./target/debug/crates-admin enqueue-job check_about_to_expire_token`
6. Check no more new emails sent by the worker

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,135 @@
...
+                };
+                let email = ExpiryNotificationEmail {
+                    name: user.gh_login.clone(),
+                    token_name: token.name.clone(),
+                };
+                emails.send(&recipient, email)?;
+                // Also update the token to prevent duplicate notifications.
+                diesel::update(token)
+                    .set(crate::schema::api_tokens::expiry_notification_at.eq(now.nullable()))
+                    .execute(conn)?;
```
eth3lbert, line 54: After a quick look, if I'm reading correctly, this code will fail fast once an email fails to send. This behavior might also block subsequent emails that could potentially be sent successfully.

I believe it would be better to warn or collect these failed emails, skip them, and continue sending the remaining emails.

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,163 @@
...
+    // Batch send notifications in transactions.
+    const BATCH_SIZE: usize = 100;
+    for chunk in expired_tokens.chunks(BATCH_SIZE) {
+        conn.transaction(|conn| {
+            for token in chunk {
+                // Send notification.
+                let user = User::find(conn, token.user_id)?;
+                let Some(recipient) = user.email(conn)? else {
+                    return Err(anyhow!("No address found"));
+                };
```
eth3lbert, line 45: I think it would be more appropriate to just warn/collect the failed tokens and continue processing the remaining pending tokens, for the same reasons mentioned previously.

src/models/token.rs
```diff
@@ -95,6 +97,26 @@ impl ApiToken {
...
+    ) -> QueryResult<Vec<ApiToken>> {
+        use diesel::dsl::{now, IntervalDsl};
+
+        api_tokens::table
+            .filter(api_tokens::revoked.eq(false))
+            .filter(
+                api_tokens::expired_at
+                    .is_not_null()
+                    .and(api_tokens::expired_at.gt(now.nullable())) // Ignore already expired tokens
+                    .and(api_tokens::expired_at.lt(now.nullable() + days_until_expiry.days())),
```
eth3lbert, line 114:
```suggestion
                    .and(api_tokens::expired_at.assume_not_null().gt(now)) // Ignore already expired tokens
                    .and(
                        api_tokens::expired_at
                            .assume_not_null()
                            .lt(now + days_until_expiry.days()),
                    ),

```

Not a blocker, but I think this is more semantic.

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,163 @@
...
+                let email = ExpiryNotificationEmail {
+                    name: user.gh_login.clone(),
+                    token_name: token.name.clone(),
+                };
```
eth3lbert, line 49: We don't need to clone `user.gh_login`.  Additionally, we could get rid of the clone entirely if we have the `ExpiryNotificationEmail` as 
```rust
struct ExpiryNotificationEmail<'a> {
    name: &'a str,
    token_name: &'a str,
}
```

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,163 @@
...
+        let token: ApiToken = diesel::insert_into(api_tokens::table)
+            .values((
+                api_tokens::user_id.eq(user.id),
+                api_tokens::name.eq("test_token"),
+                api_tokens::token.eq(token.hashed()),
+                api_tokens::expired_at.eq(now.nullable() + 2.days()),
+            ))
+            .returning(ApiToken::as_returning())
+            .get_result(&mut conn)?;
+
```
eth3lbert, line 126: It would be helpful if we also checked that non-expired tokens are not affected.

src/worker/jobs/expiry_notification.rs
eth3lbert: though I'm not sure how failed tokens will be notified to the admin or how the admin will be able to perceive them.

eth3lbert:
A couple of comments, but overall this looks good to me! 👍

eth3lbert:
I also noticed that since we don't have an index on `expired_at`, I'm not sure how long`find_tokens_expiring_within_days` will take. We might need to investigate this further in production.

src/models/token.rs
hi-rustin: Updated.

src/worker/jobs/expiry_notification.rs
hi-rustin: Updated. Thanks!

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,163 @@
...
+        // Set up a user and a token that is about to expire.
+        let user = faker.user(&mut conn, "a", Some("testuser@test.com".to_owned()))?;
+        let token = PlainToken::generate();
+
+        let token: ApiToken = diesel::insert_into(api_tokens::table)
+            .values((
+                api_tokens::user_id.eq(user.id),
+                api_tokens::name.eq("test_token"),
+                api_tokens::token.eq(token.hashed()),
+                api_tokens::expired_at.eq(now.nullable() + 2.days()),
```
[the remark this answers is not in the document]
hi-rustin: Make sense. Updated.

src/worker/jobs/expiry_notification.rs
hi-rustin: Good suggestion! Added.

src/worker/jobs/expiry_notification.rs
hi-rustin: I'm not so sure about the likelihood of this happening. As I recall, we can't use the token before verifying the email. I'll check.

migrations/2024-03-06-123525_add_expiry_notification_at/up.sql
```diff
@@ -0,0 +1,4 @@
+-- Introduce a new column `expiry_notification_at` in the `api_tokens` table.
+-- This column will hold the timestamp of when the user was informed about their token's impending expiration.
+ALTER TABLE api_tokens ADD expiry_notification_at TIMESTAMP;
+
```
Turbo87, line 4:
```suggestion
ALTER TABLE api_tokens ADD expiry_notification_at TIMESTAMP;

COMMENT ON COLUMN api_tokens.expiry_notification_at IS 'timestamp of when the user was informed about their token's impending expiration';
```

src/models/token.rs
```diff
@@ -95,6 +97,30 @@ impl ApiToken {
         .or_else(|_| tokens.select(ApiToken::as_select()).first(conn))
         .map_err(Into::into)
     }
+
+    /// Find all tokens that are not revoked and will expire within the specified number of days.
+    pub fn find_tokens_expiring_within_days(
+        conn: &mut PgConnection,
+        days_until_expiry: i64,
```
Turbo87, line 104: I think it might be better to accept a `chrono::TimeDelta` (aka. `chrono::Duration`) here to have a more expressive API.

src/models/token.rs
```diff
@@ -95,6 +97,30 @@ impl ApiToken {
...
+                api_tokens::expired_at
+                    .is_not_null()
+                    .and(api_tokens::expired_at.assume_not_null().gt(now)) // Ignore already expired tokens
+                    .and(
+                        api_tokens::expired_at
+                            .assume_not_null()
+                            .lt(now + days_until_expiry.days()),
+                    ),
+            )
+            .filter(api_tokens::expiry_notification_at.is_null())
```
Turbo87, line 120: hmm, given the name of the function I wouldn't expect this additional filter to be here. I understand why it's there for this specific purpose, but it makes the function less reusable. I'm wondering if we should move the function closer to the code that calls it, instead of putting it on the `ApiToken` impl block. 🤔

src/tests/routes/me/tokens/snapshots/all__routes__me__tokens__create__create_token_success.snap
```diff
@@ -1,13 +1,14 @@
 ---
 source: src/tests/routes/me/tokens/create.rs
-expression: response.into_json()
+expression: response.json()
```
Turbo87, line 3: any idea where these changes are coming from? I don't see these expressions change in the diff here 🤔

src/tests/routes/me/tokens/snapshots/all__routes__me__tokens__create__create_token_success.snap
```diff
@@ -1,13 +1,14 @@
...
 ---
 {
   "api_token": {
     "crate_scopes": null,
     "created_at": "[datetime]",
     "endpoint_scopes": null,
     "expired_at": null,
+    "expiry_notification_at": null,
```
Turbo87, line 11: do we want to expose this field on the public API? we can always add it later, but removing it would be tricky once we've first exposed it

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,188 @@
...
+        )
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        models::token::ApiToken, schema::api_tokens, test_util::test_db_connection,
+        typosquat::test_util::Faker, util::token::PlainToken,
```
Turbo87, line 101: I'm not sure we should rely on typosquat-specific test utilities in other parts of the codebase, especially if they require us to change the existing test utilities.

since we're only using `faker.user()` it would probably be easier to use `NewUser::new()` directly and keep the existing test utility as-is, which would also reduce the amount of unrelated changes in the PR diff.

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,139 @@
...
+use anyhow::anyhow;
+use crates_io_worker::BackgroundJob;
+use diesel::{
+    dsl::now, Connection, ExpressionMethods, NullableExpressionMethods, PgConnection, RunQueryDsl,
+};
+
+use crate::{email::Email, models::User, worker::Environment, Emails};
+
+/// The threshold in days for the expiry notification.
+const EXPIRY_THRESHOLD: i64 = 3;
```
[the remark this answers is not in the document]
Turbo87: similar to the comment above, this should probably be a `chrono::TimeDelta` (or something similar)

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,188 @@
...
+pub struct CheckAboutToExpireToken;
+
+impl BackgroundJob for CheckAboutToExpireToken {
+    const JOB_NAME: &'static str = "expiry_notification";
+
+    type Context = Arc<Environment>;
+
+    #[instrument(skip(env), err)]
+    async fn run(&self, env: Self::Context) -> anyhow::Result<()> {
+        let mut conn = env.connection_pool.get()?;
```
Turbo87, line 25: this will probably need a rebase now that #8385 is merged :)

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,188 @@
...
+        // Check if the token is about to expire
+        // If the token is about to expire, trigger a notification.
+        check(&env.emails, &mut conn)
+    }
+}
+// Check if the token is about to expire and send a notification if it is.
+fn check(emails: &Emails, conn: &mut PgConnection) -> anyhow::Result<()> {
+    info!("Checking if tokens are about to expire");
+    let expired_tokens =
+        crate::models::token::ApiToken::find_tokens_expiring_within_days(conn, EXPIRY_THRESHOLD)?;
```
Turbo87, line 35: any reason for not importing `ApiToken` at the top of the file?

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,188 @@
...
+    // Batch send notifications in transactions.
+    const BATCH_SIZE: usize = 100;
+    for chunk in expired_tokens.chunks(BATCH_SIZE) {
```
Turbo87, line 38: why are we chunking this? and if we decide to chunk it, why not chunk it at the database level?

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,188 @@
...
+    name: &'a str,
+    token_name: &'a str,
+}
+
+impl<'a> Email for ExpiryNotificationEmail<'a> {
+    const SUBJECT: &'static str = "Your token is about to expire";
+
+    fn body(&self) -> String {
+        format!(
+            r#"Hi @{},
```
Turbo87, line 83: I don't think we're using the @-prefix in any other emails

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,188 @@
...
+
+We noticed your token "{}" will expire in about {} days.
```
Turbo87, line 85: we might want to serialize the exact date instead of "in about 3 days". maybe something like:

> Your "foo" crates.io API token will expiry on 2024-04-15.

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,188 @@
...
+use std::sync::Arc;
+
+/// The threshold in days for the expiry notification.
+const EXPIRY_THRESHOLD: i64 = 3;
+
+/// A job responsible for monitoring the status of a token.
+/// It checks if the token is about to reach its expiry date.
+/// If the token is about to expire, the job triggers a notification.
+#[derive(Default, Serialize, Deserialize, Debug)]
+pub struct CheckAboutToExpireToken;
```
Turbo87, line 16: nitpick: what do you think about something like `SendTokenExpiryNotifications`?

src/worker/jobs/expiry_notification.rs
hi-rustin: I checked it. Users can generate tokens even users haven't verified their emails.

migrations/2024-03-06-123525_add_expiry_notification_at/up.sql
hi-rustin: Updated.

src/worker/jobs/expiry_notification.rs
hi-rustin: Updated.

src/worker/jobs/expiry_notification.rs
hi-rustin: Updated. Good point!

src/worker/jobs/expiry_notification.rs
hi-rustin: This is imported by the IDE. Updated.

migrations/2024-03-06-123525_add_expiry_notification_at/up.sql
Turbo87: looks like I missed an `'` in the comment which caused a syntax error 🙈

src/models/token.rs
hi-rustin: I refactored it in [`53f70bf` (#8290)](https://github.com/rust-lang/crates.io/pull/8290/commits/53f70bf814fe606526d4ca14c743f9efc04801ac)

But I'm not sure I used the right API. I used `now + days_until_expiry.num_days().day()` to check it. Do we have a better way to construct it?

src/models/token.rs
hi-rustin: I have moved it out in [53f70bf (#8290)](https://github.com/rust-lang/crates.io/pull/8290/commits/53f70bf814fe606526d4ca14c743f9efc04801ac).

src/tests/routes/me/tokens/snapshots/all__routes__me__tokens__create__create_token_success.snap
hi-rustin: This change was generated by the cargo-insta.

src/worker/jobs/expiry_notification.rs
hi-rustin: I am more concerned here about having too many tokens that need to be emailed, resulting in a large transaction. I may be overly worried?

src/worker/jobs/expiry_notification.rs
hi-rustin: If we use the date directly, should we consider the time zone issue? Or do we just say it is UTC time?

src/tests/routes/me/tokens/snapshots/all__routes__me__tokens__create__create_token_success.snap
hi-rustin: My insta version:
```console
❯ cargo insta --version
cargo-insta 1.37.0
```

src/tests/routes/me/tokens/snapshots/all__routes__me__tokens__create__create_token_success.snap
hi-rustin: Sounds reasonable to me. We can make it private first. [`cc04749` (#8290)](https://github.com/rust-lang/crates.io/pull/8290/commits/cc047491b36a5a96dfbc25c8c90c9f00d2d2b59e)

hi-rustin (commented):
I know that some commits can be squashed. I will re-organize all commits after the review. Using a separate commit to address the comment might help the reviewer understand what changed.

src/tests/routes/me/tokens/snapshots/all__routes__me__tokens__create__create_token_success.snap
hi-rustin: All reverted.

src/models/token.rs
eth3lbert: Since we're using `chrono::TimeDelta`, it might be more reasonable to calculate the actual time directly on the app side. This approach is similar to what we do in https://github.com/rust-lang/crates.io/blob/63c30c6cca0cfca1d3ad9be4816a43d3363c8b55/src/worker/jobs/downloads/clean_processed_log_files.rs#L31-L40.

src/worker/jobs/expiry_notification.rs
eth3lbert: I think what @Turbo87 means is why not just select data in batches (`select ... limit BATCH_SIZE`) instead of selecting all the data at once and then chunking it?

src/worker/jobs/expiry_notification.rs
eth3lbert: So, in this case, is it better to log the error and continue processing the next one?

What do you think, @Turbo87 ?

src/worker/jobs/expiry_notification.rs
Turbo87: good point. I guess since we're a developer-focused service it's probably fine to use UTC and the full ISO8601 format (2024-04-15T12:34:56Z).

src/worker/jobs/expiry_notification.rs
hi-rustin: > if e.g. we process 10 tokens and the 7th update fails, we have already sent out 6 emails, but since the transaction will be rolled back we wouldn't have saved that in the database

In the current implementation, we just print a log if the send is failed. We will try it in the next round.

hi-rustin: > why not just select data in batches (`select ... limit BATCH_SIZE`) instead of selecting all the data at once and then chunking it?

Is it possible that after an email fails to send, we will immediately retry it again in the next loop? (This relies on the database implementation of limit) Would this cause us to try it many times in a short period of time.
Or do you mean we order the data and use limit + offset to control this?

src/models/token.rs
hi-rustin: I tried to follow this approach. But I didn't make it compile. I will have another try. Thanks!

src/worker/jobs/expiry_notification.rs
eth3lbert: Yeah, if we need to run in batch mode, then avoiding errors that cause rollbacks would be more reasonable.

On the other hand, if we process emails one at a time, then we need a mechanism to prevent getting stuck selecting the failed email, which would block other unhandled emails.

src/worker/jobs/expiry_notification.rs
eth3lbert: Or even be explicit with the offset `+00:00`?

src/worker/jobs/expiry_notification.rs
eth3lbert: > Is it possible that after an email fails to send, we will immediately retry it again in the next loop? (This relies on the database implementation of limit) Would this cause us to try it many times in a short period of time. Or do you mean we order the data and use limit + offset to control this?

I think this part is similar to how we do seek(cursor)-based pagination. We just need to use `ORDER BY` to ensure a deterministic order and seek columns for offset.
e.g., `SELECT ... WHERE id > last_processed_id ORDER BY id LIMIT batch_size`.
P.S. There's no real need to order by ID, we could also order by expiry time if that makes more sense.

src/models/token.rs
hi-rustin: @eth3lbert Does this make sense? We cannot directly use `chrono::DateTime<chrono::Utc>` here. Because expired_at is a Timestamp, not a Timestamptz.

```rust
    let expire_data = chrono::Utc::now() + days_until_expiry;
    api_tokens::table
        .filter(api_tokens::revoked.eq(false))
        .filter(
            api_tokens::expired_at
                .is_not_null()
                .and(api_tokens::expired_at.assume_not_null().gt(now)) // Ignore already expired tokens
                .and(
                    api_tokens::expired_at
                        .assume_not_null()
                        .lt(expire_data.naive_utc()),
                ),
        )
        .filter(api_tokens::expiry_notification_at.is_null())
        .select(ApiToken::as_select())
        .get_results(conn)
```

And I found we use my current approach in our codebase https://github.com/rust-lang/crates.io/blob/81023270df96b4d0f6cf78d0b41b4a7060737d74/src/controllers/token.rs#L49

hi-rustin: So I am not sure which one is better. It seems the current one is easier to understand and we don't need to call `naive_utc`.

eth3lbert: > Because expired_at is a Timestamp

Yeah, you need to use `NaiveDateTime`.
Ref: https://github.com/diesel-rs/diesel/blob/974814c6792c534736498a0bd7abf95e21f8bf7b/diesel/src/pg/types/date_and_time/chrono.rs#L21-L48

```rust
        let now = chrono::Utc::now();
        let expired_at = now + chrono::TimeDelta::try_days(days_until_expiry).unwrap();

        api_tokens::table
            .filter(api_tokens::revoked.eq(false))
            .filter(
                api_tokens::expired_at
                    .is_not_null()
                    .and(api_tokens::expired_at.assume_not_null().gt(now.naive_utc())) // Ignore already expired tokens
                    .and(
                        api_tokens::expired_at
                            .assume_not_null()
                            .lt(expired_at.naive_utc()),
                    ),
            )
            .filter(api_tokens::expiry_notification_at.is_null())
            .select(ApiToken::as_select())
            .get_results(conn)
```

eth3lbert: > So I am not sure which one is better. It seems the current one is easier to understand and we don't need to call `naive_utc`.

Either approach is fine with me. I actually left a similar comment before https://github.com/rust-lang/crates.io/pull/8278#discussion_r1520278505. 

One reason to use the app-side timestamp might be that readers wouldn't need to know how to use PG `intervals` to work with the data.

hi-rustin: > Yeah, you need to use `NaiveDateTime`.

Yes. I already did that. I think it makes the code more complex. Seems `now + days_until_expiry.num_days().day()` is easier to understand. I am not sure how PG does the query plan optimization for different approaches. Usually, it shouldn't be a big deal, and also to evaluate the value of this simple expression won't be a big deal. 

So I prefer to keep `now + days_until_expiry.num_days().day()` to keep it simple. 

But if it has a huge performance difference, then maybe we should change it. 

Thanks!

src/worker/jobs/expiry_notification.rs
hi-rustin: > Yeah, if we need to run in batch mode, then avoiding errors that cause rollbacks would be more reasonable.

Expect the email not found case. We already did it.

hi-rustin: > On the other hand, if we process emails one at a time, then we need a mechanism to prevent getting stuck selecting the failed email, which would block other unhandled emails.

Yes, I think sending them one by one makes it more complex. It's difficult to avoid we stuck on the same one for a while.

hi-rustin: @eth3lbert @Turbo87 Based on the comments mentioned above, how about we continue using the current approach and log any errors for emails that failed or were not found? Afterward, we can proceed to send all other emails that can be sent. When we trigger the background job again, we can attempt to send the failed emails again. Does this sound reasonable to you?

Using batch + offset or sending them one by one seems all require more effort to avoid retrying the same email in a short period of time. I am also OK with the `batch + offset` solution. I'm not sure how many expired tokens are in crates.io right now. Maybe I'm overthinking and complicating the issue. 😆

src/worker/jobs/expiry_notification.rs
hi-rustin: Sounds good to me. I will update it.

src/worker/jobs/expiry_notification.rs
eth3lbert: > Yes, I think sending them one by one makes it more complex. It's difficult to avoid we stuck on the same one for a while.

The following is a simplified version for sending them one by one.

```rust
fn get_token(
    last_id: Option<i32>,
    conn: &mut PgConnection,
) -> diesel::QueryResult<crate::models::token::ApiToken> {
    use crate::schema::api_tokens;
    use diesel::prelude::*;

    let now = chrono::Utc::now();
    let expired_at = now + chrono::TimeDelta::try_days(EXPIRY_THRESHOLD).unwrap();

    let mut query = api_tokens::table
        .filter(api_tokens::revoked.eq(false))
        .filter(
            api_tokens::expired_at
                .is_not_null()
                .and(api_tokens::expired_at.assume_not_null().gt(now.naive_utc())) // Ignore already expired tokens
                .and(
                    api_tokens::expired_at
                        .assume_not_null()
                        .lt(expired_at.naive_utc()),
                ),
        )
        .into_boxed();

    if let Some(id) = last_id {
        query = query.filter(api_tokens::id.gt(id));
    }

    query
        .order_by(crate::schema::api_tokens::id.asc())
        .select(crate::models::token::ApiToken::as_select())
        .filter(api_tokens::expiry_notification_at.is_null())
        .first::<crate::models::token::ApiToken>(conn)
}

// Check if the token is about to expire and send a notification if it is.
fn check(emails: &Emails, conn: &mut PgConnection) -> anyhow::Result<()> {
    info!("Checking if tokens are about to expire");

    let mut last_id: Option<i32> = None;

    // You might want to consider changing the following loop to a while let loop
    loop {
        let Ok(token) = get_token(last_id, conn) else {
            break;
        };
        last_id = Some(token.id);

        conn.transaction(|conn| {
            let user = User::find(conn, token.user_id)?;
            let Some(recipient) = user.email(conn)? else {
                error!("No address found for token {}", token.id);
                return Ok::<_, anyhow::Error>(());
            };
            let email = ExpiryNotificationEmail {
                name: &user.gh_login,
                token_name: &token.name,
            };
            match emails.send(&recipient, email) {
                Ok(_) => {
                    // Update the token to prevent duplicate notifications.
                    diesel::update(&token)
                        .set(
                            crate::schema::api_tokens::expiry_notification_at
                                .eq(chrono::Utc::now().naive_utc()),
                        )
                        .execute(conn)?;
                }
                Err(e) => {
                    error!("Failed to send email: {:?} to {}", e, recipient);
                }
            }
            Ok(())
        })?;
    }

    Ok(())
}
```

Feel free to modify it for better readability or clarity.

src/schema.rs
```diff
@@ -81,6 +81,12 @@ diesel::table! {
         ///
         /// (Automatically generated by Diesel.)
         expired_at -> Nullable<Timestamp>,
+        /// The `expiry_notification_at` column of the `api_tokens` table.
+        ///
+        /// Its SQL type is `Nullable<Timestamp>`.
+        ///
+        /// (Automatically generated by Diesel.)
```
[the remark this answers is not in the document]
hi-rustin: Good catch!

src/worker/jobs/expiry_notification.rs
Turbo87: I thought a bit more about this and I think the simplest implementation would be to get one big chunk from the database (`LIMIT 10000` or something like that) and then process them one by one without chunking. That would allow us to get rid of the outer `for chunk in chunks` loop and the `conn.transaction()` call since there is only a single database call happening inside each loop iteration. we could potentially add a warning message to the logs if we actually receive the 10000 tokens in any run, but with the current volume such a limit will probably be sufficient.

hi-rustin (commented):

I know that some commits can be squashed. I will re-organize all commits after the review. Using a separate commit to address the comment might help the reviewer understand what changed.

src/worker/jobs/expiry_notification.rs
hi-rustin: > I thought a bit more about this and I think the simplest implementation would be to get one big chunk from the database (`LIMIT 10000` or something like that) and then process them one by one without chunking.

I have followed this approach and updated the PR.

src/worker/jobs/expiry_notification.rs
hi-rustin: Better name!

src/worker/jobs/expiry_notification.rs
```diff
@@ -41,31 +45,31 @@ impl BackgroundJob for CheckAboutToExpireToken {
 fn check(emails: &Emails, conn: &mut PgConnection) -> anyhow::Result<()> {
     info!("Checking if tokens are about to expire");
     let expired_tokens = find_tokens_expiring_within_days(conn, EXPIRY_THRESHOLD)?;
-    // Batch send notifications in transactions.
-    const BATCH_SIZE: usize = 100;
-    for chunk in expired_tokens.chunks(BATCH_SIZE) {
+    if expired_tokens.len() == MAX_ROWS as usize {
+        warn!("The number of tokens to check exceeds the limit");
+    }
```
eth3lbert, line 50: I'm not sure if warning about `expired_tokens.len() == MAX_ROWS` **exceeding the limit** is a good idea. Since it's an equality check. A better approach might be to compare with `expired_tokens.len() > MAX_ROWS`. We could then query for `MAX_ROWS + 1` elements and process only the first `MAX_ROWS`.

Turbo87: hmm, IMHO it's fine as is. though we could potentially rephrase the warning message a bit to make it clearer.

```suggestion
    if expired_tokens.len() == MAX_ROWS as usize {
        warn!("The max. number of API tokens per query has been reached. More API tokens might be processed on the next run.");
    }
```

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+// Check if the token is about to expire and send a notification if it is.
+fn check(emails: &Emails, conn: &mut PgConnection) -> anyhow::Result<()> {
+    info!("Checking if tokens are about to expire");
+    let expired_tokens = find_tokens_expiring_within_days(conn, EXPIRY_THRESHOLD)?;
+    if expired_tokens.len() == MAX_ROWS as usize {
+        warn!("The number of tokens to check exceeds the limit");
+    }
+    for token in &expired_tokens {
+        conn.transaction(|conn| {
```
Turbo87, line 52: we can probably get rid of the transaction block now since there is only a single write in the transaction and no following reads. might make sense to extract a `fn handle_expiring_token(emails, conn, token) { ... }` for that purpose so that we can use `?` inside the fn and only have to write the error handling code once.

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        // Check that the token is not about to expire.
+        let tokens = api_tokens::table
+            .filter(api_tokens::revoked.eq(false))
+            .filter(api_tokens::expiry_notification_at.is_null())
+            .select(ApiToken::as_select())
+            .load::<ApiToken>(&mut conn)?;
+        assert_eq!(tokens.len(), 3);
+
+        assert_eq!(tokens.len(), 3);
```
Turbo87, line 206:
```suggestion
```

duplicate?

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,229 @@
...
+                        .set(api_tokens::expiry_notification_at.eq(now.nullable()))
+                        .execute(conn)?;
+                }
+                Err(e) => {
+                    error!(?e, ?recipient, "Failed to send notification");
+                    return Err(anyhow!("Failed to send notification: {}", e));
+                }
+            }
+            Ok::<_, anyhow::Error>(())
+        })?;
```
Turbo87, line 76: if we use `?` here then the loop will stop on the first error, but I guess we would probably want to continue processing the remaining tokens?

src/worker/jobs/expiry_notification.rs
hi-rustin: Yes, I agree that we need to improve the warning message. `Exceeds` is quite confusing.

src/worker/jobs/expiry_notification.rs
hi-rustin: Good catch!

hi-rustin (commented):

hi-rustin:
Tested locally:

Check the steps at: https://github.com/rust-lang/crates.io/pull/8290#issuecomment-2039243197

```eml
root@e044d8a84f05:/tmp# tail bab14f97-cf75-4b66-a90d-568fca969b2c.eml 

Hi hi-rustin,

We noticed your token "Test" will expire on 2024-05-08T13:42:47Z.

If this token is still needed, visit https://crates.io/settings/tokens/new =
to generate a new one.

Thanks,
The crates.io team
```

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,231 @@
...
+    info!("Checking if tokens are about to expire");
+    let expired_tokens = find_tokens_expiring_within_days(conn, EXPIRY_THRESHOLD)?;
+    if expired_tokens.len() == MAX_ROWS as usize {
+        warn!("The maximum number of API tokens per query has been reached. More API tokens might be processed on the next run.");
+    }
+    for token in &expired_tokens {
+        if let Err(e) = handle_expiring_token(conn, token, emails) {
+            error!(?e, "Failed to handle expiring token");
+        }
+    }
```
eth3lbert, line 57: There is a possibility that expired tokens are not being processed correctly. This might be due to failed tokens reaching the `MAX_ROWS` limit and being repeatedly selected, which could prevent tokens after that limit from being processed. While this should be a rare case, it's something to be aware of.

hi-rustin: Yes, it is possible. But it should be very rare.

eth3lbert: Yeah, considering the sufficiently large `MAX_ROW` value, this should be a rare case. Additionally, if these failed tokens can be identified and recovered somehow (either before processing the job or afterwards), then there shouldn't be any problems. Therefore, I don't feel strongly about requesting a change, but I just wanted to point out this potential issue.

hi-rustin: To fully solve this problem, we can store retry information in our database. Once we reach a certain threshold, we can ignore those tokens. But we should be able to improve it in the future if we do run into this problem.

src/worker/jobs/expiry_notification.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+impl<'a> Email for ExpiryNotificationEmail<'a> {
+    const SUBJECT: &'static str = "Your token is about to expire";
+
+    fn body(&self) -> String {
+        format!(
+            r#"Hi {},
+
+We noticed your token "{}" will expire on {}.
```
Turbo87, line 123: one thing I just noticed: I don't think we are restricting the token names in any way. technically someone could put e.g. HTML in the token name resulting in a potential security risk when we put it directly in the email. we are using `text/plain` though, so unless some email client ignores that content type it would probably be fine.

since the email will only be sent to the person that created the token this shouldn't be an issue in practice anyway, but it's something to look out for in other generated emails :)

Turbo87 (approved):
Looks great! I've rebased the branch and added a couple of small commits on top. Thanks for working on this!

PR #8294, opened 2024-03-13
Remove `CDN_LOG_COUNTING_ENABLED` feature flag and related code

Turbo87:
The download counting via CDN log files has been running in production for a couple of weeks now without major issues. This PR permanently enables the feature flag and removes all code that was previously use for counting downloads on the API servers.


LawnGnome (approved):
image

_nice_

PR #8385, opened 2024-03-29
worker: Replace `r2d2` with `deadpool`

Turbo87:
This PR ports all of our background worker jobs to use the async database connection pool (based on the `deadpool` crate) instead of `r2d2` (which does not support `diesel-async`).

This was initially causing failing and flaky tests, but those issues got resolved by:

- https://github.com/rust-lang/crates.io/pull/8382
- and https://github.com/rust-lang/crates.io/pull/8381

### DIFF
--- a/src/worker/jobs/daily_db_maintenance.rs
+++ b/src/worker/jobs/daily_db_maintenance.rs
@@ -1,5 +1,5 @@
-use crate::tasks::spawn_blocking;
 use crate::worker::Environment;
+use anyhow::anyhow;
 use crates_io_worker::BackgroundJob;
 use diesel::{sql_query, RunQueryDsl};
 use std::sync::Arc;
@@ -22,14 +22,14 @@ impl BackgroundJob for DailyDbMaintenance {
     /// archive daily download counts and drop historical data, we can drop this task and rely on
     /// auto-vacuum again.
     async fn run(&self, env: Self::Context) -> anyhow::Result<()> {
-        spawn_blocking(move || {
-            let mut conn = env.connection_pool.get()?;
-
+        let conn = env.deadpool.get().await?;
+        conn.interact(move |conn| {
             info!("Running VACUUM on version_downloads table");
-            sql_query("VACUUM version_downloads;").execute(&mut *conn)?;
+            sql_query("VACUUM version_downloads;").execute(conn)?;
             info!("Finished running VACUUM on version_downloads table");
             Ok(())
         })
         .await
+        .map_err(|err| anyhow!(err.to_string()))?
     }
 }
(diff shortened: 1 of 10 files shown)


src/worker/jobs/daily_db_maintenance.rs
```diff
@@ -22,14 +22,14 @@ impl BackgroundJob for DailyDbMaintenance {
...
+        let conn = env.deadpool.get().await?;
+        conn.interact(move |conn| {
             info!("Running VACUUM on version_downloads table");
-            sql_query("VACUUM version_downloads;").execute(&mut *conn)?;
+            sql_query("VACUUM version_downloads;").execute(conn)?;
             info!("Finished running VACUUM on version_downloads table");
             Ok(())
         })
         .await
+        .map_err(|err| anyhow!(err.to_string()))?
```
LawnGnome, line 33: It's a little unfortunate we have to launder this through `Display` and lose any semantic information on the error, but looking at `InteractError`, I don't see any obvious way around that.

src/worker/jobs/downloads/process_log.rs
```diff
@@ -343,16 +343,16 @@ impl Debug for NameAndVersion {
 
 /// Checks if the given log file has already been processed.
 ///
-/// Calls [`spawn_blocking()`] and acquires a connection from the pool before
+/// Acquires a connection from the pool before
 /// passing it to the [`already_processed_inner()`] function.
```
LawnGnome, line 347: Rewrap:

```suggestion
/// Acquires a connection from the pool before passing it to the [`already_processed_inner()`]
/// function.
```

src/worker/jobs/downloads/process_log.rs
```diff
@@ -343,16 +343,16 @@ impl Debug for NameAndVersion {
...
-    let already_processed = spawn_blocking(move || {
-        let mut conn = db_pool.get()?;
-        Ok::<_, anyhow::Error>(already_processed_inner(path, &mut conn)?)
-    })
-    .await?;
+    let conn = db_pool.get().await?;
+    let already_processed = conn
+        .interact(move |conn| already_processed_inner(path, conn))
+        .await
+        .map_err(|err| anyhow!(err.to_string()))??;
```
LawnGnome, line 355: Is it worth us providing a trait to wrap this pattern, since it's coming up in every job?

LawnGnome (approved):
LGTM.

A couple of non-blocking notes, but nothing substantive.

src/worker/jobs/daily_db_maintenance.rs
Turbo87: yeah, it's unfortunate. I hope this will resolve itself once we move to `diesel-async` though, since we no longer need the `interact()` calls then

src/worker/jobs/downloads/process_log.rs
Turbo87: you mean something like `.interact_anyhow(...)`?

src/worker/jobs/downloads/process_log.rs
Turbo87: yeah, I mainly kept it like this to reduce the diff size :D

PR #8388, opened 2024-03-30
Add bioinformatics category and related subcategories

AliSajid:
Hi,

I have been working on quite a few bioinformatics related rust packages and thought it would be great to have a bioinformatics category and related subcategories. I've added those to the file here.

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -517,6 +517,30 @@ description = """
 Crates for research tools and processing of data related to the brain and nervous system.\
 """
 
+[science.categories.bioinformatics]
+name = "Bioinformatics"
+description = """
+Crates for processing large-scale biological data.\
+"""
+
+[bioinformatics.categories.genomics]
+name = "Genomics"
+description = """
+Crates for processing genetic data, including sequences, abdundance, variants and analysis.\
+"""
+
+[bioinformatics.categories.proteomics]
+name = "Proteomics"
+description = """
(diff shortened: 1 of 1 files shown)


zaeleus:
I'd be in favor of a biology category or subcategory.

Some related notes:

  1. The `science` category does already broadly include biology-related crates.

     https://github.com/rust-lang/crates.io/blob/c03b893bf63afd0d98626ca8fcb0638eafe9f55a/src/boot/categories.toml#L495-L500

  2. lib.rs currently categorizes "biology, bioinformatics, genetics" under `science::bio`: <https://lib.rs/science/bio>. See also <https://gitlab.com/lib.rs/main/-/blob/c4fc5e1e61b0c26f9034e7b36c254609eee6a94c/categories/src/categories.toml#L832-840>. This is more generalized than what is proposed here, and I'm not sure how deep crates.io category levels are allowed to be.

AliSajid:
It does make sense to a `science:biology:*` subcategories if that's what lib.rs already does. though I would argue Bioinformatics is distinct from Biology and could be nested directly under `science:`. Shall I make the change?

jtgeibel:
@AliSajid yes, lets move these under `science::*`.

Also, a heads up that there is currently a proposed RFC (rust-lang/rfcs#3488) to deprecate categories and focus our efforts on improving keywords. However, the team has not made a decision on this RFC yet and we see no reason avoid using/adding categories until that decision is made.

PR #8419, opened 2024-04-09
Implement rate limiting for e-mail verifications

LawnGnome:
This applies a burst of 3 and a refill time of 30 minutes by default per user. (Not that I can imagine a scenario where we'd ever override this for a user, but using the same machinery as other user actions is obviously much simpler.)

I don't love that this ends up essentially prop-drilling the rate limiter into a bunch of new places, but I don't see an alternative other than prop-drilling the whole app struct through, which would be worse.

This doesn't address (sorry) per-address rate limiting, but is at least a reasonable starting point.

### DIFF
--- a/src/controllers/user/me.rs
+++ b/src/controllers/user/me.rs
@@ -1,4 +1,6 @@
 use crate::auth::AuthCheck;
+use crate::rate_limiter::LimitedAction;
+use crate::rate_limiter::RateLimiter;
 use secrecy::{ExposeSecret, SecretString};
 use std::collections::HashMap;
 
@@ -159,11 +161,13 @@ pub async fn update_user(
             // an invalid email set in their GitHub profile, and we should let them sign in even though
             // we're trying to silently use their invalid address during signup and can't send them an
             // email. They'll then have to provide a valid email address.
-            let email = UserConfirmEmail {
-                user_name: &user.gh_login,
-                domain: &state.emails.domain,
+            let email = UserConfirmEmail::new(
+                &state.rate_limiter,
+                conn,
+                user,
+                &state.emails.domain,
                 token,
-            };
+            )?;
 
             let _ = state.emails.send(user_email, email);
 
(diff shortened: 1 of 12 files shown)


Turbo87:
> I don't love that this ends up essentially prop-drilling the rate limiter into a bunch of new places, but I don't see an alternative other than prop-drilling the whole app struct through, which would be worse.

I think the root cause of this is that the model layer is currently sending emails, which IMHO is a bit questionable, especially because we're using this code path in a couple of places in the test suite too. It might be better to decouple these things first before introducing the rate limiting.

LawnGnome:
> I think the root cause of this is that the model layer is currently sending emails, which IMHO is a bit questionable, especially because we're using this code path in a couple of places in the test suite too. It might be better to decouple these things first before introducing the rate limiting.

Yeah, I went back and forth on this yesterday — decoupling the e-mail sending out of the model layer is better from an architectural perspective, but it means the `NewUser::create_or_update` API is easier to misuse as it stands, since it now comes with an implicit need to figure out if the e-mail address needs verification after and action that.

The e-mail upsert code could be extracted into something that returns that state, but being in the same transaction as the user upsert is obviously a nice quality that would be bad to lose.

I'm sorta wondering if we end up with something like this (treat names as placeholders):

```rust
let new_user = NewUser::new(...);
let updater = new_user.create_or_update(&conn)?;  // returns some intermediate type holding a transaction
let verification_required = updater.set_email(&email)?;  // needs to be optional since not all updates might touch the e-mail?
updater.commit()?;  // consumes updater

// Has to be after the commit;
if verification_required {
  // send verification e-mail
}
```

Another, simpler option would obviously be to return something else from `create_or_update` than just the `User` — a struct enclosing the `User` along with a `bool` field indicating if verification is required.

Now I've typed this out, I probably have a slight preference for the second option (return something enclosing `User` and whatever other state we need), but I'm interested in what you think.

PR #8424, opened 2024-04-09
crates_io_worker: Replace `r2d2` with `deadpool`

Turbo87:
While #8385 ported the individual background worker jobs from `r2d2` to `deadpool`, this PR ports the background worker system itself to it.

Note that the default `async_pool_size` for the test suite had to be increased because the background worker system now requires us to have one more connection available.

### DIFF
--- a/crates/crates_io_worker/src/util.rs
+++ b/crates/crates_io_worker/src/util.rs
@@ -3,23 +3,6 @@ use sentry_core::Hub;
 use std::any::Any;
 use std::future::Future;
 use std::panic::PanicInfo;
-use tokio::task::JoinError;
-
-pub async fn spawn_blocking<F, R, E>(f: F) -> Result<R, E>
-where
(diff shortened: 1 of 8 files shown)


crates/crates_io_worker/src/util.rs
```diff
@@ -3,23 +3,6 @@ use sentry_core::Hub;
...
-{
-    let current_span = tracing::Span::current();
-    let hub = Hub::current();
-    tokio::task::spawn_blocking(move || current_span.in_scope(|| Hub::run(hub, f)))
-        .await
-        // Convert `JoinError` to `E`
-        .map_err(Into::into)
-        // Flatten `Result<Result<_, E>, E>` to `Result<_, E>`
-        .and_then(std::convert::identity)
-}
```
LawnGnome, line 22: Is it wrong that I'm actually most excited by this change?

LawnGnome (approved):

PR #8431, opened 2024-04-09
controllers: Replace `r2d2` with `deadpool`

Turbo87:
This PR is roughly similar to #8385, but this time all of our API endpoints are ported from the sync `r2d2` database connection pool to the async `deadpool` equivalent.

A couple of minor refactorings were needed in some places, but the semantics should still be similar to before. I guess this is best reviewed commit-by-commit... 😅


LawnGnome (approved):
22 browser tabs later, :+1:.

PR #8440, opened 2024-04-10
controllers/user/session: Replace `r2d2` with `deadpool`

Turbo87:
This changes the order of operations a bit. Previously, we were first doing the GitHub requests and then connecting to the database. Now, we are first connecting to the database before doing the **sync** OAuth exchange with GitHub. This allows us to move from `spawn_blocking()` to `conn.interact()` though, which in turn allows us to get rid of the sync database connection pools soon.


Turbo87:
@LawnGnome I'll go ahead and merge this to unblock the next PR, but feel free to review after-the-fact :)

PR #8451, opened 2024-04-11
Remove `balance_capacity` middleware

Turbo87:
The main purpose of this middleware was to keep download requests working smoothly by dropping non-download requests if the load on the system was exceeding its limits. Now that download requests are no longer hitting our API the need for this middleware has significantly decreased. Additionally, this system has been running in report-only mode on production for the past couple of years, so we never really took advantage of it anyway.


LawnGnome (approved):
Where we're going, we don't need balance.

Issue #8461, opened 2024-04-12
Unable to add owner via CLI; kinda works via web


metasim:
### Current Behavior

I tried adding a github team as a crate owner. When doing it from the command line, even after logging in with the right API key, I get a `status 403 Forbidden`. When I went to the [settings page](https://crates.io/crates/pdal/settings) and added an owner there I got this:

<img src="https://github.com/rust-lang/crates.io/assets/131013/4306e988-a37c-466c-af80-7e8c6da3f8bd" width=400/>

And after I refresh the browser page, it (successfully) shows up as an owner:

<img src="https://github.com/rust-lang/crates.io/assets/131013/23d62bde-98f6-4068-b240-7243c4a0a88b" width=400/>

### Expected Behavior

Be able to add an owner via the CLI.

### Steps To Reproduce

1. `cargo login`
2. `cargo owner --add github:georust:pdal-publishers`

Result:
```
    Updating crates.io index
error: failed to invite owners to crate `pdal` on registry at https://crates.io

Caused by:
  the remote server responded with an error (status 403 Forbidden): must be logged in to perform that action
```

### Environment

cargo 1.76.0 (c84b36747 2024-01-18)

weihanglo:
Can't reproduce on my side. I've done the followings

* cargo version 1.77.2
* Created a change-owner scope token on my account (weihanglo)
* `cargo login` with that token
* `cargo owner --add github:rust-tw:jobs-board-maintainers` succeeded for my crate `sfz`
* Confirmed owner added on crates.io. 
* `cargo owner --remove github:rust-tw:jobs-board-maintainers` succeeded for my crate `sfz`
* Confirmed owner removed on crates.io.

Could you try `cargo logout` and then `cargo login` again? 

Also need more details from the server side.

weihanglo:
```
    Updating crates.io index
error: failed to invite owners to crate `sfz` on registry at https://crates.io

Caused by:
  the remote server responded with an error (status 403 Forbidden): must be logged in to perform that action

**Comment by weihanglo** [comment]
I got the same error message when the token scope is wrong.

```
    Updating crates.io index
error: failed to invite owners to crate `sfz` on registry at https://crates.io

Caused by:
  the remote server responded with an error (status 403 Forbidden): must be logged in to perform that action
```

Seems like crates.io returns the same error for different scenario. For 403 it's fine with returning fewer details, though maybe the error message could be more general?

[...]

PR #8482, opened 2024-04-17
Improve "403 Forbidden" error messages

Turbo87:
This PR adjusts our `forbidden()` error constructor fn by removing the hardcoded `details` message and replacing it with more contextual error messages.

Resolves https://github.com/rust-lang/crates.io/issues/8461


LawnGnome (approved):
Looks good to me. I don't have any concerns with any of those messages.

PR #8484, opened 2024-04-17
Add `default_versions` database table

Turbo87:
This PR adds a new database table: `default_versions`. This table has only two columns: `crate_id` and `version_id`.

Once the table has been filled with the `update-default-versions` admin command it will contain a row for every crate in the database, pointing to the version that will be shown by default by the front-end (not implemented yet) and can be used for other queries where useful (e.g. reverse dependencies).

This calculation has to be performed on the Rust side, since it involves sorting by SemVer and the Postgres extension for SemVer is unfortunately not available on Heroku.

There are four different operations to consider to keep this table in sync:

- Publishing a new crate: we write the `version_id` straight into the database since this is the first and therefore default version.
- Publishing a new version of an existing crate: we update the default version in a background job. this leaves us with a short time of the table pointing to an older, but existing version.
- Yanking/Unyanking a version: we update the default version in a background job. (see above)
- Deleting a version (admin action): we update the default version within the same transaction that deletes the versions. the foreign key constraint would otherwise prevent us from deleting a version that is currently a default version.

Note that the code currently does not prevent the `update_default_version()` fn from running multiple times in parallel for the same crate. I'm open to implementing such a lock, I just haven't found a good way to do it yet. It currently does not deduplicate the background jobs either, but since the job is quite fast it is most likely not needed anyway.

Note also that this table is currently write-only and not used anywhere yet. Once we have had this running in production for a few days we can check whether we see any inconsistencies before we use the data anywhere.


src/admin/update_default_versions.rs
eth3lbert, line 1: Should we accept an argument to specify crate_id(s)? If one isn't provided, then we update all crates.

migrations/2024-04-17-135931_default-versions-table/up.sql
```diff
@@ -0,0 +1,24 @@
...
+    crate_id   integer not null
+        constraint default_versions_pk
+            primary key
+        constraint default_versions_crates_id_fk
+            references crates
+            on delete cascade,
+    version_id integer not null
+        constraint default_versions_versions_id_fk
+            references versions
+            on delete no action deferrable initially deferred
```
eth3lbert, line 12: I'm thinking about the edge case of deleting a version where there's only a single version in a crate. Would the crate also be deleted?
If it's not, since the `version_id` column is not null in this PR, what will happen in this case?

Turbo87: it would fail to delete the version. and I guess that's a good thing because then the admin knows that they should've deleted the whole crate instead.

jtgeibel (approved):

PR #8498, opened 2024-04-19
Convert to async tests

Turbo87:
I'm sorry for the size of this PR... really...! 😅 

If you feel like it, this is probably best reviewed commit-by-commit, though most of the changes have been fairly mechanical.

The gist of this PR is to migrate us from `#[test]` to `#[tokio::test]`. This currently involves a few places where `block_in_place()` was necessary for the migration, so the tests are using a multi-threaded runtime (instead of the default current-thread runtime).

The primary three things that are now using async/await instead of blocking are:
- anything that looks like an HTTP request
- the `run_pending_background_jobs()` fn
- the `stored_files()` fn

Everything else is still roughly the same as before, including `app.db()` calls. That means we are doing blocking IO on the runtime threads for now, but since this is limited to the test code it shouldn't be too much of an issue, and is certainly something that will be cleaned up in a follow-up PR. Consider this PR very much as an intermediate step! :)

### DIFF
--- a/src/tests/account_lock.rs
+++ b/src/tests/account_lock.rs
@@ -21,12 +21,12 @@ fn lock_account(app: &TestApp, user_id: i32, until: Option<NaiveDateTime>) {
     });
 }
 
-#[test]
-fn account_locked_indefinitely() {
+#[tokio::test(flavor = "multi_thread")]
+async fn account_locked_indefinitely() {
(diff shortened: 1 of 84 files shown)


LawnGnome (approved):
I'm not going to pretend that I reviewed this in great, line-by-line detail. But the CI seems happy, the tests pass on my laptop, and the _types_ of changes all make sense to me. :shipit:

PR #8521, opened 2024-04-22
Add e2e tests with Playwright

eth3lbert:
This PR starts porting the current test suit that includes visits to [Playwright](https://playwright.dev/).  Additionally, the tests will be implemented in **Typescript**.

#### Steps to try locally:
```console
# Install Chromium for Playwright
pnpm exec playwright install chromium

# Run e2e tests with Playwright
pnpm e2e
```

#### Items for follow-up PRs:
- Skipping QUnit tests implemented in Playwright
- Coverage
- QUnit tests integration with Playwright (?)
  - Ref: https://github.com/lifeart/demo-ember-vite/blob/master/e2e/flows/qunit.spec.ts
- Adding e2e tests to CI workflow

----

For those who are familiar with QUnit but haven't used Playwright before, you may want to skim through [intro](https://playwright.dev/docs/intro), especially the [writing-tests](https://playwright.dev/docs/writing-tests) and [running-and-debugging](https://playwright.dev/docs/running-tests) sections. Playwright offers some test and debug functionalities similar to QUnit:

- `--debug` flag: Similar to development mode.
- `--ui` flag: Similar to the QUnit tests route.

There is also an [getting-started-vscode](https://playwright.dev/docs/getting-started-vscode) if you use it as your editor.


app/sentry.js
```diff
@@ -25,6 +25,11 @@ export function init() {
     beforeSend(event, hint) {
       let error = hint?.originalException;
 
+      // Expose errors only in non-production environments for testing with Playwright
+      if (config.environment !== 'production' && window) {
+        window.__SENTRY_EVENTS = (window.__SENTRY_EVENTS ?? []).concat(error);
+      }
+
```
eth3lbert, line 32: Ember uses QUnit, which can be quite difficult to test and mock with Playwright. I'm unsure if there's a better alternative approach to address this challenge.

Turbo87: what kind of options does playwright provide? e.g. does it allow us to run custom JS before executing the JS of the page that is loaded? ideally we would inject something into the page before the Ember app boots up.

eth3lbert: There is a function called [addInitScript](https://playwright.dev/docs/api/class-page#page-add-init-script):
```
It adds a script that would be evaluated in one of the following scenarios:
    * Whenever the page is navigated.
    * Whenever a child frame is attached or navigated. In this case, the script is evaluated in the context of the newly attached frame.
```

However, I'm not sure how to use it to work with Ember.js.

Turbo87: we could use it to set up a callback, which is then called by an [application instance initializer](https://guides.emberjs.com/release/applications/initializers/#toc_application-instance-initializers). that callback could then set up any service mocks or whatever else we need for the specific test. (theoretically... I haven't tried any of this :D)

eth3lbert: If I understand you correctly, you're suggesting to set up a generic mock hook. It would look something like this:
``` js
// some.spec.ts
await page.addInitScript(() => {
  window.__mock = function(app) {
    ...
  };
});

// initializers/mock.js
export function initializer(app) {
  if (config.environment !== 'production' && typeof window?.__mock === 'function') {
    window.__mock(app);
  }
}
```

This approach seems reasonable to me, and I imagine it would work great with plain js.

However, I'm unsure how to use this with Ember's dependency injection system. Additionally, Playwright seems to have issues with Ember modules (e.g., @ember/service, @ember/xxx), causing "module not found" errors.

Turbo87:
```js
class MockSentryService extends Service {
  // ...
}

app.register('service:sentry', MockSentryService);
```

something like this is what I had in mind, which allows us to override the services that the app uses, but the imports not working might make this a bit more complicated 😅

eth3lbert: Yes, this is exactly what I wanted to do but couldn't achieve.

---
When mocking the apis, we are likely to face the same issue, especially when we want to rely on the current miragejs implementation 😅 

We have two options to mocking apis:

- Reuse the existing miragejs
    This approach would require an option to inject the config and a hook to process some operations (e.g., `server.create` or `server.createList`).
    The simplest way to achieve this might be to use something like below, along with Playwright's `addInitScript` (miragejs need to intercept request in browser side) to make it work.

```js
// mirage/config.js
export default function makeServer(config) {
  let server = createServer({
    ...config,
    routes() {
      ...
    },
    ...window['__mirage_config'],
  });

  if (window['__mirage_hook']) {
    window['__mirage_hook'](server);
  }
  return server;
}
```

- Use Playwright's route:
 This option involves using Playwright's `route` to intercept the network request. We can then choose to:
  - Use miragejs to generate mock data.
  - Create mock data directly without miragejs

  However, if we opt to reuse the existing mirage codes, we would encounter the same "module not found" issue mentioned earlier due to imports not working.  This would require some changes, which I believe are reasonable as they would decouple miragejs from Ember.

eth3lbert: I've tried it before, but unfortunately, with no luck .

Currently, I'm going to implement it using the inject approach, which seems to have the best chance of reusing the existing miragejs implementation:
``` typescript
// e2e/helper.ts
export async function prepareMirage(
  page: Page,
  hook: prepareDataFn,
  options: ServerConfig<typeof MODELS, typeof FACTORIES> = DEFAULT_MIRAGE_CONFIG,
) {
  await page.addInitScript(options => {
    window['__mirage_config'] = options;
  }, options);
  await page.addInitScript(`window['__mirage_hook'] = ${hook.toString()};`);
}

// e2e/some.spec.ts
test('listing categories', async ({ page }, testInfo) => {
  await prepareMirage(page, server => {
    server.create('category', { category: 'API bindings' });
    server.create('category', { category: 'Algorithms' });
    // @ts-expect-error
    server.createList('crate', 1, { categoryIds: ['algorithms'] });
    server.create('category', { category: 'Asynchronous' });
    // @ts-expect-error
    server.createList('crate', 15, { categoryIds: ['asynchronous'] });
    server.create('category', { category: 'Everything', crates_cnt: 1234 });
  });

  await page.goto('/categories');

  await expect(page.locator('[data-test-category="api-bindings"] [data-test-crate-count]')).toHaveText('0 crates');
  await expect(page.locator('[data-test-category="algorithms"] [data-test-crate-count]')).toHaveText('1 crate');
  await expect(page.locator('[data-test-category="asynchronous"] [data-test-crate-count]')).toHaveText('15 crates');
  await expect(page.locator('[data-test-category="everything"] [data-test-crate-count]')).toHaveText('1,234 crates');
  ...
});
```

Compare to the existing tests:
``` javascript
// tests/acceptance/categories-test.js
test('listing categories', async function (assert) {
  this.owner.lookup('service:intl').locale = 'en';

  this.server.create('category', { category: 'API bindings' });
  this.server.create('category', { category: 'Algorithms' });
  this.server.createList('crate', 1, { categoryIds: ['algorithms'] });
  this.server.create('category', { category: 'Asynchronous' });
  this.server.createList('crate', 15, { categoryIds: ['asynchronous'] });
  this.server.create('category', { category: 'Everything', crates_cnt: 1234 });

  await visit('/categories');

  assert.dom('[data-test-category="api-bindings"] [data-test-crate-count]').hasText('0 crates');
  assert.dom('[data-test-category="algorithms"] [data-test-crate-count]').hasText('1 crate');
  assert.dom('[data-test-category="asynchronous"] [data-test-crate-count]').hasText('15 crates');
  assert.dom('[data-test-category="everything"] [data-test-crate-count]').hasText('1,234 crates');
  ...
});
```

nickschot: If you do want to try, the mirage-playwright setup is a bit different than the usual. For it to work with playwright in a (from the test perspective) convenient and performant way, mirage needs to run in node, which it can't by default due to pretender-js. That's where mirage-playwright comes in.  mirage-playwright relies on playwright intercepting the actual network requests done from the ember app and translates/forwards them to a mirage-js instance running from playwright (in node). There's no mirage within ember within this setup (like ember-cli-mirage would).

I haven't set it up with an ember app yet, as I haven't had that need before, but the above is how it's supposed to work. Let me know if you'd like more pointers in case you do want to try it out.

app/initializers/app-hook.js
eth3lbert, line 1: This is the most likely approach I could come up with to make overriding Ember services possible with an initializer. What do you think, @Turbo87 ?

eth3lbert:
A script like the following, which leverages `delta`, could be helpful for reviewing the changes. This script would compare the new implementation to the old one.
```python
import sys
from subprocess import PIPE, run


def get_changed_files():
    cmd = "git diff --name-only -r HEAD^ main"
    output = run(cmd, stdout=PIPE, stderr=PIPE, shell=True, universal_newlines=True)
    files = output.stdout
    return files


for new in get_changed_files().splitlines():
    if not new.endswith(".spec.ts"):
        continue
    old = new.replace("e2e", "tests").replace(".spec.ts", "-test.js")
    cmd = f"diff -u --ignore-space-change {old} {new} | delta"
    output = run(cmd, stdout=PIPE, stderr=PIPE, shell=True, universal_newlines=True)
    print(output.stdout)
    sys.stdin.read(1)
```

Turbo87:
due to the size of the diff and my upcoming PTO I probably won't be able to review the PR this week, but I've put it on my todo list for next week :)

eth3lbert:
Thanks! This PR definitely took some time to review, so I appreciate it.

Turbo87 (approved):
LGTM! :shipit: 

Given the size of the diff, I won't pretend to have reviewed every single test in this PR, but the approach in general looks good to me.

I rebased the branch on top of `main` to get rid of the merge conflict, and I fixed a CSS selector typo in the 404 test, which I noticed in the second commit.

One thing to consider: we might want to add an `e2e/README.md` explaining the purpose of the folder and that this is still WIP. I guess the next step then would be CI integration?

eth3lbert:
Thanks for the review! I'll integrate this change into our CI in a follow-up PR.

---

After this PR is merged, should we skip the overlapping Ember tests that have already been rewritten in Playwright?

Turbo87:
> After this PR is merged, should we skip the overlapping Ember tests that have already been rewritten in Playwright?

I think it's best to keep them running in parallel for now until we're confident about the new test setup

Issue #8536, opened 2024-04-25
Failed to log in: Error obtaining token after logging out


Mirch:
### Current Behavior

I logged out from crates.io. When logging back in with GitHub, I now see a `Failed to log in: Error obtaining token` pop-up.

### Expected Behavior

I should get logged in normally.

### Steps To Reproduce

1. Logout from you account.
2. Revoke the crates.io access from your GitHub account. (optional, I assume, but I gave it a try)
3. Attempt to log back in on crates.io.

### Environment

- Browser: Chrome / Safari
- OS: macOS 12.5.1

maxswjeon:
Me too, for accessing
```
GET https://crates.io/api/private/session/authorize?code=(redacted)
```
responses 500 (Internal Server Error) with body
```
{"errors":[{"detail":"Error obtaining token"}]}
```
with headers
```
X-Amz-Cf-Pop: ICN51-C1
X-Cache: Error from cloudfront
```

Related code is
https://github.com/rust-lang/crates.io/blob/d07e7f77a36455de99933f64f44084d6fb02cc26/src/controllers/user/session.rs#L98-L106

Turbo87:
I can reproduce the issue. I'll roll back today's deployments in the hope that it resolves the immediate issue and will investigate afterwards.

glennib:
I was just able to log in.

Mirch:
Works fine now. Thanks!

maxswjeon:
Works fine too.

Turbo87:
I've found the issue and https://github.com/rust-lang/crates.io/pull/8539 should fix it :)

Turbo87:
the fix is deployed to production now and at least on my side it appears to work as intended. sorry for the interruption!

PR #8539, opened 2024-04-25
oauth2: Enable default features

Turbo87:
`deb82a6` unintentionally disabled the TLS feature of the old `reqwest` version that is still used by the `oauth2` crate, because we declared that we don't want the default features of the crate. This commit enables the default features again, which turns the TLS support in `reqwest` back on.

Resolves https://github.com/rust-lang/crates.io/issues/8536

For more debugging context see https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/failing.20logins :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3500,6 +3500,7 @@ dependencies = [
  "http 0.2.12",
  "http-body 0.4.6",
  "hyper 0.14.28",
+ "hyper-rustls 0.24.2",
  "ipnet",
  "js-sys",
  "log",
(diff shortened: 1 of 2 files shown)


carols10cents (approved):
I don't suppose there's a way to add a test for this...? I know testing oauth is hard.

Approving though.

Turbo87:
@carols10cents if we'd control the other side too it would be easier, but to test this properly in our smoke testing setup we'd need a GitHub account that we can test the login with. then we could potentially use some browser automation tool to test the full login flow. since this includes automating the GitHub side of the login flow too I'm a bit worried about the flakiness of such a smoke test.

maxswjeon:
Hi, I know seeing this can be hard (since this is a comment on the closed PR thread) but I have some ideas regarding the smoke test. Why not having a mock OAuth2 Server?

maxswjeon:
BTW, sorry for pointing out the wrong commit for the issue. Might have made you confused a lot.

Turbo87:
> Why not having a mock OAuth2 Server?

if we had implemented it with a mock server we probably would've run it locally without TLS, which wouldn't have triggered this particular issue, so the test would've passed, but it still would've failed in production. another potential issue would be if our mock server doesn't behave exactly as the GitHub server does, which could also cause similar issues.

PR #8558, opened 2024-04-29
Add `os::android-apis` category for Android bindings

MarijnS95:
Android is its own OS on top of Linux with _lots_ of custom/specific interfaces.  Many crates are wrapping (parts) of it, such as [the NDK crate].  Since there's an `os::XXX-apis` tag for many OSes already, Android also belongs in this list.

[the NDK crate]: https://crates.io/crates/ndk

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -414,6 +414,12 @@ description = """
 Bindings to operating system-specific APIs.\
 """
 
+[os.categories.android-apis]
+name = "Android APIs"
+description = """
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):
seems reasonable. thanks :)

MarijnS95:
Looks like it's [public already](https://crates.io/categories/os::android-apis), many thanks!

I published a crate with this category before it was set live, and it doesn't seem to be accounted for that category yet though: https://crates.io/crates/android-powerstats

PR #8559, opened 2024-04-29
Implement dark mode via `light-dark()` fn

Turbo87:
This implements a dark mode for the user interface, based on the system theme setting. The `color-scheme` CSS attribute could eventually be overridden if the user does not want to use the system setting.

### Examples

Screen Shot 2024-04-29 at 14 34 30
Screen Shot 2024-04-29 at 14 34 42
Screen Shot 2024-04-29 at 14 34 51
Screen Shot 2024-04-29 at 14 35 16
Screen Shot 2024-04-29 at 14 35 42
Screen Shot 2024-04-29 at 14 35 49
Screen Shot 2024-04-29 at 14 36 02

### DIFF
--- a/app/components/color-scheme-menu.hbs
+++ b/app/components/color-scheme-menu.hbs
@@ -1,0 +1,20 @@
+<Dropdown local-class="dropdown" data-test-dark-mode-menu ...attributes as |dd|>
+  <dd.Trigger @hideArrow={{true}} local-class="trigger" data-test-dark-mode-toggle>
+    {{svg-jar this.icon local-class="icon"}}
+    <span local-class="trigger-label">Change color scheme</span>
+  </dd.Trigger>
+
+  <dd.Menu local-class="menu" as |menu|>
+    {{#each this.colorSchemes as |colorScheme|}}
+      <menu.Item>
+        <button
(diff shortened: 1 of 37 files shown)


Turbo87:
looks like I might have to double check that the PostCSS plugin is working correctly (see Percy), but this should still be available for review. I hope I didn't forget any pages...

also, do we require a user override button to ship this or can we ship this without one?

eth3lbert:
I would expect a color scheme toggle button to be included with this PR. This would allow users who don't want this change to simply switch back to the light theme.

eth3lbert:
I've implemented a simple dark mode toggle that can switch between light, dark, and system. Feel free to cherry-pick or modify it as needed if you find it helpful in any way. https://github.com/eth3lbert/crates.io/commit/0dc59e496bcc4a83e1a1e941b5ce20585546de61

image
image

Turbo87:
@eth3lbert thanks! I've cherry-picked the UI from your commit, but simplified the implementation a little bit :)

eth3lbert:
That's great! Adding an icon as a visual cue to the dropdown menu is a much better implementation. I love this!  I've tested it locally on the landing page without any issues.

PR #8658, opened 2024-05-16
worker/jobs/expiry_notification: Reduce log level for missing email addresses

Turbo87:
Ideally, we wouldn't have users without associated email addresses, but for some reason we currently appear to have such cases. Instead of treating this as a hard error for the token expiry notification we can log it with `INFO` level and mark the token as processed even if no notification has been sent out.

The current logs for this case are a bit more noisy than intended:

Bildschirmfoto 2024-05-16 um 11 28 59


hi-rustin (approved):
Looks good to me. Thanks! 👍

PR #8714, opened 2024-05-24
CI: Add e2e test

eth3lbert:
This is a follow-up PR to #8521 and add e2e test into our CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,3 +195,43 @@ jobs:
 
       - if: github.repository != 'rust-lang/crates.io'
         run: pnpm test-coverage
+
+  e2e-test:
+    name: Frontend / Test (playwright)
+    runs-on: ubuntu-22.04
+    needs: changed-files
+    timeout-minutes: 60
+    if: needs.changed-files.outputs.non-rust == 'true'
+
+    env:
+      JOBS: 1 # See https://git.io/vdao3 for details.
+
+      # Percy secrets are included here to enable Percy's GitHub integration
+      # on community-submitted PRs
+      PERCY_TOKEN: web_0a783d8086b6f996809f3e751d032dd6d156782082bcd1423b9b860113c75054
+
+    steps:
+      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+
+      - uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 # v4.0.0
+        with:
+          version: ${{ env.PNPM_VERSION }}
+
+      - uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          cache: pnpm
+          node-version-file: package.json
+
+      - run: pnpm install
+
+      - run: pnpm playwright install chromium
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -195,3 +195,42 @@ jobs:
 
       - if: github.repository != 'rust-lang/crates.io'
         run: pnpm test-coverage
+
+  e2e-test:
+    name: E2E / Test
```
Turbo87, line 200:
```suggestion
    name: Frontend / Test (playwright)
```

I don't think E2E is quite right for this since it implies that the tests are also testing our backend code, which is currently not the case (and I'm not sure we'd want to do that, for performance reasons)

.github/workflows/ci.yml
```diff
@@ -195,3 +195,42 @@ jobs:
...
+
+      - if: github.repository == 'rust-lang/crates.io'
+        run: pnpm percy exec -- pnpm e2e
+
+      - uses: actions/upload-artifact@v4
+        if: ${{ !cancelled() }}
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 14
```
Turbo87, line 236: > No files were found with the provided path: playwright-report/. No artifacts will be uploaded.

🤔

.github/workflows/ci.yml
eth3lbert: Fair enough! Thanks, I'll take this change.

.github/workflows/ci.yml
eth3lbert: I'll investigate this!
I just noticed that there's also a https://playwright.dev/docs/test-reporters#github-actions-annotations available, but I haven't tried it out yet.

eth3lbert:
It seems like the screenshot doesn't match the expected behavior. This could be because Ember tests and Playwright use different screen resolutions. I'll dig into this further to investigate.

Turbo87:
it looks like it was previously not loading the FiraSans font and with playwright it is loading it, which results in different font rendering.

eth3lbert:
> it looks like it was previously not loading the FiraSans font and with playwright it is loading it, which results in different font rendering.

Thanks for looking that up! That's interesting. Should we align with the Ember side and add a fixture to block the font resource, or fix it later on the Ember side?

Turbo87:
I'd just leave it as is for now. if we notice that it causes actual problems we can still address it later. once we merge this branch the loaded font will be the new baseline.

playwright.config.ts
```diff
@@ -18,9 +18,11 @@ export default defineConfig({
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 2 : undefined,
+  workers: process.env.CI ? 1 : undefined,
```
Turbo87, line 21: @eth3lbert is there a reason why we're overriding `workers` on CI? did you notice any issues with it running in parallel? it appeared to work fine for me locally at least.

eth3lbert: I think we don't have enough CPU capacity to run both `QUnit` and `E2E` tests simultaneously.  And yeah, it's indeed safe to run them in parallel. I usually do that locally too :D

PR #8733, opened 2024-05-27
tests/dump_db: Add basic end-to-end test

Turbo87:
We haven't been testing this part of the codebase properly, so this PR adds an "end-to-end" test for the database dumping background worker job. It will create a test database, put some dummy data into it, enqueue the database dump background worker job, run the pending background jobs, assert that a tarball was uploaded to the (in-memory) object store, download the tarball, and finally compare the tarball contents against an insta snapshot. This should ensure that we are not unintentionally breaking the generation of our database dumps.

Additionally, this PR introduces a mutex on the database dump tests, since all invocations of the code currently share the same output directory. Once this has been fixed, the mutex can be removed again.

### DIFF
--- a/src/tests/dump_db.rs
+++ b/src/tests/dump_db.rs
@@ -1,8 +1,16 @@
 use crates_io::worker::jobs::dump_db;
 use crates_io_test_db::TestDatabase;
+use once_cell::sync::Lazy;
+use std::sync::Mutex;
+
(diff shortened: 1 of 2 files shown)


Turbo87:
should've known... psql is obviously creating differences for all sorts of reasons and the order of the metadata files is also inconsistent 🎉

PR #8737, opened 2024-05-27
Improve the performance of reverse dependencies using the `default_versions`

eth3lbert:
This PR is the second in a sequence of multiple PRs that improve the performance of reverse dependencies using the `default_versions` table.


src/models/krate.rs
```diff
@@ -433,18 +432,32 @@ impl Crate {
...
+            .select(dependencies::version_id);
+        let yanked_subq = versions::table
+            .filter(versions::id.eq(default_versions::version_id))
+            .filter(versions::yanked)
+            .select(sql::<Integer>("1"));
+        let total = default_versions::table
+            .filter(default_versions::version_id.eq_any(dep_versions_subq))
+            .filter(not(exists(yanked_subq)))
+            .count()
+            .get_result(conn)?;
```
Turbo87, line 458: is running two queries after another really faster than doing it in a single query? I guess it would give us the option of caching the count separately, but other than that I would expect this to be slower than the alternative 😅

src/tests/routes/crates/reverse_dependencies.rs
```diff
@@ -10,14 +11,16 @@ async fn reverse_dependencies() {
...
+        let c2 = CrateBuilder::new("c2", user.id)
             .version(VersionBuilder::new("1.0.0").dependency(&c1, None))
             .version(
                 VersionBuilder::new("1.1.0")
                     .dependency(&c1, None)
                     .dependency(&c1, Some("foo")),
             )
             .expect_build(conn);
+        let _ = update_default_version(c1.id, conn);
+        let _ = update_default_version(c2.id, conn);
```
Turbo87, line 23: might be better to run `update_default_version()` in the `expect_build()` fn instead?

Turbo87 (commented):
for those that haven't followed the discussion on Zulip, could you include a summary of the performance numbers for these changes?

src/models/krate.rs
eth3lbert: You're right, after running some local benchmarks, it seems the single query is still slightly faster (~98ms vs ~103ms). Based on this, I'll revert the change back to using a single query.

src/tests/routes/crates/reverse_dependencies.rs
eth3lbert: I've been thinking about this as well. However, I'm concerned that it might become more difficult to simulate scenarios where default_versions doesn't catch up (even though we don't have such a scenario currently).

Do you think it would be more suitable to have an option to toggle with_default_versions or another build_expect that controls syncing default_versions? Or would that be too complicated?

Turbo87: I think if we want to test such a scenario we can manually adjust the value in the `default_versions` table as needed, but by default it should update it correctly.

src/models/krate_reverse_dependencies.sql
```diff
@@ -1,42 +1,32 @@
...
-            crate_downloads.downloads AS crate_downloads,
-            crates.name AS crate_name,
-            versions.id AS version_id
-        FROM
-        -- We only want the crates whose *max* version is dependent, so we join on a
-        -- subselect that includes the versions with their ordinal position
-        (
-            SELECT DISTINCT ON (crate_id)
-               crate_id, semver_no_prerelease, id
+WITH default_versions as (
```
Turbo87, line 1: maybe use a different CTE name again here to avoid confusion with the database table :)

src/tests/routes/crates/reverse_dependencies.rs
eth3lbert: Make sense.

src/models/krate_reverse_dependencies.sql
eth3lbert: No problem!

eth3lbert:
> for those that haven't followed the discussion on Zulip, could you include a summary of the performance numbers for these changes?

Sure! I'll update the description to include the summary.

eth3lbert:
Including `crate_id` in the unique index `default_versions_version_id_uindex` can potentially further improve performance.

eth3lbert:
> I'm surprised by this. if we add the `crate_id` then we're essentially just copying the data that is in the table into an index, so what is the difference then between scanning over the index vs. scanning over the table itself? 🤔

In my understanding, an `index scan` requires one additional access compared to an `index-only scan`. Previously, this query used an `index scan`. Covering index can improve performance by enabling an `index-only scan`. This allows the query to be answered solely from the index, without accessing the underlying table data (heap).

FYI: https://www.postgresql.org/docs/current/indexes-index-only-scans.html#INDEXES-INDEX-ONLY-SCANS

Turbo87:
I ran these queries on the production read-replica with these results:

- Current query: https://explain.dalibo.com/plan/6779823819d6c6d3 averages around 2.5sec
- Proposed query: https://explain.dalibo.com/plan/23daf51cbg3334eh averages around 500ms

nice work! 😍

Turbo87 (approved):

PR #8748, opened 2024-05-28
Create `db-dump.zip` file too

Turbo87:
Zip files use compression per file, which allows users to only extract the data that they need, instead of needlessly extracting the full tarball to read the small table that they are interested in.

Resolves https://github.com/rust-lang/crates.io/issues/2078

- [x] https://github.com/rust-lang/simpleinfra/blob/54bd88a2e0dfd474ef7a0f349565b555469491a9/terragrunt/modules/crates-io/compute-static/src/main.rs#L191-L198 might have to be adjusted

### DIFF
--- a/src/worker/jobs/dump_db.rs
+++ b/src/worker/jobs/dump_db.rs
@@ -34,7 +34,7 @@ impl BackgroundJob for DumpDb {
 
         let database_url = self.database_url.clone();
 
-        let (tarball, zip) = spawn_blocking(move || {
+        let archives = spawn_blocking(move || {
             let directory = DumpDirectory::create()?;
 
             info!("Exporting database…");
@@ -49,7 +49,9 @@ impl BackgroundJob for DumpDb {
         .await?;
 
         info!("Uploading tarball…");
-        env.storage.upload_db_dump(TAR_PATH, tarball.path()).await?;
+        env.storage
+            .upload_db_dump(TAR_PATH, archives.tar.path())
+            .await?;
         info!("Database dump tarball uploaded");
 
         info!("Invalidating CDN caches…");
@@ -66,7 +68,9 @@ impl BackgroundJob for DumpDb {
         }
 
         info!("Uploading zip file…");
-        env.storage.upload_db_dump(ZIP_PATH, zip.path()).await?;
+        env.storage
+            .upload_db_dump(ZIP_PATH, archives.zip.path())
+            .await?;
         info!("Database dump zip file uploaded");
 
         info!("Invalidating CDN caches…");
@@ -221,10 +225,12 @@ pub fn run_psql(script: &Path, database_url: &str) -> anyhow::Result<()> {
     Ok(())
 }
 
-fn create_archives(
-    export_dir: &Path,
-    tarball_prefix: &Path,
-) -> anyhow::Result<(tempfile::NamedTempFile, tempfile::NamedTempFile)> {
+struct Archives {
+    tar: tempfile::NamedTempFile,
+    zip: tempfile::NamedTempFile,
+}
+
+fn create_archives(export_dir: &Path, tarball_prefix: &Path) -> anyhow::Result<Archives> {
     debug!("Creating tarball file…");
     let tar_tempfile = tempfile::NamedTempFile::new()?;
     let encoder =
@@ -293,7 +299,10 @@ fn create_archives(
     drop(tar);
     zip.finish()?;
 
-    Ok((tar_tempfile, zip_tempfile))
+    Ok(Archives {
+        tar: tar_tempfile,
+        zip: zip_tempfile,
+    })
 }
 
 mod configuration;
@@ -321,8 +330,8 @@ mod tests {
         fs::write(p.join("data").join("crate_owners.csv"), "").unwrap();
         fs::write(p.join("data").join("users.csv"), "").unwrap();
 
-        let (tarball, zip) = create_archives(p, &PathBuf::from("0000-00-00")).unwrap();
-        let gz = GzDecoder::new(File::open(tarball.path()).unwrap());
+        let archives = create_archives(p, &PathBuf::from("0000-00-00")).unwrap();
+        let gz = GzDecoder::new(File::open(archives.tar.path()).unwrap());
         let mut tar = Archive::new(gz);
 
         let entries = tar.entries().unwrap();
@@ -341,7 +350,7 @@ mod tests {
         ]
         "###);
 
-        let file = File::open(zip.path()).unwrap();
+        let file = File::open(archives.zip.path()).unwrap();
         let reader = BufReader::new(file);
 
         let archive = zip::ZipArchive::new(reader).unwrap();
--- a/src/worker/jobs/dump_db.rs
+++ b/src/worker/jobs/dump_db.rs
@@ -6,6 +6,7 @@ use crates_io_worker::BackgroundJob;
 use std::fs::{self, File};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


src/worker/jobs/dump_db.rs
```diff
@@ -202,15 +221,22 @@ pub fn run_psql(script: &Path, database_url: &str) -> anyhow::Result<()> {
...
 }
 
-fn create_tarball(export_dir: &Path, prefix: &Path) -> anyhow::Result<tempfile::NamedTempFile> {
-    debug!("Creating tarball file");
-    let tempfile = tempfile::NamedTempFile::new()?;
-    let encoder = flate2::write::GzEncoder::new(tempfile.as_file(), flate2::Compression::default());
+fn create_archives(
+    export_dir: &Path,
+    tarball_prefix: &Path,
+) -> anyhow::Result<(tempfile::NamedTempFile, tempfile::NamedTempFile)> {
```
eth3lbert, line 227: Would it be better to return a struct with explicit named fields? Without seeing the implementation, it's difficult to determine the order of zip or tarball from the current return type.

eth3lbert:
> Zip files use compression per file, which allows users to only extract the data that they need, instead of needlessly extracting the full tarball to read the small table that they are interested in.

Actually, I think this could be achieved using `--wildcards` (or `--include` on mac) which only processes files matching the provided pattern.

e.g. `tar -zxvf tmp/db-dump.tar.gz --include '*.sql'`

Turbo87:
@eth3lbert yeah, that invocation might work, but the whole file still has to be decompressed by gzip before you can access the individual files inside the tarball. with a zip file you don't have to decompress the whole archive if you want to access just one file.

src/worker/jobs/dump_db.rs
Turbo87: done :)

eth3lbert (approved):

17351269:
// Forward purge requests immediately to a backend
    // https://developer.fastly.com/learning/concepts/purging/#forwarding-purge-requests
    if request.get_method() == "PURGE" {
        return send_request_to_s3(&config, &request);
    }

    init_logging(&config);
    let mut log = collect_request(&config, &request);

    let has_origin_header = request.get_header("Origin").is_some();
    let mut response = handle_request(&config, request);

    if has_origin_header {
        add_cors_headers(&mut response);
    }

    let log = collect_response(&mut log, &response);
    build_and_send_log(log, &config);

    response
}

/// Initialize the logger
///
/// Fastly provides its own logger implementation that streams logs to pre-configured endpoints. We
/// have created one endpoint for request logs and one for service logs.
///
/// Logs are echoed to stdout as well to enable tailing the logs with the Fastly CLI.
fn init_logging(config: &Config) {
    Logger::builder()
        .max_level(LevelFilter::Debug)
        .endpoint(config.datadog_request_logs_endpoint.clone())
        .endpoint(config.s3_request_logs_endpoint.clone())
        .default_endpoint(config.s3_service_logs_endpoint.clone())
        .echo_stdout(true)
        .init();
}

/// Collect data for the logs from the request
fn collect_request(config: &Config, request: &Request) -> LogLineV1Builder {
    let http_details = HttpDetailsBuilder::default()
        .protocol(http_version_to_string(request.get_version()))
        .referer(
            request
                .get_header("Referer")
                .and_then(|s| s.to_str().ok())
                .map(|s| s.to_string()),
        )
        .useragent(
            request
                .get_header("User-Agent")
                .and_then(|s| s.to_str().ok())
                .map(|s| s.to_string()),
        )
        .build()
        .ok();

    let tls_details = TlsDetailsBuilder::default()
        .cipher(request.get_tls_cipher_openssl_name())
        .protocol(request.get_tls_protocol())
        .build()
        .ok();

    let log_line = LogLineV1Builder::default()
        .ddtags(format!("app:{},env:{}", DATADOG_APP, config.datadog_env))
        .service(DATADOG_SERVICE)
        .date_time(OffsetDateTime::now_utc())
        .edge_location(var("FASTLY_POP").ok())
        .host(request.get_url().host().map(|s| s.to_string()))
        .http(http_details)
        .ip(request.get_client_ip_addr())
        .method(Some(request.get_method().to_string()))
        .url(request.get_url_str().into())
        .tls(tls_details)
        .to_owned();

    log_line
}

/// Handle the request
///
/// This method handles the incoming request and returns a response for the client. It first ensures
/// that the request uses whitelisted request methods, then sets a TTL to cache the response, before
/// finally forwarding the request to S3.
fn handle_request(config: &Config, mut request: Request) -> Result<Response, Error> {
    if let Some(response) = limit_http_methods(&request) {
        return Ok(response);
    }

    set_ttl(config, &mut request);
    rewrite_urls_with_plus_character(&mut request);
    rewrite_download_urls(&mut request);

    // Database dump is too big to cache on Fastly
    if request.get_url_str().ends_with("db-dump.tar.gz") {
        redirect_db_dump_to_cloudfront(config)
    } else {
        send_request_to_s3(config, &request)
    }
}

/// Limit HTTP methods
///
/// Clients are only allowed to request resources using GET and HEAD requests. If any other HTTP
/// method is received, HTTP 403 Unauthorized is returned.
///
/// We don't return HTTP 405 Method Not Allowed to maintain parity with CloudFront.
fn limit_http_methods(request: &Request) -> Option<Response> {
    let method = request.get_method();

    if method != Method::GET && method != Method::HEAD {
        return Some(
            Response::from_body("Method not allowed").with_status(StatusCode::UNAUTHORIZED),
        );
    }

    None
}

/// Set the TTL
///
/// A TTL header is added to the request to ensure that the content is cached for the given amount
/// of time.
fn set_ttl(config: &Config, request: &mut Request) {
    request.set_ttl(config.static_ttl);
}

/// Rewrite URLs with a plus character
///
/// An issue was reported for crates.io where URLs that encoded the `+` character in a crate's
/// version as `%2B` were not working correctly. As a backwards-compatible fix, we are transparently
/// rewriting URLs that contain the `+` character to use `%2B` instead. This ensures that crates in
/// Amazon S3 are accessed in a consistent way across all clients and Content Delivery Networks.
///
/// See more: https://github.com/rust-lang/crates.io/issues/4891
fn rewrite_urls_with_plus_character(request: &mut Request) {
    let url = request.get_url_mut();
    let path = url.path();

    if path.contains('+') {
        let new_path = path.replace('+', "%2B");
        url.set_path(&new_path);
    }
}

/// Rewrite `/crates/{crate}/{version}/download` URLs to
/// `/crates/{crate}/{crate}-{version}.crate`
///
/// cargo versions before 1.24 don't support placeholders in the `dl` field
/// of the index, so we need to rewrite the download URL to point to the
/// crate file instead.
fn rewrite_download_urls(request: &mut Request) {
    static RE: Lazy<Regex> = Lazy::new(|| {
        Regex::new(r"^/crates/(?P<crate>[^/]+)/(?P<version>[^/]+)/download$").unwrap()
    });

    let url = request.get_url_mut();
    let path = url.path();

    if let Some(captures) = RE.captures(path) {
        let krate = captures.name("crate").unwrap().as_str();
        let version = captures.name("version").unwrap().as_str();
        let new_path = format!("/crates/{krate}/{krate}-{version}.crate");
        url.set_path(&new_path);
    }
}

/// Redirect request to CloudFront
///
/// As of early 2023, certain files are too large to be served through Fastly. One of those is the
/// database dump, which gets redirected to CloudFront.
fn redirect_db_dump_to_cloudfront(config: &Config) -> Result<Response, Error> {
    let url = format!("https://{}/db-dump.tar.gz", config.cloudfront_url);
    Ok(Response::temporary_redirect(url))
}

LawnGnome (approved):
:zipper_mouth_face:

Issue #8410, opened 2024-04-08
Crates with paths differing only by case are allowed


kornelski:
### Current Behavior

It's possible to have a tarball that contains multiple files with paths that are identical except case, e.g. `readme.md` and `README.MD`. On case-insensitive file systems on of these will overwrite the other. Tricks like these could confuse tools or even hide files from code reviews performed on macOS or Windows.

e.g. [Cargo can pacakge duplicate `cargo.lock` file](https://github.com/rust-lang/cargo/issues/13722), and [`cargo.toml` breaks docs.rs but not crates.io](https://github.com/rust-lang/docs.rs/issues/2491).

I haven't checked whether it's still allowed to have completely duplicate paths, but there are many old crates with duplicate files, e.g. two `Cargo.toml.orig` files: https://crates.io/crates/nucleo-f103rb/0.3.0

### Expected Behavior

I think it'd be best if crates.io refused to accept any tarballs with potentially duplicate paths. It's not sufficient to rely on `cargo publish` to work correctly, because attackers could upload arbitrary tarballs. Even if crates.io could safely deal with weird tarballs, they can be causing trouble in 3rd party tools.

Turbo87:
we currently only check this scenario for Cargo.toml files, but I tend to agree that we might want to restrict this to all files. 

@rust-lang/crates-io any thoughts on this?

LawnGnome:
Reopening, since we ended up reverting #8788. (See that PR for details.)

Turbo87:
this issue is blocked on:
- https://github.com/rust-lang/cargo/issues/13722 
- and https://github.com/rust-lang/cargo/issues/14020

PR #8788, opened 2024-06-04
crates_io_tarball: Disallow paths differing only by case

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/8410

This implementation uses `.to_ascii_lowercase()` on the `Path` to compare the paths. In other words: this may or may not be sufficient depending on the filesystem and the usage of non-ASCII characters in the paths. It seems to cover the majority of the existing cases though, so it's at least a step in the right direction.

In the future we could even consider rejecting non-ASCII paths, in which case this implementation would probably be sufficient.

/cc @kornelski

### DIFF
--- a/crates/crates_io_tarball/src/lib.rs
+++ b/crates/crates_io_tarball/src/lib.rs
@@ -9,7 +9,7 @@ use crate::manifest::validate_manifest;
 pub use crate::vcs_info::CargoVcsInfo;
 pub use cargo_manifest::{Manifest, StringOrBool};
 use flate2::read::GzDecoder;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 use std::io::Read;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -33,6 +33,8 @@ pub enum TarballError {
     Malformed(#[source] std::io::Error),
     #[error("invalid path found: {0}")]
     InvalidPath(String),
+    #[error("duplicate path found: {0}")]
+    DuplicatePath(String),
     #[error("unexpected symlink or hard link found: {0}")]
     UnexpectedSymlink(String),
     #[error("Cargo.toml manifest is missing")]
@@ -41,8 +43,6 @@ pub enum TarballError {
     InvalidManifest(#[from] cargo_manifest::Error),
     #[error("Cargo.toml manifest is incorrectly cased: {0:?}")]
     IncorrectlyCasedManifest(PathBuf),
-    #[error("more than one Cargo.toml manifest in tarball: {0:?}")]
-    TooManyManifests(Vec<PathBuf>),
     #[error(transparent)]
     IO(#[from] std::io::Error),
 }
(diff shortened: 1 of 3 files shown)


kornelski (approved):

LawnGnome (approved):
LGTM.

I think we'll eventually have to grapple with the terrible questions of "so, what, if any, encoding do we expect paths in crate files to be in", and "is uniqueness _actually_ defined by the NF[CD] canonicalisations of path names after case folding, if the answer to the previous question involves Unicode in some way", but let's not let perfect be the enemy of good here.

zimond:
after this, I cannot publish my crate with the following error:
```
the remote server responded with an error (status 400 Bad Request): uploaded tarball contains more than one file with the same path xxx/README.md
```
I have to delete the readme file to upload a new version, which is weird. The bundled artifact `.crate` file contains only one readme file and I don't know where this error comes from.

Turbo87:
for what crate is this? can you `cargo package` and send us the file to take a look?

zimond:
@Turbo87 it's https://crates.io/crates/fontkit
[fontkit-0.6.0-beta.5.crate.zip](https://github.com/user-attachments/files/15605400/fontkit-0.6.0-beta.5.crate.zip)

eth3lbert:
The doc.rs source page indicates that there are two `README.md` files.
https://docs.rs/crate/fontkit/0.5.2/source/

zimond:
Now I'm really confused.

zimond:
Could it be that `cargo` is generating a `README.md` for me? [link](https://github.com/rust-lang/cargo/blob/10b7d384352f6d9a39f609de44476f815fc792a2/src/doc/src/reference/manifest.md?plain=1#L230)

zimond:
Ok so with `readme = "Readme.md"` in `Cargo.toml`, I successfully uploaded a new version `0.6.0-beta.6` to crates.io.

epage:
This is likely #13722 which comes from our logic to ensure a README or a LICENSE from another location in the repo is pulled into the `.crate` file

https://github.com/rust-lang/cargo/blob/10b7d384352f6d9a39f609de44476f815fc792a2/src/cargo/ops/cargo_package.rs#L330-L338

epage:
btw this seems like this would have been a good change to coordinate with the Cargo team on as this is interrelated to Cargo's behavior and choices Cargo makes.  If nothing else, we should probably be consistent on our case sensitivity logic.

ehuss:
Filed https://github.com/rust-lang/cargo/issues/14020 about the readme specifically.

ehuss:
@Turbo87 I would suggest reverting this until things are fixed, as I suspect this will impact a large number of users.

I would also suggest doing a scan of existing crates to see how common it is. This may trigger for all sorts of reasons which we don't know about.

LawnGnome:
I've just deployed crates.io with this reverted.

Let's figure out what the right behaviour should be here and then get cargo and crates.io on the same page.

PR #8794, opened 2024-06-05
Postgres: Set `application_name` to crates.io

Turbo87:
This makes queries from the application easier to recognize in the logs.

see https://www.postgresql.org/docs/current/runtime-config-logging.html#GUC-APPLICATION-NAME


LawnGnome (approved):
I've never heard of this, but it seems like a good idea to me. :shrug:

PR #8824, opened 2024-06-10
worker/jobs/dump_db: Only export the `public` schema via `pg_dump`

Turbo87:
Postgres databases sometimes have additional "schemas" in them, occasionally with different access permissions. One example of this is the `cron` schema that the Scheduler feature on CrunchyBridge creates. This schema is not accessible by the `application`, which causes `pg_dump` to break. By restricting `pg_dump` to the `public` schema we can fix this issue.


Turbo87:
looks like this is causing our generated `schema.sql` file to suddenly include `CREATE SCHEMA public` which is not quite what we want... back to the drawing board

PR #8832, opened 2024-06-11
Add supports_attributes Flag to Storage Struct to Support ObjectStore Implementations

### LINKED ISSUE #8823: Publishing Crates Locally Fails Due to Unsupported put_opts Operation on a LocalFileSystem ObjectStore
### Current Behavior

Following the guide to publish crates locally ([found here](/Users/jonathan/dev/crates.io/local_uploads)), when I run `cargo publish --index file:///path/to/your/crates.io/tmp/index-bare --token $YOUR_TOKEN` (with the appropriate substitutions) I receive the following response from cargo:

```
error: failed to publish to registry at http://localhost:8888/

Caused by:
  the remote server responded with an error: Internal Server Error
```

And in the server logs:

```
ERROR crates_io::util::errors: Internal Server Error error=failed to upload crate: Operation not yet implemented.
 INFO http: method=PUT path="//api/v1/crates/new" request_id= ip="127.0.0.1" service=78ms status=200 user_agent="cargo/1.78.0 (54d8815d0 2024-03-26)" normalized_path="/api/v1/crates/new" crate_name="test-crate" crate_version="0.1.0" uid="1" tokenid="2" error="failed to upload crate: Operation not yet implemented."
```

### Expected Behavior

Expected: Package should publish successfully.

Cargo:
```
Published test-crate v0.1.0 at registry `/Users/jonathan/dev/crates.io/tmp/index-bare`
```

Server:
```
INFO http: method=PUT path="//api/v1/crates/new" request_id= ip="127.0.0.1" service=78ms status=200 user_agent="cargo/1.78.0 (54d8815d0 2024-03-26)" normalized_path="/api/v1/crates/new" crate_name="test-crate" crate_version="0.1.0" uid="1" tokenid="2"
```

### Steps To Reproduce

Follow instructions in the CONTRIBUTING.md file for publishing crates locally.

https://github.com/rust-lang/crates.io/blob/main/docs/CONTRIBUTING.md#publishing-a-crate-to-your-local-cratesio

### Environment

- OS: macOS 14.4.1

### Anything else?

I have identified the issue being with the object_store crate. It seems recently the object_store team added a `NotImplemented` error when using `put_opts` on a LocalFileSystem ObjectStore, under the condition that opts are provided ([found in their git repo](https://github.com/apache/arrow-rs/blob/087f34b70e97ee85e1a54b3c45c5ed814f500b0a/object_store/src/local.rs#L349)):

```rust
if !opts.attributes.is_empty() {
    return Err(crate::Error::NotImplemented);
}
```

Currently, crates.io uses the put_opts method like so (self.store is a LocalFileSystem ObjectStore when using the local file system):

```rust
#[instrument(skip(self, bytes))]
pub async fn upload_crate_file(&self, name: &str, version: &str, bytes: Bytes) -> Result<()> {
    let path = crate_file_path(name, version);
    let attributes = Attributes::from_iter([
            (Attribute::ContentType, CONTENT_TYPE_INDEX),
            (Attribute::CacheControl, CACHE_CONTROL_INDEX),
        ]);
    let opts = attributes.into();
    self.store.put_opts(&path, bytes.into(), opts).await?;
    Ok(())
}
```

A quick fix I found to work is to replace `Attributes::from_iter()` with `Attributes::new()`

```rust
#[instrument(skip(self, bytes))]
pub async fn upload_crate_file(&self, name: &str, version: &str, bytes: Bytes) -> Result<()> {
    let path = crate_file_path(name, version);
    let attributes = Attributes::new();
    let opts = attributes.into();
    self.store.put_opts(&path, bytes.into(), opts).await?;
    Ok(())
}
```

I don't think this is ideal.

mcclementines:
Resolves #8823

I liked the suggestion on the issue and implemented as described. Checked with the object_store crate for the correct values of the flag to support InMemory (true), LocalFileSystem (false), and S3 (true) object stores. 

I thought using match to set the attributes based on the flag looked cleaner, but I believe the if statement is more standard style.

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -111,6 +111,7 @@ pub struct Storage {
     cdn_prefix: Option<String>,
     store: Arc<dyn ObjectStore>,
     index_store: Arc<dyn ObjectStore>,
+    supports_attributes: bool,
 }
 
 impl Storage {
@@ -134,6 +135,8 @@ impl Storage {
 
                 let index_store = build_s3(index, Default::default());
 
+                let supports_attributes = true;
+
                 if cdn_prefix.is_none() {
                     panic!("Missing S3_CDN environment variable");
                 }
@@ -142,6 +145,7 @@ impl Storage {
                     cdn_prefix,
                     store: Arc::new(store),
                     index_store: Arc::new(index_store),
+                    supports_attributes,
                 }
             }
 
@@ -165,10 +169,13 @@ impl Storage {
                 let store: Arc<dyn ObjectStore> = Arc::new(local);
                 let index_store: Arc<dyn ObjectStore> = Arc::new(local_index);
 
+                let supports_attributes = false;
+
                 Self {
                     cdn_prefix,
                     store,
                     index_store,
+                    supports_attributes,
                 }
             }
 
@@ -176,10 +183,13 @@ impl Storage {
                 warn!("Using in-memory file storage");
                 let store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
 
+                let supports_attributes = true;
+
                 Self {
                     cdn_prefix,
                     store: store.clone(),
                     index_store: Arc::new(PrefixStore::new(store, "index")),
+                    supports_attributes,
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/storage.rs
```diff
@@ -238,10 +252,14 @@ impl Storage {
...
     pub async fn upload_readme(&self, name: &str, version: &str, bytes: Bytes) -> Result<()> {
         let path = readme_path(name, version);
-        let attributes = Attributes::from_iter([
-            (Attribute::ContentType, CONTENT_TYPE_README),
-            (Attribute::CacheControl, CACHE_CONTROL_README),
-        ]);
+        let attributes = if self.supports_attributes {
+            Attributes::from_iter([
+                (Attribute::ContentType, CONTENT_TYPE_CRATE),
+                (Attribute::CacheControl, CACHE_CONTROL_IMMUTABLE),
```
Turbo87, line 258: copy-paste mistake? 😉

Issue #8674, opened 2024-05-18
Dark mode search button colors too aggressive


alerque:
The new dark mode is great to have, but some of the color choices are a bit jaring. It feels like some things are leftover from a different theme, not a design choice for the dark theme.

For example the search button in the header and home page is way too bright and overpowers everything on any page, including the logo.

![image](https://github.com/rust-lang/crates.io/assets/173595/fcfc783f-5109-4c3b-9fa4-845901d7cb01)

In fact the other yellow button colors are much more subdued, some of them perhaps *too* muted and inconsistent:

![image](https://github.com/rust-lang/crates.io/assets/173595/af88129f-bcfe-4183-9891-fd25a0c6ea71)

Of these the two big buttons from the home page look about right, but the "follow" buttons seem too muted and the "search" buttons seem to bright.

rope-hmg:
I also find the text colour is too bright. And possibly the background is too dark?

<img width="634" alt="image" src="https://github.com/rust-lang/crates.io/assets/2173098/9e39e776-aca1-4722-90c6-520d8ed8e399">

alerque:
@rope-hmg I would suggest a different issue, that's really a different discussion.

rope-hmg:
Ah yes. The dark mode is not good, but talking about how the dark mode isn't good isn't relevant. Yep, I see it.

Seems to me like it's very reasonable to mention some of the colour choices being too bright. Especially when you yourself said "but some of the color choices are a bit jaring".

Turbo87:
> It feels like some things are leftover from a different theme, not a design choice for the dark theme.

while not necessarily intentional, the current color seems to fit okayish IMHO. since it's also one of our somewhat unofficial brand colors I'm not sure that we really want to change it.

> I also find the text colour is too bright. And possibly the background is too dark?

the current color scheme has reasonably good contrast. if we move to light-grey on dark-grey this will be significantly worse.

ultimately, this is a bikeshed discussion and we won't be able to please everyone. I'd be very happy if a professional designer got involved and helped us improve the design of the entire page, but until that is the case we all have to live with the current level of beauty 😅

PR #8853, opened 2024-06-13
Frontend (CSS changes)

Daksh-10:
With reference to #8674,
I have tried upgrading the view of the website


Turbo87:
thank you for the pull request. unfortunately I'm not sure that it really improves the situation though. IMHO the search button isn't that bright and it's rather small, so I'm not even convinced it needs "fixing" 😅

Issue #5882, opened 2023-01-07
Display `cargo install crate-name` rather than `cargo add crate-name` for binary crates


Aloso:
### Current Behavior

crates.io has an "Install" sections with copy-paste snippets to install the crate. It includes these two snippets:

```
cargo add crate-name
```

```
crate-name = "version"
```

For **binary crates** (e.g. [ripgrep](https://crates.io/crates/ripgrep)) this is not useful.

### Expected Behavior

For crates that have a binary target, but not a library target, crates.io should show instead:

```
cargo install crate-name
```

For crates that have both a binary target and a library target, crates.io should all three:

```
cargo add crate-name
```

```
crate-name = "version"
```

```
cargo install crate-name
```

### Steps To Reproduce

1. Go to https://crates.io/crates/ripgrep
2. Look at the page

### Environment

- Browser: any
- OS: irrelevant

epage:
At the moment, crates.io has no knowledge of whether a crate has a `[[bin]]` or a `[lib]` component.  You can see what `cargo publish` tells crates.io [here](https://github.com/rust-lang/cargo/blob/master/crates/crates-io/lib.rs#L42).

Adjusting crates.io's backend to take that information and cargo to report it would be the first step.  We'd then need to wait a period of time for publishes to happen to get the new data.  In theory, we could tear apart the crates and update the metadata but I suspect this is a more more complex of a case (e.g. different lib and main auto-detection rules) that we would just wait for new publishes.  Then the webpage could start sporting the new information.

I feel like there were an issue or two in cargo's backlog where having this information would be helpful but I can't find them at the moment.

epage:
I remembered what cargo's case is.

If we are already putting whether a crate has a `[[bin]]` and/or a `[lib]` in the index, it could be useful for us to enumerate what `[[bin]]`s are present.

In cargo today, if you type `cargo foo` without the `cargo-foo` bin being installed, we could check if a `cargo-foo` crate exists and suggest `cargo install cargo-foo`.  The more advanced solution is if we can walk the entire index, checking for what crate includes the `cargo-foo` bin, we can then suggest that crate name.  For example, the `cargo upgrade` bin is provided by `cargo-edit` crate.

See rust-lang/cargo#4682 for more details.

jonasbb:
> [...] it could be useful for us to enumerate what `[[bin]]`s are present.

[...]

PR #8859, opened 2024-06-13
Add `has_lib` and `bin_names` fields to the versions API

Turbo87:
https://docs.rs/cargo-manifest/latest/cargo_manifest/struct.Manifest.html#method.complete_from_abstract_filesystem is used to find implicit libraries or binaries in the crate file.

Tasks:

- [x] find libraries, binaries, etc. in the crate file
- [x] save that data in the database
- [x] expose that data in the API
- [x] adjust the install instructions on the web frontend depending on the data (see https://github.com/rust-lang/crates.io/pull/8883)
- [ ] figure out a way to backfill the data for existing versions

Related:

- https://github.com/rust-lang/crates.io/issues/5882
- https://github.com/rust-lang/crates.io/pull/8883


Turbo87:
This PR is ready for review now. The frontend part of this work will be opened in an independent PR, same for the backfilling code.

migrations/2024-06-14-081653_add-lib-and-bins-columns/up.sql
```diff
@@ -0,0 +1,6 @@
+alter table versions
+    add has_lib boolean,
+    add bin_names text[];
+
+comment on column versions.has_lib is 'TRUE if the version has a library (e.g. `src/lib.rs`), FALSE if no library was detected, or NULL if the version has not been analyzed yet.';
+comment on column versions.bin_names is 'list of the names of all detected binaries in the version. the list may be empty which indicates that no binaries were detected in the version. the column may be NULL is the version has not been analyzed yet.';
```
eth3lbert, line 6: IIUC, null is used for existing versions that need backfilling? And for newly published versions, a non-null value would be expected to be inserted? Do we also plan to alter the table to non-nullable once all values are filled?

Turbo87: > IIUC, null is used for existing versions that need backfilling? And for newly published versions, a non-null value would be expected to be inserted?

correct

> Do we also plan to alter the table to non-nullable once all values are filled?

I'm not sure yet. at the moment the detection is done inside the publish endpoint, but if we were to move the detection to an async background worker job we might want to keep this "no-yet-analyzed" state.

LawnGnome (approved):
With the obvious caveat that we're now dependent on `cargo-manifest`'s target discovery working the same way as `cargo`, this LGTM. (On the bright side, this is something that's easily re-backfilled later if there are any issues, and certainly not on any critical path for publishing or using crates.)

Turbo87:
> With the obvious caveat that we're now dependent on `cargo-manifest`'s target discovery working the same way as `cargo`, this LGTM.

yeah, that's why I'm currently working on fixing some of the finer details of the target auto-discovery in `cargo-manifest` :D

migrations/2024-06-14-081653_add-lib-and-bins-columns/up.sql
eth3lbert: Thanks for the explanation!

This actually reminds me of `default_versions`, which faces a similar situation but actually handles it quite well.

Turbo87:
https://github.com/rust-lang/crates.io/pull/8951 has been merged, so this is now unblocked :)

PR #8867, opened 2024-06-14
Remove obsolete `versions.semver_no_prerelease` database column

Turbo87:
... and the related `to_semver_no_prerelease()` database function.

The usage of this column in the reverse dependencies query has been replaced by the `default_versions` table, so the column is no longer needed.


eth3lbert (approved):
👍

LawnGnome (approved):

PR #8870, opened 2024-06-14
tests/builders/version: Simplify `license` field

Turbo87:
We're using regular `Option<String>` already for `links` and `rust_version` and while the `&str` might be slightly more efficient it is also slightly less ergonomic. Switching the field type also allows us to remove the `'a` lifetime declaration from the struct which makes it a bit easier on the eyes too. If we ever feel like we need the raw speed we can always add it back.

### DIFF
--- a/src/tests/builders/version.rs
+++ b/src/tests/builders/version.rs
@@ -9,11 +9,11 @@ use chrono::NaiveDateTime;
 use diesel::prelude::*;
 
 /// A builder to create version records for the purpose of inserting directly into the database.
-pub struct VersionBuilder<'a> {
+pub struct VersionBuilder {
     created_at: Option<NaiveDateTime>,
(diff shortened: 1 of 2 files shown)


src/tests/builders/version.rs
```diff
@@ -56,8 +56,8 @@ impl<'a> VersionBuilder<'a> {
     }
 
     /// Sets the version's `license` value.
-    pub fn license(mut self, license: Option<&'a str>) -> Self {
-        self.license = license;
+    pub fn license(mut self, license: impl Into<String>) -> Self {
+        self.license = Some(license.into());
         self
     }
```
Turbo87, line 62: and if you're wondering if I don't have to adjust all of the call sites? you're right, but this function isn't actually used anywhere at the moment so the perf hit is basically zero 😂

PR #8873, opened 2024-06-15
Remove background color from preformatted code blocks within lists

### LINKED ISSUE #8866: Dark mode: Code blocks in lists retain inline style background
### Current Behavior

A code block that is inside a list item has the CSS background for inline code applied to it in addition to the normal code block style.

### Expected Behavior

The code block should be styled the same as top-level code blocks.

### Steps To Reproduce

An example is the readme for [`delegate`](https://crates.io/crates/delegate). The first code block renders fine, while the subsequent code blocks are inside a list and look bad. This is only an issue in dark mode.

### Environment

- Browser: Firefox 127
- OS: Windows 10

### Anything else?

_No response_

eth3lbert:
This should resolve #8866. 

image

### DIFF
--- a/app/components/text-content.module.css
+++ b/app/components/text-content.module.css
@@ -38,6 +38,12 @@
             margin: 0;
             padding: var(--space-4xs) var(--space-3xs);
         }
+        /* Remove background color and padding from preformatted code blocks within lists. */
+        /* See: https://github.com/rust-lang/crates.io/issues/8866 */
+        pre code {
+            background-color: initial;
+            padding: initial;
+        }
     }
 
     code {
--- a/app/components/text-content.module.css
+++ b/app/components/text-content.module.css
@@ -38,12 +38,6 @@
             margin: 0;
             padding: var(--space-4xs) var(--space-3xs);
         }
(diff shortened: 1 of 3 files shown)


kext:
If you look closely you can see the that though the background color is gone, the first line is still slightly indented from the border radius and padding effect. Same for the last line but that is not visible.

eth3lbert:
@kext Thanks for the feedback!

> If you look closely you can see the that though the background color is gone, the first line is still slightly indented from the border radius and padding effect. Same for the last line but that is not visible.

> That is actually also true for the light version.

The indention issue has also been addressed now.
image
image


> For normal code blocks the font size is also `1em` instead of `85%`.

I'm unsure about changing the font size within the lists. While I'm open to the suggestion, I'd like to hear feedback from other reviewers before making a decision.

Turbo87:
> I'm unsure about changing the font size within the lists. While I'm open to the suggestion, I'd like to hear feedback from other reviewers before making a decision.

We're trying to roughly follow what GitHub does, and they don't decrease the font-size, so I guess we should do the same.

I've slightly adjusted the PR to apply to all `.hljs` elements and added the font-size reset. Thanks for working on this! :)

Issue #7884, opened 2024-01-05
Problem when trying to install a specific version


gabriiels:
### Current Behavior

When attempting to install a specific version of a package, the command shown is incorrect.
For example, if you want to install version 2.6.0 of MongoDB, it gives you this command:
```
cargo add mongodb@2.6.0
mongodb = "2.6.0"

```
However, that line installs version 2.8.0. In all cases, it seems to install the latest versions, which have a green icon.

![fix-1](https://github.com/rust-lang/crates.io/assets/49616724/15a27424-f76a-450a-8214-a898e8625ba2)

![fix-crates](https://github.com/rust-lang/crates.io/assets/49616724/44f7c5dd-6f3e-463f-a708-adab8cac3da7)

### Expected Behavior

In the case of specific versions, it's necessary to add the "=" symbol:
```
cargo add mongodb@=2.6.0
mongodb = "=2.6.0""
```
![fix-2](https://github.com/rust-lang/crates.io/assets/49616724/bb9a0180-1d41-4e62-9fd3-c58b11570030)

### Environment

- Browser: 
- OS: Linux 

### Anything else?

Sometimes there are unexpected behaviors. For example:

If one downloads
mongodb = "2.8.0"

And then wants to download version 2.6.0, it is necessary to put the '=' sign. This is also true if it's the first time downloading.
mongodb = "=2.6.0"

If version 2.8.0 and 2.6.0 have both been downloaded, you can use the command with or without the '=' symbol to download either of the two versions specified.

PR #8883, opened 2024-06-17
Show `cargo install` for crates with binaries

Turbo87:
This PR adds `has_lib` and `bin_names` fields to the `version` Ember Data model and then uses these new fields to display `cargo install` instructions if a crate has any binaries. If the API does not expose these fields yet, or they are returned as `null` because the crate/version has not been analyzed yet, the code will default to the current behavior of assuming the crate is a library.

Related:

- https://github.com/rust-lang/crates.io/issues/5882
- https://github.com/rust-lang/crates.io/pull/8859

### DIFF
--- a/app/components/crate-sidebar/install-instructions.hbs
+++ b/app/components/crate-sidebar/install-instructions.hbs
@@ -1,0 +1,33 @@
+<p local-class="copy-help">Run the following Cargo command in your project directory:</p>
+
+{{#if (is-clipboard-supported)}}
+  <CopyButton
+    @copyText={{this.cargoAddCommand}}
+    title="Copy command to clipboard"
+    local-class="copy-button"
+  >
+    <span local-class="selectable">{{this.cargoAddCommand}}</span>
+    {{svg-jar "copy" aria-hidden="true" local-class="copy-icon"}}
+  </CopyButton>
+{{else}}
+  <code local-class="copy-fallback">
+    {{this.cargoAddCommand}}
+  </code>
+{{/if}}
+
+<p local-class="copy-help">Or add the following line to your Cargo.toml:</p>
+
+{{#if (is-clipboard-supported)}}
+  <CopyButton
+    @copyText={{this.tomlSnippet}}
+    title="Copy Cargo.toml snippet to clipboard"
+    local-class="copy-button"
+  >
+    <span local-class="selectable">{{this.tomlSnippet}}</span>
+    {{svg-jar "copy" aria-hidden="true" local-class="copy-icon"}}
+  </CopyButton>
+{{else}}
+  <code local-class="copy-fallback">
+    {{this.tomlSnippet}}
+  </code>
+{{/if}}
--- a/app/components/crate-sidebar/install-instructions.hbs
+++ b/app/components/crate-sidebar/install-instructions.hbs
@@ -1,33 +1,82 @@
-<p local-class="copy-help">Run the following Cargo command in your project directory:</p>
+{{#if @binNames}}
+  {{#if (is-clipboard-supported)}}
+    <CopyButton
+      @copyText={{this.cargoInstallCommand}}
+      title="Copy command to clipboard"
+      local-class="copy-button"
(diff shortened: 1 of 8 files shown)


app/components/crate-sidebar/install-instructions.js
```diff
@@ -0,0 +1,21 @@
...
+    return this.args.exactVersion
+      ? `cargo install ${this.args.crate}@${this.args.version}`
+      : `cargo install ${this.args.crate}`;
+  }
+
+  get cargoAddCommand() {
+    return this.args.exactVersion
+      ? `cargo add ${this.args.crate}@=${this.args.version}`
+      : `cargo add ${this.args.crate}`;
+  }
```
hi-rustin, line 14: May I ask why we need to use `cargo add crate-name@=x.x.x` to install the library and `cargo install crate-name@x.x.x` to install the binary? Why do we need to use `=` for adding a library?

hi-rustin (commented):
Thanks!

app/components/crate-sidebar/install-instructions.js
Turbo87: my understanding is that for `cargo add` the version number includes an implicit `^` constraint, while with `cargo install` my local testing showed that an implicit `=` constraint was used.

eth3lbert: For library, in some cases, it's necessary to add `@=` to install a specific version, as initially reported in issue #7884.
On the other hand, it seems that binary is always installed with the specified version.

hi-rustin: Oh, I see. I hadn't noticed that. Thank you!

app/components/crate-sidebar/install-instructions.hbs
```diff
@@ -0,0 +1,76 @@
...
+      {{this.cargoInstallCommand}}
+    </code>
+  {{/if}}
+
+  <p local-class="copy-help">
+    Running the above command will globally install the following binaries:
+    {{#if (eq @binNames.length 1)}}
+      <span local-class="bin-name">{{get @binNames 0}}</span>
+    {{else if (eq @binNames.length 2)}}
+      <span local-class="bin-name">{{get @binNames 0}}</span> and <span local-class="bin-name">{{get @binNames 1}}</span>
```
LawnGnome, line 22:
```suggestion
      <span local-class="bin-name">{{get @binNames 0}}</span>, and <span local-class="bin-name">{{get @binNames 1}}</span>
```

_insert link to Vampire Weekend's Oxford Comma song here_

app/components/crate-sidebar/install-instructions.hbs
```diff
@@ -0,0 +1,76 @@
...
+      {{svg-jar "copy" aria-hidden="true" local-class="copy-icon"}}
+    </CopyButton>
+  {{else}}
+    <code local-class="copy-fallback">
+      {{this.cargoInstallCommand}}
+    </code>
+  {{/if}}
+
+  <p local-class="copy-help">
+    Running the above command will globally install the following binaries:
```
LawnGnome, line 18: I hate saying this, but should we make this singular/plural based on the number of binaries? (I guess if this looks like a list enough maybe it doesn't matter.)

LawnGnome (approved):

app/components/crate-sidebar/install-instructions.hbs
Turbo87: I guess in that case we might as well remove the `length == 2` special case 😂

Turbo87: nevermind, https://en.wikipedia.org/wiki/Serial_comma says "in a series of three or more terms" 🙈

app/components/crate-sidebar/install-instructions.hbs
Turbo87: I've moved the sentence into the condition blocks and special cased it for the 1- and 2-item cases

PR #8911, opened 2024-06-24
Support `c`/`xml`/`rs` languages for syntax highlighting in codeblocks

marc2332:
Closes https://github.com/rust-lang/crates.io/issues/8912

## Changes
Adds the `c`/`xml`/`rs` languages to `crates_io_markdown` allowed list of languages, so then this [function](https://github.com/rust-lang/crates.io/blob/11731460ef7e00b3c8cb3b0f7cfeb351b13226c6/app/modifiers/highlight-syntax.js#L50) passes correctly.


marc2332:
I haven't tried running the backend myself, but the tests seem to pass at least, would be great if somebody could check whether it actually ends up working in the frontend or not 😄

Turbo87 (approved):
thanks! :)

Turbo87:
btw once this is deployed it will only be enabled for newly uploaded readmes unless we rerender all existing ones, but that is a lengthy and error-prone task so we usually don't bother :)

PR #8928, opened 2024-06-25
markdown: Remove hidden lines in Rust code blocks

### LINKED ISSUE #8325: README example rendering doesn't hide lines
### Current Behavior

I just looked at [the crates.io page for my crate, `service-skeleton`](https://crates.io/crates/service-skeleton), and noticed that the lines that I *thought* I'd hidden (using [the standard rustdoc mechanism](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example)) are showing up in the rendering of the README (such as the line in the first example, under the "Usage" heading, that reads `// Yes, this is cheating` and the line before/after).

### Expected Behavior

The lines that I'm trying to hide wouldn't show up in the README rendering on crates.io.

### Steps To Reproduce

1. Go to https://crates.io/crates/service-skeleton
2. Scroll down to "Usage"
3. Look at the first example
4. Observe that there are lines starting with `# ` visible.

### Environment

_No response_

### Anything else?

It's possible I've misunderstood how doctests work, in which case, my apologies, and I'd appreciate a pointer towards what I *should* be doing instead.

Turbo87:
This PR changes our markdown rendered to treat `# `-prefixed lines in Rust code blocks as hidden, similar to how rustdoc handles them (see https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example).

This allows authors to use `#![doc = include_str!("../README.md")]` in combination with hidden lines. These lines will obviously still be rendered by the GitHub markdown renderer, but that tradeoff is up to the crate author to make.

Resolves https://github.com/rust-lang/crates.io/issues/8325

### DIFF
--- a/crates/crates_io_markdown/lib.rs
+++ b/crates/crates_io_markdown/lib.rs
@@ -101,6 +101,25 @@ impl<'a> MarkdownRenderer<'a> {
                 if let Some((before_comma, _)) = orig_annot.split_once(',') {
                     ncb.info = before_comma.to_string();
                 }
+
+                // Remove hidden lines in Rust code blocks.
+                // (see https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example)
+                if ncb.info == "rust" || ncb.info == "rs" {
+                    ncb.literal = ncb
+                        .literal
+                        .split('\n')
+                        .filter_map(|line| {
+                            if line.starts_with("# ") {
+                                None
+                            } else if line.starts_with("##") {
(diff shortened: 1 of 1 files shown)


LawnGnome (approved):
Approved to the extent that this seems to work fine, but I don't know how I feel about this in general: see https://github.com/rust-lang/crates.io/issues/8325#issuecomment-2189962292 for more context.

Turbo87:
see https://github.com/rust-lang/crates.io/issues/8325#issuecomment-2227967243

Issue #1765, opened 2019-06-14
"the following are not valid category slugs and were ignored" should be error instead of warning


vi:
Category slugs are hard find find if you are not looking specifially for them and hard to get right if you trying to type `categories` field in Cargo.toml manually.

Invalid category warning may be fixed only for next version of a crate, since publishing has already happened.

I think this warning better to be just error instead.

> I don't know why invalid categories are warnings instead of errors.

(from https://github.com/rust-lang/cargo/issues/7035)

PR #8930, opened 2024-06-25
publish: Change "unknown categories" warning to error

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/1765


src/controllers/krate/publish.rs
```diff
@@ -384,7 +384,12 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
 
             // Update all categories for this crate, collecting any invalid categories
             // in order to be able to warn about them
```
LawnGnome, line 386:
```suggestion
            // in order to be able to return an error to the user.
```

LawnGnome (approved):

Issue #2685, opened 2020-08-11
Add links information to the crates.io api response


slugclub:
The index files in rust-lang/crates.io-index have a links field which contains the name of the native library a package links with. However, this information is not exposed via the crates.io API.

For example, the bcc-sys package has a links field in its [index file](https://github.com/rust-lang/crates.io-index/blob/16b52a336eb837856b5259bc54c9a502a770df96/bc/c-/bcc-sys) but that information is not available via the [api](https://crates.io/api/v1/crates/bcc-sys).

Can you please add it?

Turbo87:
for future reference, this got implemented in https://github.com/rust-lang/crates.io/pull/1248, but was only added to the index there, but not to our database. we would need to add a new column to the versions table and save this data in the database too. we might also have to build a small script to backfill the `links` from the index into the database for already existing versions.

Turbo87:
The `links` column in the database exists at this point and has already been backfilled.

The remaining issue is that the `version` model of the API already contains a `links` property with a bunch of related URLs (see e.g. https://crates.io/api/v1/crates/libz-sys/1.1.18). Changing the meaning of the existing `links` field would be a breaking change, so we will have to come up with a different name to correspond to the `links` field in the index.

@rust-lang/crates-io any suggestions? `links_`? `Links`? `links2`? `the_real_links`? `cargo_links`?

LawnGnome:
`links_2_3_4`, but I may have listened to too much Rammstein back in the day.

That's definitely an unfortunate name clash that we should fix if there's ever a v2 of the API.

More serious suggestion: `links_to_library`?

LawnGnome:
> `lib_links`?

That works too.

eth3lbert:
> `lib_links`?

I favor this one.

PR #8945, opened 2024-06-28
Expose `versions.links` column on the API

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/2685 by adding a `lib_links` field to the version API responses. This field corresponds to the `versions.links` database column and the `links` field in the index.

As mentioned in the linked issue the name `links` is unfortunately already taken and changing the content would be a breaking change, which we are trying to avoid.


hi-rustin (approved):
I am not sure do we need to add a case that contains a real `lib_links`. But I guess it is OK.

LawnGnome (approved):

Turbo87:
> I am not sure do we need to add a case that contains a real `lib_links`.

good point, I've added a test for it :)

PR #8791, opened 2024-06-04
Add a crates.io-specific security page

carols10cents:
The security response WG feels the ecosystem help communication is better suited to a crates.io-specific security page, so this creates one with that information and some other information that is relevant to crates.io but not really relevant on the Rust security page.

This PR:

- Creates crates.io/security
- Moves some security-related content from crates.io/policies to this page (but leaves the header with the ID and a link to the security page to avoid breaking any links to that heading)
- Changes the "Security" link in the footer to go to this new page

Once this PR is merged and deployed, follow-up work will include:

- Changing the security link in the README in addition to the existing link to the Rust security page (not changed in this PR to avoid having a broken link in the window between when this PR is merged and when it is deployed)
- Linking to this page from https://www.rust-lang.org/policies/security
- Linking to the ecosystem help section in particular in places that crate authors might see it when they need it, such as:
  - A Foundation blog post announcing this notice and recommending crate authors share it
  - https://doc.rust-lang.org/cargo/
  - https://crates.io/.well-known/security.txt
  - https://rustsec.org/
  - TWiR

Background references:

- [Original PR to rust-lang.org](https://github.com/rust-lang/www.rust-lang.org/pull/1975)
- [Revert of that PR](https://github.com/rust-lang/www.rust-lang.org/pull/1986)
- [Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/Security.20help.20for.20crate.20authors)

PR #9048, opened 2024-07-09
Add `.well-known/security.txt` file

carols10cents:
Follow-up work to #8791 now that the page is live!

### DIFF
--- a/public/.well-known/security.txt
+++ b/public/.well-known/security.txt
@@ -1,0 +1,2 @@
+Contact: https://crates.io/policies/security
+Expires: 2025-06-25T00:00:00.000Z
--- a/src/middleware/common_headers.rs
+++ b/src/middleware/common_headers.rs
@@ -23,11 +23,12 @@ pub async fn add_common_headers(
 
     let path = request.uri().path();
 
-    const STATIC_FILES: [&str; 4] = [
+    const STATIC_FILES: [&str; 5] = [
         "/github-redirect.html",
         "/favicon.ico",
         "/robots.txt",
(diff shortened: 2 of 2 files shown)


carols10cents:
uhhhh the tests are failing because linking with cc failed because there's no space left..... does a cache somewhere need cleared?

Turbo87 (approved):

public/.well-known/security.txt
```diff
@@ -0,0 +1,2 @@
+Contact: https://crates.io/policies/security
+Expires: 2025-06-25T00:00:00.000Z
```
LawnGnome, line 2: We're 100% going to forget this, aren't we?

I wonder if we should have a scheduled GitHub action that monitors this and tells us if it's close. (I can write that after this gets merged.)

LawnGnome (approved):

LawnGnome:
> uhhhh the tests are failing because linking with cc failed because there's no space left..... does a cache somewhere need cleared?

(At least) #9030 also failed for the same reason. Weird.

I'll open an issue and see what I can figure out. (Decidedly not a CI or GitHub Actions expert over here.)

public/.well-known/security.txt
Turbo87: hmm, that's a good point. I guess alternatively the file is small enough that we could also generate it on startup and serve it from the axum server 🤔

carols10cents: well the point is that a person should look at it every once in a while.........

jtgeibel (approved):

Issue #9050, opened 2024-07-10
Backend test GitHub Action failing


LawnGnome:
From https://github.com/rust-lang/crates.io/actions/runs/9861815465/job/27280723990:

```
   Doc-tests crates_io

running 1 test
test src/sql/semver.rs - sql::semver::SemverVersion (line 9) ... FAILED

failures:

---- src/sql/semver.rs - sql::semver::SemverVersion (line 9) stdout ----
error: linking with `cc` failed: exit status: 1
  |
  = note: LC_ALL="C" PATH="/home/runner/.rustup/toolchains/1.79.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin:/snap/bin:/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin" VSLANG="1033" "cc" "-m64" (lots and lots of object files and libraries omitted) "-Wl,-Bdynamic" "-lz" "-lssl" "-lcrypto" "-lpq" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "/home/runner/.rustup/toolchains/1.79.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "/home/runner/work/crates.io/crates.io/target/debug/deps/background_worker-604f9be365c4d193" "-Wl,--gc-sections" "-pie" "-Wl,-z,relro,-z,now" "-
  = note: /usr/bin/ld: final link failed: No space left on device
          collect2: error: ld returned 1 exit status
          

error: aborting due to 1 previous error

Couldn't compile the test.

failures:
    src/sql/semver.rs - sql::semver::SemverVersion (line 9)
```

This is presumably going to block any backend PR until it's resolved.

PR #9051, opened 2024-07-10
CI debugging

LawnGnome:
Debugging #9050; ignore.


.github/workflows/ci.yml
```diff
@@ -119,22 +119,32 @@ jobs:
...
-      - run: cargo build --tests --workspace
-      - run: cargo test --workspace
+      # Create a working directory on /mnt, which is a larger temporary
+      # filesystem than /, that we can then point our later commands to.
+      - run: |
+          sudo mkdir /mnt/target
+          sudo chown $(id -u):$(id -g) /mnt/target
+
+      - run: cargo build --tests --workspace --target-dir=/mnt/target
+      - run: cargo test --workspace --target-dir=/mnt/target
```
Turbo87, line 139: what do you think about setting `CARGO_TARGET_DIR` instead of using CLI args?

Turbo87: I've switched it over to using the env var. Also cut down on the duplication a little bit :)

LawnGnome: Yeah, that's almost certainly better. Thanks.

Issue #9052, opened 2024-07-10
Spurious Percy test failures


LawnGnome:
Many recent Percy builds seem to have failed due to it not loading the Fira fonts that crates.io uses, which causes visual changes when the fallback fonts are used.

Example: https://percy.io/b1b3bf45/crates.io/builds/35231632/

Turbo87:
this is caused by us currently running the original frontend test suite, and the playwright test suite in parallel. these both share the same Percy snapshot names, and one loads the webfont, the other doesn't.

tl;dr it's a race condition 🙈

eth3lbert:
It might be better to place it in a separate Percy folder (project?) if possible. Would that be a valid option for us?

Turbo87:
hmm, I think there is a 1-to-1 correlation between Percy projects and GitHub projects, so making it a dedicated project would be difficult, but I guess we could prefix/suffix the playwright snapshots accordingly?

eth3lbert:
> but I guess we could prefix/suffix the playwright snapshots accordingly?

Oh, this should also work.

Then, I believe we only need to adjust the title here https://github.com/rust-lang/crates.io/blob/cb34f1974ffab39e4dde0e0871769dba1886a33c/e2e/fixtures/percy.ts#L13-L18, which I intentionally implemented with the same name as the `@percy/ember` one for consistency.

Turbo87:
it looks like https://github.com/rust-lang/crates.io/pull/9054 has fixed this :)

PR #9054, opened 2024-07-11
Separate e2e snapshots and enable parallel test suites for percy

eth3lbert:
This PR addresses #9052 by:

- Separates the e2e percy snapshots from the QUnit tests by prefixing them with `e2e` in their titles.
- Enables parallel test suites for percy in CI.

---

Changing only the title to separate snapshot sets **without** enabling parallel test suites is insufficient.  This will cause  finalized builds to not recognize each other and consider others **missing**.  Therefore, enabling parallel test suites for percy is necessary.

Thanks @Turbo87 for the information and guidance!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,8 @@ jobs:
   percy-nonce:
     name: Frontend / Percy Nonce
     runs-on: ubuntu-latest
+    needs: changed-files
+    if: needs.changed-files.outputs.non-rust == 'true'
 
     # persist job results to other jobs in the workflow
     outputs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,17 @@ jobs:
       non-rust: ${{ steps.changed-files-non-rust.outputs.any_modified }}
       rust-lockfile: ${{ steps.changed-files-rust-lockfile.outputs.any_modified }}
 
+  percy-nonce:
+    runs-on: ubuntu-latest
+    # persist job results to other jobs in the workflow
+    outputs:
+      nonce: ${{ steps.percy-nonce.outputs.nonce }}
+    steps:
+      # persist step results to other steps in the job
+      - id: percy-nonce
+        # adding a timestamp makes the nonce more unique for re-runs
+        run: echo "nonce=${{ github.run_id }}-$(date +%s)" >> $GITHUB_OUTPUT
+
   backend-lint:
     name: Backend / Lint
     runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


Turbo87:
hmmmm....... it looks like the previous snapshots are now shown as "Missing" 😅 

I'm wondering if something like https://www.browserstack.com/docs/percy/integrate/parallel-test-suites might help

eth3lbert:
> hmmmm....... it looks like the previous snapshots are now shown as "Missing" 😅

Yes, that's exactly what I have a question about 😅 .  I originally thought there would be two sets of snapshots, but I can only see the ones with the "e2e" prefix.

> I'm wondering if something like https://www.browserstack.com/docs/percy/integrate/parallel-test-suites might help

I'll take a look to this.

eth3lbert:
I noticed a deprecation warning:
https://github.com/rust-lang/crates.io/actions/runs/9891816711/job/27323060801#step:2:11
```
 Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

I will make a change to address this later.

eth3lbert:
Here are some notes based on the https://www.browserstack.com/docs/percy/integrate/parallel-test-suites:

- The `NONCE` (`PERCY_PARALLEL_NONCE`) needs to be set manually because it could fail in GitHub CI
  - Some CI providers will reuse the same value we are setting the `NONCE` to on rebuild. For example, GitHub actions reuse the workflows `GITHUB_RUN_ID`. This will cause the Percy build to fail with an error saying you can’t add new snapshots to an already finalized build (this is because the `NONCE` matches an existing build and all `NONCE`s need to be unique).
  - To work around this problem, you will need to create your own `NONCE` or find a `NONCE` that your CI provider gives which is unique for each build.
- The total value needs to be set to the number of finalized builds, otherwise the process will get stuck on "receiving". 
  - If a build is hanging in the “receiving” state, even after tests have finished, that means the build was never fully finalized.
  - If `PERCY_PARALLEL_TOTAL` is set to a specific number, make sure there were that many Percy builds finalized in the test run. For example, if `PERCY_PARALLEL_TOTAL=4` and there were only 3 builds ran/finalized, Percy will hang in a receiving state. This is because Percy is waiting for 4 finalized builds.
  - If `PERCY_PARALLEL_TOTAL` is set to-1 (using --parallel), make sure `percy build:finalize` was called. If it was, make sure `PERCY_PARALLEL_NONCE` is the same for the tests and the finalize step.

Turbo87 (approved):
👏

PR #9064, opened 2024-07-12
Create `crates/{name}.xml` RSS feeds

Turbo87:
This creates RSS feeds per crate published at `https://static.crates.io/rss/crates/{name}.xml`. The feeds are synced with the database in a background job after every successful publish of the corresponding crate. It includes the latest 10 releases of the crate, but at least all releases within the last 24 hours.

Closes https://github.com/rust-lang/crates.io/issues/127

Related:

- https://github.com/rust-lang/crates.io/issues/127
- https://github.com/rust-lang/crates.io/pull/8908
- https://github.com/rust-lang/crates.io/pull/8986


Turbo87:
@samlh I guess that would make sense, yeah :)

LawnGnome (approved):
This is pretty well trodden ground now with the earlier RSS PRs, and this looks well in line with those. :+1:

Issue #9105, opened 2024-07-17
Github login fails with invalid state parameter


twosatsmaxi:
### Current Behavior

`{"errors":[{"detail":"invalid state parameter"}]}`

### Expected Behavior

Should be able to login with github.

### Steps To Reproduce

1. Setup cratesio with docker compose in a standalone instance running locally with local ip.
2. Create app in github and setting redirection url with base `https://192.168.29.108/`
3. Try to login.
4. Redirection works but fails with `invalid state parameter` error.

### Environment

- Browser: Arc
- OS: macOS

Turbo87:
can you debug this further? last time I tried local login it worked for me, but to be fair I wasn't running it inside a container.

alpeshvas:
> can you debug this further? last time I tried local login it worked for me, but to be fair I wasn't running it inside a container.

I have setup debug log using RUST_LOG but nothing helpful there

Turbo87:
you will probably have to look at the devtools/console of the browser and insert a couple of additional logging instructions to see where the state parameter is coming from and how it's flowing through the front- and backend.

twosatsmaxi:
so there is api being called with the endpoint - `http://192.168.29.108:4200/api/private/session/authorize?code=CODEEEEE&state=XXXXXX`
and it returns 
```
{"errors":[{"detail":"invalid state parameter"}]}
```

twosatsmaxi:
I'll fork the repo and docker build with more logs in backend to see the root cause. But do you think this is because the IP is not localhost but some private local ip?

Turbo87:
> But do you think this is because the IP is not localhost but some private local ip?

I would expect this to not be a problem.

I see that you are using port 4200, which is the port that `ember serve` uses. If I remember correctly you will have to run the API server and use port 8888 in the browser to make this work properly. you might also have to include the port in the base URL at GitHub.

twosatsmaxi:
but doc says 4200 
<img width="997" alt="image" src="https://github.com/user-attachments/assets/f874bb57-a278-4fac-bc3c-4317eed6a788">
https://github.com/rust-lang/crates.io/blob/main/docs/CONTRIBUTING.md#using-your-local-cratesio-with-cargo

twosatsmaxi:
<img width="566" alt="image" src="https://github.com/user-attachments/assets/a8442147-2417-4781-88ba-88a5089cd1de">
got this in browser console

twosatsmaxi:
[...]

PR #9109, opened 2024-07-17
fix local server

twosatsmaxi:


Turbo87:
as I mentioned in https://github.com/rust-lang/crates.io/issues/9105#issuecomment-2233317343, I'm not a fan of introducing environment variables that lower the security of our system, especially when these are not needed in production or on localhost. since this PR is only needed when running the crates.io codebase in a local network without HTTPS I would lean towards not adding this environment variable.

I'll leave this PR open a couple of days for others on the @rust-lang/crates-io team to chime in though.

Issue #9093, opened 2024-07-15
Support `<source media>` in Readme.md


benruijl:
### Current Behavior

Crates.io support dark mode, but does not seem to support parsing the following Markdown/HTML for toggling images to dark mode in `Readme.md`:

```
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://symbolica.io/resources/demo.dark.svg">
  <source media="(prefers-color-scheme: light)" srcset="https://symbolica.io/resources/demo.light.svg">
  <img width="600" alt="A demo of Symbolica" src="https://symbolica.io/resources/demo.dark.svg">
</picture>
```

Github does support this.

As a result, the Readme.md looks bad in dark mode.

### Expected Behavior

I expect that the `source media` tags get parsed as happens on Github, to make `crates.io` render the Readme.md nicely in dark mode.

### Environment

- Browser: any
- OS: any

benruijl:
Yes, 

https://crates.io/crates/symbolica

versus

https://github.com/benruijl/symbolica

Turbo87:
thanks! https://github.com/rust-lang/crates.io/pull/9113 should hopefully solve this :)

Turbo87:
once that PR is deployed you will have to release a new version, which causes the readme to rerender and then the `picture` and `source` elements will no longer be removed from the rendered HTML.

Turbo87:
@benruijl it's deployed to production now :)

PR #9113, opened 2024-07-18
Add support for pictures that react to the user's color scheme choice

Turbo87:
Resolves https://github.com/rust-lang/crates.io/issues/9093

GitHub allows this via e.g.:

```html
<picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://test.crates.io/logo_dark.svg">
    <img src="https://test.crates.io/logo.svg" alt="logo" width="200">
</picture>
```

but so far we haven't allowed the `picture` and `source` elements when rendering markdown files. This PR changes that.

This PR also adjusts our frontend code to make these `picture` elements work properly. Natively, these elements only react to the device's color scheme choice, but not to any overrides on the website itself. The `color-scheme` CSS attribute unfortunately does not influence the `media` resolution of the `source` elements. https://larsmagnus.co/blog/how-to-make-images-react-to-light-and-dark-mode presents a relatively straight-forward way to make this work with theme switchers, which I've adapted to make it work with our Ember.js frontend code.

### DIFF
--- a/app/components/rendered-html.hbs
+++ b/app/components/rendered-html.hbs
@@ -5,6 +5,7 @@
 <TextContent
   ...attributes
   {{highlight-syntax selector="pre > code:not(.language-mermaid)"}}
+  {{update-source-media this.colorScheme.resolvedScheme}}
   {{render-mermaids}}
 >
   {{html-safe @html}}
--- a/app/components/rendered-html.js
+++ b/app/components/rendered-html.js
@@ -1,0 +1,6 @@
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+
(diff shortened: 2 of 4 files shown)


LawnGnome (approved):
LGTM.

One thought: do we need to add an event listener to the modifier to handle colour scheme changes after the readme has been displayed, or is there something else in our frontend that will already re-render in that case?

Turbo87:
> do we need to add an event listener to the modifier to handle colour scheme changes after the readme has been displayed, or is there something else in our frontend that will already re-render in that case?

since we pass the color choice into the modifier it will automatically rerun when the (tracked) value changes, so we don't need to manually set up any listeners :)

Issue #8717, opened 2024-05-24
API tokens: Add "Copy" functionality


Turbo87:
As suggested in https://github.com/rust-lang/crates.io/issues/8154#issue-2142123788, we would like to have a functionality on the frontend that allows us to create a new API token based on the settings of an existing token. This would be very useful for expiring tokens that need to be recreated regularly.

The rough idea is to implement support for a query parameter on the "Create API Token" page that holds the id of the existing token. The creation form will then be prefilled with the settings from the existing token.

Once that exists we could potentially add "Copy" buttons to the existing tokens. We should then also adjust the token expiry notification emails to link directly to the "Create API Token" page with the ID of the token that is expiring.

- [x] Implement prefill support on the "Create API Token" page
- [x] Adjust URL in the expiry notification emails
- [ ] Add "Copy" button to the API token list

hi-rustin:
Thanks for helping to write it down!

hi-rustin:
Finally got some time to get it started.

PR #9122, opened 2024-07-19
Add support for regenerating the token

hi-rustin:
Resolves https://github.com/rust-lang/crates.io/issues/8717

The reason I chose "regenerate" instead of "copy" is that the latter is quite confusing. Users might think it means copying the token value.


hi-rustin:
Test locally:

https://github.com/user-attachments/assets/0bccd20a-1e92-4b41-a836-71f6736ed7d9

PR #9161, opened 2024-07-24
Use `diesel-async` database pools

Turbo87:
This PR moves our codebase from `deadpool` database pools to `deadpool` database pools... essentially... well... it's complicated.

The codebase currently uses the database pools from `deadpool_diesel`. These database pools are async, but they still return a sync database connection that can be used with regular `diesel` queries. With this PR we are migrating to using database pools from `diesel-async` in their `deadpool` variant. These are returning an async connection, which can be wrapped in an `AsyncConnectionWrapper` to make them work with sync `diesel` queries (in combination with `spawn_blocking()`).

In other words: this is just an intermediate step that allows us to use async `diesel` queries (enabled by `diesel-async`), but a lot of the code is still using sync queries with the wrapper. I've started to port over some of the simpler endpoints to use async queries, but that migration is far from complete. It should however serve as a starting point to demonstrate the end goal of removing all the `spawn_blocking()` calls.

Note that I had to introduce a `Conn` trait to abstract over the `PgConnection` and `AsyncConnectionWrapper` structs. Both implement `LoadConnection<Backend = Pg>`, but since trait aliases are not stabilized yet and I didn't want to repeat that signature everywhere I decided to extract the `Conn` trait. After that I essentially searched for `&mut PgConnection` and replaced it with `&mut impl Conn`, which is equivalent, but also allows us to pass in `AsyncConnectionWrapper` instead.

I would recommend reviewing this commit-by-commit... 😅

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1012,6 +1012,7 @@ dependencies = [
  "derive_deref",
  "dialoguer",
  "diesel",
+ "diesel-async",
  "diesel_full_text_search",
  "diesel_migrations",
  "dotenvy",
@@ -1197,6 +1198,7 @@ dependencies = [
  "crates_io_test_db",
  "deadpool-diesel",
  "diesel",
+ "diesel-async",
  "futures-util",
  "sentry-core",
  "serde",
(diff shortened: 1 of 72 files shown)


eth3lbert (approved):
I've reviewed the commits one by one with `hide whitespace` turned on, and it looks good to me! Plus, the CI passed, nice works 👍

eth3lbert:
> In other words: this is just an intermediate step that allows us to use async `diesel` queries (enabled by `diesel-async`), but a lot of the code is still using sync queries with the wrapper. I've started to port over some of the simpler endpoints to use async queries, but that migration is far from complete. It should however serve as a starting point to demonstrate the end goal of removing all the `spawn_blocking()` calls.

I also see that there's still a lot of work to do for the migration, especially for all the models.

Let me know if you need more help.

Turbo87:
yeah, it's far from finished but seemed like a good first step in hopefully the right direction. I haven't found a good way yet to incrementally migrate the rest of the code. Feel free to take a look while I'm on vacation :)

PR #9192, opened 2024-07-30
Use `postgres-native-tls` to connect to the database with TLS

Turbo87:
`diesel-async` by default does not support TLS directly. Instead it offers the `custom_setup` option which allows its users to choose their TLS stack between `native-tls`, `openssl` and `rustls`.

This change implements a `custom_setup` hook based on `postgres-native-tls`, whose dependencies are mostly already in our lockfile anyway. The code was mostly adapted from the official example code in `diesel-async`.

Before `diesel-async` we relied on the regular `diesel` connection implementation, which uses `pg-sys` under the hood, which uses `libpq`, which uses OpenSSL for TLS. In other words: this slightly changes how TLS is handled for the database connections. Specifically, on OSX the server certificate is rejected as "not trusted" due to a long validity period. On production this should not be relevant, but during local development against the staging database this can cause issues, thus I've included a code comment with a temporary workaround for these cases.


Turbo87:
I've deployed and tested this on staging and everything appears to be working with this PR

Turbo87:
obviously CI is now stalling... this fix is driving me mad 🫣

LawnGnome (approved):
I assume the reasoning behind `native-tls` instead of, say, `rustls` is to minimise how much change occurs relative to the previous behaviour?

I don't _love_ the workarounds here — essentially disabling TLS validity checks to handle an iffy macOS behaviour and CI that really doesn't have a need for TLS at all isn't ideal — but I'm happy enough to merge this for now and iterate later.

Turbo87:
> I assume the reasoning behind `native-tls` instead of, say, `rustls` is to minimise how much change occurs relative to the previous behaviour?

I tried `native-tls`, `rustls` and `openssl` and they all didn't work out-of-the-box without any issues. `native-tls` was already in our lockfile, so it felt like a good candidate to use here too. `rustls` would cause us to pull in quite a few more transitive dependencies and I wasn't sure if that's really worth it.

> I don't _love_ the workarounds here — essentially disabling TLS validity checks to handle an iffy macOS behaviour and CI that really doesn't have a need for TLS at all isn't ideal — but I'm happy enough to merge this for now and iterate later.

yeah, I don't love them either... 

though to be fair, as @weiznich pointed out on Zulip, the default `sslmode` in libpq is `prefer` and we set it to `required` on production environments. if you look at https://www.postgresql.org/docs/current/libpq-ssl.html you will realize that even `required` does not verify the validity of the server cert. in other words: the implementation here is now apparently safer than before because we are now verifying the TLS certificate, even if `sslmode` is only set to `required`, due to no longer using libpq for the connection.

PR #9206, opened 2024-08-01
Unify CDN invalidation into a single method on `Environment`

LawnGnome:
The index only uses CloudFront, so that still needs to go directly through `Environment::cloudfront()`, but there's no particular reason to have a bunch of background jobs with similar patterns for static.crates.io.


Turbo87 (approved):
yeah, it was getting a bit repetitive 😅

PR #9236, opened 2024-08-08
docs: Remove deprecated lines from ARCHITECTURE.md#Backend

albohlabs:
Deleted two lines from `ARCHITECTURE.md` because `build.rs` was removed via https://github.com/rust-lang/crates.io/pull/7422 and `.rustfmt.toml` was removed via https://github.com/rust-lang/crates.io/pull/2107.


Turbo87 (approved):
good catch, thank you!

Issue #2639, opened 2020-07-15
Send email when a token or co-owner is added


jsha:
In the spirit of #1895, it would be good to notify account holders and crate owners on actions that affect authorization status, like adding an API token to their account or adding a co-owner to a crate they own.

domodwyer:
Hey @Turbo87 I'm happy to finish this ticket off (by adding the "send an email when a co-owner is added" functionality).

Before I do, in https://github.com/rust-lang/crates.io/pull/9245#discussion_r1712873879 you said:

> we could move sending the email to a background job, but that would be yet another query. this is fine as is for now :)

Would you prefer I implement the new co-owner email as a background job? Or inline with the API call like #9245 does for the token email. If background job, I'll open a PR after to do the same for #9245 too after?

Turbo87:
> Would you prefer I implement the new co-owner email as a background job?

whatever you prefer/is easier. we don't have strong reasons yet to prefer one over the other.

PR #9245, opened 2024-08-10
Add API token creation notifications

domodwyer:
This PR will make crates.io email the user's registered email address when a new token is added to their account.

An email notification helps mitigate a sneaky 3rd party with access to the account from creating a token and using it to do nasty things later. Without the notification you'd only know about the token if you were periodically checking the token page.

Part of https://github.com/rust-lang/crates.io/issues/2639

---

* feat: token creation notification email (83c83135)
      
      When a new token is created for an account, send a notification email to
      the account owner.

### DIFF
--- a/src/controllers/token.rs
+++ b/src/controllers/token.rs
@@ -142,16 +142,15 @@ pub async fn new(app: AppState, req: BytesRequest) -> AppResult<Json<Value>> {
         )?;
 
         if let Some(recipient) = recipient {
+            let email = NewTokenEmail {
+                user_name: &user.gh_login,
+                domain: &app.emails.domain,
+            };
+
             // At this point the token has been created so failing to send the
             // email should not cause an error response to be returned to the
             // caller.
-            let email_ret = app.emails.send(
-                &recipient,
-                NewTokenEmail {
-                    user_name: &user.gh_login,
-                    domain: &app.emails.domain,
-                },
-            );
+            let email_ret = app.emails.send(&recipient, email);
             if let Err(e) = email_ret {
                 error!("Failed to send token creation email: {e}")
             }
--- a/src/controllers/token.rs
+++ b/src/controllers/token.rs
@@ -130,6 +130,8 @@ pub async fn new(app: AppState, req: BytesRequest) -> AppResult<Json<Value>> {
             .transpose()
             .map_err(|_err| bad_request("invalid endpoint scope"))?;
 
+        let recipient = user.email(conn)?;
+
         let api_token = ApiToken::insert_with_scopes(
             conn,
             user.id,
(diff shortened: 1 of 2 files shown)


src/controllers/token.rs
```diff
@@ -130,6 +130,8 @@ pub async fn new(app: AppState, req: BytesRequest) -> AppResult<Json<Value>> {
             .transpose()
             .map_err(|_err| bad_request("invalid endpoint scope"))?;
 
+        let recipient = user.email(conn)?;
```
domodwyer, line 133: This performs an extra database query to resolve the user's email address - is this the best way to get it?

src/controllers/token.rs
```diff
@@ -138,9 +140,26 @@ pub async fn new(app: AppState, req: BytesRequest) -> AppResult<Json<Value>> {
...
+            let email_ret = app.emails.send(
+                &recipient,
+                NewTokenEmail {
+                    user_name: &user.gh_login,
+                    domain: &app.emails.domain,
+                },
+            );
+            if let Err(e) = email_ret {
+                error!("Failed to send token creation email: {e}")
+            }
```
domodwyer, line 157: I figured it'd be better to not cause token creation failures if the email service is throwing errors, so I put it _after_ token creation and the API still returns the OK response it does today. I can change this to a hard error if you prefer though 🙏

src/controllers/token.rs
```diff
@@ -199,3 +218,25 @@ pub async fn revoke_current(app: AppState, req: Parts) -> AppResult<Response> {
...
+
+impl<'a> crate::email::Email for NewTokenEmail<'a> {
+    const SUBJECT: &'static str = "New API token created";
+
+    fn body(&self) -> String {
+        format!(
+            "\
+Hello {user_name}!
+
+A new API token was recently added to your crates.io account.
```
domodwyer, line 235: Should this `crates.io` reference use `{domain}` too?

Turbo87: I guess that would make sense, yeah

src/controllers/token.rs
Turbo87: yeah, that's fine. we could move sending the email to a background job, but that would be yet another query. this is fine as is for now :)

Turbo87 (commented):
looks good to me at first glance, but I'm on mobile-only due to vacation until the end of the week so a proper review will have to wait until then :)

in any case, thanks for working on this!

domodwyer:
No worries! Go and enjoy vacation - this can definitely wait 🙏

Turbo87:
I've rebased and slightly tweaked it to include the token name in the notification email, but other than that this looks great. thanks again! :)

Turbo87 (approved):

PR #9252, opened 2024-08-12
Update Renovate config per the hint in #9229.

LawnGnome:
#9229 removes comments, so we don't want that, but we should handle the actual schema change.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -58,7 +58,7 @@
             labels: ["A-backend ⚙️"],
         },
         {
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):
👍

seemed hard to do on mobile so I was waiting with that one until I got home 😄

LawnGnome:
> 👍
> 
> seemed hard to do on mobile so I was waiting with that one until I got home 😄

Please actually enjoy your holiday. :laughing:

Issue #9193, opened 2024-07-30
Support yank a package with a message


hi-rustin:
As we proposed in https://rust-lang.github.io/rust-project-goals/2024h2/yank-crates-with-a-reason.html.
We want to allow other owners to yank a package with a message.

The motivation here is that when a crate is updated to address a critical issue — such as a fix for a soundness bug or a security vulnerability—it is beneficial to yank previous versions and prompts users to upgrade with a yank reason. Additionally, if a crate is renamed or deprecated, the yank message can provide guidance on the new recommended crate or version. This ensures that users are aware of necessary updates and can maintain the security and stability of their projects. 

So in crates.io, we need to at least support passing a message to the yank API.

- [ ] Add support to the crates.io's yank API
- [ ] Add support on the browser frontend for giving a reason

Turbo87:
FWIW I'm not sure whether we should add this to the existing yank API (`DELETE /api/v1/crates/:crate_id/:version/yank` and `PUT /api/v1/crates/:crate_id/:version/unyank`). I think it might be better to implement something like `PATCH /api/v1/crates/:crate_id/:version` with `yank` and `yank_reason` fields and treat the old API as legacy.

hi-rustin:
> FWIW I'm not sure whether we should add this to the existing yank API (`DELETE /api/v1/crates/:crate_id/:version/yank` and `PUT /api/v1/crates/:crate_id/:version/unyank`).

I kind of agree with this because we don't actually delete anything; we just update the yank status.


> I think it might be better to implement something like `PATCH /api/v1/crates/:crate_id/:version` with `yank` and `yank_reason` fields and treat the old API as legacy.

I am a little worried about going this way. It means that if the other registries want to follow this API spec, they will need to introduce a new API or method as well.

Turbo87:
yes, that would be the consequence, but I don't think fixing the mistakes of the past is necessarily a bad idea in this case.

Rustin170506:
Because this API change could potentially introduce a new API to all registries, I want to summarize the current situation and specify the new API we want to add.

## Current API

1. `/api/v1/crates/:crate_id/:version/yank`: [DELETE](https://httpwg.org/specs/rfc9110.html#DELETE) method without any Body or parameters
2. `/api/v1/crates/:crate_id/:version/unyank`: [PUT](https://httpwg.org/specs/rfc9110.html#PUT) method without any Body or parameters

## New API

[...]

PR #9274, opened 2024-08-17
database: Add `versions.yank_message` column

Rustin170506:
ref https://github.com/rust-lang/crates.io/issues/9193

In this PR, I added a new column to the `versions` table.

I picked the text here. For a simple text-based message, I think the length is enough.

I am not sure if we should consider using JSONB here to accommodate the possibility of needing to store more structured information in the future.

I also marked it private, I guess before we expose the new API to users, we should keep it private.

### DIFF
--- a/migrations/2024-08-17-152408_add-yank-message/up.sql
+++ b/migrations/2024-08-17-152408_add-yank-message/up.sql
@@ -1,0 +1,3 @@
+ALTER TABLE versions ADD COLUMN yank_message TEXT;
+
+COMMENT ON COLUMN versions.yank_message IS 'message associated with a yanked version';
--- a/src/worker/jobs/dump_db/dump-db.toml
+++ b/src/worker/jobs/dump_db/dump-db.toml
@@ -233,6 +233,7 @@ links = "public"
 rust_version = "public"
 has_lib = "public"
 bin_names = "public"
+yank_message = "private"
 
 [versions_published_by.columns]
 version_id = "private"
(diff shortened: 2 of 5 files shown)


Rustin170506 (commented):

migrations/2024-08-17-152408_add-yank-message/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE versions ADD COLUMN yank_message VARCHAR(255);
```
Turbo87, line 1: https://stackoverflow.com/a/4849030/1478093 seems to suggest that using `varchar(n)` could be problematic if we want to change the limit at some point:

> varchar(n) – it's problematic to change the limit in live environment (requires exclusive lock while altering table)

https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default also seems to discourage usage of `varchar(n)` in favor of `text`.

src/worker/jobs/dump_db/dump-db.toml
```diff
@@ -233,6 +233,7 @@ links = "public"
 rust_version = "public"
 has_lib = "public"
 bin_names = "public"
+yank_message = "private"
```
Turbo87, line 236: any reason for keeping this private?

Turbo87: oops, missed this:

> I also marked it private, I guess before we expose the new API to users, we should keep it private.

😅

migrations/2024-08-17-152408_add-yank-message/up.sql
Rustin170506: Make sense. Updated.

Turbo87 (approved):

PR #9300, opened 2024-08-21
fix: bound co-owner invites per request

domodwyer:
This limits an "add crate co-owner" request to inviting at most 10 users per request.

This bounds the amount of CPU time a single API request can take, and therefore limits how long the associated database connection and `spawn_blocking()` thread are monopolised for to service that request.

There's still some oddities around using a single transaction across all invites: it's possible to construct an API request that does the work to invite 9 people, and then aborts and rolls back any database changes. This can enable someone malicious to retry requests rapidly (including sending emails as a result). I think perhaps a refactor of the controller / model split is needed to solve this properly though.

---

* fix: bound co-owner invites per request (2ce40b3fa)
      
      Bound the number of co-owner invites that can be sent in a single API
      request, effectively bounding the work done per request.
      
      This limits the number of iterations the N^2 membership test loop has to
      perform, limiting the amount of CPU time and memory needed for both it
      and the subsequent database queries per invitee.

### DIFF
--- a/src/controllers/krate/owners.rs
+++ b/src/controllers/krate/owners.rs
@@ -108,6 +108,14 @@ async fn modify_owners(
 ) -> AppResult<Json<Value>> {
     let logins = body.owners;
 
+    // Bound the number of invites processed per request to limit the cost of
+    // processing them all.
+    if logins.len() > 10 {
+        return Err(bad_request(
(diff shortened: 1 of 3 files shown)


Turbo87 (approved):
good catch, thanks :)

PR #9301, opened 2024-08-21
fix: send co-owner invite emails after txn commit 

domodwyer:
Before this PR, it was possible to cause the "add crate co-owner" API request to repeatedly send spam emails to invitees.

A request can be constructed where the first N-1 invites can be processed successfully, but the Nth purposely fails (i.e. due to the username not existing, or already being an owner of the crate). 

Because the invite API request processing happens serially within one database transaction, the database modifications made by the API handler (including the successfully generated & emailed invites) are all rolled back together. Because the state is rolled back, there's nothing stopping a malicious user from sending the request again, causing the same to reoccur, and sending more emails.

This PR pulls sending emails out of the model (which seems like it probably shouldn't be doing that) and instead moves the responsibility to the caller to send the invite emails only after the transaction commits.

This all-or-nothing behaviour can be used to burn CPU and add load to the database, but it probably needs a broader refactor to solve.

---

* fix: send co-owner invite emails after txn commit (9db6b0806)
      
      This changes the co-owner API endpoint to only send invite emails after
      all invites have been processed and committed to the database.
      
      This has the added benefit of releasing the database connection back to
      the pool sooner.

### DIFF
--- a/src/controllers/krate/owners.rs
+++ b/src/controllers/krate/owners.rs
@@ -127,7 +127,11 @@ async fn modify_owners(
 
         let user = auth.user();
 
-        conn.transaction(|conn| {
+        // The set of emails to send out after invite processing is complete and
+        // the database transaction has committed.
+        let mut emails = Vec::with_capacity(logins.len());
+
+        let comma_sep_msg = conn.transaction(|conn| {
             let krate: Crate = Crate::by_name(&crate_name)
                 .first(conn)
                 .optional()?
@@ -160,8 +164,12 @@ async fn modify_owners(
                     if owners.iter().any(login_test) {
                         return Err(bad_request(format_args!("`{login}` is already an owner")));
                     }
-                    let msg = krate.owner_add(&app, conn, user, login)?;
+                    let (msg, maybe_email) = krate.owner_add(&app, conn, user, login)?;
                     msgs.push(msg);
+
+                    if let Some(v) = maybe_email {
+                        emails.push(v);
+                    }
                 }
                 msgs.join(",")
             } else {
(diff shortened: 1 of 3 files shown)


src/models/krate.rs
```diff
@@ -371,28 +373,29 @@ impl Crate {
...
-                        }
-
-                        Ok(format!(
+                        let msg = format!(
                             "user {} has been invited to be an owner of crate {}",
                             user.gh_login, self.name
-                        ))
+                        );
+
+                        Ok((msg, email))
```
domodwyer, line 391: Instead of the model sending the invite email to the invitee, it now returns the email to the caller.

I kept the email struct generation here because querying for the user's email should probably happen within the same txn to preserve atomicity, and it was the smallest change to prevent the email spam.

src/controllers/krate/owners.rs
```diff
@@ -170,8 +178,20 @@ async fn modify_owners(
...
+        // committed.
+        for email in emails {
+            let addr = email.recipient_email_address().to_string();
+
+            if let Err(e) = app.emails.send(&addr, email) {
+                warn!("Failed to send co-owner invite email: {e}");
+            }
+        }
+
+        Ok(Json(json!({ "ok": true, "msg": comma_sep_msg })))
```
domodwyer, line 194: The controller collects these invite emails generated incrementally, and then sends them after the txn commits and the invites can no longer be rolled back.

src/tests/routes/crates/owners/add.rs
```diff
@@ -353,3 +353,37 @@ async fn test_unknown_team() {
...
+    // Add an invalid username to the end of the invite list to cause the
+    // request to fail.
+    usernames.push("bananas".to_string());
+
+    let response = token.add_named_owners("crate_name", &usernames).await;
+    assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+    assert_snapshot!(response.text(), @r#"{"errors":[{"detail":"could not find user with login `bananas`"}]}"#);
+
+    // No emails should have been sent.
+    assert_eq!(app.as_inner().emails.mails_in_memory().unwrap().len(), 0);
```
domodwyer, line 379: Without this PR, this assert would fail - the API request errors, but 9 emails would still have been sent.

Turbo87 (approved):
Thank you! 👍

PR #9306, opened 2024-08-23
styles: Fix button text color when hovering

Turbo87:
Buttons were not meant to get a green text color when the cursor hovers over them...

There might be a better fix for it, but I'm not seeing it right now. Feel free to comment if you know how to improve this 😉

### DIFF
--- a/app/styles/shared/buttons.module.css
+++ b/app/styles/shared/buttons.module.css
@@ -34,7 +34,11 @@
     cursor: pointer;
 
     &:hover, &:active, &:visited {
-        color: var(--text-color);
+        /*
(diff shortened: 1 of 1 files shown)


LawnGnome:
If we could get something like this as the output, I don't think we'd need the `!important`:

```css
a.button:hover {
  color: var(--text-color);
}
```

Unfortunately, this runs into https://github.com/sass/sass/issues/1425. Maybe `&:hover:is(a)` would work here as an additional selector alongside the existing `&:hover` set?

Turbo87:
Might work, but the button class can also be applied to real buttons, not just anchor tags 😅

PR #9307, opened 2024-08-23
Add a "Report crate" button

Turbo87:
This button is a `mailto:help@crates.io` link with a prefilled subject and body, that should make it easier for users to report crates which are violating our usage policies or are otherwise worth reporting to us.

Related:

- https://github.com/rust-lang/crates.io/discussions/7081

### Button

Bildschirmfoto 2024-08-23 um 10 44 17

### Email template

Bildschirmfoto 2024-08-23 um 10 44 36

### DIFF
--- a/app/components/crate-sidebar.hbs
+++ b/app/components/crate-sidebar.hbs
@@ -106,8 +106,8 @@
     {{/if}}
   {{/unless}}
 
-  {{#if this.playgroundLink}}
-    <div>
+  <div>
+    {{#if this.playgroundLink}}
       <a
         href={{this.playgroundLink}}
         target="_blank"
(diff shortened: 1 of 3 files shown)


app/components/crate-sidebar.hbs
```diff
@@ -128,6 +128,13 @@
           try out directly in your browser.
         </p>
       {{/unless}}
-    </div>
-  {{/if}}
+    {{/if}}
+
+    <a
+      href="mailto:help@crates.io?subject=The%20%22{{@crate.name}}%22%20crate&body=I'm%20reporting%20the%20%22{{@crate.name}}%22%20crate%20because%3A%0A%0A-%20%5B%20%5D%20it%20contains%20spam%0A-%20%5B%20%5D%20it%20is%20name-squatting%20(reserving%20a%20crate%20name%20without%20content)%0A-%20%5B%20%5D%20it%20is%20abusive%20or%20otherwise%20harmful%0A-%20%5B%20%5D%20it%20contains%20a%20vulnerability%20(please%20try%20to%20contact%20the%20crate%20author%20first)%0A-%20%5B%20%5D%20it%20is%20violating%20the%20usage%20policy%20in%20some%20other%20way%20(please%20specify%20below)%0A%0AAdditional%20details%3A%0A%0A%3Cplease%20add%20more%20information%20if%20you%20can%3E"
```
paolobarbolini, line 134: Drive-by review: this would probably work better with something along the lines of (however that's done in Ember):

```js
`mailto:help@crates.io${new URLSearchParams({
    subject: 'blabla',
    body: 'blablabla'
}).toString()}`
```

Turbo87: yeah, I had it roughly like that at first, but since the allowed characters in crate names are fairly restricted and the message is unlikely to change that often I figured it would probably be sufficient to have it in the encoded form directly and save a couple of payload bytes :)

jtgeibel (approved):

PR #9322, opened 2024-08-24
test: generate per-user email

Turbo87:
When generating a test user, derive that user's email address from the username.

This makes asserting the correct recipient emails are used when inspecting emails in tests.

@domodwyer I extracted that from one of your branches, since it seems like a useful change on its own and is also used/needed for https://github.com/rust-lang/crates.io/pull/2705 :)

### DIFF
--- a/src/tests/util/test_app.rs
+++ b/src/tests/util/test_app.rs
@@ -113,8 +113,8 @@ impl TestApp {
         f(&mut self.0.test_database.connect())
     }
 
-    /// Create a new user with a verified email address in the database and return a mock user
-    /// session
+    /// Create a new user with a verified email address in the database
+    /// (`<username>@example.com``) and return a mock user session.
     ///
     /// This method updates the database directly
     pub fn db_new_user(&self, username: &str) -> MockCookieUser {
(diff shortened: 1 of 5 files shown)


src/tests/util/test_app.rs
```diff
@@ -113,16 +113,16 @@ impl TestApp {
         f(&mut self.0.test_database.connect())
     }
 
-    /// Create a new user with a verified email address in the database and return a mock user
-    /// session
+    /// Create a new user with a verified email address in the database
+    /// (`<username>@example.com``) and return a mock user session.
```
domodwyer, line 117: Whoops

```suggestion
    /// (`<username>@example.com`) and return a mock user session.
```

domodwyer (approved):
Glad it's useful!

There's also https://github.com/rust-lang/crates.io/commit/b397f42247fb62d1dc82d7113d8f6da07e26f81a which helps validate which email is sent to which user - you can see it in action here: https://github.com/rust-lang/crates.io/commit/e97e93b7b6d2072737f53549035c9b5476a3ddf7 - maybe it helps? 🙏 

FYI I do plan on polishing up that branch and submitting it soon - I figured it was better to get the previous refactor in first!

PR #9324, opened 2024-08-24
tests: Extract `TestApp::emails()` and `TestApp::emails_snapshot()` fns

Turbo87:
This makes it slightly easier to write assertions for the emails that we are sending out. For `emails_snapshot()` the `Message-ID` and `Date` headers are automatically removed and the emails are then concatenated with a separator to work directly with the `assert_snapshot!()` macro from `insta`.

### DIFF
--- a/src/tests/snapshots/all__github_secret_scanning__github_secret_alert_revokes_token-2.snap
+++ b/src/tests/snapshots/all__github_secret_scanning__github_secret_alert_revokes_token-2.snap
@@ -1,0 +1,23 @@
+---
+source: src/tests/github_secret_scanning.rs
+expression: app.emails_snapshot()
+---
+To: foo@example.com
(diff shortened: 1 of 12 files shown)


src/tests/snapshots/all__github_secret_scanning__github_secret_alert_revokes_token-2.snap
```diff
@@ -0,0 +1,23 @@
...
+
+Please review your account at https://crates.io to confirm that no
+
+unexpected changes have been made to your settings or crates.
+
+
+
+Source type: some_source
+
+URL where the token was found: some_url
```
Turbo87, line 23: This change also brought to light that our secret scanning emails aren't quite formatted as intended 😅

PR #9329, opened 2024-08-24
style: fix broken rustdoc comment

domodwyer:
Cleans up the bad comment formatting I introduced in https://github.com/rust-lang/crates.io/pull/9322, and enables a [rustdoc lint] to try and catch repeats in the future.

[rustdoc lint]: https://doc.rust-lang.org/rustdoc/lints.html#unescaped_backticks

---

* ci: lint for broken inline code / backticks (559b51b18)
      
      Enables the rustdoc "unescaped_backticks" lint[1] at warn level.
      
      Unfortunately this only covers "prod" code, so it doesn't catch the
      actual broken backticks in the test code!
      
      [1]: https://doc.rust-lang.org/rustdoc/lints.html#unescaped_backticks

* style: remove spurious backtick (5c34d6602)
      
      This breaks the rustdoc rendering.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,9 @@
 //! implemented in the [category](category/index.html), [keyword](keyword/index.html),
 //! [krate](krate/index.html), [user](user/index.html) and [version](version/index.html) modules.
 
+// <https://doc.rust-lang.org/rustdoc/lints.html#unescaped_backticks>
+#![warn(rustdoc::unescaped_backticks)]
+
 #[cfg(test)]
 #[macro_use]
 extern crate claims;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -4,6 +4,9 @@
 //! implemented in the [category](category/index.html), [keyword](keyword/index.html),
 //! [krate](krate/index.html), [user](user/index.html) and [version](version/index.html) modules.
 
+// <https://doc.rust-lang.org/rustdoc/lints.html#unescaped_backticks>
+#![warn(rustdoc::unescaped_backticks)]
+
```
Turbo87, line 9: can we enable this via something like https://github.com/rust-lang/crates.io/blob/c847b96cc64af63f57e068d8971b606cce86a6ad/Cargo.toml#L22-L24 so that it's applied to all crates in the workspace?

domodwyer: Ah definitely! I didn't notice the workspace config :+1:

Turbo87: sweeet! thank you

Turbo87 (approved):

PR #9332, opened 2024-08-25
refactor: separate `Crate::owner_add` model concerns

domodwyer:
This PR pulls out the non-database concerns from `Crate::owner_add()`, moving them to the controller instead.

This gives the controller the responsibility of generating API responses and emails, giving it better visibility into what the model call did.

Part of https://github.com/rust-lang/crates.io/issues/2639

---

* refactor: separate Crate::owner_add model concerns (f38edb075)
      
      This separates out the concerns of interacting with the database via the
      Crate model, moving management of invites (emailing, response
      generation, what is considered success) to the controller.
      
      This will simplify future work on rust-lang/crates.io#2639 by allowing
      the controller to inspect the result of the owner_add() call to see if
      an invite was generated, or already existed without having to peek at
      strings.

* test: assert error response when re-inviting team (9c88b223a)
      
      Asserting an error response was returned when adding an existing user
      was covered, but not teams.

### DIFF
--- a/src/controllers/krate/owners.rs
+++ b/src/controllers/krate/owners.rs
@@ -1,18 +1,19 @@
 //! All routes related to managing owners of a crate
 
-use crate::app::AppState;
-use crate::auth::AuthCheck;
-use crate::models::token::EndpointScope;
+use crate::models::{krate::NewOwnerInvite, token::EndpointScope};
 use crate::models::{Crate, Owner, Rights, Team, User};
 use crate::tasks::spawn_blocking;
 use crate::util::errors::{bad_request, crate_not_found, custom, AppResult};
 use crate::views::EncodableOwner;
+use crate::{app::AppState, models::krate::OwnerAddError};
+use crate::{auth::AuthCheck, email::Email};
 use axum::extract::Path;
 use axum::Json;
 use diesel::prelude::*;
 use diesel_async::async_connection_wrapper::AsyncConnectionWrapper;
 use http::request::Parts;
 use http::StatusCode;
+use secrecy::{ExposeSecret, Secret};
 use serde_json::Value;
 use tokio::runtime::Handle;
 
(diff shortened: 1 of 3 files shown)


src/models/krate.rs
```diff
@@ -539,36 +517,37 @@ impl Crate {
...
+    User(User, Secret<String>),
 
-    /// Email body variables.
-    user_name: String,
-    domain: String,
-    crate_name: String,
-    token: SecretString,
+    /// The invitee was a [`Team`], and they were immediately added as an owner.
+    Team(Team),
 }
```
domodwyer, line 529: The model now returns a success enum that describes what type of owner was added, and if it is a user, it includes the acceptance token.

src/models/krate.rs
```diff
@@ -539,36 +517,37 @@ impl Crate {
...
+    #[error("{0}")] // AppError does not impl Error
+    AppError(BoxedAppError),
+
+    /// The requested invitee already has a pending invite.
+    ///
+    /// Note: Teams are always immediately added, so they cannot have a pending
+    /// invite to cause this error.
+    #[error("user already has pending invite")]
+    AlreadyInvited(User),
 }
```
domodwyer, line 544: Importantly for https://github.com/rust-lang/crates.io/issues/2639, the error response separates out the "user already has an invite" condition. Previously this was an opaque success (string message).

src/models/krate.rs
```diff
@@ -539,36 +517,37 @@ impl Crate {
...
-impl OwnerInviteEmail {
-    pub fn recipient_email_address(&self) -> &str {
-        &self.recipient_email_address
-    }
+/// Error results from a [`Crate::owner_add()`] model call.
+#[derive(Debug, Error)]
+pub enum OwnerAddError {
+    /// An opaque [`BoxedAppError`].
+    #[error("{0}")] // AppError does not impl Error
+    AppError(BoxedAppError),
```
domodwyer, line 536: A `BoxedAppError` doesn't seem to impl `std::error::Error` - is this intentional?

If it did, this would have been a `#[from]` and `#[error(transparent)]` but I don't think it's a big deal for this usage. It was a bit surprising though.

src/controllers/krate/owners.rs
```diff
@@ -171,11 +172,44 @@ async fn modify_owners(
...
+                        )),
+
+                        // This user has a pending invite.
+                        Err(OwnerAddError::AlreadyInvited(user)) => msgs.push(format!(
+                            "user {} already has a pending invitation to be an owner of crate {}",
+                            user.gh_login, krate.name
+                        )),
+
+                        // An opaque error occurred.
+                        Err(OwnerAddError::AppError(e)) => return Err(e),
```
domodwyer, line 212: The controller can now inspect the model call result, and generate the appropriate API response / emails.

src/models/krate.rs
```diff
@@ -354,74 +355,51 @@ impl Crate {
         Ok(users.chain(teams).collect())
     }
 
-    /// Invite `login` as an owner of this crate, returning a status message and
-    /// optionally an invite email to be sent by the caller.
+    /// Invite `login` as an owner of this crate, returning the created
+    /// [`NewOwnerInvite`].
     pub fn owner_add(
```
Turbo87, line 360: it looks like the controller is the only user of this fn. I'm wondering if it's even worth keeping it on the model at all... if we moved/inlined it completely we might not even need the new enums.

src/models/krate.rs
Turbo87: yeah, the error handling still isn't 100% clean. the problem is that there is a wildcard `From` impl between the two, so `BoxedAppError` currently can't implement `Error` itself.

src/models/krate.rs
domodwyer: I started to pull this out, but there's a decent chunk of database-specific / diesel query building bits in the "team" path:

https://github.com/domodwyer/crates.io/blob/0b0db20b529cbecdf8ff31d022ead3da2c9ecf50/src/models/krate.rs#L388-L400

I'd be inclined to keep the model fn because of it? The fn does abstract the caller from "what kind of thing are you adding" which is worthwhile too.

domodwyer:
Rebased to fix the conflicts 🙏

LawnGnome (approved):
tl;dr: this feels like an improvement to me.

@Turbo87, I do see your concern around model/controller separation here, but this feels cleaner overall. I think the key question is how much knowledge do we want the controller to have of the differences between inviting users and adding teams on a database level — it's easy to envisage a more fine-grained set of model methods: one to create the user invite, one to insert the team owner, but does that push too much logic up to the caller?

Basically, it's a design question. I'm fine with this change, personally — it's better than before, IMO. But it really comes down to what we think should be in models versus controllers: my own tastes tend more towards less logic in the model layer, but I'm not the team lead. :slightly_smiling_face:

Turbo87:
> my own tastes tend more towards less logic in the model layer, but I'm not the team lead.

I 💯% agree :)

IMHO the "model" layer is supposed to be a data-access layer, but it shouldn't encode significant business logic

domodwyer:
Yeah I agree, but currently figuring out if an invite is for a team or a user is weirdly deep in the model layer (nested deeper beyond the code changed in this PR) so creating the necessary database entries for that invite is all kinda mixed down in the various model bits at the moment.

This change is intended to unblock https://github.com/rust-lang/crates.io/issues/2639 - I think further work on splitting out logic from the models would be nicer all round - I might have a go after.

PR #9341, opened 2024-08-26
Implement publish notification emails

Turbo87:
I started out rebasing https://github.com/rust-lang/crates.io/pull/2705, then updated it to the current code structure, and moved all of the email sending to a background job.

In short: when a new crate version is published, a corresponding `send_publish_notifications` background job containing the `version_id` will be queued. that background job will query the relevant publish information (crate name, version, publish time, and publisher name) from the database, will get the email addresses of all current owners of the crate, and then send publish notifications to each of these owners. If sending all notification fails we assume a network issue and retry the job. If only a subset fails we assume something else is wrong and won't retry.


src/worker/jobs/send_publish_notifications.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        let publish_time = publish_details
+            .publish_time
+            .and_utc()
+            .to_rfc3339_opts(SecondsFormat::Secs, true);
+
+        // Find names and email addresses of all crate owners
+        let recipients = crate_owners::table
+            .filter(crate_owners::deleted.eq(false))
+            .filter(crate_owners::owner_kind.eq(OwnerKind::User))
```
LawnGnome, line 45: Should we also be looking up team owners?

LawnGnome (approved):
No concerns on the code, but one question about whether teams are being accounted for here.

I predict the next feature request we get will be the option to turn this off. :laughing:

src/worker/jobs/send_publish_notifications.rs
Turbo87: we should, I'm just not sure that we can... :D

our publishing code currently performs a "get all owner teams, iterate through teams and query GitHub API to see if publisher is member of the team" operation, but that does not list the members of these teams. in other words: we'd need new GitHub API calls since we don't "cache" the team memberships on our side. if/when we have non-GitHub teams implemented this will probably become a lot easier.

Turbo87:
> I predict the next feature request we get will be the option to turn this off. 😆

sure, but when you look at other package registries, they also don't seem to allow you to turn this off. if you really wanted to ignore these notifications it should be easy to implement a corresponding filter in your email server/client.

src/worker/jobs/send_publish_notifications.rs
LawnGnome: Ugh, I forgot that we don't cache that. Sorry, some Sourcegraph slipped through there.

Not blocking for this PR, but I'm wondering if it's worth the effort to figure that[^that] out sooner rather than later — given that part of the motivation here (at least from what I see) is supply chain security, not sending notifications for crates that are entirely owned by teams (such as the AWS crate ecosystem) feels problematic.

I'll give this some thought.

[^that]: For clarity, I'm more thinking just some level of caching here; not so much a full blown decoupled-from-GitHub team design.

LawnGnome:
> sure, but when you look at other package registries, they also don't seem to allow you to turn this off. if you really wanted to ignore these notifications it should be easy to implement a corresponding filter in your email server/client.

Oh, I don't actually think we should provide the option. I just expect someone to ask. :sweat_smile:

src/worker/jobs/send_publish_notifications.rs
Turbo87: > I'm wondering if it's worth the effort to figure that out sooner rather than later

it then becomes a problem of cache invalidation. I think I'd personally rather see a proper teams concept in crates.io with a feature to optionally keep the team memberships in sync with some GitHub team. Pietro mentioned a while ago that this sync might be possible by having people install a GitHub app (?) in their orgs, but I haven't looked at any of this in detail yet.

PR #9342, opened 2024-08-26
email: Change `from` to `crates.io <no-reply@crates.io>`

Turbo87:
So far we had only used `no-reply@crates.io` without a `name` in the `Mailbox`. This changes the `Mailbox` instance to use the configured `domain` as the name, which means that users will see e.g. "crates.io" as the sender instead of "no-reply@crates.io".


LawnGnome (approved):
Makes sense to me. :+1:

PR #9345, opened 2024-08-27
Fix links without href being hard to read on dark theme

### LINKED ISSUE #9330: Dark Mode: Links without href are black on dark grey
### Current Behavior

Links in dark mode if they don't have an href (e.g. `[sometext](notactuallyalink)`) are hard to read due to them being black and the background is dark gray so there is very little contrast between them.
Example:
https://crates.io/crates/glycin/1.0.2
image
Source code of the readme https://gitlab.gnome.org/sophie-h/glycin/-/raw/1.0.1/README.md

### Expected Behavior

Expected there to be more contrast between the background color and the color of links.

### Steps To Reproduce

1. go to https://crates.io/crates/glycin
2. observe the links on page

### Environment

- Browser: Firefox
- OS: EndeavorOS (Arch based rolling distro)

### Anything else?

Originally I didn't realize this was caused by their readme not having valid links. Not as big of a deal as I originally thought, probably still good to fix the color on invalid links though.

paul-hansen:
Fixes #9330 

In dark mode, `color: initial` was causing it to revert to the browser default of black which is too similar to the dark gray background.

Now links without href will use `color: unset` so it acts as if the text color is not changed from previous context, which seems to be the original intent of setting `color: initial`.

### DIFF
--- a/app/styles/application.module.css
+++ b/app/styles/application.module.css
@@ -125,7 +125,7 @@ a, .link {
 
 /* Using `:not(...)` here for specificity reasons */
 a:not([href]) {
-    color: initial;
+    color: unset;
     cursor: initial;
 }
 


Turbo87 (approved):
awesome, thank you! :)

PR #9423, opened 2024-09-10
Add `PATCH /crates/:crate/:version` route

Rustin170506:
ref https://github.com/rust-lang/crates.io/issues/9193

This pull request introduces a new patch API for yanking and unyanking crates with a message.

The logic is very similar to the current yank and unyank API, but it supports setting a yank message.

- [ ] Add tests


Rustin170506 (commented):

Rustin170506 (commented):

src/controllers/version/metadata.rs
```diff
@@ -84,3 +104,145 @@ pub async fn show(
...
+                    "Cannot update yank message for a version that is not yanked",
+                ));
+            }
+        }
+        // If both yanked and yank_message are None, do nothing.
+        // This function only cares about updating the yanked state and yank message.
+        (None, None) => {}
+    }
+    Ok(())
+}
```
eth3lbert, line 178: For me, I'd prefer to simply verify and return early if it fails without making any changes.

src/controllers/version/metadata.rs
```diff
@@ -84,3 +104,145 @@ pub async fn show(
...
+    krate: &Crate,
+) -> AppResult<()> {
+    let auth = AuthCheck::default()
+        .with_endpoint_scope(EndpointScope::Yank)
+        .for_crate(&krate.name)
+        .check(req, conn)?;
+
+    state
+        .rate_limiter
+        .check_rate_limit(auth.user_id(), LimitedAction::YankUnyank, conn)?;
```
eth3lbert, line 194: If I understand correctly, the current implementation will not trigger a rate limit with invalid data. However, I believe we should check for a rate limit first, as data validation can (and should?) also be performed on the frontend before making a request. This can also prevent potential malicious requests.

src/controllers/version/metadata.rs
```diff
@@ -84,3 +104,145 @@ pub async fn show(
...
+        // No changes, return early
+        return Ok(());
+    }
+
+    diesel::update(version)
+        .set((
+            versions::yanked.eq(version.yanked),
+            versions::yank_message.eq(&version.yank_message),
+        ))
+        .execute(conn)?;
```
eth3lbert, line 235: I suggest returning early if there's nothing to change (`(None, None)`) first. Then, make a single update request with a `WHERE` filter that checks if the updating fields are distinct. Finally, examine the returned result to determine if the row was updated.

eth3lbert (commented):
I've left some suggestions before a full review.

src/controllers/version/metadata.rs
Rustin170506: Currently, I believe `update_version_yank_state` is the first thing called.

src/controllers/version/metadata.rs
Rustin170506: For the patch API, I believe we should consider other updates for the version. So we cannot simply return if the yank update is empty, even if we only have the yank stats update at the moment. I will check GitHub's API design to see what is its behavior if nothing changes.

src/controllers/version/metadata.rs
Rustin170506: Are you suggesting that we should first verify if there is no actual change?

src/controllers/version/metadata.rs
eth3lbert: Regarding the version part, I think both options are acceptable, as an invalid version string will result in a 404 page. 
However, for invalid payload, as I previously mentioned, I recommend consuming the rate limit token.

src/controllers/version/metadata.rs
eth3lbert: > I see, you are saying if the change is `(None, None)`, then we should skip this query and check at all.

Yeah!

Regarding to sql update part, you should be able to update only when the value is distinct and should also be able to determine if a row was actually updated.

src/controllers/version/metadata.rs
eth3lbert: I suggest either removing this function or modifying it to only verify the payload and return an early bad request. In either case, there is no need to modify the version.

Turbo87:
I've rebased this on top of https://github.com/rust-lang/crates.io/pull/9437 to get that first part out of the way and make the PR a bit easier to review :)

src/tests/krate/yanking.rs
```diff
@@ -220,3 +220,66 @@ async fn publish_after_yank_max_version() {
...
+    // Check initial state
+    let json = anon.show_version("patchable", "1.0.0").await;
+    assert!(!json.version.yanked);
+    assert_eq!(json.version.yank_message, None);
+
+    // Yank with message
+    token
+        .update_yank_status("patchable", "1.0.0", Some(true), Some("Yanking reason"))
+        .await
+        .good();
```
Turbo87, line 241: we might want to add assertions on the response JSON here

src/tests/krate/yanking.rs
```diff
@@ -220,3 +220,66 @@ async fn publish_after_yank_max_version() {
...
+
+    let json = anon.show_version("patchable", "1.0.0").await;
+    assert!(json.version.yanked);
+    assert_eq!(
+        json.version.yank_message,
+        Some("Yanking reason".to_string())
+    );
```
Turbo87, line 248: I'm wondering if these assertions could be simplified by using `insta` snapshots

src/controllers/version/metadata.rs
```diff
@@ -6,17 +6,37 @@
...
 use crate::views::{EncodableDependency, EncodableVersion};
+use crate::worker::jobs::{self, UpdateDefaultVersion};
 
 use super::version_and_crate;
 
+#[derive(Deserialize)]
+pub struct VersionUpdate {
+    yanked: Option<bool>,
+    yank_message: Option<String>,
+}
```
Turbo87, line 38: we should probably make the request and response JSON consistent by including the `version` wrapper. something like this:

```suggestion
#[derive(Deserialize)]
pub struct VersionUpdate {
    yanked: Option<bool>,
    yank_message: Option<String>,
}

#[derive(Deserialize)]
pub struct VersionUpdateRequest {
    version: VersionUpdate,
}
```

src/controllers/version/metadata.rs
Rustin170506: I just modified it to only handle the validation work.

src/controllers/version/metadata.rs
eth3lbert: I just noticed that this function is used in two places, I'm wondering if it would be better to have something similar to the following:

```rust
enum YankAction {
    UpdateWithReason {
        yanked: Option<bool>,
        yank_message: Option<String>,
    },
    Toggle {
        yanked: bool,
    },
}

impl YankAction {
    pub fn from_update(data: &VersionUpdate) -> Self {
        ...
    }

    fn validate_with_version(&self, version: &Version) -> AppResult<()> {
        ...
    }

    fn apply(
        &self,
        state: &AppState,
        req: &Parts,
        conn: &mut impl Conn,
        version: &mut Version,
        krate: &Crate,
    ) -> AppResult<()> {
      ...
    }
}
```

I'm not sure if this would overcomplicate things, but IMHO, with this approach, we could distinguish the operations more easily, making it easier to follow and debug. Additionally, it would provide the potential opportunity to optimize each case individually while still sharing most of the functionality.

(Note: this is just a brief example. The naming, methods, and their arguments can be further modified as needed.)

src/controllers/version/metadata.rs
```diff
@@ -84,3 +104,139 @@ pub async fn show(
...
+        (Some(yanked), None) => {
+            version.yanked = yanked;
+            version.yank_message = None;
+        }
+        (None, Some(message)) => {
+            version.yank_message = Some(message.clone());
+        }
+        // If both yanked and yank_message are None, do nothing.
+        _ => {}
+    }
```
eth3lbert, line 170: I think this can be simplified to:

```suggestion
    version.yanked = update_data.yanked.unwrap_or(version.yanked);
    version.yank_message = update_data.yank_message.clone();
```

src/tests/krate/yanking.rs
Rustin170506: Updated.

src/tests/krate/yanking.rs
Rustin170506: Updated.

src/controllers/version/metadata.rs
Rustin170506: Yes. You are right. I have changed the perform_version_yank_update function a little bit to make it more obvious.

src/controllers/version/metadata.rs
Rustin170506: From the design of GitHub's patch API, it seems they don't have this kind of wrapper. I'm not sure what the purpose of this kind of wrapper is.
See: https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#update-a-pull-request

src/controllers/version/metadata.rs
Rustin170506: > However, for invalid payload, as I previously mentioned, I recommend consuming the rate limit token.

From our current API design, it seems we prefer do it before the auth check.

https://github.com/rust-lang/crates.io/blob/5f8fbb9c920a7fc346c7acf556110963b251efd8/src/controllers/krate/publish.rs#L62

Rustin170506: > I'm not sure if this would overcomplicate things, but IMHO, with this approach, we could distinguish the operations more easily, making it easier to follow and debug.

I just simplified the API design to make all updates more obvious. Directly pass the 'yanked' and 'yank_message' to the function, instead of implicitly depending on the updated behavior of the 'version' parameter. What do you think?

src/controllers/version/metadata.rs
Turbo87: GitHub is using a different API response design in general. If you look at https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request you will see that there is also no wrapper object for the `GET` method.

If you compare that to https://crates.io/api/v1/crates/ripgrep/14.1.0, you can see that our `GET` responses are using a wrapper object. For this particular URL/endpoint this might not be relevant, but if you look at https://crates.io/api/v1/crates/ripgrep, you can see that it not only include the `crate` field with the primary object in the response, but also secondary included objects like `categories`, `keywords` and `versions`.

While GitHub may not use the wrapper object for responses, I think it's more important to stay somewhat consistent with our own API and use it here.

Rustin170506: Make sense. We should at least keep the structure the same as the response.

Rustin170506 (commented):

src/controllers/version/metadata.rs
Turbo87: > I believe we should check for a rate limit first, as data validation can (and should?) also be performed on the frontend before making a request. This can also prevent potential malicious requests.

I'm not sure whether I agree or not. It sounds reasonable at first, but our current rate limiting implementation performs SQL queries, and parsing a JSON payload should be quite a bit quicker than performing an SQL query. if we performed the rate limiting first then the API could be flooded with invalid data which would still take away a lot of SQL connections from the pool.

in any case this PR just continues with the same order that the other API endpoints are currently using so I think it's okay for now. we can probably (continue to) discuss and potentially address this in a dedicated issue/PR/Zulip thread.

src/controllers/version/metadata.rs
```diff
@@ -84,3 +110,135 @@ pub async fn show(
...
+        .rate_limiter
+        .check_rate_limit(auth.user_id(), LimitedAction::YankUnyank, conn)?;
+
+    let api_token_id = auth.api_token_id();
+    let user = auth.user();
+    let owners = krate.owners(conn)?;
+
+    if Handle::current().block_on(user.rights(state, &owners))? < Rights::Publish {
+        if user.is_admin {
+            let action = if version.yanked {
```
Turbo87, line 192: I *think* this should be using the `yanked` variable from below 🤔

Turbo87:
I rebased the branch and fixed the above two comments. I think this is good to go now :)

PR #9445, opened 2024-09-14
Fix display for text in new token page

GuillaumeGomez:
The text should not go under the select, this PR fixes it:

image

r? @Turbo87

### DIFF
--- a/app/templates/settings/tokens/new.hbs
+++ b/app/templates/settings/tokens/new.hbs
@@ -30,41 +30,45 @@
   <div local-class="form-group" data-test-expiry-group>
     {{#let (unique-id) as |id|}}
       <label for={{id}} local-class="form-group-name">Expiration</label>
-
-      <select
-        id={{id}}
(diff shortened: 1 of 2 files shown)


app/templates/settings/tokens/new.hbs
```diff
@@ -30,41 +30,45 @@
...
-        local-class="expiry-date-input"
-        data-test-expiry-date
-        {{on "input" this.resetExpiryDateValidation}}
-      />
-    {{else}}
-      <span local-class="expiry-description" data-test-expiry-description>
-        {{this.expiryDescription}}
-      </span>
-    {{/if}}
+    <div local-class="select-group">
```
GuillaumeGomez, line 35: I couldn't create the `div` inside the `let` as the `let` didn't include the following block (`input`/`span`). However, if you prefer I can instead move the next block inside this `if`.

Turbo87 (approved):
LGTM, thanks! :)

PR #9448, opened 2024-09-15
Add support for offset-based pagination to crate versions endpoint

eth3lbert:
This PR add support for the offset-based pagination to crate versions endpoint, which is a prerequisite for maintaining consistent pagination layout in the frontend.

### DIFF
--- a/src/controllers/krate/versions.rs
+++ b/src/controllers/krate/versions.rs
@@ -43,7 +43,7 @@ pub async fn versions(
             pagination = Some(
                 PaginationOptions::builder()
                     .enable_seek(true)
-                    .enable_pages(false)
+                    .enable_pages(true)
                     .gather(&req)?,
             );
         }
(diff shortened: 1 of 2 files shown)


Turbo87:
> which is a prerequisite for maintaining consistent pagination layout in the frontend

I'm not sure I understand this part. Why do we need page-based pagination?

My understanding is that page-based pagination with Postgres is generally slow and there is basically no way to make it fast when larger offsets are requested, which is why we are trying to move away from supporting that? 🤔

eth3lbert:
> I'm not sure I understand this part. Why do we need page-based pagination?

Given that we already use numeric (offset-based) pagination on the `Depends` page (tab), it seems logical to adopt the same for the `Versions` page (tab) as well. 

> My understanding is that page-based pagination with Postgres is generally slow and there is basically no way to make it fast when larger offsets are requested, which is why we are trying to move away from supporting that? 🤔

Indeed!  If we're open to using a load more fashion (seek-based) pagination on the versions page (tab), I think we can just close this PR.

Turbo87:
> If we're open to using a load more fashion (seek-based) pagination on the versions page (tab), I think we can just close this PR.

given the performance issues that Postgres has with page/offset-based pagination I don't think we really have a choice 😅

eth3lbert:
> > If we're open to using a load more fashion (seek-based) pagination on the versions page (tab), I think we can just close this PR.
> 
> given the performance issues that Postgres has with page/offset-based pagination I don't think we really have a choice 😅

Although we can somewhat simulate numeric pagination with seek-based for versions since we load all versions anyway (might need to discuss the response format further), do you expect implementing it this way? Or just simply use a "load more" button?

eth3lbert:
> given the performance issues that Postgres has with page/offset-based pagination I don't think we really have a choice 😅

I'm actually quite happy to close this PR. I'm also looking forward to transitioning to seek-based pagination. That's why I only implemented seek-based pagination instead of both originally. Feel free to close it anytime. 😸

Turbo87:
I think a load more button is probably sufficient. if we can load the first 100 (?) versions upfront then the button will rarely be visible anyway.

> since we load all versions anyway

I assume you mean we load all versions numbers without their details?

eth3lbert:
> I think a load more button is probably sufficient. if we can load the first 100 (?) versions upfront then the button will rarely be visible anyway.

👌 This make sense to me.

> I assume you mean we load all versions numbers without their details?

Yes, for the current implementation, this should be relatively easy to achieve with sorting by semver, as the seek only consists of the ID. However, some changes will be necessary to make it work while sorting by date.

PR #9473, opened 2024-09-19
search: check for null bytes before querying

### LINKED ISSUE #9444: DatabaseError: invalid byte sequence for encoding "UTF8": 0x00
Sentry Issue: [CRATES-IO-BACKEND-3YP](https://rust-lang.sentry.io/issues/5831992097/?referrer=github_integration)

```
DatabaseError: invalid byte sequence for encoding "UTF8": 0x00
```

LawnGnome:
[PostgreSQL has an unusual definition of `UTF8` that doesn't allow for null bytes][null-bytes], even though they are technically valid UTF-8 by any other definition (including Rust's `str` type). Sending a crate search request that includes a null byte will eventually result in a database error anyway, but let's save Sentry the trouble and just 400 out early.

I did a quick audit of other places that we use query parameters and don't see any other code paths where we use a string directly as a query parameter, but I won't pretend this was a super detailed survey. (Fun fact: I can actually see the power usage for my laptop increase on my fancy new charger when I hit "show references" on the `query` method in `RequestUtils`.)

Fixes #9444.

[null-bytes]: https://www.commandprompt.com/blog/null-characters-workarounds-arent-good-enough/

### DIFF
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -55,28 +55,34 @@ pub async fn search(app: AppState, req: Parts) -> AppResult<Json<Value>> {
         use seek::*;
 
         let params = req.query();
-        let option_param = |s| params.get(s).map(|v| v.as_str());
-        let sort = option_param("sort");
-        let include_yanked = option_param("include_yanked")
+        let option_param = |s| match params.get(s).map(|v| v.as_str()) {
+            Some(v) if v.contains('\0') => Err(bad_request(format!(
+                "parameter {s} cannot contain a null byte"
+            ))),
+            Some(v) => Ok(Some(v)),
+            None => Ok(None),
+        };
+        let sort = option_param("sort")?;
+        let include_yanked = option_param("include_yanked")?
             .map(|s| s == "yes")
             .unwrap_or(true);
 
         // Remove 0x00 characters from the query string because Postgres can not
         // handle them and will return an error, which would cause us to throw
         // an Internal Server Error ourselves.
-        let q_string = option_param("q").map(|q| q.replace('\u{0}', ""));
+        let q_string = option_param("q")?.map(|q| q.replace('\u{0}', ""));
 
         let filter_params = FilterParams {
             q_string: q_string.as_deref(),
             include_yanked,
-            category: option_param("category"),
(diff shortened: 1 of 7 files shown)


src/controllers/krate/search.rs
```diff
@@ -55,28 +55,34 @@ pub async fn search(app: AppState, req: Parts) -> AppResult<Json<Value>> {
...
+        let sort = option_param("sort")?;
+        let include_yanked = option_param("include_yanked")?
             .map(|s| s == "yes")
             .unwrap_or(true);
 
         // Remove 0x00 characters from the query string because Postgres can not
         // handle them and will return an error, which would cause us to throw
         // an Internal Server Error ourselves.
-        let q_string = option_param("q").map(|q| q.replace('\u{0}', ""));
+        let q_string = option_param("q")?.map(|q| q.replace('\u{0}', ""));
```
Turbo87, line 73: I guess we don't need this anymore? 😅 

```suggestion
        let q_string = option_param("q")?;
```

LawnGnome: Well, that's embarrassing.

src/controllers/krate/search.rs
```diff
@@ -55,28 +55,29 @@ pub async fn search(app: AppState, req: Parts) -> AppResult<Json<Value>> {
...
-        let option_param = |s| params.get(s).map(|v| v.as_str());
-        let sort = option_param("sort");
-        let include_yanked = option_param("include_yanked")
+        let option_param = |s| match params.get(s).map(|v| v.as_str()) {
+            Some(v) if v.contains('\0') => Err(bad_request(format!(
+                "parameter {s} cannot contain a null byte"
+            ))),
+            Some(v) => Ok(Some(v)),
+            None => Ok(None),
+        };
```
eth3lbert, line 64: Is it okay to simply write it as:
```suggestion
        let option_param = |s| match params.get(s).map(|v| v.as_str()) {
            Some(v) if v.contains('\0') => Err(bad_request(format!(
                "parameter {s} cannot contain a null byte"
            ))),
            v @ _ => Ok(v), // or just v => Ok(v)
        };
```
Or will it be more difficult to read in this way?

PR #9475, opened 2024-09-19
frontend: change logic of when doc links are shown in search results

### LINKED ISSUE #1484: Documentation link is sometimes missing
For crates without a `documentation` key in their `Cargo.toml`, `crates.io` is supposed to generate a `Documentation` link to `docs.rs`. However, this `Documentation` link is sometimes missing on search results.

Steps to reproduce:

1. Open a fresh browser instance. (For example, I'm using `chromium --user-data-dir="$(mktemp -d)"` on Arch Linux.)
2. Navigate to https://crates.io/search?q=ndarray-npy
3. Observe that there is no `Documentation` link for the `ndarray-npy` crate. (It should be there, but it is missing. That's what this issue is about.)
4. Click on the `ndarray-npy` crate link (URL: https://crates.io/crates/ndarray-npy).
5. Click the "back" button on your browser to navigate back to the previous page.
6. Observe that the `Documentation` link has now appeared.

I'm observing this behavior for crates that do not include a `documentation` key in their `Cargo.toml`. Currently, examples of this include the `ndarray-npy`, `py_literal`, `aho-corasick`, `unreachable`, and `void` crates, among many others.

LawnGnome:
The crate version page uses a fallback to handle crates that don't define `documentation` keys in their manifests: if docs.rs has documentation for that crate version, that those docs are used.

The search results do not implement the same logic, so crates that don't have `documentation` keys don't get the handy documentation link in the crate row.

This copies the logic from the version model into the crate model, using the default crate version as the version to look for on docs.rs.

I strongly suspect there is a better way to do this than copying a bunch of code, and would welcome suggestions from those more familiar with Ember.

This would eventually fix #1484.

### DIFF
--- a/app/components/crate-row.hbs
+++ b/app/components/crate-row.hbs
@@ -61,8 +61,8 @@
     {{#if @crate.homepage}}
       <li><a href="{{@crate.homepage}}">Homepage</a></li>
     {{/if}}
-    {{#if @crate.documentation}}
-      <li><a href="{{@crate.documentation}}">Documentation</a></li>
+    {{#if @crate.documentationLink}}
+      <li><a href="{{@crate.documentationLink}}">Documentation</a></li>
     {{/if}}
     {{#if @crate.repository}}
       <li><a href="{{@crate.repository}}">Repository</a></li>
--- a/app/controllers/search.js
+++ b/app/controllers/search.js
@@ -3,9 +3,10 @@ import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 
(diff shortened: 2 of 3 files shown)


epage:
Doesn't this run into the issue of hammering docs.rs mentioned in https://github.com/rust-lang/crates.io/issues/1484#issuecomment-552603395 ?

Turbo87:
yeah, this would create an N+1 query situation towards the docs.rs API. I'm not sure whether that's actually desirable...

slightly tangential: I'm not sure either whether we should really show these doc links in the search results at all. e.g. https://lib.rs/search?q=serde has a more concise view of search results which IMHO is better than ours

PR #9496, opened 2024-09-23
Report a crate with a form

### LINKED ISSUE #9478: "Report a crate" button doesn't require any information
### Current Behavior

[The current implementation of the "Report a crate" button](https://github.com/rust-lang/crates.io/pull/9307) is nice and simple, which I like, but it has led to us getting some support emails where the person reporting the crate has not checked any of the markdown checkboxes indicating why they're reporting the crate, nor have they edited the text that says `<please add more information if you can>`.

This adds more burden on processing support emails, because I have no idea why this person is reporting the crate. 

### Expected Behavior

Brainstorming on a solution, I'd love if there was a little web form that required at least one checkbox to be checked (and one would be "other", which would require entering text in a text box), and the submit button for the form would then generate the `mailto` link once the form was valid.

eth3lbert:
A report form will display in a dialog when the report button is clicked.
Users must select reasons and provide detail information (required when the "other" reason is chosen) before submitting.

Closes #9478 

Screenshots:

image

image

image

### DIFF
--- a/app/components/crate-report-form.module.css
+++ b/app/components/crate-report-form.module.css
@@ -1,0 +1,69 @@
+.report-form {
+  background-color: var(--main-bg);
+  padding: 0.5rem 1rem;
+}
+
+.form-group {
+  border: none;
+  margin: 0;
+  padding: 0;
+
+  & + & {
+    margin-top: 1rem;
+  }
+}
+
+.form-group-name {
+  composes: form-group-name from '../styles/settings/tokens/new.module.css';
+}
+
+.reasons-list {
+  composes: scopes-list from '../styles/settings/tokens/new.module.css';
+  label {
+    flex-wrap: nowrap;
+  }
+  input {
+    align-self: center;
+  }
+}
+
+.detail {
+  border: 1px solid var(--gray-border);
+  border-radius: var(--space-3xs);
+  resize: vertical;
+  width: 100%;
+
+  &.invalid {
+    background: light-dark(#fff2f2, #170808);
+    border-color: red;
+  }
+}
+
+.form-group-error {
+  composes: form-group-error from '../styles/settings/tokens/new.module.css';
+}
+
+.buttons {
+  composes: buttons from '../styles/settings/tokens/new.module.css';
+  justify-content: end;
+  gap: 2rem;
+}
+
+.button {
+  &:focus {
+    outline: 1px solid var(--bg-color-top-dark);
+    outline-offset: 2px;
+  }
+}
+
+.report-button {
+  composes: button;
+  composes: button small from '../styles/shared/buttons.module.css';
+  border-radius: var(--space-3xs);
+}
+
+.cancel-button {
+  composes: button;
+  composes: tan-button small from '../styles/shared/buttons.module.css';
+  border-radius: var(--space-3xs);
+}
--- a/app/components/button-with-dialog.hbs
+++ b/app/components/button-with-dialog.hbs
@@ -1,0 +1,19 @@
+<button
+  type="button"
+  ...attributes
+  {{on "click" (fn this.toggleDialog true)}}
+>
+  {{ @text }}
+</button>
+
+<div
+  local-class="dialog {{ if this.isOpen 'open' }}"
+  role="dialog"
+  aria-label="dialog"
+>
+  {{#if this.isOpen}}
+    <div local-class="dialog__content" data-test-dialog-content>
+      {{yield this.toggleDialog this.isOpen}}
+    </div>
+  {{/if}}
+</div>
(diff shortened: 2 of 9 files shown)


Turbo87:
sweet, thank you! I was about to start on something similar :D

looks mostly good to me, but I'm wondering about the decision to use a dialog. I think I would've used a fullscreen route like `/crates/foo/report` instead. do you see any advantages of using a dialog instead?

eth3lbert:
> looks mostly good to me, but I'm wondering about the decision to use a dialog. I think I would've used a fullscreen route like `/crates/foo/report` instead. do you see any advantages of using a dialog instead?

Actually, a route would also work and probably fits better with crates.io's design since we don't currently rely on modals UI. However, I also think this relatively simple operation would work with a dialog, which has the benefit of not interrupting the current workflow and avoids navigating to another location. However, some users might have trouble with a dialog on certain browsers(?). Ultimately, I think both are valid options, and I don't have a strong preference, I'm also Ok report in a new page if it's more suitable 😉 .

eth3lbert:
I'm also unsure what caused the UI difference in the README rendering test (I don't know how to set up local README rendering 🙈 ). This issue might also be resolved with a new page implementation.

app/components/crate-report-form.module.css
```diff
@@ -0,0 +1,69 @@
...
+  composes: buttons from '../styles/settings/tokens/new.module.css';
+  justify-content: end;
+  gap: 2rem;
+}
+
+.button {
+  &:focus {
+    outline: 1px solid var(--bg-color-top-dark);
+    outline-offset: 2px;
+  }
```
eth3lbert, line 56: I also included an outline for focused buttons, as I previously had difficulty distinguishing them visually. (IMHO, we might also want to apply this to some other buttons, especially those with gradients.)

Turbo87: yeah, the buttons never had great a11y...

we could try adding that to all of the buttons from the shared module, but that should probably be done in a dedicated PR :)

Turbo87:
> some users might have trouble with a dialog on certain browsers

I was mostly worried about the UX on mobile. it's possible to make dialogs work decently on mobile devices, but from my experience it's quite hard, which is why I tend to avoid dialogs if possible :D

the report button was somewhat inspired by the one on npmjs.com. there it links to something like https://www.npmjs.com/support?inquire=security&security-inquire=abuse-or-spam, which is also a full-page contact form.

> I'm also unsure what cauxsed the UI difference in the README rendering test

it seems that we have some sort of race condition between our Percy snapshots and the mermaid diagram rendering, though I've only noticed it on the playwright test suite so far. we might need another instruction to wait for the diagram to finish rendering before we take the snapshot.

app/components/crate-report-form.module.css
eth3lbert: That would be great! And it definitely should be done in separate PRs.

eth3lbert:
> I was mostly worried about the UX on mobile. it's possible to make dialogs work decently on mobile devices, but from my experience it's quite hard, which is why I tend to avoid dialogs if possible :D

Fair enough!

> the report button was somewhat inspired by the one on npmjs.com. there it links to something like https://www.npmjs.com/support?inquire=security&security-inquire=abuse-or-spam, which is also a full-page contact form.

Yeah, this is a great one that we can learn from. 
I guess we just want to implement something similar to the current one, but on a full page? Or do you lean toward implementing it as a general report page?

> it seems that we have some sort of race condition between our Percy snapshots and the mermaid diagram rendering, though I've only noticed it on the playwright test suite so far. we might need another instruction to wait for the diagram to finish rendering before we take the snapshot.

Oh, that's interesting. I haven't encountered this issue before. Thanks for the information.

Turbo87:
> I guess we just want to implement something similar to the current one, but on a full page? Or do you lean toward implementing it as a general report page?

I'm undecided on this one. What do you think? It's probably easiest to start with the former and incrementally evolve it into the latter?

We also still have https://github.com/rust-lang/crates.io/discussions/7081 open, which could be solved by either having another dedicated "Report User" page, or make it an option on the general report page.

eth3lbert:
> I'm undecided on this one. What do you think? It's probably easiest to start with the former and incrementally evolve it into the latter?
> 
> We also still have #7081 open, which could be solved by either having another dedicated "Report User" page, or make it an option on the general report page.

Then I lean toward a more general contact support page, starting with simple options and gradually expanding them later. 

Do we already have a list of the support services we offer? (We might need feedback from others who handle support.)

Turbo87:
> Do we already have a list of the support services we offer?

not really. we could take a look at the support requests that we've received and handled in the past, but we would probably need some sort of generic option anyway. the most common request is from author who would like to have their crates deleted for various reasons, but that should be covered by the crate deletion RFC implementation "soon".

eth3lbert:
@Turbo87 I have experimented full-page fashion, and most flows follow npm. I'll try to open a PR later sometime.

localhost_4200_support

localhost_4200_support_inquire=crate-violation

The designs and wording need further refinement in that PR.

Turbo87:
superseded by https://github.com/rust-lang/crates.io/pull/9529

PR #9529, opened 2024-09-27
Add a support page

### LINKED ISSUE #9478: "Report a crate" button doesn't require any information

eth3lbert:
This PR adds a support page that allows user to select a support type from a list and fill in required information. It also moves "Email Support" from the footer to the support page.

The support page is inspired by npm's support page[^1]. When a user clicks the "report violation" link, a report form will be displayed. The user must input a crate name, select reasons, and provide detailed information (required when "other" is chosen) before submitting. The name field in the report form should be automatically filled when navigating from the report button in the `CrateSidebar`.

Related #9496 . 
Closes #9478 .

[^1]: https://www.npmjs.com/support

### DIFF
--- a/app/components/support/crate-report-form.hbs
+++ b/app/components/support/crate-report-form.hbs
@@ -1,0 +1,89 @@
+<form
+  local-class="report-form"
+  data-testid="crate-report-form"
+  ...attributes
+  {{on "submit" (prevent-default this.submit)}}
+>
+  <h2>Report A Crate</h2>
+
+  <fieldset local-class="form-group" data-test-id="fieldset-crate">
+    {{#let (unique-id) as |id|}}
+      <label for={{id}} local-class="form-group-name">
+        Crate
+      </label>
+      <Input
+        id={{id}}
+        @type="text"
+        @value={{this.crate}}
+        autocomplete="off"
+        aria-required="true"
+        aria-invalid={{if this.crateInvalid "true" "false"}}
+        local-class="crate-input"
+        data-test-id="crate-input"
+        {{auto-focus}}
+        {{on "input" this.resetCrateValidation}}
+      />
+      {{#if this.crateInvalid}}
+        <div local-class="form-group-error" data-test-id="crate-invalid">
+          Please specify a crate.
+        </div>
+      {{/if}}
+    {{/let}}
+  </fieldset>
+
+  <fieldset local-class="form-group" data-test-id="fieldset-reasons">
+    <div local-class="form-group-name">Reasons</div>
+    <ul role="list" local-class="reasons-list {{if this.reasonsInvalid "invalid"}}">
+      {{#each this.reasons as |option|}}
+        <li>
+          <label>
+            <Input
+              @type="checkbox"
+              @checked={{this.isReasonSelected option.reason}}
+              name={{ option.reason }}
+              data-test-id="{{ option.reason }}-checkbox"
+              {{on "change" (fn this.toggleReason option.reason)}}
+            />
+            {{option.description}}
+          </label>
+        </li>
+      {{/each}}
+    </ul>
+    {{#if this.reasonsInvalid}}
+      <div local-class="form-group-error" data-test-id="reasons-invalid">
+        Please choose reasons to report.
+      </div>
+    {{/if}}
+  </fieldset>
+
+  <fieldset local-class="form-group" data-test-id="fieldset-detail">
+    {{#let (unique-id) as |id|}}
+      <label for={{id}} local-class="form-group-name">Detail</label>
+      <Textarea
+        id={{id}}
+        @value={{this.detail}}
+        local-class="detail {{if this.detailInvalid "invalid"}}"
+        aria-required={{if this.detailInvalid "true" "false" }}
+        aria-invalid={{if this.detailInvalid "true" "false"}}
+        rows="5"
+        data-test-id="detail-input"
+        {{on "input" this.resetDetailValidation}}
+      />
+      {{#if this.detailInvalid}}
+        <div local-class="form-group-error" data-test-id="detail-invalid">
+          Please provide some detail.
+        </div>
+      {{/if}}
+    {{/let}}
+  </fieldset>
+
+  <div local-class="buttons">
+    <button
+      type="submit"
+      local-class="report-button"
+      data-test-id="report-button"
+    >
+      Report
+    </button>
+  </div>
+</form>
--- a/app/components/support/crate-report-form.js
+++ b/app/components/support/crate-report-form.js
@@ -1,0 +1,105 @@
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+
+import window from 'ember-window-mock';
+
+const REASONS = [
+  {
+    reason: 'spam',
+    description: 'it contains spam',
+  },
+  {
+    reason: 'name-squatting',
+    description: 'it is name-squatting (reserving a crate name without content)',
+  },
+  {
+    reason: 'abuse',
+    description: 'it is abusive or otherwise harmful',
+  },
+  {
+    reason: 'security',
+    description: 'it contains a vulnerability (please try to contact the crate author first)',
+  },
+  {
+    reason: 'other',
+    description: 'it is violating the usage policy in some other way (please specify below)',
+  },
+];
+
+export default class CrateReportForm extends Component {
+  @service store;
+
+  @tracked crate = '';
+  @tracked selectedReasons = [];
+  @tracked detail = '';
+  @tracked crateInvalid = false;
+  @tracked reasonsInvalid = false;
+  @tracked detailInvalid = false;
+
+  reasons = REASONS;
+
+  constructor() {
+    super(...arguments);
+    this.crate = this.args.crate;
+  }
+
+  validate() {
+    this.crateInvalid = !this.crate || !this.crate.trim();
+    this.reasonsInvalid = this.selectedReasons.length === 0;
+    this.detailInvalid = this.selectedReasons.includes('other') && !this.detail?.trim();
+    return !this.crateInvalid && !this.reasonsInvalid && !this.detailInvalid;
+  }
+
+  @action resetCrateValidation() {
+    this.crateInvalid = false;
+  }
+
+  @action resetDetailValidation() {
+    this.detailInvalid = false;
+  }
+
+  @action isReasonSelected(reason) {
+    return this.selectedReasons.includes(reason);
+  }
+
+  @action toggleReason(reason) {
+    this.selectedReasons = this.selectedReasons.includes(reason)
+      ? this.selectedReasons.filter(it => it !== reason)
+      : [...this.selectedReasons, reason];
+    this.reasonsInvalid = false;
+  }
+
+  @action
+  submit() {
+    if (!this.validate()) {
+      return;
+    }
+
+    let mailto = this.composeMail();
+    window.open(mailto, '_self');
+  }
+
+  composeMail() {
+    let crate = this.crate;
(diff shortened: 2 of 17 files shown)


eth3lbert:
@Turbo87 I've only created the support route for now, and haven't placed it anywhere yet. I believe we want to replace "Email Support" under the "Get Help" section in the footer with this. I'm unsure whether we should simply add a "Get help" mailto link to the categorized support list or differentiate it with a horizontal separator and a subheader like "If you couldn't find a category above." followed by the "Get help" mailto link.

app/components/support/crate-report-form.hbs
```diff
@@ -0,0 +1,94 @@
...
+  {{on "submit" (prevent-default this.submit)}}
+>
+  <h2>Report A Crate</h2>
+
+  <fieldset local-class="form-group" data-testid="fieldset-name">
+    {{#let (unique-id) as |id|}}
+      <label for={{id}} local-class="form-group-name">
+        Crate name
+        {{#if this.loadCrateTask.isRunning}}
+          <LoadingSpinner />
```
eth3lbert, line 14: This looks poor in dark mode. We should add dark mode support in a separate PR.

app/components/support/crate-report-form.hbs
```diff
@@ -0,0 +1,94 @@
...
+        local-class="name-input"
+        data-testid="name-input"
+        {{auto-focus}}
+        {{on "input" this.resetNameValidation}}
+      />
+      {{#if this.nameInvalid}}
+        <div local-class="form-group-error" data-testid="name-invalid">
+          Please specify an existing crate.
+        </div>
+      {{/if}}
```
Turbo87, line 34: since we ultimately generate an email, where the user can change the name again I'm wondering if the extra name validation might be overkill here 😅

eth3lbert (commented):
I've also introduced `data-testid="some-id"` in this PR instead of the `data-test-some-id` format. `data-testid` is becoming more popular recently, which also enables `getByTestId` in playwright and the potential benefit of removing `data-testid` in production (further implementation is required) if desired. This comes with a minor drawback of being slightly more verbose in qunit tests.

app/controllers/support.js
```diff
@@ -0,0 +1,25 @@
...
+    label: 'Report a crate that violates policies',
+  },
+];
+
+const VALID_INQUIRE = new Set(SUPPORTS.map(s => s.inquire));
+
+export default class SupportController extends Controller {
+  queryParams = ['inquire', 'name'];
+
+  @tracked inquire;
```
Turbo87, line 16: since this is in a controller I think the property keeps its value across page navigations. in other words: if you press the "report crate" button the property will be set, if you navigate away from the page, then you press the support link in the footer, you will not get to the overview state, but to the "report crate" state again. I'm not 100% sure about this, but we should probably add a test to ensure that navigating to `/support` without query params resets the state.

app/components/support/crate-report-form.js
```diff
@@ -0,0 +1,135 @@
...
+    super(...arguments);
+    this.reset();
+    this.name = this.args.name;
+  }
+
+  reset() {
+    this.nameInvalid = false;
+    this.reasonsInvalid = false;
+    this.detailInvalid = false;
+  }
```
Turbo87, line 58: any reason for not setting them directly in their `@tracked` definitions above? looks like `reset()` is not called anywhere else?

app/components/support/crate-report-form.js
```diff
@@ -0,0 +1,135 @@
...
+    // Try reusing crate in the store if name is given
+    // and fallback to fetch if its not in store
+    if (!this.crate && this.args.name === this.name) {
+      let crate = this.store.peekRecord('crate', this.args.name);
+      if (crate) {
+        return crate;
+      }
+    }
+    let last = this.loadCrateTask.lastSuccessful?.value;
+    return last?.id === this.name ? last : await ajax(`/api/v1/crates/${this.name}?include=`).then(r => r.crate);
```
Turbo87, line 133: `this.store.findRecord(...)` would be easier, or am I missing something?

eth3lbert: Yes, `findRecord` definitely does the work more simply. The reason for using ajax here is that `include=` is more efficient because it doesn't query for unnecessary information.

app/components/support/crate-report-form.hbs
eth3lbert: Fair enough! It might be overkill. I implemented it primarily to prevent typos. Do you suggest simply not checking for existing names?

app/components/support/crate-report-form.js
eth3lbert: Oops, this logic is more relevant to the previous version of the dialog. I'll remove it.

Turbo87:
> the potential benefit of removing `data-testid` in production

that is already done for `data-test-*` attributes by https://github.com/mainmatter/ember-test-selectors, so using `data-testid` and not removing them from prod builds would be a (small) regression for now. we could potentially add `data-testid` support to ember-test-selectors though, if that is becoming a common pattern in other frameworks.

app/components/support/crate-report-form.js
Turbo87: we could add support for `?include=` to our Ember Data adapter though (see "Retrieving Related Model Records" in https://api.emberjs.com/ember-data/5.3/classes/Store/methods/findRecord?anchor=findRecord)

app/components/support/crate-report-form.hbs
Turbo87: yeah, I expect most users to go through the report crate button anyway, which auto-fills the field and thus shouldn't have any typos

eth3lbert:
> > the potential benefit of removing `data-testid` in production
> 
> that is already done for `data-test-*` attributes by https://github.com/mainmatter/ember-test-selectors, so using `data-testid` and not removing them from prod builds would be a (small) regression for now. we could potentially add `data-testid` support to ember-test-selectors though, if that is becoming a common pattern in other frameworks.

Thanks for pointing this out. I didn't know that before.

Yes, I suppose this is a common pattern in newer testing frameworks like testing-library, cypress, and playwright, etc. Do you suggest we stick to the `data-test-*` convention for now and introduce `data-testid` later when we implement the removal process for production?

Turbo87:
> Do you suggest we stick to the `data-test-*` convention for now and introduce `data-testid` later when we implement the removal process for production?

yeah, I think that would be best for now. I'd rather convert the whole codebase to the other pattern at once, instead of having to deal with two concepts in parallel when possible.

one thing to note, we use things like `data-test-crate="serde"` too, where the value of the test attribute is dynamic. I'm not sure how that would work with `data-testid="crate"`? I guess `data-testid="crate[serde]"` or something like that?

eth3lbert:
> one thing to note, we use things like `data-test-crate="serde"` too, where the value of the test attribute is dynamic. I'm not sure how that would work with `data-testid="crate"`? I guess `data-testid="crate[serde]"` or something like that?

Hmm, I'm not sure if I fully understand your question, but I think this can be easily achieved using a template like `data-testid="crate-{{crate.name}}"`.

app/controllers/support.js
eth3lbert: One more thing I noticed is that, if you're on the support page with query, if the support link in footer is defined as `<LinkTo @route="support" />` it will still inherit those query params. There doesn't seem to be a simple way to avoid this inheritance in the template. You need to use hash to manual overwrite all query params.

eth3lbert:
> > the potential benefit of removing `data-testid` in production
> 
> that is already done for `data-test-*` attributes by https://github.com/mainmatter/ember-test-selectors, so using `data-testid` and not removing them from prod builds would be a (small) regression for now. we could potentially add `data-testid` support to ember-test-selectors though, if that is becoming a common pattern in other frameworks.

I think I've found a possible solution. Since the `testid` attribute is configurable in Playwright https://playwright.dev/docs/locators#set-a-custom-test-id-attribute, and ember-test-selectors accepts selectors with the pattern `data-test-*`, we can use `data-test-id` as the custom `testid` attribute to satisfy both frameworks.

app/components/support/crate-report-form.hbs
eth3lbert: No longer a concern, as we do not verify the existence of the crate anymore.

app/components/support/crate-report-form.js
eth3lbert: No longer a concern, as we do not verify the existence of the crate anymore.

app/controllers/support.js
eth3lbert: I'm wondering if there's a way to test this without `<LinkTo @route="support"/>`. We already have `<LinkTo @route="support" @query={{hash inquire=null crate=null }}>Support</LinkTo>` that reset the query params by setting their values to null.

Turbo87: not sure about playwright, but in the Ember.js test suite `visit('/support')` might be sufficient as it does not reload the whole app

eth3lbert: Hmm, I just commented out `resetController` and expected the following to fail, but it didn't 😅 .
```js
  test('should not retain query params when exiting and then returning', async function (assert) {
    await visit('/support?inquire=crate-violation');
    assert.strictEqual(currentURL(), '/support?inquire=crate-violation');
    assert
      .dom('[data-test-id="support-main-content"] section')
      .exists({ count: 1 })
      .hasAttribute('data-test-id', 'crate-violation-section');

    await visit('/');
    assert.strictEqual(currentURL(), '/');
    await visit('/support');
    assert.strictEqual(currentURL(), '/support');
    assert
      .dom('[data-test-id="support-main-content"] section')
      .exists({ count: 1 })
      .hasAttribute('data-test-id', 'inquire-list-section');
  });

```
However, if there's a `<LinkTo @route="support">`, I can see that the generated URL has sticky query params.

Turbo87: interesting, maybe I was wrong then :D

Turbo87:
@eth3lbert is this PR in draft state intentionally or is it ready for review/merging?

eth3lbert:
> @eth3lbert is this PR in draft state intentionally or is it ready for review/merging?

Yes, this is _almost_ done. We just need to implement this

> I believe we want to replace "Email Support" under the "Get Help" section in the footer with this. I'm unsure whether we should simply add a "Get help" mailto link to the categorized support list or differentiate it with a horizontal separator and a subheader like "If you couldn't find a category above." followed by the "Get help" mailto link.

 The simplest and most intuitive way would be to simply place a mailto link after the `{{each}}` block in template, rather than putting it into supports. I assume that it should always appear after other supports, and when we want to add a new support, we don't need to worry about placing the new support before it. After that, I think it will be ready for review and merging.

Or do you expect this to be done in a separate PR?

Turbo87:
> Or do you expect this to be done in a separate PR?

I'm fine with either path :)

app/controllers/support.js
eth3lbert: I figured out how to test this in qunit but not playwright. I don't know how to mock the route in playwright QQ.

eth3lbert:
I just copied a screenshot from Percy for preview. :)

image

These words need to be revised. :D

Turbo87 (approved):
:shipit:

PR #9530, opened 2024-09-27
Change crates.io policy to not offer crate transfer mediation

Turbo87:
This PR implements https://github.com/rust-lang/rfcs/pull/3646. For some reason I thought this had already been implemented, but I just stumbled upon that section in the policies and was surprised that it was still in there... 😅

### DIFF
--- a/app/templates/policies/index.hbs
+++ b/app/templates/policies/index.hbs
@@ -71,10 +71,10 @@
   <p>crates.io has a first-come, first-serve policy on crate names. Upon publishing a package, the publisher will be made
     owner of the package on crates.io.</p>
 
-  <p>If you want to take over a package, we require you to first try and contact the current owner directly. If the
-    current owner agrees, they can add you as an owner of the crate, and you can then remove them, if necessary. If the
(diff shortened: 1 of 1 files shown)


app/templates/policies/index.hbs
```diff
@@ -71,10 +71,8 @@
   <p>crates.io has a first-come, first-serve policy on crate names. Upon publishing a package, the publisher will be made
     owner of the package on crates.io.</p>
 
-  <p>If you want to take over a package, we require you to first try and contact the current owner directly. If the
-    current owner agrees, they can add you as an owner of the crate, and you can then remove them, if necessary. If the
-    current owner is not reachable or has not published any contact information the crates.io team may reach out to help
-    mediate the process of the ownership transfer.</p>
+  <p>If you want to take over a package, we recommend you try and contact the current owner directly. If the
```
djc, line 74: I'd suggest making the implied explicit, something like "For security reasons, the crates.io team will not change ownership of existing usable non-malicious crates." I guess the devil is in the details of the adjectives to use here?

Issue #9352, opened 2024-08-28
Tracking Issue for "Crate Deletions"


Turbo87:
This is a tracking issue for the RFC "crates.io: Crate Deletions" (rust-lang/rfcs#3660).

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.
Discussion comments will get marked as off-topic or deleted.
Repeated discussions on the tracking issue may lead to the tracking issue getting locked.

### Steps

- [x] Implement crate deletion API endpoint (https://github.com/rust-lang/crates.io/pull/9904)
- [x] Implement crate deletion user interface (https://github.com/rust-lang/crates.io/pull/10177, https://github.com/rust-lang/crates.io/pull/10343)
- ~Implement version deletion API endpoint~
- ~Implement version deletion user interface~
- [ ] Update https://crates.io/policies to no longer mention that "Crate deletion by their owners is not possible" (https://github.com/rust-lang/crates.io/pull/10392)

### Unresolved Questions

- Should names of deleted crates be blocked so that they can't be re-used?
- Should deleted versions be blocked from being re-uploaded?
- Should we keep and mark deleted versions in the index?

carols10cents:
Two pieces of feedback now that I've tested the `/delete` route in staging:

1. The page that shows up on `/delete` says:

> Requirements:
> 
> A crate can only be deleted if:
> 
> 1. the crate has been published for less than 72 hours, or
> 2.  the crate only has a single owner, and
>         the crate has been downloaded less than 100 times for each month it has been published, and
>         the crate is not depended upon by any other crate on crates.io.

I forgot what the RFC actually decided on, and I missed that little "or" at the end of the first requirement. It would be nice if that was a bit more obvious, either put a big "OR" between 1 and 2 along the left, or maybe change "A crate can only be deleted if" to something like "A crate can only be deleted if one of the two following applies".

2. I published a new crate to test this out and I got an email when I published the crate, but I did not get an email when I deleted the crate. I think deleting should send an email too as it could be a sign of account compromise if it wasn't the owner who did it.

Turbo87:
thanks for the feedback!

I had that thought about the email as well, but somehow forgot about it again... 😅

PR #9549, opened 2024-10-01
Extract `DeleteCrateFromStorage` background job

Turbo87:
This background job can later be reused to implement https://rust-lang.github.io/rfcs/3660-crates-io-crate-deletions.html

Related:

- https://github.com/rust-lang/crates.io/issues/9352

### DIFF
--- a/src/worker/jobs/delete_crate.rs
+++ b/src/worker/jobs/delete_crate.rs
@@ -3,6 +3,7 @@ use crate::worker::Environment;
 use anyhow::Context;
 use crates_io_worker::BackgroundJob;
 use std::sync::Arc;
+use tokio::try_join;
 
 /// A background job that deletes all files associated with a crate from the storage backend.
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


src/worker/jobs/delete_crate.rs
```diff
@@ -0,0 +1,50 @@
...
+                let result = ctx.storage.delete_all_readmes(name).await;
+                result.context("Failed to delete readme files from S3")
+            },
+            async {
+                info!("{name}: Deleting RSS feed from S3…");
+                let feed_id = FeedId::Crate { name };
+                let result = ctx.storage.delete_feed(&feed_id).await;
+                result.context("Failed to delete RSS feed from S3")
+            }
+        )?;
```
Turbo87, line 45: is it sufficient for this to be in one job or should it be split into multiple jobs that can be retried individually if they fail? 🤔

eth3lbert: Yeah, what would happen if one of these failed?

Turbo87: `try_join!()` cancels the other runnings tasks if one of them fails. it then returns the error, which will make the job fail and will cause it to be retried a little later. AFAIU all of the steps in this job are idempotent though, so retrying isn't a big deal. the job does not perform CDN invalidations yet though.

eth3lbert: Let's say task 1 succeeds first, but task 2 fails, so it cancels task 3. What will the state of task 1 be after all? If it remains done, is it safe to retry?

eth3lbert: I think if every job could be retried even if it's already completed, this wouldn't be a problem.

Turbo87: > If it remains done, is it safe to retry?

yes, as far as I understand this shouldn't be an issue. I guess splitting it up is probably unnecessary over-engineering 😅

PR #9551, opened 2024-10-01
Move `features` to `feature2` that depend on features that use new feature syntax

Turbo87:
Fixes https://github.com/rust-lang/crates.io/issues/6135. Closes https://github.com/rust-lang/crates.io/pull/6168.

Compared to https://github.com/rust-lang/crates.io/pull/6168, this PR implements the fix in a way that features depending on newer syntax are recursively moved to `features2` instead of moving them all unconditionally.

Probably best reviewed commit-by-commit :)

/cc @rust-lang/cargo

### DIFF
--- a/src/models/feature.rs
+++ b/src/models/feature.rs
@@ -7,11 +7,38 @@ pub type FeaturesMap = BTreeMap<String, Vec<String>>;
 ///
 /// See <https://rust-lang.github.io/rfcs/3143-cargo-weak-namespaced-features.html>.
 pub fn split_features(features: FeaturesMap) -> (FeaturesMap, FeaturesMap) {
-    features.into_iter().partition(|(_k, vals)| {
-        !vals
-            .iter()
-            .any(|v| v.starts_with("dep:") || v.contains("?/"))
-    })
+    const ITERATION_LIMIT: usize = 100;
+
+    // First, we partition the features into two groups: those that use the new
+    // features syntax (`features2`) and those that don't (`features`).
+    let (mut features, mut features2) =
+        features
+            .into_iter()
+            .partition::<FeaturesMap, _>(|(_k, vals)| {
+                !vals
+                    .iter()
+                    .any(|v| v.starts_with("dep:") || v.contains("?/"))
+            });
+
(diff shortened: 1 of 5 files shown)


src/models/feature.rs
```diff
@@ -7,11 +7,36 @@ pub type FeaturesMap = BTreeMap<String, Vec<String>>;
...
+            });
+
+        features = split.0;
+
+        if !split.1.is_empty() {
+            features2.extend(split.1);
+        } else {
+            break;
+        }
+    }
```
Turbo87, line 37: yes, I'm a little worried about the infinite loop, but are there actually cases where this could actually get into infinite behavior? I don't think so, but a second(/third/fourth/...) pair of eyes won't hurt...

if we're worried that we may have missed something we could potentially also limit the number of iterations to a somewhat sane number roughly matching our features limit :)

Turbo87: some additional context: this function is (currently) only used by the "sync to index" background jobs. in other words: this should not be able to freeze our API server, and if the background worker freezes up we will quickly see it in our monitoring.

Turbo87: I've changed the code now to limit the number of "recursion" steps to 100 to play it safe...

eth3lbert: Do we want to add some logging when the recursion limit is reached? That way, we could examine what actually happened later.

Turbo87: good call! implemented by https://github.com/rust-lang/crates.io/pull/9580

weihanglo:
👋🏾! Is the migration of this done for all relevant crate index files.

src/models/feature.rs
weihanglo: I am also interested when the limit is reached. Where could I subscribe it? :)

Turbo87:
@weihanglo yes and no. we only triggered a manual sync for clap. the other crates will update whenever they release a new version.

src/models/feature.rs
Turbo87: unfortunately you'd need access to the crates.io logs 😅

PR #9559, opened 2024-10-02
models/token: Remove unnecessary transaction

Turbo87:
The transaction behavior was initially introduced in https://github.com/rust-lang/crates.io/commit/52862fab97dd00184a2034c4d6bffdfe241734e4 and the commit message mentions:

> Because our tests run in a single transaction, we need to wrap that update to ensure that the error doesn't poison the transaction.

but that first part is no longer true, so we don't need the transaction anymore AFAICT.

This PR also improves our test suite for this functionality a little to catch potential regressions.

### DIFF
--- a/src/models/token.rs
+++ b/src/models/token.rs
@@ -79,15 +79,13 @@ impl ApiToken {
             .filter(api_tokens::token.eq(token));
 
         // If the database is in read only mode, we can't update last_used_at.
-        // Try updating in a new transaction, if that fails, fall back to reading
-        conn.transaction(|conn| {
-            update(tokens)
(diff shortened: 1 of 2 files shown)


Turbo87:
nevermind... I now understand the purpose of this transaction: if `find_by_api_token()` is used within an outer transaction then a failing `update()` would cause the outer transaction to be poisoned on the database side, even though the error is handled in the `or_else()` invocation on the app side.

TIL: transactions with only a single statement are not always as pointless as I thought...

I'll try to come up with a couple of tests for this behavior in a follow-up PR

PR #9575, opened 2024-10-04
Correct frontend source dir

HigherOrderLogic:
According to https://github.com/rust-lang/crates.io/blob/main/docs/ARCHITECTURE.md#frontend---emberjs, the frontend source directory should be `app`.


Turbo87 (approved):
good catch, thank you!

PR #9619, opened 2024-10-10
admin/yank_version: Reduce `spawn_blocking()` usage

Turbo87:
similar to https://github.com/rust-lang/crates.io/pull/9598


Turbo87:
well... looks like this is going to be a bit more complicated... 🙈

PR #9624, opened 2024-10-10
Add `release_tracks` meta to pagination versions

eth3lbert:
This PR will add `release_tracks` information to the `meta` field for paginated versions. This is necessary for pagination because it only loads a portion of versions per page, while `release_tracks` may require all sorted versions to determine their values. Therefore, we also need to calculate these on the server side.

---

Tasks:
- [x] `release_tracks`
  - [x] sorted by semver
  - [x] sorted by date
  - [x] optional by `include` query param

### DIFF
--- a/src/controllers/krate/versions.rs
+++ b/src/controllers/krate/versions.rs
@@ -157,7 +157,7 @@ fn list_by_semver(
 ) -> AppResult<PaginatedVersionsAndPublishers> {
     use seek::*;
 
-    let (data, total) = if let Some(options) = options {
+    let (data, total, release_tracks) = if let Some(options) = options {
         // Since versions will only increase in the future and both sorting and pagination need to
         // happen on the app server, implementing it with fetching only the data needed for sorting
         // and pagination, then making another query for the data to respond with, would minimize
@@ -168,18 +168,26 @@ fn list_by_semver(
         let mut sorted_versions = IndexMap::new();
         for result in versions::table
             .filter(versions::crate_id.eq(crate_id))
-            .select((versions::id, versions::num))
-            .load_iter::<(i32, String), DefaultLoadingMode>(conn)?
+            .select((versions::id, versions::num, versions::yanked))
+            .load_iter::<(i32, String, bool), DefaultLoadingMode>(conn)?
         {
-            let (id, num) = result?;
-            sorted_versions.insert(id, (num, None));
+            let (id, num, yanked) = result?;
+            let semver = semver::Version::parse(&num).ok();
+            sorted_versions.insert(id, (semver, yanked, None));
         }
-        sorted_versions.sort_by_cached_key(|_, (num, _)| Reverse(semver::Version::parse(num).ok()));
+        sorted_versions
+            .sort_unstable_by(|_, (semver_a, _, _), _, (semver_b, _, _)| semver_b.cmp(semver_a));
 
         assert!(
             !matches!(&options.page, Page::Numeric(_)),
             "?page= is not supported"
         );
+        let release_tracks = Some(ReleaseTracks::from_sorted_semver_iter(
+            sorted_versions
+                .values()
+                .filter(|(_, yanked, _)| !yanked)
+                .filter_map(|(semver, _, _)| semver.as_ref()),
+        ));
         let mut idx = Some(0);
         if let Some(SeekPayload::Semver(Semver { id })) = Seek::Semver.after(&options.page)? {
             idx = sorted_versions
@@ -201,19 +209,24 @@ fn list_by_semver(
                 .load_iter::<(Version, Option<User>), DefaultLoadingMode>(conn)?
             {
                 let row = result?;
-                sorted_versions.insert(row.0.id, (row.0.num.to_owned(), Some(row)));
+                // The versions are already sorted, and we only need to enrich the fetched rows into them.
+                // Therefore, other values can now be safely ignored.
+                sorted_versions
+                    .entry(row.0.id)
+                    .and_modify(|entry| *entry = (None, false, Some(row)));
             }
 
             let len = sorted_versions.len();
             (
                 sorted_versions
                     .into_values()
-                    .filter_map(|(_, v)| v)
+                    .filter_map(|(_, _, v)| v)
                     .collect(),
                 len,
+                release_tracks,
             )
         } else {
-            (vec![], 0)
+            (vec![], 0, release_tracks)
         }
     } else {
         let mut data: Vec<(Version, Option<User>)> = versions::table
@@ -223,7 +236,7 @@ fn list_by_semver(
             .load(conn)?;
         data.sort_by_cached_key(|(version, _)| Reverse(semver::Version::parse(&version.num).ok()));
         let total = data.len();
-        (data, total)
+        (data, total, None)
     };
 
     let mut next_page = None;
@@ -237,7 +250,7 @@ fn list_by_semver(
         meta: ResponseMeta {
             total: total as i64,
             next_page,
-            release_tracks: None,
+            release_tracks,
         },
     })
 }
--- a/src/controllers/krate/versions.rs
+++ b/src/controllers/krate/versions.rs
@@ -133,7 +133,11 @@ fn list_by_date(
 
     Ok(PaginatedVersionsAndPublishers {
         data,
-        meta: ResponseMeta { total, next_page },
+        meta: ResponseMeta {
+            total,
+            next_page,
+            release_tracks: None,
+        },
     })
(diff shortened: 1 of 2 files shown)


src/controllers/krate/versions.rs
```diff
@@ -302,4 +354,228 @@ struct PaginatedVersionsAndPublishers {
 struct ResponseMeta {
     total: i64,
     next_page: Option<String>,
+    release_tracks: Option<ReleaseTracks>,
```
Turbo87, line 357: might make sense to use `skip_if_empty` (or whatever it is called) for this field

src/controllers/krate/versions.rs
```diff
@@ -109,6 +119,25 @@ fn list_by_date(
             )
         }
         query = query.limit(options.per_page);
+
+        if include.release_tracks {
+            let mut sorted_versions = IndexSet::new();
+            for result in versions::table
+                .filter(versions::crate_id.eq(crate_id))
+                .filter(versions::yanked.ne(true))
```
Turbo87, line 127:
```suggestion
                .filter(versions::yanked.eq(false))
```

😅

src/controllers/krate/versions.rs
```diff
@@ -302,4 +354,228 @@ struct PaginatedVersionsAndPublishers {
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+enum ReleaseSlug {
+    MajorX(u64),
+    MajorMinor(u64, u64),
+    Major(u64),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Serialize)]
+struct HighestSemver {
+    highest: SemverTriple,
```
Turbo87, line 372: why do we not use `semver::Version` here?

src/controllers/krate/versions.rs
```diff
@@ -302,4 +354,228 @@ struct PaginatedVersionsAndPublishers {
...
+        }
+    }
+}
+
+impl std::fmt::Display for ReleaseSlug {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::MajorX(major) => write!(f, "{major}.x"),
+            Self::MajorMinor(major, minor) => write!(f, "{major}.{minor}"),
+            Self::Major(major) => write!(f, "{major}"),
```
Turbo87, line 423: I'm wondering if we should really encode the `.x` thing in the backend. It's more of a presentational thing. in the `MajorMinor` case `major` is always zero, so we might not need to store the zero in the enum.

src/controllers/krate/versions.rs
```diff
@@ -302,4 +354,228 @@ struct PaginatedVersionsAndPublishers {
...
+            release_tracks: false,
+        };
+        for component in s.split(',') {
+            match component {
+                "" => {}
+                "full" => {
+                    mode = Self {
+                        release_tracks: true,
+                    }
+                }
```
Turbo87, line 478: I think we can skip the `full` option. Better to be explicit :)

Turbo87:
I thought a bit more about `first_release` too. I'm not sure we necessarily need it. If we decide to keep it, I think it would be fine to add a `first_version_id` column to the `crates` table instead of computing it on demand all the time. The value won't ever change once a crate is published, so we won't run into dead rows issues in the `crates` table.

...

and then I thought some more and realized that with version deletion being a thing soon it would be possible to delete that first version 😅 

I think I'm back on "do we actually need it" :D

eth3lbert:
> I thought a bit more about `first_release` too. I'm not sure we necessarily need it. If we decide to keep it, I think it would be fine to add a `first_version_id` column to the `crates` table instead of computing it on demand all the time. The value won't ever change once a crate is published, so we won't run into dead rows issues in the `crates` table.

Totally agree! I've actually been thinking about the same thing.


> and then I thought some more and realized that with version deletion being a thing soon it would be possible to delete that first version 😅
>
> I think I'm back on "do we actually need it" :D

The `first_release` mentioned in this PR is simply to maintain existing functionality. If it's possible to simply not display it, then I'm leaning towards not implementing it.

And due to the above reasoning, I think I'll change this PR to focus solely on `release_tracks`.

src/controllers/krate/versions.rs
eth3lbert: Oops, I have no idea what I'm thinking.

Or just
```suggestion
                .filter(not(versions::yanked))
```

src/controllers/krate/versions.rs
eth3lbert: Good call!

src/controllers/krate/versions.rs
eth3lbert: > I'm wondering if we should really encode the `.x` thing in the backend. It's more of a presentational thing.

I just implemented it to match the current frontend's implementation.
Do we want to change the representation in this PR or just change it in a follow-up PR if necessary?



> in the `MajorMinor` case `major` is always zero, so we might not need to store the zero in the enum.

Great call!

src/controllers/krate/versions.rs
eth3lbert: Since we only need the `major`, `minor`, and `patch`, we don't need to keep the full `semver::Version`.

eth3lbert: Hmm, after reconsidering, I think keeping `semver::Version` could be useful if we need a full version string. This is likely what we want, since we don't give id here, and if we need to map it back at frontend, we might require a full version string to do so. Therefore, I'll make the change!

Turbo87: yeah, exactly

src/controllers/krate/versions.rs
eth3lbert: I think I got your point now. Are you suggesting that instead of providing `1.x` we could simply provide `1` and let client to determine what to display? In that case, I think two variants `Major` and `Minor` would be sufficient.

Turbo87: yeah, exactly

eth3lbert:
All feedback addressed! Thanks for the review :)

src/controllers/krate/versions.rs
```diff
@@ -302,4 +355,208 @@ struct PaginatedVersionsAndPublishers {
...
+                    HighestSemver {
+                        highest: num.clone(),
+                    },
+                );
+            }
+        }
+
+        Self(map)
+    }
+}
```
Turbo87, line 399: small nitpick: can we keep the `impl` blocks next to the struct/enum which they are associated with? IMHO it makes it a bit easier to understand the code that way since you don't have to search for impl blocks far away from the struct/enum.

eth3lbert: Sure!

src/controllers/krate/versions.rs
```diff
@@ -302,4 +355,208 @@ struct PaginatedVersionsAndPublishers {
 struct ResponseMeta {
     total: i64,
     next_page: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    release_tracks: Option<ReleaseTracks>,
+}
+
+#[derive(Debug, Eq, PartialEq, Serialize)]
+struct ReleaseTracks(IndexMap<ReleaseSlug, HighestSemver>);
```
Turbo87, line 363:
```suggestion
struct ReleaseTracks(IndexMap<ReleaseTrackName, ReleaseTrackDetails>);
```

the other nitpick: what do you think about names like this to make it clearer that these are all related to the release tracks field?

eth3lbert: This seems acceptable to me, although the name detail is not very self-explanatory. However, the content is straightforward, so I think it should be fine.

Turbo87: > although the name detail is not very self-explanatory

that is true, but `HighestSemver` assumes that we will never add any other fields to the struct and I'm not sure whether that will stay true forever :D

eth3lbert: yeah, fair enough!

eth3lbert:
@Turbo87 all feedback has been addressed, and I believe this is ready to be merged. Let me know if anything needs to be changed.

Turbo87 (approved):
🎉

eth3lbert:
Hmm, it seems this doesn't have to be restricted to only available for paginated versions since you need to specify the include parameter anyway.

I'd like to have it available for non-paginated requests as well, which could make our migration to paginated versions smoother. Any thoughts, @Turbo87 ?

Turbo87:
I don't see any reason against it at first glance 👍

PR #9627, opened 2024-10-11
worker: Add `DEDUPLICATED` flag to the `BackgroundJob` trait

Turbo87:
This flag can be set when implementing background jobs to automatically deduplicate jobs when they are enqueued. If an unstarted job already exists in the queue the `enqueue()` fn will return `Ok(None)` instead of the job ID.

This should allow us to get rid of `enqueue_sync_to_index()` as a special case and make the same functionality available to other jobs as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1200,11 +1200,13 @@ name = "crates_io_worker"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "claims",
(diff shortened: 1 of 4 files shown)


LawnGnome (approved):
Took a few minutes where you could hear the gears grinding in my head to suss out the deduplicated query, but now my brain has kicked in, LGTM. :+1:

PR #9628, opened 2024-10-11
worker/jobs: Use `DEDUPLICATED` for jobs where it makes sense

Turbo87:
Most of our background jobs don't need to be enqueued multiple times if there is already a similar one in the queue. This commit changes most of the jobs to use the new `DEDUPLICATED` flag on the `BackgroundJob` impl.


LawnGnome (approved):
OK, so I think that means the ones that aren't going to be de-duplicated are:

* `ProcessCdnLogQueue`
* `NormalizeIndex`
* `RenderAndUploadReadme`

Feels like this should be OK.

PR #9643, opened 2024-10-12
Remove CI system detection

Turbo87:
This was interesting when we were still routing the download traffic through our API servers, but this is no longer the case, so there is very little point in still running this detection for each request.


Turbo87:
looks like there are no objections, so I'll go ahead and merge this

PR #9648, opened 2024-10-13
Extract `crates_io_database` crate

Turbo87:
This crate only contains the database schema at the moment, but could be extended with other purely database-related code in the future. This should hopefully help with untangling our mess of interdependencies in the `models` module.

This might also help a little bit with compile times of the `crates_io` crate, since all of the `table!()` macros can be expanded once and can then be cached since the schema does not change that often.

Finally, it would make it possible for other subcrates to depend on the schema, without pulling in all of the other `crates_io` code.


Turbo87:
alright, looks like no objections, so let's do this :)

PR #9696, opened 2024-10-19
Return downloads for the top 5 most downloaded versions

joshka:
This changes the downloads endpoint to return the five most downloaded
versions of a crate, instead of the five most recent versions. This
should make the endpoint more useful for users who are interested in
seeing the most popular versions of a crate.

In particular it helps with the case where a crate has many pre-release
versions, which are not as interesting to users as the stable versions.

Fixes: https://github.com/rust-lang/crates.io/discussions/4336

---

I'm not certain this is the right approach and will discuss more in the discussion thread. Putting this here as something which maybe can be tested against live data to understand the perf issues. I'm not sure how to do that.

### DIFF
--- a/src/controllers/krate/downloads.rs
+++ b/src/controllers/krate/downloads.rs
@@ -14,7 +14,6 @@ use axum::Json;
 use diesel::prelude::*;
 use diesel_async::RunQueryDsl;
 use serde_json::Value;
-use std::cmp;
 
 /// Handles the `GET /crates/:crate_id/downloads` route.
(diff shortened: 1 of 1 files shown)


Turbo87:
thanks, but this is a breaking change to our API, which makes this hard to merge as-is

joshka:
> thanks, but this is a breaking change to our API, which makes this hard to merge as-is

This is more for the discussion / idea than for actually merging as-is at this point.

PR #9712, opened 2024-10-21
Replace `derive_builder` with `bon`

Turbo87:
`bon` uses the type system to ensure all required fields have been set, which means that we can catch issues at compile time. The necessary attributes are also less verbose due to better defaults inside `bon`. 

The only downside I've found is that custom builder fns like `dummy_checksum()` don't appear easy to implement at the moment, which is why I've inlined the two custom builder fns that we had. (/cc @Veetaha)

### DIFF
--- a/src/models/version.rs
+++ b/src/models/version.rs
@@ -1,8 +1,8 @@
 use std::collections::BTreeMap;
 
+use bon::Builder;
 use chrono::NaiveDateTime;
 use crates_io_index::features::FeaturesMap;
-use derive_builder::Builder;
 use diesel::prelude::*;
 use serde::Deserialize;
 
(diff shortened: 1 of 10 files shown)


Veetaha:
@Turbo87 regarding the custom dummy checksum finishing function (function that consumes the builder and returns the result).

It should be possible to add custom methods to the builder itself in the next release. It will be a big 3.0 release, but breaking changes there won't touch 99% of the users since they break some rare obscure cases and rare-ly used apis such as `#[builder(expose_positional_fn)]`, so the upgrade will probably require just a bump in `Cargo.toml` in this case.

The builder type state API will be stabilized (the PR for that was merged into master https://github.com/elastio/bon/pull/145).
There is an example in that PR that shows how one could add methods to the builder. The documentation for this feature is not ready at the moment though

Veetaha:
Regarding the usage in test fixtures. In 3.0 it will be possible to annotate the type of the builder, so test fixtures could benefit from this by returning a builder with (potentially partially) filled values like so:

```rust
#[derive(bon::Builder)]
#[cfg_attr(test, builder(on(_, overwritable)))]
struct User {
    login: String,
    name: String,
    level: u32,
}

#[cfg(test)]
mod tests {
    use super::*;
    use user_builder::{SetLevel, SetLogin, SetName};

    // Base fixture that has dummy values for all fields
    fn user() -> UserBuilder<SetLevel<SetName<SetLogin>>> {
        User::builder()
            .level(24)
            .name("Bon Bon".to_owned())
            .login("@bonbon".to_owned())
    }

    #[test]
    fn test_with_empty_name() {
        // Build user with an empty name, and all other fields filled with dummy data
        let user = user()
            .name("".to_owned())
            .build();

        // ...
    }

    #[test]
    fn test_with_zero_level_and_admin_login() {
        let admin = user()
            .level(0)
            .login("@admin".to_owned())
            .build();

        // ...
    }
}
```

Although, I'm not sure how useful this will be. Notice the `#[builder(ovewritable)]` attribute that is also planned (https://github.com/elastio/bon/issues/149) to allow for overwrites of already set values that is useful in case of fixtures.

src/models/version.rs
```diff
@@ -80,58 +80,26 @@ impl Version {
...
 
-impl NewVersionBuilder<'_> {
-    pub fn features(
-        &mut self,
-        features: &BTreeMap<String, Vec<String>>,
-    ) -> serde_json::Result<&mut Self> {
-        self.features = Some(serde_json::to_value(features)?);
-        Ok(self)
-    }
-
```
Veetaha, line 118: There will be a new attribute in `3.0.` that would allow expressing this setter as this (apart from defning the custom method on the builder directly):
```rust
#[builder(with = |features: &BTreeMap<...>| -> serde_json::Result<_> { 
  serde_json::to_value(features) 
})]
features: serde_json::Value,
```
(yes it's possible to have a fallible setter this way if you annotate the return type of the closure with a type that has `Result` suffix)

src/models/version.rs
```diff
@@ -80,58 +80,26 @@ impl Version {
...
-    /// Set the `checksum` field to a basic dummy value.
-    pub fn dummy_checksum(&mut self) -> &mut Self {
-        const DUMMY_CHECKSUM: &str =
-            "0000000000000000000000000000000000000000000000000000000000000000";
-
-        self.checksum(DUMMY_CHECKSUM)
-    }
```
Veetaha, line 125: This method will require an MSRV of 1.79.0 in bon 3.0.

An example of how this will work that exactly mirrors this usage is described on the new comment on the cargo feature that will be avaialble in 3.0 here:

https://github.com/elastio/bon/blob/aed714815be5affea9a89a15ea728f4e1ab7e283/bon/Cargo.toml#L66-L104

Veetaha (approved):

Turbo87:
awesome, thanks for the quick feedback :)

PR #9722, opened 2024-10-22
`default_version` for publish

eth3lbert:
Since we will respond with the `default_version` to the client, this PR always upsert the `default_value` for publish.

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -23,8 +23,8 @@ use tokio::runtime::Handle;
 use url::Url;
 
 use crate::models::{
-    insert_version_owner_action, Category, Crate, DependencyKind, Keyword, NewCrate, NewVersion,
-    Rights, VersionAction,
+    default_versions::Version as DefaultVersion, insert_version_owner_action, Category, Crate,
+    DependencyKind, Keyword, NewCrate, NewVersion, Rights, VersionAction,
 };
 
 use crate::licenses::parse_license_expr;
@@ -64,7 +64,7 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
 
     Crate::validate_crate_name("crate", &metadata.name).map_err(bad_request)?;
 
-    let version = match semver::Version::parse(&metadata.vers) {
+    let semver = match semver::Version::parse(&metadata.vers) {
         Ok(parsed) => parsed,
         Err(_) => {
             return Err(bad_request(format_args!(
@@ -75,7 +75,7 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
     };
 
     // Convert the version back to a string to deal with any inconsistencies
-    let version_string = version.to_string();
+    let version_string = semver.to_string();
 
     let request_log = req.request_log();
     request_log.add("crate_name", &*metadata.name);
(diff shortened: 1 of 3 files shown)


src/controllers/krate/publish.rs
eth3lbert, line 1: Another approach would be to calculate from all versions. Since `TopVersion` already needs to calculate from all versions, we could consider pre-sorting the versions and providing methodologies for `TopVersion` and `default_versions` to determine values from the sorted versions.

src/controllers/krate/publish.rs
```diff
@@ -385,16 +385,40 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
                 .values((
                     default_versions::crate_id.eq(krate.id),
-                    default_versions::version_id.eq(version.id),
+                    default_versions::version_id.eq(default_version.id),
                 ))
-                .on_conflict_do_nothing()
+                .on_conflict(default_versions::crate_id)
+                .do_update()
+                .set(default_versions::version_id.eq(default_version.id))
                 .execute(conn)?;
```
Turbo87, line 422:
```rust
// Upsert the `default_version` determined by the existing `default_version` and the
// published version. Note that this could potentially write an outdated version
// (although this should not happen regularly), as we might be comparing to an
// outdated value.
//
// Compared to only using a background job, this prevents us from getting into a
// situation where a crate exists in the `crates` table but doesn't have a default
// version in the `default_versions` table.

if let Some(existing_default_version) = existing_default_version {
    let published_default_version = DefaultVersion {
        id: version.id,
        num: semver,
        yanked: false,
    };

    if existing_default_version < published_default_version {
        diesel::update(default_versions::table)
            .filter(default_versions::crate_id.eq(krate.id))
            .set(default_versions::version_id.eq(version.id))
            .execute(conn)?;
    }
} else {
    diesel::insert_into(default_versions::table)
        .values((
            default_versions::crate_id.eq(krate.id),
            default_versions::version_id.eq(version.id),
        ))
        .execute(conn)?;
}
```

I guess we could do something like this to skip the update query if the value does not need updating. what do you think?

src/controllers/krate/publish.rs
```diff
@@ -385,16 +385,40 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
+            // outdated value.
+            //
+            // Compared to only using a background job, this prevents us from getting into a
+            // situation where a crate exists in the `crates` table but doesn't have a default
+            // version in the `default_versions` table.
+            let default_version = if let Some(existing_default_version) = existing_default_version {
+                std::cmp::max(existing_default_version, published_default_version)
+            } else {
+                published_default_version
+            };
```
eth3lbert, line 413: This could be simplified to the following:
```suggestion
            let default_version = std::cmp::max(Some(published_default_version), existing_default_version)
                .expect("At least the published version should be selected.");
```

eth3lbert: oops this is an outdated review 😅

Turbo87: it's probably correct, but it feels arcane enough to require a comment and proper testing :D

src/controllers/krate/publish.rs
eth3lbert: Updated, thanks :)

src/controllers/krate/publish.rs
```diff
@@ -385,17 +385,43 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
+
+            // Upsert the `default_value` determined by the existing `default_value` and the
+            // published version. Note that this could potentially write an outdated version
+            // (although this should not happen regularly), as we might be comparing to an
+            // outdated value.
+            //
+            // Compared to only using a background job, this prevents us from getting into a
+            // situation where a crate exists in the `crates` table but doesn't have a default
+            // version in the `default_versions` table.
+            // let default_version = std::cmp::max(existing_default_version, Some(published_default_version))
```
Turbo87, line 403:
```suggestion
```

😉

eth3lbert: 🙈

src/controllers/krate/publish.rs
```diff
@@ -446,11 +471,9 @@ pub async fn publish(app: AppState, req: BytesRequest) -> AppResult<Json<GoodCra
...
             SendPublishNotificationsJob::new(version.id).enqueue(conn)?;
 
-            // If this is a new version for an existing crate it is sufficient
-            // to update the default version asynchronously in a background job.
-            if inserted_default_versions == 0 {
-                UpdateDefaultVersion::new(krate.id).enqueue(conn)?;
-            }
+            // Update the default version asynchronously in a background job
+            // to ensure correctness and eventual consistency.
+            UpdateDefaultVersion::new(krate.id).enqueue(conn)?;
```
Turbo87, line 476: I guess we could move this into the `if let Some(existing_default_version) = existing_default_version` condition above. we don't need to enqueue it in the `insert_into()` case.

eth3lbert: So rather than always enqueuing the update job, we only enqueue when it's not a new crate, namely when releasing a new version, right? 

Seems right to me.

Turbo87 (approved):
:)

PR #9741, opened 2024-10-23
Expose the `default_version` on the API

eth3lbert:
This PR exposes the `default_version` on the API, which was introduced in #8484.
I implements this into multiple commits, which should be easier to follow and inspect :)


src/views.rs
```diff
@@ -208,6 +208,8 @@ pub struct EncodableCrate {
     // NOTE: Used by shields.io, altering `downloads` requires a PR with shields.io
     pub downloads: i64,
     pub recent_downloads: Option<i64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub default_version: Option<String>,
```
eth3lbert, line 212: I'm unsure if we should make this `Option<String> or just a `String`. Currently, I have no idea what would happen if a crate had no versions available (crate deletion?). So, for now, I'll make it an optional string. We could also consider making it a string with either an `""` or `"0.0.0"` as the value.

Turbo87: > I have no idea what would happen if a crate had no versions available (crate deletion?).

AFAIK currently the constraints in the database are set up in a way that won't allow the deletion of the version that is referenced by the `default_versions` table. the only way to delete a default version is to assign a new default version in the same transaction.

the only other path to the `default_versions` row missing that I can think of is if a crate is created in the database without a corresponding `default_versions` entry, which is why we have the `insert_into(default_versions)` directly in the publish endpoint.

so in theory this should not be able to happen... 😅 

that said, I'm not opposed to keeping this as an `Option`, but we should probably log it or create a Sentry issue if we indeed encounter a `None` here.

eth3lbert: > AFAIK currently the constraints in the database are set up in a way that won't allow the deletion of the version that is referenced by the `default_versions` table. the only way to delete a default version is to assign a new default version in the same transaction.
> 
> the only other path to the `default_versions` row missing that I can think of is if a crate is created in the database without a corresponding `default_versions` entry, which is why we have the `insert_into(default_versions)` directly in the publish endpoint.
> 
> so in theory this should not be able to happen... 😅

Indeed, after inspecting the DDL, I see that the `version_id` column is `NOT NULL` and references the `versions` table, just as you mentioned. However, for crate deletion, do we really want to implement a restriction that prevents users from deleting a version if it only has one version available? The RFC doesn't mention this part.

> that said, I'm not opposed to keeping this as an `Option`, but we should probably log it or create a Sentry issue if we indeed encounter a `None` here.

Good call!

eth3lbert: Other notes: My implementation starts with `skip_serializer_if` because this allows me to break the implementation into multiple sub-commits. However, we can also remove this part at the end if it's more suitable.

Turbo87: > However, for crate deletion, do we really want to implement a restriction that prevents users from deleting a version if it only has one version available? The RFC doesn't mention this part.

either that, or deleting the last version automatically deletes the whole crate, but I'd rather not end up in a situation where a crate exists in our database but doesn't have any published versions. a lot of the ecosystem is expecting this to not be a possible state.

> we can also remove this part at the end if it's more suitable

yeah, I think that makes sense

eth3lbert: > deleting the last version automatically deletes the whole crate

If so, does this name become available after a crate is deleted? If it does, are there any potential issues, such as someone squatting on or impersonating the deleted crate?

Turbo87: > does this name become available after a crate is deleted?

yes

> If it does, are there any potential issues, such as someone squatting on or impersonating the deleted crate?

given the requirements that we've established for deleting crates and versions I'd expect it to be rather unlikely for this to have a significant impact. remember that versions can only be deleted within 72 hours. so if you delete the only/last version of a crate then the crate will also only have existed for max. 72 hours and most likely won't have gained enough popularity to be a meaningful target (+ the other requirements for deleting crates...).

eth3lbert: Sounds reasonable to me 👍

Turbo87 (approved):

PR #9753, opened 2024-10-25
Create devcontainer.json

mustafacco7:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,5 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
```
Turbo87, line 2: what is this? could you give a bit more context? also, why this specific image?

PR #9765, opened 2024-10-27
Migrate the frontend to use the new yank API

Rustin170506:
Moved to the new API before adding frontend support for the yank message.

Tested locally:

https://github.com/user-attachments/assets/bf3ce6c2-b6b5-499c-b878-5180e710fd7d

### DIFF
--- a/app/adapters/version.js
+++ b/app/adapters/version.js
@@ -1,0 +1,9 @@
+import ApplicationAdapter from './application';
+
+export default class VersionAdapter extends ApplicationAdapter {
+  urlForUpdateRecord(id, modelName, snapshot) {
+    let crateName = snapshot.record.crate.id;
+    let num = snapshot.record.num;
+    return `/${this.namespace}/crates/${crateName}/${num}`;
+  }
+}
--- a/app/models/version.js
+++ b/app/models/version.js
@@ -166,22 +166,42 @@ export default class Version extends Model {
   }
 
   yankTask = keepLatestTask(async () => {
-    let response = await fetch(`/api/v1/crates/${this.crate.id}/${this.num}/yank`, { method: 'DELETE' });
+    let response = await fetch(`/api/v1/crates/${this.crate.id}/${this.num}`, {
+      method: 'PATCH',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        version: {
+          yanked: true,
+        },
+      }),
+    });
     if (!response.ok) {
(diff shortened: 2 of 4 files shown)


Rustin170506 (commented):

mirage/route-handlers/crates.js
```diff
@@ -264,6 +264,21 @@ export function register(server) {
...
+    if (!crate) {
+      return notFound();
+    }
+
+    const version = schema.versions.findBy({ crateId: crate.id, num: versionNum });
+    if (!version) {
+      return notFound();
+    }
+
+    return {};
```
Turbo87, line 279: we should update the `yank` status and add the `yank_message` field here and then return the serialized version. note that there is also a test suite for the mirage route handlers to ensure that it roughly matches the real API that we are mocking here. if we implement the frontend code against an API that does not match the real API we're in for trouble 😅

Rustin170506: I followed the previous mock API 😅
```js
  server.delete('/api/v1/crates/:name/:version/yank', (schema, request) => {
    const { name, version: versionNum } = request.params;
    const crate = schema.crates.findBy({ name });
    if (!crate) {
      return notFound();
    }

    const version = schema.versions.findBy({ crateId: crate.id, num: versionNum });
    if (!version) {
      return notFound();
    }

    return {};
  });

  server.put('/api/v1/crates/:name/:version/unyank', (schema, request) => {
    const { name, version: versionNum } = request.params;
    const crate = schema.crates.findBy({ name });
    if (!crate) {
      return notFound();
    }

    const version = schema.versions.findBy({ crateId: crate.id, num: versionNum });
    if (!version) {
      return notFound();
    }

    return {};
  });
```

Do we need to update the old one as well? Or just delete it?

Turbo87: ouch, wow, yeah, those are definitely wrong 🙈

Rustin170506: Alright, I’ll update it as well. Initially, I thought we were focusing only on UI testing(we only checked the button name), so I didn’t consider the actual version status. I’ll make sure to update the version status correctly.

Turbo87: > I thought we were focusing only on UI testing(we only checked the button name), so I didn’t consider the actual version status

we have the "crates can be yanked by owner" test, but that apparently only asserts local state, which meant that the bug in the mirage implementation was hidden. and apparently we currently don't have any tests for the yank/unyank endpoints in mirage 😢

Rustin170506 (commented):

PR #9767, opened 2024-10-27
database: Change `num_no_build` to be `NOT NULL`

Turbo87:
This PR was extracted from https://github.com/rust-lang/crates.io/pull/9756, because it needs to be deploy separately from the other migrations in that PR.

The backfill SQL script in https://github.com/rust-lang/crates.io/pull/9766 has been run on the staging and production databases, so this should be good to go. `SELECT * FROM versions WHERE num_no_build IS NULL` returns zero rows.

### DIFF
--- a/src/models/version.rs
+++ b/src/models/version.rs
@@ -34,7 +34,7 @@ pub struct Version {
     pub has_lib: Option<bool>,
     pub bin_names: Option<Vec<Option<String>>>,
     pub yank_message: Option<String>,
-    pub num_no_build: Option<String>,
+    pub num_no_build: String,
 }
 
 impl Version {
@@ -85,6 +85,8 @@ pub struct NewVersion<'a> {
     crate_id: i32,
     #[builder(start_fn)]
     num: &'a str,
+    #[builder(default = strip_build_metadata(num))]
+    num_no_build: &'a str,
     created_at: Option<&'a NaiveDateTime>,
     yanked: Option<bool>,
     #[builder(default = serde_json::Value::Object(Default::default()))]
--- a/crates/crates_io_database/src/schema.rs
+++ b/crates/crates_io_database/src/schema.rs
@@ -990,7 +990,7 @@ diesel::table! {
         /// message associated with a yanked version
         yank_message -> Nullable<Text>,
         /// This is the same as `num` without the optional "build metadata" part (except for some versions that were published before we started validating this).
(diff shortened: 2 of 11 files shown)


src/models/version.rs
```diff
@@ -85,6 +85,8 @@ pub struct NewVersion<'a> {
     crate_id: i32,
     #[builder(start_fn)]
     num: &'a str,
+    #[builder(default = strip_build_metadata(num))]
+    num_no_build: &'a str,
```
Turbo87, line 89: argh, I screwed this up... this part of the diff should've been in the previous PR instead. every new version is currently inserted with `num_no_build = NULL`, which makes the `set not null` migration in this PR fail 🤦‍♂️ 

```sql
update versions
set num_no_build = split_part(num, '+', 1)
where num_no_build IS NULL;
```

I will run ⬆️ right before deploying, which should be enough for this to deploy cleanly...

Turbo87: alright, that worked. all good now! :)

eth3lbert:
When I tried to import data using `scripts/import-database-dump.sh` today, I encountered an error, as shown below:
```
...
    \copy "versions" ("bin_names", "checksum", "crate_id", "crate_size", "created_at", "downloads", "edition", "features", "has_lib", "id", "license", "links", "num", "published_by", "rust_version", "
updated_at", "yanked") FROM 'data/versions.csv' WITH CSV HEADER
ERROR:  null value in column "num_no_build" of relation "versions" violates not-null constraint
...
```

I believe this error occurred because the `"num_no_build"` column is set to `NOT_NULL` and we also exclude it from our dump data. This then violates the `NOT_NULL` constraint.

Turbo87:
ugh... unfortunately that makes a lot of sense 😞

we might have to rethink https://github.com/rust-lang/crates.io/pull/9786 🤔

@dtolnay any thoughts on this?

eth3lbert:
Another idea would be to modify the script template, changing the DDL to allow `NULL` values before copying. We could also include https://github.com/rust-lang/crates.io/pull/9767#discussion_r1818129936  in the template to set the value after copying and then set the column back to `NOT_NULL` after updating.

Turbo87:
that basic version of the backfill won't help for the full dataset though, because it would cause uniqueness issues on the corresponding index for some of the older versions where we still allowed for duplicates :-/

PR #9786, opened 2024-10-29
Exclude num_no_build from versions.csv db-dump

dtolnay:
In versions.csv, the **num_no_build** column is identical to **num** but without build metadata. So if **num**="1.0.0+example" then **num_no_build**="1.0.0".

It exists in the database only for the purpose of defining a `unique` constraint on `(crate_id, num_no_build)`. See https://github.com/rust-lang/crates.io/pull/9756.

But there is no reason to include it in db-dump.tar.gz because in downstream code it's trivial to reconstruct from the value of **num** if needed.


Turbo87 (approved):
seems reasonable, thanks for the PR :)

PR #9840, opened 2024-11-04
mirage: Add `default_version` and `yanked` to `crate` serializer

eth3lbert:
This PR adds `default_version` and `yanked` to the `crate` serializer.

Related:
- #9741 
- #9817


mirage/serializers/crate.js
```diff
@@ -19,6 +19,8 @@ export default BaseSerializer.extend({
     'id',
     'keywords',
     'links',
+    'default_version',
+    'yanked',
```
Turbo87, line 23: are you sure that we need these mentioned here? I would've thought that those are only necessary for the fields that come directly from the corresponding factory/model

eth3lbert: I saw we have https://github.com/rust-lang/crates.io/blob/a46fcdcba0e2995b47c23b3b04e7ff5104d176fe/mirage/route-handlers/summary.js#L25-L35 which leverages serializer. So I just made the change in the serializer.

eth3lbert: But after your awareness, I think we might also want to change the factory and fixture as well?

Turbo87: I'm not sure I understand what you're saying 😅 

my question is: do the tests work the same as before when these two lines are removed again? or is something breaking then?

eth3lbert: Oh, my bad, that was quite embarrassing! A quick inspection of the mirage response confirms your absolute right.

eth3lbert: I have removed the attrs part.

---

Should we add these fields to mirage/factory or mirage/fixtures?

Turbo87: > Should we add these fields to mirage/factory or mirage/fixtures?

no :)

factory is for regular fields, model is for relationships, fixtures can die for all I care, and the serializer can be used for "computed" fields 😉

eth3lbert: TIL!!

Turbo87 (approved):

PR #9842, opened 2024-11-04
frontend: Migrate to use `default_version` and `yanked`

eth3lbert:
This PR completes the transition to `default_version` in the frontend, which was first introduced in #8484.

### DIFF
--- a/app/components/crate-row.hbs
+++ b/app/components/crate-row.hbs
@@ -6,10 +6,10 @@
           {{@crate.name}}
         </a>
       {{/let}}
-      {{#if @crate.defaultVersion}}
-        <span local-class="version" data-test-version>v{{@crate.defaultVersion}}</span>
+      {{#if (and @crate.default_version (not @crate.yanked))}}
+        <span local-class="version" data-test-version>v{{@crate.default_version}}</span>
         <CopyButton
-          @copyText='{{@crate.name}} = "{{@crate.defaultVersion}}"'
+          @copyText='{{@crate.name}} = "{{@crate.default_version}}"'
           title="Copy Cargo.toml snippet to clipboard"
           local-class="copy-button"
           data-test-copy-toml-button
--- a/app/models/crate.js
+++ b/app/models/crate.js
@@ -10,6 +10,13 @@ export default class Crate extends Model {
   @attr recent_downloads;
   @attr('date') created_at;
   @attr('date') updated_at;
+  /**
+   * This is the default version that will be shown when visiting the crate
+   * details page. Note that this value can be `null`, which may be unexpected.
+   * @return {string}
+   */
+  @attr default_version;
+  @attr yanked;
   @attr max_version;
   @attr max_stable_version;
   @attr newest_version;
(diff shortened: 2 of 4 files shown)


app/models/crate.js
```diff
@@ -10,6 +10,13 @@ export default class Crate extends Model {
   @attr recent_downloads;
   @attr('date') created_at;
   @attr('date') updated_at;
+  /**
+   * This is the default version that will be shown when visiting the crate
+   * details page. Note that this value can be `null`, which may be unexpected.
+   * @return {string}
```
Turbo87, line 16:
```suggestion
   * @type {string | null}
```

I think `@return` is only meant for functions, no?

app/components/crate-row.hbs
```diff
@@ -6,10 +6,10 @@
           {{@crate.name}}
         </a>
       {{/let}}
-      {{#if @crate.defaultVersion}}
-        <span local-class="version" data-test-version>v{{@crate.defaultVersion}}</span>
+      {{#if (and @crate.default_version (not @crate.yanked))}}
+        <span local-class="version" data-test-version>v{{@crate.default_version}}</span>
```
Turbo87, line 10: is this a behavior change? did we previously hide this for yanked crates?

app/models/crate.js
eth3lbert: Indeed. I wasn't aware of this part.

app/components/crate-row.hbs
eth3lbert: Previously, we filtered out yanked versions in mirage, ensuring that `max_stable_version` always refers to a non-yanked version as seen in https://github.com/rust-lang/crates.io/pull/9840/files#diff-1145f70fab74d9fb32b469538c364a5b4fb1c663447498e92ad6a2c348f21ed6. Therefore, I believe this change doesn't alter the behavior.

eth3lbert: Hmm, but on the other hand, you might be right. Since we don't guarantee this part in the API, this could potentially be a behavioral change.

eth3lbert: hmm, in this case, should we keep the `defaultVersion` and change to something like:
```js
  get defaultVersion() {
    if (this.default_version && !this.yanked) {
      return this.default_version;
    }
  }
```

Turbo87: isn't `this.defaultVersion == this.default_version`?

eth3lbert: I think previously `defaultVersion` would be `undefined` if it was a **yanked** version. However, `default_version` could be a version string even if it's **yanked**?

Turbo87: ohhh, I think you're right. I missed the `filter(versions::yanked.eq(false))` in https://github.com/rust-lang/crates.io/blob/1f3322a2dd526f3eb2881b66eeb4788b5078db0b/src/models/krate.rs#L479 🙈 

in that case we can keep it as-is :)

Turbo87: yeah, probably won't hurt.

PR #9850, opened 2024-11-05
util/diesel: Add diesel prelude excluding `RunQueryDsl`

eth3lbert:
This change simplifies the migration to `diesel_async`. Using
`use diesel::prelude::*` alongside `use diesel_async::RunQueryDsl` can
sometimes cause ambiguity when not all loading functions are migrated
to use `diesel_async` within a file. This can be quite annoying.

To migrate:
1. Replace `use diesel::prelude::*` with `use crate::util::diesel::prelude::*`
    for files that still need to migrate.
2. Add `use diesel::RunQueryDsl` within loading functions.
3. Once all loading functions within a file have been migrated to
       `diesel_async`, you can revert to using `use diesel::prelude::*`
       alongside `use diesel_async::RunQueryDsl` without issues.


Turbo87 (approved):
seems like a good idea 👍

PR #9856, opened 2024-11-05
Inline `CrateVersions` trait fns

Turbo87:
Having these "non-yanked" and "including yanked" variants is quite confusing and doesn't provide much benefit compared to using `Version::belonging_to(...)` directly...

Related:

- https://github.com/rust-lang/crates.io/pull/1459


eth3lbert (approved):
Yeah, I got confused sometimes.

PR #8262, opened 2024-03-07
opensearch: Adjust search engine name and description

Turbo87:
Searching "on Cargo" seemed a bit confusing when the page is called crates.io. I assume this was a remainder of the time before the cargo registry was rebranded to crates.io.

PR #9860, opened 2024-11-06
opensearch: Cargo -> crates.io

ryuapp:
ShortName was changed to crates.io in #8262, so it should be changed as well.


Turbo87 (approved):
makes sense, thank you! :)

PR #9872, opened 2024-11-07
auth: Convert `check()` and related blocking fns to async

eth3lbert:
This PR focuses on converting the `check()` and related implementations to async, so there might be `spawn_blocking()`  fns in controllers that can be removed. However, I expect these changes to be made in a separate PR :)


Turbo87 (approved):
nice work! 👍

PR #9877, opened 2024-11-08
Reduce `spawn_blocking()` usage

eth3lbert:
This is a follow-up PR to #9872 that reduce/remove `spawn_blocking()` usage :)

### DIFF
--- a/src/controllers/krate/follow.rs
+++ b/src/controllers/krate/follow.rs
@@ -5,23 +5,24 @@ use crate::auth::AuthCheck;
 use crate::controllers::helpers::ok_true;
 use crate::models::{Crate, Follow};
 use crate::schema::*;
-use crate::tasks::spawn_blocking;
-use crate::util::diesel::prelude::*;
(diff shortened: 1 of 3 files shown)


src/controllers/user/update.rs
```diff
@@ -147,12 +147,16 @@ pub async fn regenerate_token_and_send(
...
             };
 
-            state.emails.send(&email.email, email1).map_err(Into::into)
-        })?;
-
-        ok_true()
+            state
+                .emails
+                .send(&email.email, email1)
+                .map_err(Into::<BoxedAppError>::into)
```
Turbo87, line 153: sending emails is still a sync operation, so I think this would need to stay within a `spawn_blocking()` block for now

eth3lbert: ah, great catch, thanks!

eth3lbert: Discarded changes.

Turbo87 (approved):

PR #9879, opened 2024-11-08
controllers/user/update: Reduce `spawn_blocking()` scope

Turbo87:
The email sending code is still sync and needs to be in a `spawn_blocking()` block for now, but the rest of the `regenerate_token_and_send()` can be converted to use async database queries.

Related:

- https://github.com/rust-lang/crates.io/pull/9877#discussion_r1834445148


eth3lbert (approved):
nice!

Turbo87:
Rebased this on top of #9888 to take advantage of the new tests and ensure that this does not unintentionally break anything.

PR #9892, opened 2024-11-10
admin/render_readmes: Use async `reqwest` client and `diesel-async` queries

Turbo87:
This should be our last place that still used a sync `reqwest` client :)

I tested this locally by hardcoding `cdn_prefix: Some("https://static.crates.io".to_string())` and then running `cargo run --bin crates-admin render-readmes --crate mustache --page-size 5`. seemed to work fine from what I could tell.


eth3lbert (approved):
LGTM 👍 !

We could further utilize [async-compression](https://crates.io/crates/async-compression) and [tokio-tar](https://crates.io/crates/tokio-tar) and remove `spawn_blocking()` in a future PR.

PR #9901, opened 2024-11-11
models/krate: Add `async_owners()` fn

Turbo87:
This is a clone of the `owners()` fn, and allows us to get rid of yet another `spawn_blocking()` block


eth3lbert (approved):
👍

PR #9926, opened 2024-11-14
database: Change `versions.crate_size` to be `NOT NULL`

Turbo87:
As mentioned on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/.60crate_size.60.20backfill/near/482313731), the `versions.crate_size` column in our database has been backfilled and does not contain any `NULL` values anymore.

This PR adjusts the column to be `NOT NULL` to avoid these values from being reintroduced in the future :)

Related:

- https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/.60crate_size.60.20backfill/near/482313731


eth3lbert (approved):
Since the backfill process has been completed, I believe this shouldn't cause any issues. For local env, IIUC, running `import-database-dump.sh` should include the backfilled values. Otherwise, users might need to perform their own backfill process.

Issue #1205, opened 2017-12-15
Bad gateway error when wanting to remove access for a team


est31:
I was trying to remove access for a team, then I got an error:
```
$ cargo owner --remove github:rustaudio:owners
    Updating registry `https://github.com/rust-lang/crates.io-index`
       Owner removing ["github:rustaudio:owners"] from crate deepspeech-sys
error: failed to remove owners from crate deepspeech-sys: failed to get a 200 OK response, got 502
headers:
    HTTP/1.1 100 Continue

    

    HTTP/1.1 502 Bad Gateway

    Connection: keep-alive

    Server: nginx

    Date: Fri, 15 Dec 2017 04:48:51 GMT

    Content-Type: text/html

    Content-Length: 166

    Strict-Transport-Security: max-age=31536000

    Via: 1.1 vegur

    

body:
<html>
<head><title>502 Bad Gateway</title></head>
<body bgcolor="white">
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

I'm both member of the team as well as owner of the crate.

carols10cents:
Strange. Because this is coming from nginx, I don't think the request is even getting to crates.io's logic around teams/members/owners.

I'm not able to reproduce. I see several 502s in the logs for deepspeech and deepspeeech-sys around when you posted this issue. I'm suspecting a transient network error, are you still able to reproduce consistently?

est31:
@carols10cents I was changing the team name from owners to cratesio-push and adding the cratesio-push team worked, but removing the owners team gave an error. Just now I've tried again and I'm still able to reproduce, so the error is consistent.

est31:
It seems the issue is not concentrated to removing access. Apparently yanking is affected as well! @centril just tried to yank something and got a 502 Bad Gateway too. The cratename is `proptest-quickcheck-interop` if you want to troubleshoot it in the logs.

Centril:
Yup, I got:

```
D:\programming\proptest\proptest-quickcheck-interop (master)
λ cargo yank --vers 1.0.0
    Updating registry `https://github.com/rust-lang/crates.io-index`
        Yank proptest-quickcheck-interop:1.0.0
error: failed to yank: failed to get a 200 OK response, got 502
headers:
    HTTP/1.1 502 Bad Gateway

    Connection: keep-alive

    Server: nginx

    Date: Sat, 16 Dec 2017 20:44:44 GMT

    Content-Type: text/html

[...]

Issue #1512, opened 2018-10-10
can't remove deleted team from crate owners


Xaeroxe:
We recently deleted the `github:amethyst:engine-devs` team, and moved most of its functionality to `github:amethyst:core`.

We can't remove the engine-devs team from our existing crates now though, when we attempt to we get the following error:

```
error: failed to remove owners from crate amethyst_animation

Caused by:
  api errors: only members of a team can add it as an owner
```

UPDATE: We resolved this on our own, but I'm leaving this issue open so crates.io can determine what to do with deleted teams.

Xaeroxe:
Come to think of it I'm not super sure we actually ran into an issue here, so I'm closing this.

PR #9930, opened 2024-11-14
Fix owner removal for duplicate logins

### LINKED ISSUE #2736: Can't remove invalid users from crate owners
This is similar to #1818 but for users.

## What happened

- A user had a GitHub account named, let's say, `foo`.
- Another user `someone_else` added them as an owner of a crate
- `foo` deleted their GitHub account (but crates.io has no real knowledge of this)
- `foo` recreated their GitHub account with the same username (because it was still available), but in this situation GitHub assigns them a new ID
- When `foo` now logs in to crates.io, it's a different account than their old `foo` crates.io account because of the new GitHub ID (and if it wasn't, this would be a security problem)
- The new `foo` account is NOT an owner of the crate
- `someone_else` can't add the new `foo` account as an owner, nor can they remove the old `foo` as an owner :(

## What should happen

- Removing an owner, whether it's valid/current or not, should always work (if performed by another valid owner, etc)
- Once that removal works, it should be possible to add the new account as an owner

Turbo87:
When users or teams on GitHub are deleted and recreated they might exist multiple times within our own database with the same `login`. When owners are deleted, only the newest matching `login` is considered for deletion though, which makes it impossible to delete older users/teams.

This commit fixes the problem by deleting **all** owners with a matching `login` instead.

Unfortunately, `diesel` does not support `UPDATE` queries with `JOINs`, so this has to be a "raw" SQL query with bindings instead…

Fixes #2736, ~~and AFAICT also fixes https://github.com/rust-lang/crates.io/issues/1205~~

Related:

- https://github.com/rust-lang/crates.io/issues/1205
- #2736
- https://github.com/rust-lang/crates.io/issues/1818
- https://github.com/rust-lang/crates.io/issues/1512
- https://github.com/rust-lang/crates.io/pull/7051

### DIFF
--- a/src/models/krate.rs
+++ b/src/models/krate.rs
@@ -2,7 +2,7 @@ use chrono::NaiveDateTime;
 use diesel::associations::Identifiable;
 use diesel::dsl;
 use diesel::pg::Pg;
-use diesel::sql_types::{Bool, Text};
+use diesel::sql_types::{Bool, Integer, Text};
 use diesel_async::AsyncPgConnection;
 use secrecy::SecretString;
 use thiserror::Error;
@@ -18,7 +18,7 @@ use crate::schema::*;
 use crate::sql::canon_crate_name;
 use crate::util::diesel::prelude::*;
 use crate::util::diesel::Conn;
-use crate::util::errors::{version_not_found, AppResult};
+use crate::util::errors::{bad_request, version_not_found, AppResult};
 use crate::{app::App, util::errors::BoxedAppError};
 
 use super::Team;
@@ -457,12 +457,44 @@ impl Crate {
     pub fn owner_remove(&self, conn: &mut impl Conn, login: &str) -> AppResult<()> {
         use diesel::RunQueryDsl;
 
-        let owner = Owner::find_by_login(conn, login)?;
+        let query = diesel::sql_query(
+            r#"WITH crate_owners_with_login AS (
+                SELECT
+                    crate_owners.*,
+                    CASE WHEN crate_owners.owner_kind = 1 THEN
+                         teams.login
+                    ELSE
+                         users.gh_login
+                    END AS login
+                FROM crate_owners
+                LEFT JOIN teams
+                    ON crate_owners.owner_id = teams.id
+                    AND crate_owners.owner_kind = 1
+                LEFT JOIN users
+                    ON crate_owners.owner_id = users.id
+                    AND crate_owners.owner_kind = 0
+                WHERE crate_owners.crate_id = $1
+                    AND crate_owners.deleted = false
+            )
+            UPDATE crate_owners
+            SET deleted = true
+            FROM crate_owners_with_login
+            WHERE crate_owners.crate_id = crate_owners_with_login.crate_id
+                AND crate_owners.owner_id = crate_owners_with_login.owner_id
+                AND crate_owners.owner_kind = crate_owners_with_login.owner_kind
+                AND lower(crate_owners_with_login.login) = lower($2);"#,
+        );
 
-        let target = crate_owners::table.find((self.id(), owner.id(), owner.kind()));
-        diesel::update(target)
-            .set(crate_owners::deleted.eq(true))
+        let num_updated_rows = query
+            .bind::<Integer, _>(self.id)
+            .bind::<Text, _>(login)
             .execute(conn)?;
+
+        if num_updated_rows == 0 {
+            let error = format!("could not find owner with login `{login}`");
+            return Err(bad_request(error));
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


carols10cents:
Wait, what about this case:

- User `foo`, github ID 1 that is Person A has logged in to crates.io 
- Person A renames their github account to `bar` but does NOT log in to crates.io
- Person B creates a new github account named `foo` with github ID 2 and logs in to crates.io
- Person A logs in to crates.io with their account that's now named `bar` but still has github ID 1, I believe the current behavior is crates.io updates the gh_login and everything Just Works because the gh_id matches

I think what you're saying is this PR would delete the record for Person A when Person B logs in??? So when Person A logs in again, their account/ownership stuff would be gone?

carols10cents:
OH WAIT I think I'm confused. I read this line:

> which makes it impossible to delete older users/teams.

and thought the records in the `users` table were being deleteed, but you're talking about the `crate_owners` table. Carry on, I think.

Turbo87:
> and thought the records in the `users` table were being deleted, but you're talking about the `crate_owners` table

yep, exactly :)

src/models/krate.rs
```diff
@@ -453,12 +453,43 @@ impl Crate {
...
-            .set(crate_owners::deleted.eq(true))
+        let num_updated_rows = query
+            .bind::<Integer, _>(self.id)
+            .bind::<Text, _>(login)
             .execute(conn)?;
+
+        if num_updated_rows == 0 {
+            let error = format!("could not find owner with login `{login}`");
+            return Err(bad_request(error));
+        }
```
eth3lbert, line 491: Should we add a `NOT deleted` condition here? However, it might be a bad idea to work with the `num_updated_rows == 0` check below.

src/tests/routes/crates/owners/remove.rs
```diff
@@ -78,5 +81,73 @@ async fn test_unknown_team() {
...
+
+    assert_eq!(github_ids.len(), 2);
+    assert_ne!(github_ids[0], github_ids[1]);
+
+    // - The new `foo` account is NOT an owner of the crate
+    let owner_ids = crate_owners::table
+        .filter(crate_owners::crate_id.eq(krate.id))
+        .filter(crate_owners::deleted.eq(false))
+        .select(crate_owners::owner_id)
+        .load::<i32>(&mut conn)?;
```
eth3lbert, line 131: Not a blocker, but should we replace this with something like
```suggestion
    let owner_ids = CrateOwner::by_owner_kind(OwnerKind::User)
        .filter(crate_owners::crate_id.eq(krate.id))
        .select(crate_owners::owner_id)
        .load::<i32>(&mut conn)?;
         // or something that map from
         // krate.owners(&mut conn)?
```
 as we usually use these?

eth3lbert (commented):
LGTM on first glance :)

src/models/krate.rs
Turbo87: > Should we add a `NOT deleted` condition here?

hmm, yeah, that seems like a good idea.

Turbo87:
> should we add some tests for the team-related parts?

I've added https://github.com/rust-lang/crates.io/pull/9943 just now, but it looks like there is no fix needed for that issue since it already works as intended.

eth3lbert:
> Wait, what about this case:
> 
>     * User `foo`, github ID 1 that is Person A has logged in to crates.io
> 
>     * Person A renames their github account to `bar` but does NOT log in to crates.io
> 
>     * Person B creates a new github account named `foo` with github ID 2 and logs in to crates.io
> 
>     * Person A logs in to crates.io with their account that's now named `bar` but still has github ID 1, I believe the current behavior is crates.io updates the gh_login and everything Just Works because the gh_id matches
> 
> 
> I think what you're saying is this PR would delete the record for Person A when Person B logs in??? So when Person A logs in again, their account/ownership stuff would be gone?

If I understand correctly, in this case, Person B is not the owner of the crate, and they also do not have permission to change the owner. So, there should not be an issue here.

src/tests/issues/issue2736.rs
```diff
@@ -0,0 +1,63 @@
...
+    let owners = krate.owners(&mut conn)?;
+    assert_eq!(owners.len(), 2);
+    assert_none!(owners.iter().find(|o| o.id() == foo2.as_model().id));
+
+    // - `someone_else` can't add the new `foo` account as an owner, nor can they remove the old `foo` as an owner :(
+    let url = "/api/v1/crates/crate1/owners";
+    let payload = json!({ "owners": ["foo"] }).to_string();
+    let response = someone_else.delete_with_body::<()>(url, payload).await;
+    assert_eq!(response.status(), StatusCode::OK);
+    assert_snapshot!(response.text(), @r#"{"msg":"owners successfully removed","ok":true}"#);
```
eth3lbert, line 56: The comment said, "Nor can they remove the old foo as an owner :("

but the response was `OK` ;)

Turbo87: fixed

eth3lbert (approved):
The regression test confirms that the issue has been fixed as expected. The SQL part leverages `bind`, which should prevent any injection attacks.
Nice work, thanks 👍

Issue #1541, opened 2018-10-26
Display "edition" from manifest on crate-page


lukaslueg:
[This](https://crates.io/crates/actix-web-async-await) is an example of a crate that has an explicit `edition = "2018"` in it's manifest. It would be nice if the crate-page would show a marker/banner/hint that this crate expects a 2018-environment

Mark-Simulacrum:
This just requires a recent enough compiler so I don't think this is something we should do this as it's really no different than "this crate requires 1.31" which is a separate discussion (there have been multiple RFCs on that)

carols10cents:
@lukaslueg and @Eh2406, could you elaborate a bit on why this would be nice/useful and different from the compiler version the crate requires? 

Given a new enough compiler, you can depend on crates that use either edition without needing to care which edition the dependency uses, barring some potential keywords used as functions in the public API that would need the raw identifier syntax.

Or do you mean as a potential contributor to the crate, rather than a potential user of it?

In other words, if this information was displayed on crates.io, what would you do differently based on it?

Eh2406:
I had not fully thought this thure when I emoted, having read the responses I agree that it is just a weaker form of MSRV.

lukaslueg:
Well, this is actually a weak argument in the first place: While there is an ongoing discussion on how to communicate rustc version requirements, the `edition`-key in the manifest is as clear as it could: Either it's  there (because you have to have it in order to use `2018`) or it's not there (which gets us "needs rust 1.28ish"). Since it's also readily available from the manifest, it's very easy to implement without any kind of ambiguity. Besides - and the arguments get even weaker here - the whole `edition` concept is probably less known to the wider audience than one may think. A hint on `crates.io` could help communicate "hey, this is a thing!"

meven:
I think it would make sense, for the users to know about editions as @lukaslueg mentioned and to advertise the lattest edition (in case someone missed the information) and slightly encourage maintainers to migrate to a more recent edition because it would grant their crate the edtion 2018 badge.
But also because a byproduct of this could be to have statistics about the number of "edtion 2018" crates and crates "edtion 2015".

kornelski:
FWIW I've implemented that on crates.rs: https://crates.rs/crates/actix-web-async-await

<img width="176" alt="screenshot 2019-01-09 at 15 46 45" src="https://user-images.githubusercontent.com/72159/50910483-ca300180-1425-11e9-93c0-76c760599ecf.png">

[...]

PR #9932, opened 2024-11-14
Add `versions.edition` column to the database and expose it on the API

Turbo87:
This addresses part of https://github.com/rust-lang/crates.io/issues/1541 by reading the `edition` field of uploaded `Cargo.toml` files, saving the value in the database and then exposing it on the API.

Note that this does not include any backfilling for previously uploaded crates. Should we move the API part to a dedicated PR and merge it only after we have backfilled the data?

Related:

- https://github.com/rust-lang/crates.io/issues/1541
- https://github.com/LukeMathWalker/cargo-manifest/pull/58


crates/crates_io_tarball/src/manifest.rs
```diff
@@ -84,3 +84,17 @@ impl IsInherited for DepsSet {
...
+
+impl EditionExt for cargo_manifest::Edition {
+    fn as_str(&self) -> &'static str {
+        match self {
+            Self::E2015 => "2015",
+            Self::E2018 => "2018",
+            Self::E2021 => "2021",
+        }
+    }
+}
```
Turbo87, line 100: this is essentially a polyfill for https://github.com/LukeMathWalker/cargo-manifest/pull/58 :)

PR #9936, opened 2024-11-14
boot/categories: Migrate `sync_with_connection()` to `diesel-async` queries

Turbo87:
... to move more code out of `spawn_blocking()` blocks :)

### DIFF
--- a/src/boot/categories.rs
+++ b/src/boot/categories.rs
@@ -1,10 +1,12 @@
 // Sync available crate categories from `src/categories.toml`.
 // Runs when the server is started.
 
-use crate::util::diesel::prelude::*;
 use anyhow::{Context, Result};
+use crates_io_database::schema::categories;
+use diesel::pg::upsert::excluded;
+use diesel::prelude::*;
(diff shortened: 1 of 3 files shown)


src/boot/categories.rs
```diff
@@ -76,10 +77,10 @@ fn categories_from_toml(
     Ok(result)
 }
 
-pub fn sync_with_connection(toml_str: &str, conn: &mut impl Conn) -> Result<()> {
+pub async fn sync_with_connection(toml_str: &str, conn: &mut AsyncPgConnection) -> Result<()> {
     use crate::schema::categories;
     use diesel::pg::upsert::excluded;
-    use diesel::RunQueryDsl;
+    use diesel_async::RunQueryDsl;
```
eth3lbert, line 83: Should this be used globally within this file now?

eth3lbert (approved):

src/boot/categories.rs
Turbo87: yep, makes sense. updated :)

PR #9937, opened 2024-11-14
typosquat: Asyncify all the things

Turbo87:
... to get rid of these pesky sync database connections.

I wish it would've been possible to split this into multiple commits, but... no 😅

### DIFF
--- a/src/typosquat/database.rs
+++ b/src/typosquat/database.rs
@@ -5,8 +5,9 @@ use std::{
     collections::{BTreeMap, HashMap, HashSet},
 };
 
-use crate::util::diesel::Conn;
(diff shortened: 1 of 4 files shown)


src/typosquat/database.rs
```diff
@@ -42,45 +42,51 @@ impl TopCrates {
...
+                        krate.name,
+                        Crate {
+                            owners: HashSet::new(),
+                        },
+                    ),
+                );
+
+                futures_util::future::ready(Ok(crates))
+            })
+            .await?;
```
LawnGnome, line 65: This is just straight up better, async or no. :+1:

LawnGnome (approved):
:+1:

PR #9973, opened 2024-11-16
rss: Remove explicit/duplicate XML escaping

Turbo87:
XML tags inside `CDATA` sections don't need to be escaped, so this explicit escaping step led to an unnecessary duplicate escaping being applied.

Related:

- https://github.com/rust-syndication/rss/issues/167#issuecomment-2480138799

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1087,7 +1087,6 @@ dependencies = [
  "paste",
  "postgres-native-tls",
  "prometheus",
- "quick-xml 0.37.0",
  "rand",
  "regex",
  "reqwest 0.12.9",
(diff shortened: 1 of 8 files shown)


eth3lbert (commented):
According to  https://en.wikipedia.org/wiki/CDATA#Nesting, if I understand correctly, we need to encode `]]>` if description contains it.

Turbo87:
I added another commit that explicitly checks for `]]>` escaping and it looks like you discovered a bug in the rss/quick-xml crate(s). It does not appear to get escaped properly. I'll report it over there and see if I can fix it.

eth3lbert (approved):

Turbo87:
Reported at:

- https://github.com/rust-syndication/rss/issues/173
- https://github.com/tafia/quick-xml/issues/831

Turbo87:
I'll merge this PR for now until the issue is fixed upstream :)

Issue #953, opened 2017-08-12
Inconsistent crate info display (mixing versions)


bluss:
About: https://crates.io/crates/petgraph

The most recently released version of petgraph is 0.2.10.

The greatest version is petgraph 0.4.5.

This is inconsistent, it should show all the information from the same version:

* _Documentation_ is taken from 0.2.10, a version that is not visible anywhere on the page.
* _Categories_ (or the lack there of) seems to be taken from 0.2.10
* _Dependencies_ are taken from petgraph 0.4.5. 
* _Last Updated_ is taken from 0.2.10

Unknown because they are the same across versions: License, Authors, Keywords, Repository, etc.

carols10cents:
> Documentation is taken from 0.2.10, a version that is not visible anywhere on the page.

The documentation URL that's provided to crates.io is http://bluss.github.io/petgraph, without any version information. Since crates.io doesn't have anything to do with the generation of those docs, I don't think there's anything to be done here about that. I do think perhaps we could improve cargo tooling on that front, [as I mused in this issue](https://github.com/rust-lang/crates.io/issues/894).

> Categories (or the lack there of) seems to be taken from 0.2.10

Categories are per-crate, not per-version, so the latest published version is what takes effect. Is there a use case for wanting a crate to be categorized differently based on the version?

> Dependencies are taken from petgraph 0.4.5.

This seems correct to me.

> Last Updated is taken from 0.2.10

It seems like you're saying "last updated" should be 0.4.5 and should say something like "highest version published at", is that correct? It doesn't seem that strange to me to have the last time the entire crate was updated shown here.... what if we labeled this as "crate last updated at"?

bluss:
The crate information page says "petgraph 0.4.5" — then it should use the documentation link from the 0.4.5 version of the package.

bluss:
> It seems like you're saying "last updated" should be 0.4.5 and should say something like "highest version published at", is that correct? It doesn't seem that strange to me to have the last time the entire crate was updated shown here.... what if we labeled this as "crate last updated at"?

I think the core of the problem is that the latest published version 0.2.10 is not visible anywhere on the page, so the information about "Last Updated" becomes very puzzling and hard to put together with the other information on the page. It would help to just put the version of the last update in this spot, if it's different from the "main" version that crates.io is displaying.

carols10cents:
> The crate information page says "petgraph 0.4.5" — then it should use the documentation link from the 0.4.5 version of the package.

[...]

PR #9998, opened 2024-11-19
Add description, homepage, documentation and repository fields for versions

Turbo87:
We currently save these fields only per crate, not per version. This means that for crates publishing bug fixes to older major versions, these fields might get overwritten with outdated metadata.

The solution to this is to read these per crate metadata fields from the "default version" of a crate instead.

This PR introduces four more metadata columns to the `versions` table (`description`, `homepage`, `documentation` and `repository`) and exposes their values on the API. The rest of the code is not using these values yet since they will need to be backfilled from the stored crate files first. Once that has happened we could remove the fields from the `crates` table and use the content from the default version of a crate instead.

Potential downsides of this:

- database size: since we are storing mostly similar data for each version this might grow the database size considerably.
- database dump size: same as above, though compression should hopefully get rid of most of the overhead.
- database dump breaking change: if we remove the existing columns of the `crates` table, or if we stop updating them, this will essentially be a breaking change to the database dump schema. while we should not see this as a hard blocker, we should be careful to not introduce breaking changes too often.
- more joins in queries needed: this might have a performance impact, but with the existing indexes I expect this to be manageable, and we're already pulling in the default version for the relevant endpoint anyway, so the additional overhead might actually be zero.

Related:

- https://github.com/rust-lang/crates.io/issues/953
- https://github.com/rust-lang/crates.io/discussions/7560


Turbo87:
RE database dump breaking change: I guess we could also denormalize the data from the `versions` to the `crates` table whenever we update the default version 🤔

Turbo87:
7 days without any objections and we talked about this PR in the crates.io team meeting last week, also without any objections, so let's merge this :)

PR #10015, opened 2024-11-20
admin/delete-crate: Use configurable 24 hour availability delay

Turbo87:
As discussed in https://github.com/rust-lang/crates.io/pull/9912 npm uses a 24 hour delay until a package name becomes available again after deletion. This PR introduces the same default for our `delete-crate` admin tool, but also lets us override the value with the `--availability-delay XXX` CLI option.

Related:

- https://github.com/rust-lang/crates.io/pull/9912


Eh2406:
Thank you! This will be very helpful in my roll at $day_job.

PR #10016, opened 2024-11-20
Add minimum version tracking to deleted crates

LawnGnome:
As discussed in #9904 and #9912, this PR:

1. Adds a column to the `deleted_crates` table that indicates the first version that can be used when publishing a new crate version with the same name.
2. Populates that column when `crates-admin delete-crate` is run.
3. Checks that column when publishing a new crate version.

---

Implementation wise, most of the noise here is actually moving the `deleted_crates` publish checks into a sub-module, mostly because the original version I wrote inline in the controller was ugly as hell. Instead, we now get to enjoy a slightly overengineered state machine. I'm open to suggestions.

(Although, on the bright side, if we ever do need [the more complex semver compatibility check](https://github.com/rust-lang/crates.io/pull/9912#issuecomment-2484477233) rather than just checking against a minimum version, having the validation logic separated out already should make that easier to implement.)

I added new integration tests for the controller, but most of the specifics are tested in unit tests in said new sub-module. Redundant? Probably. :shrug:

As for the deletion side of things, we're going to need the same `min_version` population behaviour (specifically the use of `TopVersions`) that I added to `crates-admin` when #9904 lands, but I was struggling a bit to figure out how to integrate that into `models::krate`, since `crates-admin` never actually constructs a `models::krate::Crate` when deleting a crate. I played around with doing a partial build of `NewDeletedCrateBuilder` based on a `Crate`, but the type signatures were driving me slightly bonkers, and it didn't really feel cleaner anyway, so I just did it inline for now.

Honestly, it might be simple enough that it's easiest to just copy/paste/adapt the same general logic into #9904.

---

I have a couple of other, more specific open questions that I'll throw into the inline review after I publish this, but I guess the main question here is whether this feels directionally correct.

### DIFF
--- a/src/controllers/krate/publish/deleted.rs
+++ b/src/controllers/krate/publish/deleted.rs
@@ -1,0 +1,264 @@
+use chrono::{DateTime, SecondsFormat, Utc};
+use diesel_async::AsyncPgConnection;
+use http::StatusCode;
+use semver::Version;
+
+use crate::{
+    schema::deleted_crates,
+    sql::canon_crate_name,
+    util::{
+        diesel::prelude::*,
+        errors::{custom, AppResult},
+    },
+};
+
+/// Checks the given crate name and version against the deleted crates table,
+/// ensuring that the crate version is allowed to be published.
+///
+/// If the crate version cannot be published, a
+/// [`crate::util::errors::BoxedAppError`] will be returned with a user-oriented
+/// message suitable for being returned directly by a controller.
+pub async fn validate(
+    conn: &mut AsyncPgConnection,
+    name: &str,
+    version: &Version,
+) -> AppResult<()> {
+    use diesel_async::RunQueryDsl;
+
+    // Since the same crate may have been deleted multiple times, we need to
+    // calculate the most restrictive set of conditions that the crate version
+    // being published must adhere to; specifically: the latest available_at
+    // time, and the highest min_version.
+    let mut state = State::default();
+
+    // To do this, we need to iterate over all the relevant deleted crates.
+    for (available_at, min_version) in deleted_crates::table
+        .filter(canon_crate_name(deleted_crates::name).eq(canon_crate_name(name)))
+        .select((deleted_crates::available_at, deleted_crates::min_version))
+        .load::<(DateTime<Utc>, Option<String>)>(conn)
+        .await?
+    {
+        state.observe_available_at(available_at);
+
+        // We shouldn't really end up with an invalid semver in the
+        // `min_version` field, so we're going to silently swallow any errors
+        // for now.
+        if let Some(Ok(min_version)) = min_version.map(|v| Version::parse(&v)) {
+            state.observe_min_version(min_version);
+        }
+    }
+
+    // Finally, we can check the given name and version against the built up
+    // state and see if it passes.
+    state.into_result(name, version, Utc::now())
+}
+
+#[derive(Default)]
+#[cfg_attr(test, derive(Clone))]
+struct State {
+    available_at: Option<DateTime<Utc>>,
+    min_version: Option<Version>,
+}
+
+impl State {
+    fn observe_available_at(&mut self, available_at: DateTime<Utc>) {
+        if let Some(current) = self.available_at {
+            self.available_at = Some(std::cmp::max(current, available_at));
+        } else {
+            self.available_at = Some(available_at);
+        }
+    }
+
+    fn observe_min_version(&mut self, min_version: Version) {
+        if let Some(current) = self.min_version.take() {
+            self.min_version = Some(std::cmp::max(current, min_version));
+        } else {
+            self.min_version = Some(min_version);
+        }
+    }
+
+    fn into_result(self, name: &str, version: &Version, now: DateTime<Utc>) -> AppResult<()> {
+        let mut messages = Vec::new();
+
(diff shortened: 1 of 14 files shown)


src/controllers/krate/publish/deleted.rs
```diff
@@ -0,0 +1,270 @@
...
+                    .into(),
+            )
+            .unwrap()
+        }};
+    }
+
+    macro_rules! assert_result_failed {
+        ($result:expr) => {{
+            let text = assert_result_status!($result);
+            assert_snapshot!(text);
```
LawnGnome, line 143: I decided to write the snapshots to files because long lines make my eye twitch, but I don't feel very strongly about this.

src/controllers/krate/publish/deleted.rs
```diff
@@ -0,0 +1,270 @@
...
+            assert_eq!(response.status(), StatusCode::UNPROCESSABLE_ENTITY);
+
+            String::from_utf8(
+                axum::body::to_bytes(response.into_body(), usize::MAX)
+                    .await
+                    .unwrap()
+                    .into(),
+            )
+            .unwrap()
+        }};
```
LawnGnome, line 137: I thought about extending `tests::util` to be able to use the same response helpers on a bare `axum::response::Response` without the full boilerplate to fake a full request, but decided to keep this more self-contained for now. I can bring that code back if that sounds useful, though.

src/controllers/krate/publish/deleted.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        if messages.is_empty() {
+            Ok(())
+        } else {
+            Err(custom(
+                StatusCode::UNPROCESSABLE_ENTITY,
+                format!(
+                    "A crate with the name `{name}` was previously deleted.\n\n* {}",
+                    messages.join("\n* "),
```
LawnGnome, line 111: This is basically the shitty version of our conversation about representing multiple errors in a single response in #9904. :hankey:

LawnGnome:
> as mentioned in [#9912 (comment)](https://github.com/rust-lang/crates.io/pull/9912#issuecomment-2482479967), I'm curious what the situation at other registries looks like regarding this situation, for user-issued deletions, but also when admins delete packages/versions for whatever reason.

All right, all right, I'll go actually ask some people.

_grumbles about this being as bad as having to phone someone on an actual phone_

LawnGnome:
> one potential issue with this I discovered yesterday: when someone name-squats `foo@100000000000000.0.0` and deletes the crate again. would that now cause others to only be able to publish `foo@100000000000001.0.0` and upward from there? 🫣

So, yeah, that's an issue.

On the other hand, there's also kind of an issue here with the "specifically allow semver-incompatible versions" rule as well, which is that it's potentially hard to explain. If someone publishes, say: `0.1.0, 0.2.0, 0.4.0, 1.0.0, 3.0.0`, then deletes the crate, then a new maintainer would only be able to publish `0.3.x, 0.5.x, 2.x.y, 4.x.y`, and so on. Which is probably going to be challenging to communicate effectively in an error message.

Of course, a crate having that set of versions in the first place might be a sign that the name is associated with someone with... unusual ideas on versioning. :laughing:

Rustin170506:
> I guess now comes the question: is it worth it? 😂

Here are two very specific examples. I think if we allow version re-release, we will encounter these two issues. However, if we don't allow re-releasing the same version, perhaps the two issues will merge into one: it will just become "not found."

https://github.com/rust-lang/cargo/issues/10071
https://github.com/rust-lang/cargo/issues/10063

Turbo87:
IMHO these seem like bugs on the cargo side, though admittedly they are caused by the assumption mismatch that versions/crates will just never ever disappear from crates.io, which for various reasons we cannot guarantee.

but again, with the set of requirements that we enforce for people to be able to delete crates themselves, it seems unlikely for people to regularly hit these sorts of issues. I'm not saying it will never happen, but I don't see it as being a very likely scenario for most people.

Rustin170506:
> IMHO these seem like bugs on the cargo side

Yes, I'm not suggesting that crates.io completely fix these two issues. I'm just saying that, as the default registry for Cargo, we have an opportunity to reduce the likelihood of these situations occurring. Because, as you can see, fixing mismatches requires significant effort from Cargo (e.g., modifying the fingerprints). So I'm wondering if we could reduce the chances of this happening in the default registry, wouldn't that minimize the potential impact on users as much as possible?

Turbo87:
> I'm wondering if we could reduce the chances of this happening in the default registry, wouldn't that minimize the potential impact on users as much as possible?

yeah, we could. but ultimately it's a tradeoff that needs to be balanced with all of the other arguments for and against this that were brought up in the comments of this PR (and the others).

I'm personally not convinced yet that it's worth the extra complexity for our (publishing) users, but ultimately the crates.io team needs to decide this.

LawnGnome:
> I'm personally not convinced yet that it's worth the extra complexity for our (publishing) users, but ultimately the crates.io team needs to decide this.

I see this is already on the agenda for Friday, so I guess we can defer this until then.

LawnGnome:
OK, so we discussed this at the team meeting, and I think I'm convinced enough that the restrictions on crate deletion in the RFC are sufficient to make the risk low enough that we can just scan for this ex post facto, thereby sidestepping awkward questions around how we communicate this to users.

Closing.

PR #10041, opened 2024-11-21
controllers/krate/publish: Migrate to `diesel-async` queries

Turbo87:
finally...

now, unfortunately we still have a `spawn_blocking()` call in the endpoint since I didn't feel comfortable with processing the tarball on the async worker thread, but at least we're not using the `AsyncConnectionWrapper` wrapper anymore 🎉 

This will allow us to throw away a bunch of duplicated code in some of the follow-up PRs :)


src/controllers/krate/publish.rs
eth3lbert, line 1: I think we can move back to `use diesel::prelude::*` and flatten https://github.com/rust-lang/crates.io/blob/2bad84cb8a65c7bcbd05bf9ce080eb2ed52f756a/src/controllers/krate/publish.rs#L86-L124 now that we've converted this file to `diesel-async`.

eth3lbert (approved):
LGTM 👍

PR #10064, opened 2024-11-23
Migrate `crates_io_tarball` to async/await

eth3lbert:
This PR migrates crates_io_tarball to async/await, allowing us to remove the `spawn_blocking()` call in the publish endpoint :)


Cargo.toml
```diff
@@ -122,6 +122,7 @@ tracing-subscriber = { version = "=0.3.18", features = ["env-filter", "json"] }
 typomania = { version = "=0.1.2", default-features = false }
 url = "=2.5.4"
 unicode-xid = "=0.2.6"
+krata-tokio-tar = "=0.4.2"
```
Turbo87, line 125: I assume this is a fork and there is a reason for not using the original?

crates/crates_io_tarball/src/builder.rs
```diff
@@ -47,3 +50,53 @@ impl AsMut<tar::Builder<Vec<u8>>> for TarballBuilder {
         &mut self.inner
     }
 }
+
+pub struct AsyncTarballBuilder {
```
Turbo87: do we really need the async builder? is the sync one not enough for our test suite purposes?

Cargo.toml
eth3lbert: Since the original `tokio-tar` seems unmaintained, and I know it's used by `uv` and has been reviewed!

crates/crates_io_tarball/src/builder.rs
eth3lbert: Indeed and fair enough! I implemented it without realizing it was primarily for testing 😅 
And since it's only for test purposes, I agree that we don't really need an async builder.

Turbo87: 7 changed files certainly looks easier to review 😂 

I'll take a look tomorrow :)

Turbo87 (approved):

Turbo87:
LGTM! nice work :)

Turbo87:
rebased this on top of https://github.com/rust-lang/crates.io/pull/10068 to fix the `cargo-deny` issue (i.e. `rustls` vulnerability)

PR #10069, opened 2024-11-25
controllers/krate/publish: Validate JSON metadata before reading tarball

Turbo87:
Previously, the `BytesRequest` allocated memory for the full tarball (and the JSON metadata blob) before even validating that the `name` and `vers` fields of the JSON metadata correspond to a crate and version that the user has publish access too. This commit changes the code to first read the JSON metadata from the request body stream, validate it, and then read the tarball bytes afterwards.

Unfortunately, this means that we still buffer the full tarball in memory, but since we only want to upload the tarball to S3 once we validated the content of the tarball, this is somewhat hard to fix. Ideas are welcome! 😉

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -146,39 +146,14 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
         .check_rate_limit(auth.user().id, rate_limit_action, &mut conn)
         .await?;
 
-    let tarball_len = reader.read_u32_le().await.map_err(|e| {
-        if e.kind() == std::io::ErrorKind::UnexpectedEof {
-            bad_request("invalid tarball length")
-        } else {
-            e.into()
-        }
-    })?;
-    let content_length = tarball_len as u64;
-
     let maximums = Maximums::new(
         existing_crate.as_ref().and_then(|c| c.max_upload_size),
         app.config.max_upload_size,
         app.config.max_unpack_size,
     );
 
-    if content_length > maximums.max_upload_size {
-        return Err(custom(
-            StatusCode::PAYLOAD_TOO_LARGE,
-            format!("max upload size is: {}", maximums.max_upload_size),
-        ));
-    }
-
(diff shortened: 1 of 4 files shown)


eth3lbert:
nice, I'll have a look tomorrow :D

src/controllers/krate/publish.rs
```diff
@@ -633,6 +608,36 @@ async fn read_json_metadata(
         .map_err(|e| bad_request(format_args!("invalid upload request: {e}")))
 }
 
+async fn read_tarball_bytes(
+    reader: &mut (impl AsyncReadExt + Unpin),
+    max_length: u32,
+) -> Result<Bytes, BoxedAppError> {
```
eth3lbert, line 614: I think `use tokio::io::AsyncReadExt` with the following should be enough?
```suggestion
async fn read_tarball_bytes<R: AsyncRead + Unpin>(
    reader: &mut R,
    max_length: u32,
) -> Result<Bytes, BoxedAppError> {
```

src/controllers/krate/publish.rs
```diff
@@ -631,6 +602,37 @@ async fn count_versions_published_today(
         .await
 }
 
+async fn read_json_metadata(
+    reader: &mut (impl AsyncReadExt + Unpin),
+    max_length: u32,
+) -> Result<PublishMetadata, BoxedAppError> {
```
eth3lbert, line 583: I the `use tokio::io::AsyncReadExt` with the following should be enough?
```suggestion
async fn read_json_metadata<R: AsyncRead + Unpin>(
    reader: &mut R,
    max_length: u32,
) -> Result<PublishMetadata, BoxedAppError> {
```

eth3lbert (approved):
LGTM 👍

eth3lbert:
> Unfortunately, this means that we still buffer the full tarball in memory, but since we only want to upload the tarball to S3 once we validated the content of the tarball, this is somewhat hard to fix. Ideas are welcome!

Yeah, it would be great if we could stream uploads to S3 and abort them if they're invalid. But this might incur additional costs(?).

PR #10070, opened 2024-11-25
admin/render_readmes: Migrate `render_pkg_readme()` and `find_file_by_path()` fns to async/await

eth3lbert:
This PR migrate `render_pkg_readme()` and `find_file_by_path()` fns to async/await.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ doctest = true
 
 [dependencies]
 anyhow = "=1.0.93"
+async-compression = { version = "=0.4.18", default-features = false, features = ["gzip", "tokio"] }
 async-trait = "=0.1.83"
 aws-credential-types = { version = "=1.2.1", features = ["hardcoded-credentials"] }
 aws-ip-ranges = "=0.926.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -122,6 +122,9 @@ tracing-subscriber = { version = "=0.3.18", features = ["env-filter", "json"] }
 typomania = { version = "=0.1.2", default-features = false }
 url = "=2.5.4"
 unicode-xid = "=0.2.6"
+async-compression = { version = "=0.4.18", default-features = false, features = ["gzip", "tokio"] }
+krata-tokio-tar = "=0.4.2"
+tokio-util = { version = "=0.7.12", features = ["compat"] }
```
Turbo87, line 127: can you add them with alphabetic order? 🙏

Turbo87 (commented):
is the `text_to_html()` call fast enough to be performed in an async worker thread? I'm wondering if we should keep it in a `spawn_blocking()` call

Cargo.toml
eth3lbert: sure!

eth3lbert:
> is the `text_to_html()` call fast enough to be performed in an async worker thread? I'm wondering if we should keep it in a `spawn_blocking()` call

Good call. I'll answer this question after I experiment with it after I get up :)

Turbo87:
I also kept it in `spawn_blocking()` in the corresponding background job for now: https://github.com/rust-lang/crates.io/blob/3bfe1cf5dca49c8db1a4aa88cef481d090381c28/src/worker/jobs/readmes.rs#L54-L62

eth3lbert:
After benchmarking, I believe we need to keep `spawn_blocking()`. Most markdown files should not take longer than `6ms`, but there might be some files that aren't large but still take a lot of time(maybe seconds), which could cause blocking.

eth3lbert:
Ah, it seems I didn't commit the modified part correctly for some reason.
I've rebased to the latest main and made slight modifications to align with our recent work.

Turbo87 (approved):

PR #10087, opened 2024-11-27
Enqueue background jobs in parallel

Turbo87:
This PR adjusts our `BackgroundJob::enqueue()` fn to be able to take advantage of Postgres query pipelining and then uses it in various places to reduce the amount of waiting for queries to be finished.

Related:

- https://github.com/rust-lang/crates.io/pull/9663
- https://github.com/rust-lang/crates.io/pull/10082
- https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/topic/postgres.20pipelining


LawnGnome (approved):
This LGTM, but I also am not going to pretend to be fully across all the nuances of async Diesel.

PR #10097, opened 2024-11-28
database_dump: Fix `num_no_build` import issue

Turbo87:
The `num_no_build` column in the database is `NOT NULL`, but since we don't export it in the database dump it will show up as an empty value and thus fail to get imported correctly. This commit temporarily drops the `NOT NULL` constraint during the data import, backfills the data from the `num` column, and then adds the constraint again.

Related:

- https://github.com/rust-lang/crates.io/pull/9767#issuecomment-2503128672
- https://github.com/rust-lang/crates.io/pull/9786

### DIFF
--- a/crates/crates_io_database_dump/src/dump-import.sql.j2
+++ b/crates/crates_io_database_dump/src/dump-import.sql.j2
@@ -17,12 +17,55 @@ BEGIN;
 {%- endfor %}
 
     -- Enable this trigger so that `crates.textsearchable_index_col` can be excluded from the export
+
     ALTER TABLE "crates" ENABLE TRIGGER "trigger_crates_tsvector_update";
+
+    -- Drop the `NOT NULL` and uniqueness constraints on `num_no_build` column
+    -- while importing the data.
+
+    alter table "versions"
+        alter column "num_no_build" drop not null;
+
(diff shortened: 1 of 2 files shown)


Turbo87:
This seems to fix the issue, but now that I've written the code I'm questioning the brittleness of this approach vs. reverting https://github.com/rust-lang/crates.io/pull/9786.

> because in downstream code it's trivial to reconstruct from the value of `num` if needed.

unfortunately, we have 797 versions where it is not trivial, because we have duplications in there. fixing these duplications is possible (see this PR or https://github.com/rust-lang/crates.io/pull/9767#issuecomment-2505277768), but 1) does it result in the same data as in the production database? and 2) is it worth the brittleness?

I think I'm leaning towards reverting https://github.com/rust-lang/crates.io/pull/9786 instead, but I'm open to other opinions.

/cc @rust-lang/crates-io @dtolnay

LawnGnome:
> I think I'm leaning towards reverting #9786 instead, but I'm open to other opinions.

I'd also rather do a revert, I think.

eth3lbert:
> I think I'm leaning towards reverting #9786 instead, but I'm open to other opinions.

Yeah, maybe reverting is a better choice, which would also eliminate the need for potential future maintenance.

PR #10155, opened 2024-12-07
Docker: Remove deprecated `version` attribute

Rustin170506:
You can find these warnings when you run docker-compose:
```sh
❯ docker-compose up -d
WARN[0000] /Volumes/t7/code/crates.io/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
WARN[0000] /Volumes/t7/code/crates.io/docker-compose.override.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
```

See more at: https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: "3.4"
-
 x-backend: &backend
   build:
     context: .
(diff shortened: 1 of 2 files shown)


Rustin170506 (commented):

Turbo87 (approved):

PR #10174, opened 2024-12-10
Deprecate `tracked-toolbox` dependency

eth3lbert:
Since `@cached` is built into Ember, we should prefer it over external dependency.


Turbo87 (approved):
makes sense, thanks!

PR #10177, opened 2024-12-10
UI: Implement `/crates/:name/delete` route

Turbo87:
This partially implements step 2 of https://github.com/rust-lang/crates.io/issues/9352. The missing piece after this PR is a "Delete" button in the crate settings tab, since the new route is currently only reachable by typing it directly in the URL bar. I opted for this hidden route for now to be able to test it out first before making it more public.

Related: 
- https://github.com/rust-lang/crates.io/issues/9352

### Screenshots

Screen Shot 2024-12-09 at 15 58 28
Screen Shot 2024-12-09 at 15 57 41

### DIFF
--- a/app/templates/crate/delete.hbs
+++ b/app/templates/crate/delete.hbs
@@ -1,0 +1,64 @@
+<div local-class="wrapper">
+  <div local-class="content">
+    <h1 local-class="title" data-test-title>Delete the {{@model.name}} crate?</h1>
+
+    <p>Are you sure you want to delete the crate "{{@model.name}}"?</p>
+
+    <div local-class="warning">
+      {{svg-jar "triangle-exclamation"}}
+      <p><strong>Important:</strong> This action will permanently delete the crate and its associated versions. Deleting a crate cannot be reversed!</p>
+    </div>
+
+    <div local-class="impact">
+      <h3>Potential Impact:</h3>
+      <ul>
+        <li>Users will no longer be able to download this crate.</li>
+        <li>Any dependencies or projects relying on this crate will be broken.</li>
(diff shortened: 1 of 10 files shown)


app/templates/crate/delete.hbs
```diff
@@ -0,0 +1,60 @@
...
+
+    <div class="requirements">
+      <h3>Requirements:</h3>
+      <p>A crate can only be deleted if:</p>
+      <ul>
+        <li>the crate has been published for less than 72 hours, or</li>
+        <li>the crate only has a single owner,</li>
+        <li>the crate has been downloaded less than 100 times for each month it has been published,</li>
+        <li>and the crate is not depended upon by any other crate on crates.io.</li>
+      </ul>
```
LawnGnome, line 29: I think we're going to need to make the associativity more obvious:

```suggestion
      <ol>
        <li>the crate has been published for less than 72 hours, or</li>
        <li>
          <ul>
            <li>the crate only has a single owner, and</li>
            <li>the crate has been downloaded less than 100 times for each month it has been published, and</li>
            <li>the crate is not depended upon by any other crate on crates.io.</li>
          </ul>
        </li>
      </ol>
```

Which would also need this bit in `delete.module.css`:

```css
.requirements {
    > ol > li:first-child {
        margin-bottom: 0.5em;
    }

    ol ul {
        list-style: none;
        padding-left: 0;
    }
}
```

Which then looks like:

image

LawnGnome (approved):
LGTM, and I agree that doing this as a hidden route to start with makes sense.

In terms of review, nothing substantive, but a few suggestions below.

Turbo87:
rebased and all feedback incorporated :)

Issue #741, opened 2017-05-30
Add an HTTP API reference, if stable


hcpl:
For now, there is this repo where you can grep up to date API in form of router paths and `params.get` (or `query.get` or maybe something else I haven't seen) which access keyword parameters, but this is not acceptable for everyday use.

Since there can be use cases where you need to retrieve data from [crates.io](https://crates.io), consider adding a reference page somewhere for an easy access.

hcpl:
I'm trying to do it.  WIP of the reference here: https://github.com/hcpl/crates.io-http-api-reference.

Some questions popped up:
- What does "following the crate" exactly mean? Just getting notifications for any changes of the crate or only version updates?
- Why does `/versions` return an empty list? What does it stand for?
- What are "category slugs"? [This page](https://crates.io/category_slugs) doesn't explain much. I get the feeling that they are something like PyPI classifiers/categories, but I'm not sure.

carols10cents:
Related: https://github.com/rust-lang/crates.io/issues/731

> What does "following the crate" exactly mean? Just getting notifications for any changes of the crate or only version updates?

I'm not sure what you mean by "any changes of the crate or only version updates"; what changes are on crates.io that aren't version updates right now?

The best way to see what following a crate means, I think, is to follow some crates and then go to your dashboard at https://crates.io/dashboard. The "Latest updates" section is for the crates you're following.

> Why does /versions return an empty list? What does it stand for?

I actually have no idea; [I left a note in the code](https://github.com/rust-lang/crates.io/blob/master/src/krate.rs#L1187-L1189) to remind me to investigate the history of it someday :) Let me know if you figure it out!

> What are "category slugs"? This page doesn't explain much. I get the feeling that they are something like PyPI classifiers/categories, but I'm not sure.

Yes, they are exactly like PyPI classifiers/categories, and I modeled /category_slugs after https://pypi.python.org/pypi?%3Aaction=list_classifiers specifically. That page contains the tokens that crate authors specify in their Cargo.toml in order to put their crates in [categories](https://crates.io/categories). There's a little bit of documentation [here](http://doc.crates.io/manifest.html#package-metadata) and some history on the implementation of the feature [here](https://github.com/rust-lang/crates.io/pull/473).

Please let me know if you have any other questions! This will be great to have, thank you!!

hcpl:
> Related: #731

Thank you for linking, didn't know that. Should have looked at A-docs issues first!

[...]

PR #10186, opened 2024-12-11
Use `utoipa` crates to generate and serve basic OpenAPI description

Turbo87:
Bildschirmfoto 2024-12-11 um 16 47 05

This PR implements the basic infrastructure to have an **experimental**, best-effort, automatically generated OpenAPI description for the crates.io API. The OpenAPI description will be available at https://crates.io/api/openaip.json, with a viewer available at https://crates.io/api/openaip.

Due to the viewer being on the same domain as the Ember.js frontend, it will automatically share the cookie authentication, which may or may not be a problem. We could alternatively consider serving the viewer from a different subdomain if this turns out to be problematic.

Note that this PR only migrates a single route for now to prove the viability of this approach. The remaining routes can be migrated incrementally.

Related:

- https://github.com/rust-lang/crates.io/issues/741

### DIFF
--- a/src/openapi.rs
+++ b/src/openapi.rs
@@ -1,0 +1,30 @@
+use utoipa::OpenApi;
+use utoipa_axum::router::OpenApiRouter;
+
+#[derive(OpenApi)]
+pub struct BaseOpenApi;
+
+impl BaseOpenApi {
+    pub fn router<S>() -> OpenApiRouter<S>
+    where
+        S: Send + Sync + Clone + 'static,
+    {
+        OpenApiRouter::with_openapi(Self::openapi())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::tests::util::{RequestHelper, TestApp};
+    use http::StatusCode;
+    use insta::assert_json_snapshot;
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_openapi_snapshot() {
+        let (_app, anon) = TestApp::init().empty().await;
+
+        let response = anon.get::<()>("/api/openapi.json").await;
+        assert_eq!(response.status(), StatusCode::OK);
+        assert_json_snapshot!(response.json());
+    }
+}
--- a/src/openapi.rs
+++ b/src/openapi.rs
@@ -2,14 +2,20 @@ use utoipa::OpenApi;
 use utoipa_axum::router::OpenApiRouter;
 
 #[derive(OpenApi)]
-#[openapi(info(
-    title = "crates.io",
-    description = "API documentation for the [crates.io](https://crates.io/) package registry",
-    terms_of_service = "https://crates.io/policies",
-    contact(name = "the crates.io team", email = "help@crates.io"),
-    license(),
-    version = "0.0.0",
-))]
+#[openapi(
+    info(
+        title = "crates.io",
+        description = "API documentation for the [crates.io](https://crates.io/) package registry",
+        terms_of_service = "https://crates.io/policies",
+        contact(name = "the crates.io team", email = "help@crates.io"),
+        license(),
+        version = "0.0.0",
+    ),
+    servers(
+        (url = "https://crates.io"),
+        (url = "https://staging.crates.io"),
+    ),
+)]
 pub struct BaseOpenApi;
 
 impl BaseOpenApi {
--- a/src/openapi.rs
+++ b/src/openapi.rs
@@ -2,6 +2,14 @@ use utoipa::OpenApi;
 use utoipa_axum::router::OpenApiRouter;
 
 #[derive(OpenApi)]
+#[openapi(info(
+    title = "crates.io",
+    description = "API documentation for the [crates.io](https://crates.io/) package registry",
+    terms_of_service = "https://crates.io/policies",
+    contact(name = "the crates.io team", email = "help@crates.io"),
+    license(),
+    version = "0.0.0",
+))]
 pub struct BaseOpenApi;
 
 impl BaseOpenApi {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1125,6 +1125,8 @@ dependencies = [
  "typomania",
  "unicode-xid",
  "url",
+ "utoipa",
+ "utoipa-axum",
  "zip",
 ]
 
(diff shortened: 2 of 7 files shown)


src/openapi.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[derive(OpenApi)]
+#[openapi(
+    info(
+        title = "crates.io",
+        description = "API documentation for the [crates.io](https://crates.io/) package registry",
+        terms_of_service = "https://crates.io/policies",
+        contact(name = "the crates.io team", email = "help@crates.io"),
+        license(),
+        version = "0.0.0",
```
Turbo87, line 12: the OpenAPI spec does not specify the format of the `version` field other than that it's a "string". we could potentially use the git SHA1 or commit timestamp as the `version`, if that seems useful 🤔

src/config/server.rs
```diff
@@ -162,7 +162,7 @@ impl Server {
         // the `script` in `public/github-redirect.html`
         let content_security_policy = format!(
             "default-src 'self'; \
-            connect-src 'self' *.ingest.sentry.io https://docs.rs https://play.rust-lang.org {cdn_domain}; \
+            connect-src 'self' https://crates.io https://staging.crates.io *.ingest.sentry.io https://docs.rs https://play.rust-lang.org {cdn_domain}; \
```
LawnGnome, line 165: Is this necessary?

src/openapi.rs
LawnGnome: I thinking a combination of both would probably be most informative: something like `2024-12-11-21-45-00-4b33fdda250e7a8ebb4c40eff4d6205ce09c7e2a`.

LawnGnome (commented):
As I said earlier on a call, weirdly, I've also been playing around with `utoipa` in the last week.

In general, I like the idea of specifying our API (which is a significant superset of the basic Cargo Registry API) more formally, and providing it in a form that can be used downstream by users to do things like generating API clients.[^ts] OpenAPI seems to be as good a choice as any, so :+1: for that.

That said, I'm wondering if this is _too_ bare bones right now, even for an initial exploratory PR. To my mind, the benefit of this is in showing the actual shape of the data going in and out — so, in the case of this particular controller, the query parameters that are accepted, and then the structure that comes back in the response. Unfortunately, this annotation doesn't really do that.

I [threw a rough, incomplete commit on top of this](https://github.com/LawnGnome/crates.io/commit/dc15241c5958c509b973ce97edd46cbcdcbdd0a6). This implements more or less what I think is the minimum viable OpenAPI description of this endpoint, just without some of the descriptive text we'd eventually need.

Issues that I found along the way include:

1. This doesn't really work with our current usage of `ErasedJson` for small JSON blobs in responses — the only way to really get useful types generated is to write proper `struct` definitions that derive `ToSchema`, and at that point, you might as well just derive `Serialize` as well and use them. I don't know if this is actually a problem — if anything, I probably prefer this code — but it does increase our code size a bit.
2. I'm pretty dissatisfied with how query parameters actually appear in the Swagger UI, but the schema itself looks OK enough.
3. We'd have to develop some patterns around when types get inlined into other types, naming, etc.
4. The `responses(...)` tuples are going to be repeated and mixed and matched in different ways in different controllers[^status-codes], so we'll probably need to develop some macros to handle the common cases.

Also, a bonus issue is that I had never fully grappled with the idiosyncrasies of the search API's parameters and how they fit together, but that's a separate issue.

Realistically, many (all?) of these probably growing pains that we'd have to deal with regardless if we do move towards a more explicitly defined API schema, but it's worth calling them out, I think.

So I guess my fundamental question here isn't so much around this PR (it's fine), but more whether it's complete enough for us to merge as-is, and what we consider to be our baseline for including a controller in the OpenAPI schema if we decide that we're going to go in that direction.

What do you think, @Turbo87?

[^status-codes]: Almost all controllers are going to be able to return 200, 500, and 503. From there, it gets hairier.
[^ts]: I've been playing around with wiring up auto-generation of a TypeScript client from a Rust `utoipa` server, and then using that client from a frontend webapp (in this case, using SvelteKit, but the exact details aren't important). It's _very_ slick when it's all wired up.

src/config/server.rs
Turbo87: otherwise the CSP on https://crates.io would prevent the Swagger UI from sending HTTP requests to https://staging.crates.io if that server is selected.

if we don't want to allow that then the CSP change isn't necessary.

Turbo87:
> That said, I'm wondering if this is _too_ bare bones right now, even for an initial exploratory PR.

yeah, I intentionally kept it small for now to reduce the size of the diff and focus mostly on the plumbing first. what you did in that extra commit is roughly what I had in mind for a followup.



> I guess my fundamental question here isn't so much around this PR (it's fine), but more whether it's complete enough for us to merge as-is, and what we consider to be our baseline for including a controller in the OpenAPI schema if we decide that we're going to go in that direction.

my idea was to migrate all existing routes into very basic `utoipa::path` annotated routes to migrate the `router.rs` in one go and then incrementally improve the individual endpoints in follow-up PRs. IMHO the generated docs don't need to be perfect on inception as long as we clearly indicate that this is experimental/WIP.

eth3lbert:
> The OpenAPI description will be available at https://crates.io/api/openaip.json, with a viewer available at https://crates.io/api/openaip.

I guess we could also just expose the OpenAPI JSON in production and the viewer in the development environment initially? Since I expect it would be used more for development purposes and most viewers provide a way to import from a given JSON URL or file.

Turbo87:
@eth3lbert I mostly included it for local development for now, but I just found https://zudoku.dev/demo?api-url=http://127.0.0.1:8888/api/openapi.json as an alternative, so I've remove the Swagger UI commit, including the CSP header change.

src/config/server.rs
Turbo87: this is obsolete now, see https://github.com/rust-lang/crates.io/pull/10186#issuecomment-2538377449

LawnGnome (approved):
> yeah, I intentionally kept it small for now to reduce the size of the diff and focus mostly on the plumbing first. what you did in that extra commit is roughly what I had in mind for a followup.

OK, then I'm fine merging this and figuring the rest out as we go. I just wanted to get a sense of how you were thinking about it first.

I do think we'll actually want the UI more for production in the longer term, but I'm totally fine dropping it until we have some minimal level of coverage across the API, and not just one or two routes.

PR #10198, opened 2024-12-13
Fix frontend deprecations

eth3lbert:
This PR should hopefully fix the remaining deprecations :)

### DIFF
--- a/app/routes/crate/versions.js
+++ b/app/routes/crate/versions.js
@@ -1,0 +1,12 @@
+import Route from '@ember/routing/route';
+import { waitForPromise } from '@ember/test-waiters';
+
+export default class VersionsRoute extends Route {
+  setupController(controller) {
+    super.setupController(...arguments);
+    let crate = this.modelFor('crate');
+    controller.set('crate', crate);
+    // TODO: Add error handling
+    waitForPromise(crate.loadVersionsTask.perform());
+  }
+}
--- a/app/components/crate-header.js
+++ b/app/components/crate-header.js
@@ -8,7 +8,6 @@ export default class CrateHeader extends Component {
   @service session;
 
   @alias('loadKeywordsTask.last.value') keywords;
-  @alias('loadOwnerUserTask.last.value') ownerUser;
 
   constructor() {
     super(...arguments);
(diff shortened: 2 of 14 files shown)


app/routes/crate/index.js
```diff
@@ -3,4 +3,12 @@ import VersionRoute from './version';
...
   controllerName = 'crate.version';
   templateName = 'crate/version';
+
+  setupController(controller) {
+    super.setupController(...arguments);
+
+    controller.loadDownloadsTask.perform().catch(() => {
+      // ignored
+    });
+  }
```
Turbo87, line 13: why here and not in the parent class?

eth3lbert: Good catch! You're right, `crate.version` route needs this too. I'll make the change.

eth3lbert: Updated!

eth3lbert:
I made a small change https://github.com/rust-lang/crates.io/compare/39651f8a797c6277cc55a26b42229595baac19a5..ae57cf546017b6216cc550c4fb464946b0b263af#diff-8cda1e782838bbd82b901eae0eef8750eaae70e6685898e5f51ef474459a3374
and rebased to the latest commit.

app/routes/crate/versions.js
```diff
@@ -0,0 +1,12 @@
+import Route from '@ember/routing/route';
+import { waitForPromise } from '@ember/test-waiters';
+
+export default class VersionsRoute extends Route {
+  setupController(controller) {
+    super.setupController(...arguments);
+    let crate = this.modelFor('crate');
+    controller.set('crate', crate);
+    // TODO: Add error handling
+    waitForPromise(crate.loadVersionsTask.perform());
```
Turbo87, line 10: I'm wondering if it makes more sense to call this task in the `model()` or at least `afterModel()` hooks, so that the data is fully loaded before the route displays anything. I don't know why it was originally implemented this way.

I guess this is a non-blocking concern though, since the code previously used the same pattern 🤔

Turbo87 (approved):

app/routes/crate/versions.js
eth3lbert: Although implementing it this way could potentially be more complex, it would also be more responsive, as it could display all other elements (layout, crate header) before the versions are ready (a loading indicator for this hasn't been implemented yet).

eth3lbert: But yeah, I mostly implemented it following the previous pattern. But I guess putting this either in `model()` or `afterModel()` should also work for non-blocking purposes, just note not to await for it.

PR #10201, opened 2024-12-13
utoipa: Add annotations to (almost) all endpoints

Turbo87:
This PR continues what #10186 started. It migrates (almost) all of our API endpoints to `utoipa` with a set of very basic annotations. The next step will be to add path and query parameters and request/response body schemas. Also missing is authentication schemes, but this PR should be a good starting point to add them :)

Bildschirmfoto 2024-12-13 um 18 56 33

### DIFF
--- a/src/controllers/category.rs
+++ b/src/controllers/category.rs
@@ -12,7 +12,14 @@ use diesel::QueryDsl;
 use diesel_async::RunQueryDsl;
 use http::request::Parts;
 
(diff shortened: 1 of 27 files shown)


eth3lbert (approved):
I only skimmed through the routes. I think CI would watch out for the rest of the parts, and it seems all good 👍  We now have an overview of the routes, and it also covers deprecated status. I think this is a good enough start :)

Thanks for the work!

PR #10214, opened 2024-12-16
controllers/krate/versions: Add `release_tracks` meta for non-paginated versions

eth3lbert:
Per https://github.com/rust-lang/crates.io/pull/9624#issuecomment-2544487253

### DIFF
--- a/src/controllers/krate/versions.rs
+++ b/src/controllers/krate/versions.rs
@@ -8,7 +8,6 @@ use diesel_async::{AsyncPgConnection, RunQueryDsl};
 use futures_util::{future, TryStreamExt};
 use http::request::Parts;
 use indexmap::{IndexMap, IndexSet};
-use std::cmp::Reverse;
 use std::str::FromStr;
 
 use crate::app::AppState;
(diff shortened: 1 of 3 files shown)


src/controllers/krate/versions.rs
```diff
@@ -75,6 +75,12 @@ pub async fn list_versions(state: AppState, path: CratePath, req: Parts) -> AppR
 
     Ok(match pagination {
         Some(_) => json!({ "versions": versions, "meta": versions_and_publishers.meta }),
+        None if include_release_tracks => {
+            json!({
+                "versions": versions,
+                "meta": ReleaseTracksMeta { release_tracks: versions_and_publishers.meta.release_tracks }
+            })
+        }
```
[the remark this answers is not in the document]
eth3lbert: Or perhaps always including the meta attribute would be a better choice?

src/controllers/krate/versions.rs
```diff
@@ -365,6 +397,12 @@ struct ResponseMeta {
     release_tracks: Option<ReleaseTracks>,
 }
 
+#[derive(Serialize)]
+struct ReleaseTracksMeta {
```
Turbo87, line 401:
```suggestion
struct ListVersionsMeta {
```

in case we ever want to add something else to it too 😉

src/controllers/krate/versions.rs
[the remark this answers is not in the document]
Turbo87: always including `meta` seems fine to me since we already use `skip_serializing_if` below

src/controllers/krate/versions.rs
eth3lbert: Marked as resolved due to adopting the "always include meta" approach.

src/tests/routes/crates/versions/snapshots/crates_io__tests__routes__crates__versions__list__versions.snap
```diff
@@ -1,9 +1,12 @@
 ---
 source: src/tests/routes/crates/versions/list.rs
 expression: response.json()
-snapshot_kind: text
```
Turbo87, line 4: you might want to update your `cargo-insta` version 😉

PR #10216, opened 2024-12-16
controllers/krate/search: Read query parameters via axum extractor

Turbo87:
This allows us to automatically derive OpenAPI descriptions and avoids unnecessary memory allocations from query parameters we do not care about in this endpoint.

Small caveat: the `serde_html_form` crate currently deserializes `foo=` as `None` instead of `Some("")` (see https://github.com/jplatte/serde_html_form/issues/13), which is technically a behavior change for our API. According to our logs, our frontend UI appears to be the only user of the `following` query parameter though, and that one uses `following=1`. So while this is theoretically a breaking change, I don't think in practice it's a hard blocker.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1636,6 +1636,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+ "unicode-xid",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -567,6 +567,7 @@ dependencies = [
  "multer",
  "pin-project-lite",
  "serde",
+ "serde_html_form",
  "serde_json",
  "tower",
  "tower-layer",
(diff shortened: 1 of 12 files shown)


LawnGnome (approved):
I think the behavioural change is reasonable (and thanks for digging through the logs to confirm nobody actually relies on it), so :+1:.

I do wonder if we need to adjust the wording on the operation documentation to make it clear that most of the filtering parameters are mutually exclusive in practice — only one of `all_keywords`, `keyword`, `letter`, `following`, or `ids[]` can ever be applied.

Turbo87:
> I do wonder if we need to adjust the wording on the operation documentation to make it clear that most of the filtering parameters are mutually exclusive

Good point. I missed that part from your original commit. I've adopted your wording now :)

PR #10245, opened 2024-12-19
Add a basic user admin backend

LawnGnome:
This PR adds three new routes, all of them admin only:

* `GET /api/v1/users/:user_id/admin`: returns an extended version of `EncodablePrivateUser` with information on whether the account is locked
* `PUT /api/v1/users/:user_id/lock`: locks the given account
* `DELETE /api/v1/users/:user_id/unlock`: unlocks the given account

The only part I expect to be slightly controversial here is the first one: in theory, we could extend the existing `GET /api/v1/users/:user_id` route to return the extra fields only to admins. But, practically, it feels cleaner and less error prone to add a new route that is entirely its own thing, although it reuses existing view and model infrastructure as much as it can.

Unsurprisingly, this will be followed by some frontend PRs.

## Notes for reviewers

This is probably best reviewed commit-by-commit. (I thought about splitting it into three or four PRs, but I think that's just increasing the review overhead, honestly.)

I will also note — slightly defensively — that the actual _functional_ diff is only 247 lines; the remainder is the update to the OpenAPI snapshot, integration tests for the new controllers, and snapshots related to those tests.

### DIFF
--- a/src/controllers/user/admin.rs
+++ b/src/controllers/user/admin.rs
@@ -1,12 +1,14 @@
 use axum::{extract::Path, Json};
+use chrono::NaiveDateTime;
 use crates_io_database::schema::{emails, users};
 use diesel::{pg::Pg, prelude::*};
-use diesel_async::{AsyncConnection, RunQueryDsl};
+use diesel_async::{scoped_futures::ScopedFutureExt, AsyncConnection, RunQueryDsl};
 use http::request::Parts;
+use utoipa::ToSchema;
 
 use crate::{
     app::AppState, auth::AuthCheck, models::User, sql::lower, util::errors::AppResult,
-    views::EncodableAdminUser,
+    util::rfc3339, views::EncodableAdminUser,
 };
 
 /// Find user by login, returning the admin's view of the user.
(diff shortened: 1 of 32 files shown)


src/views.rs
```diff
@@ -467,6 +467,39 @@ pub struct EncodableMe {
...
 
+/// The admin's view of the serialisable user.
+///
+/// Essentially a superset of [`EncodablePrivateUser`] with the user's current
+/// lock status.
+#[derive(Deserialize, Serialize, Debug)]
+pub struct EncodableAdminUser {
+    #[serde(flatten)]
+    pub user: EncodablePrivateUser,
+    pub lock: Option<EncodableUserLock>,
```
Turbo87, line 478: why the nested object instead of directly adding the two `account_lock_reason` and `account_lock_until` fields? seems a bit easier to handle to me 😅

src/controllers/user/admin.rs
```diff
@@ -0,0 +1,185 @@
...
+    request_body = LockRequest,
+    tags = ["admin", "users"],
+    responses((status = 200, description = "Successful Response")),
+)]
+pub async fn lock(
+    state: AppState,
+    Path(user_name): Path<String>,
+    req: Parts,
+    Json(LockRequest { reason, until }): Json<LockRequest>,
+) -> AppResult<Json<EncodableAdminUser>> {
```
Turbo87, line 73: for new APIs I would prefer to avoid such "custom action" API endpoints and use a `PATCH` API instead (`PATCH /api/v1/users/{user}/admin`?) like we started for versions too and also to some degree already have for users (using `PUT` unfortunately...).

this will allow us to "easily" extend the functionality of this endpoint in the future instead of having to add new endpoints for basically every field we want to change.

src/tests/routes/users/snapshots/crates_io__tests__routes__users__admin__get__admin-found.snap
```diff
@@ -0,0 +1,17 @@
...
+  "email_verification_sent": true,
+  "email_verified": true,
+  "id": 1,
+  "is_admin": false,
+  "lock": null,
+  "login": "foo",
+  "name": null,
+  "publish_notifications": true,
+  "url": "https://github.com/foo"
+}
```
Turbo87, line 17: the rest of the API is using e.g. `{ user: { ... } }`, which allows us to side-load other data from the same request in some cases. we should probably stay consistent with the other API endpoints.

PR #10248, opened 2024-12-19
Simplify `releaseTrackSet` using `release_tracks` meta

eth3lbert:
This is a follow-up PR to #10214 that simplifies `releaseTrackSet` using `release_tracks` meta. As part of this transition, we also ensure that `versions` are loaded from a separate request. In other words, we ask the `crate` API to *not* include the `versions` data by default.

This is also part of eliminating the need for versions to be sorted again within the app. Therefore, the next step would be to leverage sorted data from endpoint without recalculating it within the app.

### DIFF
--- a/app/models/crate.js
+++ b/app/models/crate.js
@@ -22,6 +22,22 @@ export default class Crate extends Model {
   @attr max_version;
   @attr max_stable_version;
   @attr newest_version;
+  /**
+   * @typedef {Object} VersionsMeta
+   * @property {number} total
+   * @property {string | null} next_page
+   * @property {Object.<string, ReleaseTrackDetails>} release_tracks
+   *
+   * @typedef {Object} ReleaseTrackDetails
+   * @property {string} highest
+   **/
+  /**
+   * This isn't an attribute in the crate response.
+   * It's actually the `meta` attribute that belongs to `versions`
+   * and needs to be assigned to `crate` manually.
+   * @type {VersionsMeta | null}
+   **/
+  @attr versions_meta;
 
   @attr description;
   @attr homepage;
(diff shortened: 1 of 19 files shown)


app/models/crate.js
```diff
@@ -130,8 +147,19 @@ export default class Crate extends Model {
...
+   **/
+  loadVersionsTask = task(async ({ reload = false, withReleaseTracks } = {}) => {
+    let versionsRef = this.hasMany('versions');
+    let opts = { adapterOptions: { withReleaseTracks } };
+    if (opts.adapterOptions.withReleaseTracks == null) {
+      opts.adapterOptions.withReleaseTracks = this?.versions_meta?.release_tracks == null;
+    }
+    let fut = reload === true ? versionsRef.reload(opts) : versionsRef.load(opts);
+    return (await fut) ?? [];
   });
```
eth3lbert, line 163: I think we could also consider extending this to two tasks: a parent task to calculate the options, and a child task to perform the actual API request.

The reason for this is that for `reload`, we likely want to request with the same arguments. However, with a single task and no other state to hold the calculated options (withReleaseTracks), the value might change. If a child task is introduced, we could use `last.args` as calculated options.

eth3lbert:
After rethinking this a bit, perhaps the version-related functionality might be worth moving from the crate model to a dedicated service. If we aim to eliminate the need for sorting versions within the application, we likely need to store sorted versions (or indexes should suffice). Otherwise, with only store, we would still need to perform a sorting operation to obtain sorted versions.

Therefore, it would be more beneficial to simply extract these functionalities into a dedicated service that handles loading a crate's versions, release_tracks, and sorted indexes.

I'll make a change for the above reason.

eth3lbert:
Closing as this should be covered in #10659.

PR #10249, opened 2024-12-19
real_ip: Remove Heroku router workaround

Turbo87:
We have upgraded our servers to use the "Heroku Router 2.0" which fixed the `X-Forwarded-For` bug and no longer appends the IP to the first header it sees. This means we can finally remove our workaround and parse the header appropriately.


LawnGnome (approved):
Makes sense to me.

PR #10253, opened 2024-12-20
crate/delete: tinker with the requirements styling

LawnGnome:
As discussed at today's team meeting, here is a slight tweak to how the requirements are laid out when deleting a crate.

Before:

image

After:

image

### DIFF
--- a/app/styles/crate/delete.module.css
+++ b/app/styles/crate/delete.module.css
@@ -55,7 +55,8 @@
 
 .requirements {
     .or {
-        text-align: center;
+        padding-left: 3.5em;
+        padding-bottom: 0.3em;
         font-weight: bold;
         font-variant: small-caps;
     }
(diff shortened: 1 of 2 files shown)


eth3lbert:
I feel it would be better if the "OR" were left-aligned, perhaps placing it between "the" and "(a)". This is more of a personal preference, so I might be wrong 😉

LawnGnome:
@eth3lbert I don't know where I land on that — I think I want a little more visual hierarchy, but just centering might be admittedly too much. How about this:

image

eth3lbert:
> @eth3lbert I don't know where I land on that — I think I want a little more visual hierarchy, but just centering might be admittedly too much. How about this:
> 
> 
> 
> image

This looks great to me! I love this one 😍

Turbo87 (approved):

Turbo87:
rebased to fix the conflict with #10266

carols10cents:
Sorry for how slow I was to take a look at this-- but it looks great! Thank you!!! ❤️

PR #10265, opened 2024-12-23
controllers/krate/delete: Send deletion notification email after successful deletion

Turbo87:
This PR addresses part two of https://github.com/rust-lang/crates.io/issues/9352#issuecomment-2549402681 by sending a notification email to the owner when a crate is getting deleted.

### DIFF
--- a/src/controllers/krate/delete.rs
+++ b/src/controllers/krate/delete.rs
@@ -1,6 +1,7 @@
 use crate::app::AppState;
 use crate::auth::AuthCheck;
 use crate::controllers::krate::CratePath;
+use crate::email::Email;
 use crate::models::{NewDeletedCrate, Rights};
 use crate::schema::{crate_downloads, crates, dependencies};
 use crate::util::errors::{custom, AppResult, BoxedAppError};
@@ -80,6 +81,7 @@ pub async fn delete_crate(path: CratePath, parts: Parts, app: AppState) -> AppRe
         }
     }
 
+    let crate_name = krate.name.clone();
     conn.transaction(|conn| {
         async move {
             diesel::delete(crates::table.find(krate.id))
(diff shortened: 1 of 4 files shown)


src/controllers/krate/delete.rs
```diff
@@ -147,6 +166,33 @@ async fn has_rev_dep(conn: &mut AsyncPgConnection, crate_id: i32) -> QueryResult
...
+impl Email for CrateDeletionEmail<'_> {
+    fn subject(&self) -> String {
+        format!("crates.io: Deleted \"{}\" crate", self.krate)
+    }
+
+    fn body(&self) -> String {
+        format!(
+            "Hi {},
+
+this is a confirmation email for the deletion of your \"{}\" crate.
```
LawnGnome, line 188: Wording tweak (feel free to ignore it if you want, but in that case let's at least capitalise the first letter):

```suggestion
Your \"{}\" crate has been deleted, per your request.
```

src/controllers/krate/delete.rs
```diff
@@ -116,6 +118,23 @@ pub async fn delete_crate(path: CratePath, parts: Parts, app: AppState) -> AppRe
...
+            let email = CrateDeletionEmail {
+                user: &user.gh_login,
+                krate: &crate_name,
+            };
+
+            app.emails.send(&recipient, email).await?
+        }
+
+        Ok::<_, anyhow::Error>(())
+    };
```
LawnGnome, line 132: Curiosity, rather than any actual change required: was there a reason for doing this as an `async` block instead of an `async fn`?

LawnGnome (approved):

src/controllers/krate/delete.rs
Turbo87: it would have been a fn with quite a few parameters and implementing it as a block seemed easier 😅

PR #10266, opened 2024-12-23
Raise crate deletion download limit

Turbo87:
As discussed in the last crates.io team meeting the current limit of just 100 downloads per month was too conservative. This commit raises the limit to 500 downloads per month to make self-service deletions a bit more viable in practice.

### DIFF
--- a/app/templates/crate/delete.hbs
+++ b/app/templates/crate/delete.hbs
@@ -26,7 +26,7 @@
         <li>
           <ul>
             <li>the crate only has a single owner, and</li>
-            <li>the crate has been downloaded less than 100 times for each month it has been published, and</li>
+            <li>the crate has been downloaded less than 500 times for each month it has been published, and</li>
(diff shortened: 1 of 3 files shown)


Rustin170506 (approved):
Thanks! :shipit:

Rustin170506:
The snapshot needs to be updated.

Turbo87:
> The snapshot needs to be updated.

oops, thought I had caught all of them but apparently I missed the new OpenAPI snapshot 😅

Rustin170506 (approved):

LawnGnome (approved):
:+1:, but as discussed last week, let's make sure we put a rough rubric on how we got to 500 in the ops guide.

eth3lbert (commented):
Forgot to mention that we also have an experimenter https://github.com/rust-lang/crates.io/discussions/10273 who is interested in this 😉

PR #10276, opened 2024-12-26
mirage: Respect `include` query params in `GET /api/v1/crates/:id`

eth3lbert:
Extracted from #10248 😉


Turbo87 (approved):
looks good to me at first glance

PR #10277, opened 2024-12-26
e2e/acceptance/versions: Fix potential flaky test case

eth3lbert:
This is another PR that extracted from #10248.


Turbo87:
> Fix potential flaky test case

Can you explain in what case this can be flaky?

eth3lbert:
> > Fix potential flaky test case
> 
> Can you explain in what case this can be flaky?

I recall a timeout or failure in that test case 😅 

Since `expect(...).toEqual(...)` is a non-retrying assertion, I added a check to ensure we obtain the expected results from `evaluateAll` before making the assertion.

PR #10288, opened 2024-12-29
controllers/krate/metadata: Add support for `default_version` include mode

eth3lbert:
This allows us to respond with a version data of `default_version` included, which potentially benefits apps by eliminating the need to wait for the crate response to obtain the `default_version` and then make a subsequent request to get the actual version data. This is particularly useful when we move towards not requesting with `versions` included.

### DIFF
--- a/src/controllers/krate/metadata.rs
+++ b/src/controllers/krate/metadata.rs
@@ -28,9 +28,11 @@ pub struct FindQueryParams {
     /// Additional data to include in the response.
     ///
     /// Valid values: `versions`, `keywords`, `categories`, `badges`,
-    /// `downloads`, or `full`.
+    /// `downloads`, `default_version`, or `full`.
     ///
     /// Defaults to `full` for backwards compatibility.
+    ///
+    /// **Note**: `versions` and `default_version` share the same key `versions`, therefore `default_version` will be ignored if both are provided.
     ///
     /// This parameter expects a comma-separated list of values.
     include: Option<String>,
@@ -88,7 +90,7 @@ pub async fn find_crate(
             .optional()?
             .ok_or_else(|| crate_not_found(&path.name))?;
 
-    let versions_publishers_and_audit_actions = if include.versions {
+    let mut versions_publishers_and_audit_actions = if include.versions {
         let mut versions_and_publishers: Vec<(Version, Option<User>)> =
             Version::belonging_to(&krate)
                 .left_outer_join(users::table)
@@ -117,6 +119,18 @@ pub async fn find_crate(
     let ids = versions_publishers_and_audit_actions
         .as_ref()
         .map(|vps| vps.iter().map(|v| v.0.id).collect());
+
+    // Since `versions` and `default_version` share the same key (versions), we should only settle
+    // the `default_version` when `versions` is not included.
+    if let Some(default_version) = default_version
+        .as_ref()
+        .filter(|_| include.default_version && !include.versions)
+    {
+        let version = krate.find_version(&mut conn, default_version).await?;
+        let published_by = version.published_by(&mut conn).await?;
+        let actions = VersionOwnerAction::by_version(&mut conn, &version).await?;
+        versions_publishers_and_audit_actions = Some(vec![(version, published_by, actions)]);
+    };
 
     let kws = if include.keywords {
         Some(
@@ -202,6 +216,7 @@ struct ShowIncludeMode {
     categories: bool,
     badges: bool,
     downloads: bool,
+    default_version: bool,
 }
 
 impl Default for ShowIncludeMode {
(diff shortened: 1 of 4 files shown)


src/controllers/krate/metadata.rs
```diff
@@ -118,6 +118,21 @@ pub async fn find_crate(
...
+        if let Some(versions) = versions_publishers_and_audit_actions.as_ref() {
+            default_version_meta = versions
+                .iter()
+                .find(|(version, _, _)| version.num.as_str() == default_version)
+                .cloned();
+        } else if let Ok(version) = krate.find_version(&mut conn, default_version).await {
+            let published_by = version.published_by(&mut conn).await?;
+            let actions = VersionOwnerAction::by_version(&mut conn, &version).await?;
+            default_version_meta = Some((version, published_by, actions));
+        }
```
Turbo87, line 133:
```suggestion
        } else {
            let version = krate.find_version(&mut conn, default_version).await?;
            let published_by = version.published_by(&mut conn).await?;
            let actions = VersionOwnerAction::by_version(&mut conn, &version).await?;
            default_version_meta = Some((version, published_by, actions));
        }
```

src/tests/routes/crates/snapshots/crates_io__tests__routes__crates__read__include_default_version.snap
```diff
@@ -0,0 +1,78 @@
...
+    "max_version": "0.0.0",
+    "name": "foo_default_version",
+    "newest_version": "0.0.0",
+    "recent_downloads": null,
+    "repository": null,
+    "updated_at": "[datetime]",
+    "versions": null,
+    "yanked": false
+  },
+  "default_version": {
```
Turbo87, line 38: I'm wondering whether we should add this to the `versions` array instead. if both includes are set we wouldn't be duplicating the information in the response, and I think e.g. Ember Data would expect it to be in `versions` instead of a somewhat random other key 🤔

src/controllers/krate/metadata.rs
```diff
@@ -243,13 +264,15 @@ impl FromStr for ShowIncludeMode {
                         categories: true,
                         badges: true,
                         downloads: true,
+                        default_version: true,
```
Turbo87, line 267: I guess it's a bit unfortunate that we named the legacy default value `full`. I think I'd prefer if we did not add more fields to the default case where `?include=...` is not set 😅

src/tests/routes/crates/snapshots/crates_io__tests__routes__crates__read__new_name.snap
```diff
@@ -35,6 +35,7 @@ snapshot_kind: text
     "versions": null,
     "yanked": false
   },
+  "default_version": null,
```
Turbo87, line 38: I guess this should probably use "skip if empty" 🤔

src/tests/routes/crates/snapshots/crates_io__tests__routes__crates__read__include_default_version.snap
eth3lbert: Yeah, this is quite interesting. I haven't played around with this part on the frontend yet. I'm guessing we have  versions relationship with `@hasMany` and `default_version` relationship with `@belongsTo`. But indeed, the `default_version` is definitely one of the `versions`.

So, are you suggesting we just share the same key here? Which means that with only `default_version` in the `include` query, the output would be similar to:

```json
{
"crate": {...},
"categories": null,
"keyword": null,
"versions": [{...default_version...}] 
}
```

And if both `versions` and `default_version` are included (or `full`), then we simply treat it as the `versions` and ignore the `default_version`?

Turbo87: yeah, exactly

src/controllers/krate/metadata.rs
eth3lbert: If we share the same key, I think we could simply make `default_version` default to `true`, right? Since it's effectively shadowed by `versions` by default?

Turbo87: yep, and the next comment would also become obsolete since there is no new key that would be added

Turbo87 (approved):

src/tests/routes/crates/snapshots/crates_io__tests__routes__crates__read__include_default_version.snap
```diff
@@ -0,0 +1,79 @@
...
+      "versions": "/api/v1/crates/foo_default_version/versions"
+    },
+    "max_stable_version": null,
+    "max_version": "0.0.0",
+    "name": "foo_default_version",
+    "newest_version": "0.0.0",
+    "recent_downloads": null,
+    "repository": null,
+    "updated_at": "[datetime]",
+    "versions": null,
```
eth3lbert, line 35: @Turbo87 Ah, should we also add the `default_versions` id here?

Turbo87: I think not. I think I would interpret the `versions` array here as the **full** list of versions, not a subset, which would be the case if we only added the `default_version` in it.

PR #10293, opened 2024-12-30
Add `README.md` files to subpackages

Turbo87:
Our subpackages in the `crates` folder lacked proper documentation, so this PR adds a couple of basic `README.md` files and instructs `rustdoc` to use them as the module docstrings too.


LawnGnome (approved):
Nice work!

PR #10298, opened 2024-12-31
Tidy crate metadata

eth3lbert:
This changes the sorting of `crate.versions`. Previously, we sorted versions by semver, which is more computationally expensive compared to sorting by id/date, which can be achieved more easily directly from the database. While this change alters the sorting order, it still provides consistent output and eliminates the need for double semver sorting.


Turbo87 (approved):

src/controllers/krate/metadata.rs
```diff
@@ -91,15 +90,12 @@ pub async fn find_crate(
...
-                .select(<(Version, Option<User>)>::as_select())
-                .load(&mut conn)
-                .await?;
-
-        versions_and_publishers
-            .sort_by_cached_key(|(version, _)| Reverse(semver::Version::parse(&version.num).ok()));
+        let versions_and_publishers: Vec<(Version, Option<User>)> = Version::belonging_to(&krate)
+            .left_outer_join(users::table)
+            .select(<(Version, Option<User>)>::as_select())
+            .order_by(versions::id)
```
eth3lbert, line 96: @Turbo87 Hmm, should we sort this in descending order? This more closely aligns with the previous ordering.

Turbo87: ah, yeah, that probably makes sense

PR #10306, opened 2025-01-02
controllers/version: Pipelining database queries

eth3lbert:
I just rewrote this using pipelining, but I haven't done any benchmarking (though I expect the performance improvement to be marginal).


Turbo87 (approved):
hahaha, I had the exact same idea when I saw this bit of code the other day 😂

Issue #7881, opened 2024-01-05
Use non-200 HTTP response status codes for API errors


Turbo87:
When cargo and crates.io were first implemented, the [HTTP response status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) were ignored. cargo assumed that crates.io would always respond with a "200 OK" and any error messages are detected from the response JSON body. When crates.io responded with a different status code the cargo was showing the raw JSON body instead of a regular error message.

This was changed in https://github.com/rust-lang/cargo/pull/6771, which was released in cargo 1.34 (mid-2019). Since that release, cargo now supports 4xx and 5xx status codes too and extracts the error message from the JSON response, if available.

The crates.io team has decided to phase out these "200 OK" response status codes for error responses to make our API a little more intuitive. Since this can be seen as a breaking change to our API we will need to announce this change publicly, including a cut-off date. On this date we want to switch the API from the previous "200 OK" behavior to the new 4xx/5xx behavior.

Inside the crates.io source code we have been using a `cargo_err()` fn which returns a "200 OK" JSON response with a custom error message. These calls will have to be replaced with other errors.

To allow us to incrementally migrate our codebase while maintaining the previous behavior, a middleware was introduced in #7715, which ensures that all cargo-relevant endpoints always return "200 OK". #7776 then implemented a feature flag on top of the middleware, which will allow us to flip the switch on the cut-off date and test it out ahead of time on our staging environment.

The remaining work now is to replace all calls of `cargo_err()` with 4xx/5xx error responses and adjust the corresponding tests, or write new tests if none exist for these error scenarios yet.

Eh2406:
cc: @rust-lang/cargo, for a heads up about the breaking change.

weihanglo:
Cargo might not need to change I guess? Just old cargos (<1.35) would get obscure error messages, which the versions are so old and should be fine.

ehuss:
How much of a breaking change is this? It seems like it is mostly an issue of the quality of error messages, correct?

For example, the old behaviors are:
* 403 — `received 403 unauthorized response code`, which doesn't explain why, but should be sufficient for a user to figure out what was wrong (their token was bad, or for the wrong account/crate, etc.).
* 404 — `received 404 not found response code`, which doesn't explain why, but should be sufficient for a user to figure out what was wrong (`cargo owner` for a package that doesn't exist).
* all other non-200 codes — shows the raw headers and body, which although ugly, contains the text of the error.

[...]

PR #10332, opened 2025-01-07
mirage: Maintains crate version-related handling

eth3lbert:
This PR maintains crate version-related handling by:
- Fixing the not found error status code to 404.
- Adding the missing `GET /api/v1/crates/:name/:version` endpoint.
- Rewording some request parameters or variables for consistency.

Related:
- #7881


Turbo87 (approved):
thanks!

might have been best to submit them as dedicated PRs though, since the only common thing between them is "mirage" 😉

eth3lbert:
Haha, indeed, and my bad! I actually just wanted to create a single add route, but somehow it ended up like this :dizzy_face:

PR #10335, opened 2025-01-07
e2e/acceptance/readme-rendering: Fix test flakyness

eth3lbert:
According to
https://github.com/mermaid-js/mermaid/blob/bc2cc61240987ac956c69e262c93971cd254f00a/packages/mermaid/src/mermaid.ts#L160-L164, when mermaid finishes processing a node, it will set the `data-processed` to true.


Turbo87 (approved):
let's hope this helps :D

PR #10344, opened 2025-01-08
Include `default_version` by default

eth3lbert:
This PR would ensure that `default_version` is included in the `crate` response by default. This eliminates the need to wait for the `versions` request to complete when `default_version` is the only version required. This also decouples requested version handling from versions. 

Related:
- #10288 
- #10296

### DIFF
--- a/app/adapters/crate.js
+++ b/app/adapters/crate.js
@@ -49,7 +49,7 @@ export default class CrateAdapter extends ApplicationAdapter {
 function setDefaultInclude(query) {
   if (query.include === undefined) {
     // This ensures `crate.versions` are always fetched from another request.
-    query.include = 'keywords,categories,downloads';
+    query.include = 'keywords,categories,downloads,default_version';
   }
 
   return query;
--- a/app/adapters/version.js
+++ b/app/adapters/version.js
@@ -6,4 +6,14 @@ export default class VersionAdapter extends ApplicationAdapter {
     let num = snapshot.record.num;
     return `/${this.namespace}/crates/${crateName}/${num}`;
   }
+
+  urlForQueryRecord(query) {
+    let { name, num } = query ?? {};
+    let baseUrl = this.buildURL('crate', name);
+    let url = `${baseUrl}/${num}`;
+    // The following used to remove them from URL's query string.
+    delete query.name;
+    delete query.num;
+    return url;
+  }
 }
--- a/app/models/crate.js
+++ b/app/models/crate.js
@@ -67,6 +67,14 @@ export default class Crate extends Model {
     return Object.fromEntries(versions.slice().map(v => [v.id, v]));
   }
 
+  /** @return {Map<string, import("../models/version").default>} */
+  @cached
+  get loadedVersionsByNum() {
+    let versionsRef = this.hasMany('versions');
+    let values = versionsRef.value();
+    return new Map(values?.map(ref => [ref.num, ref]));
+  }
+
   @cached get releaseTrackSet() {
     let map = new Map();
     let { versionsObj: versions, versionIdsBySemver } = this;
--- a/app/routes/crate/version-dependencies.js
+++ b/app/routes/crate/version-dependencies.js
@@ -1,25 +1,31 @@
+import { NotFoundError } from '@ember-data/adapter/error';
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 
 export default class VersionRoute extends Route {
+  @service store;
   @service router;
 
   async model(params, transition) {
     let crate = this.modelFor('crate');
 
-    let versions;
-    try {
-      versions = await crate.loadVersionsTask.perform();
-    } catch (error) {
-      let title = `${crate.name}: Failed to load version data`;
-      return this.router.replaceWith('catch-all', { transition, error, title, tryAgain: true });
-    }
-
     let requestedVersion = params.version_num;
-    let version = versions.find(version => version.num === requestedVersion);
-    if (!version) {
-      let title = `${crate.name}: Version ${requestedVersion} not found`;
-      return this.router.replaceWith('catch-all', { transition, title });
+    let version;
+    try {
+      version =
+        crate.loadedVersionsByNum.get(requestedVersion) ??
+        (await this.store.queryRecord('version', {
+          name: crate.id,
+          num: requestedVersion,
+        }));
+    } catch (error) {
+      if (error instanceof NotFoundError) {
+        let title = `${crate.name}: Version ${requestedVersion} not found`;
+        return this.router.replaceWith('catch-all', { transition, title });
+      } else {
+        let title = `${crate.name}: Failed to load version data`;
+        return this.router.replaceWith('catch-all', { transition, error, title, tryAgain: true });
+      }
     }
 
     try {
--- a/app/routes/crate/version.js
+++ b/app/routes/crate/version.js
@@ -1,22 +1,26 @@
+import { NotFoundError } from '@ember-data/adapter/error';
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';
 import { waitForPromise } from '@ember/test-waiters';
 
 import { didCancel } from 'ember-concurrency';
-import semverSort from 'semver/functions/rsort';
 
 import { AjaxError } from '../../utils/ajax';
 
 export default class VersionRoute extends Route {
   @service router;
   @service sentry;
+  @service store;
 
   async model(params, transition) {
     let crate = this.modelFor('crate');
 
-    let versions;
+    // TODO: Resolved version without waiting for versions to be resolved
+    // The main blocker for this right now is that we have a "belongsTo" relationship between
+    // `version-download` and `version` in sync mode. This requires us to wait for `versions` to
+    // exist before `version-download` can be created.
     try {
-      versions = await crate.loadVersionsTask.perform();
+      await crate.loadVersionsTask.perform();
     } catch (error) {
       let title = `${crate.name}: Failed to load version data`;
       return this.router.replaceWith('catch-all', { transition, error, title, tryAgain: true });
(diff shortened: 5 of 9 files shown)


app/models/crate.js
```diff
@@ -67,6 +67,14 @@ export default class Crate extends Model {
     return Object.fromEntries(versions.slice().map(v => [v.id, v]));
   }
 
+  /** @return {Map<string, import("../models/version").default>} */
+  @cached
+  get loadedVersionsByNum() {
+    let versionsRef = this.hasMany('versions');
+    let values = versionsRef.value();
+    return new Map(values?.map(ref => [ref.num, ref]));
+  }
```
eth3lbert, line 76: I think `versionsObj` could also be changed to implement in this way. This would allow all `versions` that belong to this `crate` to be available, rather than just those loaded via the `loadVersionsTask`.

eth3lbert:
I have removed a commit that aimed to only await `versions` when needed. This commit fails in CI because we have some `versions.length` in the template, which would potentially load `versions`. I will address this in a separate PR, which will also require that the total number of `versions` be extracted from the meta.

app/adapters/crate.js
```diff
@@ -9,7 +9,7 @@ export default class CrateAdapter extends ApplicationAdapter {
     let { include } = snapshot;
     // This ensures `crate.versions` are always fetched from another request.
     if (include === undefined) {
-      snapshot.include = 'keywords,categories,downloads';
+      snapshot.include = 'keywords,categories,default_version';
```
Turbo87, line 12: is `downloads` no longer needed? 🤔

app/adapters/version.js
```diff
@@ -6,4 +6,12 @@ export default class VersionAdapter extends ApplicationAdapter {
...
     return `/${this.namespace}/crates/${crateName}/${num}`;
   }
+
+  queryRecord(_store, _type, query) {
+    let { name, num } = query ?? {};
+    let baseUrl = this.buildURL('crate', name);
+    let url = `${baseUrl}/${num}`;
+    let data = { ...query, name: undefined, num: undefined };
+    return this.ajax(url, 'GET', { data });
+  }
```
Turbo87, line 16: wouldn't is be easier to override `urlForQueryRecord()` instead?

app/adapters/crate.js
eth3lbert: Oops, no, I did this by accident.

app/adapters/version.js
eth3lbert: Yeah, I've started with something like the following:
```javascript
  urlForQueryRecord(query) {
    let { name, num } = query ?? {};
    let baseUrl = this.buildURL('crate', name);
    let url = `${baseUrl}/${num}`;
    // The following used to remove them from URL's query string.
    delete query.name;
    delete query.num;
    return url
  }

```

I'm fine with both approaches. Do you have any opinions or preferences between them?

eth3lbert:
Rebased to make CI happy after the new Danger Zone was added :D

app/adapters/version.js
eth3lbert: Yeah, fair enough. I'll resolve the conflict and make the change then!

eth3lbert:
Rebased!

app/routes/crate/version-dependencies.js
```diff
@@ -2,21 +2,23 @@ import Route from '@ember/routing/route';
...
+    let version =
+      crate.loadedVersionsByNum.get(requestedVersion) ??
+      (await this.store
+        .queryRecord('version', {
+          name: crate.id,
+          num: requestedVersion,
+        })
+        .catch(() => {
+          // ignored
+        }));
```
Turbo87, line 21: AFAIU this change removes proper error handling. if the version could not be loaded because of a random network issue if would now show "Version not found" instead of the "Failed to load version data" with a retry button that was removed above

app/routes/crate/version-dependencies.js
```diff
@@ -32,6 +34,17 @@ export default class VersionRoute extends Route {
...
+    let crate = this.modelFor('crate');
+    // TODO: Resolved version without waiting for versions to be resolved
+    try {
+      await crate.loadVersionsTask.perform();
+    } catch (error) {
+      let title = `${crate.name}: Failed to load version data`;
+      return this.router.replaceWith('catch-all', { transition, error, title, tryAgain: true });
+    }
+  }
+
```
Turbo87, line 47: I don't understand this part. what do we gain from moving this request to `afterModel()`? from my understanding `afterModel()` is also render-blocking? 🤔

app/routes/crate/version.js
```diff
@@ -25,14 +27,33 @@ export default class VersionRoute extends Route {
...
+      version =
+        crate.loadedVersionsByNum.get(requestedVersion) ??
+        (await this.store
+          .queryRecord('version', {
+            name: crate.id,
+            num: requestedVersion,
+          })
+          .catch(() => {
+            // ignored
+          }));
```
Turbo87, line 39: same here, this should have proper error handling.

Turbo87, line 39: if we still use `crate.loadVersionsTask` above, what is the reason for this fallback? 🤔

e2e/acceptance/crate-dependencies.spec.ts
```diff
@@ -71,7 +71,7 @@ test.describe('Acceptance | crate dependencies page', { tag: '@acceptance' }, ()
   test('shows an error page if versions fail to load', async ({ page, mirage, ember }) => {
     await mirage.addHook(server => {
       let crate = server.create('crate', { name: 'foo' });
-      server.create('version', { crate, num: '2.0.0' });
+      server.create('version', { crate, num: '1.0.0' });
```
Turbo87, line 74: why was the version number changed?

app/adapters/version.js
Turbo87: > I'll resolve the conflict and make the change then

was that change lost in the rebase somehow? 😅

Turbo87:
also, sorry for keeping you waiting on this. I totally forgot about this PR 😞

app/routes/crate/version-dependencies.js
eth3lbert: I can't quite remember the intention 😅  I think it might have been related to a commit I removed, to make the diffs between commits easier to read. I've reverted this part!

app/routes/crate/version.js
eth3lbert: Yes, that's a completely fair question. We still have dependencies on the versions (especially the version count), which makes the fallback seem unnecessary. The fallback would become valuable if we could load the page without blocking to wait for versions.

e2e/acceptance/crate-dependencies.spec.ts
eth3lbert: no idea 🤦

app/adapters/version.js
eth3lbert: oops, updated!

app/routes/crate/version.js
```diff
@@ -25,14 +28,37 @@ export default class VersionRoute extends Route {
...
+            num: default_version,
+          })
+          .catch(error => {
+            if (!(error instanceof NotFoundError)) {
+              let title = `${crate.name}: Failed to load version data`;
+              this.router.replaceWith('catch-all', { transition, error, title, tryAgain: true });
+            }
+          }));
 
       if (!version) {
```
eth3lbert, line 63: Since we already have `default_version`, do we still want this fallback here?

eth3lbert: More specifically, I'm talking about this part https://github.com/rust-lang/crates.io/blob/859959f065fa873da96c494fcbed44fc70fcb727/app/routes/crate/version.js#L63-L68

app/routes/crate/version-dependencies.js
```diff
@@ -1,25 +1,39 @@
...
 
-    let versions;
+    // TODO: Resolved version without waiting for versions to be resolved
     try {
-      versions = await crate.loadVersionsTask.perform();
+      await crate.loadVersionsTask.perform();
     } catch (error) {
       let title = `${crate.name}: Failed to load version data`;
       return this.router.replaceWith('catch-all', { transition, error, title, tryAgain: true });
     }
```
Turbo87, line 18: do we still need this? 🤔

app/routes/crate/version.js
Turbo87: no, I think we can get rid of that and replace it with an error screen

app/routes/crate/version.js
```diff
@@ -25,14 +28,37 @@ export default class VersionRoute extends Route {
     let version;
     let requestedVersion = params.version_num;
```
Turbo87, line 29:
```suggestion
    let requestedVersion = params.version_num || crate.default_version;
    if (error instanceof NotFoundError) {
      let title = `${crate.name}: Failed to find default version`;
      return this.router.replaceWith('catch-all', { transition, title, tryAgain: true });
    } 
```

I think that might be able to simplify the rest of the code here

app/routes/crate/version-dependencies.js
eth3lbert: Ah, yeah, this could be removed! It should be able to auto-load if needed (primarily for version counts).

app/routes/crate/version.js
eth3lbert: Hmm, this doesn't seem to align with the current business logic. Currently, we don't fall back to the `default_version` if the user explicitly requests a specific version.

eth3lbert: I think I get what you're saying.

app/routes/crate/version-dependencies.js
eth3lbert: Ah, I remember why I kept this, as we have an acceptance case for versions. See: https://github.com/rust-lang/crates.io/actions/runs/13137385348/job/36655968746?pr=10344#step:6:90. I think we can remove this test case now.

app/routes/crate/version.js
```diff
@@ -1,44 +1,49 @@
...
 
-    let versions;
+    // TODO: Resolved version without waiting for versions to be resolved
     try {
-      versions = await crate.loadVersionsTask.perform();
+      await crate.loadVersionsTask.perform();
     } catch (error) {
       let title = `${crate.name}: Failed to load version data`;
       return this.router.replaceWith('catch-all', { transition, error, title, tryAgain: true });
     }
```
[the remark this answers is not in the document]
eth3lbert: Yes, `versions` is currently still required due to the `belongsTo` relationship between `version-download` and `version`. I'll leave a comment.

e2e/acceptance/crate-dependencies.spec.ts
```diff
@@ -62,20 +62,6 @@ test.describe('Acceptance | crate dependencies page', { tag: '@acceptance' }, ()
     await expect(page.locator('[data-test-try-again]')).toHaveCount(0);
   });
 
-  test('shows an error page if versions fail to load', async ({ page, msw, ember }) => {
-    let crate = msw.db.crate.create({ name: 'foo' });
-    msw.db.version.create({ crate, num: '2.0.0' });
-    await msw.worker.use(http.get('/api/v1/crates/:crate_name/versions', () => HttpResponse.json({}, { status: 500 })));
```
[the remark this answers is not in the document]
eth3lbert: However, the assertion would need to be changed from `foo: Failed to load version data` to `foo: Version 1.0.0 not found`, which would make the test case look a bit like https://github.com/rust-lang/crates.io/blob/5005c99e176512f47b0a8f9ecc711fdf50c38846/e2e/acceptance/crate-dependencies.spec.ts#L53-L63.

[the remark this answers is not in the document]
Turbo87: ah, nevermind, you're right. I didn't know we already had another test case that covers it. :)

Turbo87 (approved):
LGTM now :)

I guess this is good to go?

eth3lbert:
> LGTM now :)
> 
> 
> 
> I guess this is good to go?

Yes, I will file a follow-up PR tomorrow :) 

thanks for the review!

Issue #10368, opened 2025-01-10
The crate view page is empty when the only difference to a published crate is `-` vs `_`


cyrgani:
### Current Behavior

Documentation pages for nonexistent crates like [serde-json](https://crates.io/crates/serde-json) or [proc_macro2](https://crates.io/crates/proc_macro2), where the only difference in the name is that `_` was used but the crate name uses `-` (and the other way around) do not contain a 404 info (`serde-json: Crate not found`)
 and are empty, compared to a [non-existing crate](https://crates.io/crates/ajdnfiuewfsakdjf) that works fine.

### Expected Behavior

A 404 message like for other non-existing crates or ideally also a note that a similar crate with different usage of `-` or `_` exists.

### Steps To Reproduce

1. Go to [https://crates.io/crates/serde-json](https://crates.io/crates/serde-json)
2. See the (absence of the) error.

### Environment

- Browser: Chrome 131.0.6778.264, Firefox 134
- OS: Linux Mint 22

Turbo87:
@eth3lbert could this be caused by one of your latest refactorings? the API call seems to return the right thing, but the frontend then crashes afterwards.

eth3lbert:
Thank you for the report. I will look into this as soon as I get back to my computer.

cyrgani:
Another case I just encountered which causes the browser to hang and slowly causes OOM, which is probably related: 
[https://crates.io/crates/serde-codegen](https://crates.io/crates/serde-codegen)

eth3lbert:
> [@eth3lbert](https://github.com/eth3lbert) could this be caused by one of your latest refactorings? the API call seems to return the right thing, but the frontend then crashes afterwards.

I just bisected a bit and it seems it failed at HEAD is now at bbe5345fb Update dependency ember-data to v5 (#10355) 😅 

After upgrading, the name requested (`serde-json`) and the name received (`serde_json`) in the response don't match, it results in another dangling crate with no versions.

Before:

<img width="930" alt="Image" src="https://github.com/user-attachments/assets/bf220829-7ded-4867-8aa5-17b720086c60" />

After:

<img width="914" alt="Image" src="https://github.com/user-attachments/assets/2ee94d1f-1ad8-4259-8f1f-5d6fa77f35aa" />

Turbo87:
argh okay... I wonder if using `queryRecord()` instead of `findRecord()` would fix this 🤔 

in any case, we should probably add a test case for this 😅

[...]

PR #10369, opened 2025-01-10
Fix non-canonical crate name pages

Turbo87:
Before the Ember Data 5 update it was possible to visit `/crates/foo-bar` when the crate was actually called `foo_bar` and vice-versa. The Ember Data 5 update broke this behavior, and this PR is restoring it, by using `queryRecord()` instead of `findRecord()`, which can no longer deal with requesting a certain `id` and getting a different one returned from the server.

Fixes https://github.com/rust-lang/crates.io/issues/10368

### DIFF
--- a/app/routes/crate.js
+++ b/app/routes/crate.js
@@ -12,7 +12,7 @@ export default class CrateRoute extends Route {
     let crateName = params.crate_id;
 
     try {
-      return await this.store.findRecord('crate', crateName);
+      return this.store.peekRecord('crate', crateName) || (await this.store.queryRecord('crate', { name: crateName }));
     } catch (error) {
       if (error instanceof NotFoundError) {
         let title = `${crateName}: Crate not found`;
--- a/mirage/route-handlers/crates.js
+++ b/mirage/route-handlers/crates.js
@@ -3,6 +3,10 @@ import { Response } from 'miragejs';
 import { getSession } from '../utils/session';
 import { compareIsoDates, compareStrings, notFound, pageParams, releaseTracks } from './-utils';
 
+function toCanonicalName(name) {
+  return name.toLowerCase().replace(/-/g, '_');
+}
+
 export function list(schema, request) {
   const { start, end } = pageParams(request);
 
@@ -56,7 +60,8 @@ export function register(server) {
 
   server.get('/api/v1/crates/:name', function (schema, request) {
     let { name } = request.params;
-    let crate = schema.crates.findBy({ name });
+    let canonicalName = toCanonicalName(name);
+    let crate = schema.crates.all().models.find(it => toCanonicalName(it.name) === canonicalName);
     if (!crate) return notFound();
     let serialized = this.serialize(crate);
     let includes = request.queryParams?.include ?? '';
--- a/tests/acceptance/crate-test.js
+++ b/tests/acceptance/crate-test.js
@@ -131,6 +131,19 @@ module('Acceptance | crate page', function (hooks) {
     assert.dom('[data-test-try-again]').exists();
   });
 
+  test('works for non-canonical names', async function (assert) {
+    let crate = this.server.create('crate', { name: 'foo-bar' });
+    this.server.create('version', { crate });
+
+    await visit('/crates/foo_bar');
+
+    assert.strictEqual(currentURL(), '/crates/foo_bar');
+    assert.strictEqual(currentRouteName(), 'crate.index');
+    assert.strictEqual(getPageTitle(), 'foo-bar - crates.io: Rust Package Registry');
+
+    assert.dom('[data-test-heading] [data-test-crate-name]').hasText('foo-bar');
+  });
+
   test('navigating to the all versions page', async function (assert) {
     this.server.loadFixtures();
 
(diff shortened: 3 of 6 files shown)


app/routes/crate.js
```diff
@@ -12,7 +12,7 @@ export default class CrateRoute extends Route {
     let crateName = params.crate_id;
 
     try {
-      return await this.store.findRecord('crate', crateName);
+      return await this.store.queryRecord('crate', { name: crateName });
```
Turbo87, line 15: as mentioned in the issue, ideally we would redirect the user to the same route, but with the crate name replaced. this appears to be non-trivial though, since the user could also have requested a child route and we would potentially redirect them to the wrong route. I think this fix matches the previous behavior though, so it's probably good enough for now :)

eth3lbert: Fair enough! I have no idea why this would be somewhat non-trivial in Ember, especially for nested routes 😅 
Since `queryRecord` would make a request for every call, should we optimize this to something like `peekRecord` + `queryRecord`?

Turbo87: > should we optimize this to something like `peekRecord` + `queryRecord`?

hmm, I guess that makes sense, yeah

eth3lbert (approved):
Aside from the missing e2e tests, this looks good to me. The optimization part could also be implemented in a separate PR later.

tests/acceptance/crate-test.js
```diff
@@ -131,6 +131,19 @@ module('Acceptance | crate page', function (hooks) {
...
+    this.server.create('version', { crate });
+
+    await visit('/crates/foo_bar');
+
+    assert.strictEqual(currentURL(), '/crates/foo_bar');
+    assert.strictEqual(currentRouteName(), 'crate.index');
+    assert.strictEqual(getPageTitle(), 'foo-bar - crates.io: Rust Package Registry');
+
+    assert.dom('[data-test-heading] [data-test-crate-name]').hasText('foo-bar');
+  });
```
eth3lbert, line 145: I think we should also check this in e2e.

eth3lbert:
I got an alternative workaround for this issue this morning. Since we know how to normalize `crate.id`, the response of the `id` is predictable. This means we could just make a unique `lid` for it. According to RFC[^1], we could leverage the `setIdentifierGenerationMethod` to tell the store how to generate a `lid`, which should be something like the following:

```javascript
// crates.io/app/initializers/ember-data-identifier.js
import { setIdentifierGenerationMethod } from '@ember-data/store';

function toCanoncialName(name) {
  return name.toLowerCase().replace(/_/g, '-');
}

export function initialize(_app) {
  const generationMethod = (resource, _bucket) => {
    let { id, lid, type } = resource;
    if (lid) {
      return lid;
    }

    if (id) {
      if (resource.type === 'crate') {
        id = toCanoncialName(id);
      }

      return `@lid:${type}-${id}`;
    }

    return crypto.randomUUID();
  };

  setIdentifierGenerationMethod(generationMethod);
}

export default {
  name: 'app.configure-ember-data-identifier',
  initialize,
};
```

This way, IDs with dashes or underscores would be canonicalized to a single unique `lid`.

This approach has the benefit of not requiring modifications to other business logic. However, the downside is that if the response `id` ever becomes unpredictable, the issue would occur again.



[^1]: https://rfcs.emberjs.com/id/0403-ember-data-identifiers/

Turbo87:
> we could leverage the `setIdentifierGenerationMethod` to tell the store how to generate a `lid`

I like the approach, but unfortunately this seems to break our API token creation flow, at least with the proposed implementation. it seems to be missing the `NEW_IDENTIFIERS` part of https://github.com/emberjs/data/blob/v5.3.9/packages/store/src/-private/caches/identifier-cache.ts, but I'm not sure if reimplementing that is worth the complexity.

eth3lbert:
> I like the approach, but unfortunately this seems to break our API token creation flow, at least with the proposed implementation. it seems to be missing the `NEW_IDENTIFIERS` part of https://github.com/emberjs/data/blob/v5.3.9/packages/store/src/-private/caches/identifier-cache.ts,

Ah, that's indeed unfortunate. It would be nice if upstream exported the default generation method, then we could just use it as a fallback.



> but I'm not sure if reimplementing that is worth the complexity.

Fair enough, unless the generation method approach could be easier to use without modifying other places, I think the adapter approach would make more sense.

mirage/route-handlers/crates.js
```diff
@@ -3,6 +3,10 @@ import { Response } from 'miragejs';
 import { getSession } from '../utils/session';
 import { compareIsoDates, compareStrings, notFound, pageParams, releaseTracks } from './-utils';
 
+function toCanonicalName(name) {
+  return name.toLowerCase().replace(/_/g, '-');
+}
```
eth3lbert, line 8: 🙋 The canonical implementation in the backend(database) actually replaces `-` with `_`, but here we replace `_` with `-`. Is this difference intentional?

Turbo87: > Is this difference intentional?

no, it's not intentional. I guess ultimately it doesn't really matter which way around we do it, but I'll adjust it so that it matches.

PR #10380, opened 2025-01-13
crate/delete: Add required reason input field

Turbo87:
As discussed in our last team meeting, this pull requests adds a required "Reason" input field to the crate deletion page:

Bildschirmfoto 2025-01-13 um 16 22 50

### DIFF
--- a/app/adapters/crate.js
+++ b/app/adapters/crate.js
@@ -25,6 +25,18 @@ export default class CrateAdapter extends ApplicationAdapter {
     return `${baseUrl}/${crateName}`;
   }
 
+  /** Adds a `reason` query parameter to the URL, if set in the `adapterOptions`. */
+  urlForDeleteRecord(id, modelName, snapshot) {
+    let url = super.urlForDeleteRecord(...arguments);
+
+    let reason = snapshot.adapterOptions.reason;
+    if (reason) {
+      url += `?reason=${encodeURIComponent(reason)}`;
+    }
+
+    return url;
+  }
+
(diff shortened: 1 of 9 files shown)


LawnGnome (approved):

eth3lbert:
🙋 Should we add a test case for sending with no reason input?

Turbo87:
I won't say no if someone added one :D

app/adapters/crate.js
```diff
@@ -25,6 +25,18 @@ export default class CrateAdapter extends ApplicationAdapter {
     return `${baseUrl}/${crateName}`;
   }
 
+  /** Adds a `reason` query parameter to the URL, if set in the `adapterOptions`. */
+  urlForDeleteRecord(id, modelName, snapshot) {
+    let url = super.urlForDeleteRecord(...arguments);
+
+    let reason = snapshot.adapterOptions.reason;
+    if (reason) {
+      url += `?reason=${encodeURIComponent(reason)}`;
```
eth3lbert, line 34: @Turbo87 🙋  If I understand correctly, the query string is named `message` at the backend https://github.com/rust-lang/crates.io/blob/19c80e04bd4446d96cc9ebbe7dfe4ac87c4345ce/src/controllers/krate/delete.rs#L24-L29, but here we are sending it as `reason`. I think this might be a mistake?

Turbo87: damn, you're right... 🤦‍♂️

PR #10411, opened 2025-01-17
controllers/krate/search: Use provided `plainto_tsquery_with_search_config()` fn

eth3lbert:
Previously, I used the wrong function, and the correct function was not available at that time.
The correct function `plainto_tsquery_with_search_config()` respects to `plainto_tsquery(config regconfig, query text ) → tsquery` is now available [^1], and our test cases cover this part. I think it's time we switched to using it :D

Related:
- #8052 

[^1]: https://github.com/diesel-rs/diesel_full_text_search/pull/41

### DIFF
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -6,11 +6,11 @@ use axum_extra::extract::Query;
 use axum_extra::json;
 use axum_extra::response::ErasedJson;
 use derive_more::Deref;
-use diesel::dsl::{exists, sql, InnerJoinQuerySource, LeftJoinQuerySource};
+use diesel::dsl::{exists, InnerJoinQuerySource, LeftJoinQuerySource};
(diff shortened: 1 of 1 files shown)


src/controllers/krate/search.rs
```diff
@@ -5,10 +5,11 @@ use axum::extract::FromRequestParts;
 use axum_extra::extract::Query;
 use axum_extra::json;
 use axum_extra::response::ErasedJson;
+use configuration::TsConfigurationByName;
```
Turbo87, line 8: where is `configuration` coming from?

eth3lbert: Good catch! My editor auto-imports that, and it actually comes from `diesel_full_text_search` I've updated it to be more explicit.

Turbo87 (approved):

PR #10429, opened 2025-01-21
Enable Comrak support for style alerts in Markdown

kbdharun:
This PR enables `alerts` (for GitHub style alerts https://github.com/kivikakk/comrak/pull/519) and `multiline_block_quotes` (for GitLab multiline blockquotes https://github.com/kivikakk/comrak/pull/521) Comrak extensions that were added/modified in the recent release https://github.com/kivikakk/comrak/releases/tag/v0.34.0, to the Crates Markdown renderer.

**Related discussion**: https://github.com/rust-lang/crates.io/discussions/8506

---

While I was able to build the frontend as well as backend via Docker, I have not tested the rendering with a sample crates package (due to my unfamiliarity with the codebase). Any help with testing this and adding test cases for style alerts would be appreciated.

**Extension docs**:

https://docs.rs/comrak/0.35.0/comrak/struct.ExtensionOptionsBuilder.html#method.alerts
https://docs.rs/comrak/0.35.0/comrak/struct.ExtensionOptionsBuilder.html#method.multiline_block_quotes


Turbo87:
https://github.com/rust-lang/crates.io/blob/main/tests/acceptance/readme-rendering-test.js
http://localhost:4040/tests?filter=Acceptance%20|%20README%20rendering

> Any help with testing this and adding test cases for style alerts would be appreciated.

we need two test cases:

1. checking the conversion of the markdown alert into an HTML alert. this should be added at the bottom of the file in the `crates_io_markdown` crate, which already has a bunch of similar tests.

2. checking the styling of the HTML alerts. you can find a basic smoke test for the rendering at https://github.com/rust-lang/crates.io/blob/main/tests/acceptance/readme-rendering-test.js. if you run `pnpm start` you should be able to visit http://localhost:4040/tests?filter=Acceptance%20|%20README%20rendering and see it in action. if you replace `wait percySnapshot(assert);` with `await this.pauseTest();` you can pause the test and manually check if the styling matches your expectation.

Turbo87:
@kbdharun are you still working on this?

kbdharun:
Hi @Turbo87, I tried adding a test case but unfortunately it isn't functioning properly with the alerts extension, if possible can you assist fixing it. Thanks in advance.

Turbo87:
> I tried adding a test case but unfortunately it isn't functioning properly with the alerts extension, if possible can you assist fixing it.

for the Rust test: I don't think the test is wrong, it just shows that the alert extension isn't working as intended. Most likely the issue is related to the `html_sanitizer` in the same file, which only allows a certain set of tags, attributes and values. It would probably have to be adjusted to allow the corresponding CSS classes on these elements.

for the JS test: you will probably have to adjust the assertions to make that test pass.

Turbo87:
feel free to reopen if you feel like picking this back up :)

PR #10464, opened 2025-01-28
Implement dynamic OpenGraph image url support for crate pages

hdoordt:
Related discussion: https://github.com/rust-lang/crates.io/discussions/9928

I replaced `tower_http::services::ServeFile` with a manually built response, and attempted to mimic its behavior. Furthermore, I replaced the hard coded open graph image url from `app/index.html` with a Jinja2 variable called `og_image_url`. I imagine this file is served from a bucket or the like in production, so upon deployment that variable should be replaced with the original fallback URL.

The base URL of the OG images is configurable, and should work with either a local or public instance of [`og-loc`](https://github.com/mainmatter/og-loc), or, as suggested in the related discussion, a file in a bucket.

Is this going in the right direction?

To do:

- [x] cache rendered HTML with [`moka`](https://docs.rs/moka/latest/moka/)

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -12,7 +12,7 @@ use crate::config::cdn_log_storage::CdnLogStorageConfig;
 use crate::config::CdnLogQueueConfig;
 use crate::middleware::cargo_compat::StatusCodeConfig;
 use crate::storage::StorageConfig;
-use crates_io_env_vars::{list, list_parsed, required_var, required_var_parsed, var, var_parsed};
+use crates_io_env_vars::{list, list_parsed, required_var, var, var_parsed};
 use http::HeaderValue;
 use std::collections::{HashMap, HashSet};
 use std::convert::Infallible;
@@ -74,8 +74,6 @@ pub struct Server {
 
     /// Should the server serve the frontend `index.html` for all
     /// non-API requests?
-    /// Setting this parameter requires setting
-    /// [`Self::og_image_base_url`] as well.
     pub serve_html: bool,
 
     /// Base URL for the service from which the OpenGraph images
@@ -184,14 +182,6 @@ impl Server {
             cdn_domain = storage.cdn_prefix.as_ref().map(|cdn_prefix| format!("https://{cdn_prefix}")).unwrap_or_default()
         );
 
-        let serve_html = var_parsed("SERVE_HTML")?.unwrap_or(true);
-        let og_image_base_url = serve_html
-            .then(|| {
-                required_var_parsed("OG_IMAGE_BASE_URL")
-                    .context("OG_IMAGE_BASE_URL must be set when using INDEX_HTML_TEMPLATE_PATH")
-            })
-            .transpose()?;
-
         Ok(Server {
             db: DatabasePools::full_from_environment(&base)?,
             storage,
@@ -236,7 +226,7 @@ impl Server {
                 .unwrap_or(StatusCodeConfig::AdjustAll),
             serve_dist: true,
             serve_html: true,
-            og_image_base_url,
+            og_image_base_url: var_parsed("OG_IMAGE_BASE_URL")?,
             html_render_cache_max_capacity: var_parsed("HTML_RENDER_CACHE_CAP")?.unwrap_or(1024),
             content_security_policy: Some(content_security_policy.parse()?),
         })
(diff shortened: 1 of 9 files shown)


src/config/server.rs
```diff
@@ -71,9 +73,21 @@ pub struct Server {
...
     pub serve_dist: bool,
 
-    /// Should the server serve the frontend `index.html` for all
-    /// non-API requests?
-    pub serve_html: bool,
+    /// If set the server serves the frontend `index.html` for all
+    /// non-API requests using the Jinja template at the given path.
+    /// Setting this parameter requires setting
+    /// [`Self::og_image_base_url`] as well.
+    pub index_html_template_path: Option<PathBuf>,
```
Turbo87, line 80: do we need this to be configurable?

src/middleware/ember_html.rs
```diff
@@ -26,11 +67,63 @@ pub async fn serve_html(request: Request, next: Next) -> Response {
...
+                let len = suffix.find('/').unwrap_or(suffix.len());
+                let krate = &suffix[..len];
+
+                // `state.config.og_image_base_url` will always be `Some` as that's required
+                // if `state.config.index_html_template_path` is `Some`, and otherwise this
+                // middleware won't be executed; see `crate::middleware::apply_axum_middleware`.
+                if let Ok(og_img_url) = state.config.og_image_base_url.as_ref().unwrap().join(krate)
+                {
+                    break 'og Cow::from(og_img_url.to_string());
+                }
```
Turbo87, line 90: I think we might want to use `.png` file extensions?

src/middleware/ember_html.rs
```diff
@@ -26,11 +67,63 @@ pub async fn serve_html(request: Request, next: Next) -> Response {
...
+            return (StatusCode::METHOD_NOT_ALLOWED, headers).into_response();
+        }
+
+        // Come up with an Open Graph image URL. In case a crate page is requested,
+        // we use the crate's name and the OG image base URL from config to
+        // generate one, otherwise we use the fallback image.
+        let og_image_url = 'og: {
+            if let Some(suffix) = path.strip_prefix(PATH_PREFIX_CRATES) {
+                let len = suffix.find('/').unwrap_or(suffix.len());
+                let krate = &suffix[..len];
```
Turbo87, line 82: might make sense to extract a `extract_crate_name(...)` fn. that way it's a easier to write a few tests for it too.

src/middleware/ember_html.rs
```diff
@@ -26,11 +67,63 @@ pub async fn serve_html(request: Request, next: Next) -> Response {
...
+
+                // `state.config.og_image_base_url` will always be `Some` as that's required
+                // if `state.config.index_html_template_path` is `Some`, and otherwise this
+                // middleware won't be executed; see `crate::middleware::apply_axum_middleware`.
+                if let Ok(og_img_url) = state.config.og_image_base_url.as_ref().unwrap().join(krate)
+                {
+                    break 'og Cow::from(og_img_url.to_string());
```
Turbo87, line 89: TIL 😂 

but I'm wondering if this could be simplified somehow. we could extract a `generate_og_image_url()` fn or something like that, and then use `return` instead of the labeled break with return value.

src/middleware/ember_html.rs
```diff
@@ -26,11 +67,63 @@ pub async fn serve_html(request: Request, next: Next) -> Response {
...
+
         // Serve static Ember page to bootstrap the frontend
-        ServeFile::new("dist/index.html")
-            .oneshot(request)
-            .await
-            .map(|response| response.map(axum::body::Body::new))
+        Response::builder()
+            .header(header::CONTENT_TYPE, "text/html")
+            .header(header::CONTENT_LENGTH, html.len())
+            .body(axum::body::Body::new(html))
```
Turbo87, line 126: I guess we could probably use https://docs.rs/axum/latest/axum/response/struct.Html.html for this?

src/middleware/ember_html.rs
hdoordt: Neat right? ;) . But yeah I agree a `fn` would be an improvement

src/config/server.rs
hdoordt: I'm not sure, actually. It depends on the exact file structure. It that's stable, we can just hard code `dist/index.html`

Turbo87: thanks! I think since it was already hardcoded before that makes the PR a little more digestable. we can always change it later if there is a need for it :)

Turbo87 (approved):
I've rebased and simplified the PR a little bit more. Let's give this a go and see how it behaves on the server.

hdoordt:
Great, thanks!

PR #10489, opened 2025-01-31
fix: only autofocus search form on hero header

### LINKED ISSUE #10480: Don't auto-focus search input on page load
### Current Behavior

On crate pages (e.g. `https://crates.io/crates/tokio`), the search input field is automatically focused on page load, stealing focus from keyboard-based navigation tools which the user may have configured. This can be a mild annoyance to users of assistive tools and those with keyboard-centric workflows. It arguably does not provide much value to other users who can easily focus the search input by pressing `S` or `/`.

### Expected Behavior

The input should not be auto-focused on page load.

### Steps To Reproduce

_No response_

### Environment

- Browser:
- OS:

### Anything else?

_No response_

b0o:
fixes #9701
fixes #10480

### DIFF
--- a/app/components/header.hbs
+++ b/app/components/header.hbs
@@ -10,7 +10,7 @@
         The Rust community&rsquo;s crate registry
       </h1>
 
-      <SearchForm @size={{if @hero "big"}} />
+      <SearchForm @size={{if @hero "big"}} @autofocus={{@hero}} />
     </div>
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):

Turbo87 (approved):
looks good, thank you! :)

PR #10513, opened 2025-02-05
Replace `ember-tooltips` with `floating-ui`

Turbo87:
https://floating-ui.com is a low-level library to implement e.g. tooltips. This PR replaces the unmaintained `ember-tooltips` addon with a custom `Tooltip` component based on `floating-ui` that works roughly similar. This should hopefully then unblock the `@ember/string` update in #10354.

### Before

Bildschirmfoto 2025-02-05 um 13 04 55

### After

Bildschirmfoto 2025-02-05 um 13 04 31

yes, we lost the arrow, but does that really matter? 😅 

floating-ui does have an "arrow middleware", but I couldn't make it work properly in combination with `flip()`. since it would require more complexity for IMHO little gain I think it's fine to ship this without arrows.

### DIFF
--- a/app/components/tooltip.js
+++ b/app/components/tooltip.js
@@ -1,0 +1,15 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+
+export default class Tooltip extends Component {
+  @tracked hidden = true;
+
+  @action hide() {
+    this.hidden = true;
+  }
+
+  @action show() {
+    this.hidden = false;
+  }
+}
--- a/app/components/crate-header.hbs
+++ b/app/components/crate-header.hbs
@@ -9,10 +9,10 @@
           {{svg-jar "trash"}}
           Yanked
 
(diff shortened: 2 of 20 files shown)


Rustin170506:
@Turbo87 Some tests failed.

Turbo87:
> Some tests failed

should be fixed now :)

app/components/tooltip.js
```diff
@@ -0,0 +1,15 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+
+export default class UserAvatar extends Component {
```
Rustin170506, line 5: I guess this is a typo?

Rustin170506 (commented):
Thanks!

app/components/tooltip.js
Turbo87: lol, yeah, dumb copy-paste issue 😂

Turbo87: should be fixed :)

PR #10519, opened 2025-02-06
Add `default_versions.num_versions` column

eth3lbert:
The total number of versions is one of the blockers for avoiding loading full versions within the application. To address this, we can either send a separate request to retrieve the number, or, as suggested by @Turbo87, store the value in db and include it in the crate response. The latter approach seems more promising, as it eliminates the need for an extra request solely to obtain the total.

To ensure the value is up-to-date and maintain deployment convenience, the implementation will be separated into multiple PRs:

- [x] db: migration and save value
- [ ] expose on the API and leverage in the application

### DIFF
--- a/crates/crates_io_database/src/schema.rs
+++ b/crates/crates_io_database/src/schema.rs
@@ -408,6 +408,8 @@ diesel::table! {
         crate_id -> Int4,
         /// Reference to the version in the `versions` table.
         version_id -> Int4,
+        /// The total number of versions.
+        num_versions -> Nullable<Int4>,
     }
 }
 
--- a/crates/crates_io_database_dump/src/snapshots/crates_io_database_dump__tests__sql_scripts@export.sql.snap
+++ b/crates/crates_io_database_dump/src/snapshots/crates_io_database_dump__tests__sql_scripts@export.sql.snap
@@ -18,7 +18,7 @@ BEGIN ISOLATION LEVEL REPEATABLE READ, READ ONLY;
     \copy (SELECT "crate_id", "created_at", "created_by", "owner_id", "owner_kind" FROM "crate_owners" WHERE NOT deleted) TO 'data/crate_owners.csv' WITH CSV HEADER
 
     \copy "versions" ("bin_names", "categories", "checksum", "crate_id", "crate_size", "created_at", "description", "documentation", "downloads", "edition", "features", "has_lib", "homepage", "id", "keywords", "license", "links", "num", "num_no_build", "published_by", "repository", "rust_version", "updated_at", "yanked") TO 'data/versions.csv' WITH CSV HEADER
-    \copy "default_versions" ("crate_id", "version_id") TO 'data/default_versions.csv' WITH CSV HEADER
+    \copy "default_versions" ("crate_id", "num_versions", "version_id") TO 'data/default_versions.csv' WITH CSV HEADER
     \copy "dependencies" ("crate_id", "default_features", "explicit_name", "features", "id", "kind", "optional", "req", "target", "version_id") TO 'data/dependencies.csv' WITH CSV HEADER
     \copy "version_downloads" ("date", "downloads", "version_id") TO 'data/version_downloads.csv' WITH CSV HEADER
 COMMIT;
--- a/migrations/2025-02-05-083109_add-num-versions-column/up.sql
+++ b/migrations/2025-02-05-083109_add-num-versions-column/up.sql
@@ -1,0 +1,27 @@
+ALTER TABLE default_versions
+    ADD COLUMN num_versions INTEGER;
+
+COMMENT ON COLUMN default_versions.num_versions IS 'The total number of versions.';
+
+CREATE OR REPLACE FUNCTION update_num_versions_from_versions() RETURNS TRIGGER AS $$
+BEGIN
+    IF (TG_OP = 'INSERT') THEN
+        INSERT INTO default_versions (crate_id, version_id, num_versions)
+        VALUES (NEW.crate_id, NEW.id, 1)
+        ON CONFLICT (crate_id) DO UPDATE
+        SET num_versions = EXCLUDED.num_versions + 1;
+        RETURN NEW;
+    ELSIF (TG_OP = 'DELETE') THEN
+        UPDATE default_versions
+        SET num_versions = num_versions - 1
+        WHERE crate_id = OLD.crate_id;
+        RETURN OLD;
+    END IF;
+END
+$$ LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS trigger_update_num_versions_from_versions ON versions;
+CREATE TRIGGER trigger_update_num_versions_from_versions
+     AFTER INSERT OR DELETE ON versions
+     FOR EACH ROW
+     EXECUTE PROCEDURE update_num_versions_from_versions();
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -445,7 +445,7 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
         // Upsert the `default_value` determined by the existing `default_value` and the
         // published version. Note that this could potentially write an outdated version
         // (although this should not happen regularly), as we might be comparing to an
-        // outdated value.
+        // outdated value. The initial record will be handled by the trigger function.
         //
         // Compared to only using a background job, this prevents us from getting into a
         // situation where a crate exists in the `crates` table but doesn't have a default
@@ -470,14 +470,6 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
             // Update the default version asynchronously in a background job
             // to ensure correctness and eventual consistency.
             UpdateDefaultVersion::new(krate.id).enqueue(conn).await?;
-        } else {
-            diesel::insert_into(default_versions::table)
-                .values((
-                    default_versions::crate_id.eq(krate.id),
-                    default_versions::version_id.eq(version.id),
-                ))
-                .execute(conn)
-                .await?;
         }
 
         // Update all keywords for this crate
--- a/crates/crates_io_database_dump/src/dump-db.toml
+++ b/crates/crates_io_database_dump/src/dump-db.toml
@@ -105,6 +105,7 @@ dependencies = ["crates", "versions"]
 [default_versions.columns]
 crate_id = "public"
 version_id = "public"
+num_versions = "public"
 
 [deleted_crates]
 dependencies = ["users"]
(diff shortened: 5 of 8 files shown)


eth3lbert:
Admin could leverage the following SQL to set the `total` values:
```sql
WITH filtered AS (
	SELECT crate_id
	FROM default_versions
	WHERE total = 0
	ORDER BY crate_id
	-- the limit can be tuned or removed as needed.
	LIMIT 10000
), to_update AS (
	SELECT crate_id, count(*) AS total
	FROM versions
	JOIN filtered USING (crate_id)
	GROUP BY crate_id
)
UPDATE default_versions
SET total = to_update.total
FROM to_update
WHERE default_versions.crate_id = to_update.crate_id;
```

crates/crates_io_database_dump/src/snapshots/crates_io_database_dump__tests__sql_scripts@export.sql.snap
```diff
@@ -1,6 +1,7 @@
 ---
 source: crates/crates_io_database_dump/src/lib.rs
 expression: content
+snapshot_kind: text
```
Turbo87: `cargo install cargo-insta` 😉 

the latest release removed `snapshot_kind: text` noise again 😅

crates/crates_io_database/src/schema.rs
```diff
@@ -408,6 +408,8 @@ diesel::table! {
         crate_id -> Int4,
         /// Reference to the version in the `versions` table.
         version_id -> Int4,
+        /// The total number of versions.
+        total -> Int4,
```
Turbo87, line 412: I think I'd prefer something like `num_versions` since it's a bit more self-descriptive

src/controllers/krate/publish.rs
```diff
@@ -450,22 +449,45 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
...
-                    .set(default_versions::version_id.eq(version.id))
+            let update_base = diesel::update(default_versions::table)
+                    .filter(default_versions::crate_id.eq(krate.id));
+
+            // If `total` isn't set (it defaults to 0), this only occurs when a crate already
+            // exists and releases a new version before the admin sets the total.
+            // In this scenario, the `total` value is updated in a background job. This case should
+            // be safe to remove once all `total` values have been set.
+            let is_new = existing_default_version < published_default_version;
+            let default_total = existing_total == 0;
```
Turbo87, line 466: wouldn't it be easier to have the column nullable for now and work with `None/Some(..)` instead?

src/models/default_versions.rs
```diff
@@ -74,7 +74,10 @@ pub async fn update_default_version(
         ))
         .on_conflict(default_versions::crate_id)
         .do_update()
-        .set(default_versions::version_id.eq(default_version.id))
+        .set((
+            default_versions::version_id.eq(default_version.id),
+            default_versions::total.eq(greatest(total as i32, default_versions::total)),
```
Turbo87, line 79: I guess in the uncommon case of us deleting specific versions as an admin action this `greatest()` call will likely fail to produce the correct number 🤔

crates/crates_io_database/src/schema.rs
eth3lbert: fair enough! I'll update it!

src/models/default_versions.rs
Turbo87:
```suggestion
            default_versions::total.eq(total),
```

why not just that?

Turbo87: > This could happen if a new version comes out during that time. And since this is a deduplicated job, I'm not sure if it would break anything.

The job is only deduplicated if the other job is not yet started.

If I understand you correctly, you are worried about the following scenario: publish A causes default versions update job A. while job A is running or queued up publish B comes in and queues up job B.

if job A is running then job B will not be deduplicated and should eventually write the correct value to the database. if job A is still queued up then job B will indeed be deduplicated, but when job A runs the second version should have been created already which should also result in a consistent value.

there are probably other theoretical scenarios in which we return an incorrect value, but I'm not sure if they matter much in practice. if we return an off-by-one number here for a bit it's probably not the end of the world, though we might want to document in our API docs that this is an eventually-consistent value.

eth3lbert: > If I understand you correctly, you are worried about the following scenario: publish A causes default versions update job A. while job A is running or queued up publish B comes in and queues up job B.

I'm concerned about the following scenario: When there are two publishes, A and B, publish A triggers job a. While job a is running and calculating the default_version, the total count is determined. However, if publish B completes before the update statement within job a is executed, the calculated value could become stale.



> The job is only deduplicated if the other job is not yet started.

Then this would be great! This would guarantee that even if an off-by-one error occurs in the running job, it should be correct in the subsequent one. I think I'm completely fine with it set to cached value.

---

🙋‍♂️ I have another question regarding the `verify_default_version`. When do we perform the verification? Is this performed regularly?

Turbo87: > I have another question regarding the `verify_default_version`. When do we perform the verification? Is this performed regularly?

there is an admin tool that can run it, but it is not running regularly. I've used it a couple of times after we did the initial switch to the `default_versions` table to verify that everything was consistent.

src/controllers/krate/publish.rs
```diff
@@ -450,22 +449,44 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
...
+            let update_base = diesel::update(default_versions::table)
+                    .filter(default_versions::crate_id.eq(krate.id));
+
+            // If `num_versions` isn't set, this only occurs when a crate already exists and
+            // releases a new version before the admin sets the value.
+            // In this scenario, the `num_versions` value is updated in a background job. This case
+            // should be safe to remove once all `num_versions` values have been set.
+            let is_new = existing_default_version < published_default_version;
+            let fut = match (is_new, existing_num_versions.is_none()) {
+                (true, true) => {
```
Turbo87, line 467:
```suggestion
            let fut = match (is_new, existing_num_versions) {
                (true, None) => {
```

😉

Turbo87 (approved):
I think I'm good with this, but would appreciate a quick glance for @LawnGnome before we move forward :)

LawnGnome (commented):
None of this is blocking, but here are my thoughts:

Firstly, `default_versions` is now not as accurate a name for the table. It's probably not worth renaming right now, but `computed_versions` might be closer to what we're doing.

Secondly, I don't love the extra complexity in the publish endpoint, honestly. I know some of that is transitional, but I get slightly twitchy any time denormalised table updates happen in one place outside of the model layer, because that makes it easy to not do it in unusual circumstances.

For default version calculation, we don't really have a choice (since it's — at best — challenging to do in SQL), but it feels like we could calculate `num_versions` with a trigger on the `versions` table, which would get rid of that potential issue.

I'm not saying we _must_ use a trigger, and I wouldn't block the merging of this PR as-is, but @eth3lbert, I'm interested in whether you considered that here?

Turbo87:
> but it feels like we could calculate `num_versions` with a trigger on the `versions` table, which would get rid of that potential issue.

that's a good point. if we can do this within the database then we could avoid a lot of the application code complexity.

I assume a "insert version => increment num_versions, delete version => decrement num_versions" might be sufficient with a one-time query to set the initial counts.

eth3lbert:
> I'm not saying we _must_ use a trigger, and I wouldn't block the merging of this PR as-is, but @eth3lbert, I'm interested in whether you considered that here?

Good suggestion! I totally forgot about the trigger function here. Triggers can be a bit tricky sometimes, but we already have similar one for categories, so I'm happy to use a trigger. Thanks again!

Turbo87:
do we still need the second commit?

eth3lbert:
> do we still need the second commit?

I see this as a last resort, but it should be okay to remove.

Turbo87:
probably fine to drop it. but I guess we should adjust the publish endpoint to insert `num_versions: 1` for the first record or change the default to 1.

eth3lbert:
> but I guess we should adjust the publish endpoint to insert `num_versions: 1` for the first record or change the default to 1.

I added a test for this to determine if a default value of 0 is sufficient.

eth3lbert:
> but I guess we should adjust the publish endpoint to insert `num_versions: 1` for the first record or change the default to 1.

Yeah, it looks like we must default to 1 or set it to 1 in endpoint for first version. I actually lean towards setting it in the publish endpoint and making the default value `null`. This has the benefit that existing crates would remain `null` since `null + 1 = null`, making it easier to determine if this has been initialized by an admin yet. What do you think, @Turbo87 @LawnGnome?

migrations/2025-02-05-083109_add-num-versions-column/up.sql
```diff
@@ -0,0 +1,26 @@
...
+    ADD COLUMN num_versions INTEGER NOT NULL DEFAULT 0;
+
+COMMENT ON COLUMN default_versions.num_versions IS 'The total number of versions.';
+
+CREATE OR REPLACE FUNCTION update_num_versions_from_versions() RETURNS TRIGGER AS $$
+BEGIN
+    IF (TG_OP = 'INSERT') THEN
+        UPDATE default_versions
+        SET num_versions = num_versions + 1
+        WHERE crate_id = NEW.crate_id;
```
Turbo87, line 11: I'm wondering if we should use `INSERT INTO default_versions ON CONFLICT (crate_id) DO UPDATE SET` here. that should allow us to completely remove the `diesel::insert_into(default_versions::table)` part from the publish endpoint and also set the `num_versions: 1` default here at the same time.

eth3lbert: I also considered this previously :D I am fine with this.

eth3lbert: updated!

LawnGnome:
> Yeah, it looks like we must default to 1 or set it to 1 in endpoint for first version. I actually lean towards setting it in the publish endpoint and making the default value `null`. This has the benefit that existing crates would remain `null` since `null + 1 = null`, making it easier to determine if this has been initialized by an admin yet. What do you think, @Turbo87 @LawnGnome?

The only reason `num_versions` is nullable now is for the initial post-migration case where we haven't updated the records for existing crates, right?

If that's the case, then I'd rather set a default, make it `NOT NULL`, and just accept that we need to run the post-migration queries to set `num_versions` immediately after deployment. Since we're merging this PR before the PR that actually uses this value, that seems safe enough as long as we:

1. Deploy this PR.
2. Set all the `num_versions` values.
3. Deploy the PR that actually uses this in the API.

Turbo87 (approved):
> Do you also suggest using a non-null default value?

I think I prefer nullable for now

migrations/2025-02-05-083109_add-num-versions-column/up.sql
```diff
@@ -0,0 +1,27 @@
...
+
+COMMENT ON COLUMN default_versions.num_versions IS 'The total number of versions.';
+
+CREATE OR REPLACE FUNCTION update_num_versions_from_versions() RETURNS TRIGGER AS $$
+BEGIN
+    IF (TG_OP = 'INSERT') THEN
+        INSERT INTO default_versions (crate_id, version_id, num_versions)
+        VALUES (NEW.crate_id, NEW.id, 1)
+        ON CONFLICT (crate_id) DO UPDATE
+        SET num_versions = EXCLUDED.num_versions + 1;
```
Turbo87, line 12: is `EXCLUDED.num_versions` actually correct here? shouldn't this be `default_versions.num_versions`? 🤔 

maybe I'm missing something, but isn't `EXCLUDED.num_versions` referring to the `1` two lines above?

eth3lbert: Ah, 🤦 🙈 , it should definitely be `default_versions.num_versions + 1` Do you want me to send a patch migration, or will you take care of it?

Turbo87: I'm currently busy with some other stuff, so if you have time, that would be great :)

Ideally with a regression test case 😅

Turbo87:
> ```sql
> WITH filtered AS (
> 	SELECT crate_id
> 	FROM default_versions
> 	WHERE num_versions is NULL
> 	ORDER BY crate_id
> 	-- the limit can be tuned or removed as needed.
> 	LIMIT 10000
> ), to_update AS (
> 	SELECT crate_id, count(*) AS num_versions
> 	FROM versions
> 	JOIN filtered USING (crate_id)
> 	GROUP BY crate_id
> )
> UPDATE default_versions
> SET num_versions = to_update.num_versions
> FROM to_update
> WHERE default_versions.crate_id = to_update.crate_id;
> ```

✅

PR #10522, opened 2025-02-06
capitalize first sentence of crate delete mail

Uriopass:
Tiny thing, kind of bugged me when I received the mail.

### DIFF
--- a/src/controllers/krate/delete.rs
+++ b/src/controllers/krate/delete.rs
@@ -207,7 +207,7 @@ impl Email for CrateDeletionEmail<'_> {
         format!(
             "Hi {},
 
-your \"{}\" crate has been deleted, per your request.
+Your \"{}\" crate has been deleted, per your request.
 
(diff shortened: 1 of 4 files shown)


Turbo87 (commented):
I'm not a native english speaker, but since the email begins with "Hi {username}," wouldn't this still be part of the same sentence and thus not have captialization? 😅

Uriopass:
It's a very wide convention to capitalize the first sentence after the greeting, although not completely unanimous it's the majority.
See: https://english.stackexchange.com/questions/11906/should-i-capitalize-the-starting-sentence-after-a-greeting-that-ends-in-a-comma

> However, as far as I can tell, there is no good reason to oppose conventional style in this case. I would therefore recommend that you stick with the tried-and-true style of using a capital letter to start the letter proper:

Turbo87 (approved):
Seems reasonable, thanks!

I've rebased this PR and fixed the broken snapshot files in the test suite.

Uriopass:
Thanks! I should have fixed the snapshot myself. Will do if there's a next time :)

PR #10529, opened 2025-02-07
Add `?include=versions` support for `GET /api/v1/crates/{name}/downloads` endpoint

eth3lbert:
While #10344 is merged, we currently include the `default_version`. Ideally, we should not load all versions for the index/version page. However, there are still some blockers. One blocker is `VersionDownload` due to its` belongsTo` relationship with `version`. This requires us to wait for `versions` to be loaded before loading `downloads`.
This PR aims to add an include mode to the `GET /api/v1/crates/{name}/downloads` endpoint. This allows it to include the related `versions` and avoids the need to wait for `versions` to be fully loaded in app.

### DIFF
--- a/e2e/acceptance/crate.spec.ts
+++ b/e2e/acceptance/crate.spec.ts
@@ -111,22 +111,6 @@ test.describe('Acceptance | crate page', { tag: '@acceptance' }, () => {
     await expect(page.locator('[data-test-try-again]')).toHaveCount(0);
   });
 
-  test('other versions loading error shows an error message', async ({ page, msw }) => {
-    let crate = msw.db.crate.create({ name: 'nanomsg' });
-    msw.db.version.create({ crate, num: '0.6.0' });
-    msw.db.version.create({ crate, num: '0.6.1' });
-
-    await msw.worker.use(http.get('/api/v1/crates/:crate_name/versions', () => HttpResponse.json({}, { status: 500 })));
-
-    await page.goto('/');
-    await page.click('[data-test-just-updated] [data-test-crate-link="0"]');
-    await expect(page).toHaveURL('/crates/nanomsg/0.6.0');
-    await expect(page.locator('[data-test-404-page]')).toBeVisible();
-    await expect(page.locator('[data-test-title]')).toHaveText('nanomsg: Failed to load version data');
-    await expect(page.locator('[data-test-go-back]')).toHaveCount(0);
-    await expect(page.locator('[data-test-try-again]')).toBeVisible();
-  });
-
   test('works for non-canonical names', async ({ page, msw }) => {
     let crate = msw.db.crate.create({ name: 'foo-bar' });
     msw.db.version.create({ crate });
--- a/src/controllers/krate/downloads.rs
+++ b/src/controllers/krate/downloads.rs
@@ -6,17 +6,36 @@
 use crate::app::AppState;
 use crate::controllers::krate::CratePath;
 use crate::models::download::Version;
-use crate::models::VersionDownload;
-use crate::schema::{version_downloads, versions};
-use crate::util::errors::AppResult;
-use crate::views::EncodableVersionDownload;
+use crate::models::{User, Version as FullVersion, VersionDownload, VersionOwnerAction};
+use crate::schema::{version_downloads, version_owner_actions, versions};
+use crate::util::errors::{bad_request, AppResult, BoxedAppError};
+use crate::views::{EncodableVersion, EncodableVersionDownload};
+use axum::extract::FromRequestParts;
+use axum_extra::extract::Query;
 use axum_extra::json;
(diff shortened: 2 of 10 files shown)


e2e/acceptance/crate.spec.ts
```diff
@@ -111,22 +111,6 @@ test.describe('Acceptance | crate page', { tag: '@acceptance' }, () => {
...
-
-    await page.goto('/');
-    await page.click('[data-test-just-updated] [data-test-crate-link="0"]');
-    await expect(page).toHaveURL('/crates/nanomsg/0.6.0');
-    await expect(page.locator('[data-test-404-page]')).toBeVisible();
-    await expect(page.locator('[data-test-title]')).toHaveText('nanomsg: Failed to load version data');
-    await expect(page.locator('[data-test-go-back]')).toHaveCount(0);
-    await expect(page.locator('[data-test-try-again]')).toBeVisible();
-  });
-
```
eth3lbert, line 129: The versions should only be used to display the count on the crate page. Therefore, I think it should be safe to remove this test case. Let me know if this is inappropriate.

e2e/acceptance/crate.spec.ts
eth3lbert, line 1: The versions should only be used to display the count on the crate page. Therefore, I think it should be safe to remove this test case. Let me know if this is inappropriate.

src/controllers/krate/downloads.rs
```diff
@@ -75,10 +99,74 @@ pub async fn get_crate_downloads(state: AppState, path: CratePath) -> AppResult<
...
 
+    let include = params
+        .include
+        .as_ref()
+        .map(|mode| ShowIncludeMode::from_str(mode))
+        .transpose()?
+        .unwrap_or_default();
+
+    if include.versions {
+        latest_five.sort_unstable_by_key(|it| cmp::Reverse(it.id));
```
Turbo87, line 110: I assume this sort operation is the reason for the `split_at_mut()` change above? is there a reason why we need to re-sort here? can't we use `latest_five` in the reverse semver order that they should already be in?

Turbo87: `VersionOwnerAction::for_versions` is using `.grouped_by(versions)` internally to accomplish that. seems less brittle to me 🤔

eth3lbert: I'm not sure I understand the feedback. Do you mean we should encapsulate this in models, or share a hashmap for publishers and owner actions?

---

I also noticed that we should probably query minimal versions data instead of full columns for downloads.

Turbo87: I'm talking about https://docs.rs/diesel/latest/diesel/associations/trait.GroupedBy.html. If we used that for the `id_and_publishers` query below too then we wouldn't need the sorting and can guarantee that the mapping is correct, even if for some unlikely reason one of the ids couldn't be found.

eth3lbert: Rebased and rewrote with `grouped_by(latest_five)`.

src/controllers/krate/downloads.rs
```diff
@@ -80,10 +112,88 @@ pub async fn get_crate_downloads(state: AppState, path: CratePath) -> AppResult<
...
+        let actions = actions.grouped_by(latest_five);
+        let versions = versions_and_publishers
+            .into_iter()
+            .zip(actions)
+            .filter_map(|(vp, actions)| {
+                vp.into_iter().next().map(|(version, publisher)| {
+                    EncodableVersion::from(version, &path.name, publisher, actions)
+                })
+            })
+            .collect::<Vec<_>>();
```
eth3lbert, line 126: Or, we could rewrite this by grouping things ourselves instead of using `grouped_by()`:
```rust
        let mut versions_and_publishers = versions_and_publishers
            .into_iter()
            .map(|it| (it.0.id, (it, vec![])))
            .collect::<std::collections::HashMap<_, _>>();
        for it in actions {
            versions_and_publishers
                .entry(it.0.version_id)
                .and_modify(|(_, actions)| {
                    actions.push(it);
                });
        }
        let versions = latest_five
            .iter()
            .filter_map(|it| versions_and_publishers.remove(&it.id))
            .map(|((version, publisher), actions)| {
                EncodableVersion::from(version, &path.name, publisher, actions)
            })
            .collect::<Vec<_>>();
```

Turbo87: what advantage would that have?

eth3lbert: It could potentially have an advantage when we want more advanced control, such as a custom hasher (nohash/identity) for int domain hashmap. Otherwise, not really.

Turbo87: https://en.m.wikipedia.org/wiki/KISS_principle 😉

Turbo87 (approved):
let's give this a try :)

PR #10540, opened 2025-02-09
List crate features so they're readable

### LINKED ISSUE #465: Crate page should list the crate's Cargo "features"
It's currently hard to discover optional [features](http://doc.crates.io/manifest.html#the-features-section) for a crate.  It would be useful for crates.io to list these along with the dependencies.  Unfortunately there's no way to declare docs or other metadata for a feature, so this will just have to be a flat list, but maybe more info can be added in the future.

NthMetal:
fixes #465 

Just a UI update to move the version feature list to a separate togglable panel.
Currently you can only view them on hover. Which isn't readable with long lists and doesn't allow you to easily select what you wanna see.
Before: 
image

After:
image
image
image

Thoughts, suggestions, improvements, etc. appreciated!

### DIFF
--- a/app/components/version-list/row.hbs
+++ b/app/components/version-list/row.hbs
@@ -94,21 +94,15 @@
         {{/if}}
 
         {{#if @version.featureList}}
-          <span local-class="num-features" data-test-feature-list>
-            {{svg-jar "checkbox"}}
-            {{@version.featureList.length}} {{if (eq @version.featureList.length 1) "Feature" "Features"}}
-
-            <Tooltip local-class="tooltip">
-              <ul local-class="feature-list">
-                {{#each @version.featureList as |feature|}}
-                  <li>
-                    {{svg-jar (if feature.isDefault "checkbox" "checkbox-empty")}}
-                    {{feature.name}}
-                  </li>
-                {{/each}}
-              </ul>
-            </Tooltip>
-          </span>
+            <button
+              type="button"
+              local-class="features-button"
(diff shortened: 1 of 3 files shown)


NthMetal:
Ok thanks for getting back to me @Turbo87 

The intent with the pr was to make it more functional, not really pretty, that was just a bonus. The hover for features isn't readable and is hard to do things with. I can't open it and compare the list of features with my Cargo.toml. I can't select stuff and copy and paste if the list is too short. I can't open the list to even make sure i'm typing the right feature in my IDE. I can't ctrl+f and search for a feature.

Right now if I want more than just a quick glance at features I have to open up my chrome inspect menu and look at it there or search up the package on lib.rs. The data is there, it's just not usable because of the hover.

Having it in a separate tab sounds great, I contemplated doing that initially, but figured it would be easier to just make it expandable.

Were any of those other issues considered? They weren't really mentioned in this PR, but were brought up in the linked issue.

Turbo87:
> Right now if I want more than just a quick glance at features I have to open up my chrome inspect menu and look at it there or search up the package on lib.rs.

there is also https://docs.rs/crate/bevy/latest/features, but admittedly lib.rs gives a bit more information

> Were any of those other issues considered?

I'm not sure what "other issues" you mean. All of the points you brought up should be addressed if we move the features list to a dedicated tab, but as I mentioned in the comment above, this is at least partially blocked on https://rust-lang.github.io/rfcs/3416-feature-metadata.html.

PR #10554, opened 2025-02-11
Simplify `crate_owner_invitations` expiry checks

Turbo87:
This PR contains part two of #10553. It changes the `expires_at` column to be `not null` and then simplifies the expiry checks to use `expires_at` instead of "`created_at` + configured expiry duration".

As a nice side-effect, this PR removes the code dependency from the `CrateOwnerInvitation` model to our server config struct.

This should only be merged and deployed after #10553 has already been deployed and the `expires_at` column backfilled.


Turbo87:
> This should only be merged and deployed after #10553 has already been deployed and the `expires_at` column backfilled.

the `expires_at` column has been backfilled via `update crate_owner_invitations set expires_at = created_at + interval '30 day';`

PR #10558, opened 2025-02-11
db: Fix incorrect Upsert for `num_versions`

eth3lbert:
Per https://github.com/rust-lang/crates.io/pull/10519/files#r1951118907. 🤦‍♂️ 🙈 

This fixes the misused `EXCLUDED` which would cause the `num_versions` to remain at 2 during conflicts.


Turbo87 (approved):
thanks :)

PR #10572, opened 2025-02-13
database: Alter `teams.org_id` column to be `not null`

Turbo87:
Out of the roughly 1100 teams in our database only 10 have `org_id: NULL`. These teams belong to GitHub organizations that don't exist anymore, so the corresponding teams are essentially useless. See https://github.com/rust-lang/crates.io/pull/2796#issuecomment-718326067 for more context.

Since the nullability of the column makes our code more complicated than it needs to be, this commit proposes to remove these teams and turn the column into a `not null` column.


eth3lbert (commented):
The migration SQL looks good to me.

---

Looking at just the migration script, this seems like an irreversible change that could result in data loss. We probably won't need to revert it, but I'm wondering if there's a good way to undo this type of irreversible migration, including the data?

Turbo87:
> Looking at just the migration script, this seems like an irreversible change that could result in data loss. We probably won't need to revert it, but I'm wondering if there's a good way to undo this type of irreversible migration, including the data?

if it's necessary to restore it, we have the original data in the current database dump and can restore it from there.

eth3lbert (approved):
With a backup, I don't see a problem with this.

PR #10603, opened 2025-02-17
controllers/version: Load required data only

eth3lbert:
This PR primarily addresses three things:

- Avoiding loading the entire `Crate` when only the `Version` is needed.
- Loading the required data in a single query based on the request, reusing the same or similar queries as much as possible.
- Adding a helper function to handle `not found` scenarios.

First, I assume that in most cases, the given inputs are valid, then it would be great to load them in a single query.

Second, since most endpoints do not require all columns from `Crate` and `Version`, it would be nice to load only the necessary columns to reduce data loading.

Third, when only `Version` is required, it would be ideal to load only the ID of `Crate` to determine its existence.

I also want to thank @weiznich for talking with me during the all-hands to sketch out how to implement this properly using Diesel without overcomplicating things. I hope this aligns with what you suggested!

### DIFF
--- a/src/controllers/version.rs
+++ b/src/controllers/version.rs
@@ -15,7 +15,6 @@ use serde::de::Error;
 use serde::{Deserialize, Deserializer};
 use utoipa::IntoParams;
 
-use crate::controllers::krate::load_crate;
 use crate::models::{Crate, Version};
 use crate::schema::{crates, versions};
 use crate::util::errors::{AppResult, crate_not_found, version_not_found};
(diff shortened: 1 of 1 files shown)


eth3lbert:
Rebased and conflict resolved!

Turbo87 (commented):
I'm somewhat unsure whether the extra complexity is worth it. @LawnGnome any opinions?

Turbo87 (approved):
👍 

but I guess the PR title and description still need to be adjusted 😅 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/10603/07920c13aa8d428105fa2ea6fc1cf74e803775b8..7da670885eff037ffeccded5c919b9c159245659)*

eth3lbert:
> but I guess the PR title and description still need to be adjusted 😅

Yeah, this was just me trying out a new workflow with `jj` at the venue before leaving. I'll correct the title and description later to align with the proposed content.

eth3lbert:
Both the title and description should now align with the content!

PR #10604, opened 2025-02-17
Use timestamps with timezones

Turbo87:
It generally appears to be recommended to use `timestamptz` (timestamp with time zone) instead of `timestamp` (timestamp without time zone), unless the saved timestamp explicitly should not have a timezone. Since we always treat these timestamps as UTC though, we should be using `timestamptz` instead.

This commit accordingly also switches `NaiveDateTime` for `DateTime<Utc>` in most cases in our codebase. This change also allowed us to remove the `rfc3339` serde helper module, since we no longer need the `DateTime<Utc> -> NaiveDateTime` conversion performed by this module.


crates/crates_io_database/src/models/token.rs
```diff
@@ -80,7 +77,7 @@ impl ApiToken {
             .transaction(|conn| {
                 async move {
                     diesel::update(tokens)
-                        .set(api_tokens::last_used_at.eq(now.nullable()))
+                        .set(api_tokens::last_used_at.eq(now.into_sql::<Timestamptz>().nullable()))
```
Turbo87, line 80: this is unfortunately necessary due to a quirk in diesel. see https://github.com/diesel-rs/diesel/issues/1514#issuecomment-613340122.

migrations/2025-02-17-133723_timezones/up.sql
```diff
@@ -0,0 +1,45 @@
+SET timezone = 'UTC';
```
Turbo87, line 1: this part is important since Postgres would otherwise perform a full table rewrite. if we set the timezone to `UTC`, Postgres can take a shortcut and doesn't have to rewrite any data.

eth3lbert: Yeah, this is interesting, which I also wanted to point out :)
Here is a SO link[^1] for reference that provides more details about this.

[^1]: https://dba.stackexchange.com/questions/322904/alter-timestamp-column-to-timestamptz-without-converting-data/322907#322907

eth3lbert (approved):
I'm good with this, especially since it won't cause a full table rewrite :D

migrations/2025-02-17-133723_timezones/up.sql
Turbo87: migration started on staging at 2025-02-18T08:28:22.618774Z 
migration ended on staging at 2025-02-18T08:28:22.653333Z

35ms seems reasonable IMHO :)

PR #10610, opened 2025-02-18
WIP sekrit project

carols10cents:


crates/crates_io_database/src/models/user.rs
```diff
@@ -122,3 +122,78 @@ impl NewUser<'_> {
...
+#[diesel(
+    table_name = linked_accounts,
+    check_for_backend(diesel::pg::Pg),
+    primary_key(provider, account_id),
+    belongs_to(User),
+)]
+pub struct LinkedAccount {
+    pub user_id: i32,
+    pub provider: AccountProvider,
+    pub account_id: i32, // corresponds to user.gh_id
```
blm768, line 176: This could very well be a problem to deal with later,  but I could imagine having a provider that represents its IDs some other way, e.g. strings or UUIDs. We probably don't want to commit to such a design yet,  but perhaps we'll eventually need some kind of "table-per-provider" approach.

script/backfill-linked-accounts.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO linked_accounts (user_id, provider, account_id, access_token, login, avatar)
```
blm768, line 1: We'll likely want a benchmark on this to see if it takes long enough that it needs to be broken into smaller batches, but my instinct says this approach is _probably_ fine.

migrations/2025-02-19-013433_add_username_to_user/up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE users
+-- The column needs to be nullable for this migration to be fast; can be changed to non-nullable
```
blm768, line 2: Technically, we probably could get away with making this non-null with a default of `''`, but Postgres can handle changes in nullability without a table rewrite, so I don't know that it would really matter in practice.

migrations/2025-02-19-013433_add_username_to_user/up.sql
```diff
@@ -0,0 +1,6 @@
...
+-- after backfill of all records.
+ADD COLUMN username VARCHAR;
+
+CREATE INDEX lower_username ON users (lower(username));
```
blm768, line 6: This might need a benchmark, but I suspect we'd save some work by creating this after the column has been populated.

crates/crates_io_database/src/models/user.rs
carols10cents: That's a really good point. Hmm. I'm trying to think through whether I'd rather start with an `oauth_github` table and then add, say, `oauth_gitlab`, etc tables, or if i'd rather start with what's here and then need to migrate to a table-per-provider later....

script/backfill-linked-accounts.sql
carols10cents: Yeah, I'm starting to think I want a more general sort of harness for running SQL in a loop of batches of N until it returns 0 results, because a few of these migrations are going to need to do that sort of thing...

carols10cents:
Thank you for the review @blm768 !

crates/crates_io_database/src/models/user.rs
blm768: Unfortunately, it'll probably be hard to know what we need until we need it. The good news is that we shouldn't be boxing ourselves into a hole that's impossible to escape regardless of which approach we choose.

blm768:
Glad to help! Most of my experience with large database migrations is with MySQL, not PostgreSQL, so I very well could have missed some things, but as far as I can tell, most of what I learned with MySQL seems to port over well enough.

carols10cents:
Closing this to be redone slower.

PR #10621, opened 2025-02-18
ESLint: Migrate to flat config format

Turbo87:
see https://eslint.org/docs/latest/use/configure/migration-guide


Turbo87:
This should hopefully unblock https://github.com/rust-lang/crates.io/pull/10609

PR #10659, opened 2025-02-22
Migrate versions page to paginated list

eth3lbert:
This PR migrates the `versions` page to a paginated style, which loads more versions with a "Load More" button. Two sorted lists serve as different contexts, so they can display in different. Most of the computation for versions (`sorting` and `release_tracks`) on this page will no longer be needed on the app, they all done on the backend.

Some major changes are highlighted as follows:

- Drop support for first version hint (breaking change).
- Migrate the `versions` page to a paginated list.
- Simplify `releaseTrackSet` with `release_tracks` meta.

Regarding the first version hint, since we've moved to a paginated list and the API doesn't support reverse ordering, determining the first version would likely require fetching all versions. If we choose to support this rather than drop it, it would be nice to store this information in the `default_versions` table during publishing and expose it on the API, similar to #10519.

For `release_tracks`, since we now only fetch it once if it doesn't already exist, and paginated related data only resets when the controllers `crate` changes (or page reload), `release_tracks` could become outdated if it changes after the initial fetch. Two operations can affect the result of `release_tracks`:

- The highest stable version being yanked
- A new stable version being released (potential)

In the yanked scenario, a simple solution would be to refetch `release_tracks` upon detection within the app. For new releases, the update currently only occurs upon page reload.

Another possible and potentially better solution would be to, again, upsert `release_tracks` during publishing/yanking, and expose it on the API like #10519. This would be costly (offline) during backfill, as it would require calculation from scratch, but should be inexpensive during updates, as it would only need to compare to a single version. While this would complicate the publish endpoint, it would result in calculation only once, rather than request.

Related:
- #9624 
- #10248

### preview
paginated_versions

### DIFF
--- a/app/controllers/crate/versions.js
+++ b/app/controllers/crate/versions.js
@@ -14,7 +14,7 @@ export default class SearchController extends Controller {
   }
 
   get sortedVersions() {
-    let { versionIdsBySemver, versionIdsByDate, versionsObj: versions } = this.crate;
+    let { versionIdsBySemver, versionIdsByDate, loadedVersionsById: versions } = this.crate;
 
     return (this.sort === 'semver' ? versionIdsBySemver : versionIdsByDate).map(id => versions.get(id));
   }
--- a/app/controllers/crate/versions.js
+++ b/app/controllers/crate/versions.js
@@ -52,6 +52,7 @@ export default class SearchController extends Controller {
     if (next_page) {
       let params = new URLSearchParams(next_page);
       params.set('name', crate.name);
+      params.delete('include');
       query = Object.fromEntries(params.entries());
     } else {
       if (sort !== 'semver') {
@@ -62,6 +63,9 @@ export default class SearchController extends Controller {
         sort,
         per_page,
       };
+    }
+    if (crate.release_tracks == null) {
+      query.include = 'release_tracks';
     }
 
     try {
@@ -82,6 +86,18 @@ export default class SearchController extends Controller {
           next_page: meta.next_page,
         };
       }
+
+      // set release_tracks to crate
+      if (meta.release_tracks) {
+        let payload = {
+          crate: {
+            id: crate.id,
+            release_tracks: meta.release_tracks,
+          },
+        };
+        this.store.pushPayload(payload);
+      }
+
       return versions;
     } catch (error) {
       // report unexpected errors to Sentry and ignore `ajax()` errors
--- a/app/controllers/crate/versions.js
+++ b/app/controllers/crate/versions.js
@@ -16,6 +16,6 @@ export default class SearchController extends Controller {
   get sortedVersions() {
     let { versionIdsBySemver, versionIdsByDate, versionsObj: versions } = this.crate;
 
-    return (this.sort === 'semver' ? versionIdsBySemver : versionIdsByDate).map(id => versions[id]);
+    return (this.sort === 'semver' ? versionIdsBySemver : versionIdsByDate).map(id => versions.get(id));
   }
 }
--- a/app/controllers/crate/versions.js
+++ b/app/controllers/crate/versions.js
@@ -1,21 +1,88 @@
 import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 
+import { dropTask } from 'ember-concurrency';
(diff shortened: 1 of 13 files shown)


tests/models/version-test.js
```diff
@@ -231,6 +236,7 @@ module('Model | Version', function (hooks) {
       this.db.version.create({ crate, num: '0.4.3', yanked: true });
       crateRecord = await this.store.findRecord('crate', crate.name, { reload: true });
       versions = (await crateRecord.loadVersionsTask.perform({ reload: true })).slice();
+      crateRecord.release_tracks = calculateReleaseTracks(versions);
```
eth3lbert, line 239: This change was made to pass the tests, but it doesn't reflect what we do in practice (for versions page). Since the new release wouldn't be loaded without a page reload, as already mentioned in the description, I'm not sure what the best approach is in this case.

app/templates/crate/versions.hbs
```diff
@@ -1,10 +1,10 @@
-<CrateHeader @crate={{this.model}} />
+<CrateHeader @crate={{this.crate}} />
 
 <div local-class="results-meta">
   <span local-class="page-description" data-test-page-description>
-    All <strong>{{ this.model.num_versions }}</strong>
-    versions of <strong>{{ this.model.name }}</strong> since
-    {{date-format this.model.created_at 'PPP'}}
+    <strong>{{ this.sortedVersions.length }}</strong> of <strong>{{ this.crate.num_versions }}</strong>
+    <strong>{{ this.crate.name }}</strong> versions since
```
eth3lbert, line 6: This currently seems a bit odd, since the bolded number and name are positioned directly beside each other. There might be better wording or display methods.

app/controllers/crate/versions.js
```diff
@@ -1,21 +1,88 @@
...
+  return { data: [], next_page: undefined };
+}
+
 export default class SearchController extends Controller {
-  queryParams = ['sort'];
+  @service store;
 
+  queryParams = ['per_page', 'sort'];
   @tracked sort;
+  @tracked per_page = 10;
```
Turbo87, line 20: I think we should increase this a little bit. I don't think we need to display 1000 versions, but 100 should probably be fine? That way we don't show the pagination for the majority of crates.

app/templates/crate/versions.hbs
```diff
@@ -23,3 +23,15 @@
...
 </ul>
+{{#if this.next_page}}
+  <div local-class="load-more">
+    <button type="button" data-test-id="load-more" disabled={{this.loadMoreTask.isRunning}}
+      {{on "click" (perform this.loadMoreTask)}}
+    >
+      Load More
+      {{#if this.loadMoreTask.isRunning}}
+        <LoadingSpinner />
+      {{/if}}
```
Turbo87, line 34: it looks like the spinner isn't vertically centered correctly. it looks a bit off.

Turbo87 (commented):
tested it out locally and looks like it works as intended :)

app/controllers/crate/versions.js
eth3lbert: fair enough!

app/templates/crate/versions.hbs
eth3lbert: Thanks, I'll take a look!

tests/acceptance/crate-test.js
```diff
@@ -130,13 +131,25 @@ module('Acceptance | crate page', function (hooks) {
...
+
+  test('navigating to the versions page with custom per_page', async function (assert) {
+    loadFixtures(this.db);
+
+    await visit('/crates/nanomsg/versions?per_page=10');
+
+    assert.dom('[data-test-page-description]').hasText(/10 of 13\s+nanomsg versions since\s+December \d+th, 2014/);
+
+    await click('[data-test-id="load-more"]');
+    assert.dom('[data-test-page-description]').hasText(/13 of 13\s+nanomsg versions since\s+December \d+th, 2014/);
```
eth3lbert, line 152: Since our `per_page` (100) is much bigger than the test records (13), I split it into two test cases. This can effectively test if 'load more' works.

eth3lbert:
Rebased and feedback addressed!

Turbo87 (approved):
alright, let's give this a try

PR #10666, opened 2025-02-23
routes/crate: Fetch crate when peeked crate lacks included values

### LINKED ISSUE #10663: Keywords are sometimes missing
### Current Behavior

When I search for a crate, e.g. `enigo` and click on its name, the description of the crate and other metadata gets displayed. However the keywords are missing. If I open the crate details in a new tab instead, the keywords get displayed.

Missing keywords:
Image

Correctly displayed keywords:

Image

### Expected Behavior

The keywords of a crate should always get displayed, no matter if it gets opened in a new tab or in the current tab

### Steps To Reproduce

- Open crates.io
- Search for "enigo" (anything will do)
- Right click on the result and open in a new tab
- Left click on the result and open it in the current tab
- Compare the content of the tabs

### Environment

- Browser: Librewolf 135.0.1-1
- OS: Arch Linux

### Anything else?

When the keywords are missing, the "Toggle Reader view" is available. If the keywords are present, the toggle is missing. That's not important to me, but I thought that's curious too

eth3lbert:
Fixes #10663 .


e2e/acceptance/crate.spec.ts
```diff
@@ -245,4 +245,14 @@ test.describe('Acceptance | crate page', { tag: '@acceptance' }, () => {
...
+
+  test('keywords are shown when navigating from search', async ({ page, msw }) => {
+    loadFixtures(msw.db);
+
+    await page.goto('/search?q=nanomsg');
+    await page.getByRole('link', { name: 'nanomsg', exact: true }).click();
+
+    await expect(page).toHaveURL('/crates/nanomsg');
+    await expect(page.locator('[data-test-keyword]')).toBeVisible();
+  });
```
eth3lbert, line 257: There's another case that would cause the same issue: navigating from dependencies that have more than one dependency (as this will query with `?ids=`). We would need to expand the fixtures to have at least two dependencies. Let me know if we'd like to include this as well.

eth3lbert:
I re-ran the failed tests, and those failures seem unrelated to this PR.

e2e/acceptance/crate.spec.ts
Turbo87: probably fine as-is. if anything, IMHO we shouldn't extend the hardcoded fixtures and instead setup a dedicated scenario for the test. I would prefer to get rid of the fixtures in general.

Turbo87 (approved):

PR #10721, opened 2025-02-28
serializers/crate: Prevent `categories` and `keywords` from being cleared

### LINKED ISSUE #10711: Tags are not displayed when navigating from Crate -> Tag -> Same Crate
### Current Behavior

If you click a tag on a crate on crates.io it brings you to a list of crates with that tag. If you then navigate back to the same crate that brought you to the list, its tags are no longer displayed.

### Expected Behavior

The list of tags should be repopulated.

### Steps To Reproduce

Using the back button: 
1. Open https://crates.io/crates/trunk
2. Click on "#build-tool" (opens https://crates.io/keywords/build-tool)
3. Click "back" on your browser
4. The crate's tag list is now blank

---
Navigating via crates:
1. Open https://crates.io/crates/trunk
2. Click on "#build-tool" (opens https://crates.io/keywords/build-tool)
3. Click on "trunk" at the top of the list of crates (returns to https://crates.io/crates/trunk)
4. The crate's tag list is now blank.
 
This happens even if you navigate through multiple pages of results to find the original crate (ex - trunk is on page 3 of "#compilers").

### Environment

- Browser: Firefox 135, Edge 133
- OS: Windows 11
---
- Browser: Firefox 135
- OS: Android 13

### Anything else?

Image
Image

eth3lbert:
Previously, if `categories` and `keywords` of `crate` had existing values, they would be cleared from query results if those value returned null. This caused issues when `crate` was peeked from the store. This commit detects null values and ignores the fields instead of updating them.

Fixes #10711 .

### DIFF
--- a/app/serializers/crate.js
+++ b/app/serializers/crate.js
@@ -1,5 +1,7 @@
 import ApplicationSerializer from './application';
 
+const SKIP_NULL_FIELDS = new Set(['categories', 'keywords']);
+
 export default class CrateSerializer extends ApplicationSerializer {
   isNewSerializerAPI = true;
 
@@ -9,5 +11,21 @@ export default class CrateSerializer extends ApplicationSerializer {
     }
 
     return super.extractRelationships(...arguments);
+  }
+
+  normalizeQueryResponse(_store, _modelClass, payload) {
+    // We don't want existing relationships overwritten by results with null values.
+    // See: https://github.com/rust-lang/crates.io/issues/10711
+    if (payload.crates) {
+      payload.crates = payload.crates.map(crate => {
+        for (const rel of SKIP_NULL_FIELDS) {
+          if (crate[rel] === null) {
+            delete crate[rel];
+          }
+        }
+        return crate;
+      });
+    }
+    return super.normalizeQueryResponse(...arguments);
   }
 }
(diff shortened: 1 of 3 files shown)


app/serializers/crate.js
```diff
@@ -10,4 +12,23 @@ export default class CrateSerializer extends ApplicationSerializer {
 
     return super.extractRelationships(...arguments);
   }
+
+  normalizeQueryResponse(_store, _modelClass, payload) {
+    // We don't want existing relationships overwritten by results with null values.
+    // See: https://github.com/rust-lang/crates.io/issues/10711
+    if (payload.crates) {
+      payload = {
+        ...payload,
```
Turbo87, line 21: is that assignment doing anything? seems like the `payload` variable isn't being used afterwards

app/serializers/crate.js
```diff
@@ -10,4 +12,23 @@ export default class CrateSerializer extends ApplicationSerializer {
...
+        crates: payload.crates.map(crate => {
+          for (const rel of SKIP_NULL_FIELDS) {
+            if (crate[rel] == null) {
```
Turbo87, line 24:
```suggestion
            if (crate[rel] === null) {
```

I guess we should be explicit here?

eth3lbert: yeah, fair enough!

app/serializers/crate.js
eth3lbert: yeah, `payload.crates` should be enough!

eth3lbert:
those failures in CI seem unrelated to this PR.

PR #10740, opened 2025-03-03
versions: Improve the loading indicator

eth3lbert:
An excerpt from https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/swc/near/503007257
>I noticed that while loading the versions list there is no loading indicator and it looks like the crate simply has no versions. might be worth implementing some kind of loading screen for that state.

This shows a loading indicator for the initial load, and a loading button when loading more :D

https://github.com/user-attachments/assets/7386c5ae-99dd-4939-8d4f-220fb16e753a

### DIFF
--- a/app/templates/crate/versions.hbs
+++ b/app/templates/crate/versions.hbs
@@ -1,7 +1,10 @@
 <CrateHeader @crate={{this.crate}} />
 
 <div local-class="results-meta">
-  <span local-class="page-description" data-test-page-description>
+  <span local-class="page-description"
+    class="{{if (and this.loadMoreTask.isRunning (not this.sortedVersions)) 'is-empty'}}"
+    data-test-page-description
+  >
     <strong>{{ this.sortedVersions.length }}</strong> of <strong>{{ this.crate.num_versions }}</strong>
     <strong>{{ this.crate.name }}</strong> versions since
     {{date-format this.crate.created_at 'PPP'}}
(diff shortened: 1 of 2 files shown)


eth3lbert:
I need to fix those flaky CI issues that aren't related to this PR. It's quite annoying :/

app/templates/crate/versions.hbs
```diff
@@ -16,22 +16,27 @@
   </div>
 </div>
 
-<ul local-class="list">
+<ul local-class="list" class="{{unless this.sortedVersions 'is-empty'}}">
```
Turbo87, line 19: what is the purpose of the `'is-empty'` here?

app/templates/crate/versions.hbs
```diff
@@ -16,22 +16,27 @@
...
   {{#each this.sortedVersions as |version|}}
     <li>
       <VersionList::Row @version={{version}} local-class="row" data-test-version={{version.num}} />
     </li>
   {{/each}}
 </ul>
-{{#if this.next_page}}
+{{#if this.loadMoreTask.isRunning}}
```
Turbo87, line 26: should we also hide the `.page-description` element above while `this.loadMoreTask.isRunning && this.sortedVersions.count === 0`?

app/templates/crate/versions.hbs
eth3lbert: I use this to determine if it's the initial load. Therefore, when there are empty versions (indicating loading is still in progress), it provides a margin and displays a different loading style (text style) compared to when versions are already present (button style).

Turbo87: ohhhh, that is an inline `unless`. nevermind 😅

app/templates/crate/versions.hbs
eth3lbert: So, this is regarding the initial load (like the 'is-empty' check mentioned above)? I presume this could be achieved via style="{{unless this.sortedVersions 'visibility:hidden'}}" (which could also be implemented using a CSS class and style).

I don't really have a preference on this, but I think the reasoning behind it is to make it more noticeable that it's still loading and is an initial load. So, I'm okay with this.

I'm also wondering if we should prevent the top loading indicator (the yellow one) from completing until the first versions have finished loading.

eth3lbert: > should we also hide the `.page-description` element above while `this.loadMoreTask.isRunning && this.sortedVersions.count === 0`?

done!

Turbo87 (approved):
looks good, thanks 👍

PR #10743, opened 2025-03-03
Fix flaky tests

eth3lbert:
Attempting to fix the following flaky tests:

https://github.com/rust-lang/crates.io/actions/runs/13632195641/job/38102274110?pr=10740#step:6:239
```
not ok 156 Chrome 133.0 - [147 ms] - Exam Partition 1 - Acceptance | support > reporting a crate from crate page: valid form without detail
    ---
        stack: >
            Error: Element not found when calling `click('[data-test-id="link-crate-report"]')`.
                at eval (webpack://crates-io/../../.pnpm/@ember+test-helpers@5.1.0_@babel+core@7.26.9_ember-source@6.0.1_@glimmer+component@2.0.0_rsvp@4.8.5_webpack@5.98.0_/node_modules/@ember/test-helpers/dist/dom/click.js?:107:13)
                at async prepare (webpack://crates-io/./tests/acceptance/support-test.js?:3035:7)
                at async Object.eval (webpack://crates-io/./tests/acceptance/support-test.js?:3091:7)
        message: >
            Promise rejected during "valid form without detail": Element not found when calling `click('[data-test-id="link-crate-report"]')`.
        negative: >
            false
        browser log: |
    ...
```
and
https://github.com/rust-lang/crates.io/actions/runs/13632195641/job/38102274110?pr=10740#step:6:254
```
not ok 159 Chrome 133.0 - [90 ms] - Exam Partition 1 - Acceptance | support > reporting a crate from crate page: valid form with required detail
    ---
        stack: >
            Error: Element not found when calling `click('[data-test-id="link-crate-report"]')`.
                at eval (webpack://crates-io/../../.pnpm/@ember+test-helpers@5.1.0_@babel+core@7.26.9_ember-source@6.0.1_@glimmer+component@2.0.0_rsvp@4.8.5_webpack@5.98.0_/node_modules/@ember/test-helpers/dist/dom/click.js?:107:13)
                at async prepare (webpack://crates-io/./tests/acceptance/support-test.js?:3035:7)
                at async Object.eval (webpack://crates-io/./tests/acceptance/support-test.js?:3132:7)
        message: >
            Promise rejected during "valid form with required detail": Element not found when calling `click('[data-test-id="link-crate-report"]')`.
        negative: >
            false
        browser log: |
    ...
```

### DIFF
--- a/tests/acceptance/support-test.js
+++ b/tests/acceptance/support-test.js
@@ -1,4 +1,4 @@
-import { click, currentURL, fillIn, findAll } from '@ember/test-helpers';
+import { click, currentURL, fillIn, findAll, waitFor } from '@ember/test-helpers';
 import { module, test } from 'qunit';
 
 import percySnapshot from '@percy/ember';
@@ -205,6 +205,8 @@ test detail
       };
 
       await visit('/crates/nanomsg');
+      assert.strictEqual(currentURL(), '/crates/nanomsg');
+      await waitFor('[data-test-id="link-crate-report"]');
       await click('[data-test-id="link-crate-report"]');
       assert.strictEqual(currentURL(), '/support?crate=nanomsg&inquire=crate-violation');
       assert.dom('[data-test-id="crate-input"]').hasValue('nanomsg');


tests/acceptance/support-test.js
```diff
@@ -205,6 +205,8 @@ test detail
       };
 
       await visit('/crates/nanomsg');
+      assert.strictEqual(currentURL(), '/crates/nanomsg');
+      assert.dom('[data-test-id="crate-sidebar"]').exists();
```
Turbo87, line 209: how would this fix it?

eth3lbert: Ah, wait, sorry. I should have used `waitFor` rather than `assert.dom` 🤦.

eth3lbert: I just changed it to wait for the button to exist.

Turbo87: > I should have used `waitFor` rather than `assert.dom` 🤦.

I guess that should work, though I'm wondering why it is needed here. iirc the `link-crate-report` button should be rendered directly and there are no network requests or other async things that would prevent that. feels like something else might be wrong.

ultimately, the assertion architecture of playwright where `waitFor()` is the default is probably better than the test waiter architecture of Ember.js 😅

eth3lbert: > irc the `link-crate-report` button should be rendered directly and there are no network requests or other async things that would prevent that. . feels like something else might be wrong.

Yeah, I have no clue what's going on here either! It only depends on the crate (@crate.name), which should be resolved in the routes, but it's failing somehow for no obvious reason. This led me to wait for it explicitly as a naive solution. I should investigate the root cause in more detail in the future.



> ultimately, the assertion architecture of playwright where `waitFor()` is the default is probably better than the test waiter architecture of Ember.js 😅

Indeed, the auto-retry is much more ergonomic.

Turbo87 (approved):
I guess we can merge this for now until we find a better solution (or migrate the whole codebase to use `waitFor()` instead :P)

Turbo87:
> Error: waitFor timed out waiting for selector "[data-test-id="link-crate-report"]"

https://github.com/rust-lang/crates.io/actions/runs/13762342974/job/38481079622?pr=10794

eth3lbert:
> > Error: waitFor timed out waiting for selector "[data-test-id="link-crate-report"]"
> 
> https://github.com/rust-lang/crates.io/actions/runs/13762342974/job/38481079622?pr=10794

I'm not sure what the issue is here. It would be great if we could take a screenshot when it fails to find the DOM.

PR #10755, opened 2025-03-05
VersionList::Row: Limit the length of the features list

Turbo87:
as discussed in the team meeting:

Bildschirmfoto 2025-03-05 um 09 52 47

Related:

- https://github.com/rust-lang/crates.io/pull/10540


LawnGnome (approved):
LGTM. I think we should do this rather than having a more heavyweight UI, as we discussed.

PR #10763, opened 2025-03-06
Add `semver_ord(num)` SQL function and `version.semver_ord` column

Turbo87:
We currently rely on the Rust `semver` crate to implement our "sort by semantic versioning" functionality, that is used by web interface, but also to determine the "default version". This has the downside that we need to load the full list of version numbers for a crate from the database to the API server, sort it and then throw away the ones that we don't need.

Ideally, we would be using https://pgxn.org/dist/semver/ or https://github.com/pgcentralfoundation/pgrx to solve this, but unfortunately most PostgreSQL hosters don't allow/support custom extensions, and it would also make local development a bit more challenging.

As a workaround, this PR implements a `semver_ord(num)` pgSQL function that returns a JSONB array, which has the same ordering precedence as the Semantic Versioning spec (https://semver.org/#spec-item-11), with the small caveat that it only supports up to 15 prerelease parts. The maximum number of prerelease parts in our current dataset is 7, so 15 should be plenty.

> **Update:** After discussion in this PR we changed the maximum number of prerelease specifiers to 10, but the remaining specifiers are now appended to the array as a string, to at least ensure uniqueness and a stable sort order.

The database migration in this commit also adds a new `semver_ord` column to the `versions` table, and an on-insert trigger function that automatically derives the `semver_ord` column from the `num` column value:

Bildschirmfoto 2025-03-06 um 08 46 02

Once this migration has run, the existing versions can be backfilled by running the following SQL script, until all versions are processed:

```sql
with versions_to_update as (
    select id, num
    from versions
    where semver_ord = 'null'::jsonb
    limit 1000
)
update versions
    set semver_ord = semver_ord(num)
    where id in (select id from versions_to_update);
```

This PR does not yet implement any code to actually use this new column, since it will need to be backfilled first. Once that has happened and verified to produce correct results, we can start to migrate our codebase to move the semver sorting into the database and potentially even calculate the "default version" directly in the database too.


migrations/2025-03-06-060640_semver_ord/up.sql
eth3lbert, line 1: I'm wondering, rather than ignoring them, would it be better if we stored everything beyond 15 parts as a single part? This way, while it's still not perfectly accurate, the beyond part would potentially be sorted lexically, which should be much better than ignoring it?

--

And maybe we could also log those with more than 15 part versions, then we could revisit whether to raise the limit or ignore them.

Turbo87: > would it be better if we stored everything beyond 15 parts as a single part? This way, while it's still not perfectly accurate, the beyond part would potentially be sorted lexically, which should be much better than ignoring it?

yeah, that seems like a good idea. in that case I guess we could even reduce the number of prerelease parts to 10 or even 5. 🤔

eth3lbert: sounds great 👍

eth3lbert (approved):
Thank you for this excellent PR. This will significantly simplify the work involving version sorting. 💪 💪 💪 
 I've left some non-blocking nitpicks, but everything is currently great! 👍

crates/crates_io_database/tests/semver_ord.rs
```diff
@@ -0,0 +1,100 @@
...
+        1.0.0-alpha,
+        1.0.0-alpha.1,
+        1.0.0-alpha.beta,
+        1.0.0-beta,
+        1.0.0-beta.2,
+        1.0.0-beta.11,
+        1.0.0-rc.1,
+        1.0.0,
+    ]
+    ");
```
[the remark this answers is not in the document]
Turbo87: done ✅

Turbo87:
@Gankra you know a lot about semver edge cases. do you want to try and break this before we actually start using the new column? :D

Gankra:
ahahahaha

I *think* as long as it handles 
https://crates.io/crates/cursed-trying-to-break-cargo/
ok there's not much else for me to throw at it?

Gankra:
Also dang I know that page induces NaNs (improved from "bootloops crates.io's client") but what a chaotic sort haha.

Turbo87:
> Also dang I know that page induces NaNs (improved from "bootloops crates.io's client") but what a chaotic sort haha.

are you sure you were in semver sorting mode? https://crates.io/crates/cursed-trying-to-break-cargo/versions?sort=semver looks okayish to me 😅

Gankra:
Oh you're totally right, it's so rare to see date and semver not be 1:1 for crates I forgot date is the default :)

crates/crates_io_database/tests/semver_ord.rs
```diff
@@ -0,0 +1,118 @@
...
+            .await
+            .unwrap()
+            .output
+            .unwrap_or_default()
+    };
+
+    insta::assert_snapshot!(check("0.0.0").await, @r#"[0, 0, 0, {}]"#);
+    insta::assert_snapshot!(check("1.0.0-alpha.1").await, @r#"[1, 0, 0, [true, "alpha", false, 1, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, ""]]"#);
+
+    // see https://crates.io/crates/cursed-trying-to-break-cargo/1.0.0-0.HDTV-BluRay.1020p.YTSUB.L33TRip.mkv – thanks @Gankra!
```
LawnGnome, line 32: I love that this is represented as-is. :laughing:

migrations/2025-03-06-060640_semver_ord/up.sql
```diff
@@ -0,0 +1,109 @@
...
+    prerelease_part text;
+    i int := 0;
+begin
+    if match_result is null then
+        return null;
+    end if;
+
+    if match_result[4] is null then
+        -- A JSONB object has higher precedence than an array, and versions with
+        -- prerelease specifiers should have lower precedence than those without.
```
LawnGnome, line 27: I went to check that we're not relying on any undocumented behaviour, and found [this gem](https://www.postgresql.org/docs/17/datatype-json.html#JSON-INDEXING):

> The btree ordering for jsonb datums is seldom of great interest[...]

Little did they know!

LawnGnome (approved):
My PL/pgSQL is very rusty, but this LGTM. :+1:

migrations/2025-03-06-060640_semver_ord/up.sql
Turbo87: yep, see https://github.com/rust-lang/crates.io/pull/10763#discussion_r1983818620 :)

migrations/2025-03-06-060640_semver_ord/up.sql
```diff
@@ -0,0 +1,109 @@
...
+
+
+-- Populate the `semver_ord` column for existing versions.
+-- This query should be run manually in small batches to avoid locking the
+-- table for too long.
+
+-- with versions_to_update as (
+--     select id, num
+--     from versions
+--     where semver_ord = 'null'::jsonb
```
Turbo87, line 104: I just noticed that I didn't update this correctly. this line should now be `where semver_ord is null`.

PR #10780, opened 2025-03-07
Exclude deleted crate owners from admin tool output

carols10cents:
I just tried to use the admin tool to delete a crate and it LIED TO MY FACE and included the username of a DELETED OWNER in the output. Uuuuunaaaaaaccceptable


Turbo87 (approved):
👍

PR #10792, opened 2025-03-10
app/routes/crate/settings: hide the settings page for non-owners

LawnGnome:
@technetos pinged me via DM on this just in case it was a security issue. It wasn't, since the backend controllers all have appropriate permission checks, but there's no reason for us to even show this page to someone who doesn't own a crate.

I implemented some quick regression tests for the authentication side of the frontend controller based on the delete crate tests, but not a full set of tests for the page. It's Sunday night.

### DIFF
--- a/app/routes/crate/settings.js
+++ b/app/routes/crate/settings.js
@@ -1,6 +1,23 @@
+import { inject as service } from '@ember/service';
+
 import AuthenticatedRoute from '../-authenticated-route';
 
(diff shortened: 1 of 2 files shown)


Turbo87 (approved):
I've been aware of this for a while but figured that since the frontend doesn't actually know whether the user is a team owner, it's fine. But then again, it doesn't make much sense to use different logic for showing the tab and allowing the user to visit the page.

tl;dr LGTM :D

PR #10793, opened 2025-03-10
Add backward seek-based pagination support for the search endpoint

### LINKED ISSUE #10349: `prev_page` field is always `null` on search
### Current Behavior

`prev_page` is always `null`.

### Expected Behavior

It should return the `seek` for the previous page if there is one.

### Steps To Reproduce

You can see the issue on docs.rs:

Go to https://docs.rs/releases/search?query=sysinfo
Click "next page"

Then you can see that there is no "previous page" button. I originally suspected a bug on docs.rs side but after investigating, it looks like the `["api", "v1", "crates"]` query always return a `null` `prev_page` in the JSON response.

### Environment

_No response_

### Anything else?

_No response_

eth3lbert:
Previously, we were returning `null` for `prev_page` in search endpoint. With this PR, we now provide the URL for `prev_page`. For other endpoints to support backward seeking, they must have the page options with seek backward enabled (`.enable_seek_backward(true)`). Otherwise, without this, any seek param starting with `-` (indicating backward seeking) will result in an error!

It would be easier to review this PR commit by commit 😉 

Resolves #10349.


src/tests/routes/crates/list.rs
eth3lbert, line 1: This is currently only tested with a page size of 1. It would be better to expand the tests to include varying page sizes.

Turbo87 (commented):
I'm still wondering if all the additional complexity is really worth it... 😅 

I'm curious what the rest of the @rust-lang/crates-io team thinks about this.

LawnGnome:
Sorry, I skimmed this yesterday, but didn't have time to give it a full review. I still haven't, but my rough thought process was basically "oh, this is way better than I would have thought", then I realised that GitHub had collapsed the `krate/search.rs` changes, and I didn't feel that way so much after that.

That said, the changes are mostly mechanical, and feel like we could eventually abstract at least some of them behind a macro or two. Against that, `krate/search.rs` is _already_ terrifying, and adding two new sets of branches for each sort option doesn't spark joy.

So, I guess I wouldn't block this, but I'm not super enthusiastic about it, either. Do we have a strong use case for this? (Maybe more a question for @GuillaumeGomez.) I'm probably -0 right now, but could be convinced, particularly if we came up with a way to abstract some of the details out of the controllers proper.

eth3lbert:
Thanks for the feedback! 

> Sorry, I skimmed this yesterday, but didn't have time to give it a full review. I still haven't, but my rough thought process was basically "oh, this is way better than I would have thought", then I realised that GitHub had collapsed the `krate/search.rs` changes, and I didn't feel that way so much after that.
> 
> That said, the changes are mostly mechanical, and feel like we could eventually abstract at least some of them behind a macro or two. Against that, `krate/search.rs` is _already_ terrifying, and adding two new sets of branches for each sort option doesn't spark joy.

Cursor-based pagination does indeed complicate things, especially with `NULLS LAST` ordering and ordering with a "tie-breaker". Abstraction should be doable. Given the columns and their ordering, we should be able to build the filter query. However, that would require more work and potentially become more complex when more tables or even aliases are involved, and it could also be more difficult to implement with diesel. Ultimately, this would result in something similar to SeaORM's general cursor-based pagination solution, which should be great :thinking:.

> So, I guess I wouldn't block this, but I'm not super enthusiastic about it, either. Do we have a strong use case for this? (Maybe more a question for @GuillaumeGomez.) I'm probably -0 right now, but could be convinced, particularly if we came up with a way to abstract some of the details out of the controllers proper.

This currently serves mainly for docs.rs, and I'm not personally pressing for it :D

GuillaumeGomez:
Well, the navigation experience is suboptimal, so that would be very welcome. However if you think that a better (technical) solution could be achieved, then we can wait for it.

PR #10819, opened 2025-03-13
Use `semver_ord` column for sorting versions

Turbo87:
This PR changes how we return sorted versions from the `GET /api/v1/crates/{name}/versions` API endpoint. 

Previously, we needed to download all version numbers from the database to the API server and perform the sorting and filtering there, since the database did not support sorting by semantic version. With https://github.com/rust-lang/crates.io/pull/10763 that changed, so now we can perform the sorting and filtering directly in the database, which simplifies the code that the API server needs to run and the data that needs to be transferred between the API server and the database.

Note that I have not added any database indexes for the `semver_ord` column yet, so there is potentially some more room for performance improvements.

The implementation as-is has a small caveat, in that it is a breaking change to the `seek` query parameter format. We could potentially build a compatibility layer to make old `next_page` links still work with the new implementation, but since seek pagination for this endpoint is rarely used outside of our own frontend, I'm not sure whether it's really worth it for the short transition period between the two implementations.

Related:

- https://github.com/rust-lang/crates.io/pull/10763

### DIFF
--- a/src/controllers/krate/versions.rs
+++ b/src/controllers/krate/versions.rs
@@ -99,13 +99,8 @@ pub async fn list_versions(
         None => None,
     };
 
-    // Sort by semver by default
-    let versions_and_publishers = match &params.sort.as_ref().map(|s| s.to_lowercase()).as_deref() {
-        Some("date") => {
-            list_by_date(crate_id, pagination.as_ref(), &params, &req, &mut conn).await?
-        }
-        _ => list_by_semver(crate_id, pagination.as_ref(), &params, &req, &mut conn).await?,
-    };
+    let versions_and_publishers =
+        list(crate_id, pagination.as_ref(), &params, &req, &mut conn).await?;
 
     let versions = versions_and_publishers
         .data
@@ -126,12 +121,12 @@ pub async fn list_versions(
     }))
 }
 
-/// Seek-based pagination of versions by date
+/// Seek-based pagination of versions
 ///
 /// # Panics
 ///
-/// This function will panic if `option` is built with `enable_pages` set to true.
-async fn list_by_date(
+/// This function will panic if `options` is built with `enable_pages` set to true.
+async fn list(
     crate_id: i32,
     options: Option<&PaginationOptions>,
     params: &ListQueryParams,
@@ -139,6 +134,11 @@ async fn list_by_date(
     conn: &mut AsyncPgConnection,
 ) -> AppResult<PaginatedVersionsAndPublishers> {
     use seek::*;
+
+    let seek = match &params.sort.as_ref().map(|s| s.to_lowercase()).as_deref() {
+        Some("date") => Seek::Date,
+        _ => Seek::Semver,
+    };
 
     let make_base_query = || {
         let mut query = versions::table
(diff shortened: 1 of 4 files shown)


src/tests/routes/crates/versions/list.rs
```diff
@@ -291,9 +291,9 @@ async fn test_seek_based_pagination_semver_sorting() -> anyhow::Result<()> {
     assert_eq!(json.meta.total as usize, expects.len());
     assert_eq!(json.meta.release_tracks, None);
 
-    // A decodable seek value, MTAwCg (100), but doesn't actually exist
+    // A decodable seek value, WyIwLjAuMCIsMTAwXQ (["0.0.0",100]), but doesn't actually exist
     let json: VersionList = anon
-        .get_with_query(url, "per_page=10&sort=semver&seek=MTAwCg")
+        .get_with_query(url, "per_page=10&sort=semver&seek=WyIwLjAuMCIsMTAwXQ")
```
Turbo87, line 296: with the old link the API will now return a "400 Bad Request" response

src/controllers/krate/versions.rs
```diff
@@ -249,131 +247,93 @@ async fn list_by_semver(
...
+        if options.is_some() {
             versions::table
                 .filter(versions::crate_id.eq(crate_id))
-                .left_outer_join(users::table)
-                .select(<(Version, Option<User>)>::as_select())
-                .filter(versions::id.eq_any(ids))
-                .load_stream::<(Version, Option<User>)>(conn)
+                .filter(not(versions::yanked))
+                .select(versions::num)
+                .load_stream::<String>(conn)
```
eth3lbert: Is a `.order(versions::semver_ord.desc())` call missing here, or am I missing something?

eth3lbert:
Ah, I just reviewed an outdated version :sweat_smile: I guess you decided not to introduce the change to release_tracks in this PR after all?

LawnGnome (approved):
I was tempted to just approve this without even reviewing it, because:

image

But I promise I did review it.

I was a bit nervous about the `seek` behaviour change, so I decided to dig through the last week's worth of logs to see if we were getting any significant programmatic use of the `seek` parameter with `sort=semver`, and the answer is: as far as I can tell, we have literally zero. There were seven requests in the last week with `sort=semver` and a non-empty `seek` parameter: two of them were me just now trying to remember how that page worked, and the other five were all from real browsers that I have no reason to believe weren't actually real people clicking on the load more button.

So I think we're good.

eth3lbert:
> not sure what happened, but I only pushed this branch once so far 😅

I'm not sure either. I just saw an "outdated" label on my previously deleted input 🤷 


> AFAIK I didn't touch the `release_tracks` code (yet). should I?

I think a separate PR is fine, but it's also okay if you'd like to make the changes in this PR, since at least part of it would be related to `semver_ord`.

Turbo87:
> But I promise I did review it.

thanks! 😆

> I think a separate PR is fine, but it's also okay if you'd like to make the changes in this PR, since at least part of it would be related to `semver_ord`.

I think a separate PR makes more sense. since I'll be on vacation starting tomorrow, that will have to wait until after my vacation though. feel free to take over though, if you want :)

PR #10824, opened 2025-03-13
controllers/krate/versions: Use `semver_ord` for `release_tracks` where applicable

eth3lbert:
This the follow-up PR to #10819 for the `release_tracks` part. :D

### DIFF
--- a/src/controllers/krate/versions.rs
+++ b/src/controllers/krate/versions.rs
@@ -204,6 +204,7 @@ async fn list(
                 .filter(versions::crate_id.eq(crate_id))
                 .filter(not(versions::yanked))
                 .select(versions::num)
+                .order(versions::semver_ord.desc())
                 .load_stream::<String>(conn)
                 .await?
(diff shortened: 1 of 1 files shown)


Turbo87 (approved):
ah, I see. I thought you wanted to move the entire release tracks calculation into the database 😅

eth3lbert:
> I actually experimented with this locally, and it is indeed doable.

This is something I tried locally:
```sql
SELECT DISTINCT ON (rt_name) num
FROM (
    SELECT num,
           CASE WHEN left(num, 1) = '0' THEN '.' || split_part(num, '.', 2)
                ELSE split_part(num, '.', 1)
           END AS rt_name,
           semver_ord
    FROM versions
    WHERE crate_id = (SELECT id FROM crates WHERE name = 'rand_core')
        AND NOT yanked -- no yanked
        AND semver_ord[3] = '{}' -- no pre-release
) t
ORDER BY rt_name, semver_ord DESC;
```

PR #10838, opened 2025-03-17
Replace `krata-tokio-tar` with `astral-tokio-tar`

eth3lbert:
Astral recently published `astral-tokio-tar`, which is a fork of `krata-tokio-tar` that preserves permissions and is actively maintained. I don't see any reason not to switch to it.


Turbo87 (approved):
sounds reasonable 👍

PR #10868, opened 2025-03-21
Make regenerate button visible for expired tokens

seqre:
The issue was reported by @fasterthanlime on https://hachyderm.io/@fasterthanlime/114199813804811002 and it was introduced by accident in https://github.com/rust-lang/crates.io/pull/9122. I moved the button outside the if statement and it seems to work!

Screenshot From 2025-03-21 23-22-55

Clicking on `Regenerate` redirects to the token creation as expected.

### DIFF
--- a/app/components/settings/api-tokens.hbs
+++ b/app/components/settings/api-tokens.hbs
@@ -112,15 +112,15 @@
         {{/if}}
 
         <div local-class="actions">
+          <LinkTo
+            @route="settings.tokens.new"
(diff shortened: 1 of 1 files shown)


fasterthanlime:
@seqre thanks for looking into this!

Turbo87 (approved):
thanks!

I've opened https://github.com/rust-lang/crates.io/pull/10873 to also add a couple of regression assertions to the test suite :)

PR #10874, opened 2025-03-22
CSS: Remove `composes` usage

Turbo87:
https://github.com/BlueCutOfficial/css-modules-to-scoped-css documents how to migrate from ember-css-modules to ember-scoped-css, which unlocks building the frontend application with https://vite.dev/ instead of webpack.

Unfortunately, ember-scoped-css does not support the `composes` declaration from ember-css-modules though.

This PR removes/replaces the 97 usages of `composes` in our codebase. I'm not in love with the solution, but it seems good enough for now to unlock that migration mentioned above.


Turbo87:
looks like there are a bunch of CSS specificity issues that only show up when restarting the `ember serve` command... 😩

PR #10888, opened 2025-03-25
Ensure CDN invalidations occur when crates or crate versions are deleted

### LINKED ISSUE #10887: CDNs may still serve up the old version of a replaced crate
It looks like we have at least one case where a crate has been published, deleted, and then republished where at least one CDN node has held onto the old version of the crate file and is still serving it up. (The rough timeline in this specific case is: published on the 13th, deleted on the 17th, republished on the 19th.)

The index file was correctly invalidated, and is correct across the CDNs, as far as I can tell.

I thought we were invalidating on delete, but I'm not so sure after 30 seconds with `rg`, so I'm going to dig into this further. If we're not, I'll add a job to ensure that we do so.

---

(For those with Zendesk access, this is derived from [#2140](https://rust-lang.zendesk.com/agent/tickets/2140).)

LawnGnome:
This can be reviewed commit-by-commit if you want, but it's probably small enough to digest without doing so.

Basically: we need to invalidate the CDNs when crate files (and related artifacts) are deleted, otherwise a crate that reuses the same name might still have the deleted crate file served from a CDN node.

I initially debated implementing this inline in the two different places we'd need it via a new `Environment::invalidate_cdns_many` function, but it felt cleaner to implement it as a new background job (and it doesn't feel like it's _super_ time critical, most of the time), so here we are.

I have one mild concern about this: namely, because Fastly requires one API request per path to be purged — and we have to purge two paths for each crate version, plus the RSS feed if we're deleting the whole crate — we may go somewhat closer to hitting the rate limit than previously. But I think let's try it and see what happens in practice.

I'd like to give this a test on staging before we deploy it, ideally.

Fixes #10887.


src/worker/jobs/invalidate_cdns.rs
```diff
@@ -0,0 +1,57 @@
...
+        //
+        // For now, we won't parallelise: most crate deletions are for new crates with one (or very
+        // few) versions, so the actual number of paths being invalidated is likely to be small, and
+        // this is all happening from either a background job or admin command anyway.
+        if let Some(fastly) = ctx.fastly() {
+            for path in self.paths.iter() {
+                fastly
+                    .invalidate(path)
+                    .await
+                    .with_context(|| format!("Fastly: {path}"))?;
```
Turbo87, line 44:
```suggestion
                    .with_context(|| format!("Failed to invalidate path on Fastly CDN: {path}"))?;
```

This makes the resulting error message look a bit easier to comprehend :)

src/worker/jobs/invalidate_cdns.rs
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+
+        if let Some(cloudfront) = ctx.cloudfront() {
+            cloudfront
+                .invalidate_many(self.paths.clone())
+                .await
+                .context("CloudFront")?;
```
Turbo87, line 52:
```suggestion
                .context("Failed to invalidate paths on CloudFront CDN")?;
```

src/cloudfront.rs
```diff
@@ -35,17 +35,21 @@ impl CloudFront {
     /// Invalidate a file on CloudFront
     ///
     /// `path` is the path to the file to invalidate, such as `config.json`, or `re/ge/regex`
-    #[instrument(skip(self))]
     pub async fn invalidate(&self, path: &str) -> anyhow::Result<()> {
-        let path = if path.starts_with('/') {
-            path.to_string()
-        } else {
-            format!("/{path}")
-        };
```
Turbo87, line 44: do we not need this logic anymore? are we sure that all invalidated paths always start with `/`?

Turbo87 (approved):
LGTM at first glance

src/cloudfront.rs
LawnGnome: Good catch! Not sure enough; I'll reinstate it.

LawnGnome:
Tested on staging; seems to work as expected, both for version and full crate deletions.

Should be good to deploy whenever we next deploy.

PR #10913, opened 2025-03-28
markdown: Add tasklist test cases

jendrikw:
The test are currently failing, which indicates a bug. From my understanding, the test should pass.

As per this URLO thread: https://users.rust-lang.org/t/missing-list-item-names-in-the-readme-on-crates-io/127558

You can see the bug live in action at
https://crates.io/crates/Yoda/0.11.5

### DIFF
--- a/crates/crates_io_markdown/lib.rs
+++ b/crates/crates_io_markdown/lib.rs
@@ -670,4 +670,20 @@ There can also be some text in between!
         </picture>
         "#);
     }
+
+    #[test]
+    fn nested_checkbox_lists() {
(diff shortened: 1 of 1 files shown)


Turbo87:
@jendrikw we are using https://github.com/kivikakk/comrak for the markdown rendering. if the conversion fails here then its probably best to report this upstream in their bug tracker, then the bugfix should automatically land on crates.io once it's fixed.

Turbo87:
Reported at https://github.com/kivikakk/comrak/issues/558

PR #10956, opened 2025-04-07
remove ember-fetch

mansona:
This is a really basic change assuming all network goes through your helper function

See more info here: https://rfcs.emberjs.com/id/1065-remove-ember-fetch/


Turbo87:
thanks @mansona! I've added additional `waitForPromise()` wrappers around the `response.text()` and `response.json()` calls, and it looks like CI is happy with this now! 🎉

Turbo87 (approved):

PR #10967, opened 2025-04-09
CI: Fix `changed-files` action usage

Turbo87:
Unfortunately this action needs to perform `git fetch` operations in some cases, which apparently requires `persist-credentials` to be `true`.


Turbo87:
huh... apparently that does not fix it, even though the error message says so 😂

PR #10990, opened 2025-04-11
Split "loading" placeholder from "load more" button

Turbo87:
Restyling the button to not be visible as a button makes the CSS more complicated than it needs to be. This commit splits the load more button into a loading placeholder for the initial load, and the actual load more button with a loading indicator for subsequent loads.

### DIFF
--- a/app/styles/crate/versions.module.css
+++ b/app/styles/crate/versions.module.css
@@ -33,21 +33,27 @@
     }
 }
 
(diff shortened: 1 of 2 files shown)


eth3lbert (approved):
Thanks, this LGTM! I've tested it locally, and it works as expected! :shipit:

PR #10995, opened 2025-04-11
sentry: remove `Authorization` header values before send

LawnGnome:
These get redacted in the Sentry ingest, but why send them at all?

### DIFF
--- a/src/sentry/mod.rs
+++ b/src/sentry/mod.rs
@@ -1,4 +1,5 @@
 use crate::config::SentryConfig;
+use http::header::AUTHORIZATION;
 use sentry::protocol::Event;
 use sentry::{ClientInitGuard, ClientOptions, TransactionContext};
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


src/sentry/mod.rs
```diff
@@ -53,10 +54,14 @@ pub fn init() -> Option<ClientInitGuard> {
...
-        // Remove cookies from the request to avoid sending sensitive
-        // information like the `cargo_session`.
         if let Some(request) = &mut event.request {
+            // Remove cookies from the request to avoid sending sensitive information like the
+            // `cargo_session`.
             request.cookies.take();
+
+            // Also remove `Authorization`, just so it never even gets sent to Sentry, even if
+            // they're redacting it downstream.
+            request.headers.remove(AUTHORIZATION.as_str());
```
Turbo87, line 64: mhhh... is this case-insensitive?

LawnGnome: Ugh, no — my brain vapour locked on `AUTHORIZATION` being `HeaderName`, never mind that it's obviously becoming a `&str` to handle the `BTreeMap<String, String>` that we actually have.

Thanks. Good catch. I'd blame Friday brain, but it's even more Friday for you.

Turbo87 (approved):
hard to review the tests on mobile but the actual code LGTM 👍

LawnGnome:
> hard to review the tests on mobile but the actual code LGTM 👍

All good, thanks. Just wanted the basic sanity check after my earlier Friday morning brain.

I owe you a beer or two.

src/sentry/mod.rs
Turbo87: Balmer peak kicked in 😉

Issue #10247, opened 2024-12-19
Tracking Issue for "Trusted Publishing Support"


Turbo87:
This is a tracking issue for the RFC "crates.io: Trusted Publishing Support" (rust-lang/rfcs#3691).

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.
Discussion comments will get marked as off-topic or deleted.
Repeated discussions on the tracking issue may lead to the tracking issue getting locked.

### Steps

- [x] Implement backend support in crates.io
  - [x] Add database table for (GitHub Actions) trusted publishing configs (https://github.com/rust-lang/crates.io/pull/11062)
  - [x] Implement API endpoint to create trusted publishing config (incl. email notification) (https://github.com/rust-lang/crates.io/pull/11113)
  - [x] Implement API endpoint to delete trusted publishing config (incl. email notification) (https://github.com/rust-lang/crates.io/pull/11209)
  - [x] Implement API endpoint to list trusted publishing configs (https://github.com/rust-lang/crates.io/pull/11230)
  - [x] Add database table for temporary access tokens (https://github.com/rust-lang/crates.io/pull/11062)
  - [x] Add database table for used OIDC token IDs (https://github.com/rust-lang/crates.io/pull/11062)
  - [x] Implement API endpoint to exchange GHA OIDC token for temporary access token (https://github.com/rust-lang/crates.io/pull/11131)
  - [x] Adjust exchange API endpoint to save used OIDC token IDs and reject replays (https://github.com/rust-lang/crates.io/pull/11131)
  - [x] Implement API endpoint to revoke temporary access token (https://github.com/rust-lang/crates.io/pull/11234)
  - [x] Adjust publish API endpoint to accept temporary access tokens (https://github.com/rust-lang/crates.io/pull/11294)
  - [x] Implement regular background job to delete expired temporary access tokens and OIDC token IDs (https://github.com/rust-lang/crates.io/pull/11222)
  - [x] Adjust GitHub Secret Scanning endpoints to work with temporary access tokens too (https://github.com/rust-lang/crates.io/pull/11405)
- [x] Implement frontend UI in crates.io
[...]

PR #11062, opened 2025-04-25
Create "Trusted Publishing" database tables

Turbo87:
This PR creates the three necessary database tables to implement [Trusted Publishing](https://github.com/rust-lang/crates.io/issues/10247) on crates.io:

- The `trustpub_configs_github` table is used to store the Trusted Publishing configurations for GitHub Actions. Users will be able to use the web interface to create new configurations, or remove existing ones. The `repository_owner_id` column will be looked up from the GitHub API when the configuration is created to avoid account resurrection attacks.
- The `trustpub_tokens` table contains temporary access tokens that can be used to publish corresponding crates. They can be created by exchanging a GitHub Actions OIDC token for a temporary access token, if the OIDC token corresponds to an existing Trusted Publising configuration.
- The `trustpub_used_jtis` table is used to ensure that OIDC tokens can only once be exchanged for temporary access tokens to avoid potential replay attacks.

Related:


migrations/2025-04-25-090000_trusted-publishing/up.sql
```diff
@@ -0,0 +1,64 @@
+create table github_oidc_configs
```
Turbo87, line 1: does it make sense to have a table per OIDC provider? or would it be better to save them all in the same table?

since they all use different claims I lean towards table per provider, but I'm open to the opposite if there are good reasons for it.

Turbo87, line 1: should this be names `oidc_github_configs` instead so that all the related tables use a `oidc_` prefix?

migrations/2025-04-25-090000_trusted-publishing/up.sql
```diff
@@ -0,0 +1,64 @@
...
+    crate_id int not null references crates on delete cascade,
+    hashed_token bytea not null,
+    expires_at timestamptz not null
+);
+
+comment on table oidc_tokens is 'Temporary access tokens for OIDC-based publishing (aka. Trusted Publishing)';
+comment on column oidc_tokens.id is 'Unique identifier of the `oidc_tokens` row';
+comment on column oidc_tokens.created_at is 'Date and time when the token was created';
+comment on column oidc_tokens.crate_id is 'Unique identifier of the crate that can be published using this token';
+comment on column oidc_tokens.hashed_token is 'SHA256 hash of the token that can be used to publish the crate';
```
Turbo87, line 41: a JWT can generally be used to publish multiple crates, if they all have a similar Trusted Publishing configuration for the same repository. during the exchange request the API server would create a dedicated `oidc_tokens` row for each crate with the same `hashed_token`.

would it be better to use a `crate_ids` array instead of row-per-crate? I'm not sure if an array would be able to support the `references crates` above.

on the other hand, the current design does not prevent hash collisions since a uniqueness constraint on only the `hashed_token` column is not possible.

migrations/2025-04-25-090000_trusted-publishing/up.sql
eth3lbert: I lean towards starting with a single table for two reasons: 1. It's the simplest way to begin. 2. After a quick glance at the given columns, I don't anticipate significant variations between different providers, and it seems like a good starting point for all providers already.

eth3lbert: Since this would only be used for a trusted publishing service, should we add trusted publishing (or its abbreviation/alias) to the table name? This is considering we might have other services that also have their own OIDC-related configurations. However, we could also rename it later when we actually need to, though.

migrations/2025-04-25-090000_trusted-publishing/up.sql
```diff
@@ -0,0 +1,64 @@
...
+comment on column github_oidc_configs.created_at is 'Date and time when the configuration was created';
+comment on column github_oidc_configs.crate_id is 'Unique identifier of the crate that this configuration is for';
+comment on column github_oidc_configs.repository_owner is 'GitHub name of the user or organization that owns the repository';
+comment on column github_oidc_configs.repository_owner_id is 'GitHub ID of the user or organization that owns the repository';
+comment on column github_oidc_configs.repository_name is 'Name of the repository that this configuration is for';
+comment on column github_oidc_configs.workflow_filename is 'Name of the workflow file inside the repository that will be used to publish the crate';
+comment on column github_oidc_configs.environment is 'GitHub Actions environment that will be used to publish the crate (if `NULL` the environment is unrestricted)';
+
+create index github_oidc_configs_repository_owner_id_repository_name_index
+    on github_oidc_configs (repository_owner_id, repository_name);
```
eth3lbert, line 24: I think it would be better to delay creating the index until we have our endpoint and SQL query, as we might change it to better suit our queries (e.g., if we want it to be a covering index).

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: > I don't anticipate significant variations between different providers

PyPI started to support Google as an OIDC provider for Trusted Publishing, which only uses two fields: email address (required) and subject (optional).

in other words, this is significantly different from what GitHub/GitLab support.

eth3lbert: Ah, good to know! I was only considering the repo holder previously :sweat_smile:. Then this indeed makes more sense to me, as it might vary for each provider!

I haven't investigated this yet, but if those columns won't be used for filtering and mainly serve as data, or if we only operate on them within the app, another possible solution could be to save those configurations as JSONB to accommodate different providers. Though the flow chart doesn't include this, I see an index clause (`github_oidc_configs (repository_owner_id, repository_name)`) in the schema, which means we might need these as filtering columns rather than just relying on either `id` or `crate_id`, so this might not be suitable?

Turbo87: > which means we might need these as filtering columns

the way it works is that we receive a JWT with various "claims". the `iss` claim (issuer) tells us where the JWT is coming from (e.g. GitHub Actions), which lets us select the correct table to search for corresponding Trusted Publishing configurations. the rest of the provider-specific claims are then used to find one or more matching configurations (multiple crates could be in the same repository). from these configurations we then have a list of crate IDs, that the newly generated temporary access token is valid for.

so yes, these columns are mainly used for filtering to find matching configurations for a JWT.

Turbo87: I've avoided `trusted_publishing_` so far because it seems quite long, and for `tp_` it isn't really obvious what it stands for, but I'm open to suggestions :)

eth3lbert: maybe we could add `_publisher` after the provider (e.g., `github_publisher`), similar to how PyPI has the `_publishers` suffix.

migrations/2025-04-25-090000_trusted-publishing/up.sql
jtgeibel: Yeah, I forget how arrays work when used as foreign keys. Seems simpler to leave it the way you have it here with multiple rows and no constraint on the `hashed_token`.

migrations/2025-04-25-090000_trusted-publishing/up.sql
jtgeibel: > should this be names `oidc_github_configs` instead so that all the related tables use a `oidc_` prefix?

I think that makes sense.

crates/crates_io_database/src/schema.rs
```diff
@@ -561,6 +561,28 @@ diesel::table! {
...
+    /// Trusted Publisher OIDC configuration for GitHub Actions
+    github_oidc_configs (id) {
+        /// Unique identifier of the `github_oidc_configs` row
+        id -> Int4,
+        /// Unique identifier of the crate that this configuration is for
+        crate_id -> Int4,
+        /// GitHub user or organization that owns the repository
+        repository_owner -> Varchar,
+        /// Unique identifier of the user or organization that owns the repository
+        repository_owner_id -> Int4,
```
jtgeibel, line 574: Is there a reason we need to store both string and ID versions of this? Also, since this isn't a foreign key, I assume it is the GH ID and not any of our own internal IDs. If so, maybe we should tweak the comment to clarify that.

jtgeibel (approved):
Looks good to me overall, just a few minor comments/questions.

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: > I've avoided `trusted_publishing_` so far because it seems quite long, and for `tp_` it isn't really obvious what it stands for, but I'm open to suggestions :)

I just came up with `trustpub_`. Short and somewhat self-describing with enough context. I think I prefer that to `oidc_`.

crates/crates_io_database/src/schema.rs
Turbo87: > since this isn't a foreign key, I assume it is the GH ID and not any of our own internal IDs. If so, maybe we should tweak the comment to clarify that.

good catch. I actually updated the comments in the migration script already to reflect that, but forgot to regenerate the schema file 😅 



> Is there a reason we need to store both string and ID versions of this?

storing the string is useful to be able to show it in the list of active trusted publishing configs for a crate. but also, if you rename a repository then the ID will stay the same, but the name will be different. in that case we probably want the publish to fail and require a reconfiguration?

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: > Yeah, I forget how arrays work when used as foreign keys

apparently they are not implemented at all according to the internet, but since the tokens are short-lived anyway it's probably not a huge deal to have IDs in the array that no longer correspond to an existing crate. since the crate IDs are unique and won't get reassigned we shouldn't be vulnerable to resurrection attacks or anything like that.

I've now adjusted the PR and the prototype to use the one-row-per-token approach with a `crate_ids: int[]` column, which then gives us the extra safety regarding hash collisions.

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: you're probably right. I've removed the non-unique index for now.

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: I changed the names to `trustpub_configs_github`, `trustpub_tokens` and `trustpub_used_jtis`

crates/crates_io_database/src/schema.rs
```diff
@@ -765,6 +765,58 @@ diesel::table! {
...
+    trustpub_tokens (id) {
+        /// Unique identifier of the `trustpub_tokens` row
+        id -> Int8,
+        /// Date and time when the token was created
+        created_at -> Timestamptz,
+        /// Date and time when the token will expire
+        expires_at -> Timestamptz,
+        /// SHA256 hash of the token that can be used to publish the crate
+        hashed_token -> Bytea,
+        /// Unique identifier of the crate that can be published using this token
```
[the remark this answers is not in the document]
Turbo87: that's actually a good catch. nice job AI 😅

migrations/2025-04-25-090000_trusted-publishing/up.sql
```diff
@@ -0,0 +1,61 @@
+create table trustpub_configs_github
+(
+    id serial primary key,
+    created_at timestamptz not null default now(),
+    crate_id int not null references crates on delete cascade,
+    repository_owner varchar not null,
+    repository_owner_id int not null,
+    repository_name varchar not null,
```
LawnGnome, line 8: Should we also be tracking repository ID?

migrations/2025-04-25-090000_trusted-publishing/up.sql
```diff
@@ -0,0 +1,61 @@
+create table trustpub_configs_github
+(
+    id serial primary key,
+    created_at timestamptz not null default now(),
+    crate_id int not null references crates on delete cascade,
```
LawnGnome, line 5: This is fine on a database level, but we will have to ensure that there's a reasonable path for multi-crate workspaces to be configured without necessarily having to do each one with many clicks in the UI.

migrations/2025-04-25-090000_trusted-publishing/up.sql
```diff
@@ -0,0 +1,61 @@
...
+    on trustpub_tokens (hashed_token);
+
+-------------------------------------------------------------------------------
+
+create table trustpub_used_jtis
+(
+    id bigserial primary key,
+    jti varchar not null,
+    used_at timestamptz not null default now(),
+    expires_at timestamptz not null
```
LawnGnome, line 51: Do we need to track this?

migrations/2025-04-25-090000_trusted-publishing/up.sql
LawnGnome: My initial reaction here was the same as @eth3lbert's, but I think I'm convinced that provider-specific tables are the way to go here.

LawnGnome (approved):

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: we use the owner ID to protect against resurrection attacks, but I'm not sure if the same would apply for the repository ID. if the user recreates the repository with the same name, do we really want to break the publishing workflow because of it?

aside from that, we might not be able to get the repository ID for private repositories, unless we use the auth token of a repository owner. that might work right now with GitHub being our only auth provider, but would cause some issues if that were to change eventually.

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: from what I can tell, PyPI requires you to set it up for each individual project too. let's start simple, this is what was agreed upon in the RFC.

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: otherwise we wouldn't be able to clean up this table and it would grow forever. I was planning on having a background job that regularly deletes expired tokens and jtis.

migrations/2025-04-25-090000_trusted-publishing/up.sql
Turbo87: we could open up the configuration creation for API tokens in the future and not just cookie auth, then those that want to automate it can do so without us having to make the web UI more complex for the simple use cases.

migrations/2025-04-25-090000_trusted-publishing/up.sql
woodruffw: Just chiming in from PyPI's side: the reasons @Turbo87 are the same ones we don't track the repository ID either -- in the best case it's no stronger than the resurrection guarantee of the owner ID, but it's hard/annoying to obtain in the private repo case 🙂

PR #11113, opened 2025-05-05
Implement `PUT /api/v1/trusted_publishing/github_configs` API endpoint

Turbo87:
This PR implements the first API endpoint for "Trusted Publishing" support in crates.io.

`PUT /api/v1/trusted_publishing/github_configs` can be used to create Trusted Publishing configurations specific to GitHub Actions. As discussed in https://github.com/rust-lang/crates.io/pull/11062#discussion_r2059728214, it seems best to make the configurations provider-specific, and only have the token exchange endpoint be generic.

The endpoint is currently limited to cookie authentication, so that we don't have to commit to this API yet, since our frontend will be the only user. Once we have validated that we are happy with the API we can open it up to API token authentication, but that will likely mean introducing another token scope, which seems out of scope for this PR.

This PR is probably best reviewed commit by commit.

Related:

### DIFF
--- a/crates/crates_io_trustpub/src/github/validation.rs
+++ b/crates/crates_io_trustpub/src/github/validation.rs
@@ -1,0 +1,154 @@
+use std::sync::LazyLock;
+
+const MAX_FIELD_LENGTH: usize = 255;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ValidationError {
+    #[error("GitHub repository owner name may not be empty")]
+    OwnerEmpty,
+    #[error("GitHub repository owner name is too long (maximum is {MAX_FIELD_LENGTH} characters)")]
+    OwnerTooLong,
+    #[error("Invalid GitHub repository owner name")]
+    OwnerInvalid,
+
+    #[error("GitHub repository name may not be empty")]
+    RepoEmpty,
+    #[error("GitHub repository name is too long (maximum is {MAX_FIELD_LENGTH} characters)")]
+    RepoTooLong,
+    #[error("Invalid GitHub repository name")]
+    RepoInvalid,
+
+    #[error("Workflow filename may not be empty")]
+    WorkflowFilenameEmpty,
+    #[error("Workflow filename is too long (maximum is {MAX_FIELD_LENGTH} characters)")]
+    WorkflowFilenameTooLong,
+    #[error("Workflow filename must end with `.yml` or `.yaml`")]
+    WorkflowFilenameMissingSuffix,
+    #[error("Workflow filename must be a filename only, without directories")]
+    WorkflowFilenameContainsSlash,
+
+    #[error("Environment name may not be empty (use `null` to omit)")]
+    EnvironmentEmptyString,
+    #[error("Environment name is too long (maximum is {MAX_FIELD_LENGTH} characters)")]
+    EnvironmentTooLong,
+    #[error("Environment name may not start with whitespace")]
+    EnvironmentStartsWithWhitespace,
+    #[error("Environment name may not end with whitespace")]
+    EnvironmentEndsWithWhitespace,
+    #[error(r#"Environment name must not contain non-printable characters or the characters "'", """, "`", ",", ";", "\""#)]
+    EnvironmentInvalidChars,
+}
+
+pub fn validate_owner(owner: &str) -> Result<(), ValidationError> {
+    static RE_VALID_GITHUB_OWNER: LazyLock<regex::Regex> =
+        LazyLock::new(|| regex::Regex::new(r"^[a-zA-Z0-9][a-zA-Z0-9-]*$").unwrap());
+
+    if owner.is_empty() {
+        Err(ValidationError::OwnerEmpty)
+    } else if owner.len() > MAX_FIELD_LENGTH {
+        Err(ValidationError::OwnerTooLong)
+    } else if !RE_VALID_GITHUB_OWNER.is_match(owner) {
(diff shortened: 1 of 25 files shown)


crates/crates_io_trustpub/src/github/validation.rs
```diff
@@ -0,0 +1,154 @@
...
+pub fn validate_owner(owner: &str) -> Result<(), ValidationError> {
+    static RE_VALID_GITHUB_OWNER: LazyLock<regex::Regex> =
+        LazyLock::new(|| regex::Regex::new(r"^[a-zA-Z0-9][a-zA-Z0-9-]*$").unwrap());
+
+    if owner.is_empty() {
+        Err(ValidationError::OwnerEmpty)
+    } else if owner.len() > MAX_FIELD_LENGTH {
+        Err(ValidationError::OwnerTooLong)
+    } else if !RE_VALID_GITHUB_OWNER.is_match(owner) {
+        Err(ValidationError::OwnerInvalid)
```
eth3lbert, line 51: This doesn't seem to cover the following cases:

1. Ends with `-`
2. Length exceeds `39`

crates/crates_io_trustpub/src/github/validation.rs
```diff
@@ -0,0 +1,154 @@
...
+pub fn validate_repo(repo: &str) -> Result<(), ValidationError> {
+    static RE_VALID_GITHUB_REPO: LazyLock<regex::Regex> =
+        LazyLock::new(|| regex::Regex::new(r"^[a-zA-Z0-9-_.]+$").unwrap());
+
+    if repo.is_empty() {
+        Err(ValidationError::RepoEmpty)
+    } else if repo.len() > MAX_FIELD_LENGTH {
+        Err(ValidationError::RepoTooLong)
+    } else if !RE_VALID_GITHUB_REPO.is_match(repo) {
+        Err(ValidationError::RepoInvalid)
```
eth3lbert, line 66: This doesn't seem to cover the following cases:

1. Reserved names `.` and `..`
2. Length exceeds `100`

eth3lbert (commented):
I haven't looked at all the other commits yet, but here's some feedback on the GitHub owner and repository validation parts (a bit more hardening).

crates/crates_io_trustpub/src/github/validation.rs
eth3lbert: I think the currently proposed regex might be sufficient, as the `regex` crate doesn't support lookaround features. If we'd like to harden this part, although there might be some workarounds for lookaround, I think it would be best implemented with simple string-related checks for maintenance reasons.

crates/crates_io_trustpub/src/github/validation.rs
```diff
@@ -0,0 +1,154 @@
...
+    } else if filename.contains('/') {
+        Err(ValidationError::WorkflowFilenameContainsSlash)
+    } else {
+        Ok(())
+    }
+}
+
+pub fn validate_environment(env: &str) -> Result<(), ValidationError> {
+    static RE_INVALID_ENVIRONMENT_CHARS: LazyLock<regex::Regex> =
+        LazyLock::new(|| regex::Regex::new(r#"[\x00-\x1F\x7F'"`,;\\]"#).unwrap());
```
LawnGnome, line 88: I'm betting you had fun figuring out what "non-printable character" meant in this context (from the GitHub error).

src/controllers/trustpub/github_configs/emails.rs
```diff
@@ -0,0 +1,50 @@
...
+Trusted Publishing configuration:
+
+- Repository owner: {repository_owner}
+- Repository name: {repository_name}
+- Workflow filename: {workflow_filename}
+- Environment: {environment}
+
+If you did not make this change and you think it was made maliciously, you can remove the configuration from the crate via the \"Settings\" tab on the crate's page.
+
+If you are unable to revert the change and need to do so, you can email help@crates.io to communicate with the crates.io support team."
```
LawnGnome, line 47: This is fine for now, but we should remember to link this to the documentation we write when this actually gets released.

LawnGnome (approved):
I think @eth3lbert stole my thunder on the only actual things I was going to ask about, so otherwise: LGTM. :zap:

Turbo87:
@eth3lbert @LawnGnome probably should've mentioned that in the PR message: the validation rules and email template are more or less taken directly from the PyPI implementation 😅 (@woodruffw please let me know if I need to adjust anything to comply with licensing etc.)

I guess technically we don't need the validations at all, since we only need to compare the values with the claims from the OIDC token. They are however useful for:

- avoiding GitHub API calls for user names that are known to be invalid
- avoiding common user mistakes when entering these values

I don't think we need to be quite as strict as GitHub though, so now I'm wondering why PyPI implemented e.g. the non-printable characters validation 🤔

eth3lbert:
> I guess technically we don't need the validations at all, since we only need to compare the values with the claims from the OIDC token. They are however useful for:
> 
>     * avoiding GitHub API calls for user names that are known to be invalid
> 
>     * avoiding common user mistakes when entering these values
> 
> 
> I don't think we need to be quite as strict as GitHub though

Yeah, fair enough. This was something I had in mind before leaving the comments, but I was unsure about the level of validation we should be concerned with, so I just went with a more hardening approach, which wouldn't be wrong, just perhaps unnecessary :D. But I'm genuinely okay with basic validation 👍

Turbo87:
@eth3lbert any objections on merging as-is? :)

PR #11131, opened 2025-05-08
Add `PUT /api/v1/trusted_publishing/tokens` API endpoint

Turbo87:
This is a big one... I'm sorry! 😅 

This PR implements the token exchange API endpoint used for the Trusted Publishing flow. This is roughly how it works:

- The endpoint accepts a JSON web token, issued by the GitHub Actions runner.
- It looks at the `iss` claim (Issuer) to figure out who issued this token and compares it against a list of supported Trusted Publishing providers (configurable via `TRUSTPUB_PROVIDERS`, but currently only supporting GitHub Actions).
- If a supported provider is found, it loads the JWKS (JSON Web Key Set) from their server, to validate that the JWT was indeed issued by that provider.
- It uses the `jti` claim from the JWT to prevent token reuse by attempting to insert a new row into the `trustpub_used_jtis` table, or returns an error if a uniqueness constraint is triggered.
- It then checks the `trustpub_configs_github` table for a configuration matching the other claims in the JWT
- If any configurations are found a temporary access token is generated, saved to the `trustpub_tokens` table, and then returned as the API response.

As with the previous PRs, this implementation is largely inspired by the corresponding implementation in https://github.com/pypi/warehouse. 

I've prepared the code to be somewhat provider-agnostic, but at this point the endpoint is still hardcoded to only support GitHub Actions. Refactoring this to completion can be done once we add support for another provider.

With that context I hope the PR is somewhat reviewable. It is definitely best reviewed commit by commit.

Related:


crates/crates_io_trustpub/src/access_token.rs
```diff
@@ -0,0 +1,110 @@
...
+use sha2::digest::Output;
+use sha2::{Digest, Sha256};
+
+/// A temporary access token used to publish crates to crates.io using
+/// the "Trusted Publishing" feature.
+#[derive(Debug)]
+pub struct AccessToken(SecretString);
+
+impl AccessToken {
+    const PREFIX: &str = "cio_tp_";
```
Turbo87, line 11: for regular tokens we use `cio` plus 32 alphanumeric character. when we implemented GitHub Secret Scanning we were told to use a more unique prefix with a separator. therefore the temporary access tokens are using the `cio_to_` prefix to make them easier to recognize with less false positives.

crates/crates_io_trustpub/src/access_token.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /// Create an access token from two u64 values.
+    ///
+    /// This is used internally by the `generate()` fn and is extracted
+    /// to a separate function for testing purposes.
+    fn from_u64s(r1: u64, r2: u64) -> Self {
+        let plaintext = format!("{}{r1:016x}{r2:016x}", Self::PREFIX);
+        Self(SecretString::from(plaintext))
+    }
```
Turbo87, line 25: - is 128bits of randomness enough?
- https://docs.github.com/en/code-security/secret-scanning/secret-scanning-partnership-program/secret-scanning-partner-program#identify-your-secrets-and-create-regular-expressions suggests adding a (32bit) checksum at the end for validation purposes. That would require us to find and add yet another dependency. Not sure if that's really worth it? 🤔

crates/crates_io_trustpub/src/github/test_helpers.rs
```diff
@@ -0,0 +1,128 @@
...
+            runner_environment: "github-hosted".into(),
+            actor: "octocat".into(),
+            workflow: "example-workflow".into(),
+            head_ref: "".into(),
+            base_ref: "".into(),
+            event_name: "workflow_dispatch".into(),
+            ref_type: "branch".into(),
+            job_workflow_ref: format!(
+                "{owner_name}/{repository_name}/.github/workflows/{workflow_filename}@refs/heads/main"
+            ),
```
Turbo87, line 87: see https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token

crates/crates_io_trustpub/src/github/workflows.rs
```diff
@@ -0,0 +1,109 @@
+use std::sync::LazyLock;
+
+/// Extracts the workflow filename from a GitHub workflow reference.
+///
+/// In other words, it turns e.g. `rust-lang/regex/.github/workflows/ci.yml@refs/heads/main`
+/// into `ci.yml`, or `None` if the reference is in an unexpected format.
+pub(crate) fn extract_workflow_filename(workflow_ref: &str) -> Option<&str> {
+    static WORKFLOW_REF_RE: LazyLock<regex::Regex> =
+        LazyLock::new(|| regex::Regex::new(r"([^/]+\.(yml|yaml))(@.+)").unwrap());
```
Turbo87, line 9: PyPI uses a "lookahead match" pattern, but that does not appear to be supported by the `regex` crate. I've slightly adjusted the pattern to make it work, but since all the test cases from PyPI are passing I'm fairly confident in the change.

crates/crates_io_trustpub/src/keystore/impl.rs
```diff
@@ -0,0 +1,82 @@
...
+/// The main implementation of the [`OidcKeyStore`] trait.
+///
+/// This struct fetches OIDC keys from a remote provider and caches them. If
+/// a key is not found in the cache, it will attempt to refresh the cached
+/// key set, unless the cache has just recently been refreshed.
+pub struct RealOidcKeyStore {
+    issuer_uri: String,
+    client: reqwest::Client,
+    cache: RwLock<Cache>,
+}
```
Turbo87, line 18: This implementation roughly matches how PyPI caches the JWKS, though they save it in Redis with an expiration time, while we cache them in memory on each API server.

I hope the code comments make it clear enough how the implementation works.

crates/crates_io_trustpub/src/test_keys.rs
```diff
@@ -0,0 +1,70 @@
...
+//! testing purposes. The keys are not secure and must not be used
+//! in production!
+
+use jsonwebtoken::errors::Error;
+use jsonwebtoken::{Algorithm, DecodingKey, EncodingKey};
+use serde::Serialize;
+use std::sync::LazyLock;
+
+const PRIVATE_KEY_PEM: &[u8] = br#"
+-----BEGIN PRIVATE KEY-----
```
Turbo87, line 11: I'm somewhat surprised GitHub allowed me to push this 😂 

As the code comment above says: **DO NOT USE THESE IN PRODUCTION!!!1**

src/bin/server.rs
```diff
@@ -37,6 +38,7 @@ fn main() -> anyhow::Result<()> {
         .databases_from_config(&config.db)
         .github(github)
         .github_oauth_from_config(&config)
+        .trustpub_providers(&list("TRUSTPUB_PROVIDERS")?)
```
Turbo87, line 41: this essentially acts as a feature flag for each of the providers. it allows us to enable e.g. `github` on the staging environment, but not yet on production. once we add new providers it will also allow us to gradually enable them.

if a provider is breached it would also allow us to temporarily disable them.

src/config/server.rs
```diff
@@ -186,6 +190,9 @@ impl Server {
                 .unwrap_or_default()
         );
 
+        let domain_name = dotenvy::var("DOMAIN_NAME").unwrap_or_else(|_| "crates.io".into());
+        let trustpub_audience = var("TRUSTPUB_AUDIENCE")?.unwrap_or_else(|| domain_name.clone());
```
Turbo87, line 194: this allows us to configure the expected `aud` claim, if necessary. it defaults to the domain name (crates.io / staging.crates.io).

src/controllers/trustpub/tokens/exchange/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    put,
+    path = "/api/v1/trusted_publishing/tokens",
+    request_body = inline(json::ExchangeRequest),
+    tag = "trusted_publishing",
+    responses((status = 200, description = "Successful Response", body = inline(json::ExchangeResponse))),
+)]
+pub async fn exchange_trustpub_token(
+    state: AppState,
+    json: json::ExchangeRequest,
+) -> AppResult<Json<json::ExchangeResponse>> {
```
Turbo87, line 32: in case you didn't find it in the large diff: this is the main entrypoint of the new API endpoint 😉

src/tests/util/test_app.rs
```diff
@@ -388,6 +392,17 @@ impl TestAppBuilder {
...
+    /// Add a new OIDC keystore to the application
+    pub fn with_oidc_keystore(
+        mut self,
+        issuer_url: impl Into<String>,
+        keystore: MockOidcKeyStore,
+    ) -> Self {
+        self.oidc_key_stores
+            .insert(issuer_url.into(), Box::new(keystore));
+        self
+    }
```
Turbo87, line 404: by default, all tests start with no Trusted Publishing providers enabled, but this fn can be used to explicitly add them (with `MockOidcKeyStore` implementations).

crates/crates_io_trustpub/src/access_token.rs
```diff
@@ -0,0 +1,110 @@
...
+/// the "Trusted Publishing" feature.
+#[derive(Debug)]
+pub struct AccessToken(SecretString);
+
+impl AccessToken {
+    const PREFIX: &str = "cio_tp_";
+
+    /// Generate a new access token.
+    pub fn generate() -> Self {
+        Self::from_u64s(rand::random(), rand::random())
```
woodruffw, line 15: Small thing, but two questions:

1. Why two `u64`s instead of a single `[u8; 16]`? The per-call overhead for the RNG is minuscule, but a single call would eliminate it and would prevent any implied separation between the halves of the uniformly random token.
2. Does `crates.io` have a forking/multi-process model? If so `rand::random()` _might_ be risky -- I believe it uses `ThreadRng` by default, which isn't guaranteed to be self-healing over process forks (i.e., it needs to be explicitly reseeded to prevent duplicated randomness between forks). If there are multiple processes at play that can call this `generate()` function, then my recommendation would be to use `OsRng` instead.

(The historical reason for not using `OsRng` is performance concerns around the system calls it can incur, but on Linux `getrandom` has been in the vDSO since 6.11.)

src/config/server.rs
woodruffw: Nice! This may or may not be applicable in your case, but noting: PyPI intentionally chose _not_ to use the domain here, since we thought we might want to change the authentication and/or exchange subdomain at some point. So PyPI uses `pypi` and TestPyPI uses `testpypi`.

(This _probably_ doesn't matter in practice, just flagging as a very small difference! In practice, any identifier that serves as a domain separator is fine, DNS name or whatever else 🙂)

src/controllers/trustpub/tokens/exchange/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+use diesel_async::scoped_futures::ScopedFutureExt;
+use diesel_async::{AsyncConnection, RunQueryDsl};
+use secrecy::ExposeSecret;
+
+#[cfg(test)]
+mod tests;
+
+/// Exchange an OIDC token for a temporary access token.
+#[utoipa::path(
+    put,
```
mdtro, line 23: To be compliant with [RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693#name-request), this should be done via a POST.

crates/crates_io_trustpub/src/access_token.rs
Turbo87: > Does `crates.io` have a forking/multi-process model?

unless I'm misunderstanding our application, crates.io only uses a single process, but then has multiple worker threads inside the process. since we don't have a GIL in Rust, or are generally single-threaded like in JS, there is usually no need for multiple processes.

> Why two `u64`s instead of a single `[u8; 16]`?

no real reason actually. it seemed easiest to get to the hex output from there, but I'm totally open to alternatives 😅

src/controllers/trustpub/tokens/exchange/mod.rs
Turbo87: seems like that RFC requires quite a few other things like using `application/x-www-form-urlencoded` (which PyPI also doesn't do). I wasn't aware that we were supposed to follow that RFC? 🤔 

the rest of the API generally uses `PUT` to create resources and `POST` for editing existing resources, which is why I chose `PUT` here.

src/controllers/trustpub/tokens/json.rs
```diff
@@ -0,0 +1,13 @@
+use axum::Json;
+use axum::extract::FromRequest;
+
+#[derive(Debug, Deserialize, FromRequest, utoipa::ToSchema)]
+#[from_request(via(Json))]
+pub struct ExchangeRequest {
+    pub jwt: String,
+}
```
mdtro, line 8: We probably want to expand and adjust this to be inline with [RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693#name-request).

The `jwt` here in the original implementation will be the `subject_token` in the spec.

Here's what I'd expect an example token exchange request to look like (a lot of implementations also support JSON content types here too):
```
POST /api/v1/trusted_publishing/tokens
Content-Type: application/x-www-form-urlencoded

grant_type=urn:ietf:params:oauth:grant-type:token-exchange&
subject_token=eyJhbGciOiJSUzI1NiIs...github-oidc-id-token...&
subject_token_type=urn:ietf:params:oauth:token-type:id_token&
audience=crates.io&
requested_token_type=urn:ietf:params:oauth:token-type:access_token
```

src/controllers/trustpub/tokens/exchange/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+            if crate_ids.is_empty() {
+                warn!("No matching Trusted Publishing config found");
+                let message = "No matching Trusted Publishing config found";
+                return Err(bad_request(message));
+            }
+
+            let new_token = AccessToken::generate();
+
+            let new_token_model = NewToken {
+                expires_at: chrono::Utc::now() + chrono::Duration::minutes(30),
```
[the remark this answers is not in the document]
Turbo87: > Would it make sense to make this configurable via an environment variable to make it easy for us to adjust?

I could make it configurable, though I'm wondering whether it would be better to accept it as a request parameter. I guess 30min is probably fine for this MVP though? :)

src/controllers/trustpub/tokens/exchange/mod.rs
woodruffw: Yeah, I don't believe PyPI follows that RFC either (at least, I didn't follow it while implementing Trusted Publishing 😅).

IMO the state/parameters used here are pretty distinct from the standard OAuth2 Token Exchange anyways, so deviating here seems like not a huge deal.

src/controllers/trustpub/tokens/exchange/mod.rs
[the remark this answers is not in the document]
woodruffw: As a datapoint: we currently do 15 minutes for PyPI, but all of you are eminently more qualified to know what makes more sense here!

https://github.com/pypi/warehouse/blob/c92eaf2777cb26586f2817136e80037c6697e55c/warehouse/oidc/views.py#L288

(Whatever expiry you'd pick, I'd recommend not making it client configurable -- IMO TP's benefit is easiest to justify when expiries are consistent, i.e. users can't configure their way into a longer-lived credential by accident.)

crates/crates_io_trustpub/src/github/claims.rs
```diff
@@ -0,0 +1,397 @@
...
+    pub aud: String,
+    #[serde(with = "ts_seconds")]
+    pub iat: DateTime<Utc>,
+    #[serde(with = "ts_seconds")]
+    pub exp: DateTime<Utc>,
+    pub jti: String,
+
+    pub repository_owner_id: String,
+    pub repository: String,
+    pub job_workflow_ref: String,
```
woodruffw, line 24: Sorry, I should have caught this earlier: I _think_ you want the `workflow_ref` claim here (and elsewhere below), instead of the `job_workflow_ref` claim.

For context: these two claims have the same value in most CI workflows, but diverge when a reusable workflow is used: `workflow_ref` always contains the "top-level" workflow reference, while `job_workflow_ref` either container the "top-level" reference *or* the reusable workflow reference, if one is used.

In other words, if a user has a single `publish.yml`, then they might see the following:

| `workflow_ref` | `job_workflow_ref` |
|--------|--------|
| `foo/bar/.github/workflows/publish.yml@SOMEREF` | `foo/bar/.github/workflows/publish.yml@SOMEREF` |


Whereas if they have a workflow structure like this, where `publish-reusable.yml` performs the actual publishing operation:

```
publish.yml
|
--> third-party/repo/publish-reusable.yml@OTHERREF
```

...then the claims will look like:


| `workflow_ref` | `job_workflow_ref` |
|--------|--------|
| `foo/bar/.github/workflows/publish.yml@SOMEREF` | `third-party/repo/.github/workflows/publish-reusable.yml@OTHERREF` |

PyPI's implementation made the mistake of conflating these two claims -- we end up doing so in a way that isn't exploitable, but that makes reusable workflows for publishing not work except for in a small handful of cases (i.e., where the org and repo already match, defeating the value proposition of third-party reusable workflows). See https://github.com/pypi/warehouse/issues/11096 for a bit more context there :slightly_smiling_face: 

TL;DR: You probably want to check the `workflow_ref` claim consistently, rather than the `job_workflow_ref` claim. The latter diverges from the former when a reusable workflow is used, meaning that checking it for consistency with the `repository` and `repository_owner_id` claims will "work" but will preclude use of crates.io's Trusted Publishing feature with third-party reusable workflows :slightly_smiling_face:

Turbo87: oh interesting. I think I copied that from PyPI, but if you're saying that that impl is flawed then I guess we should change it while we still can 😄

just to confirm: we only need to replace one with the other, correct?

woodruffw: > oh interesting. I think I copied that from PyPI, but if you're saying that that impl is flawed then I guess we should change it while we still can 😄

Yep -- the implementation from PyPI is flawed in such a way that makes it not-exploitable, but very difficult to modify for support for reusable workflows. So I would definitely recommend not making the same mistake we did 😅 

> just to confirm: we only need to replace one with the other, correct?

Correct! That will give you verification of the "top-level" workflow, which is the thing that Trusted Publishing conceptualizes as the publishing identity.

Turbo87: I've switched it over to `workflow_ref` now :)

crates/crates_io_trustpub/src/access_token.rs
Turbo87: I've completely rebuilt the `AccessToken` struct now.

It now uses 31 random alphanumeric characters and another alphanumeric character as a checksum, plus the prefix mentioned above. 

This is obviously not a cryptographic hash/checksum, but for the purposes that we want to use it for (preventing database interaction if the checksums don't match for a user-provided token) this shouldn't matter (unless I'm missing something 😅).

Another review on this part of the PR would be appreciated :)

crates/crates_io_trustpub/src/access_token.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Debug)]
+pub struct AccessToken(SecretString);
+
+impl AccessToken {
+    pub const PREFIX: &str = "cio_tp_";
+    const RAW_LENGTH: usize = 31;
+
+    /// Generate a new access token.
+    pub fn generate() -> Self {
+        let raw = Alphanumeric.sample_string(&mut rand::rng(), Self::RAW_LENGTH);
```
woodruffw, line 17: Just noting for cross-checking purposes: `log2(62) * 31` is approximately 185 bytes of entropy, so more than sufficient 🙂

crates/crates_io_trustpub/src/access_token.rs
```diff
@@ -0,0 +1,149 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AccessTokenError {
+    MissingPrefix,
+    InvalidLength,
+    InvalidCharacter,
+    InvalidChecksum { claimed: char, actual: char },
+}
+
+fn checksum(raw: &[u8]) -> char {
```
woodruffw, line 68: I also went ahead and checked the distribution on this checksum with a Python script:

image

(The vertical axis is the ASCII codepoint, so this suggests that a basic XOR here produces a pretty good distribution -- the only gaps are in the non-alphanumeric ranges, which is exactly what's expected.)

The script I used:

```python
#!/usr/bin/env python3

import random

ALPHANUMERIC = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"


def checksum(raw: str) -> str:
    checksum_value = 0
    for b in raw:
        checksum_value ^= ord(b)

    return ALPHANUMERIC[checksum_value % len(ALPHANUMERIC)]


def token() -> str:
    # Generate a random token of 31 characters
    return "".join(random.choices(ALPHANUMERIC, k=31))


for _ in range(10_000):
    raw = token()
    cksum = checksum(raw)
    print(f"{raw} {ord(cksum)}")
```

and then for plotting:

```bash
./checksum-distribution.py | awk '{ print $2 }' | gnuplot -p -e 'plot "/dev/stdin"'
```

crates/crates_io_trustpub/src/access_token.rs
Turbo87: yeah, we're using roughly the same pattern for the regular API keys, though without the checksum, so I figured it should be sufficient :)

crates/crates_io_trustpub/src/access_token.rs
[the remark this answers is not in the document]
Turbo87: oh, yeah, good point. I meant to add a couple more comments to the code, but forgot before I pushed 😅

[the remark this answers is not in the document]
Turbo87: done ✅

src/controllers/trustpub/tokens/exchange/mod.rs
mdtro: Ah, fair enough. :) My opinion here is loosely held, there's not hard requirement to implement this to the RFC. I don't believe I mentioned this RFC in the Rust RFC either. 😅

src/controllers/trustpub/tokens/json.rs
mdtro: Resolving this. See https://github.com/rust-lang/crates.io/pull/11131#discussion_r2080144926.

src/controllers/trustpub/tokens/exchange/mod.rs
LawnGnome: I think matching PyPI is probably more sensible right now than RFC 8693. It doesn't seem like implementing that RFC in the future would prevent us from implementing this the way we have right now anyway, so this is a changeable decision.

LawnGnome (approved):
Ugh, I just realised I put my review for this on the wrong PR yesterday. Fuck. Sorry.

Anyway, approved. Same notes as before.

crates/crates_io_trustpub/src/github/claims.rs
```diff
@@ -0,0 +1,399 @@
...
+    validation.required_spec_claims.insert("iss".into());
+    validation.required_spec_claims.insert("exp".into());
+    validation.required_spec_claims.insert("aud".into());
+    validation.validate_exp = true;
+    validation.validate_aud = true;
+    validation.validate_nbf = true;
+    validation.set_issuer(&[GITHUB_ISSUER_URL]);
+    validation.set_audience(&[audience]);
+    validation
+}
```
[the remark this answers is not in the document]
woodruffw: From PyPI's experience, you can go either way here: validating the `sub` claim would be essentially duplicate/cross-check the other claims, since the relevant parts of `sub` are also in `repository` and (indirectly) `repository_owner_id`.

(I've actually thought about removing the `sub` claim check for GitHub Actions from PyPI, since GitHub Actions actually allows users to [customize the claim](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#customizing-the-subject-claims-for-an-organization-or-repository) in ways that result in confusing errors being surfaced to users.)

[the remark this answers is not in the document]
eth3lbert: Ah, fair point! And thanks for sharing this!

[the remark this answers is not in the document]
Turbo87: yeah, I figured since the data in `sub` is redundant it's probably sufficient to validate the GitHub-specific claims instead :)

src/config/server.rs
Turbo87: `aud: crates` seems more confusing and unspecific to me than `aud: crates.io`. `aud: crates-io` has basically the same problem as `aud: crates.io`. since `aud: crates.io` would keep working even if we actually lost the ccTLD at some point I don't think it's worth introducing another name for it. if we had to migrate we could still allow both ccTLDs for the migration period.

Turbo87:
I think all comments/threads are addressed now, so let's :shipit: :)

PR #11167, opened 2025-05-13
Use same database pool creation code for API server and background worker

Turbo87:
This PR extracts a `create_database_pool()` fn from the API server initialization code and then subsequently uses it in the `bin/background-worker` binary.

This ensures that things like the `connection_timeout` and `statement_timeout` are consistently applied to both the API server and the background worker.


LawnGnome (approved):
LGTM.

PR #11209, opened 2025-05-20
Add `DELETE /api/v1/trusted_publishing/github_configs/{id}` API endpoint

Turbo87:
This PR implements another item on the todo list in https://github.com/rust-lang/crates.io/issues/10247: an API endpoint to delete Trusted Publishing configurations for GitHub Actions.

The implementation should be relatively straight-forward. It checks that the authenticated users is an owner of the crate, and if the configuration exists, it deletes it and sends a notification email to all the owners.

Related:

### DIFF
--- a/src/controllers/trustpub/github_configs/delete/mod.rs
+++ b/src/controllers/trustpub/github_configs/delete/mod.rs
@@ -1,0 +1,95 @@
+use crate::app::AppState;
+use crate::auth::AuthCheck;
+use crate::controllers::trustpub::github_configs::emails::ConfigDeletedEmail;
+use crate::util::errors::{AppResult, bad_request, not_found};
+use axum::extract::Path;
+use crates_io_database::models::OwnerKind;
+use crates_io_database::models::trustpub::GitHubConfig;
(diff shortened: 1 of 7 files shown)


src/controllers/trustpub/github_configs/delete/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+        return Err(bad_request("You are not an owner of this crate"));
+    }
+
+    // Delete the configuration from the database
+    diesel::delete(trustpub_configs_github::table.filter(trustpub_configs_github::id.eq(id)))
+        .execute(&mut conn)
+        .await?;
+
+    // Send notification emails to crate owners
+
```
eth3lbert, line 71: Whoops, I guess I found an unnecessary blank line :D

eth3lbert (approved):
Other than that blank line, the implementation looks solid to me, and it seems all the forbidden cases I'm aware of are already covered in the tests!
Nice work, thanks!

src/controllers/trustpub/github_configs/delete/mod.rs
eth3lbert: ah, fair enough!

PR #11294, opened 2025-06-04
controllers/krate/publish: Add support for Trusted Publishing access tokens

Turbo87:
This is the last major missing puzzle piece for the Trusted Publishing backend work: the ability to use temporary access tokens to publish new versions of crates.

Related:

- https://github.com/rust-lang/crates.io/pull/11131

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -1,7 +1,7 @@
 //! Functionality related to publishing a new crate or version of a crate.
 
 use crate::app::AppState;
-use crate::auth::AuthCheck;
+use crate::auth::{AuthCheck, Authentication};
 use crate::worker::jobs::{
     self, CheckTyposquat, SendPublishNotificationsJob, UpdateDefaultVersion,
 };
@@ -11,7 +11,7 @@ use cargo_manifest::{Dependency, DepsSet, TargetDepsSet};
 use chrono::{DateTime, SecondsFormat, Utc};
 use crates_io_tarball::{TarballError, process_tarball};
 use crates_io_worker::{BackgroundJob, EnqueueError};
-use diesel::dsl::{exists, select};
+use diesel::dsl::{exists, now, select};
 use diesel::prelude::*;
 use diesel::sql_types::Timestamptz;
 use diesel_async::scoped_futures::ScopedFutureExt;
@@ -19,8 +19,8 @@ use diesel_async::{AsyncConnection, AsyncPgConnection, RunQueryDsl};
 use futures_util::TryFutureExt;
 use futures_util::TryStreamExt;
 use hex::ToHex;
-use http::StatusCode;
 use http::request::Parts;
+use http::{StatusCode, header};
 use sha2::{Digest, Sha256};
 use std::collections::HashMap;
 use tokio::io::{AsyncRead, AsyncReadExt};
@@ -38,12 +38,13 @@ use crate::middleware::log_request::RequestLogExt;
 use crate::models::token::EndpointScope;
 use crate::rate_limiter::LimitedAction;
 use crate::schema::*;
-use crate::util::errors::{AppResult, BoxedAppError, bad_request, custom, internal};
+use crate::util::errors::{AppResult, BoxedAppError, bad_request, custom, forbidden, internal};
 use crate::views::{
     EncodableCrate, EncodableCrateDependency, GoodCrate, PublishMetadata, PublishWarnings,
 };
-use crates_io_database::models::versions_published_by;
+use crates_io_database::models::{User, versions_published_by};
 use crates_io_diesel_helpers::canon_crate_name;
+use crates_io_trustpub::access_token::AccessToken;
 
 const MISSING_RIGHTS_ERROR_MESSAGE: &str = "this crate exists but you don't seem to be an owner. \
      If you believe this is a mistake, perhaps you need \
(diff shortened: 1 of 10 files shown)


Turbo87:
damn... looks like the email redaction code isn't handling line breaks 😅

src/controllers/krate/publish.rs
```diff
@@ -126,35 +146,69 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
...
         .optional()?;
 
-    let endpoint_scope = match existing_crate {
-        Some(_) => EndpointScope::PublishUpdate,
-        None => EndpointScope::PublishNew,
-    };
+    let access_token = req
+        .headers
+        .get(header::AUTHORIZATION)
+        .and_then(|h| h.as_bytes().strip_prefix(b"Bearer "))
```
LawnGnome: [`auth-scheme` is specified as being case-insensitive](https://httpwg.org/specs/rfc9110.html#auth.scheme), so we should probably match this case insensitively. (RFC 6750 expresses no opinion either way.)

src/controllers/krate/publish.rs
```diff
@@ -126,35 +146,69 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
...
+        .map(AccessToken::from_byte_str)
+        .transpose()
+        .map_err(|_| forbidden("Invalid authentication token"))?;
+
+    let auth = match (access_token, &existing_crate) {
```
LawnGnome: It took me a couple of reads to remember how this worked and why it was OK for the `Some(...), Some(...)` to not fall through. As I understand it:

1. Normal token requests from `cargo` don't include an `auth-scheme`, so their `Authorization` headers can't start with `Bearer ` (the whitespace being important).
2. Therefore, if `Bearer ` is not present, then `access_token` is `None`.
3. So all non-TP publish requests will fall through to the `_` arm.

You'd probably word it better than I would, but let's maybe add a comment explaining that, and maybe we could rename `access_token` to something more specific (`bearer_token`, `trustpub_token`, I dunno)?

src/controllers/krate/publish.rs
```diff
@@ -369,17 +420,24 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
...
-            return Err(custom(StatusCode::FORBIDDEN, MISSING_RIGHTS_ERROR_MESSAGE));
-        }
+            let owners = krate.owners(conn).await?;
+            if Rights::get(user, &*app.github, &owners).await? < Rights::Publish {
+                return Err(custom(StatusCode::FORBIDDEN, MISSING_RIGHTS_ERROR_MESSAGE));
+            }
+
+            krate
+        } else {
+            // Trusted Publishing does not support creating new crates
```
LawnGnome, line 448: :+1:

LawnGnome (approved):
A couple of non-binding suggestions, but LGTM in general. I like the new `trustpub` test, in particular! :+1

src/controllers/krate/publish.rs
Turbo87: I've simplified the logic a bit from the two value match to a simpler `if trustpub_token` :)

src/controllers/krate/publish.rs
Turbo87: good catch, thanks. I've adjusted the code to handle auth schemes case-insensitively and also allow for additional whitespace between the auth scheme and the token.

eth3lbert:
🙋  I was wondering if we could use the creator of the trustpub config as `published_by` but then I realized we don't record that information :sweat_smile:. Is this a bad idea, or are we not supposed to do it this way?

Turbo87:
I think that could be misleading, since it is not the person that triggered the release itself. It might be possible to figure out the releaser from the `actor` or `actor_id` fields of the JWT, but that seems like a future enhancement to me and it would also only work for GitHub Actions, but not for GitLab CI or other Trusted Publishers.

eth3lbert:
:raising_hand: Another question came up. I noticed that only the regular (non-TrustPub) AuthType requires a verified email for publishing, while the TrustPub AuthType does not seem to require this. Would it be bad to ensure email is verified before creating a trustpub config?

Turbo87:
> Would it be bad to ensure email is verified before creating a trustpub config?

yeah, we should probably implement that 👍

src/controllers/krate/publish.rs
```diff
@@ -126,35 +146,76 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
         .await
         .optional()?;
 
-    let endpoint_scope = match existing_crate {
-        Some(_) => EndpointScope::PublishUpdate,
-        None => EndpointScope::PublishNew,
-    };
+    let trustpub_token = req
```
LawnGnome, line 152: I might still like a small explanatory comment here along the lines of:

```suggestion
    // Trusted publishing tokens are distinguished from regular crates.io API tokens because they
    // use the Bearer auth scheme, so we look for that specific prefix.
    let trustpub_token = req
```

But if you think that's unnecessary, that's not a blocker.

src/controllers/krate/publish.rs
```diff
@@ -126,35 +146,76 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
...
+            let mut split = h.as_bytes().splitn(2, |b| *b == b' ');
+            Some((split.next()?, split.next()?))
+        })
+        .filter(|(scheme, _token)| scheme.eq_ignore_ascii_case(b"Bearer"))
+        .map(|(_scheme, token)| token.trim_ascii())
+        .map(AccessToken::from_byte_str)
+        .transpose()
+        .map_err(|_| forbidden("Invalid authentication token"))?;
+
+    let auth = if let Some(trustpub_token) = trustpub_token {
```
LawnGnome, line 165: This is definitely clearer than before, so :+1:.

src/controllers/krate/publish.rs
Turbo87: done 👍 

FWIW this is a temporary state anyway. I have a WIP branch here to clean up our auth code, which should hopefully clarify the whole thing quite a bit.

eth3lbert:
🙋 A random question just came to mind. IIUC, the supported issuer is checked during the key exchange stage. This means that once they've exchanged the token, they can then ensure to publish a new version with it this time, even if we were to roll back support for this issuer right after the exchange, right?

Turbo87:
> they can then ensure to publish a new version with it this time, even if we were to roll back support for this issuer right after the exchange, right?

yes, that's how it's currently implemented. I guess if we wanted to change that we would need back-references from the temporary tokens to their configurations (i.e. one nullable column per provider?) and set them to "on delete cascade".

PR #11301, opened 2025-06-05
crate/settings: Hide "Add Owner" form behind "Add Owner" button

Turbo87:
This form does not need a dedicated header and does not need to be visible all the time. This commit hides it until the user explicitly clicks the "Add Owner" button. This frees up space on the page to add the list of "Trusted Publishing" configurations.

### Before

Bildschirmfoto 2025-06-05 um 16 04 31

### After

Bildschirmfoto 2025-06-05 um 16 02 14
Bildschirmfoto 2025-06-05 um 16 02 24

### DIFF
--- a/app/templates/crate/settings.hbs
+++ b/app/templates/crate/settings.hbs
@@ -2,17 +2,22 @@
 
 <CrateHeader @crate={{this.crate}} />
 
-<h2>Add Owner</h2>
+<div local-class="owners-header">
+  <h2>Owners</h2>
+  {{#unless this.addOwnerVisible}}
+    <button type="button" class="button button--small" data-test-add-owner-button {{on "click" this.showAddOwnerForm}}>Add Owner</button>
+  {{/unless}}
+</div>
 
-<form local-class="email-form" {{on "submit" (prevent-default (perform this.addOwnerTask))}}>
-  <label local-class="email-input-label" for='new-owner-username'>
-    Username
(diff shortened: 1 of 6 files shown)


app/templates/crate/settings.hbs
```diff
@@ -2,17 +2,20 @@
...
+</div>
 
-<h2>Owners</h2>
+{{#if this.addOwnerVisible}}
+  <form local-class="email-form" {{on "submit" (prevent-default (perform this.addOwnerTask))}}>
+    <label local-class="email-input-label" for='new-owner-username'>
+      Username
+    </label>
+    <Input @type="text" id="new-owner-username" @value={{this.username}} placeholder="Username" local-class="email-input" name="username" />
+    <button type="submit" disabled={{not this.username}} class="button button--small" data-test-save-button>Save</button>
```
LawnGnome, line 18: I tend to like specific verbs on forms, so maybe:

```suggestion
    <button type="submit" disabled={{not this.username}} class="button button--small" data-test-save-button>Add Owner</button>
```

LawnGnome (approved):
This is definitely better. :+1:

app/templates/crate/settings.hbs
Turbo87: hahaha, I have a commit here already for a follow-up PR that pretty much does that 😂

Issue #11347, opened 2025-06-12
Trusted Publishing: Access tokens are only accepted with `Bearer` auth scheme


Turbo87:
### Current Behavior

see title

### Expected Behavior

The temporary access tokens from the Trusted Publishing flow are still going to be used with `cargo publish` which just sends `Authorization: <token>` instead of `Authorization: Bearer <token>`, so our current Trusted Publishing token detection logic doesn't actually make much sense...

PR #11350, opened 2025-06-12
Allow empty and `Bearer` auth schemes in `Authorization` headers

Turbo87:
`cargo publish` was built without using an auth scheme in the `Authorization` header. Regular HTTP client libraries often only support either basic or bearer authentication, but not empty auth schemes.

The regular API tokens currently only support an empty auth scheme, while Trusted Publishing was (mistakenly) built in a way to only support the `Bearer` auth scheme.

This PR adjusts both code paths to use a shared `Authorization` header extractor which allows usage of an empty auth scheme, or the `Bearer` auth scheme, regardless of the type of token that is used or the API endpoint it is used on.

Related:

- Resolves https://github.com/rust-lang/crates.io/issues/11347

### DIFF
--- a/src/controllers/trustpub/tokens/revoke/tests.rs
+++ b/src/controllers/trustpub/tokens/revoke/tests.rs
@@ -62,6 +62,27 @@ async fn test_happy_path() -> anyhow::Result<()> {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn test_happy_path_without_auth_scheme() -> anyhow::Result<()> {
+    let (app, _client) = TestApp::full().empty().await;
+    let mut conn = app.db_conn().await;
+
+    let token1 = new_token(&mut conn, 1).await?;
+    let _token2 = new_token(&mut conn, 2).await?;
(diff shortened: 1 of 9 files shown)


src/controllers/trustpub/tokens/revoke/tests.rs
```diff
@@ -61,13 +61,34 @@ async fn test_happy_path() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_happy_path_with_auth_scheme() -> anyhow::Result<()> {
```
eth3lbert: Nit: It would be great if the function name explicitly included `empty_auth_scheme`.

eth3lbert (approved):
Yeah, I just checked that there's even an [open issue in cargo about the auth scheme](https://github.com/rust-lang/cargo/issues/4989). But anyway, this LGTM. I just left a small nit. Maybe we also want to add some comments in the codebase, even linking to issues, to provide context and explain why we need to support this.

src/controllers/trustpub/tokens/revoke/tests.rs
Turbo87: lol, yeah, it was supposed to be `without`, not `with` 😅

PR #11351, opened 2025-06-12
trustpub: Disallow config creation if email is not verified

Turbo87:
see https://github.com/rust-lang/crates.io/pull/11294#issuecomment-2949033375

### DIFF
--- a/src/controllers/trustpub/github_configs/create/mod.rs
+++ b/src/controllers/trustpub/github_configs/create/mod.rs
@@ -3,7 +3,7 @@ use crate::auth::AuthCheck;
 use crate::controllers::krate::load_crate;
 use crate::controllers::trustpub::github_configs::emails::ConfigCreatedEmail;
 use crate::controllers::trustpub::github_configs::json;
-use crate::util::errors::{AppResult, bad_request};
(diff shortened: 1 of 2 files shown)


src/controllers/trustpub/github_configs/create/mod.rs
```diff
@@ -61,8 +61,14 @@ pub async fn create_trustpub_github_config(
         .load::<(i32, String, String, bool)>(&mut conn)
         .await?;
 
-    if !user_owners.iter().any(|owner| owner.0 == auth_user.id) {
-        return Err(bad_request("You are not an owner of this crate"));
+    let auth_owner = user_owners
```
eth3lbert: Nit: Instead of `auth_owner.3` It might be more readable to destructure it as a variable.
```suggestion
    let (.., verified) = user_owners
```
or
```suggestion
    let (_, _, _, verified) = user_owners
```

eth3lbert (approved):
Thanks, it looks great! Just a small nit, but it's not a blocker.

src/controllers/trustpub/github_configs/create/mod.rs
Turbo87: that's a good idea. thanks! :)

PR #11366, opened 2025-06-16
Replace yank button with a drop down menu

### LINKED ISSUE #11139: Yank button isn’t visible on mobile in portrait orientation
### Current Behavior

The yank button isn’t visible in portrait orientation on my phone (iPhone 11). I had to yank a crate version on my phone today and couldn’t because there was no button. (Later I found out the button becomes visible in landscape orientation)

### Expected Behavior

The yank button should always be visible

### Steps To Reproduce

1. Go to versions of a crate you control
2.  Observe there’s no yank button if the phone is in portrait orientation

### Environment

- Browser: Safari
- OS: iOS 18.4.1

### Anything else?

iPhone 11

eth3lbert:
We plan to add more actions in the future, and to prevent too many buttons from showing on each row and on small-screen devices, @Turbo87 suggests we should use a drop down menu instead.

Resolves #11139.

### DIFF
--- a/app/components/privileged-action.hbs
+++ b/app/components/privileged-action.hbs
@@ -1,14 +1,14 @@
 {{#if this.isPrivileged}}
-  <div>
+  <div ...attributes>
     {{yield}}
   </div>
 {{else if this.canBePrivileged}}
   {{#if (has-block 'placeholder')}}
-    <div>
+    <div ...attributes>
       {{yield to='placeholder'}}
     </div>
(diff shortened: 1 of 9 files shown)


Turbo87:
thanks!

Bildschirmfoto 2025-06-18 um 09 56 09

I slightly adjusted the look of it a bit :)

Turbo87 (approved):

Issue #10611, opened 2025-02-18
Refactoring to disentangle account creation from only one identity provider


carols10cents:
- [ ] Start writing `gh_id`, `gh_access_token`, `gh_login`, `gh_avatar` to both the `users` table and a new table with more generic names that also has a column indicating the OAuth provider that is the source of the values (which will always be set to GitHub for now. Deploy and manually verify values are written/updated as expected, but otherwise don't use the values in the new table.
- [ ] Write a script to backfill the new table with data from all existing records as appropriate. Do an offline test with a production database dump to know how long it will take the script to run and what the load on the database will be like. If necessary, batch updates done by the script to minimize production app impacts.
- [ ] Change all code that uses the users table's `gh_id`, `gh_access_token`, `gh_login`, `gh_avatar` fields to instead use the new table.
- [ ] In a migration, rename the users table's `gh_id`, `gh_access_token`, `gh_login`, `gh_avatar` columns to start with `deprecated_` to allow for easy rollback; deploy and monitor.
- [ ] When satisfied that nothing is using the deprecated fields, write another migration removing the fields entirely.

MarkusEicher:
First let me say thanks to all people working on making crates.io and dependency management better and more secure in the Rust ecosystem. Maybe I am outing myself as a junior dev (which I am) with this question, but could you imagine to have something like the trusted publisher workflow for publishing crates to crates.io as a user too? What I mean, is that I would prefer to use OIDC from a IDP I trust (eg. Google, GitHub....) and having the same short term access token for every single push to the crates.io repo, instead of a long term api token.

stacksjb:
> First let me say thanks to all people working on making crates.io and dependency management better and more secure in the Rust ecosystem. Maybe I am outing myself as a junior dev (which I am) with this question, but could you imagine to have something like the trusted publisher workflow for publishing crates to crates.io as a user too? What I mean, is that I would prefer to use OIDC from a IDP I trust (eg. Google, GitHub....) and having the same short term access token for every single push to the crates.io repo, instead of a long term api token.

Is your request at all related to https://crates.io/docs/trusted-publishing ? That already allows publishing without long term token.

MarkusEicher:
Yes. Maybe I did not understand it correctly. Afaik this is available with GitHub Actions, right? I meant if it could be used for user authentication from the cli with an IDP  other than GitHub and by using claims and short time tokens. As I said, it may be I have to do more research, if I got some stuff wrong.

[...]

PR #11379, opened 2025-06-18
Refactor auth code

Turbo87:
Up until this point our "auth" code was scattered all over the place and mixed up authentication (who are you?) and authorization (what are you allowed to do?) in multiple places:

- Whether a user account was locked was checked as part of authentication, although it is more of an authorization concern.
- Whether a user has a verified email address was checked in the individual endpoints, if at all.
- Whether a user is an admin was also checked in the individual endpoints that were adjusted to support admin users.
- Whether a user is an owner (or team member) of a crate was also checked in the individual endpoints, in various different ways.
- Whether an API token has the right scopes was also checked as part of authentication, but it is an authorization concern.

This commit centralizes these checks with a three step approach:

1. In the first step the endpoint uses one of the `Credentials` structs as an axum request extractor to extract the raw credentials from the request metadata (user ID from session cookie, API token, Trusted Publishing token).
2. The credentials are used to look up a corresponding user (or Trusted Publishing token) from the database.
3. The authenticated entity is used to check for authorization of the requested operation.

The second and third step are folded into one fn call (`.validate(permission)`) to make it harder to use an authenticated entity without knowing if they are authorized for the operation.

While the auth code and the endpoints have changed quite a bit due to this refactoring, the test code is mostly the same except for a few minor details caused by using the request extractors now.


Turbo87:
@rust-lang/crates-io I've opened this as a draft since I'm mostly looking for feedback on this idea. Please take a look and let me know if this is actually an improvement or not. I'm a little worried that I've over-complicated it, but I'm not quite sure if it can be simplified in a significant way.

Turbo87:
@rust-lang/crates-io it's been almost 6 months now. does anyone have an opinion on this?!

A248:
@steveklabnik @alexcrichton this is a semi-blocker for non-Github account creation

Turbo87:
> this is a semi-blocker for non-Github account creation

no, it's not

A248:
Okay, sorry then. However @carols10cents in #10611 seemed to think that it was.

carols10cents:
> Okay, sorry then. However @carols10cents in #10611 seemed to think that it was.

[I said "might"](https://github.com/rust-lang/crates.io/issues/10611#issuecomment-3254881232)

Turbo87:
almost one year and still no feedback... guess I'll throw the work away then 🤷‍♂️

Issue #6335, opened 2023-04-14
Require verified email address to accept a crate tansfer


jtgeibel:

Turbo87:
implemented by https://github.com/rust-lang/crates.io/pull/11404

PR #11404, opened 2025-06-22
Require verified email address to accept a crate ownership invitation

Turbo87:
This PR implements https://github.com/rust-lang/crates.io/issues/6335, by returning errors to users with unverified email addresses that attempt to accept ownership invitations.


eth3lbert (approved):
LGTM 👍  
The `crate_name` seems to be used only when errors occur. I'm not sure if it's worth extracting it as a function and only calling it inside the error block. However, given that this should be a rare endpoint, I think it's also fine to keep it as is for now.

LawnGnome (approved):

PR #11405, opened 2025-06-22
controllers/github/secret_scanning: Add support for Trusted Publishing tokens

Turbo87:
This implements the final todo item of the Trusted Publishing backend work (see https://github.com/rust-lang/crates.io/issues/10247).

If a passed in token is successfully parsed as a Trusted Publishing token, the token is automatically revoked and a warning is logged.

Since these tokens belong to a crate (or multiple) instead of a user I have not implemented email notifications for them (yet). Should we email all owners of the crate in case a Trusted Publishing token is leaked? I guess we could also implement that in a follow-up PR, if we decide that we want that.

### DIFF
--- a/src/controllers/github/secret_scanning.rs
+++ b/src/controllers/github/secret_scanning.rs
@@ -8,7 +8,9 @@ use anyhow::{Context, anyhow};
 use axum::Json;
 use axum::body::Bytes;
 use base64::{Engine, engine::general_purpose};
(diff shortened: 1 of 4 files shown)


LawnGnome (approved):
LGTM.

For what it's worth, I do think we should figure out a notification strategy here, since it'll probably indicate some sort of CI compromise, but that can be a follow up PR. (E-mailing all the owners feels fine to me as a first step.)

PR #11413, opened 2025-06-23
msw: Implement simplified docs rebuild endpoint

eth3lbert:
This commit implements the `/api/v1/crates/{name}/{version}/rebuild_docs` endpoint, which is relatively simple but good enough for frontend testing.


Turbo87 (approved):
generally LGTM, but seems that CI is unhappy 😅

PR #11416, opened 2025-06-23
Add Package URL (PURL) display to crate sidebar

Turbo87:
This PR adds [Package URL (PURL)](https://github.com/package-url/purl-spec) support to crate pages, enabling better integration with software supply chain and security tools. PURLs follow the format `pkg:cargo/{name}@{version}` and are displayed in the sidebar metadata section with click-to-copy functionality.

This is a frontend-only implementation that generates PURLs client-side using existing crate data. The display uses ellipsis overflow with a tooltip showing the full PURL.

On `staging.crates.io` (and everywhere that is no `crates.io` itself) it will append a `repository_url` qualifier (see https://github.com/package-url/purl-spec/issues/215).

Bildschirmfoto 2025-06-23 um 21 05 07
Bildschirmfoto 2025-06-23 um 21 04 30

### DIFF
--- a/app/utils/purl.js
+++ b/app/utils/purl.js
@@ -1,0 +1,22 @@
+import window from 'ember-window-mock';
+
+/**
+ * Adds a repository_url query parameter to a PURL string based on the host.
+ *
+ * @param {string} purl - The base PURL string (e.g., "pkg:cargo/serde@1.0.0")
+ * @param {string} [host] - The host to use for repository URL. Defaults to current window location host.
+ * @returns {string} The PURL with repository_url parameter added, or unchanged if host is crates.io
+ */
+export function addRegistryUrl(purl) {
+  let host = window.location.host;
+
+  // Don't add repository_url for the main crates.io registry
+  if (host === 'crates.io') {
+    return purl;
+  }
+
+  // Add repository_url query parameter
+  const repositoryUrl = `https://${host}/`;
+  const separator = purl.includes('?') ? '&' : '?';
+  return `${purl}${separator}repository_url=${encodeURIComponent(repositoryUrl)}`;
+}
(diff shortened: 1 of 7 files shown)


app/utils/purl.js
```diff
@@ -0,0 +1,22 @@
...
+  let host = window.location.host;
+
+  // Don't add repository_url for the main crates.io registry
+  if (host === 'crates.io') {
+    return purl;
+  }
+
+  // Add repository_url query parameter
+  const repositoryUrl = `https://${host}/`;
+  const separator = purl.includes('?') ? '&' : '?';
```
LawnGnome, line 20: It feels like this is redundant, since we can't ever generate a package URL that includes a `?` right now?

(I guess, more broadly, I'm not quite clear on why this is separated from the basic PURL generation in the model.)

LawnGnome (approved):
:+1:

One thought I do have is whether we should add some sort of help explaining what this is, since PURLs aren't a super well known concept right now. One thought: could we add a question mark on the right hand side of the PURL that has a hover explaining what a PURL is and why it's useful? Or maybe we have some other way of linking to more information. :shrug:

app/utils/purl.js
Turbo87: > I'm not quite clear on why this is separated from the basic PURL generation in the model

I initially implemented the same property on the `crate` model too and wanted to DRY it up 😉

Turbo87:
I've added a small help icon next to the PURL now:

Bildschirmfoto 2025-06-24 um 09 28 35
Bildschirmfoto 2025-06-24 um 09 30 43

LawnGnome (approved):
Thanks for that!

Turbo87:
with one approval and no objections during the team meeting, I guess this is ready to merge now :)

since this is a frontend-only feature for now, we can easily revert this if we notice any issues.

PR #11419, opened 2025-06-24
Implement Trusted Publishing token exposure notifications

Turbo87:
As requested in https://github.com/rust-lang/crates.io/pull/11405, this PR implements email notifications to all related crate owners if a Trusted Publishing token is publicly exposed and reported to us by the GitHub secret scanning program.

One thing to note with the proposed implementation: In the scenario where user 1 is an owner of crate 1 and 2, and user 2 is an owner of only crate 2, then the notification of user 2 will only mention crate 2. I wonder if that is the right thing to do or not. We could also expose the list of all publishable crates to all users, independent from their ownership status 🤔

### DIFF
--- a/src/controllers/github/secret_scanning.rs
+++ b/src/controllers/github/secret_scanning.rs
@@ -1,23 +1,26 @@
 use crate::app::AppState;
 use crate::email::Email;
 use crate::models::{ApiToken, User};
-use crate::schema::api_tokens;
+use crate::schema::{api_tokens, crate_owners, crates, emails};
 use crate::util::errors::{AppResult, BoxedAppError, bad_request};
 use crate::util::token::HashedToken;
 use anyhow::{Context, anyhow};
 use axum::Json;
 use axum::body::Bytes;
 use base64::{Engine, engine::general_purpose};
+use crates_io_database::models::OwnerKind;
 use crates_io_database::schema::trustpub_tokens;
 use crates_io_github::GitHubPublicKey;
 use crates_io_trustpub::access_token::AccessToken;
 use diesel::prelude::*;
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
+use futures_util::TryStreamExt;
 use http::HeaderMap;
 use p256::PublicKey;
 use p256::ecdsa::VerifyingKey;
 use p256::ecdsa::signature::Verifier;
 use serde_json as json;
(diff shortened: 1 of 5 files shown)


src/controllers/github/secret_scanning.rs
```diff
@@ -138,19 +141,40 @@ async fn alert_revoke_token(
...
+        let Some(crate_ids) = token_data else {
             debug!("Unknown Trusted Publishing token received (false positive)");
             return Ok(GitHubSecretAlertFeedbackLabel::FalsePositive);
+        };
+
+        // Delete the token
+        diesel::delete(trustpub_tokens::table)
+            .filter(trustpub_tokens::hashed_token.eq(hashed_token.as_slice()))
+            .execute(conn)
+            .await?;
```
eth3lbert: Hmm, I guess this could still be done with a single `DELETE` statement, but by using `returning()` to get the `crate_ids`?

Turbo87: yeah, that's a good idea. thanks! :)

Turbo87: I've implemented it with `returning()` now

src/controllers/github/secret_scanning.rs
```diff
@@ -224,6 +239,76 @@ async fn send_notification_email(
...
+                .push(crate_name.clone());
+        }
+    }
+
+    // Send notifications in sorted order by email for consistent testing
+    let mut sorted_notifications: Vec<(String, Vec<String>)> = notifications.into_iter().collect();
+    sorted_notifications.sort_by(|a, b| a.0.cmp(&b.0));
+
+    for (email, mut crate_names) in sorted_notifications {
+        crate_names.sort();
```
eth3lbert: Have you considered just using a sorted container like `BTreeMap` and `BTreeSet`?

Turbo87: they use insertion sorting though, right? I guess for stable snapshots that should be sufficient, but having the crate names sorted alphabetically might make sense?

Turbo87: > they use insertion sorting though

ah, nevermind. I misread the docs 😅

Turbo87: I've changed it to used `BTreeMap` and `BTreeSet` instead 👍

PR #11420, opened 2025-06-24
Use `minijinja` templates for emails

Turbo87:
This PR refactors our email sending code to use `minijinja` templates instead of `format!()` and string concatenation. In the long term this even enables us to add dedicated templates for HTML emails too, or use template inheritance to use consistent email footers, etc.

**Before**:
```rust
impl Email for UserConfirmEmail<'_> {
    fn subject(&self) -> String { /* ... */ }
    fn body(&self) -> String { /* ... */ }
}

emails.send(recipient, user_confirm_email).await?;
```

**After** (with templates):
```rust
let email_message = EmailMessage::from_template("user_confirm", context! {
    user_name => user_name,
    domain => domain,
    token => token.expose_secret()
})?;

emails.send(recipient, email_message).await?;
```

```jinja2
Hello {{ user_name }}! Welcome to crates.io. Please click the
link below to verify your email address. Thank you!

https://{{ domain }}/confirm/{{ token }}
```


Turbo87:
rebased to include the new email type from https://github.com/rust-lang/crates.io/pull/11419

src/controllers/session.rs
```diff
@@ -170,13 +172,19 @@ async fn create_or_update_user(
...
                     // Swallows any error. Some users might insert an invalid email address here.
-                    let email = UserConfirmEmail {
-                        user_name: &user.gh_login,
-                        domain: &emails.domain,
-                        token,
-                    };
-                    let _ = emails.send(user_email, email).await;
+                    if let Ok(email) = email {
+                        let _ = emails.send(user_email, email).await;
+                    }
```
eth3lbert: Nit: I assume this is intended to only swallow errors related to sending, so it would be nice to log a warning when rendering fails, just like with others.

src/controllers/user/update.rs
```diff
@@ -106,45 +108,23 @@ pub async fn update_user(
...
+            context! {
+                user_name => user.gh_login,
+                domain => state.emails.domain,
+                token => token.expose_secret()
+            },
+        );
+
+        if let Ok(email) = email {
+            let _ = state.emails.send(user_email, email).await;
+        }
```
eth3lbert: Nit: same here, probably log a warning when rendering fails.

src/worker/jobs/sync_admins.rs
```diff
@@ -138,72 +138,30 @@ impl BackgroundJob for SyncAdmins {
...
-
-        Ok(())
-    }
+async fn send_email(
+    ctx: &Environment,
+    address: &str,
+    context: &minijinja::Value,
+) -> anyhow::Result<()> {
+    let email = EmailMessage::from_template("admin_account", context)?;
+    Ok(ctx.emails.send(address, email).await?)
```
eth3lbert: Nit: Hmm, I'm wondering why this isn't adding `.context()` with either a rendering error or a sending error like the others.

src/controllers/token.rs
```diff
@@ -286,28 +283,17 @@ pub async fn revoke_current_api_token(app: AppState, req: Parts) -> AppResult<Re
...
+    token_name: &str,
+    user_name: &str,
+) -> anyhow::Result<()> {
+    let context = context! {
+        token_name,
+        user_name,
+        domain => emails.domain,
+    };
+    let email = EmailMessage::from_template("new_token", context)?;
+    Ok(emails.send(recipient, email).await?)
```
eth3lbert: Nit: Why is the template context not built outside this function and passed into it, as is done with others, but rather constructed within this function?

eth3lbert: Nit: Same here, `.context()` with either rendering error or a sending error.

Turbo87:
Thanks for the review! All of the feedback should be addressed now :)

src/controllers/token.rs
```diff
@@ -286,28 +289,13 @@ pub async fn revoke_current_api_token(app: AppState, req: Parts) -> AppResult<Re
...
-    }
+async fn send_creation_email(
+    emails: &crate::Emails,
+    recipient: &str,
+    context: impl Serialize,
+) -> anyhow::Result<()> {
+    let email = EmailMessage::from_template("new_token", context);
+    let email = email.context("Failed to render email template")?;
+    let result = emails.send(recipient, email).await;
+    result.context("Failed to send email")
```
eth3lbert, line 300: Nit: use chaining, as is done with others.
```suggestion
    let email = EmailMessage::from_template("new_token", context)
        .context("Failed to render email template")?;
    emails
        .send(recipient, email)
        .await
        .context("Failed to send email")
```

Turbo87:
```suggestion
    let email = EmailMessage::from_template("new_token", context)
        .context("Failed to render email template")?;
    emails.send(recipient, email).await
        .context("Failed to send email")
```

I'd be okay with this formatting, but rustfmt seems very keen on giving `.await` its own line, which makes it look a bit ugly IMHO 😅

eth3lbert: maybe change other chained statements in other files to separate statements? I'm fine if they all follow the same style.

eth3lbert: These are all just nitpicks, not blockers, so feel free to merge if you want! :D

LawnGnome:
I defer to @eth3lbert's excellent review on the details.

I definitely think that moving to a templating engine is the right call, but I'm not so sure I like the loss of the explicit input fields for each template, both from the perspective of type safety and for knowing what might be interpolated into the e-mails. That said, I don't know that I have a great alternative: the thought of having a proc macro or something to ensure that templates and contexts are in sync feels... terrifying, and well beyond the scope of what we should do here, `minijinja` doesn't really give us a lot of alternatives, and I don't like any of the `minijinja` alternatives more than `minijinja` itself.

So, :+1:, but I also think we should remain open to continuing to iterate on the details over time.

Turbo87:
@LawnGnome yeah, I understand the concern. technically `minijinja` works with anything that is `Serializable` and `context!()` just happens to be one such things. we could also go back to explicit structs, but that still wouldn't solve the problem of the template referring to fields that don't exist. I know there are some template engines that check this at compile time, but IMHO that seems like overkill for this use case.

I guess we can try it out like this and see whether it is sufficient or not. If we see any issues with this approach we can always go back to explicit structs as the template inputs and/or improve our test suite for the templates.

PR #11422, opened 2025-06-24
Add `Rebuild Docs` action button

eth3lbert:
This PR implements the frontend part of #11169, which adds a `Rebuild Docs` action button to allow owners to enqueue the rebuild task.

cc @syphar

### DIFF
--- a/app/models/version.js
+++ b/app/models/version.js
@@ -200,4 +200,8 @@ export default class Version extends Model {
     this.store.pushPayload(payload);
     await waitForPromise(this.releaseTracks.refreshTask.perform(this.crateName, false));
   });
+
+  async rebuildDocs() {
+    return await waitForPromise(apiAction(this, { method: 'POST', path: 'rebuild_docs' }));
+  }
 }
--- a/app/components/version-list/row.hbs
+++ b/app/components/version-list/row.hbs
@@ -128,6 +128,18 @@
         <menu.Item>
           <YankButton @version={{@version}} class="button-reset" local-class="menu-button" />
         </menu.Item>
+        <menu.Item>
+          <button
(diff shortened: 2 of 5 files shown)


app/models/version.js
```diff
@@ -200,4 +201,16 @@ export default class Version extends Model {
     this.store.pushPayload(payload);
     await waitForPromise(this.releaseTracks.refreshTask.perform(this.crateName, false));
   });
+
+  rebuildDocsTask = keepLatestTask(async () => {
+    try {
+      await waitForPromise(apiAction(this, { method: 'POST', path: 'rebuild_docs' }));
+      this.notifications.success('Docs rebuild task was enqueued successfully!');
```
Turbo87: can you move the notification code into the component itself? the model layer shouldn't be aware of any user interface concerns :)

app/models/version.js
```diff
@@ -200,4 +201,16 @@ export default class Version extends Model {
...
+    } catch (error) {
+      console.error(error);
```
Turbo87:
```suggestion
```

app/models/version.js
eth3lbert: yeah, indeed!

syphar:
@Turbo87 we shouldn't forget to set the secret before deploying this

Turbo87:
good point! let me DM you 😉

Turbo87:
the docs.rs API token is now set on both staging.crates.io and crates.io

Turbo87:
I've manually added a docs.rs rebuild background job for my flarmnet crate to our queue and it seems like it worked:

Bildschirmfoto 2025-06-25 um 12 15 29

PR #11436, opened 2025-06-26
Add OpenGraph image generation crate

Turbo87:
This PR adds a new `crates_io_og_image` crate that generates OpenGraph images for crates.io packages using the Typst typesetting system.

### Features
- Generates 1200x630px images with crate metadata (name, description, version, license, tags, metrics)
- Downloads and renders author avatars from URLs
- Handles text truncation and overflow for long content
- Uses Typst typesetting system for layout and rendering
- Uses snapshot tests for visual regression detection

crates_io_og_image__tests__generated_og_image snap

(more screenshots available in the snapshots folder in the diff view 😉)

The generated images are used for social media previews when crates.io links are shared.

Note that the crate is not integrated into the main crates.io application yet. This will be done in a follow-up pull request.


Turbo87:
hmmmmm..... looks like the PNG snapshot tests aren't quite as portable as I had hoped... 😅

Turbo87:
> hmmmmm..... looks like the PNG snapshot tests aren't quite as portable as I had hoped... 😅

I take that back. The issue was caused by the CI machines not having the Fira Sans font pre-installed. I've added that to the CI workflow now and it seems that CI is happy with it 🎉

saharshxyz:
Great new edition using modern tooling – love to see it! Having fun talking about this in the Typst discord :)

.github/workflows/ci.yml
```diff
@@ -167,6 +169,26 @@ jobs:
...
+          typst --version
+
+      - name: Install Fira Sans font
+        run: |
+          # Download and install Fira Sans
+          wget -q "https://github.com/mozilla/Fira/archive/4.202.zip"
+          unzip -q "4.202.zip"
+          sudo mkdir -p /usr/local/share/fonts/fira-sans
+          sudo cp Fira-4.202/otf/FiraSans-*.otf /usr/local/share/fonts/fira-sans/
+          sudo fc-cache -f -v
```
NiklasEi: FYI: Instead of installing the font, you could also give the CLI a path to it:
https://github.com/typst/typst/blob/e8ce894ee7317654a0de1c822d14b3e169dde4d9/crates/typst-cli/src/args.rs#L360

You might be able to save quite some time by telling the CLI to not search for system fonts if FiraSans is all you need.

Turbo87: huh, good to know, thanks! :)

Turbo87: I've spend about an hour trying to get this to work on CI, but unfortunately it keeps failing. Since the system fonts approach works well enough for us, I guess this is fine for now.

Turbo87: turns out I was running into another problem and didn't notice it. I've now reintroduced support for setting a `TYPST_FONT_PATH` env var to switch from system font discovery to using `--font-path` 🎉

NiklasEi:
Typst's PNG export is (understandably) optimized for speed and not file size. If I understand the plan correctly, the images will be rendered once and then cached. So it would probably be worth it to run something like `oxipng` over the results, trading a bit more time for a reduction in file size.

<details>

<summary>A quick test with the output of the example from this PR => 56% smaller file size.</summary>

```
time oxipng -o 2 --strip safe test_og_image.png 
Processing: test_og_image.png
55478 bytes (56.00% smaller): test_og_image.png

real	0m0.136s
user	0m0.130s
sys	0m0.042s
```

```
time oxipng -o 4 --strip safe test_og_image.png 
Processing: test_og_image.png
53965 bytes (57.20% smaller): test_og_image.png

real	0m0.325s
user	0m0.948s
sys	0m0.028s
```

</details>


_I hope this is the right place for a comment like that. I really like the idea of this PR!_

Turbo87:
> I hope this is the right place for a comment like that

absolutely. that's a great idea, thanks! :)

Turbo87:
> So it would probably be worth it to run something like `oxipng` over the results, trading a bit more time for a reduction in file size.

`oxipng` is now integrated. these are the results from a quick analysis of our options:

### Regular Image (111,800 bytes)
| Tool | Size (bytes) | Reduction | Processing Time | Quality |
|------|-------------|-----------|-----------------|---------|
| **Original** | 111,800 | - | - | Lossless |
| **optipng** | 49,823 | 55.4% | ~0.30s | Lossless |
| **oxipng** | 46,627 | 58.3% | ~0.16s | Lossless |
| **pngquant** | 17,648 | **84.2%** | ~0.09s | Near-lossless |

### Example Image (135,109 bytes)
| Tool | Size (bytes) | Reduction | Processing Time | Quality |
|------|-------------|-----------|-----------------|---------|
| **Original** | 135,109 | - | - | Lossless |
| **optipng** | 59,231 | 56.2% | ~0.33s | Lossless |
| **oxipng** | 55,364 | 59.0% | ~0.15s | Lossless |
| **pngquant** | 22,841 | **83.1%** | ~0.07s | Near-lossless |

### Overflow Image (284,544 bytes)
| Tool | Size (bytes) | Reduction | Processing Time | Quality |
|------|-------------|-----------|-----------------|---------|
| **Original** | 284,544 | - | - | Lossless |
| **optipng** | 101,764 | 64.2% | ~0.45s | Lossless |
| **oxipng** | 92,863 | 67.4% | ~0.19s | Lossless |
| **pngquant** | 43,194 | **84.8%** | ~0.09s | Near-lossless |


I tried out `pngquant`, but it caused quite some graining on the avatar images, so I went with oxipng instead. 

Similar to Typst, we are using the oxipng CLI to avoid the compile time increase.

Turbo87:
@rust-lang/crates-io unless anyone voices any objections, I plan on merging this PR tomorrow and then work on integrating the new crate into the rest of crates.io

PR #11441, opened 2025-06-26
Add support for rendering alerts in README

eth3lbert:
This PR continues the work on #10429 and makes it functional by adding `allowed_classes`. This ensures that the generated alert blocks won't be cleaned by the sanitizer. Tests have also been added to confirm everything works as expected.

Thanks, @kbdharun, for the work on #10429.

preview:
image
image


eth3lbert:
Currently, alert icon support is handled via pure CSS for the following reasons:

- `Comrak` does not currently insert alert icon nodes.
- While we could somehow insert svg alert icon with a custom renderer, our sanitizer will remove SVG. :/ We could still use this method to inject a placeholder node and load icon via CSS, though. (the point would be: how important is it to have a placeholder node for the alert icon in the rendered HTML?)

There's also an alternative solution which doesn't require serving the SVG icon via CSS. This might require parsing the sanitized document and inserting the SVG into it.

crates/crates_io_markdown/lib.rs
```diff
@@ -429,7 +443,7 @@ mod tests {
     #[test]
     fn text_with_forbidden_class_attribute() {
         let text = "<p class='bad-class'>Hello World!</p>";
-        assert_snapshot!(markdown_to_html(text, None, ""), @"<p>Hello World!</p>");
+        assert_snapshot!(markdown_to_html(text, None, ""), @r#"<p class="">Hello World!</p>"#);
```
eth3lbert, line 446: This seems to be caused by the `allowed_classes`. I'm not sure if there's any way to get around it.

Turbo87:
LGTM, though in the screenshot it looks like there is a lot of padding at the bottom of the alert. is that intentional? I couldn't find anything in the CSS that would explicitly set it like this. 🤔

eth3lbert:
> LGTM, though in the screenshot it looks like there is a lot of padding at the bottom of the alert. is that intentional? I couldn't find anything in the CSS that would explicitly set it like this. 🤔

Ah, good catch! It shouldn't have a `margin-bottom` for the last child.

eth3lbert:
Thanks for the review! All the feedback should be addressed now. (I also rebased to the latest) :D

PR #11453, opened 2025-06-27
Add SLoC (Source Lines of Code) metric to versions

Turbo87:
This PR introduces basic source code analysis for newly published versions. A new `crates_io_linecount` workspace crate uses the [`tokei`](https://github.com/XAMPPRocky/tokei) crate to analyze source files during the publish process. The system collects language breakdowns and line count statistics, storing them as JSON in a new `linecounts` column on the `versions` table.

~The analysis runs during tarball processing and excludes test directories and non-programming files. All existing functionality remains unchanged, with the new column being optional for backward compatibility.~

**Update:** The analysis has been moved to a dedicated background job that is triggered by the publish flow or manually via the `crates-admin` tool.

Note that this is only the first step in a series of pull requests. The follow-up PRs will:
- ~implement a background job to backfill the existing versions~
- adjust the API responses to expose the data
- show the total SLoC count in the crate sidebar of the website
- adjust the OpenGraph images to show the SLoC count

### DIFF
--- a/src/bin/crates-admin/analyze_crates.rs
+++ b/src/bin/crates-admin/analyze_crates.rs
@@ -1,0 +1,152 @@
+use anyhow::Context;
+use crates_io::db;
+use crates_io::schema::background_jobs;
+use crates_io::schema::{crates, default_versions, versions};
+use crates_io::worker::jobs::AnalyzeCrateFile;
+use crates_io_worker::BackgroundJob;
+use diesel::prelude::*;
+use diesel_async::{AsyncPgConnection, RunQueryDsl};
+use indicatif::{ProgressBar, ProgressStyle};
+
+const CHUNK_SIZE: usize = 100;
+
+#[derive(clap::Parser, Debug, Eq, PartialEq)]
+#[clap(
+    name = "analyze-crates",
+    about = "Queue background jobs to analyze uploaded crate file.",
+    group(clap::ArgGroup::new("mode").required(true))
+)]
+pub struct Options {
+    /// Backfill *all* versions that are missing line count statistics.
+    #[clap(long, group = "mode")]
+    backfill: bool,
+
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):
No specific concerns on the implementation here.

I do wonder a little if we want to do this synchronously during publish. I don't expect line count calculation to be terribly expensive in the common case, but to defend against potential pathological cases, I might feel better about this if it was a background job. Not sure if anyone else on @rust-lang/crates-io has strong feelings here.

Turbo87:
I ran a quick benchmark of the `process_tarball()` fn using the `mozjs_sys-0.67.1.crate` file:

### Before

```
tarball_processing/process_mozjs_tarball
                        time:   [365.17 ms 368.25 ms 373.35 ms]
```

### After

```
tarball_processing/process_mozjs_tarball
                        time:   [990.60 ms 1.0022 s 1.0289 s]
                        change: [+173.34% +177.54% +181.29%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Admittedly, that is a more severe performance impact than I had anticipated. It might still be small compared to the time it takes to upload the crate file to our server, but still quite significant. I guess going with a background job makes more sense in this case, even though it requires us to download the crate file from the CDN again right after uploading it.

Turbo87:
@LawnGnome thanks for nudging this in the right direction. I've changed the implementation to run in a background job instead which should resolve the performance concerns.

@rust-lang/crates-io unless there are any objections I plan on merging this towards the end of the week.

epage:
I'm curious, in what way are you looking at using this or do places that inspried this use this?  I'm not seeing any linked issue or comment to provide the bigger picture.

Turbo87:
the main usages are for spam detection and nice-to-have metadata display

PR #11455, opened 2025-06-27
Added crate ID for background jobs

### LINKED ISSUE #11349: Background jobs are referencing crates/versions by name instead of ID
### Current Behavior

Some of our background jobs (e.g. `sync_crate_feed`) are referencing crates (and sometimes versions) by names. This can cause issues when crates are deleted and later reregistered while a background job is still in the queue.

### Expected Behavior

Background jobs should use crate/version IDs instead of names to avoid these issues, since the IDs will always be unique.

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

This is not much of an issue in practice since regular crate deletions block re-registrations for 24 hours and within that timeframe the background jobs should have finished. It's still worth fixing though...

YohDeadfall:
Should the jobs store crate names for logging purposes? Or should they just show the provided ID instead?

### DIFF
--- a/src/bin/crates-admin/delete_crate.rs
+++ b/src/bin/crates-admin/delete_crate.rs
@@ -83,38 +83,50 @@ pub async fn run(opts: Opts) -> anyhow::Result<()> {
     let available_at = now + chrono::TimeDelta::hours(opts.availability_delay);
 
     for name in &crate_names {
-        if let Some(crate_info) = existing_crates.iter().find(|info| info.name == *name) {
-            let id = crate_info.id;
+        let crate_id =
+            if let Some(crate_info) = existing_crates.iter().find(|info| info.name == *name) {
+                let id = crate_info.id;
 
-            let deleted_crate = NewDeletedCrate::builder(name)
-                .created_at(&crate_info.created_at)
-                .deleted_at(&now)
-                .deleted_by(deleted_by.id)
-                .maybe_message(opts.message.as_deref())
-                .available_at(&available_at)
-                .build();
+                let deleted_crate = NewDeletedCrate::builder(name)
+                    .created_at(&crate_info.created_at)
+                    .deleted_at(&now)
+                    .deleted_by(deleted_by.id)
+                    .maybe_message(opts.message.as_deref())
+                    .available_at(&available_at)
+                    .build();
 
-            info!("{name}: Deleting crate from the database…");
-            let result = conn
-                .transaction(|conn| delete_from_database(conn, id, deleted_crate).scope_boxed())
-                .await;
+                info!("{name}: Deleting crate from the database…");
+                let result = conn
+                    .transaction(|conn| delete_from_database(conn, id, deleted_crate).scope_boxed())
+                    .await;
 
-            if let Err(error) = result {
-                warn!(%id, "{name}: Failed to delete crate from the database: {error}");
+                if let Err(error) = result {
+                    warn!(%id, "{name}: Failed to delete crate from the database: {error}");
+                };
+
+                Some(id)
+            } else {
+                info!("{name}: Skipped missing crate");
+
+                None
             };
-        } else {
-            info!("{name}: Skipped missing crate");
-        };
 
         info!("{name}: Enqueuing background jobs…");
         let git_index_job = jobs::SyncToGitIndex::new(name);
         let sparse_index_job = jobs::SyncToSparseIndex::new(name);
-        let delete_from_storage_job = jobs::DeleteCrateFromStorage::new(name.into());
+        let delete_from_storage_job = async |conn: &mut AsyncPgConnection| {
+            if let Some(crate_id) = crate_id {
+                jobs::DeleteCrateFromStorage::new(name.into(), crate_id)
+                    .enqueue(conn)
+                    .await?;
+            }
+            Ok(())
+        };
 
         if let Err(error) = tokio::try_join!(
             git_index_job.enqueue(&mut conn),
             sparse_index_job.enqueue(&mut conn),
-            delete_from_storage_job.enqueue(&mut conn),
+            delete_from_storage_job(&mut conn),
         ) {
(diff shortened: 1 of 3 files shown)


src/worker/jobs/rss/sync_crate_feed.rs
```diff
@@ -43,8 +47,13 @@ impl BackgroundJob for SyncCrateFeed {
...
         info!("Loading latest {NUM_ITEMS} version updates for `{name}` from the database…");
         let mut conn = ctx.deadpool.get().await?;
+        let crate_id = if let Some(crate_id) = self.crate_id {
+            crate_id
+        } else {
+            load_crate_id(name, &mut conn).await?
+        };
 
-        let version_updates = load_version_updates(name, &mut conn).await?;
+        let version_updates = load_version_updates(crate_id, &mut conn).await?;
```
Turbo87: it's probably easiest to keep using `name` for now and migrate over to `crate_id` in the next PR when this PR here has been deployed to production. That way we won't need the `load_crate_id()` fn at all :)

Turbo87 (commented):
> Fixes https://github.com/rust-lang/crates.io/issues/11349

I will remove that part of the issue description, since a) this is only covering one of the problematic jobs, and b) this job will also need a follow-up PR to migrate away from `Option<i32>` to `i32`.

YohDeadfall:
> I will remove that part of the issue description

That's why it's only a draft. I have some questions and would like to deal with them first on a single job, and then do the same for the rest.

src/worker/jobs/rss/sync_crate_feed.rs
YohDeadfall: So, the first step is only for adding identifiers to the jobs without changing how they work, right?

Turbo87: yeah, exactly. that should make it easier to migrate them.

YohDeadfall:
A side question. Does it makes sense to provide configuration files for the analyzer as it's done in Miri and Rust itself? An example: https://github.com/rust-lang/miri/commit/d835aee60770c0a41582a37c2623e2ae53a6da26.

Turbo87:
> Does it makes sense to provide configuration files for the analyzer as it's done in Miri and Rust itself?

idk, I use the JetBrains IDEs, so I don't use rust-analyzer 😅

YohDeadfall:
If I'm not wrong then it's still rust-analyzer based, but it can be provided by the IDE itself. Anyway, we can live without adding these configs as no one complained except me, and I fixed it locally.

YohDeadfall:
So, `DeleteCrateFromStorage` uses name only as it never fetches something by it. Meanwhile, it's created when there's no crate exists:

https://github.com/rust-lang/crates.io/blob/77226386100054fc95e9147521e0942bf33d290b/src/bin/crates-admin/delete_crate.rs#L106-L112

Is it a bug or a feature?

Turbo87:
> `DeleteCrateFromStorage` uses name only as it never fetches something by it.

hmm, I guess in that case we should probably keep using the name since we can't delete by ID anyway.

> Meanwhile, it's created when there's no crate exists. Is it a bug or a feature?

that is actually intentional. we had a couple of cases in the past where a crate was deleted from the database by our support team but wasn't removed from S3 storage. this implementation allowed us to remove it from storage even though it didn't exist in the database anymore.

rust-toolchain.toml
```diff
@@ -1,2 +1,3 @@
 [toolchain]
 channel = "1.88.0"
+components = ["rust-analyzer"]
```
YohDeadfall: That's unintentional and will be removed. It fixes the issue with helix at least.

Turbo87:
I thought a bit more about #11349. It looks like we currently have six background jobs that use the crate name:

  1. DeleteCrateFromStorage - /src/worker/jobs/delete_crate.rs:14
  2. DocsRsQueueRebuild - /src/worker/jobs/docs_rs_queue_rebuild.rs:12
  3. SyncToGitIndex - /src/worker/jobs/index/sync.rs:16
  4. SyncToSparseIndex - /src/worker/jobs/index/sync.rs:84
  5. SyncCrateFeed - /src/worker/jobs/rss/sync_crate_feed.rs:25
  6. CheckTyposquat - /src/worker/jobs/typosquat.rs:20

These can be split into two categories: idempotent and non-idempotent tasks:

- `SyncToGitIndex`, `SyncToSparseIndex` and `SyncCrateFeed` are idempotent. They just sync what is saved in the database with the git index and S3, and it doesn't matter how often you run them. If they run on a crate that has been deleted and then recreated, they will correctly update to use the new crate.
- `DeleteCrateFromStorage`, `DocsRsQueueRebuild` and `CheckTyposquat` are one-off tasks that perform an action. Since they all work directly with the crate name and don't use the database, I wonder if adding the crate ID to them actually gives us any benefits.

Out of these jobs the `DeleteCrateFromStorage` is the only destructive one that could cause troubles if it would run with a significant delay. If, by the time the job runs, someone had already re-registered a crate, we would potentially delete the contents of the new crate. Unfortunately though, using an ID in this job is tricky since the crate files are saved with the crate name as folder name, not the ID. Since by default the crate names after deletion are only available at least 24 hours after the deletion this shouldn't be a big issue in practice though.

tl;dr out of these six jobs I'm not longer sure if any of them would actually benefit from using the crate ID instead of name 🙈 

Feel free to tell me I'm wrong though! :)

In any case, I'm sorry if I wasted your time by working on this issue. I should've thought about it a bit more before opening that issue... 😞

YohDeadfall:
> Feel free to tell me I'm wrong though! :)

Well, yo have an index for `name`, so it doesn't really matter if one uses `name` or `id` if we forget about performance for the moment. Even if we consider it, the difference won't be large enough to notice it.

> In any case, I'm sorry if I wasted your time by working on this issue. I should've thought about it a bit more before opening that issue... 😞

No worries, I'm doing it for fun and practice.

PR #11481, opened 2025-07-01
Integrate `crates_io_og_image` into the application

Turbo87:
This pull request integrates the `crates_io_og_image` crate (introduced in #11436) into the main crates.io application to enable automated OpenGraph image generation for crates. The integration includes adding the OG image generator to the worker environment, implementing background job processing for image generation, and extending storage functionality to handle OG image uploads. Images are automatically generated when crates are published or default versions are updated, with admin commands available for backfilling existing crates.


src/tests/util/test_app.rs
```diff
@@ -306,6 +307,8 @@ impl TestAppBuilder {
                 credentials: Credentials::Missing,
             };
 
+            let og_image_generator = Arc::new(OgImageGenerator::from_environment().unwrap());
```
Turbo87: This change basically requires everyone to have Typst and oxipng installed to run the tests since all publishing tests are now creating og-images. I'm wondering whether we should make this generator optional and default it to `None` in the test suite. That would also reduce the overhead of the image creation for most of the tests. 🤔

eth3lbert: > I'm wondering whether we should make this generator optional and default it to None in the test suite. 

yeah, I think it's a good idea. :D

Turbo87: implemented it like that this morning :)

Turbo87:
@rust-lang/crates-io unless anyone has objections I plan on merging this tomorrow

PR #11500, opened 2025-07-02
Skip running a job if the crate/version deleted

### LINKED ISSUE #11348: Background jobs are infinitely retried if crates are deleted
### Current Behavior

If a crate is deleted while a related background job is still in the queue, the job will eventually fail, but then get retried forever because the SQL query to find the corresponding crate will always fail.

### Expected Behavior

The job should exit early, without failure, but with a log warning.

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

YohDeadfall:
Fixes #11348.

After diffing in I found only 3 jobs that can fail because of deleted data. The rest should be fine.

### DIFF
--- a/src/worker/jobs/readmes.rs
+++ b/src/worker/jobs/readmes.rs
@@ -9,7 +9,7 @@ use diesel_async::AsyncConnection;
 use diesel_async::scoped_futures::ScopedFutureExt;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
-use tracing::{info, instrument};
+use tracing::{info, instrument, warn};
 
 #[derive(Clone, Serialize, Deserialize)]
 pub struct RenderAndUploadReadme {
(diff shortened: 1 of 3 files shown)


src/worker/jobs/readmes.rs
```diff
@@ -70,13 +70,31 @@ impl BackgroundJob for RenderAndUploadReadme {
...
+                                .await,
+                        ) {
+                        Ok(r) => r,
+                        Err(diesel::result::Error::DatabaseError(
+                            diesel::result::DatabaseErrorKind::ForeignKeyViolation,
+                            ..,
+                        ))
+                        | Err(diesel::result::Error::NotFound) => {
+                            warn!(
+                                "Skipping rendering README for vesion {}: no version found",
```
Turbo87:
```suggestion
                                "Skipping rendering README for version {}: no version found",
```

src/worker/jobs/update_default_version.rs
```diff
@@ -28,8 +28,16 @@ impl BackgroundJob for UpdateDefaultVersion {
...
         info!("Updating default version for crate {crate_id}");
         let mut conn = ctx.deadpool.get().await?;
-        update_default_version(crate_id, &mut conn).await?;
+        let res = update_default_version(crate_id, &mut conn).await;
+        if let Err(diesel::result::Error::DatabaseError(
+            diesel::result::DatabaseErrorKind::ForeignKeyViolation,
+            ..,
+        )) = res
+        {
+            warn!("Skipping update default version for crate for {crate_id}: no crate found",);
```
Turbo87:
```suggestion
            warn!("Skipping update default version for crate for {crate_id}: no crate found");
```

src/worker/jobs/send_publish_notifications.rs
```diff
@@ -39,7 +39,12 @@ impl BackgroundJob for SendPublishNotificationsJob {
         let mut conn = ctx.deadpool.get().await?;
 
         // Get crate name, version and other publish details
-        let publish_details = PublishDetails::for_version(version_id, &mut conn).await?;
+        let Some(publish_details) = PublishDetails::for_version(version_id, &mut conn).await?
+        else {
+            warn!("Skipping publish notifications for {version_id}: no version found",);
```
Turbo87:
```suggestion
            warn!("Skipping publish notifications for {version_id}: no version found");
```

Turbo87 (commented):
nice work! could you try to write a few regression tests for this too? idk if that is viable for the `RenderAndUploadReadme` job, but for the other two I think this should be possible.

YohDeadfall:
You could just ask not to use higher order functions and make it simple (:

PR #11503, opened 2025-07-02
Dedupe expired JTI deletion job

YohDeadfall:
It's a job without any parameters that always cleanups tokens expired to the moment of running. Therefore, there's no reason to enqueue another one if the existing hasn't been executed yet.


Turbo87 (approved):
good catch. not sure how I missed that 😅

PR #11506, opened 2025-07-03
db-dump: Reset sequences after data import

Turbo87:
This automatically resets all `id` column sequences to the maximum `id` column value of the corresponding table after the data was imported.

Closes https://github.com/rust-lang/crates.io/pull/10204

### DIFF
--- a/script/reset-sequences.sql
+++ b/script/reset-sequences.sql
@@ -1,0 +1,38 @@
+-- Reset all ID column sequences to their maximum values or 1
+-- This script dynamically discovers all sequences associated with 'id' columns
+-- and sets them to either the maximum id value in the table or 1 if empty
+
+DO $$
+DECLARE
+    rec RECORD;
+    max_id BIGINT;
+BEGIN
+    FOR rec IN
+        WITH id_sequences AS (
+            SELECT
+                pg_class.relname AS sequence_name,
+                pg_class_tables.relname AS table_name
+            FROM pg_class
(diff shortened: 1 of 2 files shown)


script/reset-sequences.sql
```diff
@@ -0,0 +1,35 @@
...
+            AND pg_attribute.attname = 'id'  -- only 'id' columns
+            AND pg_namespace.nspname = 'public' -- only public schema
+        )
+        SELECT * FROM id_sequences
+    LOOP
+        -- Get the maximum id value from the table
+        EXECUTE format('SELECT COALESCE(MAX(id), 1) FROM public.%I', rec.table_name) INTO max_id;
+
+        -- Reset the sequence to the max value
+        PERFORM setval('public.' || rec.sequence_name, max_id);
```
eth3lbert: FWIW, it might be worth commenting that the next value will be the set `max_id + 1`. This is because `setval` has two modes controlled by its optional third argument (which defaults to true), determining whether to add 1 to the next value.

eth3lbert (approved):
The SQL looks correct to me :D

script/reset-sequences.sql
Turbo87: I slightly changed it to explicitly use the third argument. Without this for empty tables the next value would have been `2` instead of `1`. I hope that makes it a bit clearer.

eth3lbert: ah yeah, you're right, thanks for the noticing!

PR #11508, opened 2025-07-03
Replace rebuild docs button with dedicated confirmation page

Turbo87:
Screen Shot 2025-07-03 at 19 04 39

Resolves https://github.com/rust-lang/crates.io/discussions/11499

### DIFF
--- a/app/routes/crate/rebuild-docs.js
+++ b/app/routes/crate/rebuild-docs.js
@@ -1,0 +1,34 @@
+import { service } from '@ember/service';
+
+import AuthenticatedRoute from '../-authenticated-route';
+
+export default class RebuildDocsRoute extends AuthenticatedRoute {
+  @service router;
+  @service session;
+  @service store;
+
+  async model(params) {
+    // Get the crate from parent route
+    let crate = this.modelFor('crate');
+
+    // Load the specific version
+    let version = await this.store.queryRecord('version', {
+      name: crate.id,
+      num: params.version_num,
+    });
+
+    return { crate, version };
+  }
+
+  async afterModel(model, transition) {
+    let user = this.session.currentUser;
+    let owners = await model.crate.owner_user;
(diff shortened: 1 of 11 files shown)


app/routes/crate/rebuild-docs.js
```diff
@@ -0,0 +1,34 @@
...
+    });
+
+    return { crate, version };
+  }
+
+  async afterModel(model, transition) {
+    let user = this.session.currentUser;
+    let owners = await model.crate.owner_user;
+    let isOwner = owners.some(owner => owner.id === user.id);
+    if (!isOwner) {
```
GuillaumeGomez, line 27: Couldn't we instead show the message on the page instead of redirecting the user?

Turbo87: as mentioned in the DMs, this is how the rest of crates.io behaves and I don't see a reason why we should do it differently for this route. note that the URL will stay the same and the user will see a "This page is only accessible by crate owners" message.

GuillaumeGomez: If URL is the same, then all good for me. :+1:

GuillaumeGomez:
Thanks for the feature!

eth3lbert (approved):
LGTM! I'm just not sure if we want the buttons to have a square style with slightly rounded corners, similar to those on the "Add a new Trusted Publisher" or "Report A Crate" pages, instead of the current fully-rounded buttons (not a blocker :D).

Turbo87:
> I'm just not sure if we want the buttons to have a square style with slightly rounded corners, similar to those on the "Add a new Trusted Publisher" or "Report A Crate" pages, instead of the current fully-rounded buttons (not a blocker :D).

I guess either is fine for now, but we should probably settle on one style for the whole page in the future to ensure a somewhat consistent UI 😅

PR #11513, opened 2025-07-04
Save and show JWT claims subset when Trusted Publishing is used

Turbo87:
For regular publishes we save and show the user that published the new version. For Trusted Publishing we currently don't do anything like that.

This PR proposes a new `versions.trustpub_data` JSONB column with a `provider` field and a subset of the relevant JWT claims. Specifically for GitHub Actions we save: `repository` (e.g. "rust-lang/crates.io"), the `run_id` (to be able to create URLs like https://github.com/rust-lang/crates.io/actions/runs/16069002415), and the `sha` (not visibly exposed yet, but could be quite useful in the future).

The frontend is then showing the data on the version list, roughly like we do for the user publishers:

Bildschirmfoto 2025-07-04 um 15 46 07

The "GitHub" label in this screenshot is clickable and leads to the corresponding job page on GitHub.

Similarly, the publish notification email template is adjusted to also show if versions were published by GitHub Actions via Trusted Publishing.

Note that this PR also adds a similar column to the `trustpub_tokens` table. At the time of publishing the JWT is no longer available, so we need to save the data during the token exchange and then copy it from one table to the other when crates are published with a Trusted Publishing token.

Related:

- https://github.com/bytecodealliance/wasm-component-ld/pull/69#issuecomment-3032604217


Turbo87:
looks like there are no objections, so let's get this merged and deployed

PR #11523, opened 2025-07-05
Migrate from `ember-css-modules` to `ember-scoped-css`

Turbo87:
The former isn't particularly well maintained anymore and blocks all sorts of other migration.

see https://github.com/soxhub/ember-scoped-css/


eth3lbert (approved):
I have to admit that I only inspected the deleted files and noticed they were just moved to a new folder and rewritten with CSS `@layer`. For the others, I only took a brief look, but given that the Percy visual snapshot doesn't show major differences, I believe this should all work well! Although some changes, like `browserlist` and `normalize.css`, could probably be split into separate commits, it's not really a big deal. And I love the change that we no longer need the `local-class`, we just need to place the CSS file beside the component which looks more natural. So, overall, LGTM 👍  Thanks for the work! :D

---

And don't forget, there's still a lock conflict before merging 😉

Issue #10387, opened 2025-01-14
Admin view for assessing whether crates are empty


carols10cents:
In dealing with reports of empty crates that are now against our policy, I'm now manually doing a bunch of clicking around and gathering info from a bunch of different sources, then either locking the account and deleting the crates if it appears the owner is only squatting, or emailing to see if the owner is still using the crates if the user has some legitimate usage of crates.io, according to our policies.

It would save me a ton of work if the info I needed for this task was on one-ish screen in the Prophesied Admin Interface of Dreams.

This page would be something like:

- crates.io/users/username/admin, so that I could still view the regular user page even when I'm in sudo mode
- display the user's verified email address so that I can email them if needed without needing to log in to the database
- list all crates, including those with all versions yanked (which the regular user page doesn't currently list), in a concise table format that contains the columns:
  - crate name
  - number of versions
  - publish date of most recent version
  - number of reverse dependencies (if the number is nonzero, link to the crate's `/reverse_dependencies` page so I can see if the dependencies are only this user's other crates)
  - link to `docs.rs/crate/cratename/latest/source/lib.rs` or `main.rs` so i can assess whether there's useful content or not
- sort by number of versions, low to high, then publish date earliest to latest
- ideally show all the user's crates without paging, realistically 50 on the page instead of 10 would probably be sufficient

just that would save a TON of time. bonus points:

- checkboxes that populate two markdown numbered list of crate names, one list of the checked names and one of the unchecked names. that is, if the table contains the crates foo, bar, and baz, and I tick the checkboxes for foo and baz, somewhere on the page are these two text lists I can copy:

```
1. foo
2. baz
```

```
1. bar
```

I'd like this functionality because in the emails I've been writing to people, I list out their empty crates that I want to know if they're still using, and I also list the crates of theirs that are fine and won't be deleted.

PR #11528, opened 2025-07-05
Add an API route for admins to assess crates

carols10cents:
Connects to #10387.

This API route, under `/private/` rather than `v1`, will make it a lot easier to assess whether crates contain useful functionality or are squatting.

This is only the backend; some of the functionality described in the issue will be frontend-only, but I wanted to get the API in first.

The API route will only return data if the currently authenticated user is an admin. This is important because the crate owner's verified email address is part of the returned data so that the admin can contact the owner if necessary.

Another reason this route is limited to admins is that some of the queries may be slow.

### DIFF
--- a/src/controllers/admin.rs
+++ b/src/controllers/admin.rs
@@ -1,0 +1,124 @@
+use crate::{
+    app::AppState,
+    auth::AuthCheck,
+    models::{CrateOwner, OwnerKind, User, Version},
+    schema::*,
+    util::errors::{AppResult, custom},
+};
+use axum::{Json, extract::Path};
+use chrono::{DateTime, Utc};
+use diesel::{dsl::count_star, prelude::*};
+use diesel_async::RunQueryDsl;
+use http::{StatusCode, request::Parts};
+use serde::Serialize;
+use std::collections::HashMap;
+
+/// Handles the `GET /api/private/admin_list/{username}` endpoint.
+pub async fn list(
+    state: AppState,
+    Path(username): Path<String>,
+    req: Parts,
+) -> AppResult<Json<AdminListResponse>> {
+    let mut conn = state.db_write().await?;
+
+    let auth = AuthCheck::default().check(&req, &mut conn).await?;
+    let logged_in_user = auth.user();
+
+    if !logged_in_user.is_admin {
+        return Err(custom(
+            StatusCode::FORBIDDEN,
+            "must be an admin to use this route",
+        ));
+    }
+
+    let (user, verified, email) = users::table
+        .left_join(emails::table)
+        .filter(users::gh_login.eq(username))
+        .select((
+            User::as_select(),
+            emails::verified.nullable(),
+            emails::email.nullable(),
+        ))
+        .first::<(User, Option<bool>, Option<String>)>(&mut conn)
+        .await?;
+
+    let crates: Vec<(i32, String, DateTime<Utc>, i64)> = CrateOwner::by_owner_kind(OwnerKind::User)
+        .inner_join(crates::table)
+        .filter(crate_owners::owner_id.eq(user.id))
+        .select((
+            crates::id,
+            crates::name,
+            crates::updated_at,
+            rev_deps_subquery(),
+        ))
+        .order(crates::name.asc())
+        .load(&mut conn)
+        .await?;
+
+    let crate_ids: Vec<_> = crates.iter().map(|(id, ..)| id).collect();
+
+    let versions: Vec<Version> = versions::table
+        .filter(versions::crate_id.eq_any(crate_ids))
+        .select(Version::as_select())
+        .load(&mut conn)
+        .await?;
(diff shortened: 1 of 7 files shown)


src/controllers/admin.rs
```diff
@@ -0,0 +1,124 @@
...
+use serde::Serialize;
+use std::collections::HashMap;
+
+/// Handles the `GET /api/private/admin_list/{username}` endpoint.
+pub async fn list(
+    state: AppState,
+    Path(username): Path<String>,
+    req: Parts,
+) -> AppResult<Json<AdminListResponse>> {
+    let mut conn = state.db_write().await?;
```
Turbo87: do we need the write access for this endpoint? I think it's read-only or am I missing something?

src/controllers/admin.rs
```diff
@@ -0,0 +1,124 @@
...
+            AdminCrateInfo {
+                name,
+                updated_at,
+                num_rev_deps,
+                num_versions: versions.map(|v| v.len()).unwrap_or(0),
+                crate_size: last_version.map(|v| v.crate_size).unwrap_or(0),
+                bin_names: last_version
+                    .map(|v| v.bin_names.clone())
+                    .unwrap_or_default(),
+            }
```
Turbo87: based on the SQL query I'm usually running in these situations, I recommend to also include:

- crate description
- default version (could probably be included in the `let crates` query)
- crate yanked or not (same as default_version yanked or not)
- total crate downloads

src/controllers/admin.rs
```diff
@@ -0,0 +1,124 @@
...
+        })
+        .collect();
+    Ok(Json(AdminListResponse {
+        user_email: verified.then_some(email).flatten(),
```
Turbo87: maybe we should have `user_email` and `user_email_verified` fields?

src/controllers/admin.rs
```diff
@@ -0,0 +1,124 @@
...
+    let versions: Vec<Version> = versions::table
+        .filter(versions::crate_id.eq_any(crate_ids))
+        .select(Version::as_select())
+        .load(&mut conn)
+        .await?;
+    let mut versions_by_crate_id: HashMap<i32, Vec<Version>> = HashMap::new();
+    for version in versions {
+        let crate_versions = versions_by_crate_id.entry(version.crate_id).or_default();
+        crate_versions.push(version);
+    }
```
Turbo87: if we use the default version instead of the last version then we only need the versions list for the "number of versions" field and that information is also available in the `default_versions` table. in other words it could be folded in the crates query above.

Turbo87 (approved):
LGTM, just a few minor comments :)

src/controllers/admin.rs
carols10cents: WHOOPS bad copypasta

src/controllers/admin.rs
carols10cents: > * crate yanked or not (same as default_version yanked or not)

Wouldn't this be better represented as "all versions yanked or not"? Or is the default version yanked iff all versions are yanked?

Turbo87: it is the same thing due to how we determine the default version. the default version can only be yanked if all versions are yanked, otherwise it would be chosen from the set of non-yanked versions.

carols10cents:
Ok I added a bunch of little commits that should hopefully be easy-ish to review one-by-one, and i think addressed everything! great suggestions, thank you!!

src/tests/routes/crates/admin.rs
```diff
@@ -0,0 +1,93 @@
...
+    assert_eq!(json_crate_0.num_rev_deps, 0);
+
+    let json_crate_1 = &json.crates[1];
+    assert_eq!(json_crate_1.name, "unyanked");
+    assert_eq!(json_crate_1.description.as_ref().unwrap(), "My Fun Crate");
+    assert_eq!(json_crate_1.downloads, 536);
+    assert_eq!(json_crate_1.num_versions, 3);
+    assert!(!json_crate_1.yanked);
+    assert_eq!(json_crate_1.default_version_num, "1.0.0");
+    assert_eq!(json_crate_1.num_rev_deps, 1);
```
Turbo87: any reason for going with individual assertions instead of `assert_json_snapshot!()`? I guess the `updated_at` field could make it slightly difficult, but with the redactions feature of insta it should be relatively straight-forward to redact that field from the response.

carols10cents: mostly because i added the fields one at a time, so i added the assertions one at a time 😅 switching to insta is a good idea, will do!

Turbo87: FWIW I usually use `serde_json::Value` instead of custom structs for the snapshot assertions to ensure that the assertions include all fields and not just the ones that happen to be in the struct.

carols10cents: Done!

PR #11531, opened 2025-07-06
og_image: Use binary prefixes for `format_bytes`

T0mstone:
Usage of SI prefixes to denote powers of 1024 is [incorrect](https://en.wikipedia.org/wiki/Binary_prefix):
> [T]he major standards organizations have expressly disapproved the use of SI prefixes to denote binary multiples, and recommended or mandated the use of the IEC prefixes for that purpose[.]

This just changes them to use the proper `kiB` and `MiB` instead of `kB` and `MB`.

An alternative would be changing the code instead, to use powers of 1000 like `format_number` does, but that might be a bit ambiguous precisely because this error is still so common.


T0mstone:
Oh, looking at the Wiki page again, I guess it should be `Ki` instead of `ki`.

T0mstone:
I don't understand why the tests are failing in CI but passing on my machine.

Turbo87:
> I don't understand why the tests are failing in CI but passing on my machine.

Might have been related to https://github.com/rust-lang/crates.io/pull/11530

I've rebased it and CI seems to be happy now

PR #11556, opened 2025-07-11
Update link and body colors to be WCAG AAA compliant

### LINKED ISSUE #1233: Enable color-contrast accessibility test and fix all violations
Once #1227 is merged in, address [this comment](https://github.com/rust-lang/crates.io/pull/1227/commits/ed4333ae06fbfe9ad769901eaffa070b9265430e#r160559295):

> There are multiple elements that currently fail this test. I have disabled it since I think it needs to be addressed by someone more inclined towards design.

by changing tests/axe-config.js to remove:

```
'color-contrast': {
            enabled: false,
}
```

run `npm test` and make changes to fix all violations where we're using colors that don't contrast enough.

kailan:
This PR updates the following colors to have a contrast ratio of at least 7:1 in the environments in which they are used:
- `main-color-light` is now slightly darker in light mode
- `link-color` and `link-hover-color` are now darker in light mode and lighter in dark mode

It also updates some colors throughout the application to reference these color variables where applicable.

Screenshot 2025-07-11 at 15 49 50

Screenshot 2025-07-11 at 15 51 56  

Look at those links pop!

---

On the category page, the entire row is now a hoverable and clickable link, so that there is no need to differentiate the title as a link (and thus the accessibility checker is happy)

Screenshot 2025-07-11 at 15 47 49

---

Listed dependencies and dependents now have increased contrast.

Screenshot 2025-07-11 at 16 36 28

I'm sure you can determine which one is before and after :)

---

I added underlines in places where differentiating the links from the text via color at a ratio that the checker is happy with is not feasible.

Screenshot 2025-07-11 at 16 14 18

---

Finally, the size of the tap targets for pagination has been slightly increased to pass a Lighthouse audit.  

Screenshot 2025-07-11 at 15 54 07

---
Tested with the built-in test suite `pnpm test` (which now all pass) as well as with Lighthouse in both dark and light mode.

Closes #1233.

### DIFF
--- a/app/components/dependency-list/row.hbs
+++ b/app/components/dependency-list/row.hbs
@@ -2,8 +2,8 @@
   local-class="
     row
     {{if @dependency.optional "optional"}}
-    {{if this.focused "focused"}}
   "
+  class="{{ if this.focused "focused" }}"
   data-test-dependency={{@dependency.crate_id}}
   ...attributes
 >
--- a/app/components/dependency-list/row.module.css
+++ b/app/components/dependency-list/row.module.css
@@ -1,33 +1,14 @@
 .row {
-    --bg-color: var(--grey200);
-    --hover-bg-color: light-dark(hsl(217, 37%, 98%), hsl(204, 3%, 11%));
-    --range-color: light-dark(var(--grey900), var(--grey400));
-    --crate-color: light-dark(var(--grey900), var(--grey400));
     --placeholder-opacity: 0.35;
-    --shadow: 0 1px 3px light-dark(hsla(51, 90%, 42%, .35), #232321);
+
+    composes: row from global;
 
     display: flex;
     align-items: center;
-    position: relative;
-    font-size: 18px;
-    padding: var(--space-s) var(--space-m);
-    background-color: light-dark(white, #141413);
-    border-radius: var(--space-3xs);
-    box-shadow: var(--shadow);
-    transition: all var(--transition-slow);
-
-    &:hover, &.focused {
-        background-color: var(--hover-bg-color);
-        transition: all var(--transition-instant);
-    }
-
-    &.focused {
-        box-shadow: 0 0 0 3px var(--yellow500), var(--shadow);
-    }
 
     &.optional {
-        --range-color: light-dark(var(--grey800), var(--grey600));
-        --crate-color: light-dark(var(--grey800), var(--grey600));
+        --row-bg: light-dark(#f6f6f6, #292726);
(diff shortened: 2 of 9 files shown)


Turbo87:
thanks for the PR!

due to the number of changes it's a bit hard to review/discuss them individually though. I like some of the changes, but others will probably need a discussion first 😅 

- Screenshot 1 (metadata title and icon): 👍 
- Screenshot 2 (link color): 👍 
- Screenshot 3 (categories): probably fine, but might have to check against other clickable lists that we have
- Screenshot 4 (dependencies): not sure about this one. the color is now very close to the color of regular dependencies.
- Screenshot 5 (underlined links): probably fine? it does add inconsistencies though. I'm wondering if we should do it everywhere instead 🤷‍♂️ 
- Screenshot 6 (pagination): 👍

kailan:
> thanks for the PR!

no problem! it was fun to get to know the codebase.

> due to the number of changes it's a bit hard to review/discuss them individually though. I like some of the changes, but others will probably need a discussion first 😅

Sorry about this. Happy to split into multiple PRs if that would be helpful.  
<br/><br/>
> * Screenshot 3 (categories): probably fine, but might have to check against other clickable lists that we have

I have created a reusable `row` class that now gets included in clickable lists I'm aware of (categories, versions, dependencies, dependents). That way, accessible styles can be applied uniformly across the site. (also means less CSS, yay!)

Screenshot 2025-07-16 at 15 42 33

Screenshot 2025-07-16 at 15 42 44

<br/>
Minor note: I updated the hardcoded `18px` font size value used in the row styles to `1.1rem` which roughly matches, but means that browser-configured font size will be respected.

<br/><br/>

> * Screenshot 4 (dependencies): not sure about this one. the color is now very close to the color of regular dependencies.

I've addressed this by changing the background color for optional dependencies, see below:

Screenshot 2025-07-16 at 15 41 35

Screenshot 2025-07-16 at 15 41 45  

<br/><br/>

> * Screenshot 5 (underlined links): probably fine? it does add inconsistencies though. I'm wondering if we should do it everywhere instead 🤷‍♂️

It certainly could be added everywhere, but I didn't want to rock the boat too much on my first contribution 😅  Let me know what you think.

Turbo87:
I just realized that #11523 caused quite a bit of conflict with this PR... sorry! 😞 

> Happy to split into multiple PRs if that would be helpful.

yeah, I guess that would probably make it easier to review, discuss and rebase them 😅 

> I've addressed this by changing the background color for optional dependencies, see below:

for the dark mode screenshot that seems okay, but the grey of light yellow in light mode seems a bit off

PR #11580, opened 2025-07-16
frontend: Redirect when space in crate name

Its-Just-Nans:
Hi

When there is leading or trailing space in the crate name, the crate is not found (obviously)

https://crates.io/crates/syn Working

https://crates.io/crates/%20syn Not found

I think this could be useful, mostly if we paste a crate name

### DIFF
--- a/app/routes/crate.js
+++ b/app/routes/crate.js
@@ -10,6 +10,11 @@ export default class CrateRoute extends Route {
 
   async model(params, transition) {
     let crateName = params.crate_id;
+    if (crateName.trim() != crateName) {
+      // If the crate name has leading/trailing whitespace, redirect to the trimmed version.
+      this.router.transitionTo('crate', crateName.trim());
+      return;
+    }
 
     try {
       // We would like the peeked crate to include information (such as keywords) for further


Turbo87:

LawnGnome:
I'm assuming the use case here is something like a keyword bookmark `https://crates.io/crates/%s`.

I can see why it would be useful (since I use keywords all the time), but I'm not sure it makes sense to complicate the frontend route handling to support this, since the search does ignore leading or trailing whitespace correctly. I'm probably -0.5 on changing this, but I also don't feel very strongly.

Its-Just-Nans:
On my side it was more a UX improvement

If I go to https://crates.io/crates/syn

image


But If I make a small typo (and a bad copy-paste) and go here 

https://crates.io/crates/%20syn

image


The message "syn: Crate not found" is not great

image


So either
- change to have a redirection
- add quote to the message


image

image


Both fixes are easy and seems OK
But I prefer the redirection since there is a lot probability that the user modify his typo

eth3lbert:
I'm okay with it. It doesn't overcomplicate things and brings some improvement.

PR #11585, opened 2025-07-16
Encrypt GitHub access tokens

Turbo87:
We currently save the GitHub OAuth access tokens in plaintext in our database, which isn't a huge issue as long as nobody unauthorized gains access to our database. But from a security standpoint it's better to encrypt them so that an attacker would need both access to the database **and** to the encryption key.

This PR implements step 1 of the migration towards using encrypted access tokens. It adds a new column to the database, changes the code to start saving encrypted tokens to that column and adds an admin tool to backfill the column based on the current plaintext tokens.

This also adds a new required `GITHUB_TOKEN_ENCRYPTION_KEY` environment variable without which the server and background worker won't start. The env var expects a 64 character hex string, which is then decoded to a 32 byte AES key. I've already generated corresponding values for our staging and production environments via `openssl rand -hex 32`.

The encryption uses "AES-256-GCM" which from my understanding should be a reasonable algorithm for this situation and requires no extra dependencies since it was already used transitively.

For completeness: Phase 2 will turn the column non-nullable and will start using the encrypted tokens instead of the plaintext tokens. Phase 3 will then remove the plaintext token column. These phases should be deployed separately for obvious reasons 😅

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -72,6 +72,10 @@ export GIT_REPO_URL=file://$PWD/tmp/index-bare
 export GH_CLIENT_ID=
 export GH_CLIENT_SECRET=
 
+# Key for encrypting/decrypting GitHub tokens. Must be exactly 64 hex characters.
+# Used for secure storage of GitHub tokens in the database.
+export GITHUB_TOKEN_ENCRYPTION_KEY=0af877502cf11413eaa64af985fe1f8ed250ac9168a3b2db7da52cd5cc6116a9
+
 # Credentials for configuring Mailgun. You can leave these commented out
 # if you are not interested in actually sending emails. If left empty,
 # a mock email will be sent to a file in your local '/tmp/' directory.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1348,6 +1348,7 @@ dependencies = [
 name = "crates_io"
 version = "0.0.0"
 dependencies = [
+ "aes-gcm",
  "anyhow",
  "astral-tokio-tar",
  "async-compression",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ name = "crates_io"
 doctest = true
 
 [dependencies]
+aes-gcm = { version = "=0.10.3", features = ["std"] }
 anyhow = "=1.0.98"
 astral-tokio-tar = "=0.5.2"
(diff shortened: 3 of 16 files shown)


mdtro:
Not something to block this PR, but we should think about what would be needed to support key rotation here in the future.

Using a KMS would make this easier, but if we want to keep this all localized there are some options.

1. Adjust the environment variable to support a list/dictionary.
2. Assign a random key ID (`kid`) to each key.
3. Embed or save the `kid` in a separate column used to encrypt the token in the database. It could be embedded into the "associated data" of the AEAD Payload (https://docs.rs/aead/0.5.2/aead/struct.Payload.html).
4. Annotate which key is the "active" one for new encryption operations.

Decryption operations would perform the lookup via the `kid`.
Encryption operations would always use the key that's marked as "active".

LawnGnome:
Without doing a full review, my gut reaction is the same as @mdtro's — I'd prefer to use a KMS for rotation and logging purposes. Not going to call it a blocker just yet, but I do want think about this some more.

@walterhpearce, you've been thinking about this kind of problem recently in the context of signing things. Any thoughts?

Turbo87:
I'm not opposed to using a KMS, but I have zero experience with it. You are both welcome to work on it and coordinate the deployment with the infra team though 😅 

I'm a bit worried that we're letting "perfect" be the enemy of "good" here (or whatever the saying is). I would assume that we could still retrofit a KMS later even with the current design. We also have a few other keys that should maybe be moved into a KMS, so this seems like a bigger task that shouldn't necessarily prevent this quick win proposed here.

Aside from that, the use of per-user GitHub access tokens might potentially turn out to not be viable for us anymore in the future. Once we support other identity providers then not every user might have a corresponding GitHub OAuth token anymore, in which case we might not even need to save them since we will need to use some other method anyway. In other words, I currently don't expect us to keep these around for the next ten years, so key rotation might not be quite as critical?

Turbo87:
so how do we continue here? should we ship this as an incremental improvement? or does one of you want to look into the KMS thing?

Turbo87:
unless anyone has any major objections I plan on moving forward with this early next week. AFAICT we can still retrofit key IDs and KMS usage later if we want it and this PR is certainly an improvement compared to saving the keys in plaintext in the database.

Issue #11597, opened 2025-07-18
Allow for more than one verified email address


carols10cents:
We currently only allow people to set and verify one email address.

Because we use verified email addresses for account recovery, it would increase the likelihood of successful recovery if people could add and verify more than one email address with their account.

Implementation considerations:

- Are there uniqueness constraints on the emails table that would need to be removed/changed?
- Allowing publish should be gated on "there exists at least one verified email address"
- When configuring whether you'd like to receive publish notification emails, you should be able to configure which email address the notifications go to

PR #11629, opened 2025-07-24
Allow users to add multiple email addresses to their account

kailan:
👋  Hello crates.io team,

This pull request updates the `emails` table to support multiple email addresses per user, as well as adding API endpoints for managing emails (`/api/v1/users/{user}/emails`), and the relevant UI changes to go with it, closing #11597.

Please see the below video for a demonstration:

https://github.com/user-attachments/assets/2297fd9a-e10d-4d4c-9106-2af6198f131f

## Example user response

```json
{
  "id": 1,
  "login": "arbitrary_username",
  "emails": [
    {
      "id": 1,
      "email": "foo@example.com",
      "verified": true,
      "verification_email_sent": true,
      "primary": true
    },
    {
      "id": 2,
      "email": "bar@example.com",
      "verified": false,
      "verification_email_sent": true,
      "primary": false
    }
  ],
  "name": null,
  "email_verified": true,
  "email_verification_sent": true,
  "email": "foo@example.com",
  "avatar": null,
  "url": "https://github.com/arbitrary_username",
  "is_admin": false,
  "publish_notifications": true
}
```

### API Stability

I have implemented this in a way that all existing API endpoints still work and previously-used properties are still available – specifically:
* PUT `/api/v1/users/{user}` will still respect the `email` parameter, adding the given email to the account. This endpoint is still needed for toggling publish notifications at the user level.
* GET `/api/v1/me` still returns `email`, `email_verified`, `email_verification_sent` – these correspond to the matching properties on the primary email. Internally these are now prefixed with `primary_` and marked as deprecated.

Is backwards compatibility expected here, or would it be okay to consider this a breaking change, removing the legacy behaviours and properties? The UI has been updated to not use them.

### Duplicate Emails

This implementation allows an email address to exist on more than one user account. Is this wanted or should there be a constraint to restrict this?

### GitHub Email Sync

Upon login, users are now prompted to also provide the `user:email` scope. This allows us to fetch _all_ emails from the user's GitHub account and add them to their crates.io account. This will also happen on subsequent logins in the case that a user has added more emails to their GitHub account.

Is this behaviour wanted? Will users be concerned about providing this extra permission?

> **NOTE:** The session controller will still work as it did before if fetching the extra emails from GitHub fails, in the event that the `user:email` scope is not granted, but I cannot see a way for a user to opt out of this while still going through the login process, despite GitHub documentation suggesting that this is possible: https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps#requested-scopes-and-granted-scopes

### Email Verification

With the extra `user:email` scope, the app can now determine if an email has been verified by GitHub. As long as the crates.io team are comfortable trusting that GitHub has verified the email correctly, this means that crates.io would no longer have to send a verification email, which reduces the friction to sign up and would also reduce the usage of the email provider.

If you'd prefer to verify them manually anyway, that change is straightforward to revert.

---

It was a pleasure to work on this and I look forward to collaborating with you to get this shipped. I have not previously worked with Ember so it is likely I have not followed best practices while implementing the UI elements. Please feel free to reach out with any questions or concerns!


kailan:
Going to work on resolving the failing tests.

migrations/2025-07-22-091706_multiple_emails/up.sql
```diff
@@ -0,0 +1,127 @@
...
+  IF OLD.send_notifications IS TRUE THEN
+    -- Allow deletion if this is the only email for the user
+    IF (SELECT COUNT(*) FROM emails WHERE user_id = OLD.user_id) = 1 THEN
+      RETURN OLD;
+    END IF;
+    RAISE EXCEPTION 'Cannot delete email: send_notifications is set to true';
+  END IF;
+  RETURN OLD;
+END;
+$$ LANGUAGE plpgsql;
```
Turbo87: why do we allow deletion if it's the only email address?

migrations/2025-07-22-091706_multiple_emails/up.sql
```diff
@@ -0,0 +1,127 @@
...
+CREATE TRIGGER trigger_enforce_max_emails_per_user
+BEFORE INSERT ON emails
+FOR EACH ROW
+EXECUTE FUNCTION enforce_max_emails_per_user();
+
+-- Add a unique constraint for the combination of user_id and email
+ALTER TABLE emails ADD CONSTRAINT unique_user_email UNIQUE (user_id, email);
+
+-- Add a new column for identifying if an email should receive notifications
+ALTER TABLE emails ADD COLUMN send_notifications BOOLEAN DEFAULT FALSE NOT NULL;
```
Turbo87: I'm wondering about the `send_notifications` naming, since it could be confused with the "please send me publish notifications" flag on the users table. GitHub is using "primary" naming, which might be harder to confuse 🤔

migrations/2025-07-22-091706_multiple_emails/up.sql
Turbo87, line 1: since this migration introduces a lot of new database constraints it might be worth it to write dedicated tests for them. in other words, set up a test database, trigger these constraints with regular diesel calls and then use snapshot assertions to check that the errors are returned as expected.

package.json
```diff
@@ -46,6 +46,7 @@
     "@sentry/ember": "9.40.0",
     "chart.js": "4.5.0",
     "date-fns": "4.1.0",
+    "ember-autofocus-modifier": "^7.0.1",
```
Turbo87: we already use `@zestia/ember-auto-focus`. I'm not opposed to switching if that one is better maintained, but that should be done in a dedicated PR

crates/crates_io_database/src/schema.rs
```diff
@@ -538,6 +538,12 @@ diesel::table! {
         ///
         /// (Automatically generated by Diesel.)
         token_generated_at -> Nullable<Timestamptz>,
+        /// The `send_notifications` column of the `emails` table.
+        ///
+        /// Its SQL type is `Bool`.
+        ///
+        /// (Automatically generated by Diesel.)
```
Turbo87: if you add a comment to the column in the migration then diesel will automatically use that for the Rust schema field comment too 😉

Turbo87 (commented):
The PR is rather large, which generally makes it a bit harder to review, so I would recommend splitting it up, which should also make the deployment a bit more resilient to any issues. 

My recommendation, I think, would be three (or four) steps: 
1. database changes, without affecting the API output
2. API changes, without affecting the frontend
3. frontend changes
4. (maybe the GitHub API changes should be a dedicated step?)

Then we can review and deploy these steps separately and address any potential issues before continuing with the next one.

I've focused the initial round of review on the database changes, since that would be step one of the deployment.

kailan:
Thanks very much for the feedback @Turbo87 – I'll address each point in this branch so I have everything working together, then I'll split it out into separate PRs for review.

now go enjoy your holiday!!

migrations/2025-07-22-091706_multiple_emails/up.sql
kailan: You can't delete a user row if there is an email row referencing it, so I did this to allow for users to be deleted.

Whether that's beneficial in development or production is a question to be answered I suppose.

Turbo87: hmm, I'm not that familiar with that constraint but that seems like a job for "on cascade delete" unless I'm missing something 🤔

migrations/2025-07-22-091706_multiple_emails/up.sql
kailan: This was a great idea because it identified some cases that were failing, thanks.

```
     Running tests/email_constraints.rs (target/debug/deps/email_constraints-7d177f18cbf0d518)

running 12 tests
test create_primary_email ... ok
test create_secondary_email ... ok
test create_primary_email_when_one_exists ... ok
test create_same_email_for_different_users ... ok
test create_secondary_email_without_primary ... ok
test create_secondary_email_with_same_email_as_primary ... ok
test delete_only_email ... ok
test create_too_many_emails ... ok
test delete_primary_email_with_secondary ... ok
test promote_secondary_to_primary ... ok
test demote_primary_without_new_primary ... ok
test delete_secondary_email ... ok
```

migrations/2025-07-22-091706_multiple_emails/up.sql
kailan: Yeah there is a constraint preventing users from being deleted that was introduced in https://github.com/rust-lang/crates.io/pull/1090

```
update or delete on table "users" violates foreign key constraint "fk_emails_user_id" on table "emails"
Detail: Key (id)=(2) is still referenced from table "emails".
```

We can discuss that when the schema changes are split into a separate PR.

kailan:
Closing in favour of smaller PRs.

PR #11642, opened 2025-07-26
Allow `emails` table to contain multiple emails per user

kailan:
This PR is split from #11629 – it contains just the database schema-related changes, with no changes to the API, frontend, or any user-visible behaviour, barring a change to the description of three properties of the `/api/v1/me` response in the OpenAPI schema.

I previously had a note here about it being possible for an email address to exist on more than one user account, but this is the current behaviour in production so I have removed it.

Relates to #11597.

### DIFF
--- a/crates/crates_io_database/tests/email_constraints.rs
+++ b/crates/crates_io_database/tests/email_constraints.rs
@@ -39,7 +39,7 @@ async fn insert_test_user(conn: &mut AsyncPgConnection) -> i32 {
         .name(&format!("testuser{}", user_count + 1))
         .gh_id(user_count as i32 + 1)
         .gh_login(&format!("testuser{}", user_count + 1))
-        .gh_access_token("token")
+        .gh_encrypted_token(&[])
         .build()
         .insert(conn)
         .await


crates/crates_io_database/src/schema.rs
```diff
@@ -538,6 +538,9 @@ diesel::table! {
         ///
         /// (Automatically generated by Diesel.)
         token_generated_at -> Nullable<Timestamptz>,
+        /// Whether this email is the primary email address for the user.
+        #[sql_name = "is_primary"]
```
kailan: `primary` is a reserved keyword in Postgres, so we use `is_primary` as the column name and then alias it to `primary` in the Diesel schema to be consistent with `verified`.

crates/crates_io_database/src/models/email.rs
```diff
@@ -24,46 +25,65 @@ pub struct NewEmail<'a> {
...
+        .returning(Email::as_returning())
+        .get_result(conn)
+        .await
+        .optional()?;
+
+        if let Some(email) = updated_email {
+            Ok(email)
+        } else {
+            // Otherwise, insert a new email
+            self.insert(conn).await
```
Turbo87, line 92: probably not worth the extra complexity fixing it, but worth mentioning: if `insert_or_update_primary()` is called on a `NewEmail`  with `primary: false`, then this code path would not add it as a primary email. the only call site of this method is currently setting `primary: true`, but I'm wondering whether we should at least `error!(...)` log that case for now.

crates/crates_io_database/src/schema.patch
```diff
@@ -45,6 +45,22 @@
...
+         /// Whether this email is the primary email address for the user.
+-        is_primary -> Bool,
++        #[sql_name = "is_primary"]
++        primary -> Bool,
+     }
+ }
+
+ diesel::table! {
+     /// Representation of the `follows` table.
+     ///
```
Turbo87: since the patch file is already a bit hard to maintain, I think it might be easier to just use `is_primary` directly in our code? 😅

migrations/2025-07-22-091706_multiple_emails/up.sql
```diff
@@ -0,0 +1,102 @@
...
+
+-- Add a unique constraint for the combination of user_id and email
+ALTER TABLE emails ADD CONSTRAINT unique_user_email UNIQUE (user_id, email);
+
+-- Add a new column for identifying the primary email
+ALTER TABLE emails ADD COLUMN is_primary BOOLEAN DEFAULT FALSE NOT NULL;
+comment on column emails.is_primary is 'Whether this email is the primary email address for the user.';
+
+-- Set `is_primary` to true for existing emails
+UPDATE emails SET is_primary = true;
```
Turbo87: we currently have 72k rows in the `emails` table. I'm a bit scared about running this as part of the migration, since the running migration will block the API server from restarting.

@eth3lbert any thoughts on this?

I guess we could extract the `add column is_primary` above to a dedicated migration/PR, deploy that to run the migration, then manually run the `update is_primary = true` without blocking the API server, and then merge/deploy the rest of this migration 🤔

migrations/2025-10-10-123902_multiple_emails/up.sql
```diff
@@ -0,0 +1,102 @@
...
+-- Set `is_primary` to true for existing emails
+UPDATE emails SET is_primary = true;
+
+-- Limit primary flag to one email per user
+-- Evaluation of the constraint is deferred to the end of the transaction to allow for replacement of the primary email
+CREATE EXTENSION IF NOT EXISTS btree_gist;
+ALTER TABLE emails ADD CONSTRAINT unique_primary_email_per_user
+EXCLUDE USING gist (
+  user_id WITH =,
+  (is_primary::int) WITH =
```
Turbo87: is this line needed? if we filter below using `WHERE is_primary` then I guess the above line would be redundant?

migrations/2025-10-10-123902_multiple_emails/up.sql
```diff
@@ -0,0 +1,102 @@
...
+  END IF;
+
+  RETURN COALESCE(NEW, OLD);
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trigger_verify_exactly_one_primary_email
+AFTER INSERT OR UPDATE ON emails
+FOR EACH ROW
+EXECUTE FUNCTION verify_exactly_one_primary_email();
```
Turbo87, line 66: isn't this redundant with the `unique_primary_email_per_user` constraint above? 🤔

migrations/2025-07-22-091706_multiple_emails/up.sql
eth3lbert: > I guess we could extract the `add column is_primary` above to a dedicated migration/PR, deploy that to run the migration, then manually run the `update is_primary = true` without blocking the API server, and then merge/deploy the rest of this migration 🤔

Yeah, I second this! I think this would be the most proper way to minimize potential issues caused by the introduced database migration.

kailan:
Haven't forgotten about this! Thanks for your review – aiming to address your comments soon.

migrations/2025-07-22-091706_multiple_emails/up.sql
kailan: I split this into a separate migration and will open a separate PR including just that.

migrations/2025-10-10-123902_multiple_emails/up.sql
kailan: Turns out this constraint was redundant so I've removed it entirely

migrations/2025-10-10-123902_multiple_emails/up.sql
kailan: This function actually covers more cases than the above `unique_primary_email_per_user`, so I've removed that in favour of keeping this around.

crates/crates_io_database/src/models/email.rs
kailan: I made it return an error. Let me know if you'd prefer to just log.

PR #11653, opened 2025-07-28
Handle team update scenario where GitHub team has been recreated

### LINKED ISSUE #6949: Deleting and renaming a GitHub team gives it a new GitHub ID, which causes uniqueness constraint violations
Related but slightly distinct:

- https://github.com/rust-lang/crates.io/issues/4281
- https://github.com/rust-lang/crates.io/issues/1818
- https://github.com/rust-lang/crates.io/issues/4252

To reproduce:

- Create a github team in an org and run `cargo owners --add github:org:team` to add the team as an owner of the crate.
- Delete the team in github and recreate it with the same name. This should give the team a new github ID.
- Attempt to add the team to a different crate, issuing the same `cargo owners --add github:org:team` command.
- Crates.io will return a uniqueness constraint violation because we enforce unique team names.

If we get the same ID but a different name, we upsert and update the name. But we aren't doing that for teams. I think as long as the *org's github id* is the same (that is, the entire organization hasn't been deleted and recreated), it isn't a security problem if we update the team ID when the names are the same (as it indicates the same organization has deleted and recreated the team).

kailan:
If a team owns a crate, that team is deleted from GitHub, then recreated with the same name, a conflict would occur when adding the team to another crate as a team with the same `login` already exists, but the `github_id` is different so an update is not triggered.

This PR modifies the `NewTeam::create_or_update` function to attempt an update of the team record matching the (`org_id`, `login`) pair, allowing existing team records to be updated with new GitHub team IDs.

It also adds a test for the above described scenario.

Resolves #6949.

### DIFF
--- a/src/tests/team.rs
+++ b/src/tests/team.rs
@@ -124,6 +124,61 @@ async fn add_renamed_team() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Test adding a team that has the same name as an existing team, but a different ID
+/// This would happen if a team was deleted in GitHub and then recreated with the same name.
+#[tokio::test(flavor = "multi_thread")]
+async fn add_team_with_same_name_different_id() -> anyhow::Result<()> {
+    let (app, anon) = TestApp::init().empty().await;
+    let mut conn = app.db_conn().await;
+    let user = app.db_new_user("user-all-teams").await;
+    let token = user.db_new_token("arbitrary token name").await;
+    let owner_id = user.as_model().id;
+
+    CrateBuilder::new("foo_same_name_different_id", owner_id)
+        .expect_build(&mut conn)
+        .await;
+
+    let original_team = NewTeam::builder()
+        .login("github:test-org:core")
+        .org_id(1000)
+        .github_id(2001)
+        .build()
+        .create_or_update(&mut conn)
+        .await?;
+
+    let new_team = NewTeam::builder()
+        // same team name
+        .login(&original_team.login)
+        // same org ID
+        .org_id(original_team.org_id)
+        // different team ID
+        .github_id(original_team.github_id + 1)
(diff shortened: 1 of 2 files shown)


kailan:
ahhh....

@rustbot label "A-backend ⚙️" "C-bug 🐞"

Turbo87 (commented):
This seems like a good incremental improvement, thanks! 👍 

For a bit more context: the way teams currently work is a bit problematic in various way. One is that they are somewhat tightly coupled to GitHub, which makes it hard(er) for us to support other identity providers. IMHO in the long term we should move team management into crates.io itself to avoid all of these issues, potentially with some functionality to sync GitHub team memberships with crates.io teams.

src/tests/team.rs
```diff
@@ -124,6 +124,61 @@ async fn add_renamed_team() -> anyhow::Result<()> {
...
+    let new_team = NewTeam::builder()
+        // same team name
+        .login(&original_team.login)
+        // same org ID
+        .org_id(original_team.org_id)
+        // different team ID
+        .github_id(original_team.github_id + 1)
+        .build()
+        .create_or_update(&mut conn)
+        .await?;
```
Turbo87, line 158: I'm a bit worried about these manual database mutations in the test, since they don't quite guarantee that the same logic is applied as it would through the API. I think the issue might be that this is mixing unit testing of the `create_or_update()` fn with integration testing of the API endpoint behavior.

I would suggest putting the unit test(s) into the database crate via `mod tests` and then adjusting the test here to only use the API endpoints. It might also make sense to use a custom GitHub API mock for this test so that we can better control the returned IDs.

We should probably also explicitly test the case where the org IDs do not match to avoid accidentally enabling resurrection attacks.

PR #11697, opened 2025-08-04
Use encrypted GitHub OAuth access tokens

Turbo87:
This PR implements the second phase described in https://github.com/rust-lang/crates.io/pull/11585. It makes the `gh_encrypted_token` column non-nullable and adjusts our code to use the encrypted tokens instead of the plaintext ones.

The PR also removes the now obsolete `encrypt-github-tokens` admin tool again, since the backfill has been performed on both environments and since it was filtering on `gh_encrypted_token IS NULL`, it wasn't compatible with the migration anymore.


Turbo87:
@rust-lang/crates-io unless anyone has objections, I plan on merging and deploying this on Thursday

PR #11701, opened 2025-08-05
add proxy-server as a subcategory of web-programming

GlenDC:
This change adds a new proxies subcategory under web-programming to better classify crates focused on building or managing various types of proxies (such as HTTP, HTTPS, SOCKS5, transparent, forward, and reverse proxies).

I believe this addition is justified because proxy-related functionality is distinct from existing http-client and http-server categories. While it is closely related to web protocols, proxy crates serve unique use cases like traffic routing, interception, anonymization, or load balancing. These use cases span from development tools to production-grade networking infrastructure, and a dedicated category will help crate authors and users better discover relevant tooling.

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -635,3 +635,11 @@ name = "WebSocket"
 description = """
 Crates to communicate over the WebSocket protocol.\
 """
+
+[web-programming.categories.proxy-server]
+name = "Proxy server"
+description = """
+Crates for building and managing proxies,\
+including HTTP, HTTPS, SOCKS5, transparent, reverse, and forward proxies.\
+Useful for traffic inspection, anonymization, load balancing, and network routing in web applications.\
+"""


Turbo87:
I'm a bit skeptical whether there are enough crates to justify such a category vs. just using the "web-programming" parent category.

GlenDC:
> I'm a bit skeptical whether there are enough crates to justify such a category vs. just using the "web-programming" parent category.
> 
> @rust-lang/crates-io any thoughts?

If we label anything under web programming you might as well get rid of http client and http server... I on purpose made it just `proxy-server` to make it an umbrella for all proxies. Rather than doing `http-proxy`, `socks5-proxy`, `mitm-proxy`, etc...

There are almost 7k crates in https://crates.io/categories/web-programming ... No way anybody is gonna find a proxy in there in specific. Of course you could say why not search for "proxy" but that could be said again for any sub-category.

FWIW I'll respect whatever choice you make though. You are all doing an amazing job and I have 0 entitlement in asking for a new sub-category. Just laying out how I see it. But perhaps I am just wrong or have not enough justification (if anything at all). As such, I'll respect however this plays out and take it as a good call.

Turbo87:
we finally discussed this proposal in our crates.io team meeting today. we don't have a very strong opinion on this, but we tend towards rejection. to be able to judge this properly, can you compile a list of a dozen or so crates with reasonably high download numbers that would fit in such a category?

GlenDC:
What is reasonably high? What is the limit?

Turbo87:
there is no hard limit, but more than a couple thousand downloads usually means that it's downloaded by more than just a few mirrors and bots 😅

GlenDC:
Dunno if my framework counts: https://crates.io/crates/rama

but next to that you can easily find a dozen crates by using keywords such as https://crates.io/keywords/proxy. Problem with that keyword is that not all of them are proxies indeed or purely proxies. Some are just connectors or small components of what can be a proxy. However on most of those pages you do find some candidates that are full fledged proxy frameworks / libraries that would be served well if they can put proxy as (sub)category correctly.

jtgeibel:
I just posted a longer comment in [this thread](https://github.com/rust-lang/crates.io/pull/11856#issuecomment-3288437438), but from the summary:

> As a user I really like the discoverability of categories, but if popular crates don't opt-in to the appropriate category then it can just as easily push users towards a worse selection than if they had done a more general search. In that sense, new categories have a steep hill to climb before they are a net value add and I think that is one of the reasons the team has been more reluctant to adding new categories recently.

I'm speaking only for myself there, but as a rough temperature check I think this (and the other bullet points listed there) helps explain why we've been setting a higher bar for new categories lately.

GlenDC:
No problem. Personally I find it impossible as a user to find proxy frameworks on crates.io. As the proxy tag has so many crates that are just offering 1 proxy-related freature but not really e2e ready.

Either way, all good for me. I can see the opinion is already pretty much settled here. I'll close it for now and save you the trouble.

PR #11745, opened 2025-08-13
add windows signal handlers to server executable

2ndDerivative:
I am aware that windows isn't supposed to supported officially, but this would at least help the server executable compile on windows.

I hope it's not too inconvenient if I go through some of the hiccups a windows user might have when trying to get this running on their platform

### DIFF
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -13,7 +13,10 @@ use reqwest::Client;
 use std::io::Write;
 use std::net::SocketAddr;
 use tokio::net::TcpListener;
+#[cfg(unix)]
 use tokio::signal::unix::{SignalKind, signal};
+#[cfg(windows)]
+use tokio::signal::windows::{ctrl_break, ctrl_c, ctrl_close, ctrl_logoff, ctrl_shutdown};
 use tower::Layer;
 
 const CORE_THREADS: usize = 4;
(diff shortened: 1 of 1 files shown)


src/bin/server.rs
```diff
@@ -111,6 +115,52 @@ async fn shutdown_signal() {
...
+            .await;
+    };
+
+    tokio::select! {
+        _ = ctrl_break => {},
+        _ = ctrl_c => {},
+        _ = ctrl_close => {},
+        _ = ctrl_logoff => {},
+        _ = ctrl_shutdown => {},
+    }
```
Turbo87, line 161: do we need all of them? if this is just for local development, shouldn't e.g. `ctrl_c` be sufficient? 🤔

2ndDerivative: I don't know, I just figured I'd use them all while they're not placed.

To be fair I didn't find the recommended semantics for the signals to be particularly well-documented

Turbo87: since I don't have a Windows machine here, can you check if the following is sufficient and compiles?

```rs
#[cfg(windows)]
async fn shutdown_signal() {
    ctrl_c()
        .expect("failed to install signal handler")
        .recv()
        .await
}
```

it's a bit unfortunate that CI won't be able to check that this actually compiles since we run CI only on Unix, but I don't think it makes sense for us to run CI on Windows too.

PR #11789, opened 2025-08-25
database: Remove obsolete `users.gh_access_token` column

Turbo87:
This PR follows #11697 and finally removes the plaintext GitHub access tokens from our database, since our code is no longer using them.


Turbo87:
@rust-lang/crates-io reviews are welcome. 

I would propose to keep the current state running for a week or two to ensure the current system with the encrypted tokens works as intended and then ship this PR to remove the old unencrypted tokens.

LawnGnome (approved):
LGTM, and the migration plan sounds solid to me. Let's do it!

PR #11814, opened 2025-08-30
Vite

mansona:


package.json
```diff
@@ -17,21 +17,22 @@
     "test": "tests"
   },
   "scripts": {
-    "build": "ember build --environment=production && node ./script/precompress-assets.mjs",
+    "build": "vite build",
```
Turbo87, line 20: we should probably keep the `precompress-assets.mjs` call 😉

mansona: yea there is going to be loads that needs fixing on this branch, and a bunch of stuff like this was automatically changed with the codemod.

For this specific thing... shouldn't that work be done with a rollup/vite plugin rather than a "post build script" ?

mansona: something like this: https://www.npmjs.com/package/rollup-plugin-gzip (this was literally the first result when searching so I'm not actually proposing this explicitly)

Turbo87: probably, but this worked well enough and we only need it for the production deployment so spending more time on the investigation wasn't worth it 😉

Issue #11892, opened 2025-09-12
Have a way to disable creation of new API tokens


carols10cents:
For temporary use during security incidents. Could limit the damage because you can't get access to existing API tokens at the expense of temporary inconvenience to some users (but less inconvenience than putting the entire site in read-only mode, which is the only ability like this we have at the moment)

PR #11894, opened 2025-09-12
Add `DISABLE_TOKEN_CREATION` flag

Turbo87:
Setting this environment variable will cause crates.io to reject new API token creation requests with the error message set inside the environment variable. This can be useful to e.g. temporarily pause API token creation during ongoing phishing campaigns.

Resolves https://github.com/rust-lang/crates.io/issues/11892

### DIFF
--- a/src/controllers/token.rs
+++ b/src/controllers/token.rs
@@ -6,6 +6,7 @@ use anyhow::Context;
 
 use crate::app::AppState;
 use crate::auth::AuthCheck;
+use crate::middleware::real_ip::RealIp;
 use crate::models::token::{CrateScope, EndpointScope};
 use crate::util::errors::{AppResult, bad_request, custom};
 use crate::util::token::PlainToken;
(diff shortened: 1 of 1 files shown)


walterhpearce:
Can we add logging during the disablement period, to explicitly capture all attempts to create a token during the disabled period?

carols10cents:
> Can we add logging during the disablement period, to explicitly capture all attempts to create a token during the disabled period?

It's there: https://github.com/rust-lang/crates.io/pull/11894/files#diff-13b69ea8d077bfdc7f18964826bd9ca405a8225da8fe8b3373dc0a3c7f964aa5R132-R135

src/controllers/token.rs
```diff
@@ -127,6 +127,17 @@ pub async fn create_api_token(
 
     let user = auth.user();
 
+    // Check if token creation is disabled
+    if let Some(disable_message) = &app.config.disable_token_creation {
+        warn!(
+            "Blocked token creation for user `{}` (id: {}) due to disabled flag (token name: `{}`)",
+            user.gh_login, user.id, new.api_token.name
+        );
```
Turbo87: @walterhpearce isn't that what we're doing here? 😅

walterhpearce: I missed it because I am not smart. Can we log the full request during this period? Headers +  addresses.

Turbo87: I've added the IP address and the headers, minus cookie and authorization header to not unnecessarily expose these in our logs.

walterhpearce: Lgtm

PR #11926, opened 2025-09-17
Add API token support to endpoints for managing Trusted Publishing configurations

Turbo87:
This PR adds a new `trusted-publishing` endpoint scope to our API tokens. If this scope is used, then the API token can be used to create, list and delete Trusted Publishing configurations for the configured crates. As with the other scopes, a legacy token without any scopes can also be used for the same purpose now.

This change will allow our users with larger workspaces to automate the setup of creating Trusted Publishing configurations, which should help with adoption of Trusted Publishing by power users.

### DIFF
--- a/src/controllers/trustpub/github_configs/create/tests.rs
+++ b/src/controllers/trustpub/github_configs/create/tests.rs
@@ -2,6 +2,7 @@ use crate::tests::builders::CrateBuilder;
 use crate::tests::util::{RequestHelper, Response, TestApp};
 use anyhow::anyhow;
 use bytes::Bytes;
+use crates_io_database::models::token::{CrateScope, EndpointScope};
 use crates_io_database::schema::{emails, trustpub_configs_github};
 use crates_io_github::{GitHubError, GitHubUser, MockGitHubClient};
 use diesel::prelude::*;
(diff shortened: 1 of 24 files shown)


src/controllers/trustpub/github_configs/create/tests.rs
```diff
@@ -243,9 +244,141 @@ async fn test_token_auth() -> anyhow::Result<()> {
...
+
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_token_auth_with_wildcard_crate_scope() -> anyhow::Result<()> {
+    let (app, _client, cookie_client, token_client) = TestApp::full()
+        .with_github(simple_github_mock())
+        .with_scoped_token(
+            Some(vec![CrateScope::try_from("f*").unwrap()]),
```
eth3lbert:
```suggestion
            Some(vec![CrateScope::try_from("*").unwrap()]),
```

It would be great if this could be changed to match the others.

eth3lbert (approved):
Thanks for making this. Aside from a nitpick about the un-unified wildcard scope and seemingly missing e2e tests for the UI, it LGTM overall!

Turbo87:
both fixed, thanks for the review! ❤️

PR #11927, opened 2025-09-17
settings/tokens: Fix scope list formatting

Turbo87:
Previously the page would show lists of crate name patterns as: `Crates: serde , serde-* , and serde_*` (with an additional whitespace in front of the comma). This was apparently introduced unintentionally when we refactored our tooltip implementation.

This PR removes the excess whitespace so that the page now shows `Crates: serde, serde-*, and serde_*`. This PR also adds corresponding tests to avoid future regressions.

### DIFF
--- a/app/components/tooltip.gjs
+++ b/app/components/tooltip.gjs
@@ -64,10 +64,12 @@ export default class Tooltip extends Component {
     return () => cleanup();
   });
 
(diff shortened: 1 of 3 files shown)


Turbo87:
grafik

turns out the above also works, but the IntelliJ Handlebars parser is unhappy about it, so I guess we'll keep it as is for now 😅

PR #11955, opened 2025-09-22
Show SLoC values in crate page sidebars

Turbo87:
#11453 is merged and the backfill has finished, so now we can start showing the SLoC estimates next to the crate file size as another data point:

Bildschirmfoto 2025-09-22 um 15 48 38

### Related

- #11453


src/views.rs
```diff
@@ -924,6 +924,15 @@ pub struct EncodableVersion {
...
     pub trustpub_data: Option<TrustpubData>,
+
+    /// Line count statistics for this version.
+    ///
+    /// Status: **Unstable**
+    ///
+    /// This field may be `null` until the version has been analyzed, which
+    /// happens in an asynchronous background job.
+    #[schema(value_type = Object)]
+    pub linecounts: Option<serde_json::Value>,
```
Turbo87, line 935: I wonder about calling this `experimental_linecounts` or `unstable_linecounts` to make that status clear, even for API users that don't know about the OpenAPI documentation 🤔

jtgeibel: The documentation route seems fine to me. Otherwise you end up breaking users when stabilizing it, or keeping an alias around forever.

LawnGnome: I also think this is fine, although I wonder if we should say a little more about the shape of this field in the documentation, since it's just going to appear as an `Object` in the API.

LawnGnome (approved):

src/views.rs
Turbo87: > I wonder if we should say a little more about the shape of this field in the documentation, since it's just going to appear as an `Object` in the API.

I explicitly didn't want to document the exact content yet while this is being considered unstable 😅 

I think once we stabilize this we should document it, but as long as we're the only users this might actually be counter-productive

LawnGnome: That's fair, and I figured that was your logic. I'd have a mild preference to document it anyway with warnings, but I don't feel strongly enough about this for it to be a blocker.

Issue #11893, opened 2025-09-12
Have a way to quickly put up a banner on every page


carols10cents:
To be used for security incidents

PR #11967, opened 2025-09-23
Add `BANNER_MESSAGE` environment variable

Turbo87:
This PR resolves https://github.com/rust-lang/crates.io/issues/11893 by introducing a `BANNER_MESSAGE` environment variable. If set, the content will be available at https://crates.io/api/v1/site_metadata, which is already checked by our frontend to show a banner if the service is in read-only mode. The third commit adjusts the frontend logic to display the `banner_message` content instead, if it is set.

/cc @carols10cents

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -64,13 +64,16 @@ export default class ApplicationRoute extends Route {
     let timeout = Ember.testing ? 0 : 1000;
     await rawTimeout(timeout);
 
-    let { read_only: readOnly } = await ajax('/api/v1/site_metadata');
-    if (readOnly) {
-      let message =
+    let { read_only, banner_message } = await ajax('/api/v1/site_metadata');
+
+    if (!banner_message && read_only) {
+      banner_message =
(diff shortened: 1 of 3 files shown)


eth3lbert (commented):
I haven't reviewed the content yet, but from the description, it seems this proposal leverages an environment variable to change the banner message. I'm not sure how much I like this, as it implies that every time a banner changes, we'd need to restart the application. I wonder if it would be more appealing to use a table accompanied by either a queue or pg_notify for this job, where an admin could just make a request to change the banner. However, this might be overkill if we only need to change the banner very rarely.

Turbo87:
> However, this might be overkill if we only need to change the banner very rarely.

yeah, I assume we would rarely use this. the only advantage would be that regular admins could set the message without needing infra admin permission, but I'd say we can probably retrofit this if we really need this. since we have an on-call rotation with pagers the security WG will be able to quickly notify us to set the message for them, if necessary :)

PR #11989, opened 2025-09-26
Always print download numbers when admin deleting a crate

carols10cents:
Requested by Pietro in zulip.

So that if we're deleting the crate for security reasons but forgot to grab the downloads from the page, the person deleting it at least has the download numbers handy in their console.

Rather than making _display_ conditional on how high the download numbers are, make the _red and bold_ styling conditional on how high the download numbers are, using the same formula that we use when deciding if a user should be allowed to delete their own crate.

### DIFF
--- a/src/bin/crates-admin/delete_crate.rs
+++ b/src/bin/crates-admin/delete_crate.rs
@@ -168,7 +168,7 @@ impl Display for CrateInfo {
         let mut downloads = format!("downloads={}", self.downloads);
         let age = Utc::now().signed_duration_since(self.created_at);
         if self.downloads as u64 > max_downloads(&age) {
-            downloads = downloads.bright_red().bold().to_string();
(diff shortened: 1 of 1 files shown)


Turbo87:
> make the _red and bold_ styling conditional on how high the download numbers are

now I'm curious, has the red/bold styling actually worked for you in the past when running this on Heroku? I could get it to work locally, but it seemed that when run via `heroku run -a crates-io -- target/release/crates-admin` it does not detect that it should be using color output. I've mostly been relying on "if the number is shown, then double check the crate" so far 😅

carols10cents:
> now I'm curious, has the red/bold styling actually worked for you in the past when running this on Heroku?

ooooh you're right. it totally doesn't work coming through heroku. wdyt about the latest commit i pushed instead?

LawnGnome (approved):

PR #11999, opened 2025-09-28
Add ARIA heading role to crate search results

### LINKED ISSUE #11759: Accessibility: Crate names should be headings
### Current Behavior

I am blind user who uses VoiceOver on macOS and JAWs, NVDA on Windows. When searching for crates, there is no easy way to jump to the different crates because they are all links, rather than being differentiated semantically somehow.

### Expected Behavior

I expect crates to be headings at an appropriate level to make it easy to jump between crate names without having to go through every unrelated link to find them.

### Steps To Reproduce

1. Open your preferred screen reader
2. Go to crates.io on any accessible browser
3. Search for any crate, such as leptos
4. Try navigation commands to find only crate names
5. Observe that the best you can do is jump by link, where there will be many links between crate namesf

### Environment

- Browser: Any browser compatible with your preferred screen reader.
- OS: Anyf

### Anything else?

_No response_

DataTriny:
Closes #11759 

Making name and version appear as headings inside search results list is important for screen-reader users. It allows for jumping between elements of a list by pressing a single key on the keyboard.

Level 2 headings seem to be more appropriate given the current site structure.

I went with ARIA roles and attributes rather than `h2` tag so as to not mess up the visual appearance. If using the semantic tag is preferred, please provide the right CSS rules as I am not able to verify this myself.

### DIFF
--- a/app/components/crate-row.gjs
+++ b/app/components/crate-row.gjs
@@ -16,7 +16,7 @@ import truncateText from 'crates-io/helpers/truncate-text';
 <template>
   <div data-test-crate-row ...attributes class='crate-row'>
     <div class='description-box'>
-      <div class='crate-spec'>
+      <div data-test-crate-spec class='crate-spec' role='heading' aria-level='2'>
         {{#let (link_ 'crate' @crate.id) as |l|}}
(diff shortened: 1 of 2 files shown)


Turbo87 (approved):
thanks! let's give this a try :)

PR #12022, opened 2025-09-30
me/crates: Redirect to user profile page to eliminate duplication

Turbo87:
There is currently basically no difference between https://crates.io/me/crates and https://crates.io/users/Turbo87, so let's remove the former and replace it with a redirect to the latter.


eth3lbert (approved):
:shipit:

Issue #11987, opened 2025-09-25
Trusted Publishing support for GitLab CI


Turbo87:
Similar to GitHub Actions, GitLab CI also supports OIDC authentication (see https://docs.gitlab.com/ci/secrets/id_token_authentication/). This issue tracks the work that is necessary to build Trusted Publishing support for it.

### Steps

- [x] Implement backend support in crates.io
  - [x] Add database table for GitLab CI trusted publishing configs (https://github.com/rust-lang/crates.io/pull/11988)
  - [x] Adjust token exchange API endpoint to support GitLab CI OIDC authentication (https://github.com/rust-lang/crates.io/pull/12026)
  - [x] Implement API endpoint to create GitLab CI trusted publishing configs (incl. email notification) (https://github.com/rust-lang/crates.io/pull/12268)
  - [x] Implement API endpoint to delete GitLab CI trusted publishing config (incl. email notification) (https://github.com/rust-lang/crates.io/pull/12285)
  - [x] Implement API endpoint to list GitLab CI trusted publishing configs (https://github.com/rust-lang/crates.io/pull/12292)
  - [x] Adjust `trustpub_config_created/deleted` emails to support GitLab CI configurations (https://github.com/rust-lang/crates.io/pull/12048)
  - [x] Adjust `SendPublishNotifications` background job to show GitLab CI information (https://github.com/rust-lang/crates.io/pull/12312)
- [x] Implement frontend UI in crates.io
  - [x] Adjust trusted publishing configs list to include GitLab CI configs (https://github.com/rust-lang/crates.io/pull/12335)
  - [x] Implement button to delete GitLab CI trusted publishing configs (https://github.com/rust-lang/crates.io/pull/12335)
  - [x] Adjust trusted publishing config creation form to support GitLab CI configs (https://github.com/rust-lang/crates.io/pull/12335)
  - [x] Adjust `version` model to support GitLab CI in `trustpubPublisher` and `trustpubUrl` fields (https://github.com/rust-lang/crates.io/pull/12345)
  - [x] Adjust `VersionList` component to show GitLab icons (https://github.com/rust-lang/crates.io/pull/12345)
  - [x] Write documentation for the Trusted Publishing feature (https://github.com/rust-lang/crates.io/pull/12346)
- [x] Figure out how to use it on the GitLab CI side
- [ ] Run a public beta period for a couple of weeks before declaring this feature stable

### Related

- rust-lang/rfcs#3691
[...]

PR #12026, opened 2025-10-02
trustpub/tokens/exchange: Implement GitLab CI support

Turbo87:
This PR adjusts our `POST /api/v1/trusted_publishing/tokens` endpoint to also accept GitLab OIDC tokens.

As discussed in https://github.com/rust-lang/crates.io/pull/11988, we are using a trust-on-first-use principle for the namespace IDs for GitLab due to some API constraints. In other words: the token exchange endpoint will fill in the `namespace_id` column from the OIDC JWT claims if it was empty before and on each subsequent token exchange it will only accept it if the IDs are matching.

Other than that, this is essentially the same implementation as for GitHub Actions.

Worth noting also is that, just like PyPI, we only support gitlab.com for now, but not any self-hosted GitLab instances.

### Related

- https://github.com/rust-lang/crates.io/issues/11987


src/controllers/trustpub/tokens/exchange/mod.rs
```diff
@@ -221,3 +227,162 @@ async fn handle_github_token_inner(
...
+}
+
+async fn handle_gitlab_token_inner(
+    conn: &mut AsyncPgConnection,
+    signed_claims: GitLabClaims,
+) -> AppResult<Json<json::ExchangeResponse>> {
+    insert_jti(conn, &signed_claims.jti, signed_claims.exp).await?;
+
+    let project_path = &signed_claims.project_path;
+    let Some((namespace, project)) = project_path.rsplit_once('/') else {
```
LawnGnome, line 260: I know why this is `rsplit_once` (the GitLab optional project thing), but it might be worth a comment explaining why this differs from GitHub, since `rsplit_once` versus `split_once` is pretty subtle.

src/controllers/trustpub/tokens/exchange/mod.rs
```diff
@@ -221,3 +227,162 @@ async fn handle_github_token_inner(
...
+                .namespace_id
+                .as_ref()
+                .is_some_and(|stored| stored != &signed_claims.namespace_id)
+        })
+        .filter_map(|config| config.namespace_id)
+        .collect();
+
+    if repo_configs.is_empty() {
+        let message = format!(
+            "The Trusted Publishing config for repository `{project_path}` does not match the namespace ID ({}) in the JWT. Expected namespace IDs: {}. Please recreate the Trusted Publishing config to update the namespace ID.",
```
LawnGnome, line 316: This is a really good explanation. :+1:

src/controllers/trustpub/tokens/exchange/mod.rs
Turbo87: oh, I thought I had left a comment on that, but apparently not. I'll add one. 👍

PR #12044, opened 2025-10-06
fix: correct the schema of the versions table

0xPoe:
I think the order of the files in the versions table was broken, because Diesel runs migrations based on their creation time. Although we merged [a325520a](https://github.com/rust-lang/crates.io/commit/a325520a52f6baa2ef118ceda2f65a410c9b6518) a long time ago, its migration file is newer than [c977a31](https://github.com/rust-lang/crates.io/commit/c977a317234c896c6f14965bb7f7b9b80d32f6d7).
As a result, if you run the migrations locally, you’ll end up with a reordered schema file.

### DIFF
--- a/crates/crates_io_database/src/schema.rs
+++ b/crates/crates_io_database/src/schema.rs
@@ -1109,10 +1109,10 @@ diesel::table! {
         keywords -> Array<Nullable<Text>>,
         /// JSONB representation of the version number for sorting purposes.
         semver_ord -> Nullable<Jsonb>,
-        /// JSONB data containing JWT claims from the trusted publisher (e.g., GitHub Actions context like repository, run_id, sha)
-        trustpub_data -> Nullable<Jsonb>,
         /// Source Lines of Code statistics for this version, stored as JSON with language breakdown and totals.
         linecounts -> Nullable<Jsonb>,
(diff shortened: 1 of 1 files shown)


0xPoe (commented):
We may need to figure out how to prevent this kind of issue from happening again. It seems that our CI didn’t catch this issue.

Turbo87:
huh, that's a good catch. I'm not sure about the proposed solution though. on production the currently committed schema is probably correct since the migrations were applied in merge order. as long as we query by column name this order mismatch shouldn't matter too much, but if we query by column order this might be problematic.

@weiznich do you have any suggestions for situations like this? :)

weiznich:
There is a [`column_sorting`](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/config.rs#L461) option for `diesel.toml` files that could be used to address this. At currently has [these options](https://github.com/diesel-rs/diesel/blob/master/diesel_cli/src/print_schema.rs#L15). I really need to get around documenting all the missing options on the web page…

That written: Diesel always requests columns by name in your select statement, so a different order in the `table!` definition shouldn't cause any issues as long as you don't rely on deserializing the data with `#[derive(Queryable)]` only (so no `Selectable`) and the fields in question all have the same type or as long as you don't write any custom select via `QueryFragment` and use `*` there. If you using `Selectable` (or the newer `HasQuery` derive) it should never be an issue.

Issue #11834, opened 2025-09-03
Accessing reverse dependencies for serde returns an HTTP 500 and redirects to /


alex:
### Current Behavior

When I go to https://crates.io/crates/serde -> press Dependents an HTTP 500 is returned (bug #1) and then the whole page redirects to / (bug #2)

### Expected Behavior

a) Obviously the 500 is a bug
b) If a 500 occurs, we should stay on the current serde page, not redirect to /

kailan:
I think the 500 is a timeout. It seems to happen at the 10 second mark each time. For what it's worth, I got the page to load after a few refreshes.

On a local instance of crates.io it takes about 2 seconds but loads fine. It makes sense that this would take longer in production due to the load on the database.

Perhaps for more popular crates the reverse dependency index needs to be generated in a background job rather than on demand.

kailan:
Fixed the redirect on error. The core issue of reverse dependencies taking a long time to load or not loading at all is still present.

alex:
Thank you!

Turbo87:
this was resolved by https://github.com/rust-lang/crates.io/pull/13897 AFAICT :)

PR #12080, opened 2025-10-10
Remain on current page if navigating to reverse dependencies page fails

kailan:
This resolves one of the reports in #11834.

Widely depended-on crates can cause timeouts when loading the reverse dependencies page.

The reverse dependencies route is the only one (bar email verification, which makes sense) that sends you to the index page upon error.

This pull request updates the error handling logic to use the `catch-all` redirect, matching the behaviour of the standard (forward?) dependencies page.

### DIFF
--- a/app/routes/crate/reverse-dependencies.js
+++ b/app/routes/crate/reverse-dependencies.js
@@ -10,7 +10,7 @@ export default class ReverseDependenciesRoute extends Route {
     page: { refreshModel: true },
   };
 
-  async model(params) {
+  async model(params, transition) {
     params.reverse = true;
     params.crate = this.modelFor('crate');
     let crateName = params.crate.name;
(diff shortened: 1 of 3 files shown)


kailan:
Not sure how to decode this 🤔 

Screenshot 2025-10-10 at 17 47 24

Turbo87:
oh boy... looks gnarly 😅

I can take a look once I have a quiet minute again :)

Turbo87:
my first thought would be that this might have to use https://github.com/rust-lang/crates.io/blob/main/tests/helpers/visit-ignoring-abort.js, but it's just an educated guess 😄

kailan:
> my first thought would be that this might have to use https://github.com/rust-lang/crates.io/blob/main/tests/helpers/visit-ignoring-abort.js, but it's just an educated guess 😄

Thanks for that – turns out that helper was already being used but this helped point me in the right direction!

I've updated the PR so that the behaviour of the reverse dependencies page now matches the standard dependencies page. This seems to make the most sense to me for now.

PR #12084, opened 2025-10-10
Replace `Ember.testing` with `@embroider/macros`

locks:
## Ember.testing
- [x] replace Ember.testing with embroider/macro

## inject as service
- [ ] ember-cli-head
- [ ] ember-cli-notifications
- [ ] [ember-error-route](https://github.com/mainmatter/ember-error-route/pull/762)


Turbo87:
@locks it looks like the Playwright test suite is unhappy about that change. could you check if we can fix that somehow? :)

Turbo87:
@locks any updates?

PR #12088, opened 2025-10-11
Replace – with - to prevent strange unicode issues in OpenAPI JSON rendering

Evian-Zhang:
The former "–" is U+2013, En Dash, the latter is the normal hyphen can be directly typed from keyboard, which is U+002D.

The U+2013 dash makes the JSON output of OpenAPI page look strange:

image

The actual rendering output may differ based on locale.

### DIFF
--- a/src/snapshots/crates_io__openapi__tests__openapi_snapshot-2.snap
+++ b/src/snapshots/crates_io__openapi__tests__openapi_snapshot-2.snap
@@ -1422,7 +1422,7 @@ expression: response.json()
     },
     "/api/private/session/authorize": {
       "get": {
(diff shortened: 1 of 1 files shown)


Turbo87:
> the JSON output of OpenAPI page

I'm not sure what you mean by "OpenAPI page" but this seems like a bug in that page, not in our description 🤔

Evian-Zhang:
It's <https://crates.io/api/openapi.json>. I found this link from the [Data Access](https://crates.io/data-access#api) page of crates.io

Turbo87:
> It's https://crates.io/api/openapi.json

ah, I see.

Bildschirmfoto 2025-10-11 um 12 01 57

seems to work fine here though. what browser are you using to access that file?

Evian-Zhang:
It's Safari 26.0.1 (21622.1.22.11.15) with default settings. I think this is related to encodings and locale. The default encoding setting of Safari in my region is GB2312 (a Chinese-specific encoding), if I changed it to Unicode (UTF - 8), then it renders fine.

Turbo87:
> The default encoding setting of Safari in my region is GB2312 (a Chinese-specific encoding)

is that really the default in Safari? 😳 

I would have expected that it defaulted to UTF-8 everywhere at this point 😅 

anyways, JSON is UTF-8 by default, so this seems like a bug in Safari, if anything :)

Evian-Zhang:
😂 Only Apple can do.

Anyway, I will close the PR. Thank you!

PR #12217, opened 2025-10-28
Limit pagination in the user interface

Turbo87:
This pull request adds support for limiting the maximum number of pagination pages across several user interface controllers and improves the user experience when the page limit is reached. The changes introduce configurable `maxPages` options, update the pagination utility, and add a helpful tooltip for users.

**Pagination logic improvements:**

* The `pagination` utility in `app/utils/pagination.js` now accepts an optional `maxPages` parameter, which limits the number of available pages. The pagination calculation uses this limit if provided. [[1]](diffhunk://#diff-33f16ff085a2bcc130ceb1adceeccc3c3ed08db2c973ea9ef9586dcf408bf746L5-R20) [[2]](diffhunk://#diff-33f16ff085a2bcc130ceb1adceeccc3c3ed08db2c973ea9ef9586dcf408bf746R59)

* The `@pagination` decorator is updated in `crates`, `keyword`, `search`, `team`, and `user` controllers to specify a `maxPages` value (20 for most, 50 for team and user), enforcing a maximum number of pages in those views. [[1]](diffhunk://#diff-639620c4f205afded7044b93ea0974aeda93fa4d48c2f0764e9bbf9ca9c5d11cR6-R14) [[2]](diffhunk://#diff-9e09f2ffda292f6736774997268d458981aec2901a7f2cd5a87443ad368c00ceR6-R14) [[3]](diffhunk://#diff-ca2ec0ae83510380e9a0e973d82d2606011a42e7a81de43426347cf1ad7ce1e4R12-R13) [[4]](diffhunk://#diff-ca2ec0ae83510380e9a0e973d82d2606011a42e7a81de43426347cf1ad7ce1e4L32-R34) [[5]](diffhunk://#diff-6cfa4dbf4e8e6f6fd6ba8bf60c9b8d9424f08fed77c8c623e0cca473a0384a18R6-R14) [[6]](diffhunk://#diff-3f6fded6c0945c887bc172de5765bbad4241d9ece9eb8c411b5e2fa17c4a959dR8-R16)

**User interface enhancements:**

* In the `pagination.gjs` component, a tooltip is displayed when the user is on the last available page, informing them that the page limit is for performance reasons and providing a link to bulk data access. [[1]](diffhunk://#diff-8ba390d30d159df08ca30a2bac8f9f518f28383946281dd13afd235f829fe3edR7-R8) [[2]](diffhunk://#diff-8ba390d30d159df08ca30a2bac8f9f518f28383946281dd13afd235f829fe3edR39-R44)

### DIFF
--- a/app/components/pagination.gjs
+++ b/app/components/pagination.gjs
@@ -38,7 +38,7 @@ import Tooltip from './tooltip';
       {{svgJar 'right-pag'}}
       {{#if (eq @pagination.currentPage @pagination.maxPages)}}
         <Tooltip>
-          Page limit reached for performance reasons. For bulk data access, please visit
+          For performance reasons, no more pages are available. For bulk data access, please visit
           <a href='https://crates.io/data-access' target='_blank' rel='noopener noreferrer'>crates.io/data-access</a>.
         </Tooltip>
       {{/if}}


eth3lbert:
I haven't taken a close review, but I'm wondering: beyond the UI restriction, should we also either redirect to the page with `maxPage` or show an error page for those who access the URL directly with a page number exceeding `maxPage` (whether from history or manual input)🤔

Turbo87:
if an error is returned from the backend then we automatically display it, otherwise it's only a soft limit on the frontend in that we don't link higher pager numbers, but don't prevent them from explicitly being requested. not sure it's worth the extra complexity to make it a hard limit on the frontend too.

app/components/pagination.gjs
```diff
@@ -34,6 +36,12 @@ import { eq } from 'ember-truth-helpers';
       data-test-pagination-next
     >
       {{svgJar 'right-pag'}}
+      {{#if (eq @pagination.currentPage @pagination.maxPages)}}
+        <Tooltip>
+          Page limit reached for performance reasons. For bulk data access, please visit
```
LawnGnome: I might massage this _very_ slightly into:

```suggestion
          For performance reasons, no more pages are available. For bulk data access, please visit
```

But I don't care very much either way. This is fine as it is.

app/components/pagination.gjs
```diff
@@ -34,6 +36,12 @@ import { eq } from 'ember-truth-helpers';
...
+          <a href='https://crates.io/data-access' target='_blank' rel='noopener noreferrer'>crates.io/data-access</a>.
```
LawnGnome, line 42: All these years later, it still kills me that the `rel` keyword and header have different spellings.

PR #12219, opened 2025-10-28
trustpub: Block `pull_request_target` and `workflow_run` triggers

Turbo87:
Add validation to reject GitHub Actions workflows using `pull_request_target` or `workflow_run` event triggers during trusted publishing token exchange.

These triggers are dangerous because they run in the target repository's security context while being triggerable by forked repositories, creating a privilege escalation vector.

The validation occurs after JWT decoding and JTI replay prevention but before any config matching, ensuring early rejection of potentially dangerous requests.

Changes:
- Add `event_name` field to `GitHubClaims` struct
- Implement validation check in `handle_github_token_inner()`
- Add tests for `pull_request_target` and `workflow_run` rejection
- Update all existing tests to include `event_name` field

### DIFF
--- a/crates/crates_io_trustpub/src/github/claims.rs
+++ b/crates/crates_io_trustpub/src/github/claims.rs
@@ -25,6 +25,7 @@ pub struct GitHubClaims {
     pub repository: String,
     pub workflow_ref: String,
     pub environment: Option<String>,
+    pub event_name: String,
     pub run_id: String,
     pub sha: String,
 }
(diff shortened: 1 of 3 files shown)


marcoieni:
nice that you can block this from the server side 👍

walterhpearce:
I've discussed this with some of the other ecosystems, and they are also moving to block these event triggers for trusted publishing actions. They may be more broadly attacking some other event types - but I think we are best scoping this down to the most critical cases of nefarious action.

LawnGnome (approved):
Implementation LGTM. :+1:

PR #12224, opened 2025-10-28
Integrate `aws-ip-ranges` as workspace crate

Turbo87:
This moves the `aws-ip-ranges` crate into the workspace at `crates/aws_ip_ranges` to eliminate the need for frequent crate publishes to crates.io.

The crate is now a path dependency that gets compiled with the main project. AWS IP ranges are still embedded at compile time via the build script, maintaining zero runtime cost.

A GitHub Actions workflow runs hourly to download the latest IP ranges from AWS and commit any changes directly to the main branch.

This eliminates ecosystem noise from 1500+ published versions while keeping the same compile-time performance characteristics.

This essentially deprecates https://github.com/Turbo87/aws-ip-ranges.


Turbo87:
closing in favor of #12225

PR #12235, opened 2025-10-29
Limit GitHub Trusted Publishing configurations to 5 per crate

Turbo87:
Adds a validation to prevent creating more than 5 GitHub Trusted Publishing configurations per crate. When the limit is reached, returns a 400 Bad Request with a clear error message.

We shouldn't allow people to create an infinite number of configurations and bloat our database in doing so. This is also relevant to the listing API endpoint, which currently does not support pagination.

### DIFF
--- a/src/controllers/trustpub/github_configs/create/mod.rs
+++ b/src/controllers/trustpub/github_configs/create/mod.rs
@@ -3,7 +3,7 @@ use crate::auth::AuthCheck;
 use crate::controllers::krate::load_crate;
 use crate::controllers::trustpub::emails::{ConfigCreatedEmail, ConfigType};
 use crate::controllers::trustpub::github_configs::json;
-use crate::util::errors::{AppResult, bad_request, forbidden, server_error};
+use crate::util::errors::{AppResult, bad_request, custom, forbidden, server_error};
 use anyhow::Context;
 use axum::Json;
 use crates_io_database::models::OwnerKind;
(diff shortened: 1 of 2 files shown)


Turbo87:
I checked on production and most crates only have 1 or 2 configurations. The only exception is https://crates.io/crates/cargo-binstall with 4 configurations. I currently don't see why a crate would need that many, but if this turns out to be insufficient for good reasons, then we can still bump the limit.

Turbo87:
yeah, it's certainly possible to exploit the race condition, but the limit is low enough that it probably won't matter too much in practice. if we do see this become a problem we can always move this check into the database itself.

src/controllers/trustpub/github_configs/create/mod.rs
```diff
@@ -53,6 +55,16 @@ pub async fn create_trustpub_github_config(
...
     let krate = load_crate(&mut conn, &json_config.krate).await?;
 
+    // Check if the crate has reached the maximum number of configs
+    let config_count = GitHubConfig::count_for_crate(&mut conn, krate.id).await?;
+    if config_count >= MAX_CONFIGS_PER_CRATE as i64 {
+        let message = format!(
+            "This crate already has the maximum number of GitHub Trusted Publishing configurations ({})",
+            MAX_CONFIGS_PER_CRATE
+        );
+        return Err(bad_request(&message));
```
LawnGnome: I wonder if this should be a 409, rather than a 400. (But I don't feel very strongly about it.)

Turbo87: no objections from my side. implemented it.

PR #12236, opened 2025-10-29
trustpub: Add cursor-based pagination to GitHub configs list endpoint

Turbo87:
Add cursor-based pagination to the `/api/v1/trusted_publishing/github_configs` endpoint using the `id` field as the cursor. Pagination is always enabled with a default of 10 items per page (following the codebase standard).

The implementation follows existing pagination patterns from the versions endpoint, using the `seek!()` macro for type-safe cursor serialization.

### DIFF
--- a/src/controllers/trustpub/github_configs/list/mod.rs
+++ b/src/controllers/trustpub/github_configs/list/mod.rs
@@ -1,7 +1,11 @@
 use crate::app::AppState;
 use crate::auth::AuthCheck;
+use crate::controllers::helpers::pagination::{
+    Page, PaginationOptions, PaginationQueryParams, encode_seek,
+};
 use crate::controllers::krate::load_crate;
-use crate::controllers::trustpub::github_configs::json::{self, ListResponse};
+use crate::controllers::trustpub::github_configs::json::{self, ListResponse, ListResponseMeta};
+use crate::util::RequestUtils;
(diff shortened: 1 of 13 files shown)


src/controllers/trustpub/github_configs/list/mod.rs
```diff
@@ -64,10 +69,13 @@ pub async fn list_trustpub_github_configs(
         return Err(bad_request("You are not an owner of this crate"));
     }
 
-    let configs = GitHubConfig::query()
-        .filter(trustpub_configs_github::crate_id.eq(krate.id))
-        .load(&mut conn)
-        .await?;
+    let pagination = PaginationOptions::builder()
+        .enable_seek(true)
+        .enable_pages(false)
```
eth3lbert, line 74: nit: This actually disables the offset-based pagination, but the current title doesn't suggest that.

eth3lbert (approved):
Only a little nit about the PR title, the implementation LGTM 👍 Thanks!

src/controllers/trustpub/github_configs/list/mod.rs
Turbo87: previously it wasn't supporting pagination at all though 🤔

eth3lbert: ah, fair enough!

PR #12241, opened 2025-10-29
CI: Fix `git config --global` calls

Turbo87:
Oops... single quotes does not allow variable expansion. This changes it to double quotes, which should fix the issue.

### DIFF
--- a/.github/workflows/update-aws-ip-ranges.yml
+++ b/.github/workflows/update-aws-ip-ranges.yml
@@ -40,11 +40,11 @@ jobs:
           APP_SLUG: ${{ steps.app-token.outputs.app-slug }}
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/update-aws-ip-ranges.yml
```diff
@@ -40,11 +40,11 @@ jobs:
           APP_SLUG: ${{ steps.app-token.outputs.app-slug }}
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
 
-      - run: git config --global user.name '${APP_SLUG}[bot]'
+      - run: git config --global user.name "${APP_SLUG}[bot]"
         env:
           APP_SLUG: ${{ steps.app-token.outputs.app-slug }}
 
-      - run: git config --global user.email '${USER_ID}+${APP_SLUG}[bot]@users.noreply.github.com'
```
devbrett90-prog: alot

PR #12255, opened 2025-10-31
Add authorization logging across requests 

walterhpearce:
This adds: 
- Masked (SHA256 hashed) logging of Authorization header
- Masked (SHA256 hashed) logging of Cookie header
- new custom_metadata.auth_type log entry for the authorization type, being: trustpub, cookie, token

Using a masked version of the authorization or cookie headers across the logs, we can specifically look for cases of tokens used across requests. This can let us detect cases such as: someone attempting tokens/cookies across multiple users, a token used across various crates, or cookie theft across users. 

This came up when responding to the recent phishing incidents, where I wanted to quickly investigate, via logs, actions conducted across users and crates with specific tokens, or correlate past vs. future actions of a given user matching on token/cookie and not specifically just IP/user.

### DIFF
--- a/src/middleware/log_request.rs
+++ b/src/middleware/log_request.rs
@@ -12,8 +12,9 @@ use axum::response::IntoResponse;
 use axum_extra::TypedHeader;
 use axum_extra::headers::UserAgent;
 use derive_more::Deref;
-use http::{Method, Uri};
+use http::{Method, Uri, header};
 use parking_lot::Mutex;
+use sha2::{Digest, Sha256};
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::fmt::Display;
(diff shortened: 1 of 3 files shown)


src/middleware/log_request.rs
```diff
@@ -83,6 +94,8 @@ pub async fn log_requests(
         http.matched_path = %matched_path,
         http.request_id = %request_metadata.request_id.as_ref().map(|h| h.as_str()).unwrap_or_default(),
         http.useragent = %request_metadata.user_agent.as_ref().map(|h| h.as_str()).unwrap_or_default(),
+        http.masked.authorization = hashed_auth_header.unwrap_or_default(),
+        http.masked.cookie = hashed_cookie_header.unwrap_or_default(),
```
Turbo87:
```suggestion
        http.request.headers.masked_authorization = hashed_auth_header.unwrap_or_default(),
        http.request.headers.masked_cookie = hashed_cookie_header.unwrap_or_default(),
```

this would align it with https://docs.datadoghq.com/standard-attributes/?search=http.request to some degree. @walterhpearce wdyt?

Turbo87:
```suggestion
        http.request.headers.hashed_authorization = hashed_auth_header.unwrap_or_default(),
        http.request.headers.hashed_cookie = hashed_cookie_header.unwrap_or_default(),
```

or maybe even this to align with the variable name? seems closer to what it actually is. masking sounds to me like we would just redact parts of the value.

walterhpearce:
Applied your naming change, works for me. I don't really care at all, so whatever makes the most sense from your perspective.

Issue #12281, opened 2025-11-06
Git index syncs can back up and cause incidents


LawnGnome:
As we've seen twice in the last two months — both of which caused pages to go out to the on-call staff member — large numbers of crate publications can cause the Git index sync jobs to back up, since they each require an exclusive lock on the index repo and each sync performs a fetch, commit, and push.

The fix here _may_ be to batch the pending syncs (so fetch once, commit many, then push once), but we need more data before we can answer that.

Turbo87:
resolved by https://github.com/rust-lang/crates.io/pull/13445 :)

PR #12282, opened 2025-11-06
worker: add more fine-grained logging for Git index syncs

LawnGnome:
This should be reverted once we have the data we need, but in the short term, this will be useful to decide our next step with #12281.

The total extra cost here is four log lines per crate version publish, which I think we can comfortably incur for now.


LawnGnome:
I _think_ I'm going to want a slightly more fine grained view than that, but let's merge and deploy #12284 first and see if that gets us enough data to make a decision on #12281. (We may as well leave this open until that's in place, though.)

Turbo87:
Looks like we're running into rate limits with Sentry, so we may as well merge this and rely on Datadog instead.

PR #12284, opened 2025-11-06
Fix Sentry child span capture by binding Hub correctly

Turbo87:
The root cause was that `with_sentry_transaction()` created a new Hub with the transaction context, but then in `worker.rs` the code called `.bind_hub(Hub::current())` which bound the future to the original Hub instead of the new one. This meant all child tracing spans were attached to the wrong Hub that lacked the transaction context, so Sentry never saw them as children of the transaction.

The fix moves the `.bind_hub(hub)` call inside `with_sentry_transaction()` so the callback future executes with the correct Hub that has the transaction, ensuring all child spans are properly associated with the parent transaction.

---

Disclaimer: Claude Code figured this out. At first it looked in entirely the wrong places, but once I nudged it into looking at the right areas it found the issue. I didn't trust it, so I tested it out on staging and it indeed seems to solve the problem

### Before

Bildschirmfoto 2025-11-06 um 12 50 50

### After

Bildschirmfoto 2025-11-06 um 12 50 40

### DIFF
--- a/crates/crates_io_worker/src/util.rs
+++ b/crates/crates_io_worker/src/util.rs
@@ -1,5 +1,5 @@
 use anyhow::anyhow;
-use sentry_core::Hub;
+use sentry_core::{Hub, SentryFutureExt};
 use std::any::Any;
 use std::future::Future;
 use std::panic::PanicHookInfo;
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):
I obviously don't know the Sentry API well, but that makes intuitive sense to me based on my past life as an observability developer, so combined with it working on staging: :+1:

PR #12310, opened 2025-11-12
Remove unused `/settings/email-notifications` route

Turbo87:
The notification emails are configured through `/settings/profile` and this additional settings page has never made it into production and hasn't been maintained for years. Since we don't have bundle splitting per routes, we should remove it to not send extra bytes to our users that aren't actually needed.


eth3lbert (approved):
Sounds reasonable to me :shipit:

PR #12311, opened 2025-11-12
Initial stab at a security tab showing advisories

djc:
Here's my first take at implementing the security tab as proposed in [RFC 3872](https://github.com/rust-lang/rfcs/pull/3872). Resolves https://github.com/rust-lang/crates.io/issues/12507.

This is intentionally light on details to get an initial review on my first baby Ember.js steps. Not sure how much we need to add before this gets merged.

Keeping as draft while the RFC has not been merged.

Screenshot 2025-11-12 at 11 53 29

### DIFF
--- a/e2e/acceptance/security.spec.ts
+++ b/e2e/acceptance/security.spec.ts
@@ -19,7 +19,7 @@ test.describe('Acceptance | crate security page', { tag: '@acceptance' }, () => 
       },
     ];
 
-    msw.worker.use(http.get('https://rustsec.org/packages/:crateId.json', () => HttpResponse.json(advisories)));
+    await msw.worker.use(http.get('https://rustsec.org/packages/:crateId.json', () => HttpResponse.json(advisories)));
 
     await page.goto('/crates/foo/security');
 
@@ -48,7 +48,7 @@ test.describe('Acceptance | crate security page', { tag: '@acceptance' }, () => 
     let crate = await msw.db.crate.create({ name: 'safe-crate' });
     await msw.db.version.create({ crate, num: '1.0.0' });
 
-    msw.worker.use(
+    await msw.worker.use(
       http.get('https://rustsec.org/packages/:crateId.json', () => HttpResponse.text('not found', { status: 404 })),
     );
 
@@ -62,7 +62,7 @@ test.describe('Acceptance | crate security page', { tag: '@acceptance' }, () => 
     let crate = await msw.db.crate.create({ name: 'error-crate' });
     await msw.db.version.create({ crate, num: '1.0.0' });
 
-    msw.worker.use(
+    await msw.worker.use(
       http.get('https://rustsec.org/packages/:crateId.json', () =>
         HttpResponse.text('Internal Server Error', { status: 500 }),
       ),
@@ -89,7 +89,7 @@ test.describe('Acceptance | crate security page', { tag: '@acceptance' }, () => 
       },
     ];
 
-    msw.worker.use(http.get('https://rustsec.org/packages/:crateId.json', () => HttpResponse.json(advisories)));
+    await msw.worker.use(http.get('https://rustsec.org/packages/:crateId.json', () => HttpResponse.json(advisories)));
 
     await page.goto('/crates/xss-test/security');
 


djc:
I've implemented Markdown formatting with GitHub-flavored Markdown support for the advisory details.

Seems like a decent MVP from which we can iterate further?

app/controllers/crate/security.js
```diff
@@ -0,0 +1,21 @@
+import Controller from '@ember/controller';
+import { service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+
+export default class SecurityController extends Controller {
+  @service releaseTracks;
+  @service sentry;
```
Turbo87:
```suggestion
```

these services are not used by anything AFAICT, so I think we can safely remove them

app/controllers/crate/security.js
```diff
@@ -0,0 +1,21 @@
...
+  constructor() {
+    super(...arguments);
+    this.reset();
+  }
+
+  reset() {
+    this.crate = undefined;
+    this.data = undefined;
+  }
+}
```
Turbo87:
```suggestion
```

Since we explicitly assign the controller fields in the `setupController()` fn in the route, I don't think we need an explicit controller implementation at all and can just rely on the automatically generated empty one.

app/routes/crate/security.js
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+  }
+
+  setupController(controller, { crate, advisories, convertMarkdown }) {
+    super.setupController(...arguments);
+    // reset when crate changes
+    if (crate && crate !== controller.crate) {
+      controller.reset();
+    }
```
Turbo87:
```suggestion
```

The `reset()` fn resets the `crate` and `advisories` fields, but, since we immediately reassign them below, I don't think we need the `reset()` call, right?

app/templates/crate/security.gjs
```diff
@@ -0,0 +1,25 @@
...
+          <h3>
+            <a href='https://rustsec.org/advisories/{{advisory.id}}.html'>{{advisory.id}}</a>:
+            {{advisory.summary}}
+          </h3>
+          <p>{{htmlSafe (@controller.convertMarkdown advisory.details)}}</p>
+        </li>
+      {{/each}}
+    </ul>
+  {{else}}
+    <div class='no-results'>
```
Turbo87: There is no corresponding styling block in the CSS file, or did I miss it? 😅

app/routes/crate/security.js
```diff
@@ -0,0 +1,59 @@
...
+  } else {
+    throw new Error(`HTTP error! status: ${response}`);
+  }
+}
+
+export default class SecurityRoute extends Route {
+  queryParams = {
+    sort: { refreshModel: true },
+  };
+
```
Turbo87:
```suggestion
```

The sort query param does not appear to be used anywhere, so I guess we can remove it? I assume the JSON file that we request is not dynamically generated, so that probably won't support different sort orders anyway.

app/routes/crate/security.js
```diff
@@ -0,0 +1,59 @@
...
+        return micromarkModule.micromark(markdown, {
+          extensions: [gfmModule.gfm()],
+          htmlExtensions: [gfmModule.gfmHtml()],
+        });
+      };
+
+      return { crate, advisories, convertMarkdown };
+    } catch (error) {
+      // report unexpected errors to Sentry and ignore `ajax()` errors
+      if (!didCancel(error) && !(error instanceof AjaxError)) {
```
Turbo87: the `didCancel()` is only relevant for ember-concurrency tasks, but since we don't use such tasks here we don't need to check for it.

the `AjaxError` is an error thrown by the `ajax()` utility function in this project, but `fetchAdvisories()` is using raw `fetch()` instead. we have two options here: 1. use `ajax()` instead of `fetch()` and catch the 404 error in `fetchAdvisories()`, or 2. we check for the "HTTP error" error message prefix from the custom error thrown in `fetchAdvisories()`.

if an error is thrown in the try block above we currently only handle it by reporting it to Sentry, but we then implicitly `return undefined;` since there is no other handling of the error. since `setupController()` is using destructuring, and destructuring of `undefined` won't work, this will probably lead to further errors. instead we should show a proper error page like we do for most of the other routes. an example of how this can be accomplished is https://github.com/rust-lang/crates.io/blob/main/app/routes/crate/version-dependencies.js#L34-L35.

Turbo87 (commented):
looks like a good start! 👍 

one more suggestion that did not fit into the diff: we should add tests for this new feature to https://github.com/rust-lang/crates.io/tree/main/e2e/acceptance. at least one test for displaying 2+ advisories, one for the empty state, and one for the rustsec server returning e.g. a 500 error. we use `msw` for network mocking, which should make it relatively straight-forward to implement. it's probably sufficient to implement the msw request handlers in the tests instead of creating a generic handler in the `crates-io-msw` package.

app/templates/crate/security.css
```diff
@@ -0,0 +1,18 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.row {
+  margin-top: var(--space-2xs);
+  background-color: light-dark(white, #141413);
+  padding: var(--space-m) var(--space-l);
+  list-style: none;
+}
```
eth3lbert, line 19:
```suggestion
.row {
  margin-top: var(--space-2xs);
  background-color: light-dark(white, #141413);
  border-radius: var(--space-3xs);
  padding: var(--space-m) var(--space-l);
  list-style: none;
}
```

We often add a little rounding to it.

app/routes/crate/security.js
djc: That's right, removing.

app/controllers/crate/security.js
djc: Removed the entire controller.

app/controllers/crate/security.js
djc: Nice, done.

app/routes/crate/security.js
djc: Fair, removed this block.

app/templates/crate/security.gjs
djc: Added.

app/routes/crate/security.js
djc: Sticking with `fetch()` which seems simpler for this; tried return something that can be destructured even in the case of an error.

app/templates/crate/security.css
djc: Added, thanks.

app/templates/crate/security.gjs
```diff
@@ -0,0 +1,29 @@
...
+            {{advisory.summary}}
+          </h3>
+          {{htmlSafe (@controller.convertMarkdown advisory.details)}}
+        </li>
+      {{/each}}
+    </ul>
+  {{else if @controller.error}}
+    <div class='no-results' data-error>
+      An error occurred while fetching advisories.
+    </div>
```
Turbo87: is there a reason for not using the `replaceWith('catch-all', ...)` pattern suggested in https://github.com/rust-lang/crates.io/pull/12311#discussion_r2618507011? that way we get consistent error page styling across the application and a "Try Again" button to retry the failed request.

app/templates/crate/security.gjs
```diff
@@ -0,0 +1,29 @@
...
+  {{#if @controller.advisories.length}}
+    <h2 class='heading'>Advisories</h2>
+    <ul class='advisories' data-test-list>
+      {{#each @controller.advisories as |advisory|}}
+        <li class='row'>
+          <h3>
+            <a href='https://rustsec.org/advisories/{{advisory.id}}.html'>{{advisory.id}}</a>:
+            {{advisory.summary}}
+          </h3>
+          {{htmlSafe (@controller.convertMarkdown advisory.details)}}
```
Turbo87, line 16: I wonder if we should add a couple of smoke tests for the `micromark` library, to ensure that is sanitizes correctly. if the library changes their defaults we should probably have some kind of notification mechanism to let us know about it (e.g. a failing test case).

e2e/acceptance/security.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    await expect(page.locator('[data-test-list] li').nth(0).locator('h3')).toContainText('First test advisory');
+    await expect(page.locator('[data-test-list] li').nth(0).locator('p')).toContainText('markdown');
+
+    // Check second advisory
+    await expect(page.locator('[data-test-list] li').nth(1).locator('h3 a')).toHaveAttribute(
+      'href',
+      'https://rustsec.org/advisories/TEST-002.html',
+    );
+    await expect(page.locator('[data-test-list] li').nth(1).locator('h3 a')).toContainText('TEST-002');
+    await expect(page.locator('[data-test-list] li').nth(1).locator('h3')).toContainText('Second test advisory');
```
Turbo87:
```suggestion
    // Check first advisory
    let advisory1 = page.locator('[data-test-list] li').nth(0);
    await expect(advisory1.locator('h3 a')).toHaveAttribute('href', 'https://rustsec.org/advisories/TEST-001.html');
    await expect(advisory1.locator('h3 a')).toHaveText('TEST-001');
    await expect(advisory1.locator('h3')).toContainText('First test advisory');
    expect(await advisory1.locator('p').innerHTML()).toBe(
      'This is the first test advisory with <strong>markdown</strong> support.',
    );

    // Check second advisory
    let advisory2 = page.locator('[data-test-list] li').nth(1);
    await expect(advisory2.locator('h3 a')).toHaveAttribute('href', 'https://rustsec.org/advisories/TEST-002.html');
    await expect(advisory2.locator('h3 a')).toHaveText('TEST-002');
    await expect(advisory2.locator('h3')).toContainText('Second test advisory');
    expect(await advisory2.locator('p').innerHTML()).toBe('This is the second test advisory with more details.');
```

app/templates/crate/security.gjs
djc: The reason is that I failed to closely read all of the comment, I suppose? Should be fixed now.

app/templates/crate/security.gjs
djc: Added a test for this.

PR #12315, opened 2025-11-12
Add `pubtime` field to index entries with environment variable toggle

Turbo87:
This adds support for including publication timestamps in index entries, controlled by the `INDEX_INCLUDE_PUBTIME` environment variable.

The field uses `skip_serializing_if` for backward compatibility - when disabled or `None`, the field is omitted from JSON output.

Usage:

```bash
INDEX_INCLUDE_PUBTIME=true  # Enable timestamps
INDEX_INCLUDE_PUBTIME=false # Disable (default)
```

### Related

- https://github.com/rust-lang/cargo/issues/15491

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1651,6 +1651,7 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
+ "chrono",
  "claims",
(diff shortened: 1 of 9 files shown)


eth3lbert (approved):
:shipit:

PR #12332, opened 2025-11-14
Extract `crates_io_version` crate with `commit()` and `user_agent()` fns

Turbo87:
This PR adds a new workspace crate to provide platform-agnostic version detection.

It currently delegates to `crates_io_heroku::commit()` but is designed to support additional deployment platforms in the future.

The PR updated all consumers to use `crates_io_version::commit()` instead of `crates_io_heroku::commit()`, abstracting away the Heroku-specific implementation details.

It also introduces a `user_agent()` fn, that can (and should) be used by all of our `reqwest::Client` instances, so that we can send requests with a consistent user-agent header like `crates.io/1234abc (https://crates.io)`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1607,6 +1607,7 @@ dependencies = [
  "async-trait",
  "claims",
  "crates_io_env_vars",
+ "crates_io_version",
  "http 1.3.1",
  "mockall",
  "mockito",
@@ -1634,6 +1635,7 @@ dependencies = [
 name = "crates_io_fastly"
 version = "0.0.0"
 dependencies = [
+ "crates_io_version",
  "reqwest",
  "secrecy",
  "thiserror 2.0.17",
@@ -1726,6 +1728,7 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "clap",
+ "crates_io_version",
  "reqwest",
  "secrecy",
  "serde",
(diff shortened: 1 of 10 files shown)


crates/crates_io_version/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+/// use crates_io_version::commit;
+///
+/// if let Ok(Some(commit)) = commit() {
+///     println!("Running commit: {}", commit);
+/// } else {
+///     println!("Commit SHA unknown");
+/// }
+/// ```
+pub fn commit() -> anyhow::Result<Option<String>> {
+    crates_io_heroku::commit()
```
eth3lbert, line 23: Maybe we could also introduce a new, alternative, platform-agnostic environment variable, such as `CRATESIO_COMMIT_HASH`, either as a fallback or as a first-party option? Later, we could potentially and gradually make `crates_io_heroku` an optional feature?

eth3lbert (approved):
I left a comment that isn't a blocker. The transition to a more platform-agnostic approach is great, thanks!

crates/crates_io_version/src/lib.rs
Turbo87: I thought about using https://docs.rs/vergen-gitcl/latest/vergen_gitcl/ instead, but that would mean recompiling the crate on every commit, which causes all the crates that use it to recompile too, which seemed a bit wasteful 😅

PR #12335, opened 2025-11-14
trustpub: Implement user interface for GitLab support

Turbo87:
This PR implements the majority of the frontend work for supporting GitLab in our Trusted Publishing flows. Specifically, it adjusts the crate settings section to list both GitHub and GitLab configurations, it allows crate owners to delete GitLab configurations, and it allows them to create new configurations for GitLab too.

### Related

- https://github.com/rust-lang/crates.io/issues/11987

### DIFF
--- a/app/templates/crate/settings/new-trusted-publisher.gjs
+++ b/app/templates/crate/settings/new-trusted-publisher.gjs
@@ -26,6 +26,7 @@ import LoadingSpinner from 'crates-io/components/loading-spinner';
           class='publisher-select base-input'
           data-test-publisher
           {{on 'change' @controller.publisherChanged}}
+          {{on 'change' (perform @controller.verifyWorkflowTask)}}
         >
           {{#each @controller.publishers as |publisher|}}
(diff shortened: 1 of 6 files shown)


app/templates/crate/settings/new-trusted-publisher.gjs
```diff
@@ -32,7 +34,7 @@ import LoadingSpinner from 'crates-io/components/loading-spinner';
       {{/let}}
 
       <div class='note'>
-        crates.io currently only supports GitHub, but we are planning to support other platforms in the future.
+        Select the CI/CD platform where your publishing workflow is configured.
```
LawnGnome, line 38: I don't know if we want to continue to hint that we may support other platforms in the future here. I'm fine with this as it is, but just flagging that we're removing that from here.

Turbo87: while supporting more platforms would be good, I don't think I can justify spending more time on adding additional ones, so that implementation would have to be done by the community.

since the previous statement said "planning to support other platforms in the future" I think it's probably fine to remove that part now that we do support an additional platform.

Issue #12361, opened 2025-11-19
Trusted Publishing Enforcement


Turbo87:
## Goal
Allow crate owners to restrict their crate to only accept publications via Trusted Publishing, preventing publishes using API tokens. This protects against leaked developer credentials.

## Key Design Decisions

**Visibility:** Public - The `trustpub_only` flag is visible to all users via `GET /api/v1/crates/{name}` for supply chain transparency.

**Authorization:** Any crate owner can toggle the flag using cookie auth or API token auth (non-legacy tokens with `trustpub` scope and matching crate scopes).

**Validation:** No validation when enabling - owners can enable the flag even without trusted publishers configured. The UI will show a warning in this case.

**Default:** `FALSE` for all crates - explicit opt-in required.

### GET `/api/v1/crates/{name}` (modified)
- Add field: `trustpub_only: bool`

### PATCH `/api/v1/crates/{name}` (new)
- Request: `{ "trustpub_only": true/false }`
- Auth: Crate owner only
- Response: 200 OK with updated crate data, or 403 if not owner
- Side effects: Email notification to all owners, audit log entry

### PUT `/api/v1/crates/new` (modified)
- Reject with 403 if `trustpub_only: true` and publish uses API token
- Error: "You tried to publish with an API token but this crate requires trusted publishing."

### Crate settings page
- Always show checkbox for `trustpub_only` flag
- If checked AND no trusted publisher configured: Show warning alert
- Checkbox triggers `PATCH /api/v1/crates/{name}`

### Backend
1. Database migration: Add `trustpub_only BOOLEAN NOT NULL DEFAULT FALSE` column
2. Update `Crate` model struct in `crates/crates_io_database/src/models/krate.rs`
3. Add `trustpub_only` field to `EncodableCrate` in `crates/crates_io_api_types/src/lib.rs`
4. Update publish endpoint to check `trustpub_only` flag and reject token-based publishes (with tests)
5. Implement `PATCH /api/v1/crates/{name}` endpoint with authorization check (crate owner + appropriate token scopes), email notifications, audit logging, and OpenAPI schema updates (with tests)

(implemented by https://github.com/rust-lang/crates.io/pull/12365)

### Frontend
1. Set up Ember Data to allow `PATCH /api/v1/crates/{name}` operations (with tests)
2. Add checkbox to crate settings page (with tests)
3. Add warning alert when checkbox is checked but no trusted publisher configured (with tests)

[...]

PR #12365, opened 2025-11-20
Add `trustpub_only` flag

Turbo87:
This PR implements the backend part of https://github.com/rust-lang/crates.io/issues/12361.

It adds a new `trustpub_only` column to the `crates` table, exposes the field in the API responses, adjusts the publish endpoint to check it, and implements a new `PATCH /api/v1/crates/{name}` endpoint to toggle it.

The UI implementation will follow in a dedicated PR.

### DIFF
--- a/src/controllers/krate/publish.rs
+++ b/src/controllers/krate/publish.rs
@@ -222,7 +222,7 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
         && matches!(auth, AuthType::Regular(_))
     {
         return Err(forbidden(
-            "You tried to publish with an API token but this crate requires trusted publishing.",
+            "New versions of this crate can only be published using Trusted Publishing (see https://crates.io/docs/trusted-publishing).",
         ));
     }
 
--- a/src/tests/krate/publish/auth.rs
+++ b/src/tests/krate/publish/auth.rs
@@ -101,7 +101,7 @@ async fn publish_with_token_rejected_when_trustpub_only() {
     let crate_to_publish = PublishBuilder::new("foo_trustpub_only", "1.0.0");
     let response = token.publish_crate(crate_to_publish).await;
     assert_snapshot!(response.status(), @"403 Forbidden");
-    assert_snapshot!(response.text(), @r#"{"errors":[{"detail":"You tried to publish with an API token but this crate requires trusted publishing."}]}"#);
+    assert_snapshot!(response.text(), @r#"{"errors":[{"detail":"New versions of this crate can only be published using Trusted Publishing (see https://crates.io/docs/trusted-publishing)."}]}"#);
 
     assert_that!(app.stored_files().await, is_empty());
     assert_that!(app.emails().await, is_empty());


src/controllers/krate/publish.rs
```diff
@@ -216,6 +216,16 @@ pub async fn publish(app: AppState, req: Parts, body: Body) -> AppResult<Json<Go
         AuthType::Regular(Box::new(auth))
     };
 
+    // Check if crate requires trusted publishing
+    if let Some(existing_crate) = &existing_crate
+        && existing_crate.trustpub_only
+        && matches!(auth, AuthType::Regular(_))
+    {
+        return Err(forbidden(
+            "You tried to publish with an API token but this crate requires trusted publishing.",
```
LawnGnome: I think people may be confused by "API token" as a term of art, so how about:

```suggestion
            "New versions of this crate can only be published using Trusted Publishing.",
```

src/email/templates/trustpub_only_changed/body.txt.j2
```diff
@@ -0,0 +1,25 @@
...
+Hello {{ recipient }}!
+
+{% if recipient == auth_user.gh_login -%}
+You changed the publishing method restriction for your crate "{{ krate.name }}".
+{%- else -%}
+crates.io user {{ auth_user.gh_login }} changed the publishing method restriction for a crate that you manage ("{{ krate.name }}").
+{%- endif %}
+
+{% if trustpub_only -%}
+This crate can now ONLY be published via Trusted Publishing. Publishing with API tokens has been disabled.
```
LawnGnome, line 13: (to follow on from my previous comment, I don't have the same concern over the use of "API token" here because the user is likely to have immediate context, since they or another owner were just thinking about trusted publishing)

src/tests/routes/crates/update.rs
```diff
@@ -0,0 +1,262 @@
...
+    CrateBuilder::new("foo", owner_id)
+        .expect_build(&mut conn)
+        .await;
+
+    let url = "/api/v1/crates/foo";
+    let body = serde_json::json!({ "trustpub_only": true });
+    let response = user.patch::<()>(url, body.to_string()).await;
+    assert_snapshot!(response.status(), @"200 OK");
+    assert_json_snapshot!(response.json(), {
+        ".crate.created_at" => "[datetime]",
```
LawnGnome, line 37: Should we also explicitly be checking `.crate.trustpub_only` at this point? (Here and below.)

src/controllers/krate/publish.rs
Turbo87: hmm... I specifically wanted to include the "API token" term to make it more obvious to the user what the problem is. I'm not sure what exactly you mean by "term of art". Do you think that crate owners that try to publish something are not aware of what the term "API token" means? 🤔

src/tests/routes/crates/update.rs
Turbo87: done :)

src/controllers/krate/update.rs
```diff
@@ -0,0 +1,199 @@
...
+                    krate,
+                    trustpub_only,
+                };
+
+                if let Err(err) = email.send(app, email_address).await {
+                    warn!("Failed to send trustpub_only notification to {email_address}: {err}");
+                }
+            }
+        }
+    }
```
eth3lbert, line 145: Not a blocker, but I'm wondering if we should check whether the given value is different from the existing value since we are sending a notification?

---

Nevertheless, I especially love the introduced notification mechanism here. Without it, a malicious user with a leaked API token could change the setting and publish a malicious version, which the crate owner might not notice. But with this mechanism, the crate owner has a chance to rotate the token and make amends!

eth3lbert (approved):
Left a minor nit. Overall LGTM! thanks!

src/controllers/krate/publish.rs
LawnGnome: > Do you think that crate owners that try to publish something are not aware of what the term "API token" means?

Honestly, yes, especially in a scenario where this might be an error message they get some time after they configured their crate to require TP. I could see a crate owner — particularly one that's stressed because they're trying to publish a fix for a vulnerability or critical bug — throwing their hands up and wondering how a TP token differs from an API token.

I don't mind including it somewhere in there in a second sentence, but I think we need to frontload the actionable part, which is that Trusted Publishing is required.

Turbo87: Alright, I've changed it to your suggestion, but with a link to https://crates.io/docs/trusted-publishing so that people can actually read up on what "Trusted Publishing" means in this error message.

src/controllers/krate/update.rs
Turbo87: > Not a blocker, but I'm wondering if we should check whether the given value is different from the existing value since we are sending a notification?

very good point, I've fixed that :)

src/controllers/krate/publish.rs
LawnGnome: Yeah, that's a good change. Makes it even more actionable. :+1:

PR #12387, opened 2025-11-24
CI: Don't run scheduled job on forks

marcoieni:
This job fails on forks, so you get an email about it, which is annoying


Turbo87:
> so you get an email about it, which is annoying

hmm, interesting. I haven't gotten any such emails so far for my own fork 😅 

but I'm not against it, so thanks! :)

PR #12410, opened 2025-11-28
templates/index: Reorder sections

teksturi:
When crates.io was new, it made sense to highlight newly published or recently updated crates. These days, newcomers are more likely looking for established and widely-used crates on the front page.

Reorder the front-page sections so the layout is more helpful and pleasant for new users.

image

### DIFF
--- a/app/templates/index.gjs
+++ b/app/templates/index.gjs
@@ -75,6 +75,71 @@ import placeholders from 'crates-io/helpers/placeholders';
     </button>
   {{else}}
     <div class='lists' data-test-lists>
+      <section data-test-most-downloaded>
+        <h2><LinkTo @route='crates' @query={{hash sort='downloads'}}>Most Downloaded</LinkTo></h2>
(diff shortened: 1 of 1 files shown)


Turbo87:
@rust-lang/crates-io any opinions?

I agree that our frontpage is currently not great. I'm not sure if just shuffling the order of the sections is the right solution though, since it causes churn without fixing the root cause of the problem. It might be worth checking out what other package registries show on their front pages.

LawnGnome:
I'm not opposed to thinking about what we should actually be presenting on the front page, but I think I agree that merely changing the order of the sections probably isn't _that_ useful in and of itself. (Moreover, I actually like the new crate list being fairly prominent; I think it encourages contribution.)

Turbo87:
we've discussed this pull request in the crates.io team meeting last week. the conclusion was that while we think that our frontpage could use some improvements, just reordering the sections would currently cause more disruption than it helps. we are open to other proposals, but will close this PR for now.

PR #12417, opened 2025-11-29
owners: explain GitHub org access policy link in team error

joshka:
Often this problem comes up when a github org is set to disallow third party access to the org.
This change is a messaging only change with a direct link to navigate to the offending problem rather than having to read and interpret the docs and then translate that into a github change.

### DIFF
--- a/src/controllers/krate/owners.rs
+++ b/src/controllers/krate/owners.rs
@@ -450,12 +450,18 @@ pub async fn create_or_update_github_team(
                 format!("Failed to decrypt GitHub token: {err}"),
             )
         })?;
-    let team = gh_client.team_by_name(org_name, team_name, &token).await
+    let team = gh_client
+        .team_by_name(org_name, team_name, &token)
+        .await
         .map_err(|_| {
             bad_request(format_args!(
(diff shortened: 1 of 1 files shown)


Turbo87:
is that the only case in which this can happen?

joshka:
> is that the only case in which this can happen?

The current `.map_err(|_| {...})`swallows the real error so currently there's no indication about why the failure happens that's exposed to the user. This makes it a tiny bit better, by giving a reasonable solution to one possible failure but it's not a solution to the error being swallowed.

I may be overindexing on the couple of times I've seen this error message (and had to look up how to fix it).

Turbo87 (commented):
@rust-lang/crates-io any thoughts on this?

IMHO it might be better to add that information to https://doc.rust-lang.org/cargo/reference/publishing.html#github-permissions, since we already link to that. It seems better suited to me to have it there since it allows us to have a longer explanation text there.

LawnGnome:
I wonder if we should be recording these errors somewhere so we can figure out if this is in fact the most common case. (It seems plausible to me, and if it is then I think we should both update the docs and tweak the message, but we don't really have the data to prove that right now.)

PR #12420, opened 2025-11-30
Add monthly and weekly downloads to recent downloads calculation

pksunkara:
We calculate the recent downloads using a materialized view that we refresh every day. But it's only for last 90 days.

Most of the other ecosystems like [NPM](https://shields.io/badges/npm-downloads), [PyPi](https://shields.io/badges/py-pi-downloads), [Packagist](https://shields.io/badges/packagist-downloads), [Hex](https://shields.io/badges/hex-pm-downloads) provide daily, weekly and monthly downloads. So, I added a weekly and monthly column in the recent downloads view and surfaced the data in the `/api/v1/crates/{name}` endpoint.

I have attached links for shields.io badges that show weekly and monthly downloads for each ecosystem. Once this is merged, I will add the same for Crates.io badges.

### DIFF
--- a/crates/crates_io_api_types/src/lib.rs
+++ b/crates/crates_io_api_types/src/lib.rs
@@ -314,6 +314,14 @@ pub struct EncodableCrate {
     #[schema(example = 456_789)]
     pub recent_downloads: Option<i64>,
 
+    /// The total number of downloads for this crate in the last month.
+    #[schema(example = 123_456)]
+    pub monthly_downloads: Option<i64>,
+
+    /// The total number of downloads for this crate in the last week.
+    #[schema(example = 56_789)]
+    pub weekly_downloads: Option<i64>,
+
     /// The "default" version of this crate.
     ///
     /// This version will be displayed by default on the crate's page.
@@ -380,6 +388,8 @@ impl EncodableCrate {
         exact_match: bool,
         downloads: i64,
         recent_downloads: Option<i64>,
+        monthly_downloads: Option<i64>,
+        weekly_downloads: Option<i64>,
     ) -> Self {
         let Crate {
             name,
@@ -440,6 +450,8 @@ impl EncodableCrate {
             created_at,
             downloads,
             recent_downloads,
+            monthly_downloads,
+            weekly_downloads,
             versions,
             keywords: keyword_ids,
             categories: category_ids,
@@ -477,6 +489,8 @@ impl EncodableCrate {
         exact_match: bool,
         downloads: i64,
         recent_downloads: Option<i64>,
+        monthly_downloads: Option<i64>,
+        weekly_downloads: Option<i64>,
     ) -> Self {
         Self::from(
             krate,
@@ -490,6 +504,8 @@ impl EncodableCrate {
             exact_match,
             downloads,
             recent_downloads,
+            monthly_downloads,
+            weekly_downloads,
         )
     }
 }
@@ -1187,6 +1203,8 @@ mod tests {
                 .and_utc(),
             downloads: 0,
             recent_downloads: None,
+            monthly_downloads: None,
+            weekly_downloads: None,
             default_version: None,
             num_versions: 0,
             yanked: false,
--- a/crates/crates_io_database/src/schema.patch
+++ b/crates/crates_io_database/src/schema.patch
@@ -40,7 +40,7 @@
          /// The `id` column of the `dependencies` table.
          ///
          /// Its SQL type is `Int4`.
-@@ -773,6 +767,24 @@
+@@ -773,6 +767,32 @@
  }
  
  diesel::table! {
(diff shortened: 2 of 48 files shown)


pksunkara:
I will fix the CI once a review is done.

pksunkara:
Refresh using the current query was taking 13.3s

```sql
SELECT crate_id, SUM(version_downloads.downloads) FROM version_downloads
INNER JOIN versions
  ON version_downloads.version_id = versions.id
WHERE version_downloads.date > date(CURRENT_TIMESTAMP - INTERVAL '90 days')
GROUP BY crate_id;
```

The query in this PR is taking 21.5s

```sql
SELECT
  crate_id,
  SUM(version_downloads.downloads),
  SUM(version_downloads.downloads) FILTER (WHERE version_downloads.date > date(CURRENT_TIMESTAMP - INTERVAL '30 days')),
  SUM(version_downloads.downloads) FILTER (WHERE version_downloads.date > date(CURRENT_TIMESTAMP - INTERVAL '7 days'))
FROM version_downloads
  INNER JOIN versions
    ON version_downloads.version_id = versions.id
  WHERE version_downloads.date > date(CURRENT_TIMESTAMP - INTERVAL '90 days')
  GROUP BY crate_id;
```

I improved it to take 13s by removing the timestamp conversions.

```sql
SELECT
  crate_id,
  SUM(version_downloads.downloads),
  SUM(version_downloads.downloads) FILTER (WHERE version_downloads.date > CURRENT_DATE - 30),
  SUM(version_downloads.downloads) FILTER (WHERE version_downloads.date > CURRENT_DATE - 7)
FROM version_downloads
  INNER JOIN versions
    ON version_downloads.version_id = versions.id
  WHERE version_downloads.date > CURRENT_DATE - 90
  GROUP BY crate_id;
```

---

**NOTE1**: Removing the timestamp conversions in the older query made it take 11.5s

**NOTE2**: We should probably remove the `(version_id, date)` index in `version_downloads` table because it's not actually being used since older data is archived and almost all the rows are used for calculation, thus postgres uses seq scan. That way, you can save some space. But, I will leave it upto you guys.

Turbo87:
interesting. on our production database this operation often take multiple minutes 😅

@eth3lbert any thoughts on this?

eth3lbert:
> interesting. on our production database this operation often take multiple minutes 😅

I'm also a bit worried about the time this will take in production, so if we're sure it won't take longer than before, I'm not opposed to it! :D"

---

Referring to the downloads table, I should probably find time to migrate our downloads table to `Declarative Partitioning`. However, before actually doing this, it would be better to test whether a new database with the same scale, specifications, and data runs much faster than the current one. This would suggest that our performance is likely being affected by the (BRIN) index.

pksunkara:
> interesting. on our production database this operation often take multiple minutes 😅

That would mean that we get more savings on actual numbers :)

Do you need to confirm the benchmark yourself before merging this? Or is there anything else I can do?

Turbo87:
I thought about this some more. The 90-day download data for the crates is already available on the API, just not in summed up form, but anyone who wants to calculate these values for a certain crate can already do so. If shields.io wants to display such values, can't they subset the per-day download numbers themselves and sum them up?

As I mentioned above, the materialized view is one of the more expensive pieces of our database and I'm slightly hesitant to extend it since we can't go back once we offer these values in our API.

I'm adding this issue to the agenda for our crates.io team meeting today.

pksunkara:
> If shields.io wants to display such values, can't they subset the per-day download numbers themselves and sum them up?

How would they get it though? I saw some endpoints but don't think any of them solved this.

Turbo87:
> I saw some endpoints but don't think any of them solved this.

https://crates.io/api/v1/crates/syn/downloads

pksunkara:
👍. If deciding to use it, it would need a filter parameter for dates.

Turbo87:
> If deciding to use it, it would need a filter parameter for dates.

why? you could filter it on the client side, no?

pksunkara:
It would be better to pass the filter to the db to reduce the load. In fact, I am wondering how much load constantly requesting this endpoint would add to the db versus the current PR. But, I can't prove anything without having metrics.

Since the current PR keeps the time taken for refresh at least the same, without knowing metrics for that endpoint, I would prefer the current PR from a pure performance viewpoint.

Turbo87:
we just discussed this PR again in the crates.io team meeting and concluded that we don't want to offer these values on the API right now, for the reasons given above:

> the materialized view is one of the more expensive pieces of our database and I'm slightly hesitant to extend it since we can't go back once we offer these values in our API

pksunkara:
I understand.

But I am curious by what you mean:

> the materialized view is one of the more expensive pieces of our database and I'm slightly hesitant to extend it

Wasn't the time the same after my PR? Are you talking about storage?

PR #12433, opened 2025-12-01
trustpub/list: Add support for `user_id` query parameter

Turbo87:
This allows users to avoid N+1 queries for Trusted Publishing configurations by adding support for a `?user_id=123` query parameter to the Trusted Publishing config listing endpoints. This user ID only accepts the user ID of the currently authenticated user and will list the configurations of all the crates that the user is an owner of. If a scoped API token is used, then only configurations for crates matching the crate scopes are returned.

### DIFF
--- a/src/controllers/trustpub/gitlab_configs/list.rs
+++ b/src/controllers/trustpub/gitlab_configs/list.rs
@@ -6,13 +6,13 @@ use crate::controllers::helpers::pagination::{
 use crate::controllers::krate::load_crate;
 use crate::controllers::trustpub::gitlab_configs::json::{self, ListResponse, ListResponseMeta};
 use crate::util::RequestUtils;
-use crate::util::errors::{AppResult, bad_request};
+use crate::util::errors::{AppResult, bad_request, forbidden};
 use axum::Json;
 use axum::extract::{FromRequestParts, Query};
 use crates_io_database::models::OwnerKind;
 use crates_io_database::models::token::EndpointScope;
 use crates_io_database::models::trustpub::GitLabConfig;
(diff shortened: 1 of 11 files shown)


eth3lbert:
🙋‍♂️ I'm curious if there's any difference between a token with its crate scope set to `*` and the proposed `allow_any_crate_scope`.

Turbo87:
> I'm curious if there's any difference between a token with its crate scope set to `*` and the proposed `allow_any_crate_scope`.

the logic is a bit complicated at the moment. without `allow_any_crate_scope()`, if you have an API token with crate scopes, but the endpoint is not using/setting a crate name, then the request will be denied.

I guess what I'm saying is that API tokens with `crate_scopes: ['*']` and `crate_scopes: null` currently behave differently. I'm not sure if that is good or not though 😅

eth3lbert:
> the logic is a bit complicated at the moment. without `allow_any_crate_scope()`, if you have an API token with crate scopes, but the endpoint is not using/setting a crate name, then the request will be denied.

Ah, I think I understand a bit about the introduced `allow_any_crate_scope()` here. Without it, a token with non-empty `crate_scopes` would probably be denied access to endpoints where the crate name isn't set or includes multiple crates. This would necessitate using a looser token without `crate_scopes` instead. With `allow_any_crate_scope()`, we have a chance to bypass the `crate_scope` during the checking phase (as determined by the developer), but the developer needs to handle the subsequent filtering themselves. This allows the user to make the request without being denied or needing to create a new token.

Turbo87:
exactly, yeah! :)

eth3lbert (approved):
I'm fine with this after understanding it, thanks for the explanation! :D

PR #12446, opened 2025-12-04
Enable use of Unix sockets to connect to DB  in tests

jtgeibel:
Currently it is possible to use a Unix socket to connect to the database by providing a URL such as `postgres:///db_name?host=/run/postgresql`. This works as expected, however the 5 `unhealthy_database` tests fail because the `ChaosProxy` assumes a TCP socket will be used. This is addressed in the 2nd commit which adds Unix socket support.

However, with that change it is now briefly possible that while running the tests, another user on the same localhost could potentially connect to the test database without credentials. It is unlikely that this is a realistic thread model for crates.io developers, however I address this in the 4th and 5th commits. See the commit descriptions for additional details and rationale.

The downside of this change is that the test configuration no longer mirrors the TCP configuration used in production. Alternatively, we could fail these 5 tests if a risky configuration is requested. (Or even decide to not worry about this "threat" at all.)

While working on this I noticed that `PgConnection` will automatically fallback to a Unix socket if no hostname is provided, however `AsyncPgConnection` does not. Some fallback logic is added so that the `postgres:///db_name` shorthand is consistently supported.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -26,6 +26,11 @@ pub async fn oneoff_connection() -> anyhow::Result<AsyncPgConnection> {
 pub fn connection_url(config: &config::DbPoolConfig) -> String {
     let mut url = Url::parse(config.url.expose_secret()).expect("Invalid database URL");
 
+    // Support `postgres:///db_name` shorthand for easier local development.
+    if url.host().is_none() {
+        maybe_append_url_param(&mut url, "host", "/run/postgresql");
+    }
+
     if config.enforce_tls {
         maybe_append_url_param(&mut url, "sslmode", "require");
     }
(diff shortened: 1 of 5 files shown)


src/db.rs
```diff
@@ -26,6 +26,11 @@ pub async fn oneoff_connection() -> anyhow::Result<AsyncPgConnection> {
 pub fn connection_url(config: &config::DbPoolConfig) -> String {
     let mut url = Url::parse(config.url.expose_secret()).expect("Invalid database URL");
 
+    // Support `postgres:///db_name` shorthand for easier local development.
+    if url.host().is_none() {
+        maybe_append_url_param(&mut url, "host", "/run/postgresql");
+    }
```
Turbo87, line 32: while I'm okay with this default for the testing setup, I'm not sure we want to use the same for production. Is there anything preventing you from specifying `/run/postgresql` directly in the `DATABASE_URL` env var?

PR #12487, opened 2025-12-13
Create new Svelte project

Turbo87:
In the last crates.io team meeting we decided to explore Svelte 5 as a replacement for our Ember.js frontend.

We evaluated a variety of options including SSR via Rust (and e.g. minijinja) and came to the conclusion that Svelte is currently our least bad option, mostly for DX reasons due to our tiny team size.

This PR:
- creates a basic example app in the `svelte/` directory
- adjusts our Rust middlewares to serve the app at `/svelte/`
- adds another `svelte` job to the GitHub Actions workflow, that runs when anything in the Svelte app is changed.

This allows us to build the Svelte app incrementally alongside the Ember.js app without needing an unreviewable big bang release. Once the new app is on par with the old one we can switch over to it.

### Related

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 dist/
+svelte/


eth3lbert (approved):
Looks like a good start 👍 thanks!

PR #12492, opened 2025-12-14
Improve contributing instructions

Its-Just-Nans:
Hi

This MR improves the contritbuting file(s)

I'm not sure about the importation steps

```
You can then import the database with

./script/import-database-dump.sh
```

Do you use the tmp/index-bare ? How?

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -268,7 +268,7 @@ postgres`). Generally, the problem is that by default the postgres server is
 > we'll help fix the problem and will add the solution to these
 > instructions!
 
-Another options is to use a standalone Docker container for Postgres.
+Another option is to use a standalone Docker container for Postgres:
 
 ```sh
 # example using postgres 16
@@ -399,6 +399,7 @@ example:
 
 ```console
 RUST_LOG=debug cargo run --bin background-worker
+```
 Then start a frontend that uses this backend by running this command in another
 terminal session (the frontend picks up frontend changes using live reload
 without a restart needed, and you can leave the frontend running while you


Its-Just-Nans:
Also, second question, is there a public roadmap for crates.io website?

script/import-database-dump.sh
```diff
@@ -6,7 +6,6 @@
 
 set -o errexit
 set -o nounset
-set -o pipefail
```
Its-Just-Nans, line 9: `#!/bin/sh` does not support pipefail

https://www.shellcheck.net/wiki/SC3040

package.json
```diff
@@ -186,8 +186,8 @@
     "last 1 years"
   ],
   "engines": {
-    "node": "24.12.0",
-    "pnpm": "10.25.0"
+    "node": "^24.12.0",
+    "pnpm": "^10.25.0"
```
Its-Just-Nans: To handle

```
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

Your pnpm version is incompatible with "tmp/crates.io".

Expected version: 10.23.0
Got: 10.25.0

This is happening because the package's manifest has an engines.pnpm field specified.
To fix this issue, install the required pnpm version globally.

To install the latest version of pnpm, run "pnpm i -g pnpm".
To check your pnpm version, run "pnpm -v".

Your Node version is incompatible with "tmp/crates.io".

Expected version: 24.11.1
Got: v24.12.0

This is happening because the package's manifest has an engines.node field specified.
To fix this issue, install the required Node version.
```

Turbo87: the "engines" are explicitly pinned because we use those fields to determine which versions to use during the deploy process. we are aware that this a little annoying at the moment, but this is a change that needs a bit more design before we can unpin them.

script/import-database-dump.sh
Turbo87: this does not seem to be related to the contributing instructions, so it should go into a dedicated PR. I also wonder if it would be better to instead change the shell to `bash` if `sh` does indeed not support `pipefail`.

docs/CONTRIBUTING.md
```diff
@@ -268,6 +268,15 @@ postgres`). Generally, the problem is that by default the postgres server is
 > we'll help fix the problem and will add the solution to these
 > instructions!
 
+Another options is to use a standalone Docker container for Postgres.
+
+```sh
+# example using postgres 15.15-trixie
+docker run -e POSTGRES_PASSWORD=password -p 5432:5432 postgres:15.15-trixie
```
Turbo87: we are using Postgres 16 in production, so these instructions should probably reflect that

docs/CONTRIBUTING.md
```diff
@@ -361,6 +370,12 @@ Set up the git repo for the crate index by running:
 ./script/init-local-index.sh
 ```
 
+You can then import the database with
+
+```console
+./script/import-database-dump.sh
+```
+
```
Turbo87, line 380: While it's probably a good idea to document this option, I'm not sure the "Setting up the git index" section is the right place for it 😅

docs/CONTRIBUTING.md
```diff
@@ -376,6 +391,14 @@ Then start the background worker (which will process uploaded READMEs):
...
 ```
 
+Since these programs use the logging framework, you can enable debug logging by
+setting the `RUST_LOG` environment variable to `debug` before running them, for
+example:
+
+```console
+RUST_LOG=debug cargo run --bin background-worker
+```
+
```
Turbo87:
```suggestion
Since crates.io is using the `tracing` crate, you can enable debug logging by
setting the `RUST_LOG` environment variable to `debug` before running them, for
example:

```console
RUST_LOG=debug cargo run --bin background-worker
```

```

**Comment by Its-Just-Nans** [comment]
About https://github.com/rust-lang/crates.io/pull/12492#discussion_r2618558932

I did it in this MR because this script is used in CONTRIBUTING.md

I will split in another MR



Also: `Do you use the tmp/index-bare ? How?`

File: script/import-database-dump.sh (hunk 1 above)
Comment (Its-Just-Nans) [reply script/import-database-dump.sh:9]: Split in https://github.com/rust-lang/crates.io/pull/12513

File: package.json (hunk 2 above)
Comment (Its-Just-Nans) [reply package.json]: yes that's kinda not helping where trying to contribute :/

File: docs/CONTRIBUTING.md [hunk 6]
```diff
@@ -376,6 +393,12 @@ Then start the background worker (which will process uploaded READMEs):
 cargo run --bin background-worker
 ```
 
+Since crates.io is using the `tracing` crate, you can enable debug logging by
+setting the `RUST_LOG` environment variable to `debug` before running them, for
+example:
+
+```console
+RUST_LOG=debug cargo run --bin background-worker
```
Turbo87, line 401:
````suggestion
RUST_LOG=debug cargo run --bin background-worker
```
````

File: docs/CONTRIBUTING.md [hunk 7]
```diff
@@ -268,6 +268,15 @@ postgres`). Generally, the problem is that by default the postgres server is
 > we'll help fix the problem and will add the solution to these
 > instructions!
 
+Another options is to use a standalone Docker container for Postgres.
```
Turbo87:
```suggestion
Another option is to use a standalone Docker container for Postgres:
```

package.json
Turbo87: https://github.com/rust-lang/crates.io/pull/12538 should hopefully help fix this issue

Its-Just-Nans: Nice!

Thanks for the update

PR #12513, opened 2025-12-16
Use `env bash` in `script/import-database-dump.sh`

Its-Just-Nans:
Related to https://github.com/rust-lang/crates.io/pull/12492#discussion_r2618558932

- https://www.shellcheck.net/wiki/SC3040

Other options is to use `/bin/bash` instead of `/bin/sh`

### DIFF
--- a/script/import-database-dump.sh
+++ b/script/import-database-dump.sh
@@ -1,4 +1,4 @@
-#!/bin/env bash
+#!/usr/bin/env bash
 
 # Downloads the database dump tarball from crates.io and imports it
 # into the `cargo_registry` database. If the database already exists it


Turbo87:
Thanks for the PR! You're correct that `pipefail` is not POSIX-compliant.

However, I'd prefer changing the shebang to `#!/usr/bin/env bash` instead of removing `pipefail`. The script has a pipeline on line 73:

```sh
python3 "$(dirname "$0")/infer-database-dump-version.py" -m "$DUMP_PATH/metadata.json" | psql -a "$DATABASE_NAME"
```

Without `pipefail`, if the Python script fails, `psql` might still exit 0 (having received partial or no input), and the script would continue silently. With `pipefail`, we'd catch this failure.

We already use bash in `script/create-diff-link.sh`, so there's precedent for it in this repo.

Its-Just-Nans:
Okay thanks for the info

Fixed

Issue #12515, opened 2025-12-16
Migrate frontend to Svelte 5


Turbo87:
In a recent crates.io team meeting, we decided to explore Svelte 5 as a replacement for our Ember.js frontend. We evaluated several options including server-side rendering via Rust (e.g., with minijinja) and concluded that Svelte is currently our best option, primarily for developer experience reasons given our small team size.

Rather than a big-bang rewrite, we're taking an incremental approach. The new Svelte app lives in the `/svelte/` directory and is served at `/svelte/` alongside the existing Ember.js app. This allows us to migrate components and routes one at a time, keeping each PR reviewable. Once the Svelte app reaches feature parity with the Ember.js app, we'll swap it to be the primary frontend.

The Svelte app is a work in progress and not yet production-ready. Review feedback on architecture and patterns is welcome as we establish conventions.

This issue serves as a tracking issue for the migration. Related pull requests will reference it to provide context.

webJose:
Hello.  I know I'm butting in where nobody called me.  Apologies in advance.  I just wanted to tell you guys that, generally speaking, one can migrate from one front-end framework/library to another at any pace by using micro-frontend technologies.  Not Module Federation, though, as exporting one Svelte component is not immediately usable in Ember as I understand it.  Module Federation is great when mixing micro-frontends of the same technology.

If you guys ever need something, me and my team successfully used `single-spa` micro-frontends to migrate away from React and into Svelte.

I'm also happening to be creating a more compact and modern micro-frontends set of libraries and Vite plug-ins.  The timing is almost right, or perhaps I'm a bit late:  [CollageJS](https://github.com/collagejs/collagejs).  Since I haven't documented the library, just ask questions directly in the Discussions section of the repository.  I'll answer and help out.

*CollageJS* is currently under the final steps of development.  It has passed the PoC and is just waiting for the finalization of the Vite plug-ins and the creation of adapter NPM packages for the various frameworks.  Currently, I only have the Svelte v5 adapter ready.  For your case, we would need an Ember.js adapter.

Thanks for the time you lent me reading this.  Happy coding in Svelte.  Svelte is amazing.

Turbo87:
thanks for the comment, but micro-frontends does not seem like a direction we want to go into for this. I've considered using Svelte components from within the Ember.js app, but ultimately that would just complicate the stack unnecessarily and solve very little of the actual migration problems (data loading, auth, etc.).

daniel-pfeiffer:
[...]

PR #12516, opened 2025-12-16
svelte: Add stub routes for all current Ember.js routes

Turbo87:
This commit/PR recreates the current [Ember.js route structure](https://github.com/rust-lang/crates.io/blob/main/app/router.js) in the Svelte app with stub routes. These routes currently only show the URL parameters. This allows us to use the Svelte `resolve()` fn from now on to create type-checked in-app links.

### Related


Turbo87:
> except the catch-all

yeah, I wasn't sure about that one yet. I have a feeling that we might not need it if we use the native error route support.

PR #12518, opened 2025-12-16
svelte: Migrate `SearchForm` component

Turbo87:
Bildschirmfoto 2025-12-16 um 15 06 20

### Related


svelte/src/lib/components/SearchForm.svelte
```diff
@@ -0,0 +1,183 @@
...
+  let inputElement: HTMLInputElement | undefined = $state();
+
+  function updateSearchValue(event: Event) {
+    searchValue = (event.target as HTMLInputElement).value;
+  }
+
+  function search(event: SubmitEvent) {
+    event.preventDefault();
+    // eslint-disable-next-line svelte/no-navigation-without-resolve -- resolve() doesn't support query params
+    goto(`${resolve('/search')}?q=${encodeURIComponent(searchValue)}&page=1`);
```
eth3lbert, line 25: This should be resolved once the ongoing PR that tried to make `resolve()` support an optional search params is merged.

svelte/src/lib/components/SearchForm.svelte
```diff
@@ -0,0 +1,183 @@
...
+<svelte:window onkeydown={handleKeydown} />
+
+<!-- svelte-ignore a11y_autofocus -->
+<form
+  action="/search"
+  role="search"
+  autocapitalize="off"
+  autocomplete="off"
+  spellcheck="false"
+  data-test-search-form
```
eth3lbert, line 54: Ah, I just thought of this: we currently don't remove the test attribute for SvelteKit, do we? And I'm not sure if it's time to rewrite the test id to the generally known `data-testid` format (we currently use `data-test-id` for compatibility). The `data-testid` format would offer more native tool support and be easier to operate.

svelte/src/lib/components/SearchForm.svelte
Turbo87: yeah, I had found a related issue, but it looked like the discussion had been ongoing for a while already :-/

svelte/src/lib/components/SearchForm.svelte
Turbo87: hmm... good point. 

I think I'd rather keep the diff between the apps as small as possible though. so either we migrate the Ember.js app to `data-testid` too first, or we use `data-test-...` in Svelte for now and switch over once we removed the Ember.js app.

eth3lbert: fair enough!

PR #12538, opened 2025-12-18
Use `.tool-versions` file to pin Node.js and pnpm versions

Turbo87:
The `.tool-versions` file is a conventional format from the `asdf` tool (see https://asdf-vm.com/guide/getting-started.html#_6-set-a-version) and is essentially a simple `<tool> <version>` mapping. 

https://github.com/Turbo87/heroku-buildpack-crates-io/commit/76b621d6189070f88dbf0f90879c4843fcbf3e08 adjusts our buildpack to read the Node.js and pnpm versions from that files instead of `pnpm`, which also means that we no longer need to install `jq` since `grep` + `awk` is sufficient. Note that the parser is very basic at the moment and currently does not support trailing comments after the version string.

As for the "why": it has been fairly annoying to locally always have to update the Node.js and pnpm versions to match what crates.io has pinned. This PR also relaxes the `engines` declarations, so that now only the major version needs to match for local development, while the deployment process and CI are using the pinned versions.


djc (approved):
Looks like a nice win!

PR #12549, opened 2025-12-18
svelte: Improve `$props()` usage

Turbo87:
This replaces `[key: string]: unknown;` in the `Props` interfaces with `HTMLAttributes<HTMLDivElement>`. That allows us to get proper auto-completion for the regular HTML element props.

This PR also ensures that the `class` attributes of the components are merged correctly when passed in from the outside.

### Related

### DIFF
--- a/svelte/src/lib/components/PageHeader.svelte
+++ b/svelte/src/lib/components/PageHeader.svelte
@@ -11,10 +11,10 @@
     children?: Snippet;
   }
 
-  let { title, suffix, showSpinner = false, children, ...others }: Props = $props();
+  let { title, suffix, showSpinner = false, children, class: className, ...others }: Props = $props();
 </script>
 
-<div data-test-page-header class="header" {...others}>
+<div data-test-page-header class={['header', className]} {...others}>
   {#if children}
     {@render children()}
   {:else}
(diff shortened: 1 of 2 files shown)


svelte/src/lib/components/PageHeader.svelte
```diff
@@ -1,20 +1,20 @@
...
     children?: Snippet;
-    [key: string]: unknown;
   }
 
-  let { title, suffix, showSpinner = false, children, ...others }: Props = $props();
+  let { title, suffix, showSpinner = false, children, class: className, ...others }: Props = $props();
 </script>
 
-<div data-test-page-header class="header" {...others}>
+<div data-test-page-header class="header {className}" {...others}>
```
eth3lbert:
```suggestion
<div data-test-page-header class={["header", className]} {...others}>
```

svelte/src/lib/components/TextContent.svelte
```diff
@@ -1,16 +1,16 @@
...
     children: Snippet;
-    [key: string]: unknown;
   }
 
-  let { boxed = false, children, ...others }: Props = $props();
+  let { boxed = false, children, class: className, ...others }: Props = $props();
 </script>
 
-<div class="wrapper" class:boxed {...others}>
+<div class="wrapper {className}" class:boxed {...others}>
```
eth3lbert:
```suggestion
<div class={["wrapper", className, { boxed }]} {...others}>
```

Turbo87: thanks, that's a good idea! :)

PR #12561, opened 2025-12-20
svelte: Port `LoadingSpinner` component from Ember.js app

Turbo87:
Bildschirmfoto 2025-12-20 um 09 24 21

Bildschirmfoto 2025-12-20 um 09 24 37

### Related

### DIFF
--- a/svelte/src/lib/components/PageHeader.svelte
+++ b/svelte/src/lib/components/PageHeader.svelte
@@ -2,7 +2,7 @@
   import type { Snippet } from 'svelte';
   import type { HTMLAttributes } from 'svelte/elements';
 
-  // TODO: import LoadingSpinner from './LoadingSpinner.svelte';
+  import LoadingSpinner from '$lib/components/LoadingSpinner.svelte';
 
   interface Props extends HTMLAttributes<HTMLDivElement> {
     title?: string;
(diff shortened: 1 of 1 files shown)


svelte/src/lib/components/PageHeader.svelte
```diff
@@ -24,8 +24,7 @@
         <small class="suffix">{suffix}</small>
       {/if}
       {#if showSpinner}
-        <!-- TODO: <LoadingSpinner class="loading-spinner" data-test-spinner /> -->
-        <span class="loading-spinner" data-test-spinner>Loading...</span>
+        <LoadingSpinner style="margin-left: var(--space-2xs)" data-test-spinner />
```
eth3lbert, line 27: It's a pity that Svelte doesn't have many options for overriding a child component's styles. I hope a better solution comes along someday! 🥲

Turbo87: yeah, that part is certainly annoying... :-/

svelte/src/lib/components/LoadingSpinner.stories.svelte
```diff
@@ -0,0 +1,25 @@
...
+  import LoadingSpinner from './LoadingSpinner.svelte';
+
+  const { Story } = defineMeta({
+    title: 'LoadingSpinner',
+    component: LoadingSpinner,
+    tags: ['autodocs'],
+    argTypes: {
+      theme: {
+        control: 'inline-radio',
+        options: [undefined, 'light'],
```
eth3lbert, line 13: Ah, is the undefined option expected here? I thought it would be `dark` instead. 🤔

Turbo87: it basically means "don't assign a value and use the default"

PR #12562, opened 2025-12-20
svelte: Port `Placeholder` component from Ember.js app

Turbo87:
Bildschirmfoto 2025-12-20 um 09 27 43

### Related

### DIFF
--- a/svelte/src/lib/components/Placeholder.stories.svelte
+++ b/svelte/src/lib/components/Placeholder.stories.svelte
@@ -1,0 +1,32 @@
+<script module>
+  import { defineMeta } from '@storybook/addon-svelte-csf';
+
+  import Placeholder from './Placeholder.svelte';
+
+  const { Story } = defineMeta({
+    title: 'Placeholder',
+    component: Placeholder,
+    tags: ['autodocs'],
+  });
+</script>
+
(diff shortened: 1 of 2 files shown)


svelte/src/lib/components/Placeholder.stories.svelte
```diff
@@ -0,0 +1,32 @@
...
+  });
+</script>
+
+<Story name="Default" args={{ style: 'width: 200px; height: 20px; border-radius: 4px;' }} />
+
+<Story name="Multiple Lines" asChild>
+  <div style="display: flex; flex-direction: column; gap: 8px;">
+    <Placeholder style="width: 100%; height: 20px; border-radius: 4px;" />
+    <Placeholder style="width: 80%; height: 20px; border-radius: 4px;" />
+    <Placeholder style="width: 90%; height: 20px; border-radius: 4px;" />
```
eth3lbert: Based on the example shown, if this scenario is common, I'd prefer providing props or CSS variables over an inline style approach.

Turbo87: I've checked the current Ember.js usage and there we use `class` to set the width, height, margins, opacity, and border-radius.

I've added `width`, `height`, `opacity` and `radius` props now. margins and other stuff like that should still be set from outside the component IMHO.

PR #12579, opened 2025-12-22
playwright: Replace `ember` import with submodule imports

Turbo87:
Importing from `ember` was causing a few IDE warnings for me. First importing `ember-source/types` and then using `@ember/...` appears to be the currently recommended way for importing types.


Turbo87:
sigh... apparently it does not work 🙄

Turbo87:
looks like using `/// <reference types="ember-source/types" />` instead of a regular import works as intended 🎉 

thank you Claude for figuring this out... 😅

PR #12583, opened 2025-12-23
Implement HTML emails

Turbo87:
One of the original goals of moving to a template system for our emails was to support HTML emails. This PR finally implements them.

Each email type now has an additional `body.html.j2` template that is used to generate the HTML content for the email. If the `HTML_EMAILS_ENABLED` environment variable is set to `true` then the generated emails will contain multipart content with plaintext and HTML content. Once we have verified that this is not causing any issues in production this feature flag can be removed again.

Note that the email templates have intentionally been kept quite basic for now. We can start to extend them once we verified that everything works as intended.

/cc @paolobarbolini

### DIFF
--- a/src/email.rs
+++ b/src/email.rs
@@ -2,8 +2,8 @@ use crate::Env;
 use crate::config;
 use lettre::address::Envelope;
 use lettre::message::Mailbox;
+use lettre::message::MultiPart;
 use lettre::message::header::ContentType;
-use lettre::message::{MultiPart, SinglePart};
 use lettre::transport::file::AsyncFileTransport;
 use lettre::transport::smtp::AsyncSmtpTransport;
 use lettre::transport::smtp::authentication::{Credentials, Mechanism};
@@ -215,11 +215,7 @@ impl Emails {
             .subject(subject);
 
         let message = if self.html_emails_enabled {
-            builder.multipart(
-                MultiPart::alternative()
-                    .singlepart(SinglePart::plain(body_text))
-                    .singlepart(SinglePart::html(body_html)),
-            )?
+            builder.multipart(MultiPart::alternative_plain_html(body_text, body_html))?
         } else {
             builder.header(ContentType::TEXT_PLAIN).body(body_text)?
         };


src/email.rs
```diff
@@ -184,19 +208,28 @@ impl Emails {
...
-            .header(ContentType::TEXT_PLAIN)
-            .body(body)?;
+            .subject(subject);
+
+        let message = if self.html_emails_enabled {
+            builder.multipart(
+                MultiPart::alternative()
+                    .singlepart(SinglePart::plain(body_text))
+                    .singlepart(SinglePart::html(body_html)),
+            )?
```
paolobarbolini: There's a small shortcut you can take by using [`MultiPart::alternative_plain_html`](https://docs.rs/lettre/0.11.19/lettre/message/struct.MultiPart.html#method.alternative_plain_html).

```suggestion
            builder.multipart(MultiPart::alternative_plain_html(body_text, body_html))?
```

src/email/templates/base.html.j2
paolobarbolini, line 1: This seems very basic, and could create problems with some anti-spam systems which expect a proper HTML document. If I remember correctly even tools like https://www.mail-tester.com pointed this aspect out.

Another basic test I found interesting in the past was to send emails to Fastmail, download the .eml for the message and look for the [`X-Spam-hits` header](https://www.fastmail.com/blog/x-spam-hits-header-has-spam-scores-added/). This is what it will looks like for a crates.io email from a few weeks ago:

```
X-Spam-hits: BAYES_40 -0.001, DCC_CHECK 1.1, ME_SENDERREP_HAM -2,
  RCVD_IN_DNSWL_NONE -0.0001, RCVD_IN_MSPIKE_H5 0.001,
  RCVD_IN_MSPIKE_WL 0.001, SPF_HELO_NONE 0.001, SPF_PASS -0.001,
  LANGUAGES en, BAYES_USED user, SA_VERSION 4.0.1
```

You can then search for the meaning of rules with a negative score or have an LLM explain it to you.

paolobarbolini (changes requested):

src/email/templates/base.html.j2
Turbo87: > This seems very basic, and could create problems with some anti-spam systems which expect a proper HTML document.

interesting. I looked at some of the emails GitHub was sending and they do not seem to use a full HTML structure either, which is why I implemented it this way 🤔

Turbo87: > If I remember correctly even tools like https://www.mail-tester.com pointed this aspect out.

I just ran a quick test and that website appears to be happy with the current state (see https://www.mail-tester.com/test-d5sh981cg).

I used https://htmltest.email/ to send the email to https://www.mail-tester.com, but I assume that shouldn't matter for this specific check?

paolobarbolini: Interesting. I guess what I may have seen in the past was a rule in `X-Spam-hits` penalizing emails with incomplete or broken HTML.

Turbo87: let's give it a try with this simple structure for now. if we notice on staging that it doesn't work as intended we can still adjust it later :)

Turbo87: I've tested it out on staging and it appears to be working fine so far. I'll enable it on prod too. If we see any complaints we can always disable it again via the env var.

PR #12597, opened 2025-12-27
email: Add view actions

Turbo87:
This commit adds "view actions" based on schema.org linked data to some of our email templates. This causes some email clients like Gmail to display a corresponding button in the email list.

This was inspired by GitHub doing something similar with e.g. "View Discussion" view actions:

Bildschirmfoto 2025-12-27 um 07 27 26

/cc @paolobarbolini 

### Related

- https://github.com/rust-lang/crates.io/pull/12583

### DIFF
--- a/src/email.rs
+++ b/src/email.rs
@@ -353,6 +353,24 @@ mod tests {
         <p>You can also go to <a href="https://crates.io/me/pending-invites">https://crates.io/me/pending-invites</a> to manage all of your crate ownership invitations.</p>
 
         <p>--<br>The crates.io Team</p>
+        <script type="application/ld+json">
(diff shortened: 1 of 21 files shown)


Turbo87:
turns out this needs more than just the metadata in the email since domains also need to be whitelisted on the GMail side 😿 

more information available at https://www.twilio.com/en-us/blog/insights/how-to-use-gmail-actions-in-your-email-program

I'll work on the registration process in the next couple of days and then we'll see if we can get whitelisted.

Turbo87:
Bildschirmfoto 2026-01-02 um 10 37 24

and now we wait...

PR #12598, opened 2025-12-27
tests/routes/support: Simplify footer support link tests

Turbo87:
This PR simplifies our tests for the "Support" link in the page footer. Previously it seems that we were primarily testing the behavior of a mocked application, instead of the actual application behavior. Instead, we now check for the support link `href` in a couple of relevant application states (startup, on support page, navigated away again).

This also brings these tests closer to black box testing, which allows us to use a different application implementation (i.e. Svelte) underneath the tests.


eth3lbert (approved):
Awesome, thanks!

PR #12607, opened 2025-12-27
svelte: Implement `/keywords` route

Turbo87:
This PR ports the `/keywords` route from Ember.js to Svelte, including the two necessary components: `ResultsCount` and `Pagination`.

Not included in this PR are the `SortDropdown` and `Tooltip` components, which are left as `TODO` comments.

### Related

### DIFF
--- a/svelte/src/routes/keywords/+page.svelte
+++ b/svelte/src/routes/keywords/+page.svelte
@@ -1,2 +1,67 @@
-<h1>Keywords</h1>
-<p>Stub route for /keywords</p>
+<script lang="ts">
+  import { resolve } from '$app/paths';
+
+  import PageHeader from '$lib/components/PageHeader.svelte';
+  import Pagination from '$lib/components/Pagination.svelte';
+  import ResultsCount from '$lib/components/ResultsCount.svelte';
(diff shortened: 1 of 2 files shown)


svelte/src/lib/utils/pagination.ts
```diff
@@ -0,0 +1,66 @@
...
+export function calculatePagination(
+  currentPage: number,
+  perPage: number,
+  totalItems: number,
+  maxPages?: number,
+): PaginationState {
+  let currentPageStart = totalItems === 0 ? 0 : (currentPage - 1) * perPage + 1;
+  let currentPageEnd = Math.min(currentPage * perPage, totalItems);
+
+  let calculatedPages = Math.ceil(totalItems / perPage || 1);
```
eth3lbert, line 23: This would result in an `infinite` value if `perPage` is `0`.

Turbo87: same as before: https://github.com/rust-lang/crates.io/blob/62209ecba038bfaae97edcf514aea85a9e55f92c/app/utils/pagination.js#L19

😅

eth3lbert: I guess it's good to go then.

PR #12610, opened 2025-12-27
svelte: Add `Dropdown` components

Turbo87:
Roughly similar implementation to the Ember.js variant, but a slightly more Svelty API:

Bildschirmfoto 2025-12-27 um 19 56 19

### Related

### DIFF
--- a/svelte/src/lib/components/dropdown/Dropdown.svelte
+++ b/svelte/src/lib/components/dropdown/Dropdown.svelte
@@ -1,0 +1,51 @@
+<script lang="ts">
+  import type { Snippet } from 'svelte';
+  import type { HTMLAttributes } from 'svelte/elements';
+  import type { DropdownContext } from './context';
+
+  import { clickOutside } from '$lib/attachments/click-outside';
+  import { setDropdown } from './context';
+
+  interface Props extends HTMLAttributes<HTMLDivElement> {
+    children: Snippet;
+  }
+
+  let { children, class: className, ...restProps }: Props = $props();
+  let uniqueId = $props.id();
+
+  let isExpanded = $state(false);
+
+  let context: DropdownContext = {
+    get isExpanded() {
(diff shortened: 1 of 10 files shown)


svelte/src/lib/components/dropdown/Dropdown.svelte
```diff
@@ -0,0 +1,52 @@
...
+
+  let context: DropdownContext = {
+    get isExpanded() {
+      return isExpanded;
+    },
+    triggerId,
+    contentId,
+    toggle: () => (isExpanded = !isExpanded),
+    close: () => (isExpanded = false),
+  };
```
eth3lbert, line 26: use [`$props.id()`](https://svelte.dev/docs/svelte/$props#$props.id()) instead.

```suggestion
  let id = $props.id();
  let isExpanded = $state(false);
  let triggerId = $derived(`dropdown-trigger-${id}`);
  let contentId = $derived`dropdown-content-${id}`);

  let context: DropdownContext = {
    get isExpanded() {
      return isExpanded;
    },
    get triggerId() {
      return triggerId
    },
    get contentId() {
      return contentId
    },
    toggle: () => (isExpanded = !isExpanded),
    close: () => (isExpanded = false),
  };
```

svelte/src/lib/components/dropdown/clickOutside.ts
eth3lbert, line 1: I think this is more of a general utility function, so it might be reasonable to move it to a global utils folder. Also, consider rewriting this using [attachments](https://svelte.dev/docs/svelte/@attach) as recommended by the [official docs](https://svelte.dev/docs/svelte/use).
> Since Svelte 5.29, attachments are preferred because they are more flexible and composable.

Turbo87: done :)

PR #12620, opened 2025-12-29
cloudfront: Route invalidations to correct distribution

### LINKED ISSUE #12609: Current `db-dump.tar.gz` looks old
The current `db-dump.tar.gz` available at `https://static.crates.io/db-dump.tar.gz` looks like it was created on December 18 (nine days ago):
```sh
$ tar tf db-dump.tar.gz 
2025-12-18-020009
2025-12-18-020009/README.md
2025-12-18-020009/export.sql
2025-12-18-020009/import.sql
2025-12-18-020009/metadata.json
...
```
The `metadata.json` file seems to support this as well:
```sh
$ cat ./metadata.json
{
  "timestamp": "2025-12-18T02:00:09.233254004Z",
  "crates_io_commit": "a111eb647195af7365a36296a62233dc1db29360"
}
```
Is this expected?

Sorry if this issue is unnecessary noise.

Turbo87:
Add support for two CloudFront distributions (index.crates.io and static.crates.io) instead of using a single distribution for all invalidations.

- Add `distribution` column to `cloudfront_invalidation_queue` table
- Create `CloudFrontDistribution` enum with Diesel traits for type-safe database operations
- Update `CloudFront` client to require both distribution IDs via `CLOUDFRONT_DISTRIBUTION_ID_INDEX` and `CLOUDFRONT_DISTRIBUTION_ID_STATIC`
- Route index metadata invalidations to the index distribution
- Route static file invalidations (crates, readmes, OG images, db dumps, RSS feeds) to the static distribution

Fixes #12609

### DIFF
--- a/src/worker/jobs/process_cloudfront_invalidation_queue.rs
+++ b/src/worker/jobs/process_cloudfront_invalidation_queue.rs
@@ -1,7 +1,7 @@
 use crate::cloudfront::{CloudFront, CloudFrontError};
 use crate::worker::Environment;
 use anyhow::Context;
-use crates_io_database::models::CloudFrontInvalidationQueueItem;
+use crates_io_database::models::{CloudFrontDistribution, CloudFrontInvalidationQueueItem};
 use crates_io_worker::BackgroundJob;
 use diesel_async::AsyncPgConnection;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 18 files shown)


src/worker/jobs/process_cloudfront_invalidation_queue.rs
```diff
@@ -68,12 +79,17 @@ impl ProcessCloudfrontInvalidationQueue {
     async fn invalidate_with_backoff(
         &self,
         cloudfront: &CloudFront,
+        distribution: CloudFrontDistribution,
```
LawnGnome, line 82: This implements backoff as global, not per-distribution. Do we know how the rate limits apply?

(That said, this _probably_ won't matter in practice.)

LawnGnome (approved):
One question, but LGTM otherwise.

src/worker/jobs/process_cloudfront_invalidation_queue.rs
Turbo87: > Do we know how the rate limits apply?

no idea, to be honest. but as long as we only have two distributions, I also assume this won't matter too much.

PR #12622, opened 2025-12-29
svelte: Extract `SearchForm` context

Turbo87:
This allows us to overwrite the search field value from within other routes (e.g. the category route like in the Ember.js app).

### Related

### DIFF
--- a/svelte/src/routes/search/+page.svelte
+++ b/svelte/src/routes/search/+page.svelte
@@ -1,2 +1,19 @@
+<script lang="ts">
+  import { onMount } from 'svelte';
+  import { page } from '$app/state';
+
+  import { getSearchFormContext } from '$lib/search-form.svelte';
+
+  let searchFormContext = getSearchFormContext();
+
+  // Sync URL ?q= param to header search state
+  $effect(() => {
+    searchFormContext.value = page.url.searchParams.get('q') ?? '';
+  });
+  onMount(() => () => {
+    searchFormContext.value = '';
+  });
+</script>
+
 <h1>Search</h1>
 <p>Stub route for /search</p>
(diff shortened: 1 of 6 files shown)


svelte/src/routes/search/+page.svelte
```diff
@@ -1,2 +1,15 @@
...
+  import { onMount } from 'svelte';
+  import { page } from '$app/state';
+
+  import { getHeaderSearch } from '$lib/header-search.svelte';
+
+  let headerSearch = getHeaderSearch();
+
+  // Sync URL ?q= param to header search state
+  $effect(() => headerSearch.setValue(page.url.searchParams.get('q') ?? ''));
+  onMount(() => () => headerSearch.clear());
```
Turbo87: this is using `$effect()` so that it a) runs on the server too (compared to `onMount()`) and b) updates the value also when staying on the same page but the query param changes.

this is using `onMount()` to set up the destructor that clears the search term because `onMount()` does **not** run on the server and on the server there is no need for us to reset the value.

svelte/src/routes/search/+page.svelte
```diff
@@ -1,2 +1,19 @@
...
+
+  let searchFormContext = getSearchFormContext();
+
+  // Sync URL ?q= param to header search state
+  $effect(() => {
+    searchFormContext.value = page.url.searchParams.get('q') ?? '';
+  });
+  onMount(() => () => {
+    searchFormContext.value = '';
+  });
```
eth3lbert, line 15:
```suggestion
  onDestroy(() => {
    searchFormContext.value = '';
  });
```

eth3lbert (approved):

svelte/src/routes/search/+page.svelte
Turbo87: [see above](https://github.com/rust-lang/crates.io/pull/12622#discussion_r2651286331) 

I hope my reasoning makes sense 😅

eth3lbert: ah my bad, I forgot that `onDestroy()` is the only one that runs inside a server-side component. 😓

PR #12624, opened 2025-12-29
svelte: Implement `ProgressBar` component and corresponding context class

Turbo87:
This implements the global loading state indicator at the top of the page, similar to the one in the Ember.js app. While the Ember.js app uses ember-concurrency tasks, the Svelte implementation uses raw `setInterval()` and `setTimeout()`, but ultimately achieves the same functionality.

In contrast to the Ember.js implementation, this one also contains a couple of basic unit tests for the functionality.

### Related

### DIFF
--- a/svelte/src/lib/progress.test.ts
+++ b/svelte/src/lib/progress.test.ts
@@ -1,0 +1,103 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { ProgressState } from './progress.svelte';
+
+describe('ProgressState', () => {
+  beforeEach(() => vi.useFakeTimers());
+  afterEach(() => vi.useRealTimers());
+
+  it('starts animation when tracking a promise', () => {
+    let state = new ProgressState();
+    let promise = new Promise(() => {});
+
+    state.trackPromise(promise);
+    expect(state.style).toBe('width: 0%');
+
+    vi.advanceTimersByTime(200);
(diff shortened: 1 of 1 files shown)


svelte/src/lib/components/ProgressBar.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+  import { getProgressContext } from '$lib/progress.svelte';
+
+  let progress = getProgressContext();
+</script>
+
+<div class="progress-bar" style={progress.style}></div>
```
eth3lbert, line 7: I'm wondering if it would be better to decouple the progress context from this component. We could move the styling logic into the component and have it accept a value prop to set the width, along with other required or optional props. What do you think?

---

There is also a built-in tween[^1] motion that could simplify things. However, since it's a JS solution, it might increase the bundle size a bit compared to a CSS transition solution.

[^1]: https://svelte.dev/tutorial/svelte/tweens

Turbo87: hmm... I'm not necessarily opposed to it, but given the tight coupling with the `transition` CSS attribute, I'm not sure if it actually has much of an advantage in the end. it's definitely worth exploring, but I think I wouldn't want to block the Svelte migration on that 😅

eth3lbert: Makes sense! I understand your point and am happy to see this merged.

PR #12625, opened 2025-12-29
svelte: Add `CrateRow` and `CrateList` components

Turbo87:
Bildschirmfoto 2025-12-29 um 17 57 35

### Related

### DIFF
--- a/svelte/src/lib/components/CrateList.stories.svelte
+++ b/svelte/src/lib/components/CrateList.stories.svelte
@@ -1,0 +1,74 @@
+<script module lang="ts">
+  import type { components } from '@crates-io/api-client';
+
+  import { defineMeta } from '@storybook/addon-svelte-csf';
+
+  import CrateList from './CrateList.svelte';
+
+  const { Story } = defineMeta({
+    title: 'CrateList',
+    component: CrateList,
+    tags: ['autodocs'],
+  });
+
+  type Crate = components['schemas']['Crate'];
+
+  function createCrate(overrides: Partial<Crate> & { id: string; name: string }): Crate {
+    return {
+      badges: [],
+      created_at: '2014-11-05T00:00:00Z',
+      default_version: '1.0.0',
+      description: 'A sample crate description',
+      documentation: 'https://docs.rs/example',
+      downloads: 1000000,
+      exact_match: false,
+      homepage: 'https://example.com',
+      links: {
+        owner_team: `/api/v1/crates/${overrides.id}/owner_team`,
+        owner_user: `/api/v1/crates/${overrides.id}/owner_user`,
+        reverse_dependencies: `/api/v1/crates/${overrides.id}/reverse_dependencies`,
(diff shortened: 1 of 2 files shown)


svelte/src/lib/components/CrateRow.svelte
```diff
@@ -0,0 +1,250 @@
...
+        >
+          <CopyIcon aria-label="Copy Cargo.toml snippet to clipboard" />
+        </button>
+      {/if}
+    </div>
+    {#if crate.description}
+      <div class="description text--small" data-test-description>
+        {truncateText(crate.description)}
+      </div>
+    {/if}
```
eth3lbert, line 60: Is this conditional intentional, since it isn't in Ember's implementation?

svelte/src/lib/components/CrateRow.svelte
```diff
@@ -0,0 +1,250 @@
...
+      {#if crate.documentation}
+        <!-- eslint-disable-next-line svelte/no-navigation-without-resolve -->
+        <li><a href={crate.documentation}>Documentation</a></li>
+      {/if}
+      {#if crate.repository}
+        <!-- eslint-disable-next-line svelte/no-navigation-without-resolve -->
+        <li><a href={crate.repository}>Repository</a></li>
+      {/if}
+    </ul>
+  {/if}
```
eth3lbert, line 118: The same applies here.

svelte/src/lib/utils/truncate-text.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Truncates text to a maximum length, appending "..." if truncated.
+ */
+export function truncateText(value: string, maxLength = 200): string {
+  if (value.length > maxLength) {
+    return `${value.slice(0, maxLength).trim()} …`;
```
eth3lbert, line 6: Is the `trim()` here intentional? I noticed it isn't in the Ember version.

svelte/src/lib/components/CrateRow.svelte
Turbo87: yeah, I added it to make TypeScript happy. in the Ember app `truncateText()` deals with `null` by returning an empty string, but I figured why render the wrapper element if we don't show any content inside. I've verified manually that it visually looks the same either way.

svelte/src/lib/components/CrateRow.svelte
Turbo87: admittedly, this one was unintentional, but looks like it is visually the same, so we might as well skip these elements?

svelte/src/lib/utils/truncate-text.ts
Turbo87: yeah, slight deviation from the Ember.js impl after I noticed in the tests that the Ember.js behavior isn't particularly useful 😅

svelte/src/lib/components/CrateRow.svelte
eth3lbert: yeah, LGTM 👍

PR #12628, opened 2025-12-29
svelte: Add color scheme picker

Turbo87:
This PR implements the color scheme dropdown in the `Header` component. It is implemented using a `ColorSchemeState` context class which reads and writes the selected value from/to local storage and uses a mediaquery if the `system` value is used. This roughly matches the implementation of the Ember.js application.

This PR also introduces a "decorator" for the Storybook setup, which automatically sets up the context for all storybook pages.

### Related

### DIFF
--- a/svelte/src/lib/color-scheme.svelte.ts
+++ b/svelte/src/lib/color-scheme.svelte.ts
@@ -1,0 +1,45 @@
+import { createContext } from 'svelte';
+import { MediaQuery } from 'svelte/reactivity';
+
+import * as storage from '$lib/utils/local-storage';
+
+export type ColorScheme = 'light' | 'dark' | 'system';
+export type ResolvedScheme = 'light' | 'dark';
+
+const STORAGE_KEY = 'color-scheme';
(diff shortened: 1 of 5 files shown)


svelte/src/lib/color-scheme.svelte.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const STORAGE_KEY = 'color-scheme';
+const VALID_SCHEMES = new Set<ColorScheme>(['light', 'dark', 'system']);
+
+export class ColorSchemeState {
+  scheme = $state<ColorScheme>('system');
+
+  #mediaQuery = new MediaQuery('prefers-color-scheme: dark', false);
+
+  readonly resolvedScheme: ResolvedScheme = $derived(
```
eth3lbert, line 17: What about using `current` here? However, I understand that `resolvedScheme` matches Ember's implementation, so it’s fine to keep it as it is.

Turbo87: you mean renaming `resolvedScheme` to `current`? I'm open to it if you think that it's easier to understand, but for now it's probably best to keep the same naming as in the Ember impl.

PR #12630, opened 2025-12-30
Add disclaimer about not being able to republish package soon after deletion

alichtman:
I recently deleted a test crate I published. When I tried to re-publish it, I got this error:

```
$ cargo publish
    Updating crates.io index
   Packaging ... v0.1.0 (/Users/alichtman/Desktop/Development/...)
   Uploading ... v0.1.0 (/Users/alichtman/Desktop/Development/...)
error: failed to publish ... v0.1.0 to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 400 Bad Request): A crate with the name `...` was recently deleted. Reuse of this name will be available after 2025-12-31T09:35:20Z.
```

Having this warning message would have prevented me from making this mistake

### DIFF
--- a/app/templates/crate/delete.gjs
+++ b/app/templates/crate/delete.gjs
@@ -19,7 +19,6 @@ import LoadingSpinner from 'crates-io/components/loading-spinner';
       <Alert @variant='warning'>
         <strong>Important:</strong>
         This action will permanently delete the crate and its associated versions. Deleting a crate cannot be reversed!
-        Additionally, you will be blocked from republishing a crate with this name for up to 36 hours.
       </Alert>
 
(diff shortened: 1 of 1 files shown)


alichtman:
@Turbo87 would it be possible to waive the reuse restriction on the package name `bunnylol`? I'd publish a new version now if I could

Thanks for the quick review!

Turbo87:
the restriction exists partially also for technical reasons. we can only lift it for exceptional reasons.

alichtman:
Np, not critical. Out of interest: what are the technical reasons?

Turbo87:
Cache and mirror TTLs and stuff like that

PR #12633, opened 2025-12-30
svelte: Add `Tooltip` component

Turbo87:
This roughly matches the implementation of the `Tooltip` component in the Ember.js application, except for the usage of the `TooltipContext`. That part is needed primarily for the Storybook implementation, to not create multiple conflicting tooltip containers with the same element ID.

### Related


Turbo87:
> although we still need to fix some linting issues

oops, yeah. guess that's a rebase issue that I missed 😅 

> It would be great if we didn't have to create an extra element as an anchor, but I think this is good enough! thanks!

agreed! we might be able to do this in the future if we choose to change the API of how tooltips are attached to elements, but I wanted to keep the external API of the tooltips close to how they are used in the Ember.js app to make the other components that use tooltips easier to port.

PR #12636, opened 2025-12-31
svelte: Implement `/support` route

Turbo87:
Playwright test suite is passing:

Bildschirmfoto 2025-12-31 um 04 28 35

(expect for reporting crates from the crate page, because the crate page hasn't been implemented yet 😅)

### Related

### DIFF
--- a/svelte/src/lib/components/Footer.svelte
+++ b/svelte/src/lib/components/Footer.svelte
@@ -21,7 +21,7 @@
       <h1>Get Help</h1>
       <ul role="list">
         <li><a href="https://doc.rust-lang.org/cargo/">The Cargo Book</a></li>
-        <li><a href={resolve('/support')}>Support</a></li>
+        <li><a href={resolve('/support')} data-test-support-link>Support</a></li>
         <li><a href="https://status.crates.io/">System Status</a></li>
         <li><a href="https://github.com/rust-lang/crates.io/issues/new/choose">Report a bug</a></li>
       </ul>


svelte/src/lib/components/support/CrateReportForm.svelte
```diff
@@ -0,0 +1,308 @@
...
+      oninput={() => (crateInvalid = false)}
+    />
+    {#if crateInvalid}
+      <div class="form-group-error" data-test-id="crate-invalid">Please specify a crate.</div>
+    {/if}
+  </fieldset>
+
+  <fieldset class="form-group" data-test-id="fieldset-reasons">
+    <div class="form-group-name">Reasons</div>
+    <ul role="list" class="reasons-list scopes-list" class:invalid={reasonsInvalid}>
```
eth3lbert, line 134:
```suggestion
    <ul role="list" class={["reasons-list scopes-list", { invalid: reasonsInvalid }]}>
```

I'm suggesting this primarily because of the following note in the official documentation.
> Unless you’re using an older version of Svelte, consider avoiding class:, since the attribute is more powerful and composable. [^1]

[^1]: https://svelte.dev/docs/svelte/class#The-class:-directive

svelte/src/lib/components/support/CrateReportForm.svelte
```diff
@@ -0,0 +1,308 @@
...
+    </div>
+  {/if}
+
+  <fieldset class="form-group" data-test-id="fieldset-detail">
+    <label for="{id}-detail" class="form-group-name">Detail</label>
+    <textarea
+      id="{id}-detail"
+      bind:value={detail}
+      class="detail"
+      class:invalid={detailInvalid}
```
eth3lbert, line 180:
```suggestion
      class={["detail", { invalid: detailInvalid }]}
```

eth3lbert (approved):
Just some minor suggestions about the class attributes, but otherwise it all LGTM. Thanks!

svelte/src/lib/components/support/CrateReportForm.svelte
Turbo87: I noticed that the array form doesn't appear to be well supported by the Svelte plugin for the JetBrains IDEs. It ends up showing me the corresponding class declarations as unused, which is why I'm leaning towards the using the "old" API for now. but tbh I'm not seeing that much of a benefit from using the array-based API anyway. I guess it's more useful if you assemble the array in the script tag in a dynamic way, but that also makes the "is this CSS class actually used" analysis more complicated, so idk if that's really worth it.

also, Claude prefers it this way... 😆😅

eth3lbert: > also, Claude prefers it this way... 😆😅

I assume this is because the attribute-based approach is a new feature that needs a bit more time to mature.

---

Your point is fair, though. I'm not strongly opposed to the old way, and we can always retrofit it later if needed!

PR #12646, opened 2026-01-02
svelte: Implement `/categories` routes

Turbo87:
This PR implements the `/categories`, `/categories/[category_id]`, and `/category_slugs` routes, similar to the corresponding keyword routes.

One small difference is that, similar to the Ember.js implementation, the category detail page sets the value of the search form in the header through the `SearchFormContext` that we introduced in an earlier PR.

### Related

### DIFF
--- a/svelte/src/routes/categories/[category_id]/+page.svelte
+++ b/svelte/src/routes/categories/[category_id]/+page.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { onMount } from 'svelte';
   import { resolve } from '$app/paths';
 
   import CrateList from '$lib/components/CrateList.svelte';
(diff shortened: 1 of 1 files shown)


svelte/src/routes/categories/+page.svelte
```diff
@@ -1,2 +1,97 @@
...
+      <SortDropdown.Option query={{ sort: 'crates' }}># Crates</SortDropdown.Option>
+    </SortDropdown.Root>
+  </div>
+</div>
+
+<div class="list">
+  {#each data.categories.categories as category (category.id)}
+    <div class="row" data-test-category={category.slug}>
+      <div>
+        <a href={resolve('/categories/[category_id]', { category_id: category.slug })}>{category.category}</a>
```
eth3lbert, line 43: There is a class name `category-link` in the Ember version which is missing here. However, since it doesn't actually do anything, it's fine to skip it.

svelte/src/routes/categories/[category_id]/+page.svelte
```diff
@@ -1,6 +1,144 @@
...
+    <h2>Subcategories</h2>
+    <div class="subcategories">
+      {#each data.category.subcategories as subcategory (subcategory.id)}
+        <div class="subcategory">
+          <div>
+            <a href={resolve('/categories/[category_id]', { category_id: subcategory.slug })}>
+              {subcategory.category}
+            </a>
+            <span class="text--small">
+              {formatNumber(subcategory.crates_cnt)}
```
Comment (eth3lbert) [inline svelte/src/routes/categories/[category_id]/+page.svelte]: The `formatNumber` function isn't used in the categories page, but it is used here. It would be great to unify the coding style one way or the other..

Comment (Turbo87) [reply svelte/src/routes/categories/[category_id]/+page.svelte]: good point. fixed :)

Issue #4176, opened 2021-11-14
https://crates.io/authorize/github fails to report authorization errors


kornelski:
**Describe the bug**
Redirect from github login page back to `https://crates.io/authorize/github?code=[redacted]&state=[redacted]` fails to log me in, and it doesn't explain why. When it runs in a pop-up, the popup just disappears (#3810). If I trick it to run in the same window (by running `window.open = () => window`), I get an empty page like in the attached screenshot.

**To Reproduce**
1. Click on 'Log in with GitHub'
2. Nothingness

**Expected behavior**
I expect to be logged in, OR receive an explanation of what went wrong, e.g. authorization failure, wrong token, missing cookies, expected referrer, or whatever is preventing the site from working. I suspect the failure is caused by my browser, but I don't have enough information to fix it.

**Screenshots**

<img width="450" alt="Screen Shot 2021-11-14 at 21 12 19" src="https://user-images.githubusercontent.com/72159/141698809-5f945dff-2d05-4bb6-8996-c3095f4dd635.png">

**Desktop (please complete the following information):**
 - macOS 12
 - Firefox 96.0a1

Turbo87:
hmm... that's definitely unintentional. I'll see if I can reproduce that.

k8scat:
![image](https://user-images.githubusercontent.com/55872984/147105144-1cb7a0ba-816b-4f28-ace4-c478faec0e12.png)

Failed to log in: Service unavailable

k8scat:
![image](https://user-images.githubusercontent.com/55872984/147106728-fc87ae0e-581a-4f95-86c3-a7baa66fdc9c.png)

carols10cents:
@k8scat Our service providers are currently experiencing an outage. Please see https://status.crates.io/incidents/r6qckdr6wtq1 for updates.

BlackGlory:
Still can't log in.

onbjerg:
Experiencing this, too. Firefox 96.0, no extensions installed, on Manjaro. I've tried adding an exception for pop-ups manually as well for crates.io, but that didn't resolve the issue.

pranavarora1895:
Kindly let me know when the bug gets fixed as I need to log in to it and want to learn how to publish my crates. Thanks.

pranavarora1895:
When I was using the **Chrome Browser**, I was unable to log in. I was able to log in when I used **Brave Browser**. Hope it helps while debugging.

[...]

Issue #3810, opened 2021-07-30
Login without pop-ups


kornelski:
I can't log in to crates.io from Firefox 92. The login pop-up opens, and then disappears, and literally nothing changes. No errors logged in the console. I can click "Log in" any number of times, and it will just keep opening the popup window, redirecting, and closing it.

It's very likely the fault of my browser or its extensions. However, the ephemeral pop-up window combined with JS-heavy page makes debugging of the problem very difficult. Top-level page doesn't log anything about GitHub auth process or its result, and the pop-up window is impossible to catch before it closes.

**Describe the solution you'd like**
Could you redirect to GitHub in the normal, top-level browsing context, rather than a pop-up window?

I suspect a more standard flow may also help with compatibility when multiple browsing context and or window openers aren't involved.

**Describe alternatives you've considered**
I can log in Chrome, but I'd prefer not to use it.

**Additional context**
![Screen Recording 2021-07-30 at 22 33 48](https://user-images.githubusercontent.com/72159/127713751-ca99441a-b939-43fe-bcb8-76468d95b598.gif)

xkikeg:
+1 with Google Chrome Version 93.0.4577.82 (Official Build) (x86_64)

PR #12650, opened 2026-01-03
Fix OAuth popup sending Origin: null to GitHub

robcohen:
## Summary

The current OAuth popup implementation causes GitHub to reject login attempts with HTTP 422 in certain browsers. This is because `document.write()` on a blank popup window creates an [opaque origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque), causing the browser to send `Origin: null` in subsequent requests.

## The Problem

Current flow in `app/services/session.js`:

```javascript
let win = window.open('', '_blank', windowDimensions);
win.document.write('<html>...</html>');  // ← Creates opaque origin
win.document.close();

let { url } = await ajax('/api/private/session/begin');
win.location = url;  // Navigation now sends Origin: null
```

When the popup navigates to GitHub and the OAuth form is submitted, the browser sends `Origin: null`. GitHub's CSRF protection rejects this with:

> "What‽ Your browser did something unexpected. Please try again."

This has been reported multiple times:
- #4320
- #3810
- #4176

## Root Cause

Per the [HTML spec](https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#document-open-steps), calling `document.open()` (implicitly via `document.write()`) on an about:blank document can result in an opaque origin. Chromium treats `document.write()` on a popup as creating a sandboxed context.

The comment in the current code states:

> "we can't call `window.open()` with this URL directly, because it might trigger the popup window prevention mechanism"

This is a valid concern, but the workaround breaks CSRF protection—a worse tradeoff.

## The Fix

Open the popup to a same-origin loading page instead of a blank window. This preserves the origin while still avoiding popup blockers.

- **New file:** `public/github-auth-loading.html` - fetches OAuth URL and redirects
- **Modified:** `app/services/session.js` - opens popup to loading page instead of blank

## References

- [HTML Spec: Opaque Origins](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque)
- [HTML Spec: document.open() steps](https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#document-open-steps)
- [Chromium Bug #635400](https://issues.chromium.org/issues/41268543)

## Testing

1. Use Chromium/ungoogled-chromium with default settings
2. Navigate to crates.io
3. Click "Log in with GitHub"
4. Verify popup opens, redirects to GitHub, and login completes
5. Check DevTools Network tab: POST to `github.com/login/oauth/authorize` should have proper `Origin` header (not `null`)

### DIFF
--- a/e2e/acceptance/login.spec.ts
+++ b/e2e/acceptance/login.spec.ts
@@ -7,7 +7,7 @@ test.describe('Acceptance | Login', { tag: '@acceptance' }, () => {
     await page.addInitScript(() => {
       globalThis.open = (url, target, features) => {
         globalThis.openKwargs = { url, target, features };
-        return { document: { write() {}, close() {} }, close() {} } as ReturnType<(typeof globalThis)['open']>;
+        return { close() {} } as ReturnType<(typeof globalThis)['open']>;
       };
     });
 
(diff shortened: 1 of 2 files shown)


robcohen:
Great! Seems to work for me now.
Yes, I used Claude code to assist me.

PR #12657, opened 2026-01-04
svelte: Migrate `UserAvatar` and `OwnersList` components

Turbo87:
Bildschirmfoto 2026-01-04 um 13 05 17
Bildschirmfoto 2026-01-04 um 13 05 11

### Related

### DIFF
--- a/svelte/src/lib/components/OwnersList.svelte
+++ b/svelte/src/lib/components/OwnersList.svelte
@@ -1,0 +1,99 @@
+<script lang="ts">
+  import type { components } from '@crates-io/api-client';
+
+  import { resolve } from '$app/paths';
+
+  import UserAvatar from './UserAvatar.svelte';
+
+  type Owner = components['schemas']['Owner'];
+
+  interface Props {
+    owners: Owner[];
+  }
+
+  let { owners }: Props = $props();
+
+  let showDetailedList = $derived(owners.length <= 5);
+
+  function displayName(owner: Owner): string {
+    if (owner.kind === 'team') {
+      // For teams, compute display_name as org_name/name
+      // login format is "github:org_name:team_name"
+      let orgName = owner.login.split(':')[1];
+      return owner.name ? `${orgName}/${owner.name}` : owner.login;
+    }
+    return owner.name ?? owner.login;
+  }
+</script>
+
+<ul
+  role="list"
+  class="list"
+  class:detailed={showDetailedList}
+  data-test-owners={showDetailedList ? 'detailed' : 'basic'}
+>
+  {#each owners as owner (owner.id)}
+    {@const isTeam = owner.kind === 'team'}
+    {@const href = isTeam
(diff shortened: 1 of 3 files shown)


svelte/src/lib/components/UserAvatar.svelte
```diff
@@ -0,0 +1,36 @@
...
+
+  let { user, size = 'small', ...rest }: Props = $props();
+
+  let sizeValue = $derived.by(() => {
+    if (size === 'medium') return 85;
+    if (size === 'medium-small') return 32;
+    return 22;
+  });
+
+  let alt = $derived(user.name ? `${user.name} (${user.login})` : `(${user.login})`);
```
eth3lbert, line 21: Although the expression isn't identical to the Ember one, I think it should be fine!

svelte/src/lib/components/OwnersList.svelte
```diff
@@ -0,0 +1,99 @@
...
+  let { owners }: Props = $props();
+
+  let showDetailedList = $derived(owners.length <= 5);
+
+  function displayName(owner: Owner): string {
+    if (owner.kind === 'team') {
+      // For teams, compute display_name as org_name/name
+      // login format is "github:org_name:team_name"
+      let orgName = owner.login.split(':')[1];
+      return owner.name ? `${orgName}/${owner.name}` : owner.login;
```
eth3lbert, line 23: This expression looks different from Ember's. 🤔

Turbo87: this more or less comes from:
 https://github.com/rust-lang/crates.io/blob/d1ee657f60a8eae1092bf98b8caf7d21c14fa1da/app/models/team.js#L12-L20

but the Ember impl ignored that `name` could theoretically be `null`. I assume our backend prevents this situation from actually happening, but TS ensures we handle this case in a defensive manor. I guess we could also do `${split[1]}/${split[2]}` instead of `login` in that case. wdyt?

eth3lbert: Yeah, I mainly just wanted to point out that there could be a difference, but I'm fine either way since they all LGTM! Do you think we should add a spec test for the `OwnerList` component during the migration stage, similar to the one we have in Ember (Or I guess you treat Storybook as a kind of specification test!)?

Turbo87: > Do you think we should add a spec test for the `OwnerList` component during the migration stage, similar to the one we have in Ember

good point! I've migrated the tests to vitest now. I've skipped the Ember Data parts of the tests since they don't seem relevant to the thing that the tests are actually testing.

> Or I guess you treat Storybook as a kind of specification test

yeah, to some degree. AFAIK it is possible to integrate Storybook with Percy or use their native visual regression testing, I just haven't gotten around to it yet, but it feels like we should probably make use of that in the future.

eth3lbert (approved):
LCTM 👍

PR #12660, opened 2026-01-04
svelte: Add `crate-sidebar/Link` component

Turbo87:
Bildschirmfoto 2026-01-04 um 14 03 10

### Related

### DIFF
--- a/svelte/src/lib/components/crate-sidebar/Link.svelte
+++ b/svelte/src/lib/components/crate-sidebar/Link.svelte
@@ -1,0 +1,82 @@
+<script module lang="ts">
+  export function simplifyUrl(url: string): string {
+    if (url.startsWith('https://')) {
+      url = url.slice('https://'.length);
+    }
+    if (url.startsWith('www.')) {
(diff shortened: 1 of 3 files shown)


svelte/src/lib/components/crate-sidebar/Link.svelte
```diff
@@ -0,0 +1,80 @@
...
+    title: string;
+    url: string;
+  }
+
+  let { url, title }: Props = $props();
+
+  let text = $derived(simplifyUrl(url));
+</script>
+
+<div>
```
eth3lbert: Should we add spread attributes here? Looking at the Ember implementation, we'll probably want to include `data-test` attributes on the links to make them easier to target later on.

eth3lbert (approved):
Just one small nit question. Otherwise, LGTM!

svelte/src/lib/components/crate-sidebar/Link.svelte
Turbo87: good point! added :)

PR #12669, opened 2026-01-05
svelte: Migrate `/users/[user_id]` and `/teams/[team_id]` routes

Turbo87:
These are basically the same as for categories and keywords, except they also load and show basic user/team information. The `include_yanked` behavior is marked as `TODO` for now until we have authentication support.

Bildschirmfoto 2026-01-05 um 16 32 57

### Related

- https://github.com/rust-lang/crates.io/issues/12515

PR #12672, opened 2026-01-06
svelte: Migrate `/crates` route

Turbo87:
### Related

- https://github.com/rust-lang/crates.io/pull/12669

### DIFF
--- a/svelte/src/routes/crates/+page.ts
+++ b/svelte/src/routes/crates/+page.ts
@@ -1,0 +1,42 @@
+import type { paths } from '@crates-io/api-client';
+
+import { createClient } from '@crates-io/api-client';
+import { error } from '@sveltejs/kit';
+
+export async function load({ fetch, url }) {
+  let client = createClient({ fetch });
+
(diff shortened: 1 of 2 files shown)


svelte/src/routes/crates/+page.ts
```diff
@@ -0,0 +1,26 @@
+import { createClient } from '@crates-io/api-client';
+
+export async function load({ fetch, url }) {
+  let client = createClient({ fetch });
+
+  let pageStr = url.searchParams.get('page') ?? '1';
+  let page = parseInt(pageStr, 10);
+  let perPage = 50;
+  let sort = url.searchParams.get('sort') ?? 'recent-downloads';
+  let letter = url.searchParams.get('letter') ?? undefined;
```
eth3lbert, line 13: Just curious, do we really need the `?? undefined` here?

Turbo87: `get()` apparently returns `null` if the key is not found and the API client expects `string | undefined`, so unless we want to make TS unhappy, I guess we do need the `?? undefined`

PR #12676, opened 2026-01-07
svelte: Migrate error page

Turbo87:
This PR first migrates the `catch-all` route from the Ember.js app and then adjusts the data loading code of the `/users/[user_id]` route to handle errors in a way that replicates the behavior of the Ember.js app. The Playwright tests for this route that are testing the error handling behavior are now passing for both apps.

**Update:** I've extended the PR to improve the error handling for the other `load()` fns too. It's probably easiest to review this commit-by-commit :)


eth3lbert (approved):
Awesome! It’s great to have more error handling included.

PR #12686, opened 2026-01-08
svelte: Migrate `CrateSidebar` component

Turbo87:
This PR migrates the majority of the `CrateSidebar` component to the Svelte app:

Bildschirmfoto 2026-01-08 um 16 46 46

Notable exception that will be implemented in follow-up PRs and are replaced by `// TODO` comments for now:

- Package URLs
- docs.rs links
- Categories
- Rust Playground buttons

### DIFF
--- a/svelte/src/lib/components/crate-sidebar/CrateSidebar.stories.svelte
+++ b/svelte/src/lib/components/crate-sidebar/CrateSidebar.stories.svelte
@@ -1,0 +1,132 @@
+<script module lang="ts">
+  import type { components } from '@crates-io/api-client';
+
+  import { defineMeta } from '@storybook/addon-svelte-csf';
+
+  import CrateSidebar from './CrateSidebar.svelte';
+
+  const { Story } = defineMeta({
+    title: 'crate-sidebar/CrateSidebar',
+    component: CrateSidebar,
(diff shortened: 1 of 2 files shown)


svelte/src/lib/components/crate-sidebar/InstallInstructions.svelte
```diff
@@ -0,0 +1,134 @@
...
+    return `${crate} = "${exact}${v}"`;
+  });
+</script>
+
+<div class="install-instructions">
+  {#if binNames && binNames.length > 0}
+    <CopyButton copyText={cargoInstallCommand} title="Copy command to clipboard" class="copy-button">
+      <span class="selectable">{cargoInstallCommand}</span>
+      <CopyIcon aria-hidden="true" class="copy-icon" />
+    </CopyButton>
```
eth3lbert: I noticed that the Ember version includes logic for clipboard detection and fallbacks. Since it isn't implemented here yet, I wanted to check if that was intentional.

Turbo87: good catch! I've restored the original behavior. these days most browsers support it anyway, but let's keep the diff small :)

PR #12693, opened 2026-01-09
Hide withdrawn advisories

djc:
Happy to take any feedback on the design/style/layout.

Screenshot 2026-01-09 at 12 32 07

### DIFF
--- a/app/routes/crate/security.js
+++ b/app/routes/crate/security.js
@@ -12,7 +12,9 @@ async function fetchAdvisories(crateId) {
     let advisories = await response.json();
     return advisories
       .filter(
-        advisory => !advisory.affected?.some(affected => affected.database_specific?.informational === 'unmaintained'),
+        advisory =>
+          !advisory.withdrawn &&
+          !advisory.affected?.some(affected => affected.database_specific?.informational === 'unmaintained'),
       )
       .map(advisory => ({
         ...advisory,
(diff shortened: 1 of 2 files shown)


Turbo87:
if they're withdrawn, should we display them at all?

djc:
> if they're withdrawn, should we display them at all?

Fair question. I was worried that it might be weird to have advisories appear and then disappear again on crates.io (especially if there is something high-profile going on) -- it feels like something that should be append-only? But I can see how it could make sense to just hide them.

Turbo87:
or maybe something like a "5 withdrawn advisories" annotation somewhere that links to the full list at rustsec? not really sure what makes the most sense here 😅

djc:
That works for me too. Do you prefer it? Should we ask other people?

Turbo87:
> Should we ask other people?

probably a good idea. maybe a zulip topic? or bluesky/mastodon?

djc:
Posted in https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/How.20to.20display.20withdrawn.20advisories/with/567500085.

djc:
Changed the PR to just hide withdrawn advisories.

LawnGnome (approved):
I think this is the right decision.

Turbo87 (approved):

PR #12695, opened 2026-01-09
svelte: Migrate `DownloadChart` component

Turbo87:
Bildschirmfoto 2026-01-09 um 14 32 27

The API of the component unfortunately needed to be slightly different from the Ember.js variant because the latter is using Ember Data model instances, while in the Svelte app we are working with raw objects at the moment, where relationships are not automatically resolved.

Aside from that, the component behaves essentially the same as in the Ember.js app. I've tried to keep the diff as small as possible compare to the Ember.js implementation, so there are probably areas of improvements here, but we can tackle those after the migration.

### Related


Turbo87:
FYI I've rebased this PR and now also included the commit that integrates the new graph component into the crate details page.

PR #12700, opened 2026-01-09
Add version ranges to advisory listings

djc:
Screenshot 2026-01-09 at 17 12 19

### DIFF
--- a/app/routes/crate/security.js
+++ b/app/routes/crate/security.js
@@ -1,6 +1,8 @@
 import Route from '@ember/routing/route';
 import { service } from '@ember/service';
 
+import { versionRanges } from 'crates-io/utils/version-ranges';
+
 async function fetchAdvisories(crateId) {
   let url = `https://rustsec.org/packages/${crateId}.json`;
   let response = await fetch(url);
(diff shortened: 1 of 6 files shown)


app/routes/crate/security.js
```diff
@@ -1,16 +1,69 @@
...
+//
+// ```
+// "events": [
+//   { introduced: "0.0.0-0" },
+//   { fixed: "0.7.46" },
+//   { introduced: "0.8.0" },
+//   { fixed: "0.8.13" }
+// ]
+// ```
+function versionRanges(advisory) {
```
Turbo87: what do you think about moving this into a file in `app/utils/` and then writing a couple of dedicated tests for the fn?

djc: Happy to write some unit tests. Moving it out of the single caller's module isn't really my style, but then I don't know Ember's/JS conventions so happy to defer to your judgement.

Turbo87: I think I'd move it out since we'd have to export the function and with the tests it won't be the only user either 😄

Turbo87 (approved):

PR #12726, opened 2026-01-14
svelte: Migrate README loading

Turbo87:
This PR migrates the async crate README loading from the Ember.js app to the Svelte app. Both relevant `+page.ts` files return a `readmePromise` to async load the README file while the page skeleton with the basic information has already rendered. This should roughly match the behavior of the Ember.js app.

### Related

### DIFF
--- a/svelte/src/lib/components/ReadmePlaceholder.svelte
+++ b/svelte/src/lib/components/ReadmePlaceholder.svelte
@@ -1,76 +1,20 @@
 <script lang="ts">
   import Placeholder from '$lib/components/Placeholder.svelte';
+
+  let radius = 'var(--space-3xs)';
+  let text = { width: '100%', height: '16px', opacity: 0.3, style: 'margin-top: var(--space-xs)' };
 </script>
 
 <div>
-  <Placeholder
-    width="30%"
-    height="25px"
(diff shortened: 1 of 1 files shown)


svelte/src/lib/components/CrateVersionPage.svelte
```diff
@@ -38,11 +49,95 @@
...
-    README content goes here.
+    {#await activeReadmePromise}
+      <div class="readme-spinner">
+        <Placeholder
+          width="30%"
+          height="25px"
+          radius="var(--space-3xs)"
+          opacity={0.6}
+          style="margin: var(--space-s) 0 var(--space-m)"
+        />
```
Turbo87: this part is not particularly pretty... not sure what to do about it 🤔

eth3lbert: Since only two of the props are different, I’m wondering if we could either create an array of props and loop through them, or create a readmePlaceholder snippet that accepts props to override the defaults.

Turbo87: I've extracted a `ReadmePlaceholder` component and used loops and spread props to reduce the duplication now :)

Issue #12668, opened 2026-01-05
Metadata (update date and bytes size) do not have labels


VojtaStanek:
### Current Behavior

On the frontend some of the metadata do not have any labels. It may not be understandable what: 
- the date means (first crate publish or latest version?),
- the number of bytes mean

<img width="750" height="424" alt="Image" src="https://github.com/user-attachments/assets/8a8fcc70-abbe-4850-b468-ca06dc27f482" />

On mobile there is for MSRV since hover does not work well on mobile.

### Expected Behavior

It should be understandable what each item means - it should have labels such as "Last update" or "Download size" (or whatever the bytes mean).

### Steps To Reproduce

1. Go to https://crates.io/crates/syn
2. Look to the sidebar

### Environment

- Browser: MS Edge
- OS: Windows

kishore08-07:
@rustbot claim
I would like to work on this issue.

PR #12746, opened 2026-01-16
Add tooltips to crate sidebar release date and size

psg-19:
This PR addresses #12668  by adding descriptive labels ("Last Updated" and "Size") to the metadata fields in the crate sidebar. This improves clarity for users by explicitly identifying what the date and byte values represent.

- app/components/crate-sidebar.gjs -> Added label elements for "Last Updated" and "Size" within the metadata section.
- app/components/crate-sidebar.css -> Added a ._label class to style the new labels with consistent font weight and spacing.
- tests/acceptance/crate-test.js -> Updated acceptance tests to verify that the new labels are correctly rendered on the crate page.

### DIFF
--- a/svelte/src/lib/components/crate-sidebar/CrateSidebar.svelte
+++ b/svelte/src/lib/components/crate-sidebar/CrateSidebar.svelte
@@ -110,7 +110,10 @@
       <CalendarIcon />
       <span>
         {formatDistanceToNow(version.created_at, { addSuffix: true })}
-        <Tooltip text={format(version.created_at, 'PPP')} />
+        <Tooltip>
+          Release date:
+          {format(version.created_at, 'PPP')}
+        </Tooltip>
       </span>
     </time>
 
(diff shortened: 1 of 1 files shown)


Turbo87:
why did you use visible labels instead of tooltips like some of the other metadata fields are using?

psg-19:
Thanks for the review.
The issue suggested `"hover does not work well on mobile"`, so I thought to use visible labels.
I will make the changes using tooltips.

Turbo87:
> The issue suggested `"hover does not work well on mobile"`

yeah, but our tooltips are built in a way where clicking on the thing also displays the tooltips, so I'm not sure that comment actually applied. admittedly tooltips are still not a perfect pattern, but changing that is a bigger discussion, so in this case consistency seems more important for now.

Turbo87:
I've rebased the branch and applied the changes to the new Svelte frontend instead. I've also slightly tweaked the labels to "Release date" and "Compressed package size".

PR #12767, opened 2026-01-20
svelte: Load crate owners from API

Turbo87:
This resolves the two "load owners" `TODO` comments. The owners list is loaded in `routes/crates/[crate_id]/+layout.ts?` so that it can be shared with all subpages. This will allow the crate header to only show the "Settings" tab if the authenticated user (still TODO) is part of the owners list, similar to the Ember.js app.

### Related

### DIFF
--- a/svelte/src/lib/components/CrateVersionPage.svelte
+++ b/svelte/src/lib/components/CrateVersionPage.svelte
@@ -23,13 +23,26 @@
     crate: Crate;
     version: Version;
     keywords?: Keyword[];
-    owners: Owner[];
+    ownersPromise: Promise<Owner[]>;
     requestedVersion?: string;
     readmePromise: Promise<string | null>;
     downloadsPromise: Promise<DownloadChartData>;
   }
 
-  let { crate, version, keywords = [], owners, requestedVersion, readmePromise, downloadsPromise }: Props = $props();
+  let {
+    crate,
+    version,
+    keywords = [],
+    ownersPromise,
+    requestedVersion,
+    readmePromise,
+    downloadsPromise,
+  }: Props = $props();
+  let owners: Owner[] = $state([]);
+
+  $effect(() => {
+    ownersPromise.then(data => (owners = data)).catch(() => (owners = []));
+  });
 
   let numberFormat = new Intl.NumberFormat();
   let stackedGraph = $state(true);
--- a/svelte/src/routes/crates/[crate_id]/+layout.ts
+++ b/svelte/src/routes/crates/[crate_id]/+layout.ts
@@ -6,11 +6,12 @@ export async function load({ fetch, params }) {
 
   let crateName = params.crate_id;
 
-  let [crateData, owners] = await Promise.all([loadCrate(client, crateName), loadOwners(client, crateName)]);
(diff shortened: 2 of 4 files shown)


svelte/src/lib/components/CrateVersionPage.svelte
```diff
@@ -23,15 +23,21 @@
     crate: Crate;
     version: Version;
     keywords?: Keyword[];
+    owners?: Owner[];
```
JayanAXHF: I'm quite curious: why does owners need to be optional? Is it for the loading stage? I don't think crates can be without owners, right?

Turbo87: good point. there is no reason to make it optional since both usages of the component supply the prop. I've changed it to non-optional. 👍

JayanAXHF: Alright, wonderful. Would the same apply to the requestedVersion prop as well?

Turbo87: > Would the same apply to the requestedVersion prop as well?

nope, because that one isn't used by https://github.com/rust-lang/crates.io/blob/main/svelte/src/routes/crates/%5Bcrate_id%5D/%2Bpage.svelte#L11-L17, only by the route that includes the version number.

svelte/src/routes/crates/[crate_id]/+layout.ts
```diff
@@ -6,11 +6,30 @@ export async function load({ fetch, params }) {
 
   let crateName = params.crate_id;
 
-  let { crate, categories, keywords, defaultVersion } = await loadCrate(client, crateName);
+  let [crateData, owners] = await Promise.all([loadCrate(client, crateName), loadOwners(client, crateName)]);
```
Comment (eth3lbert) [inline svelte/src/routes/crates/[crate_id]/+layout.ts]: Not a blocker: I guess this makes the behavior slightly different from the Ember version, as we now need to wait for owners to load before we can render the page. However, I think it's fine to merge this if you want to migrate first and retrofit it later.

svelte/src/lib/components/CrateVersionPage.svelte
eth3lbert: This change would force us to load it before rendering the page. However, we could potentially treat it as a loading state when the owners list is empty.

svelte/src/routes/crates/[crate_id]/+layout.ts
Comment (Turbo87) [reply svelte/src/routes/crates/[crate_id]/+layout.ts]: yeah, that was a deliberate decision. we don't have any kind of loading state for the owners list, so it felt a bit strange to load that data async. do you think we should return `ownersPromise` instead?

Comment (eth3lbert) [reply svelte/src/routes/crates/[crate_id]/+layout.ts]: Yeah, I think using an `ownerPromise` or an empty list as a loading state would be better(if we want to avoid blocking).

Comment (Turbo87) [reply svelte/src/routes/crates/[crate_id]/+layout.ts]: I've added a second commit that implements this. is this what you had in mind?

Comment (eth3lbert) [reply svelte/src/routes/crates/[crate_id]/+layout.ts]: exactly!

PR #12772, opened 2026-01-21
svelte: Add `dev:msw` script

Turbo87:
This script launches the Vite dev server with the `VITE_MSW_ENABLED` environment variable set, which will cause our app to initialize a basic MockServiceWorker setup including authentication. This should make it easier in the future to implement features for authenticated users, since it can be tested and developed locally.

### Related


eth3lbert (approved):
As for the typing workaround, we could just rename `@crates-io/msw/index.js` to `index.ts`. That ought to work without needing the `no-explicit-any` fix, and it should give the `db` proper types too.

Turbo87:
last time I tried that it didn't work for the QUnit/Ember test suite. once we get rid of that we can try that though :)

PR #12777, opened 2026-01-22
Add aliases and CVSS vectors to Security tab

djc:
Screenshot 2026-01-22 at 15 38 10

### DIFF
--- a/app/templates/crate/security.gjs
+++ b/app/templates/crate/security.gjs
@@ -2,6 +2,24 @@ import { htmlSafe } from '@ember/template';
 
 import CrateHeader from 'crates-io/components/crate-header';
 
+function aliasUrl(alias) {
+  if (alias.startsWith('CVE-')) {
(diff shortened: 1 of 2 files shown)


app/templates/crate/security.gjs
```diff
@@ -2,6 +2,24 @@ import { htmlSafe } from '@ember/template';
 
 import CrateHeader from 'crates-io/components/crate-header';
 
+function aliasUrl(alias) {
```
djc, line 5: Much nicer, thanks! Wasn't sure if this kind of logic was allowed to be in templates.

Turbo87: In old school ember it would've been a "template helper" but these days regular fns can be imported into the templates. if the logic is only used in one template then IMHO it's fine to put it into the same file. if you want to write tests for it then it could make sense to put it in a dedicated file, but it's not a hard rule :)

PR #12780, opened 2026-01-22
svelte: Implement Rust playground data loading and "Try" button

Turbo87:
This PR implements the "Try on Rust Playground" button on the crate details page sidebar, similar to how it is implemented in the Ember.js app.

In the root `load()` fn in the `+layout.ts` file we perform a single request to https://play.rust-lang.org/meta/crates, which is saved as a promise and then inherited by all of the child pages. The crate details page then passes the promise to the `CrateSidebar` component, which displays the button if the promise has resolved and the crate is part of the returned list of supported crates.

Don't be scared by the size of the diff, most of the changes are related to the Storybook and test suite. The actual implementation should be relatively straight forward.

### Related

### DIFF
--- a/svelte/src/lib/components/crate-sidebar/CrateSidebar.svelte
+++ b/svelte/src/lib/components/crate-sidebar/CrateSidebar.svelte
@@ -1,9 +1,11 @@
 <script lang="ts">
   import type { components } from '@crates-io/api-client';
+  import type { PlaygroundCrate } from '$lib/utils/playground';
 
   import { resolve } from '$app/paths';
   import { format, formatDistanceToNow, formatISO } from 'date-fns';
   import prettyBytes from 'pretty-bytes';
+  import { MediaQuery } from 'svelte/reactivity';
 
   import CalendarIcon from '$lib/assets/calendar.svg?component';
   import CircleQuestionIcon from '$lib/assets/circle-question.svg?component';
@@ -17,11 +19,15 @@
   import OwnersList from '$lib/components/OwnersList.svelte';
   import Tooltip from '$lib/components/Tooltip.svelte';
   import { formatShortNum } from '$lib/utils/format-short-num';
+  import { buildPlaygroundLink } from '$lib/utils/playground';
   import { getPurl } from '$lib/utils/purl';
   import Edition from './Edition.svelte';
   import InstallInstructions from './InstallInstructions.svelte';
   import Link, { simplifyUrl } from './Link.svelte';
   import Msrv from './Msrv.svelte';
+
+  const PLAYGROUND_TOOLTIP =
+    'The top 100 crates are available on the Rust Playground for you to try out directly in your browser.';
 
   type Crate = components['schemas']['Crate'];
   type Version = components['schemas']['Version'];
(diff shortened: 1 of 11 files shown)


svelte/src/lib/utils/playground.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Builds a Rust Playground URL with pre-populated code for the given crate.
+ *
+ * @param id - The module identifier used in Rust import statements (e.g., "aho_corasick").
+ * @returns A URL string to the Rust Playground with example code.
+ */
+export function buildPlaygroundLink(id: string): string {
+  let code = `use ${id};\n\nfn main() {\n    // try using the \`${id}\` crate here\n}`;
+  return `https://play.rust-lang.org/?edition=2021&code=${encodeURIComponent(code)}`;
```
LawnGnome, line 50: Admittedly irrelevant for this PR, but should we change the default edition to 2024?

svelte/src/lib/components/crate-sidebar/CrateSidebar.svelte
```diff
@@ -17,12 +19,16 @@
...
   import { formatShortNum } from '$lib/utils/format-short-num';
+  import { buildPlaygroundLink } from '$lib/utils/playground';
   import { getPurl } from '$lib/utils/purl';
   import Edition from './Edition.svelte';
   import InstallInstructions from './InstallInstructions.svelte';
   import Link, { simplifyUrl } from './Link.svelte';
   import Msrv from './Msrv.svelte';
 
+  const PLAYGROUND_TOOLTIP =
+    'The top 100 crates are available on the Rust Playground for you to try out directly in your browser.';
```
LawnGnome, line 30: Similarly, we seem to have considerably more than 100 crates available at this point.

svelte/src/lib/components/crate-sidebar/CrateSidebar.svelte
```diff
@@ -181,19 +190,30 @@
...
-      href={playgroundLink}
-      target="_blank"
-      rel="noopener noreferrer"
-      class="playground-button button button--small"
-      data-test-playground-button
-    >
-      Try on Rust Playground
-    </a>
-    -->
+    {#await playgroundCratesPromise then playgroundCrates}
```
LawnGnome, line 193: Do we want a loading state here?

svelte/src/lib/utils/playground.ts
Turbo87: yep, we should. just need to figure out whether we want to do it before or after the migration.

I guess we could update it on both apps in parallel.

svelte/src/lib/components/crate-sidebar/CrateSidebar.svelte
Turbo87: yeah, I noticed that too. I assume it is the top 100 plus their dependencies.

svelte/src/lib/components/crate-sidebar/CrateSidebar.svelte
Turbo87: probably, though not during the migration while we try to mimic the behavior of the Ember app. my goal is to make the playwright test suite pass when run against the Svelte app, but every intentional difference to the Ember app will make this slightly harder 😅

PR #12792, opened 2026-01-25
One migration towards detangling GH

carols10cents:
Connects to https://github.com/rust-lang/crates.io/issues/10611. 

This PR contains ~~a framework for running arbitrary checked-in+reviewed SQL files to do data migrations in a background job,~~ one schema migration to start the process of detangling GitHub oauth info from the users table, and one data migration ~~using the new framework~~ duplicating data from `users` into the `oauth_github` table to be run ~~in a background job~~ manually in `psql` sometime after this PR is deployed.

## Data migration framework

Based on feedback, the background job framework isn't really necessary, so I've changed this PR to just include some SQL to review then run manually.

~~The framework is a background job runner that expects the job to specify what SQL to run from a file in the `sql` directory, to be enqueued manually by an admin.~~

~~The main advantage I see in doing the migrations in a background job on-demand, as opposed to just smooshing data migration in with the schema migrations, is that the data migration can run some other time than whenever we decide to deploy. Deploys are generally non-events that can be done at whatever time of day and I think it's good to keep it that way. I'm happy to stay on top of the data migration deploy status and enqueue a background job on a weekend or other time of low usage.~~

## `oauth_github` table

So on to the code/schema changes: After thinking about [this comment](https://github.com/rust-lang/crates.io/pull/10610#discussion_r2527421846) on my first PR attempt, I decided to indeed go with a provider-specific table to allow for different information/type specification. I think the only time we'd need to query across all providers is on the page where a user can manage their linked accounts, and i think that's acceptable. When logging in, it'll always be from a known provider.

See the commit message for more info on why the `oauth_github` table is the way it is. For now, the code is only writing to that table when someone logs in to keep the `users` table and `oauth_github` table in sync-- the `users` table is still the source of truth and nothing is reading from `oauth_github`.

If needed, the migration can be safely rolled back, everything dropped from the `oauth_github` table, and this code reverted, again because the data isn't used anywhere yet.

## Duplication of `users` to `oauth_github`

Eventually, though, we _will_ want to read from `oauth_github`, which means all the github-related info in the `users` table needs to be duplicated to the `oauth_github` table regardless of whether the user has logged in.

## Testing

Locally, I ran crates.io on `main` and logged in, then switched to this branch, and I was able to log in again and have my info duplicated into `oauth_github` correctly.

I also downloaded a database dump, loaded up the `users` table's data (~57k records), and ran the ~~background job with the~~ SQL in `oauth_github.sql` with `psql`. It copied all the records over basically instantly.

### DIFF
--- a/migrations/2026-01-20-162913_oauth_github_table/up.sql
+++ b/migrations/2026-01-20-162913_oauth_github_table/up.sql
@@ -15,9 +15,15 @@ CREATE TABLE IF NOT EXISTS oauth_github (
 );
 -- safety-assured:end
 
+-- safety-assured:start
+-- This table doesn't exist yet, so creating this index concurrently isn't necessary.
+CREATE INDEX IF NOT EXISTS index_oauth_github_user_id ON oauth_github (user_id);
+-- safety-assured:end
+
 comment on table oauth_github is 'GitHub-specific account information associated with a crates.io account';
 comment on column oauth_github.account_id is 'GitHub ID returned from the oAuth response';
 comment on column oauth_github.user_id is 'Crates.io user ID foreign key';
 comment on column oauth_github.encrypted_token is 'Encrypted GitHub access token';
 comment on column oauth_github.login is 'GitHub username';
 comment on column oauth_github.avatar is 'GitHub avatar URL';
+


src/worker/jobs/run_sql.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl RunSql {
+    pub fn new(file_name: impl AsRef<Path>) -> Self {
+        Self {
+            file_name: file_name.as_ref().into(),
+        }
+    }
+}
+
+impl BackgroundJob for RunSql {
```
Turbo87: I'm wondering if we really need/want the complexity of this being a background job. Since we're already starting a Heroku dyno to enqueue the job, we could also run the SQL script directly from that dyno.

I think the background job system makes sense when the migration needs to access the git index, since that needs exclusive access. But the SQL migration only needs access to the database, so I guess this requirement doesn't apply here?

migrations/2026-01-20-162913_oauth_github_table/up.sql
```diff
@@ -0,0 +1,31 @@
...
+  -- API documentation says its IDs are int64, so let's future-proof the table while we migrate.
+  account_id BIGINT NOT NULL,
+  -- Corresponds to users.gh_encrypted_token
+  encrypted_token bytea NOT NULL,
+  -- Corresponds to users.gh_login
+  login VARCHAR NOT NULL,
+  -- Corresponds to users.gh_avatar
+  avatar VARCHAR,
+
+  PRIMARY KEY (user_id, account_id)
```
Turbo87: why are we using a composite primary key and not just `account_id`?

migrations/2026-01-20-162913_oauth_github_table/up.sql
```diff
@@ -0,0 +1,31 @@
...
+comment on column oauth_github.encrypted_token is 'Corresponds to users.gh_encrypted_token';
+comment on column oauth_github.login is 'Corresponds to users.gh_login';
+comment on column oauth_github.avatar is 'Corresponds to users.gh_avatar';
+
+-- safety-assured:start
+-- This table is being created in this migration and nothing is reading it, so there's no concern
+-- about locking. If we wanted to use CONCURRENTLY, we'd have to not be in a transaction.
+CREATE UNIQUE INDEX oauth_github_account_id
+  ON oauth_github (account_id)
+  WHERE account_id > 0;
```
Turbo87: if we are extracting the GitHub data into a dedicated table, do we still want the same `gh_id: 0` handling? I wonder if it would be better to just not have a row in this table instead in that case.

migrations/2026-01-20-162913_oauth_github_table/up.sql
```diff
@@ -0,0 +1,31 @@
+-- safety-assured:start
+CREATE TABLE IF NOT EXISTS oauth_github (
+  -- Safe (and required) to use INTEGER rather than BIGINT because this is a foreign key to
+  -- users.id, which is INTEGER
+  user_id INTEGER NOT NULL REFERENCES users (id) ON DELETE CASCADE,
```
Turbo87, line 8: We currently have a (mostly) one-to-one relationship between crates.io users and github users. The exception is the `gh_id: 0` case, where the same github user can be used by multiple crates.io users.

An alternative way of modelling this relationship would be to keep the `gh_id` column in the `users` table and have it refer to the `oauth_github` table for the details like token, login and avatar. That way it would be easier to model the `gh_id: 0` case.

any thoughts on this?

migrations/data_oauth_github.sql
```diff
@@ -0,0 +1,14 @@
...
+INSERT INTO oauth_github
+(user_id, account_id, encrypted_token, login, avatar)
+SELECT
+  id as user_id,
+  gh_id as account_id,
+  gh_encrypted_token as encrypted_token,
+  gh_login as login,
+  gh_avatar as avatar
+FROM users
+ON CONFLICT DO NOTHING;
```
Turbo87, line 15: I'm a little hesitant to adding another top-level folder for something that I assume we'll rarely use. Can we make this a subfolder of `migrations`? Alternatively, maybe make it a subfolder of the background job / admin command module?

migrations/2026-01-20-162913_oauth_github_table/up.sql
carols10cents: Because `account_id` isn't unique-- it's -1 for GitHub accounts that have been deleted.

I suppose we could take this opportunity to change that convention -- that is, rather than setting `gh_id`/`account_id` to -1, the associated `users` record just wouldn't have any `oauth_github` records if there's no more github. Wdyt?

carols10cents:
> Currently we have three ways to run SQL on our database: 1) manually connecting to the database and running the script, 2) schema migrations via diesel, 3) custom admin commands and background jobs. Ultimately, this PR implements a fourth way and I wonder if it's worth the extra complexity.

Isn't this PR "3) custom admin commands and background jobs" though?

I think another advantage is having something standardized in the logs around who ran what SQL when, which you don't get with 1).

> I'm having a bit of a hard time coming up with significant advantages though. The only advantage I see would be that it could potentially prevent typos maybe? But since the SQL files here are not being tested in an automated way, they could ultimately still contain errors too, if they are not caught during review. If that error is discovered when attempting to run the data migration then the way to fix it would be to: open a PR to fix the SQL file, get it reviewed and merged, get it deployed, run the migration again. This seems like a lot of friction to me (admittedly friction can have positive aspects in such cases too 😅).

Yeah, I'm trying to introduce more structure and procedure here-- I feel like logging into the database and running SQL directly should become less frequent over time, not more. If we continue with adding the SQL in the migration file but commented out (which loses helpful things like syntax highlighting), then when we go to run it in the database directly and discover there's a problem, I'm assuming what happens is attempting to fix the problem live, without review, which makes me more nervous 😅 I think it's better to stop and have the friction of a review in that case.

I'm open to changing this if I'm in the minority here though!

migrations/2026-01-20-162913_oauth_github_table/up.sql
carols10cents: Ahaaa yep I should have read all the comments before responding to the previous one ;)

migrations/data_oauth_github.sql
carols10cents: Sure!

Turbo87:
> I think another advantage is having something standardized in the logs around who ran what SQL when, which you don't get with 1).

FWIW if everyone is using their user accounts to access the database instead of the application account then we get the audit log in the database logs too :)

> which loses helpful things like syntax highlighting

yeah, that's fair. I guess a `data.sql` in the same folder would be better 😄

eth3lbert:
Regarding the SQL file, I’m not sure if adding a background job is worth the extra complexity, but including the SQL file itself could be beneficial. Beyond just syntax highlighting, it provides testing capabilities and allows operators to execute the file directly without the risk of copy-paste errors, so I’m not against that part.

I do have one question regarding the background job: how can we ensure the required data migration is finished before we enable the related features? It seems we would either have to deploy the migration first, wait for it to complete, and then deploy the feature—which is similar to our current process—or introduce a mechanism to ensure the feature is only provided after the data migration is done. However, that second option might add another level of complexity. Let me know if I’m missing anything here.

carols10cents:
@Turbo87 @eth3lbert Ok, I think this is ready for re-review-- I've taken out the background job stuff and just put the data migration SQL in `migrations/data-oauth-github.sql` instead. What do you think?

> I do have one question regarding the background job: how can we ensure the required data migration is finished before we enable the related features? It seems we would either have to deploy the migration first, wait for it to complete, and then deploy the feature—which is similar to our current process—or introduce a mechanism to ensure the feature is only provided after the data migration is done. However, that second option might add another level of complexity. Let me know if I’m missing anything here.

I think it depends a lot on what the feature is and how it's implemented exactly. I'm attempting to structure this PR and all the auth-related refactoring in such a way that it doesn't matter exactly when schema changes get made or data migrations get run. When this PR gets deployed, users who log in will immediately have their `oauth_github` info duplicated into the new table. The way the SQL is written with `ON CONFLICT DO NOTHING`, any users who haven't logged in since the deploy will get their github info duplicated and the SQL won't error for records that have already been duplicated during login.

This "feature" doesn't really have a way to turn "on" yet-- I haven't implemented any code that tries to read the `oauth_github` table.

crates/crates_io_database/src/schema.rs
```diff
@@ -747,6 +772,24 @@ diesel::table! {
...
+        ///
+        /// Its SQL type is `Integer`.
+        crate_id -> Integer,
+        /// The `downloads` column of the `recent_crate_downloads` table.
+        ///
+        /// Its SQL type is `BigInt`.
+        downloads -> BigInt,
+    }
+}
+
```
Turbo87: hmm... I'm surprised that the diesel CLI moved this table around. It should be ordering them alphabetically and `recent_crate_downloads` seems like it should come after `readme_renderings` 😅 

can you check if you're on the latest version of their CLI?

crates/crates_io_database/src/schema.rs
```diff
@@ -664,6 +664,31 @@ diesel::table! {
...
+        /// Corresponds to users.gh_encrypted_token
+        encrypted_token -> Bytea,
+        /// Corresponds to users.gh_login
+        login -> Varchar,
+        /// The `user_id` column of the `oauth_github` table.
+        ///
+        /// Its SQL type is `Int4`.
+        ///
+        /// (Automatically generated by Diesel.)
+        user_id -> Int4,
```
Turbo87: looks like the table itself and the `user_id` column are missing comments. the other columns are currently using "Corresponds to users.foo" comments, but once we remove those columns of the `users` table these comments won't be helpful anymore. maybe we should use the final descriptions directly?

crates/crates_io_database/src/models/user.rs
```diff
@@ -124,3 +124,72 @@ impl NewUser<'_> {
...
     }
 }
+
+/// Represents an OAuth GitHub account record linked to a user record.
+/// Stored in the `oauth_github` table.
+#[derive(Associations, Identifiable, Selectable, Queryable, Debug, Clone)]
+#[diesel(
+    table_name = oauth_github,
+    check_for_backend(diesel::pg::Pg),
+    primary_key(account_id),
```
Turbo87, line 134: do we need to set this explicitly? seems like we've only been doing this for composite primary keys so far. or is it because we don't have an `id` column? (if that is the case, should we rename `account_id` to `id`?)

crates/crates_io_database/src/models/user.rs
```diff
@@ -124,3 +124,72 @@ impl NewUser<'_> {
...
+}
+
+impl NewOauthGithub<'_> {
+    /// Inserts the associated GitHub account info into the database, or updates an existing record.
+    ///
+    /// This is to be used for logging in when there is no currently logged-in user, as opposed to
+    /// adding another linked GitHub to a currently-logged-in user. The logic for adding another
+    /// GitHub account (when that ability gets added) will need to ensure that a particular
+    /// `account_id` (ex: GitHub account with GitHub ID 1234) is only associated with one crates.io
+    /// account, so that we know what crates.io account to log in when we get an oAuth request from
```
Turbo87: now that we have the primary key just on the `account_id`, I think the database will ensure that a GitHub account can only be linked to a single crates.io account, right?

crates/crates_io_database_dump/src/dump-db.toml
```diff
@@ -164,6 +164,15 @@ created_at = "public"
...
 
+[oauth_github.columns]
+user_id = "public"
+account_id = "public"
+encrypted_token = "private"
+login = "public"
+avatar = "public"
+[oauth_github.column_defaults]
+encrypted_token = "''"
+
```
Turbo87, line 175: do we want to make this new table public already, while it's still experimental and not used by our codebase yet?

migrations/2026-01-20-162913_oauth_github_table/up.sql
```diff
@@ -0,0 +1,21 @@
+-- safety-assured:start
+CREATE TABLE IF NOT EXISTS oauth_github (
```
Turbo87, line 2: one more thing I just thought about: do we need an index on the `user_id` column? I assume that we want to use joined queries for converting most of the existing `users` table queries?

crates/crates_io_database/src/schema.rs
carols10cents:
```
 diesel --version
diesel
 Version: 2.3.6
 Supported Backends: postgres
```

and i deleted schema.rs and my db and reran the migrations just to be sure :-/

carols10cents: aha `recent_crate_downloads` isn't from diesel, it's a view that's added in schema.patch https://github.com/rust-lang/crates.io/blob/6a538fa79245c79f9f94b9f3e55101d9a38b8ab7/crates/crates_io_database/src/schema.patch#L47-L64

Turbo87: ohhhh, yeah, you're right. I guess because we're adding a table above it and the line numbers somehow match up, it's now being added at a different spot? 😅 

AFAIK the diesel team is working on proper support for views, so hopefully this won't be an issue that much longer :)

crates/crates_io_database/src/models/user.rs
carols10cents: I don't think `account_id` should be renamed to `id` because I think that will make people assume it's a database-assigned sequential id, but it's not. So I do think this is needed to set explicitly

carols10cents: Just checked and if I take out `primary_key(account_id)`, it indeed won't compile because it's not named `id`:

```
error: no field with column name `id`
   --> crates/crates_io_database/src/models/user.rs:130:24
    |
130 | #[derive(Associations, Identifiable, Selectable, Queryable, Debug, Clone)]
    |                        ^^^^^^^^^^^^
```

crates/crates_io_database/src/schema.rs
carols10cents: I've updated the `.patch` file to put `recent_crate_downloads` in the right spot again 😅

crates/crates_io_database/src/models/user.rs
carols10cents: I've updated this comment to hopefully be accurate and clear 🙏🏻

crates/crates_io_database_dump/src/dump-db.toml
carols10cents: Good call, I've made it all private for now.

carols10cents:
@Turbo87 I believe i've addressed all your comments in new commits or responses!

PR #12802, opened 2026-01-26
admin: Alphabetize commands

Turbo87:
Extracted from https://github.com/rust-lang/crates.io/pull/12792 since it seemed independently useful :)

/cc @carols10cents


carols10cents:
🎉 Thank you!!!

PR #12803, opened 2026-01-26
database: Add `deleted_crates` table to the database dump

Turbo87:
This addresses https://github.com/rust-lang/crates.io/discussions/12716.

As discussed in our team meeting, I've looked through a bunch of the `message` column values. I did not find any sensitive content, so I think we can just make the whole table public, without having to hide the `message` contents.


jlizen:
Thanks for your work on this! Excited to consume this new data.

PR #12814, opened 2026-01-27
svelte: Add current user and dropdown menus to the `Header` component

Turbo87:
Bildschirmfoto 2026-01-27 um 23 38 49

### Related

### DIFF
--- a/svelte/src/lib/components/Header.svelte
+++ b/svelte/src/lib/components/Header.svelte
@@ -6,11 +6,10 @@
   import logo from '$lib/assets/cargo.png';
   import LockIcon from '$lib/assets/lock.svg?component';
   import ColorSchemeMenu from './ColorSchemeMenu.svelte';
+  import * as Dropdown from './dropdown';
+  import LoadingSpinner from './LoadingSpinner.svelte';
   import SearchForm from './SearchForm.svelte';
-
-  // TODO: import Dropdown from './Dropdown.svelte';
-  // TODO: import LoadingSpinner from './LoadingSpinner.svelte';
-  // TODO: import UserAvatar from './UserAvatar.svelte';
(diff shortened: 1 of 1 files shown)


svelte/src/routes/+layout.svelte
```diff
@@ -42,6 +42,9 @@
 
   let notifications = new NotificationsState();
   setNotifications(notifications);
+
+  // TODO: fetch current user
```
LawnGnome, line 46: TODO: draw the rest of the owl 😆

svelte/src/lib/components/Header.svelte
```diff
@@ -1,33 +1,59 @@
...
+
+  let isAdmin = $derived(currentUser?.is_admin ?? false);
 
   // TODO: implement session state
-  // let currentUser = $derived(session.currentUser);
-  // let isAdmin = $derived(session.isAdmin);
-  // let isSudoEnabled = $derived(session.isSudoEnabled);
-  // let sudoEnabledUntil = $derived(session.sudoEnabledUntil);
+  let isLoggingIn = $state(false);
+  let isLoggingOut = $state(false);
```
LawnGnome, line 27: Are these actually independent? Could this be represented as an enum of something like `{LOGGED_OUT, LOGGING_IN, LOGGED_IN, LOGGING_OUT}`? (Preferably with words that don't rely as much on English's broken verb conjugation system.)

LawnGnome (approved):
One non-blocking question, but LGTM.

svelte/src/lib/components/Header.svelte
Turbo87: should be possible, but they're just placeholders for now until we've drawn the rest of the owl 😄

PR #12820, opened 2026-01-29
Add CVSS scores to Security tab

djc:
Here's a local implementation.

+ The other options on npmjs kinda suck -- not a lot of tests, lack of updates, no CVSS v4
- Quite a bit of complexity, I did a bunch of spot testing but not very comprehensive

Any thoughts? If we move it into a separate module, we could potentially get test vectors from Red Hat, and once it passes those we declare it complete? But, when a new CVSS version comes out, bunch more stuff to add.

### DIFF
--- a/app/templates/crate/security.css
+++ b/app/templates/crate/security.css
@@ -65,3 +65,43 @@
 .cvss strong {
   margin-right: var(--space-2xs);
 }
+
+.cvss a {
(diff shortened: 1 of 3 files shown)


djc:
Going to close this for now. I don't think there is strong consensus on if, and how, to implement this.

PR #12825, opened 2026-01-30
Add CVSS scores to Security tab (WASM version)

djc:
This seems like a potentially interesting alternative to

- #12820 

Pros/cons:

- Use the same score calculation as RustSec upstream
- No forked scoring implementation to maintain
- Does need extra WASM infrastructure

Not sure if CI needs tweaks to install wasm-pack? Otherwise, I quite like this setup.


djc:
Going to close this for now. I don't think there is strong consensus on if, and how, to implement this.

PR #12828, opened 2026-01-30
version-list: Migrate `releaseTrackTitle` logic into the template

Turbo87:
Using `htmlSafe` for this was a bit dirty for no good reason. This moves the logic into the template, which incidentally also makes it easier to port this to Svelte later.

### DIFF
--- a/app/components/version-list/row.gjs
+++ b/app/components/version-list/row.gjs
@@ -3,12 +3,12 @@ import { on } from '@ember/modifier';
 import { action } from '@ember/object';
 import { LinkTo } from '@ember/routing';
 import { service } from '@ember/service';
-import { htmlSafe } from '@ember/template';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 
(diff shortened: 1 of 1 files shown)


app/components/version-list/row.gjs
```diff
@@ -102,7 +67,21 @@ export default class VersionRow extends Component {
...
+            {{else if @version.invalidSemver}}
+              Failed to parse version
+              {{@version.num}}
+            {{else}}
+              Release Track:
+              {{@version.releaseTrack}}
+              {{#if (or @version.isPrerelease @version.isHighestOfReleaseTrack)}}
+                ({{#if @version.isPrerelease}}<span class='rt-prerelease'>prerelease</span>{{/if}}{{#if
+                  (and @version.isPrerelease @version.isHighestOfReleaseTrack)
+                }}, {{/if}}{{#if @version.isHighestOfReleaseTrack}}<span class='rt-latest'>latest</span>{{/if}})
```
Turbo87, line 82: not pretty, but whitespace control is complicated 🙈😅

PR #12836, opened 2026-01-31
ember: Use `ember-cli-deprecation-workflow` to silence noisy deprecation warnings

Turbo87:
Our frontend CI logs are currently flooded with deprecation warnings, primarily from the Ember Data project. This PR sets up the `ember-cli-deprecation-workflow` addon to silence these warnings since we are migrating away from Ember.js and Ember Data anyway.


Turbo87:
seems to work well. the CI logs are much more readable. :shipit:

PR #12840, opened 2026-01-31
svelte: Implement `/crates/[crate_id]/versions` page

Turbo87:
Bildschirmfoto 2026-01-31 um 14 55 20

This roughly matches the corresponding Ember.js route implementation, but it's using the SvelteKit `load()` fn to load the initial set of versions and then a regular async fn to load additional pages. The `Row` component is also similar in terms of styling and behavior, but the API is slightly different since it accepts a raw `version` object, without a relationship to the corresponding crate or any derived properties.

I've also included a Storybook file for the `Row` component. Compare to the other Storybook files it only contains a single Story with all of the variations though. The reason for that is that I'm trying to reduce the number of snapshots that we use for visual regression testing, since that number is somewhat limited per month.

### DIFF
--- a/svelte/src/routes/crates/[crate_id]/versions/+page.svelte
+++ b/svelte/src/routes/crates/[crate_id]/versions/+page.svelte
@@ -1,6 +1,176 @@
 <script lang="ts">
+  import type { components } from '@crates-io/api-client';
+
+  import { createClient } from '@crates-io/api-client';
+  import { format } from 'date-fns';
+
+  import CrateHeader from '$lib/components/CrateHeader.svelte';
+  import LoadingSpinner from '$lib/components/LoadingSpinner.svelte';
+  import * as SortDropdown from '$lib/components/sort-dropdown';
+  import Row from '$lib/components/version-list/Row.svelte';
+
+  type Version = components['schemas']['Version'];
+
   let { data } = $props();
+
+  let isLoading = $state(false);
+
+  // Client-side state for extra loaded versions and pagination
+  let extraVersions = $state<Version[]>([]);
+  let extraNextPage = $state<string | null>(null);
+
+  // Reset client-side state when sort changes
+  $effect(() => {
+    void data.sort;
+
+    extraVersions = [];
+    extraNextPage = null;
+  });
+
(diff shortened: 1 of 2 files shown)


svelte/src/lib/components/version-list/Row.svelte
```diff
@@ -0,0 +1,520 @@
...
+      {#if version.yanked}
+        <TrashIcon />
+      {:else if !semver}
+        ?
+      {:else}
+        {releaseTrack}
+      {/if}
+
+      <Tooltip side="right">
+        <div class="rt-tooltip">
```
eth3lbert, line 133: A non-blocking nit: it seems we're missing a `data-test-release-track-title` attribute here. Since it isn't being used anywhere currently, I'm not against leaving it out, but I just wanted to give you a heads-up.

svelte/src/lib/components/version-list/Row.svelte
```diff
@@ -0,0 +1,520 @@
...
+          via
+          {#if trustpubUrl}
+            <!-- eslint-disable-next-line svelte/no-navigation-without-resolve -->
+            <a href={trustpubUrl} target="_blank" rel="nofollow noopener noreferrer">
+              {#if trustpubProvider === 'github'}
+                <GitHubIcon />
+              {:else if trustpubProvider === 'gitlab'}
+                <GitLabIcon />
+              {/if}
+              {trustpubPublisher}
```
eth3lbert, line 186: Do you think this is worth extracting into a snippet or a component?

eth3lbert (approved):
I’ve just left a few minor non-blocking nits, but overall it LGTM! Thanks!

svelte/src/lib/components/version-list/Row.svelte
Turbo87: yeah, I noticed that it wasn't used, so I left it out for now :)

svelte/src/lib/components/version-list/Row.svelte
Turbo87: I thought about it, yeah. but for now it's just 6 relatively straight-forward lines, so I wasn't sure yet whether it would be worth it.

PR #12861, opened 2026-02-03
Use oauth_github table for user info for crates.io/users/[username]

carols10cents:
Connects to https://github.com/rust-lang/crates.io/issues/10611. 

Rather than using the users table.

The queries to the `oauth_github` table should produce the exact same JSON for the API requests as before; this only changes where the information is coming from in service of eventually being able to deprecate `users.gh_id`, `users.gh_login`, `users.gh_encrypted_token`, and `users.avatar` in favor of the `oauth_github` table data.

The `EncodablePublicUser` JSON is currently getting its data from the first returned linked account, which for now can only be GitHub and there should only ever be 1 (or 0). What I'm thinking for the future is to continue populating the `login`, `avatar`, and `url` fields with the first linked account to maintain backwards compatibility, but eventually introduce a `linked_accounts` field that will contain all linked accounts (and we'll have to discuss how to display multiple linked accounts when that's possible).

When this is deployed, a difference will be that if you visit the page of a user whose gh_id is currently -1 in the users table (ex: https://crates.io/users/GBGamer), you get a page with the gh_login and gh_avatar that we've saved, and we link to github.com/gh_login, but we know those things are no longer valid. This PR changes the behavior to return a "user not found" page in that case, because the `oauth_github` table won't have a record with the login. In some ways, this is more correct (the user doesn't actually exist in github anymore), in some ways it's less correct (there's still info in our database right now and we'll be saying there isn't, but eventually we should get rid of those user records IMO).

### DIFF
--- a/crates/crates_io_api_types/src/lib.rs
+++ b/crates/crates_io_api_types/src/lib.rs
@@ -8,8 +8,9 @@ pub use self::krate_publish::{EncodableCrateDependency, PublishMetadata};
 
 use chrono::{DateTime, Utc};
 use crates_io_database::models::{
-    ApiToken, Category, Crate, Dependency, DependencyKind, Keyword, Owner, ReverseDependency, Team,
-    TopVersions, TrustpubData, User, Version, VersionDownload, VersionOwnerAction,
+    ApiToken, Category, Crate, Dependency, DependencyKind, Keyword, LinkedAccount, Owner,
+    ReverseDependency, Team, TopVersions, TrustpubData, User, UserWithLinkedAccounts, Version,
+    VersionDownload, VersionOwnerAction,
 };
 use serde::{Deserialize, Serialize};
 
@@ -791,6 +792,39 @@ impl From<User> for EncodablePublicUser {
             id,
             avatar: gh_avatar,
             login: gh_login,
+            name,
+            url,
+        }
+    }
+}
+
+/// Converts a `UserWithLinkedAccounts` model into an `EncodablePublicUser` for JSON serialization.
+///
+/// Uses the first `LinkedAccount` and ignores any others for now for JSON API compatibility; when
+/// there may be multiple linked accounts, we'll need to rethink this. Uses placeholders if there
+/// aren't any linked accounts.
+impl From<UserWithLinkedAccounts> for EncodablePublicUser {
+    fn from(user: UserWithLinkedAccounts) -> Self {
+        let UserWithLinkedAccounts {
+            user: User { id, name, .. },
+            linked_accounts,
+            ..
+        } = user;
+
+        let first_linked_account = linked_accounts.first();
+        let avatar = first_linked_account
+            .and_then(LinkedAccount::avatar)
+            .map(ToOwned::to_owned);
+        let login = first_linked_account
+            .map(LinkedAccount::login)
+            .unwrap_or("ghost")
+            .to_owned();
+        let url = format!("https://github.com/{login}");
+
+        EncodablePublicUser {
+            id,
+            avatar,
+            login,
             name,
             url,
         }
--- a/crates/crates_io_database/src/models/mod.rs
+++ b/crates/crates_io_database/src/models/mod.rs
@@ -18,7 +18,9 @@ pub use self::owner::{CrateOwner, Owner, OwnerKind};
 pub use self::team::{NewTeam, Team};
 pub use self::token::ApiToken;
 pub use self::trustpub::TrustpubData;
-pub use self::user::{NewOauthGithub, NewUser, OauthGithub, User};
+pub use self::user::{
+    LinkedAccount, NewOauthGithub, NewUser, OauthGithub, User, UserWithLinkedAccounts,
+};
 pub use self::version::{NewVersion, TopVersions, Version};
 
 pub mod helpers;
--- a/crates/crates_io_database/src/models/user.rs
+++ b/crates/crates_io_database/src/models/user.rs
@@ -133,6 +133,20 @@ pub enum LinkedAccount {
     Github(OauthGithub),
 }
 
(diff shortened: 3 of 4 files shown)


Turbo87:
> The `EncodablePublicUser` JSON is currently getting its data from the first returned linked account, which for now can only be GitHub and there should only ever be 1 (or 0). What I'm thinking for the future is to continue populating the `login`, `avatar`, and `url` fields with the first linked account to maintain backwards compatibility, but eventually introduce a `linked_accounts` field that will contain all linked accounts (and we'll have to discuss how to display multiple linked accounts when that's possible).

tbh I'm not sure this is the right way forward, though it's possible that I'm misunderstanding the proposal.

the current situation is that the crates.io account name is the same as the GitHub account name and we only have a single database column for them.

if I understand correctly, this PR proposes to instead read the GitHub (and crates.io) account name from the `oauth_github` table instead, but potentially also from a `oauth_gitlab`, `oauth_codeberg`, ... table in the future. how will that deal with conflicts? we already have such conflicts in the GitHub case with account resurrections, but if we use the account names from other identity providers this problem will become even worse.

I think the way forward is to decouple the GitHub and crates.io account names. when you first login we should make you choose a crates.io account name. that name can default to your GitHub (or GitLab, Codeberg, ...) account name, but if that account name is already taken you will have to choose a different one. I expect us to save that account name in the `users` table, since it's unrelated to OAuth. I would also expect the `/users/[username]` page to use that account name from the `users` table, and not the names in the OAuth tables, which is why I'm sceptical about this proposal.

The problem with my proposal is that people are used to the `GitHub account name == crates.io account name` relationship and when you `cargo owner --add foo` everyone assumes that `foo` is the GitHub account name. One way around that would be to require a prefix (e.g. `crates.io:foo` vs. `github:foo`) for a migration period until this change has settled in peoples minds.

carols10cents:
Ah, yeah, I've been thinking of a new direction here, and also thinking of this as a smaller step maybe.

I've been thinking about how a large percentage of our users are probably going to have one GitHub account associated with crates.io, and those GitHub usernames will likely be unique, even when there are possible other logins. 

So to optimize for the common case, if we search all the `oauth_*` tables for the username from `crates.io/users/[username]` and only get one result, we can just show that crates.io user.

If we get more than one result matching a username when we query all the `oauth_*` tables (after grouping by crates.io user because the next most likely case will be someone who has the same username on github and gitlab and connects them both), then we show a disambiguation page like on wikipedia, something like:

```
There are multiple users with username `[username]`. Did you mean:

- [username on GitHub](https://crates.io/users/[username]/[crates.io user id])
- [username on GitLab]((https://crates.io/users/[username]/[different crates.io user id])
```

(and we'd implement those routes to look up users based on the unambiguous crates.io ID in the URL)

I could see some users being annoyed that their username needs an extra click to get to their page because someone else took the username on a different service, but it feels like this would be less vulnerable to impersonation and there would be fewer complaints about claimed/"squatted" crates.io usernames.

Similarly, when adding by username, if there's only one result for the specified username then it's unambiguous-- if there are multiple, we can send back a response like the disambiguation page that says something like:

```
There are multiple users with username `[username]`.

- If you meant `http://github.com/username`, run `cargo owner add [username]:[crates.io user id]`
- If you meant `http://gitlab.com/username`, run `cargo owner add [username]:[different crates.io user id]`
```

Or as you said we could support `cargo owner add github:[username]` which would only look in the oauth_github table for that username, and would be different from teams because teams are specified with `github:org:team`.

I previously was thinking we'd need the crates.io username as you laid out. I don't think this PR necessarily commits us to one way or the other, but this is a step forward to being able to remove `users.gh_login`, which we'd need to do in either case, and this PR doesn't actually change any user-facing behavior yet (other than for invalid github users).

If this doesn't seem to be a useful step, I'm happy to go in a different direction. Maybe we can talk about this on friday (i'm adding it to the agenda)?

Turbo87:
> but this is a step forward to being able to remove `users.gh_login`, which we'd need to do in either case

I'm not sure that we actually need or want to remove it. I think it would need to be renamed to e.g. `login` instead though, since it would be the crates.io account name going forward, instead of the GitHub account name. The only small complication with this is that the column is currently not marked as unique due to account resurrections, so we would have to clean up the dead accounts first before we mark it as unique. I guess we would also have to figure out what to do about the ghost accounts for deleted users.

> I've been thinking about how a large percentage of our users are probably going to have one GitHub account associated with crates.io, and those GitHub usernames will likely be unique, even when there are possible other logins.

yeah, that's true. I guess we could still support e.g. `cargo owner --add foo` in the case where the crates.io and GitHub account names match up, and only fail if they diverge. That should likely cover 90% of the cases and for the others we have a secure way of disambiguating.

> Maybe we can talk about this on friday

yeah, I think that would be good. although, @LawnGnome is still on PTO this week, and I think he had some thoughts on the topic as well.

eth3lbert:
> I would also expect the `/users/[username]` page to use that account name from the `users` table

I’m also leaning toward his approach, but I agree that doing nothing leaves room for impersonation and name squatting. 


In cases of conflict, I’m not sure we need a disambiguation page. Perhaps just adding a clear icon near the username or prefixing it with `crates.io:` would be enough to tell them apart.

Turbo87:
> Perhaps just adding a clear icon near the username or prefixing it with `crates.io:` would be enough to tell them apart.

I was thinking maybe a warning banner of some sort that clearly indicates if there is a mismatch between crates.io and GitHub username?

carols10cents:
We talked about this in our meeting on Friday, and decided to probably go the direction @Turbo87 and @eth3lbert were expecting, but that it's time to get an RFC like thing going and choosing a direction rather than making suboptimal code changes in an attempt to delay that decision 😂

PR #12949, opened 2026-02-12
Add more tests around user account behavior

carols10cents:
These are tests I extracted (plus some additional ones) from this PR we decided to close:

<https://github.com/rust-lang/crates.io/pull/12861>

The tests are still good and add coverage for user account edge cases we'll need to consider as we work on supporting multiple login services.

### DIFF
--- a/src/tests/routes/users/read.rs
+++ b/src/tests/routes/users/read.rs
@@ -4,6 +4,7 @@ use crates_io::models::NewUser;
 use crates_io::schema::users;
 use crates_io::views::EncodablePublicUser;
 use diesel_async::RunQueryDsl;
+use insta::assert_snapshot;
 use serde::Deserialize;
 
 #[derive(Deserialize)]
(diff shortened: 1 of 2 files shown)


src/tests/owners.rs
```diff
@@ -867,6 +867,51 @@ async fn highest_gh_id_is_most_recent_account_we_know_of() {
...
+
+    // Adding a user that doesn't exist at all also fails
+    let response = owner_token
+        .add_named_owner(krate_name, "nonexistent_username")
+        .await;
+    assert_snapshot!(response.status(), @"400 Bad Request");
+    assert_snapshot!(
+        response.text(),
+        @r#"{"errors":[{"detail":"could not find user with login `nonexistent_username`"}]}"#
+    );
```
Turbo87: there is some overlap here with: https://github.com/rust-lang/crates.io/blob/aa4548c32eb1b0a8d5626f88397903cda54204b9/src/tests/routes/crates/owners/add.rs#L296-L308

do you think we need both? I'm not sure if anything in this setup differs significantly 😅

carols10cents: oooooooops nope, i'll take the new one back out!

PR #12964, opened 2026-02-13
Use oauth_github table for sync_admins

carols10cents:
This is a tiny, low-risk way to start using the `oauth_github` table as the source of truth for GitHub IDs in the background process that syncs from the team repo to set the admin flag on users.

I volunteer as tribute to be removed (and then readded 🤞🏻 ) as an admin in the team repo to test this out 😁

### DIFF
--- a/src/worker/jobs/sync_admins.rs
+++ b/src/worker/jobs/sync_admins.rs
@@ -119,8 +119,12 @@ impl BackgroundJob for SyncAdmins {
             info!("Granted admin access: {added_admins}");
         }
 
-        // New admins from the team repo that don't have a crates.io
-        // account yet, so we can't find their GitHub ID in the database.
+        // From the set of all GitHub IDs of admins from the team repo,
+        // subtract all GitHub IDs we found in the database (whether they were already admins or
+        // whether we just added them; both are included in `database_user_github_ids`).
+        // This leaves the GitHub IDs from the team repo that we _weren't_ able to find in the
+        // database to make them admins, likely because they don't have a crates.io account
+        // associated with their GitHub account yet.
         let skipped_new_admin_github_ids = repo_admin_github_ids
             .difference(&database_user_github_ids)
             .copied()


src/worker/jobs/sync_admins.rs
```diff
@@ -77,12 +81,31 @@ impl BackgroundJob for SyncAdmins {
...
+                .returning(users::id)
                 .get_results::<i32>(&mut conn)
-                .await?
+                .await?;
+            new_admin_user_account_ids
+                .into_iter()
+                .filter_map(|(user_id, account_id)| {
+                    updated_user_ids.contains(&user_id).then_some(account_id)
+                })
+                .collect()
```
Turbo87: I admit I'm having a hard time following what the code does here. I wonder if it would be easier to do a `gh_id -> user_id` lookup for all the repo admins first and then only work with the `users.id` fields afterwards. do you think that could simplify the logic a bit?

carols10cents: Sure, that'd be a change in behavior though, is that ok? Currently a lot of this logic is to be able to print out the admins' GitHub IDs.

Turbo87: > Currently a lot of this logic is to be able to print out the admins' GitHub IDs.

hmm... I guess I chose that mostly just because it was available already. Since `users.id` would be similarly unambiguous, I think displaying those would be okay too?

carols10cents:
@Turbo87 Ok, I've substantially rewritten this change. Instead of doing a bunch of complex queries, this is now doing one SELECT for all the information this job needs from the database, then filtering through that and collecting user IDs for the various situations, and sending UPDATEs using only user IDs.

Also nevermind my comment before about needing to switch the output from github ID to user ID, because all the information is in the `Vec<UserData>`, we can filter those instances by user ID but display the github ID from the instance.

The only case that can't use user IDs is for github IDs in the team repo that have never logged into crates.io and therefore don't have crates.io users we can find.

src/worker/jobs/sync_admins.rs
```diff
@@ -27,132 +27,159 @@ impl BackgroundJob for SyncAdmins {
...
+            gh_login: String,
+            is_admin: bool,
+            account_id: i64,
+            email: Option<String>,
+        }
 
-        let database_admins = users::table
+        // Fetch all database info for all accounts that are either currently marked as admins
+        // or are admins according to the team repo.
+        let database_user_data = (users::table.inner_join(oauth_github::table))
```
Turbo87: since we are working on allowing user accounts that don't have an associated GitHub account, I wonder about the following future scenario: 

user X is logged in via GitHub, user X is an admin, user X decides to login with non-GitHub now and disassociates their account from GitHub, user X leaves the crates.io admin team. 

in this case the inner join would cause the user to not be selected in the query, making it look like the admin flag was already reset to `false`.

I assume this could be resolved by making `UserData::account_id` an `Option` and using `left_join(oauth_github::table)` instead. does that make sense?

src/worker/jobs/sync_admins.rs
```diff
@@ -27,132 +27,159 @@ impl BackgroundJob for SyncAdmins {
...
 
-        // Existing admins from the database.
+        #[derive(Debug, Queryable)]
+        struct UserData {
+            id: i32,
+            gh_login: String,
+            is_admin: bool,
+            account_id: i64,
+            email: Option<String>,
+        }
```
Turbo87: I've started using the `HasQuery` derive from diesel a bit more lately, which allows you to do this:

```suggestion
        #[derive(Debug, HasQuery)]
        #[diesel(base_query = users::table.inner_join(oauth_github::table).left_join(emails::table))]
        struct UserData {
            #[diesel(select_expression = users::id)]
            id: i32,
            #[diesel(select_expression = users::gh_login)]
            gh_login: String,
            #[diesel(select_expression = users::is_admin)]
            is_admin: bool,
            #[diesel(select_expression = oauth_github::account_id)]
            account_id: i64,
            #[diesel(select_expression = emails::email.nullable())]
            email: Option<String>,
        }
```

in that case you can simplify the query code below to:

```rust
        let database_user_data = UserData::query()
            .filter(
                users::is_admin
                    .eq(true)
                    .or(oauth_github::account_id.eq_any(&repo_admin_github_ids)),
            )
            .get_results(&mut conn)
            .await?;
```

the primary advantage AFAICT is that is avoids field order mismatches between the struct and the `.select(...)` column tuple.

carols10cents: Oh nice! I kept trying to derive `QueryableByName` and failing 😪

src/worker/jobs/sync_admins.rs
carols10cents: Great catch! I added a failing test and then changed from inner to left join to handle this!

carols10cents:
@Turbo87 ready for rereview!

src/worker/jobs/sync_admins.rs
```diff
@@ -27,132 +27,171 @@ impl BackgroundJob for SyncAdmins {
...
-        let added_admin_ids = if new_admin_ids.is_empty() {
+        let new_admin_user_ids = database_user_data
+            .iter()
+            .filter_map(|u| {
+                (u.account_id
+                    .is_some_and(|account_id| repo_admin_github_ids.contains(&account_id))
+                    && !u.is_admin)
+                    .then_some(u.id)
+            })
+            .collect::<HashSet<i32>>();
```
Turbo87, line 98:
```suggestion
        let new_admin_user_ids = database_user_data
            .iter()
            .filter(|u| !u.is_admin)
            .map(|u| u.id)
            .collect::<HashSet<i32>>();
```

correct me if I'm wrong, but I think this should be equivalent? 

if I understand correctly, `database_user_data` contains all users that are currently admins, or that the team repo defines as admins but still need to be promoted on our side. so if we filter out all of the existing admins from `database_user_data` we should automatically end up with the ones that need to be promoted.

src/worker/jobs/sync_admins.rs
```diff
@@ -27,132 +27,171 @@ impl BackgroundJob for SyncAdmins {
...
-
-        let removed_admin_ids = if obsolete_admin_ids.is_empty() {
+        let obsolete_admin_user_ids = database_user_data
+            .iter()
+            .filter_map(|u| {
+                (u.is_admin
+                    && u.account_id
+                        .is_none_or(|account_id| !repo_admin_github_ids.contains(&account_id)))
+                .then_some(u.id)
+            })
```
Turbo87: similar to the above comment, I think the `is_admin` condition is probably redundant.

```suggestion
            .filter(|u| u.account_id.is_none_or(|account_id| !repo_admin_github_ids.contains(&account_id)))
            .map(|u| u.id)
```

src/worker/jobs/sync_admins.rs
```diff
@@ -27,132 +27,171 @@ impl BackgroundJob for SyncAdmins {
...
         }
 
         // New admins from the team repo that don't have a crates.io
-        // account yet.
-
-        let skipped_new_admin_ids = new_admin_ids
-            .difference(&added_admin_ids)
+        // account yet, so we can't find their GitHub ID in the database.
+        let skipped_new_admin_github_ids = repo_admin_github_ids
+            .difference(&database_user_github_ids)
```
Turbo87, line 129: took me a second to understand why this is correct. I was reading `database_admin_github_ids` instead of `database_user_github_ids`, but since `database_user_github_ids` also contains the admins that were just added, it should work as intended.

src/worker/jobs/sync_admins.rs
carols10cents: no you're right! nice catch here and below!!

src/worker/jobs/sync_admins.rs
carols10cents: I've added a commit expanding on this comment a bit to hopefully help next time someone has to read this?

PR #13010, opened 2026-02-20
real_ip: Add Fastly CDN support

Turbo87:
This extends the IP filtering to cover Fastly in addition to CloudFront. The update script now fetches both AWS CloudFront and Fastly IP ranges, generating separate Rust source files for each. The CI workflow is renamed to reflect the broader CDN scope.

Note that this PR intentionally excludes two Fastly CIDRs, so that we can test that the CI workflow works as intended.


Turbo87:
I'll trigger the workflow manually

Turbo87:
ah nevermind, changes on the main branch trigger it automatically

Turbo87:
and it did what it was supposed to do: https://github.com/rust-lang/crates.io/commit/933efb1310191b40d90b63591b5a10ceb79a5dc3

next step, wait until it's deployed to staging, then test it out on staging and then promote to production

PR #13023, opened 2026-02-21
Only show the report crate button for logged in users

carols10cents:
Experiment to see if this affects the quantity of spam reports we get.

### DIFF
--- a/app/components/support/crate-report-form.gjs
+++ b/app/components/support/crate-report-form.gjs
@@ -40,6 +40,7 @@ const REASONS = [
 
 export default class CrateReportForm extends Component {
   @service store;
+  @service session;
 
   @tracked crate = '';
   @tracked selectedReasons = [];
@@ -114,6 +115,9 @@ ${reasons}
 Additional details:
 
 ${this.detail}
+
+Thanks,
+${this.session.currentUser.login}
 `;
     let subject = `The "${crate}" crate`;
     if (isMaliciousCodeReport) {
--- a/e2e/acceptance/support.spec.ts
+++ b/e2e/acceptance/support.spec.ts
@@ -2,7 +2,12 @@ import { expect, test } from '@/e2e/helper';
 
 test.describe('Acceptance | support page', { tag: '@acceptance' }, () => {
   test.beforeEach(async ({ page, msw }) => {
-    let user = await msw.db.user.create({});
+    let user = await msw.db.user.create({
+      login: 'johnnydee',
+      name: 'John Doe',
+      email: 'john@doe.com',
+      avatar: 'https://avatars2.githubusercontent.com/u/1234567?v=4',
+    });
     await msw.authenticateAs(user);
 
     let crate = await msw.db.crate.create({ name: 'nanomsg' });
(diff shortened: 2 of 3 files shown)


carols10cents:
@Turbo87 wdyt? 

- First commit: only show the "report crate" button if you're logged in (no change to the "support" link in the footer though
- Second commit: only show the support page's content if you're logged in
- Third commit: Add the current user's username to the generated mailto's text

Now, the user could edit the message to remove their username, but if the spam we're getting is a crawler/bot/agent taking action without any human intervention it seems likely that either submissions will just stop (because it won't be logged in) or will send the report with the logged-in username.

And if we keep getting the same emails with no change, that means the spam is being sent directly via a script outside of interactions with the site, which, i guess we're just out of luck there but this might prevent new spam?

Also I don't seem to have permissions to see the changes in chromatic, and I don't understand how to fix the Svelte error I'm getting:

```
Svelte error: missing_context
Context was not set in a parent component
https://svelte.dev/e/missing_context

	in <unknown>
```

app/components/crate-sidebar.gjs
```diff
@@ -214,14 +216,16 @@ export default class CrateSidebar extends Component {
...
+        {{#if this.session.currentUser}}
+          <LinkTo
+            @route='support'
+            @query={{hash inquire='crate-violation' crate=@crate.name}}
+            data-test-id='link-crate-report'
+            class='report-button button button--red button--small'
+          >
+            Report crate
+          </LinkTo>
+        {{/if}}
```
LawnGnome: I think we might want to keep the button for non-logged-in users, but disabled with a tooltip, so that people with accounts who just aren't logged in know there's still a path to report crates.

LawnGnome: Oh, I see @eth3lbert said the same thing. Sorry; don't know how I missed that!

Turbo87:
> Experiment to see if this affects the quantity of spam reports we get.

how do we measure it to see if the experiment is successful? I'm not sure we get a consistent enough number of spam emails to be able to properly compare it 😅 

aside from that latest wave, I think the signal to noise ratio was relatively good and even that latest wave was just a couple of clicks to move them into spam.

> @Turbo87 wdyt?

I'm honestly still not a big fan of this. IMHO the real spam is easily filtered out, but this PR adds friction to the whole reporting process, which I think has been working quite well since we introduced it.

I think with https://github.com/rust-lang/team/pull/2259 being merged, I just don't see the need to complicate the reporting process for our legitimate users.

> but if the spam we're getting is a crawler/bot/agent taking action without any human intervention

I would be quite surprised to learn that any of the usual crawlers/bots/agents are handling `mailto:` links on website by automatically sending emails to them, and especially if they first have to fill out a form to even generate the `mailto:` URL.

If we are getting automated spam emails, then these are more likely to be email-specific bots that are unrelated to our support form on the website.

> I don't understand how to fix the Svelte error I'm getting

hmm, I don't see any Svelte changes in this PR?

> Add the current user's username to the generated mailto's text

I think that part is useful. I wonder if we should make that an input field for unauthenticated users, which is automatically filled if the user is authenticated. Then again, it's easy to pretend to be someone else that way, and it would also be easy to change the generated email text, so idk how useful it would ultimately be.

carols10cents:
I removed the first commit that hid the "report crate" button if you weren't logged in (I think that's where the Svelte changes were).

But seeing as @Turbo87 isn't sure this is worth it, I'll just close :)

Issue #13032, opened 2026-02-23
Parent Category Crate Count is Wrong when Crates Belong to Multiple Subcategories


TKanX:
### Current Behavior

The crate count shown on a parent category is the sum of all subcategory counts. If a crate belongs to more than one subcategory under the same parent, it gets counted more than once.

Real example from `Science::Computational Chemistry`:

The category listing shows **5 crates**:

<img width="753" height="65" alt="Image" src="https://github.com/user-attachments/assets/1954c237-16e2-440a-9ee8-368076f51bcb" />

But inside the category page, the subcategories are:
- `Molecular Simulation`: 3 crates
- `Cheminformatics`: 2 crates
- `Electronic Structure`: 0 crates

And the actual crate list says **"Displaying 1-3 of 3 total results"**:

<img width="768" height="939" alt="Image" src="https://github.com/user-attachments/assets/c75f66a6-a71c-462e-8517-cdb16d3b3bcd" />

So the count shown in the listing (5) does not match the real number of unique crates (3).

### Expected Behavior

The parent category count should show the number of unique crates across all its subcategories. Every crate should be counted only once.

### Steps To Reproduce

1. Go to the categories listing page (`/categories`).
2. Find a parent category with subcategories, e.g. `Computational Chemistry`.
3. Note the crate count shown in the listing.
4. Click into the category and compare with the actual crate list total shown at the bottom.
5. The count in the listing is higher than the real number of unique crates.

### Environment

- Browser: Any
- OS: Any

### Anything else?

The root cause is in `toplevel.sql`. It uses `sum(c2.crates_cnt)` to calculate the parent count, which adds up per-subcategory counters without deduplication. The fix would be to count distinct crate IDs instead.

PR #13051, opened 2026-02-25
Fix double-counting in parent category crate counts

Aakash-Naidu:
This PR resolves issue #13032, where parent categories incorrectly inflated crate counts when a crate belonged to multiple subcategories under the same parent.

The root cause was the use of SUM(c2.crates_cnt) on cached subcategory counters, which caused duplicate counting when the same crate appeared in multiple subcategories.
__________________________________________________________________________________________________________________________________
**Implementation Details**

The aggregation logic has been updated to count unique crate IDs directly from the crates_categories relationship table.

SQL Changes

The following queries were updated to use:

`COUNT(DISTINCT cc.crate_id)`

with a join to crates_categories:

toplevel.sql — fixes counts on the main Categories landing page.

subcategories.sql — ensures accurate counts for child categories.

parent_categories.sql — fixes counts in parent breadcrumb navigation.

This ensures each crate contributes at most once to a parent category count.
____________________________________________________________________________________________________________________
**Performance Considerations**

Since the updated logic introduces a join with crates_categories, an index was added to maintain query performance:

Migration: 2026-02-24-171800_add_category_id_index_to_crates_categories

Adds a B-Tree index on crates_categories(category_id)

This allows PostgreSQL to use an index scan instead of a sequential scan when resolving category joins.
_____________________________________________________________________________________________________________________
**Testing & Verification**
Reproduction Test

Added category_toplevel_unique_count_reproduction, which:

Creates a crate assigned to multiple subcategories

Verifies the parent category counts it only once

This test fails under the previous implementation and passes with the updated logic.
_____________________________________________________________________________________________________
**Regression Updates**

Existing tests in models/category.rs were updated to use real crate–category relationships instead of relying on manually set crates_cnt values.

Status

All database tests in crates_io_database pass successfully.

### DIFF
--- a/crates/crates_io_database/src/models/category.rs
+++ b/crates/crates_io_database/src/models/category.rs
@@ -198,23 +198,74 @@ mod tests {
     #[tokio::test]
     async fn category_toplevel_orders_by_crates_cnt_when_sort_given() {
         use self::categories;
-
-        let new_cat = |category, slug, crates_cnt| {
-            (
-                categories::category.eq(category),
-                categories::slug.eq(slug),
-                categories::crates_cnt.eq(crates_cnt),
-            )
-        };
+        use crate::schema::crates;
+        use crate::schema::crates_categories;
 
         let test_db = TestDatabase::new();
         let mut conn = test_db.async_connect().await;
 
         diesel::insert_into(categories::table)
             .values(&vec![
-                new_cat("Cat 1", "cat1", 0),
-                new_cat("Cat 2", "cat2", 2),
-                new_cat("Cat 3", "cat3", 1),
+                (
+                    categories::category.eq("Cat 1"),
+                    categories::slug.eq("cat1"),
+                ),
+                (
+                    categories::category.eq("Cat 2"),
+                    categories::slug.eq("cat2"),
+                ),
+                (
+                    categories::category.eq("Cat 3"),
+                    categories::slug.eq("cat3"),
+                ),
+            ])
+            .execute(&mut conn)
+            .await
+            .unwrap();
+
+        let _cat1: Category = Category::by_slug("cat1")
+            .select(Category::as_select())
+            .first(&mut conn)
+            .await
+            .unwrap();
+        let cat2: Category = Category::by_slug("cat2")
+            .select(Category::as_select())
+            .first(&mut conn)
+            .await
+            .unwrap();
+        let cat3: Category = Category::by_slug("cat3")
+            .select(Category::as_select())
+            .first(&mut conn)
+            .await
+            .unwrap();
+
+        let insert_crate = |name: &'static str| {
+            diesel::insert_into(crates::table)
+                .values(crates::name.eq(name))
+                .returning(crates::id)
+        };
+
+        // Cat 1: 0 crates
+        // Cat 2: 2 crates
+        let k1: i32 = insert_crate("k1").get_result(&mut conn).await.unwrap();
+        let k2: i32 = insert_crate("k2").get_result(&mut conn).await.unwrap();
+        // Cat 3: 1 crate
+        let k3: i32 = insert_crate("k3").get_result(&mut conn).await.unwrap();
+
+        diesel::insert_into(crates_categories::table)
+            .values(&vec![
+                (
+                    crates_categories::crate_id.eq(k1),
+                    crates_categories::category_id.eq(cat2.id),
+                ),
+                (
+                    crates_categories::crate_id.eq(k2),
+                    crates_categories::category_id.eq(cat2.id),
+                ),
+                (
+                    crates_categories::crate_id.eq(k3),
+                    crates_categories::category_id.eq(cat3.id),
+                ),
             ])
             .execute(&mut conn)
             .await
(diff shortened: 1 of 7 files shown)


Turbo87:
> Since the updated logic introduces a join with crates_categories, an index was added to maintain query performance

did you measure the performance impact of this PR on all of the affected queries with a production dataset? I'd be somewhat surprised if this didn't impact performance at all 😅 

@eth3lbert any thoughts on this PR?

eth3lbert:
> > Since the updated logic introduces a join with crates_categories, an index was added to maintain query performance
> 
> did you measure the performance impact of this PR on all of the affected queries with a production dataset? I'd be somewhat surprised if this didn't impact performance at all 😅
> 
> @eth3lbert any thoughts on this PR?

I've tested it locally and it seems to have only a small impact! The following shows the most impactful EXPLAIN ANALYZE result.
<details>

<summary> old </summary>

```sql
Sort  (cost=84.36..84.37 rows=5 width=144) (actual time=0.031..0.032 rows=1 loops=1)
  Sort Key: c.path
  Sort Method: quicksort  Memory: 25kB
  InitPlan 2
    ->  Index Scan using categories_slug_key on categories  (cost=0.15..8.17 rows=1 width=32) (actual time=0.006..0.006 rows=1 loops=1)
          Index Cond: ((slug)::text = 'science::computational-chemistry'::text)
  ->  Bitmap Heap Scan on categories c  (cost=4.18..76.14 rows=5 width=144) (actual time=0.029..0.029 rows=1 loops=1)
        Recheck Cond: (path @> (InitPlan 2).col1)
        Filter: ((slug)::text <> 'science::computational-chemistry'::text)
        Rows Removed by Filter: 1
        Heap Blocks: exact=2
        ->  Bitmap Index Scan on path_gist_categories_idx  (cost=0.00..4.18 rows=5 width=0) (actual time=0.014..0.014 rows=2 loops=1)
              Index Cond: (path @> (InitPlan 2).col1)
        SubPlan 1
          ->  Aggregate  (cost=12.68..12.69 rows=1 width=4) (actual time=0.010..0.010 rows=1 loops=1)
                ->  Bitmap Heap Scan on categories c2  (cost=4.18..12.67 rows=5 width=4) (actual time=0.007..0.008 rows=16 loops=1)
                      Recheck Cond: (path <@ subltree(c.path, 0, 2))
                      Heap Blocks: exact=3
                      ->  Bitmap Index Scan on path_gist_categories_idx  (cost=0.00..4.18 rows=5 width=0) (actual time=0.005..0.005 rows=16 loops=1)
                            Index Cond: (path <@ subltree(c.path, 0, 2))
Planning Time: 0.054 ms
Execution Time: 0.054 ms
```

</details>

<details>

<summary> new </summary>

```sql
Sort  (cost=12589.40..12589.41 rows=5 width=144) (actual time=3.022..3.023 rows=1 loops=1)
  Sort Key: c.path
  Sort Method: quicksort  Memory: 25kB
  InitPlan 2
    ->  Index Scan using categories_slug_key on categories  (cost=0.15..8.17 rows=1 width=32) (actual time=0.003..0.003 rows=1 loops=1)
          Index Cond: ((slug)::text = 'science::computational-chemistry'::text)
  ->  Bitmap Heap Scan on categories c  (cost=4.18..12581.18 rows=5 width=144) (actual time=3.021..3.021 rows=1 loops=1)
        Recheck Cond: (path @> (InitPlan 2).col1)
        Filter: ((slug)::text <> 'science::computational-chemistry'::text)
        Rows Removed by Filter: 1
        Heap Blocks: exact=2
        ->  Bitmap Index Scan on path_gist_categories_idx  (cost=0.00..4.18 rows=5 width=0) (actual time=0.011..0.011 rows=2 loops=1)
              Index Cond: (path @> (InitPlan 2).col1)
        SubPlan 1
          ->  Aggregate  (cost=2513.69..2513.70 rows=1 width=4) (actual time=3.004..3.004 rows=1 loops=1)
                ->  Sort  (cost=2503.69..2508.69 rows=2000 width=4) (actual time=2.692..2.827 rows=6593 loops=1)
                      Sort Key: cc.crate_id
                      Sort Method: quicksort  Memory: 193kB
                      ->  Nested Loop Left Join  (cost=15.76..2394.03 rows=2000 width=4) (actual time=0.010..2.136 rows=6593 loops=1)
                            ->  Bitmap Heap Scan on categories c2  (cost=4.18..12.67 rows=5 width=4) (actual time=0.007..0.008 rows=16 loops=1)
                                  Recheck Cond: (path <@ subltree(c.path, 0, 2))
                                  Heap Blocks: exact=3
                                  ->  Bitmap Index Scan on path_gist_categories_idx  (cost=0.00..4.18 rows=5 width=0) (actual time=0.005..0.005 rows=16 loops=1)
                                        Index Cond: (path <@ subltree(c.path, 0, 2))
                            ->  Bitmap Heap Scan on crates_categories cc  (cost=11.58..466.87 rows=940 width=8) (actual time=0.017..0.115 rows=412 loops=16)
                                  Recheck Cond: (category_id = c2.id)
                                  Heap Blocks: exact=1537
                                  ->  Bitmap Index Scan on index_crates_categories_category_id  (cost=0.00..11.34 rows=940 width=0) (actual time=0.011..0.011 rows=412 loops=16)
                                        Index Cond: (category_id = c2.id)
Planning Time: 0.106 ms
Execution Time: 3.046 ms
```

</details>


 If this impact is also acceptable on the production server, maybe we could consider calculating this within the trigger function to update `crates_cnt`. However, a downside is that it could add a tiny bit of overhead to the publishing process.

Turbo87:
> it seems to have only a small impact

0.054 ms vs 3.046 ms seems like a pretty big impact to me 😅 

it might be small in absolute terms, but I assume that on the production server this could end up having a bigger impact 🤔

eth3lbert:
> > it seems to have only a small impact
> 
> 0.054 ms vs 3.046 ms seems like a pretty big impact to me 😅
> 
> it might be small in absolute terms, but I assume that on the production server this could end up having a bigger impact 🤔

Fair enough! That’s why I brought up whether the impact on the production server is acceptable. If so, we could change it to a trigger so it only runs once during the publish stage.

Aakash-Naidu:
> > Since the updated logic introduces a join with crates_categories, an index was added to maintain query performance
> 
> did you measure the performance impact of this PR on all of the affected queries with a production dataset? I'd be somewhat surprised if this didn't impact performance at all 😅
> 
> @eth3lbert any thoughts on this PR?

I don’t have access to a production-sized dataset to benchmark this directly. I validated the logic locally and added a B-Tree index on crates_categories(category_id) to ensure the join can use an index scan rather than a sequential scan.

I’m happy to adjust the approach if there’s a preferred alternative for minimizing runtime impact.

Turbo87:
> I don’t have access to a production-sized dataset to benchmark this directly

see https://crates.io/data-access#database-dumps

PR #13061, opened 2026-02-27
Improve GitHub OAuth error messages

### LINKED ISSUE #932: Show more informative error message if crate authors are not fully known
So today I tried publishing the app_units crate.

I got the error `error: api errors: crate name has already been claimed by another user`.

Except I've published this crate before.

I poked deeper and the crate owners are glennw and `servo:cargo-publish`. I _am_ a member of the cargo-publish org.

It turns out that Servo had tightened up org security, and one thing that was done was disallow access to all OAuth applications by default, with opt-in. So crates.io was unable to figure out who `servo:cargo-publish` consists of, and instead of giving a helpful error, it just told me I wasn't an owner.

It should give an error like "If you are a member of the cargo-publish group on the servo github org, please get servo to approve crates.io by following these steps" instead, I suspect GitHub does return a particular error code when an OAuth app is unable to access an org.

(crates.io issue because I'm sure cargo doesn't have enough info to compute this)

Aakash-Naidu:
## Overview
This PR improves the error message shown to users when crate publishing fails due to GitHub organization OAuth restrictions. Instead of a generic "permission denied" error, users now receive a clear, actionable message identifying the specific organization that needs attention.

## Problem
When a GitHub organization has OAuth restrictions enabled, `crates.io` cannot verify team membership unless an admin approves the `crates.io` OAuth application. This previously resulted in a vague error message that left users confused.

## Changes
- **Client**: Added `GitHubError::AccessDenied { org_name }` variant to handle 403 Forbidden responses.
- **Authorization**: Updated `Rights::get` to catch permission errors and extract the organization name from the team login.
- **Error Handling**: Added a specific error translator to provide the informative message:
  > "GitHub organization '{org_name}' has restricted OAuth access. To publish, a '{org_name}' administrator must approve the 'crates.io' application in the organization's 'Third-party access' settings."

## Verification
- Added a new integration test [src/tests/krate/publish/github_org_restrictions.rs](cci:7://file:///home/mcw/Aakash/crates.io/src/tests/krate/publish/github_org_restrictions.rs:0:0-0:0) that mocks a 403 Forbidden error from GitHub.
- Verified that the test passes and the correct error message is returned.
- Ran `cargo fmt` and `cargo clippy` to ensure code quality.

Resolves #932

### DIFF
--- a/crates/crates_io_github/src/lib.rs
+++ b/crates/crates_io_github/src/lib.rs
@@ -169,7 +169,9 @@ impl GitHubClient for RealGitHubClient {
 #[derive(Debug, thiserror::Error)]
 pub enum GitHubError {
     #[error(transparent)]
-    Permission(anyhow::Error),
+    Unauthorized(anyhow::Error),
+    #[error(transparent)]
+    Forbidden(anyhow::Error),
     #[error(transparent)]
     NotFound(anyhow::Error),
     #[error(transparent)]
@@ -181,7 +183,8 @@ impl From<reqwest::Error> for GitHubError {
         use reqwest::StatusCode as Status;
 
         match error.status() {
-            Some(Status::UNAUTHORIZED) | Some(Status::FORBIDDEN) => Self::Permission(error.into()),
+            Some(Status::UNAUTHORIZED) => Self::Unauthorized(error.into()),
+            Some(Status::FORBIDDEN) => Self::Forbidden(error.into()),
             Some(Status::NOT_FOUND) => Self::NotFound(error.into()),
             _ => Self::Other(error.into()),
         }
--- a/src/controllers/helpers/authorization.rs
+++ b/src/controllers/helpers/authorization.rs
@@ -50,7 +50,7 @@ impl Rights {
                         .await
                     {
                         Ok(membership) => membership.is_some_and(|m| m.is_active()),
-                        Err(GitHubError::Permission(_)) => {
+                        Err(GitHubError::Forbidden(_)) => {
                             let org_name = team
                                 .split_login()
                                 .map(|(_, org, _)| org.to_string())
@@ -60,7 +60,7 @@ impl Rights {
                                 StatusCode::FORBIDDEN,
                                 format!(
                                     "GitHub organization '{org_name}' has restricted OAuth access. \
-                                     To publish, a '{org_name}' administrator must approve the 'crates.io' \
+                                     A '{org_name}' administrator must approve the 'crates.io' \
                                      application in the organization's 'Third-party access' settings."
                                 ),
                             ));
--- a/src/tests/krate/publish/github_org_restrictions.rs
+++ b/src/tests/krate/publish/github_org_restrictions.rs
@@ -23,7 +23,7 @@ async fn publish_with_org_restrictions() {
     github
         .expect_team_membership()
         .returning(|_org_id, _team_id, _username, _token| {
-            Err(GitHubError::Permission(anyhow::anyhow!("403 Forbidden")))
+            Err(GitHubError::Forbidden(anyhow::anyhow!("403 Forbidden")))
         });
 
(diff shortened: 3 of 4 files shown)


crates/crates_io_github/src/lib.rs
```diff
@@ -168,6 +168,8 @@ impl GitHubClient for RealGitHubClient {
 
 #[derive(Debug, thiserror::Error)]
 pub enum GitHubError {
+    #[error("GitHub organization '{org_name}' has restricted OAuth access")]
+    AccessDenied { org_name: String },
```
Turbo87: the error variant is not used anywhere in `crates_io_github`, so it shouldn't live here

Aakash-Naidu: Yes, I used an LLM while working on this PR, primarily to help reason about the issue and explore all the possible implementations. All changes were reviewed, Implemented and tested manually.

You're correct that this variant isn't used within `crates_io_github` I'll remove it and ensure the error is defined in appropriate layer.

src/controllers/helpers/authorization.rs
```diff
@@ -43,10 +45,28 @@ impl Rights {
...
+                                .split_login()
+                                .map(|(_, org, _)| org.to_string())
+                                .unwrap_or_else(|| "unknown".to_string());
+
+                            return Err(custom(
+                                StatusCode::FORBIDDEN,
+                                format!(
+                                    "GitHub organization '{org_name}' has restricted OAuth access. \
+                                     To publish, a '{org_name}' administrator must approve the 'crates.io' \
+                                     application in the organization's 'Third-party access' settings."
```
Turbo87, line 64: this error message seems a bit misleading since `Rights::get()` is not only called by the publish endpoint, but by other operations too

src/controllers/helpers/authorization.rs
```diff
@@ -43,10 +45,28 @@ impl Rights {
...
                     // private membership information here.
-                    let is_team_member = gh_client
+                    let is_team_member = match gh_client
                         .team_membership(team.org_id, team.github_id, &user.gh_login, &token)
-                        .await?
-                        .is_some_and(|m| m.is_active());
+                        .await
+                    {
+                        Ok(membership) => membership.is_some_and(|m| m.is_active()),
+                        Err(GitHubError::Permission(_)) => {
```
Turbo87: I'm not sure if `GitHubError::Permission` is a sufficient enough filter for this. it currently captures both 401 and 403 responses, but I guess the OAuth org issue would only be relevant for 403 responses, while 401 could happen when a user has revoked OAuth access of the crates.io app to their whole account, making their access token invalid.

I don't know what the best way forward here would be. Maybe splitting up `GitHubError::Permission` into two variants? Maybe exposing the `reqwest` error underneath? 🤔

src/controllers/helpers/authorization.rs
Aakash-Naidu: Yes, you're right that Rights::get() is used by multiple operations, so the message mentioning publishing specifically could be misleading. I can update the wording to make it more generic.

src/controllers/helpers/authorization.rs
Aakash-Naidu: yes, GitHubError::Permission capturing both 401 and 403 responses. I've updated the implementation to split it into Unauthorized and Forbidden variants based on the underlying reqwest::StatusCode.

This allows the OAuth organization restriction case to be handled specifically for 403 Forbidden, while 401 Unauthorized now returns a message indicating that the user's GitHub session may have expired or the token was revoked.

Turbo87 (approved):
alright, let's give this a try 👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13061/e154597d3e9a2b835726ac855c349a60ddb22436..e113ad562117e24ff946d250591bb6f57d2db7a5)*

PR #13063, opened 2026-02-27
svelte: Implement API token settings pages

Turbo87:
This PR implements the API token listing and creation pages for the Svelte app, based on the corresponding implementation in the Ember.js app. The behavior and styling should be identical to the Ember.js app. With this PR `pnpm e2e e2e/acceptance/api-token.spec.ts` now also passes when run against the Svelte app. 

### Related


Turbo87:
yeah, it failed because it didn't run the playwright test suite. I screwed up when I rewrote the job filtering and haven't gotten around to fixing it yet 😅

PR #13064, opened 2026-02-27
Document publishing rate limits

carols10cents:
Link to this doc from the footer and the rate-limited error message.


svelte/src/routes/rate-limits/+page.svelte
```diff
@@ -0,0 +1,42 @@
...
+      For new versions of existing crates, allow a burst of 30 new versions published at once by a single user account,
+      with a rate limit of 1 crate per minute allowed after that burst
+    </li>
+  </ul>
+
+  <p>The rate limiting error message will give the time at which you may send the next publish.</p>
+
+  <p>
+    We are unlikely to grant exceptions to these rate limits, to preserve the performance of the system and quantity of
+    crates published for everyone's benefit.
```
Turbo87: while I agree with the unlikelyness, I wonder if we should mention that we do so "only in exceptional cases" or something like that and link to the support page?

app/router.js
```diff
@@ -62,6 +62,7 @@ Router.map(function () {
   // of the URL to be `/security`.
   this.route('security');
   this.route('data-access');
+  this.route('rate-limits');
```
Turbo87: I added `/docs/trusted-publishing` below as a `/docs` subpage to move away from all these top-level documentation pages. What do you think about making this `/docs/rate-limits` or `/policies/rate-limits`?

carols10cents:
@Turbo87 done and done!

Turbo87 (approved):
I had to fix a couple of things, but this is good to go now 😉 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13064/7313b99f813f1b58f32325dc922cc9f3ec0bb105..736dded02e5b84cb8839b259f2cfbb893eb904fa)*

PR #13093, opened 2026-03-04
svelte: Implement `/me/following` page

Turbo87:
As usual, this implements the same behavior and styling as the corresponding page in the Ember.js app :)

### Related

### DIFF
--- a/svelte/src/routes/me/following/+page.svelte
+++ b/svelte/src/routes/me/following/+page.svelte
@@ -1,2 +1,45 @@
-<h1>Following</h1>
-<p>Stub route for /me/following</p>
+<script lang="ts">
+  import CrateList from '$lib/components/CrateList.svelte';
+  import PageHeader from '$lib/components/PageHeader.svelte';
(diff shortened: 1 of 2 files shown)


svelte/src/routes/me/following/+page.svelte
```diff
@@ -1,2 +1,45 @@
...
+
+  let { data } = $props();
+
+  let pagination = $derived(calculatePagination(data.page, data.perPage, data.cratesResponse.meta.total));
+
+  let currentSortBy = $derived(data.sort === 'downloads' ? 'Downloads' : 'Alphabetical');
+</script>
+
+<PageHeader title="Followed Crates" />
+
```
eth3lbert, line 17: I noticed a comment in the Ember.js version. Has it been resolved already, or should we keep it?

Turbo87: good question. I'm not sure if there is much more duplication we can remove from the route without completely removing the route :D

PR #13103, opened 2026-03-05
svelte: Implement dashboard page

Turbo87:
This PR first ports the `CrateDownloadsList` component from Ember.js to Svelte (including a Storybook story), and then it ports the data loading and user interface of the dashboard page. Behavior and styling should match the Ember.js page.

This can be manually tested by running the `dev:msw` script in the Svelte package.

### Related

- https://github.com/rust-lang/crates.io/pull/13099

### DIFF
--- a/svelte/src/routes/dashboard/+page.svelte
+++ b/svelte/src/routes/dashboard/+page.svelte
@@ -1,2 +1,250 @@
-<h1>Dashboard</h1>
-<p>Stub route for /dashboard</p>
+<script lang="ts">
+  import type { components } from '@crates-io/api-client';
+
+  import { resolve } from '$app/paths';
+  import { formatDistanceToNow } from 'date-fns';
+
+  import DownloadIcon from '$lib/assets/download.svg?component';
+  import FollowingIcon from '$lib/assets/following.svg?component';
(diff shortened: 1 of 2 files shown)


svelte/src/routes/dashboard/+page.svelte
```diff
@@ -1,2 +1,250 @@
...
+    display: flex;
+    align-items: center;
+  }
+
+  .page-header-content :global(svg.header-icon) {
+    flex-shrink: 0;
+    margin-right: var(--space-2xs);
+    width: 32px;
+    height: 32px;
+  }
```
eth3lbert: Not a blocker, just curious: why is this one flattened while the others are nested?

Turbo87: no strong reason. in [the original `dashboard.css` file](https://github.com/rust-lang/crates.io/blob/5e5133fa7e4e80ef314cf2a51a8ae0a7118c0c6f/app/templates/dashboard.css#L1-L11) it also wasn't nested, but that was because `.header-icon` was a top-level class, which unfortunately doesn't quite work in Svelte 😅

PR #13117, opened 2026-03-06
msw: Add `GET /api/v1/crates/:name/owners` handler

Turbo87:
The Svelte layout loads owners via the combined `/owners` endpoint, but MSW only had handlers for the `owner_user` and `owner_team` endpoints. Without this handler, the request fell through to the generic version handler which interpreted `owners` as a version number 😅

### DIFF
--- a/packages/crates-io-msw/handlers/crates.js
+++ b/packages/crates-io-msw/handlers/crates.js
@@ -4,6 +4,7 @@ import downloads from './crates/downloads.js';
 import followCrate from './crates/follow.js';
 import following from './crates/following.js';
 import getCrate from './crates/get.js';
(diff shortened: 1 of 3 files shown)


eth3lbert (approved):
Nice catch! are you finding this by chance?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13117/73883093134ac9d5b109b0a3fb34df5e1efd8655..29c5a016748910de59963191546af4af2f00ec6f)*

Turbo87:
more or less. I was wondering why the request wasn't working as expected and after a bit of debugging found that bug 😅

PR #13119, opened 2026-03-06
svelte: Use composite key in `OwnersList` `{#each}` block

Turbo87:
Use `${owner.kind}:${owner.id}` instead of `owner.id` as the key, since users and teams can share the same numeric ID, which would cause a conflict and throw an error.

### Related


eth3lbert (approved):
Great catch, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13119/73883093134ac9d5b109b0a3fb34df5e1efd8655..cff2b5e2d07d472d7c96633cf8a4f349f7cf5c30)*

Issue #1584, opened 2018-12-27
It is possible for more than one user record to have the same gh_login


sgrif:
Right now we do not have a unique index on gh_login. Multiple users can have the same value for that column if the following sequence of events occurs:

- user A renames their account
- user B creates an account with the same name as A
- user B logs in to crates.io
- user A does not log back in to crates.io

We currently have 23 duplicate gh_login records across 72 user records. We do have cases where both users are owners of separate crates. We will need to manually figure out on a case by case basis for the existing records. We should prevent this from occurring in the future by checking to see if an existing user has the same gh_login, and changing that record if it exists.

The question is what do we change it to? We currently use gh_login as the identifier for the /users route, so we can't just blank it out and call it a day. We need some way to continue to link to that user.

carols10cents:
This is intentional, and why we have gh_id

carols10cents:
See https://github.com/rust-lang/crates.io/pull/1109 and https://groups.google.com/forum/#!topic/rustlang-security-announcements/ALbMvU-q9Qs

sgrif:
To be clear, I think any solution to this requires marking that the user is no longer actually the owner of that gh_login somehow (e.g. the implementation of #1586), which makes it unaffected by #1109. The index we'd add would be `UNIQUE ON gh_login WHERE NOT gh_deleted`, since we can have infinitely many deleted accounts for a single username (we have one username that has apparently deleted and recreated its github account 26 times).

But we do need *some* way to continue to identify that user after they've deleted their GH account. In #1586 I've proposed using `deleted_{id}` as it seems like the simplest solution, but we could also do something like `deleted_{gh_login}_{n}` where `n` is `SELECT ROW_NUMBER() OVER (ORDER BY id) FROM users WHERE gh_login = $1` if we want to include the old login or avoid coupling to the ID.

This issue is basically just an extension to #1585, and if #1586 is merged, the solution to this issue is just running `UPDATE users SET gh_deleted = TRUE WHERE gh_login = $1 AND gh_id != $2` whenever a user is created or changes its `gh_login` (this can be done in `NewUser::create_or_update` or as a trigger on `users`)

ssokolow:
[...]

PR #13143, opened 2026-03-10
Check GitHub API proactively for renamed or deleted users

carols10cents:
Also to hopefully get the database to a state where `users.gh_login` is unique.

# Rates

The part that I haven't decided/completed yet is how often to run the `UpdateFromGithub` job and at what pace. Currently, the job must be enqueued completely manually and it does a batch of 100 crates.io users then stops at a pace under the GitHub API rate limits.

We definitely need to run the job against all users once before thinking about implementing https://github.com/rust-lang/rfcs/pull/3946, so that 1. we start all accounts off with their current GitHub username as their crates.io username and 2. the crates.io username based on the GitHub username can be unique.

What I'm imagining is that once we decide to implement the part of https://github.com/rust-lang/rfcs/pull/3946 that stops updating your crates.io username in sync with your GitHub username, then the `UpdateFromGithub` job will only update `oauth_github.login` and not `users.gh_login` (or whatever that column ends up getting renamed to when it becomes the crates.io username).

Reasons to be cautious about how often and how fast this job runs:

- I don't know how the queries and updates to our database will affect database load in production. I hope I've written this in such a way as to not be noticeable, but I could be wrong! If i'm wrong, are there better ways to write the queries? Or will we need to schedule this task to run during times that our traffic is lower?
- We haven't been doing these proactive updates, so there isn't a huge reason to rush into doing them as fast as possible.

Reasons to turn this on to run all the time as fast as possible:

- To get the best mitigation of such long-standing issues as:
  - https://github.com/rust-lang/crates.io/issues/1585
  - https://github.com/rust-lang/crates.io/issues/1584
- We're confident it doesn't affect performance or cause issues (once we've established that confidence however we need to)

And of course there are places between "as slow as possible" and "as fast as possible" that we could land; I'm interested to get peoples' thoughts on the tradeoffs.

# User IDs

@LawnGnome You mentioned the other day something about the crates.io user account ID being PII that we should be cautious about using, mostly because archives exist. I'm interested in your thoughts about using `ghost_{user id}` as the crates.io username for deleted GitHub accounts. I think this is fine, and here's why:

- Our API responses to `https://crates.io/api/v1/users/{username}` include the user's crates.io ID; it's not secret information
- I think the information that could be gleaned from using the user ID  in conjunction with an archive is that crates.io says there's an account `ghost_123` that owns crate `foo`; the archive can tell us through user ID association that this user previously had the username `LawnGnome`, and perhaps this person deleted their GitHub account because they didn't want people to know they published crate `foo`. I don't think this is actually a reason not to use user IDs in this way because:
  - The archive may go back to a point in time where you could see the owner of `foo` was `LawnGnome` anyway
  - Users can request that crates.io delete all data we have associated with them, including their user record and sometimes including their crate, which we're likely to grant if the reason is to remove PII the user doesn't want to be exposed
  
Am I missing a way the user ID could be used that would make `ghost_{user id}` a bad idea?

# Other

Any other concerns I haven't addressed?

### DIFF
--- a/crates/crates_io_database/src/schema.rs
+++ b/crates/crates_io_database/src/schema.rs
@@ -676,6 +676,8 @@ diesel::table! {
         avatar -> Nullable<Varchar>,
         /// Encrypted GitHub access token
         encrypted_token -> Bytea,
+        /// The last time we verified with GitHub what the GitHub username for this user was, and whether the account was valid
+        last_sync -> Timestamptz,
         /// GitHub username
         login -> Varchar,
         /// Crates.io user ID foreign key
--- a/crates/crates_io_github/src/lib.rs
+++ b/crates/crates_io_github/src/lib.rs
@@ -26,6 +26,7 @@ type Result<T> = std::result::Result<T, GitHubError>;
 pub trait GitHubClient: Send + Sync {
     async fn current_user(&self, auth: &AccessToken) -> Result<GitHubUser>;
     async fn get_user(&self, name: &str, auth: &AccessToken) -> Result<GitHubUser>;
+    async fn get_user_by_id(&self, account_id: i64) -> Result<GitHubUser>;
     async fn org_by_name(&self, org_name: &str, auth: &AccessToken) -> Result<GitHubOrganization>;
     async fn team_by_name(
         &self,
@@ -211,6 +212,11 @@ impl GitHubClient for RealGitHubClient {
     async fn get_user(&self, name: &str, auth: &AccessToken) -> Result<GitHubUser> {
         let url = format!("/users/{name}");
         self.request(&url, auth).await
+    }
+
+    async fn get_user_by_id(&self, account_id: i64) -> Result<GitHubUser> {
+        let url = format!("/user/{account_id}");
+        self._request(&url, std::convert::identity).await
     }
 
     async fn org_by_name(&self, org_name: &str, auth: &AccessToken) -> Result<GitHubOrganization> {
--- a/migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
+++ b/migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
@@ -1,0 +1,8 @@
+ALTER TABLE IF EXISTS oauth_github
+-- Set existing rows' last sync value to 1970
+ADD COLUMN last_sync timestamptz NOT NULL DEFAULT to_timestamp(0),
+-- Set new rows' last sync value to the current time going forward, because creating a new
+-- crates.io account fetches the user info from GitHub
+ALTER COLUMN last_sync SET DEFAULT now();
+
+comment on column oauth_github.last_sync is 'The last time we verified with GitHub what the GitHub username for this user was, and whether the account was valid';
--- a/migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
+++ b/migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
@@ -1,8 +1,8 @@
+-- safety-assured:start
+-- Adding a column with a constant value default is safe because we're using Postgres > 11.
 ALTER TABLE IF EXISTS oauth_github
--- Set existing rows' last sync value to 1970
-ADD COLUMN last_sync timestamptz NOT NULL DEFAULT to_timestamp(0),
--- Set new rows' last sync value to the current time going forward, because creating a new
--- crates.io account fetches the user info from GitHub
-ALTER COLUMN last_sync SET DEFAULT now();
+-- Set existing rows' last sync value to 1970.
+ADD COLUMN last_sync timestamptz NOT NULL DEFAULT to_timestamp(0);
+-- safety-assured:end
 
 comment on column oauth_github.last_sync is 'The last time we verified with GitHub what the GitHub username for this user was, and whether the account was valid';
--- a/migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
+++ b/migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
@@ -1,8 +1,5 @@
--- safety-assured:start
--- Adding a column with a constant value default is safe because we're using Postgres > 11.
-ALTER TABLE IF EXISTS oauth_github
--- Set existing rows' last sync value to 1970.
-ADD COLUMN last_sync timestamptz NOT NULL DEFAULT to_timestamp(0);
--- safety-assured:end
+ALTER TABLE oauth_github
+    -- Set existing rows' last sync value to 1970.
+    ADD COLUMN last_sync timestamptz NOT NULL DEFAULT 'epoch';
 
 comment on column oauth_github.last_sync is 'The last time we verified with GitHub what the GitHub username for this user was, and whether the account was valid';
--- a/src/bin/crates-admin/enqueue_job.rs
+++ b/src/bin/crates-admin/enqueue_job.rs
@@ -190,23 +190,16 @@ pub async fn run(command: Command) -> Result<()> {
                 .await?;
 
             for oauth_github in oldest_oauth_github_records {
-                let user_id = oauth_github.user_id;
-                let github_id = oauth_github.account_id;
-                let old_username = oauth_github.login.clone();
-
                 let job = jobs::UpdateUserFromGithub {
                     dry_run,
                     account_id: oauth_github.account_id,
                 };
 
-                // Don't stop the whole batch if one user update errors, but do log the error
+                // Don't stop the whole batch if one enqueue errors, but do log the error
                 if let Err(e) = job.enqueue(&conn).await {
                     error!(
-                        error = %e,
-                        user_id,
-                        github_id,
-                        old_username,
-                        "Error running UpdateUserFromGithub"
+                        "Error enqueueing UpdateUserFromGithub for user_id {}, github_id {}, old username `{}`: {e}",
+                        oauth_github.user_id, oauth_github.account_id, oauth_github.login,
                     );
                 }
             }
--- a/src/bin/crates-admin/enqueue_job.rs
+++ b/src/bin/crates-admin/enqueue_job.rs
@@ -1,7 +1,8 @@
 use anyhow::Result;
 use chrono::NaiveDate;
 use crates_io::db;
-use crates_io::schema::{background_jobs, crates};
+use crates_io::models::OauthGithub;
+use crates_io::schema::{background_jobs, crates, oauth_github};
 use crates_io::worker::jobs;
 use crates_io_worker::BackgroundJob;
 use diesel::dsl::exists;
@@ -56,6 +57,14 @@ pub enum Command {
     SyncUpdatesFeed,
     TrustpubCleanup,
     UpdateDownloads,
+    /// Sync the oldest batch of users with GitHub
+    UpdateUserBatch {
+        #[arg(long = "dry-run")]
+        dry_run: bool,
+
+        #[arg(long = "batch-size", default_value = "100")]
+        batch_size: usize,
+    },
 }
 
 pub async fn run(command: Command) -> Result<()> {
@@ -167,6 +176,36 @@ pub async fn run(command: Command) -> Result<()> {
                 );
             } else {
                 jobs::UpdateDownloads.enqueue(&conn).await?;
+            }
+        }
+
+        Command::UpdateUserBatch {
+            dry_run,
+            batch_size,
+        } => {
+            let oldest_oauth_github_records = oauth_github::table
+                .order(oauth_github::last_sync.asc())
+                .limit(batch_size as i64)
+                .load::<OauthGithub>(&mut conn)
+                .await?;
+
+            for oauth_github in oldest_oauth_github_records {
+                let user_id = oauth_github.user_id;
+                let github_id = oauth_github.account_id;
+                let old_username = oauth_github.login.clone();
+
+                let job = jobs::UpdateUserFromGithub::new(dry_run, oauth_github);
+
+                // Don't stop the whole batch if one user update errors, but do log the error
+                if let Err(e) = job.enqueue(&conn).await {
+                    error!(
+                        error = %e,
+                        user_id,
+                        github_id,
+                        old_username,
+                        "Error running UpdateUserFromGithub"
+                    );
+                }
             }
         }
     };
--- a/src/bin/crates-admin/enqueue_job.rs
+++ b/src/bin/crates-admin/enqueue_job.rs
@@ -194,7 +194,10 @@ pub async fn run(command: Command) -> Result<()> {
                 let github_id = oauth_github.account_id;
                 let old_username = oauth_github.login.clone();
 
-                let job = jobs::UpdateUserFromGithub::new(dry_run, oauth_github);
+                let job = jobs::UpdateUserFromGithub {
+                    dry_run,
+                    account_id: oauth_github.account_id,
+                };
 
                 // Don't stop the whole batch if one user update errors, but do log the error
                 if let Err(e) = job.enqueue(&conn).await {
--- a/src/worker/jobs/update_user_from_github.rs
+++ b/src/worker/jobs/update_user_from_github.rs
@@ -41,21 +41,20 @@ impl BackgroundJob for UpdateUserFromGithub {
             .await?;
 
         info!(
-            dry_run = self.dry_run,
-            user_id = oauth_github.user_id,
-            github_id = self.account_id,
-            old_username = oauth_github.login,
-            "Starting UpdateUserFromGithub"
+            "Starting UpdateUserFromGithub ({}): user_id {}, github_id {}, old username {}",
+            if self.dry_run { "DRY RUN" } else { "FOR REAL" },
+            oauth_github.user_id,
+            self.account_id,
+            oauth_github.login,
         );
 
         let github_user = self.refresh_user(&ctx, &oauth_github).await?;
 
         if self.dry_run {
             info!(
-                user_id = oauth_github.user_id,
-                old_username = oauth_github.login,
-                new_username = github_user.login,
-                "Dry run UpdateUserFromGithub proposed update"
+                "Dry run UpdateUserFromGithub proposed update for crates.io user {} \
+                from username `{}` to username `{}`",
+                oauth_github.user_id, oauth_github.login, github_user.login,
             );
         } else {
             self.apply_update(&oauth_github, &github_user, &mut conn)
--- a/src/worker/jobs/update_user_from_github.rs
+++ b/src/worker/jobs/update_user_from_github.rs
@@ -1,0 +1,50 @@
+use crate::worker::Environment;
+use crates_io_github::GitHubUser;
+use crates_io_worker::BackgroundJob;
+use diesel_async::AsyncPgConnection;
+use serde::{Deserialize, Serialize};
+use std::sync::Arc;
+
+#[derive(Serialize, Deserialize)]
+pub struct UpdateUserFromGithub {
+    /// Crates.io user ID
+    user_id: i32,
+    /// GitHub ID
+    account_id: i32,
+    /// Encrypted GitHub token
+    encrypted_token: Vec<u8>,
+}
+
+impl BackgroundJob for UpdateUserFromGithub {
+    const JOB_NAME: &'static str = "update_user_from_github";
+    const DEDUPLICATED: bool = true;
+
+    type Context = Arc<Environment>;
+
+    /// For the specified user, query the GitHub API for the user's current information to see if
+    /// their account has been deleted or renamed. Update the `users` and `oauth_github` tables,
+    /// saving the current time in `last_sync` even if the user information hasn't changed.
+    async fn run(&self, ctx: Self::Context) -> anyhow::Result<()> {
+        let mut conn = ctx.deadpool.get().await?;
+
+        let github_user = self.refresh_user(&ctx).await?;
+
+        self.apply_update(&github_user, &mut conn).await;
+
+        Ok(())
+    }
+}
+
+impl UpdateUserFromGithub {
+    /// Given the current environment's context, request information from GitHub using the user's
+    /// API token.
+    async fn refresh_user(&self, _ctx: &Arc<Environment>) -> anyhow::Result<GitHubUser> {
+        todo!();
+    }
+
+    /// Given the information from GitHub about the current user, make the appropriate changes to
+    /// the `users` and `oauth_github` tables.
+    async fn apply_update(&self, _github_user: &GitHubUser, _conn: &mut AsyncPgConnection) {
+        todo!();
+    }
+}
--- a/src/worker/jobs/update_user_from_github.rs
+++ b/src/worker/jobs/update_user_from_github.rs
@@ -16,15 +16,9 @@ use tracing::{error, info};
 pub struct UpdateUserFromGithub {
     /// Dry run will fetch updates from GitHub and log what it would change, but does not actually
     /// update the database.
-    dry_run: bool,
-    /// Crates.io user ID
-    user_id: i32,
+    pub dry_run: bool,
     /// GitHub ID
-    account_id: i64,
-    /// Encrypted GitHub token
-    encrypted_token: Vec<u8>,
-    /// Username currently in the database
-    old_username: String,
+    pub account_id: i64,
 }
 
 impl BackgroundJob for UpdateUserFromGithub {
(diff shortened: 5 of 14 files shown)


src/worker/jobs/update_from_github.rs
```diff
@@ -0,0 +1,598 @@
...
+    // Exclude crates.io `ghost_{id}` users here; if we have a GitHub ID for them, there's a chance
+    // their GitHub account deletion could be undone so we should check in the next arm.
+    } else if user.gh_login.contains('_') && !user.gh_login.starts_with("ghost_") {
+        // We can't get account status for enterprise managed users via GitHub's API, so don't try.
+        Ok(None)
+    } else {
+        // At the time of writing, if authenticated, GitHub's rate limit is 5,000 requests per
+        // minute. 5000 / 60 is about 80 requests a second. Waiting 15 ms between requests should
+        // keep us at about 60 requests a second; well under the rate limit.
+        tokio::time::sleep(std::time::Duration::from_millis(15)).await;
```
Turbo87: unless GitHub has changed things, the limit is 5000 request per **hour**, not per minute. that signficantly changes the calculation and would require a timeout of something more like 700ms.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/aabff361a5611649e360bcb29b0b33b2ee7e1a51..a4b72c9345b5b7b72a5378641f79dd5c35e475d4)*

crates/crates_io_github/src/lib.rs
```diff
@@ -213,6 +219,16 @@ impl GitHubClient for RealGitHubClient {
...
     }
 
+    async fn get_user_by_id(
+        &self,
+        account_id: i64,
+        username: &str,
+        password: &str,
+    ) -> Result<GitHubUser> {
+        let url = format!("/user/{account_id}");
+        self.request_basic(&url, username, password).await
```
Turbo87: why are we using the `client_id/secret` auth for these calls instead of the `AccessToken`? this causes the requests to all use the same rate limit. if we used the per-user `AccessToken` I think they would use a per-user rate limit, which basically means they would not be rate limited at all 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/aabff361a5611649e360bcb29b0b33b2ee7e1a51..a4b72c9345b5b7b72a5378641f79dd5c35e475d4)*

src/worker/jobs/update_from_github.rs
```diff
@@ -0,0 +1,598 @@
...
+    /// GitHub API using the GitHub ID to see if their account has been deleted or renamed. Gather
+    /// all the changes and update the `users` and `oauth_github` tables. Save the last user ID
+    /// checked to know where to start for the next batch.
+    async fn run(&self, ctx: Self::Context) -> anyhow::Result<()> {
+        info!(dry_run = self.dry_run, "Starting UpdateFromGithub");
+
+        let mut conn = ctx.deadpool.get().await?;
+        let user_agent = crates_io_version::user_agent();
+        let client = Client::builder().user_agent(user_agent).build()?;
+        let github = RealGitHubClient::new(client);
```
Turbo87:
```suggestion
        let github = &env.github;
```

the `Environment` contains a `GitHubClient` these days. this then would also allow us to write full tests for this background job, since the `TestApp` by default puts a `mockall` mock into `env.github`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/aabff361a5611649e360bcb29b0b33b2ee7e1a51..a4b72c9345b5b7b72a5378641f79dd5c35e475d4)*

crates/crates_io_database/src/schema.rs
```diff
@@ -664,6 +664,19 @@ diesel::table! {
...
+    use diesel_full_text_search::Tsvector;
+
+    /// Track where we are in refreshing user GitHub info from the GitHub API in batches in jobs
+    metadata_github_refresh (highest_processed_user_id) {
+        /// The number of users to request in the next batch.
+        batch_size -> Int8,
+        /// The highest crates.io user ID that was processed in a previously completed batch. The next run will request a batch of users with IDs greater than this.
+        highest_processed_user_id -> Int4,
+    }
+}
```
Turbo87: can you explain why you chose this design?

my initial naive thought would've been to add a `last_sync` column or something like that to the `oauth_github` table and then do a query like `SELECT gh_id FROM oauth_github ORDER BY last_sync LIMIT 100`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/aabff361a5611649e360bcb29b0b33b2ee7e1a51..a4b72c9345b5b7b72a5378641f79dd5c35e475d4)*

src/worker/jobs/update_from_github.rs
```diff
@@ -0,0 +1,598 @@
...
+
+        let metadata = get_state_params(&mut conn).await?;
+
+        let crates_io_users = next_user_batch(metadata, &mut conn).await?;
```
Turbo87: I wonder what the rationale is for batching these updates. the GitHub request needs to happen per user and the database update as it is implemented right now also happens per user. in that case, why not have a background job that runs for just one user and have the batching happen in the code that enqueues the background jobs? that could also parallelize the updates and wouldn't block the background worker for long stretches of time.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/aabff361a5611649e360bcb29b0b33b2ee7e1a51..a4b72c9345b5b7b72a5378641f79dd5c35e475d4)*

src/worker/jobs/update_from_github.rs
carols10cents: ooooops. this seems moot though if we use the per-user access token as you suggested below!

crates/crates_io_github/src/lib.rs
carols10cents: hmmmmmm i think i was just thinking we'd only need public information so there wasn't a reason to auth as each user. I wonder if using the per-user token would handle the enterprise managed account situation too! I will switch this over and see :)

crates/crates_io_database/src/schema.rs
carols10cents: I think my brain thought this was the simplest possible thing, and would keep the info this background job needed separate from the user info :)

I do like the extra information that a `last_sync` column would give us, but that would mean we'd have to update every row every time we refresh, even if nothing changed. I guess that's fine? Let's try it and see!

src/worker/jobs/update_from_github.rs
carols10cents: 🤷🏻‍♀️ batching seemed better to my brain. your suggestion makes sense though, i'll give it a try!

crates/crates_io_database/src/schema.rs
Turbo87: > but that would mean we'd have to update every row every time we refresh, even if nothing changed. I guess that's fine? Let's try it and see!

yeah, that's definitely a downside of it. I guess it depends a bit on how often we want to run this. alternatively we could pull that column into a dedicated one-to-one relationship table to avoid having to update the rest of the row for every sync column update. my gut feeling says it should be fine as-is, but idk either 😅

src/worker/jobs/update_from_github.rs
Turbo87: oops, sorry, I guess it was `ctx.github` in this job 😅

carols10cents: no worries :)

crates/crates_io_database/src/schema.rs
eth3lbert: I also second Tobi's approach since it should be more manageable when errors occur. Perhaps we could also record the `last_start`, `last_end`, and the `error_count`? This way, we could track how long the checks take and easily identify which ones are completed, in progress, or pending. It might also allow us to potentially skip the ones that failed.

eth3lbert (commented):
I'll try to review this over the weekend! 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/ca8b3f14dbb16a79de62a38bc96931ad623cf429..82d36bb483fb5756ca1345573c0539244c6d30f6)*

crates/crates_io_database/src/schema.rs
carols10cents: Ok, I've updated the PR to use the approach Tobi suggested, but I just have `last_sync` for now (rather than the additional `last_start`, `last_end`, and `error_count`). I'm happy to add those fields if they seem useful to everyone!

crates/crates_io_github/src/lib.rs
carols10cents: Done, but I'm not quite sure what "Unauthorized" and "Forbidden" mean (if they mean something different in these cases). And assuming that one of those means the token we have for that user is invalid, does that mean we need to refresh the token or that the user has revoked crates.io's access? I'm going to have to try some dry runs out, I think.

src/worker/jobs/update_from_github.rs
carols10cents: Ok, for now I've put the fetching of a batch into the admin `enqueue-job` command.... which doesn't feel like where we actually want it, but would let us deploy this and try out some batches by running the command.

carols10cents:
Ok new approach here, leaving the failing migration check for now so that this doesn't get merged (i think there are still a few details to work out) and because I have to go 😅 

But I'd love thoughts!

migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE IF EXISTS oauth_github
+-- Set existing rows' last sync value to 1970
+ADD COLUMN last_sync timestamptz NOT NULL DEFAULT to_timestamp(0),
+-- Set new rows' last sync value to the current time going forward, because creating a new
+-- crates.io account fetches the user info from GitHub
+ALTER COLUMN last_sync SET DEFAULT now();
```
eth3lbert: This won't trigger a table rewrite and should be instantaneous in PostgreSQL v11 or later. I'm totally fine with this approach!

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

src/worker/jobs/update_user_from_github.rs
```diff
@@ -0,0 +1,167 @@
...
+                .decrypt(&self.encrypted_token)?;
+
+            match github.current_user(&token).await {
+                Ok(github_user) => Ok(github_user),
+                // If the user is not found, the account has been deleted. Update to the ghost
+                // username.
+                Err(GitHubError::NotFound(_)) => Ok(self.ghost_user()),
+                // Does unauthorized/forbidden mean the user has been deleted?
+                // Or does it mean the user removed crates.io's authorization?
+                // Or that the token we have for the user is bad?
```
eth3lbert: I'm not sure if converting the user into a ghost account is the best move in these cases, as it makes me a bit uneasy. I think it makes more sense to either force re-authorization or error and retry after reaching a specific error threshold.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

src/worker/jobs/update_user_from_github.rs
```diff
@@ -0,0 +1,167 @@
...
+        // next user rather than stopping-- hopefully we'll get that user updated next time.
+        if let Err(e) = conn
+            .transaction(async |conn| {
+                // This will be removed when we no longer sync crates.io usernames with GitHub.
+                // (The transaction can be removed when this is removed as well)
+                diesel::update(users::table)
+                    .filter(users::id.eq(self.user_id))
+                    .set(users::gh_login.eq(&github_user.login))
+                    .execute(conn)
+                    .await?;
```
eth3lbert: This could potentially be rewritten to exclude empty updates when the `gh_login` remains the same. However, I’m also fine with leaving it as it is, since it might just be removed in the end anyway.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

src/bin/crates-admin/enqueue_job.rs
```diff
@@ -173,6 +182,36 @@ pub async fn run(command: Command) -> Result<()> {
...
+                    error!(
+                        error = %e,
+                        user_id,
+                        github_id,
+                        old_username,
+                        "Error running UpdateUserFromGithub"
+                    );
+                }
+            }
+        }
```
eth3lbert, line 206: This seems to be syncing one by one. I'm wondering if we could make it update multiple values at once instead.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
Turbo87: are you sure about that?! 🤔 

in my understanding the `ADD COLUMN last_sync timestamptz NOT NULL DEFAULT to_timestamp(0)` itself would probably not cause a table rewrite since the result in a constant value and can just be saved in the table metadata as the default value. but if we then `ALTER COLUMN last_sync SET DEFAULT now()` we change the default value to a dynamic value (`now()` is not constant), but I assume we also change the value for existing rows that relied on the default value? or the query would actually have to rewrite all existing rows with the old default value before it switches to the new default?

src/bin/crates-admin/enqueue_job.rs
Turbo87: are you talking about the actual syncing or that background job enqueuing? 

the former needs one GitHub API request per user, so it's kinda hard to batch that in a useful way. we could theoretically batch the database writes, but I'm not sure whether the complexity is worth it. I'd start simple and then we can address performance issues if we actually see them.

the latter should be possible though. we might be able to run a single SQL query that transforms `oauth_github` rows directly into background jobs instead of enqueuing them one by one.

migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
eth3lbert: Yeah, I actually assumed it would need a rewrite before testing for that same reason. However, after testing (using `select pg_relation_filenode('tablename');`) the results showed that the table's file remained unchanged.

If we’re still not 100% sure, we could just go with the safe multi-step backfill method instead.

src/bin/crates-admin/enqueue_job.rs
eth3lbert: Fair enough! I was imagining that we could either enqueue multiple values or fetch them in a background job. We could run several checks asynchronously and then update everything at once. However, you’re right, we can just take the simple route for now and improve it later if we encounter performance issues.

src/worker/jobs/update_user_from_github.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// update the database.
+    dry_run: bool,
+    /// Crates.io user ID
+    user_id: i32,
+    /// GitHub ID
+    account_id: i64,
+    /// Encrypted GitHub token
+    encrypted_token: Vec<u8>,
+    /// Username currently in the database
+    old_username: String,
```
Turbo87: do we need all of those fields in the job payload? it feels like the `account_id` as the primary key plus maybe `dry_run` should be sufficient and then the job can query the token and other stuff from the database?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

src/worker/jobs/update_user_from_github.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// their account has been deleted or renamed. Update the `users` and `oauth_github` tables,
+    /// saving the current time in `last_sync` even if the user information hasn't changed.
+    async fn run(&self, ctx: Self::Context) -> anyhow::Result<()> {
+        info!(
+            dry_run = self.dry_run,
+            user_id = self.user_id,
+            github_id = self.account_id,
+            old_username = self.old_username,
+            "Starting UpdateUserFromGithub"
+        );
```
Turbo87, line 51: I've stopped using structured fields for most log messages, since Datadog isn't showing them by default. you can click the log message row and then they will show up, but in the default view it would just show "Starting UpdateUserFromGithub" without any of the additional information. I'm still using it for fields that I would like to filter on or for `duration` which has implicit meaning in Datadog, but most other stuff I just put in the log message to make it easy to see.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

src/worker/jobs/update_user_from_github.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    /// Given the current environment's context, request information from GitHub using the user's
+    /// API token.
+    async fn refresh_user(&self, ctx: &Arc<Environment>) -> anyhow::Result<GitHubUser> {
+        // if the user's gh_id isn't positive, we don't even need to ask github about this,
+        // we know this user is invalid. Just make sure their username is the ghost username.
+        if self.account_id < 1 {
+            Ok(self.ghost_user())
```
Turbo87: I wonder if we should move this check into the `run()` fn so that we don't even request a database connection from the pool in that case

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

src/worker/jobs/update_user_from_github.rs
Turbo87: yeah, the most likely cause for this error is a revoked authorization. I guess in that case we could also try an unauthenticated request, which shouldn't return such errors. If the unauthenticated request fails due to rate limiting we can fail the background job and then the worker would retry it later.

src/worker/jobs/update_user_from_github.rs
```diff
@@ -0,0 +1,167 @@
...
+    account_id: i64,
+    /// Encrypted GitHub token
+    encrypted_token: Vec<u8>,
+    /// Username currently in the database
+    old_username: String,
+}
+
+impl BackgroundJob for UpdateUserFromGithub {
+    const JOB_NAME: &'static str = "update_user_from_github";
+    const DEDUPLICATED: bool = true;
```
Turbo87, line 26: I wonder if we should set a negative default priority for this job. the consequence of that would be that the on-call team won't get actively paged for this job if it is in the queue for more than 60min. if we ever enqueue the whole `users` table then I assume the full sync would likely take more than an hour. the other advantage is that these jobs are then lower priority compared to the publish-related ones.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

src/worker/jobs/update_user_from_github.rs
```diff
@@ -0,0 +1,167 @@
...
+                // This will be removed when we no longer sync crates.io usernames with GitHub.
+                // (The transaction can be removed when this is removed as well)
+                diesel::update(users::table)
+                    .filter(users::id.eq(self.user_id))
+                    .set(users::gh_login.eq(&github_user.login))
+                    .execute(conn)
+                    .await?;
+
+                diesel::update(oauth_github::table)
+                    .filter(oauth_github::user_id.eq(self.user_id))
```
Turbo87: the primary key is the `account_id`, so we should probably use that for filtering, just in case we want to support N GitHub accounts to one crates.io account in the future 😅 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

src/worker/jobs/update_user_from_github.rs
```diff
@@ -0,0 +1,167 @@
...
+                Ok::<(), diesel::result::Error>(())
+            })
+            .await
+        {
+            // Database update failed; it's ok to not update this user this round.
+            // Better luck next time.
+            error!(
+                "Could not update user ID {} from username {} to username {}: {e}",
+                self.user_id, self.old_username, github_user.login,
+            );
```
Turbo87, line 171: any reason for logging the error instead of failing the background job so that it stays in the queue and will be retried?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/85f2b0eb22a51771ca6889e3e951a35b5f14ba4e..f7cfc39103469d3902d0129bc7d697e874ae5c67)*

migrations/2026-05-05-180007-0000_add_last_sync_to_oauth_github/up.sql
carols10cents: I've marked the `DEFAULT to_timestamp(0)` part as `safety-assured` and I removed the part that changed the default to `now()` -- I think I've convinced myself we don't actually need that because we set the value for `last_sync` to `Utc::now()` in the Rust code that's inserting new rows _anyway_.

src/worker/jobs/update_user_from_github.rs
carols10cents: ok, then in that case, I think we'll need the exclusion of GitHub Enterprised Managed Users again.

crates/crates_io_github/src/lib.rs
jplatte: I don't really have any context on the question, but often 'unauthorized' = "missing / invalid credentials", 'forbidden' = "insufficient permissions".

src/worker/jobs/update_user_from_github.rs
carols10cents: I've just pushed a commit to fall back to an unauthenticated request for the user info by ID in these cases!

crates/crates_io_github/src/lib.rs
carols10cents: This is for the user requesting their own information from GitHub using the oauth token we have saved for them, so we definitely have credentials, but those credentials may have expired or been revoked. i'm not sure what it would mean for a user to have insufficient permissions to view their own info. It's entirely possible github never returns one of these for this API call, but their API docs leave much to be desired.

Anyway, it doesn't much matter now-- I've just pushed a commit that falls back to an anonymous request to get the user's info using their github ID in both the unauthorized and forbidden cases here.

crates/crates_io_database/src/schema.rs
carols10cents: Resolving this conversation-- let's try with only `last_sync` and see if we wish we had more visibility into failure cases.

src/worker/jobs/update_user_from_github.rs
carols10cents: That's a great catch! Who knows how long this will be needed, though! I've added the extra `if` to avoid as much database traffic as possible :)

src/worker/jobs/update_user_from_github.rs
carols10cents: shoot, i just realized that the way i wrote the enqueue job query won't get `user` records that have a github id of < 1, which I do want to get right now to get us to `user.gh_login` uniqueness.... but maybe I should update those manually. i think that means i can get rid of the <1 check in the job then too, because the `oauth_github` table just shouldn't have records for those at all

but yes, i agree this can just have the account_id and do a query for the rest :)

src/worker/jobs/update_user_from_github.rs
carols10cents: yes good catch! i just noticed that too while changing the job to only have account_id 😅 fixed in the commit that made that change!

src/worker/jobs/update_user_from_github.rs
carols10cents: [as noted in another comment](https://github.com/rust-lang/crates.io/pull/13143#discussion_r3237183109), we're not even going to get records like this because `user` records with a `gh_id` < 1 don't even have `oauth_github` records, and`oauth_github` is the only table queried for jobs to enqueue 😅 so i've just removed this check entirely.

src/worker/jobs/update_user_from_github.rs
carols10cents: Done!

src/worker/jobs/update_user_from_github.rs
carols10cents: I was trying to be conservative and only retrying potentially the next time we enqueue a batch with the admin job, rather than retrying a bunch of times via the background job retry, so that if something is completely wrong when we try this for real, we aren't doing a whole bunch of wrong stuff a bunch of times (and not needing to scramble to get the jobs out of the queue).

What do you think?

src/worker/jobs/update_user_from_github.rs
carols10cents: That makes sense, i set the default priority to -15 to be less important than rendering og images, but more important than analyze-crates jobs.

carols10cents:
@Turbo87 @eth3lbert this is ready for rereview!

LawnGnome:
I'll try to do a full review on this next week, but don't block on me if @Turbo87 and/or @eth3lbert approve it; I'm happy to review this ex post facto in my copious amounts of spare time.

Just on your question for me in your post, though:

> Am I missing a way the user ID could be used that would make `ghost_{user id}` a bad idea?

I think your analysis is basically correct, and you've captured my concern: that having the ID _might_ make it easier to work backwards to identify a user who no longer wants to be identified.[^gdpr]

But, as you also said, the API is returning the ID anyway right now, unless we decide to take extra steps to merge deleted users into a singular `ghost` user (which I am explicitly _not_ suggesting here, and think we'd want to RFC if we ever did go down that road), so having it also be in the user name seems reasonable enough for now.

So no, I don't think you're missing a way, and I don't think it's a blocker here.

[^gdpr]: You didn't ask, but to get it off my chest: having just gone through the GDPR erasure process, I think it's a slightly unfortunate combination of well-intentioned, rooted in a real concern, and fundamentally useless on an Internet where public things are archived and mirrored willy-nilly. I think it's reasonable for us to enable steps that will make it _harder_ for a bad person to figure out who a deleted user is, but it's basically always a balancing act, since I can pretty much guarantee there are people archiving things like database dumps and mirroring metadata whether we — or the deleted users — like it or not.

src/worker/jobs/update_user_from_github.rs
Turbo87: okay. I'm fine with it either way. just wanted to make sure it was a conscious decision :)

Turbo87 (approved):
I've rebased this and adjusted a few minor things, but I think this is good to go now. if we notice any issues during the initial dry runs we can still fix them later.

@carols10cents let me know if you are okay with my changes or if you have any feedback :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13143/cc661a3157abac2d9faa510167367f27ec0808bd..e962f4a0124eafe5d41e98474e2fc6080ccfa3e7)*

carols10cents:
@Turbo87 your changes lgtm! I'll let you merge :)

Turbo87:
let's get this party started! 🎉

PR #13173, opened 2026-03-13
security: Prevent overflows without wrapping

miketheman:


miketheman:
Noticed on https://crates.io/crates/sqlx/security

## Before

Screenshot 2026-03-13 at 12 37 05

## After

Screenshot 2026-03-13 at 12 36 56

PR #13174, opened 2026-03-13
simplify docs.rs build-status request to safe a redirect

syphar:
( draft so I can see what fails in CI). 

in docs.rs, `/=1.2.3/` is equivalent to `/1.2.3/`, and we redirect one to the other.


syphar:
@Turbo87 ready for review now. 

I didn't manually test locally, my guess is from the change & the tests that we're fine? 

( happy to dig into how to run it locally if that's needed though)

PR #13201, opened 2026-03-18
Replace and_then(identity) with flatten

Aakash-Naidu:
Replace `and_then(std::convert::identity)` with `.flatten()`.
`flatten()` has been stable for a long time.
Also removes the outdated TODO comment.


LawnGnome:
Thanks!

PR #13219, opened 2026-03-21
Replace `deadpool-diesel` with `deadpool-runtime`

Turbo87:
The only import from `deadpool-diesel` was `deadpool_diesel::Runtime`, which is a re-export of `deadpool_runtime::Runtime`. The `postgres` feature only enabled `diesel/postgres`, which we already depend on directly. The `tracing` feature added spans in `deadpool-sync`'s `SyncWrapper`, which is not part of our async connection pool setup.

Depending on `deadpool-runtime` directly removes `deadpool-diesel` and `deadpool-sync` from the dependency tree.

This makes https://github.com/rust-lang/crates.io/pull/13215 a bit easier to fix too :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1492,7 +1492,7 @@ dependencies = [
  "crates_io_version",
  "crates_io_worker",
  "csv",
- "deadpool-diesel",
+ "deadpool-runtime",
  "derive_more",
  "dialoguer",
  "diesel",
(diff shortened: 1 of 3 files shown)


Turbo87:
oops, this was merged as part of https://github.com/rust-lang/crates.io/pull/13215 😅

PR #13240, opened 2026-03-22
svelte: Implement accept-invite page

Turbo87:
As usual, this should behave and look the same as the Ember.js app, and both apps now pass the corresponding Playwright tests.

### Related

### DIFF
--- a/svelte/src/routes/accept-invite/[token]/+page.svelte
+++ b/svelte/src/routes/accept-invite/[token]/+page.svelte
@@ -1,7 +1,54 @@
 <script lang="ts">
+  import { onMount } from 'svelte';
+  import { resolve } from '$app/paths';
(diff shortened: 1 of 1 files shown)


svelte/src/routes/accept-invite/[token]/+page.svelte
```diff
@@ -1,7 +1,54 @@
...
+      if (response.response.ok) {
+        result = 'success';
+      } else {
+        errorText = (response.error as unknown as { errors?: { detail?: string }[] })?.errors?.[0]?.detail;
+        result = 'error';
+      }
+    } catch {
+      result = 'error';
+    }
+  });
```
Comment (eth3lbert) [inline svelte/src/routes/accept-invite/[token]/+page.svelte:31]: I checked it twice to ensure there is no cleanup function! :D

PR #13259, opened 2026-03-26
svelte: Implement crate security page

Turbo87:
This PR ports the last remaining route from Ember.js to Svelte: the crate security page. As usual, behavior and styling should match the Ember.js implementation and the Playwright test suite passes against both apps now.

/cc @djc 

### Related


djc:
Congrats on completing the route porting!

PR #13274, opened 2026-03-26
svelte: Preload mermaid library before rendering README

Turbo87:
This loads the `mermaid` library after fetching the README HTML but before rendering it, so that the `renderMermaids` attachment can use the library synchronously. This matches the Ember.js implementation where the mermaid library is loaded in the component before the modifier runs.

### DIFF
--- a/svelte/src/lib/components/CrateVersionPage.svelte
+++ b/svelte/src/lib/components/CrateVersionPage.svelte
@@ -7,6 +7,7 @@
 
   import CrateIcon from '$lib/assets/crate.svg?component';
   import DownloadIcon from '$lib/assets/download.svg?component';
+  import { loadMermaid } from '$lib/attachments/mermaid';
   import CrateSidebar from '$lib/components/crate-sidebar/CrateSidebar.svelte';
   import CrateHeader from '$lib/components/CrateHeader.svelte';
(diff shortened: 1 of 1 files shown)


svelte/src/lib/components/CrateVersionPage.svelte
```diff
@@ -58,6 +59,17 @@
   let retryReadmePromise = $state<typeof readmePromise | null>(null);
   let activeReadmePromise = $derived(retryReadmePromise ?? readmePromise);
 
+  // Ensure the mermaid library is loaded before the README is rendered,
+  // so that the `renderMermaids` attachment can use it synchronously.
+  let readmeWithMermaid = $derived(
+    activeReadmePromise.then(async html => {
+      if (html?.includes('language-mermaid')) {
+        await loadMermaid().catch(e => console.error(e));
```
LawnGnome, line 67: Is it worth hoisting this to start loading Mermaid concurrently with the readme promise, or does that not matter in practice?

Turbo87: very few readmes use mermaid diagrams, so it would be somewhat wasteful to load it early if the majority of cases don't actually need the code. this is also roughly how it's currently implemented in the Ember.js app.

PR #13276, opened 2026-03-26
svelte: Handle search errors inline with retry support

Turbo87:
This moves the data fetching from the `load` function into the page component to better match the Ember.js implementation. Errors are displayed inline with a "Try Again" button, previous successful results are preserved while a new search is loading, and in-flight requests are cancelled when a new search is triggered.

### DIFF
--- a/svelte/src/routes/search/+page.svelte
+++ b/svelte/src/routes/search/+page.svelte
@@ -1,6 +1,9 @@
 <script lang="ts">
-  import { onMount } from 'svelte';
+  import type { paths } from '@crates-io/api-client';
+
+  import { onMount, untrack } from 'svelte';
   import { page } from '$app/state';
+  import { createClient } from '@crates-io/api-client';
 
   import Alert from '$lib/components/Alert.svelte';
   import CrateList from '$lib/components/CrateList.svelte';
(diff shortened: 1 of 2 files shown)


svelte/src/routes/search/+page.svelte
```diff
@@ -27,7 +30,67 @@
...
+      lastCompletedHadError = true;
+    } finally {
+      if (!signal.aborted) {
+        loading = false;
+        hasCompleted = true;
+      }
+    }
+  }
+
+  let firstResultPending = $derived(!hasCompleted && loading);
```
eth3lbert, line 79: It seems like this can be derived from 
```typescript
cratesResponse === undefined && loading
```
as well?

Turbo87: that was a good question, and I was inclined to agree, but I double checked with Claude and I think it made a good point:

> If you used `cratesResponse === undefined` instead, here's where it breaks:
> 
> 1. First request errors → `cratesResponse` is still `undefined`, `lastCompletedHadError = true` → error state shown (OK so far)
> 2. User changes query params → new fetch starts, `loading = true`
>    - With `hasCompleted`: `firstResultPending = false` → stays on error state, header spinner shows loading
>    - With `cratesResponse === undefined`: `firstResultPending = true` → switches to "Loading search results..." text, hiding the error

It's not terrible behavior either, but it doesn't quite match the current Ember.js implementation, so I guess we'll stay with the `hasCompleted` approach for now.

PR #13295, opened 2026-03-28
svelte: Add actions menu to version list rows

Turbo87:
This implements the dropdown menu with the yank/unyank and rebuild docs items. The sudo mode is out of scope for now, replaced by a TODO comment, and will be implemented in a follow-up PR.

### Related

### DIFF
--- a/svelte/src/lib/components/version-list/Row.svelte
+++ b/svelte/src/lib/components/version-list/Row.svelte
@@ -10,6 +10,7 @@
   import CalendarIcon from '$lib/assets/calendar.svg?component';
   import CheckboxEmptyIcon from '$lib/assets/checkbox-empty.svg?component';
   import CheckboxIcon from '$lib/assets/checkbox.svg?component';
+  import EllipsisCircleIcon from '$lib/assets/ellipsis-circle.svg?component';
   import GitHubIcon from '$lib/assets/github.svg?component';
   import GitLabIcon from '$lib/assets/gitlab.svg?component';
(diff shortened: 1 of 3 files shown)


svelte/src/lib/components/YankButton.svelte
```diff
@@ -0,0 +1,58 @@
...
+        let { response } = await client.DELETE('/api/v1/crates/{name}/{version}/yank', { params });
+        if (!response.ok) return;
+        version.yanked = true;
+      }
+
+      await onChanged?.();
+    } finally {
+      isLoading = false;
+    }
+  }
```
eth3lbert, line 39: I haven't finished reviewing yet, but it seems like the part that refreshes `releaseTrack` after a yank/unyank toggle is missing.

Turbo87: that should be handled by the `onChanged()` which triggers an `invalidate()` call, which will reload the versions data including the release tracks metadata

it's not perfect due to the pagination, but I'm not sure if the current Ember.js code handles that edge case perfectly either 😅

eth3lbert: Ah, indeed! That should also work and it makes sense to me, although I didn't expect that refactor to land during the migration stage.

Turbo87: yeah, the migration to the load fns makes it a bit more tricky in some cases. but we're close to switching over and then we can clean this all up a bit more since we're no longer constrained by making the playwright test suite work for both apps :)

PR #13333, opened 2026-04-03
Reject crate tarball uploads containing device files

joshtriplett:
We shouldn't allow these in new crates. Devices won't work unless unpacked with privileges *and* running on a system that has the device. A fifo won't work unless something is listening on them, and test suites wanting to test those should be creating them.

Add tests for each of these cases.


Turbo87 (approved):
CI was complaining about using `assert_snapshot!()` inside a "loop", but I've added `insta::allow_duplicates` around it which fixed the problem.

Everything else looks good. Thank you! :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13333/d3389986304fac01987fb6f4a5a40644acd0a0c7..0961b1e6c9989f548a37be1549183243341299dd)*

PR #13334, opened 2026-04-03
svelte: Implement sudo mode

Turbo87:
This adds sudo state management to the `SessionState` with localStorage persistence and auto-expiry via `setTimeout`. It wires the `Header` and `PrivilegedAction` components to the session so admin users can enable/disable elevated privileges matching the Ember app behavior. This ensures that the corresponding Playwright tests for the sudo behavior now also pass against the Svelte app.

### Related

### DIFF
--- a/svelte/src/lib/utils/session.svelte.ts
+++ b/svelte/src/lib/utils/session.svelte.ts
@@ -121,12 +121,8 @@ export class SessionState {
     if (user?.is_admin) {
       let stored = localStorage.getItem(SUDO_KEY);
       if (stored) {
-        let remaining = Number(stored) - Date.now();
-        if (remaining > 0) {
-          this.setSudo(remaining);
-        } else {
-          localStorage.removeItem(SUDO_KEY);
-        }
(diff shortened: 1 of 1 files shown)


svelte/src/lib/utils/session.svelte.ts
```diff
@@ -111,6 +117,47 @@ export class SessionState {
...
+
+    if (user?.is_admin) {
+      let stored = localStorage.getItem(SUDO_KEY);
+      if (stored) {
+        let remaining = Number(stored) - Date.now();
+        if (remaining > 0) {
+          this.setSudo(remaining);
+        } else {
+          localStorage.removeItem(SUDO_KEY);
+        }
```
eth3lbert: Can this be further reduced to the following?
```js
let remaining = max(0, Number(stored) - Date.now());
this.setSudo(remaining);
```
This would ensure a cleanup at the end. Since `disableSudo()` also just calls `setSudo(0)`, I think this makes sense.

PR #13341, opened 2026-04-03
app: enable HTML for banner messages

LawnGnome:
Since banner messages can only be set by crates.io admins, enabling HTML should be safe enough.

This adds a new `htmlContent` option to Svelte notification messages which gives the same behaviour as the `ember-cli-notifications` option of the same name. This is only used when adding banner messages; other notifications are unchanged, as shown in the storybook update.

---

Obviously the motivation here is [this Zulip message](https://rust-lang.zulipchat.com/#narrow/channel/356853-t-docs-rs/topic/only.20build.20default.20target.20by.20default.3F/near/583384440), since I'd like to link to the relevant blog post directly.

### DIFF
--- a/svelte/src/lib/components/notifications/NotificationMessage.svelte
+++ b/svelte/src/lib/components/notifications/NotificationMessage.svelte
@@ -58,7 +58,12 @@
   </div>
 
   <div class="notification__content">
-    {notification.message}
+    {#if notification.htmlContent}
+      <!-- eslint-disable-next-line svelte/no-at-html-tags -->
+      {@html notification.message}
+    {:else}
+      {notification.message}
+    {/if}
   </div>
 
   <button type="button" class="notification__close" onclick={handleClose} aria-label="Dismiss notification">
@@ -122,11 +127,9 @@
   }
 
   .notification__content {
-    display: flex;
     flex: 1 1 auto;
     min-width: 0;
     min-height: 0;
-    justify-content: space-between;
     padding: 0.5rem 1rem;
     word-break: break-word;
     line-height: 1.5;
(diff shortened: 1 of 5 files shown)


svelte/src/lib/components/notifications/NotificationMessage.svelte
```diff
@@ -122,11 +127,9 @@
   }
 
   .notification__content {
-    display: flex;
```
LawnGnome, line 125: This was unnecessary in practice — without HTML content, the `.notification__content` element can't have any children other than a single text node, which isn't affected by `display: flex` in this case, and with HTML content, we want it to flow normally. This was likely carried over from the Ember implementation, which nests the close button differently and does require the content to be `display: flex` (note the hack above in the Ember app to ensure the banner message is in a single element to work around that).

svelte/src/lib/components/notifications/Notifications.stories.svelte
```diff
@@ -38,6 +38,23 @@
         >
           Show Persistent
         </button>
+        <button
+          type="button"
+          onclick={() =>
+            notifications.info('The last word in this message is <strong>stronk</strong>.', { autoClear: false })}
```
Turbo87, line 44: idk if `stronk` was intentional, but let's keep it 😂

LawnGnome:
> I think I remember trying links in notifications before and something wasn't quite working as intended

Your memory is correct; this is https://github.com/mansona/ember-cli-notifications/issues/269.

I'm going to work around this with an `onclick` in the banner message `<a>`. It works correctly on Svelte, so this probably isn't worth trying to fix or patch, since we'll probably only ever have one HTML banner between now and when we retire the Ember frontend anyway.

LawnGnome:
Ugh, of course this doesn't work on a real deployment because of the content security policy. OK, I'll think of something else.

Issue #13193, opened 2026-03-16
503 from `cargo publish`


joshlf:
When running `cargo publish`:

```
   Uploading cargo-hermes v0.1.0-alpha.1 (/usr/local/google/home/joshlf/workspace/tmp/hermes-publish/zerocopy/tools/hermes)
error: failed to publish cargo-hermes v0.1.0-alpha.1 to registry at https://crates.io

Caused by:
  failed to get a 200 OK response, got 503
  headers:
        HTTP/2 503
        server: Varnish
        retry-after: 0
        content-type: text/html; charset=utf-8
        accept-ranges: bytes
        date: Mon, 16 Mar 2026 18:58:01 GMT
        via: 1.1 varnish
        x-served-by: cache-iad-kiad7000142-IAD
        x-cache: MISS
        x-cache-hits: 0
        x-timer: S1773687481.010830,VS0,VE787
        strict-transport-security: max-age=31536000; includeSubDomains
        content-length: 455
  
  body:

  <?xml version="1.0" encoding="utf-8"?>
  <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
   "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
  <html>
    <head>
      <title>503 backend write error</title>
    </head>
    <body>
      <h1>Error 503 backend write error</h1>
      <p>backend write error</p>
      <h3>Error 54113</h3>
      <p>Details: <redacted></p>
      <hr>
      <p>Varnish cache server</p>
    </body>
  </html>
```

carols10cents:
In our logs, I'm seeing publish requests for `cargo-hermes` returning `413 Payload Too Large`, so there's a problem here with getting the correct error returned to cargo somehow. Not sure exactly where the problem is occurring yet.

How large is the `.crate` file in `target/package` for this version?

carols10cents:
I'm discussing this with other members of our team-- this might be an issue with one of our CDNs as well. If your crate is under 10MB, it might be the CDN; we'll update soon.

joshlf:
Ah good catch! It was 26MB because I was including binary test blobs. After excluding them, [it worked](https://crates.io/crates/cargo-hermes/0.1.0-alpha.1).

Could be good to have a clearer error message, I suppose? So I'll leave this issue open for that, but my issue has been resolved.

LawnGnome:
[...]

PR #13381, opened 2026-04-10
smoke-test: Use Trusted Publishing for staging token

Turbo87:
Replace the long-lived `STAGING_SMOKE_TEST_TOKEN` secret with a short-lived token obtained via `rust-lang/crates-io-auth-action` pointed at staging.crates.io.

As a nice side effect, each smoke test run now also exercises the Trusted Publishing GitHub Action and the token exchange endpoint end-to-end, giving us continuous coverage of that flow against staging.

Dropping the repo secret also resolves the `secrets-outside-env` zizmor warning on this workflow.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -16,6 +16,9 @@ jobs:
   test:
     runs-on: ubuntu-24.04
 
+    permissions:
+      id-token: write
+
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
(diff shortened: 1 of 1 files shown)


Turbo87:
I'm merging this directly to test it out, since, due to the nature of Trusted Publishing, I couldn't really test this properly otherwise 😅

Turbo87:
well... it worked (see https://staging.crates.io/crates/crates-staging-test-tb/0.69.286), but the run was also hitting https://github.com/rust-lang/crates.io/issues/13193 so it failed for unrelated reasons 🙈

Turbo87:
@rust-lang/infra the `STAGING_SMOKE_TEST_TOKEN` secret can be removed from the repo now 🙏

marcoieni:
done! 
image

PR #13445, opened 2026-04-17
index: Switch the crate-index clone to a bare repository

Turbo87:
`reset_head()` used to run `git fetch` + `git reset --hard` before every index job. On Heroku's overlay FS the reset is slow. It had to stat all ~254k tracked files (one per published crate). A bare clone has neither a worktree nor a staging index, so the cost disappears.

Migrating meant refactoring `Repository` to no longer use the filesystem:
- `read_entry` used to `fs::read` a path in the checkout. It now reads the corresponding blob from the HEAD tree.
- a `CommitBuilder` was introduced to allow us to write blobs straight to the ODB and compose the new tree with
  `TreeUpdateBuilder`.
- `reset_head` and `squash_to_single_commit` both depended on `git reset --hard` to move HEAD, and now use a forced ref write on `refs/heads/master` instead.

With those rewrites in place I was able to switch the `open()` call to use `--bare` and no longer write any files to disk.

I temporarily deployed this branch to staging.crates.io for manual testing. Regular `cargo publish` behaves identically to the current backend. `SquashIndex` produces the expected collapsed history too.

I recommend reviewing this commit-by-commit, to avoid being overwhelmed by the diff 😅

### DIFF
--- a/crates/crates_io_index/commit_builder.rs
+++ b/crates/crates_io_index/commit_builder.rs
@@ -9,13 +9,22 @@ use tracing::{info, instrument};
 /// [`Repository::commit_builder_to`]. Stage changes through [`Self::upsert_entry`]
 /// and [`Self::remove_entry`], then call [`Self::commit_and_push`] to write the
 /// commit and push it to the target branch. Dropping the builder without
-/// calling `commit_and_push` leaves the staged worktree writes in place; they
-/// will be cleaned up by the next [`Repository::reset_head`] call.
+/// calling `commit_and_push` discards the staged operations; any blobs that
+/// were written to the ODB become unreachable and will be cleaned up by
+/// `git gc`.
+///
+/// ### Limitation
+///
+/// Each path may be touched at most once per builder: mixing or repeating
+/// `upsert_entry` / `remove_entry` for the same `name` is unsupported and
+/// will fail at commit time. libgit2's `git_tree_create_updated` rejects
+/// duplicate operations on the same path, upserts of previously-removed
+/// paths, and type-changing upserts.
 pub struct CommitBuilder<'a> {
     repo: &'a Repository,
     msg: String,
     branch: String,
-    index: git2::Index,
+    tub: git2::build::TreeUpdateBuilder,
 }
 
 impl<'a> CommitBuilder<'a> {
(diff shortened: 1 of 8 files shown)


crates/crates_io_index/repo.rs
```diff
@@ -164,6 +164,33 @@ impl Repository {
         Self::relative_index_file_helper(&name).join("/")
     }
 
+    /// Returns the crate names of all entries currently stored in the index.
+    ///
+    /// Top-level files (e.g. `config.json`) are excluded; only blobs nested
+    /// under the sharded `N[/prefix]/name` layout are returned.
+    pub fn list_entries(&self) -> anyhow::Result<Vec<String>> {
```
arlosi, line 171: There's also the [`.github`](https://github.com/rust-lang/crates.io-index/tree/master/.github/workflows) folder that should be excluded.

crates/crates_io_index/repo.rs
```diff
@@ -212,15 +212,25 @@ impl Repository {
...
+            Ok(entry) => {
+                let blob = entry
+                    .to_object(&self.repository)
+                    .context("Failed to resolve tree entry")?
+                    .peel_to_blob()
+                    .context("Failed to peel tree entry to blob")?;
+                Ok(Some(blob.content().to_vec()))
+            }
+            Err(error) if error.code() == git2::ErrorCode::NotFound => Ok(None),
+            Err(error) => Err(error).context("Failed to look up tree entry"),
```
arlosi, line 233: Maybe this is already exposed in the error chain, but it might be useful to know what `name` is failing to look up.

crates/crates_io_index/commit_builder.rs
```diff
@@ -9,13 +9,22 @@ use tracing::{info, instrument};
...
-/// calling `commit_and_push` leaves the staged worktree writes in place; they
-/// will be cleaned up by the next [`Repository::reset_head`] call.
+/// calling `commit_and_push` discards the staged operations; any blobs that
+/// were written to the ODB become unreachable and will be cleaned up by
+/// `git gc`.
+///
+/// ### Limitation
+///
+/// Each path may be touched at most once per builder: mixing or repeating
+/// `upsert_entry` / `remove_entry` for the same `name` is unsupported and
```
arlosi, line 19: We may want to lift this limitation if we start batching commits. But it's fine for now.

arlosi (approved):
This looks good to me!

I think I recommend continuing to use the `git` CLI to do the network operations (fetch and push), as it seems more stable.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13445/b27faa2ad38a2db5f63836842e3a146316d9dbf1..359259fa5b16f329077594e3c40d000fe766c5ad)*

crates/crates_io_index/commit_builder.rs
Turbo87: I think even if we batch commits, we would only ever sync the same crate once in a job, but I guess we'll see when we get there. I do wonder how necessary the batching still is with this PR applied though.

Byron:
I love seeing this PR, as there really isn't a good reason to maintain a worktree checkout if all work on the repository is automated anyway.

With that said, `gix` also supports everything you are doing and if there are ever any problems with `git2` you'd have an alternative.

PR #13462, opened 2026-04-20
svelte: Bring HTML `<head>` closer to the Ember app

Turbo87:
The Svelte app's `200.html` currently ships with only SvelteKit's defaults, which means the social previews, favicons, OpenSearch discovery and most of the iOS chrome integration are all missing compared to what the Ember app serves today. This PR closes most of that gap.

Most of the work lives in `svelte/src/app.html` and is just the missing tags: fallback title, favicon and app icons, the Apple iOS PWA bits, `theme-color`, the OpenSearch `<link>`, and the Open Graph / Twitter card meta tags. The OpenSearch descriptor itself is generated by a prerendered `+server.ts` endpoint rather than a static file, so the URLs inside automatically follow `kit.paths.base` (`/svelte` while we coexist with Ember, `""` once Ember goes away). The origin used during prerender comes from `DOMAIN_NAME` if set, falling back to `crates.io`, which means staging advertises its own URL and production advertises the canonical one.

The per-crate `og:image` already had backend infrastructure in `src/middleware/ember_html.rs`: the middleware was loading `svelte/build/200.html` as a Jinja template but the template never had a placeholder. Adding `{{ og_image_url }}` to `app.html` was enough to wire it up (the placeholder survives the SvelteKit build untouched, since SvelteKit only processes `%sveltekit.*%`). Alongside that, `extract_crate_name` now also recognises the temporary `/svelte/crates/...` prefix, with a comment marking the constant for removal once the Svelte app moves to `/`.

A few things from the Ember `<head>` were intentionally not ported. The `manifest.webmanifest` and `browserconfig.xml` files are gone, since crates.io has no PWA install flow, no service worker, and no offline support, and the browserconfig has been dead markup since IE11 retired. `X-UA-Compatible` falls in the same bucket. The blanket `notranslate` attributes were left out too: Svelte 5 is affected by the same Google-Translate hydration bug as Ember (sveltejs/kit#13272), but disabling translation site-wide is an a11y anti-pattern, and the better fix is targeted `translate="no"` on identifier-bearing elements (crate names, versions, code blocks), which is more naturally done as part of component work. Finally, the Fira font `<link>` and the CSS `@layer global, components;` declaration aren't needed: the Svelte app already imports the font from `global.css` and its component scoped styles are unlayered, so the explicit layer order isn't required.
### Related

- https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/Crates.20and.20svelte/near/586343012


eth3lbert (approved):
LGTM! Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13462/55155c753d26323f86d6c0ad63e8f487c0f1ebab..43d9fc233e6925491e3e3a1f45bcd34b5bb0c17a)*

PR #13463, opened 2026-04-20
svelte: Fix search input focus stealing on client-side navigation

Turbo87:
Fixes a regression reported on Zulip where interacting with a filter, internal link, or navigation tab in the Svelte app steals focus back to the main search bar.

Root cause: Svelte 5's `{autofocus}` handler seems to set the attribute on mount but never remove it when the value changes. `SearchForm` lives in `+layout.svelte` and is first rendered on `/` with `autofocus=true`, so the attribute sticks. SvelteKit's `reset_focus()` then refocuses any `[autofocus]` element after each client-side navigation.

Replaces the attribute with an `afterNavigate` callback that focuses the input once on the initial navigation. Passes `keepFocus: true` to `goto()` in the submit handler so pressing Enter keeps focus on the input, matching the Ember app. Also switches `SearchForm` to `bind:value`, which fixes a separate reactive-state race on fast typing followed by Enter, and adds regression tests in `e2e/acceptance/search.spec.ts`.
### Related

- https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/Crates.20and.20svelte/near/587669795

### DIFF
--- a/e2e/acceptance/search.spec.ts
+++ b/e2e/acceptance/search.spec.ts
@@ -59,33 +59,60 @@ test.describe('Acceptance | search', { tag: '@acceptance' }, () => {
     await expect(page.locator('[data-test-search-input]')).toHaveValue('');
   });
 
-  test('pressing S key to focus the search bar', async ({ page, msw }) => {
-    await loadFixtures(msw.db);
+  test.describe('search input focus behavior', () => {
+    test('pressing S key focuses the search input', async ({ page, msw }) => {
+      await loadFixtures(msw.db);
 
-    await page.goto('/');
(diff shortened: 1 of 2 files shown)


eth3lbert (approved):
It's great that users reported the regression. The fix and the added test case have everything back in working correctly again! Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13463/4e880b0eb652238576112b8083b3e51b26e386b9..f2e0ecb2362dbb3d2f36ff3919cbb24f742ba8c9)*

PR #13470, opened 2026-04-21
svelte: Drop `scroll-behavior: smooth` from global CSS

Turbo87:
On Firefox, especially at narrow viewports, clicking a crate on the `/crates` listing page sometimes left the crate detail page scrolled near the previous page's offset instead of at the top. The global `scroll-behavior: smooth` rule on `html, body` interacts poorly with SvelteKit's programmatic scroll-to-top on navigation, animating the reset in a way Firefox often fails to complete.

Dropping the rule restores reliable scroll reset on route change. Smooth in-page anchor scrolling can be reintroduced per use-site if needed via `scrollIntoView({ behavior: 'smooth' })`.

### Related

- https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/Crates.20and.20svelte/near/587821713

### DIFF
--- a/svelte/src/lib/css/global.css
+++ b/svelte/src/lib/css/global.css
@@ -109,7 +109,6 @@
   html,
   body {
     margin: 0;
-    scroll-behavior: smooth;
   }
 
   body {


eth3lbert (approved):
Thanks! I haven't run into this yet. However, after checking the SvelteKit issues, I see that others are reporting the same thing and mentioning scroll-behavior too, so the move makes sense to me. 👍

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13470/d0ae9603260b9da57b0e8e982d0ee8686a369d77..339f12eec21366c74fef134e10c89f8c07fbbb62)*

PR #13472, opened 2026-04-21
Extract archive push into `ArchiveIndexBranch` background job

Turbo87:
Extract the archive push from `SquashIndex` into a new `ArchiveIndexBranch` background job. The job reads `GIT_ARCHIVE_REPO_URL` from `config::Server` (loaded once at startup) and is a no-op if no URL is configured. `SquashIndex` enqueues it after a successful squash-and-push, warning but not failing if the enqueue itself fails. The job is also manually enqueuable via `crates-admin enqueue-job archive_index_branch <branch>`.

This change ensures that a failing archival will not unintentionally retrigger another index squash. It also allows us in the future to move the `ArchiveIndexBranch` job to a low-priority job on the default background worker queue, which unlocks the regular index repository earlier, allowing the git index sync jobs to continue earlier too.

### Related

### DIFF
--- a/crates/crates_io_index/testing.rs
+++ b/crates/crates_io_index/testing.rs
@@ -1,6 +1,6 @@
 use anyhow::anyhow;
 use git2::build::TreeUpdateBuilder;
-use git2::{ErrorCode, FileMode, Repository, Sort};
+use git2::{ErrorCode, FileMode, Oid, Repository, Sort};
 use std::path::{Path, PathBuf};
 use std::sync::Mutex;
 use std::thread;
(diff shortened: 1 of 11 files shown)


Turbo87 (commented):
I've tested this on our staging environment and everything seems to work as intended

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13472/6fe76ad222715f001b9bd804cc6c1c025a79e27e..aaaf561cc2895c318a342bc547f2fcbd5a8627fa)*

PR #13487, opened 2026-04-24
Remove the PostgreSQL `random_string` function

LawnGnome:
`random_string` is implemented in terms of the built-in PostgreSQL `random` function, which is explicitly documented as not to be relied upon for anything security-sensitive. While we migrated API token generation to use a more secure alternative several years ago, e-mail verification and crate invitation tokens are still generated using `random_string`.

In practice, I'm not terribly concerned about this (hence why I'm opening it as a regular PR and not via the GitHub security features), but we may as well clean this up. I've basically done this the same way API tokens are implemented: by creating a new `GenericToken` type that re-uses the same generator as API tokens, and implements the right traits to be used as a field type directly with Diesel.

One user-visible change here is that tokens created after this is deployed will be 32 characters instead of 26, but I think that's fine. (We could change `GenericToken::generate` to take a length, but that feels unnecessary.)


Turbo87:
out of curiosity, is this a mythical PR?

LawnGnome:
> out of curiosity, is this a mythical PR?

Yes, the issue came out of a Mythos security scan of crates.io.

Turbo87:
if that's the worst it could find then I'd be pretty happy 😂

LawnGnome:
> if that's the worst it could find then I'd be pretty happy 😂

Don't worry; I told it to try harder.

migrations/2026-04-23-232304-0000_remove-random-string-usage/up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE crate_owner_invitations
+ALTER COLUMN token
+DROP DEFAULT;
+
+ALTER TABLE emails
+ALTER COLUMN token
+DROP DEFAULT;
+
+DROP FUNCTION random_string;
```
Turbo87, line 9: just so it doesn't get lost in Slack DMs:

the `random_string` function is still being used by the `reconfirm_email_on_email_change` database trigger that has not been replaced yet by the first commit in this PR. that is why the CI runs are currently failing with `function random_string(integer) does not exist` (not visible in CI, but locally with small adjustments to make the error visible).

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13487/150d165e4d5ecda24b60f146589db4def7fd34bf..7c54ddb7470b62f8a98dbf9702f3c60c8d2de0ff)*

Turbo87:
> `random_string` is implemented in terms of the built-in PostgreSQL `random` function, which is explicitly documented as not to be relied upon for anything security-sensitive

I wonder if using https://www.postgresql.org/docs/current/pgcrypto.html#PGCRYPTO-RANDOM-DATA-FUNCS might be an easier fix 🤔

LawnGnome:
Closed in favour of #13656.

PR #13489, opened 2026-04-24
jobs/index/archive: Authenticate archive push via GitHub App

Turbo87:
The index SSH key used in production is a deploy key for the index repository, so it cannot push to the separate archive repository. Instead of minting a second SSH key plus user account with access to both repos, we registered a GitHub App scoped to both the index and archive repos. This job is the first consumer; the remaining index-writing jobs may switch from the deploy key to the app later.

Mints an installation access token from `GH_INDEX_SYNC_APP_CLIENT_ID` and `GH_INDEX_SYNC_APP_PRIVATE_KEY` after fetching the snapshot branch, then pushes `FETCH_HEAD` to the archive repository over HTTPS via a temporary `archive` remote carrying `x-access-token` credentials. For URL schemes that do not accept userinfo (e.g. `file://` in tests), the job logs a warning and falls back to pushing without credentials.

Fails loudly when `index_archive_url` is set but no GitHub App is configured, so a misconfigured worker does not silently skip the push.

### Related

- https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/index.20archive.20deploy.20key/with/590674283
- https://github.com/rust-lang/crates.io/pull/13472


Turbo87:
Note that I've tested this with https://github.com/rust-lang/staging.crates.io-index and https://github.com/rust-lang/staging.crates.io-index-archive and after an initial hickup (see below), everything worked as intended.

The hickup was this:

```
Failed to run job: Running git command failed with: To https://github.com/rust-lang/staging.crates.io-index-archive.git
 ! [remote rejected] FETCH_HEAD -> snapshot-2026-04-24 (refusing to allow a GitHub App to create or update workflow `.github/workflows/update-dl-url.yml` without `workflows` permission)
error: failed to push some refs to 'https://github.com/rust-lang/staging.crates.io-index-archive.git'
```

Apparently GitHub Apps are not allowed to push anything that changes GitHub Actions workflows, and since the index currently contained a workflow, but the archive didnt, the push failed. I've removed the workflow from the staging index and then it worked. The workflow isn't particularly helpful anymore anyway, since it only updates the `config.json` of the git index, but not the sparse index.

crates/crates_io_github_app/src/test_keys.rs
walterhpearce, line 1: Can you variable rename these variables `TEST_KEY_PEM....` etc? Automated secret scanning will alert on these otherwise. Not required, just one less piece of noise for an exception.



*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13489/8f520adaa8ab39681142e666128f79baffb41873..803015ddf0b7217a7c3b0d1774d3512e37a5f917)*

crates/crates_io_index/repo.rs
```diff
@@ -354,6 +354,47 @@ impl Repository {
...
+/// associated remote from the repository when dropped.
+pub struct TemporaryRemote<'a> {
+    repo: &'a Repository,
+    name: String,
+}
+
+impl Drop for TemporaryRemote<'_> {
+    fn drop(&mut self) {
+        if let Err(error) = self.repo.remove_remote(&self.name) {
+            warn!("Failed to remove `{}` remote: {error}", self.name);
```
LawnGnome: Nit, but I'd be tempted to make this an `error!()`, since it probably implies something else has gone wrong that needs investigation beyond just having a dangling remote.

LawnGnome (approved):
:shipit: 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13489/8f520adaa8ab39681142e666128f79baffb41873..803015ddf0b7217a7c3b0d1774d3512e37a5f917)*

PR #13517, opened 2026-04-27
jobs/index: Add `SquashIndexViaApi` job

Turbo87:
The existing `SquashIndex` job has been OOMing the production worker because `git push` regenerates the full pack on every squash. This adds an alternative job that drives the same squash through GitHub's API (`get_ref` -> `get_commit` -> `create_commit` -> `create_ref` -> drift re-check -> `update_ref`), so the server can reuse objects it already has instead of receiving a fresh pack.

The original `SquashIndex` job stays in place as a fallback. Both share the `repository` queue so index-writing jobs continue to serialize, even though the new job does not write to the local bare repo.

Along the way the `crates_io_github` crate gains:

- `parse_github_slug()` helper
- A configurable base URL on `RealGitHubClient` plus a mockito-based test harness
- `get_ref()` / `get_commit()` trait methods (unauthenticated, sufficient for the public index repos)
- `create_commit()` / `create_ref()` / `update_ref()` write methods authenticated via `AccessToken`

`Environment` now carries an `Arc<dyn GitHubClient>` so jobs can talk to GitHub directly, and `crates-admin enqueue-job` exposes the new job for manual operator use.

### Related

- https://github.com/rust-lang/crates.io/pull/13472
- https://github.com/rust-lang/crates.io/pull/13489

### DIFF
--- a/src/worker/jobs/index/squash.rs
+++ b/src/worker/jobs/index/squash.rs
@@ -1,14 +1,20 @@
 use crate::tasks::spawn_blocking;
 use crate::worker::Environment;
 use crate::worker::jobs::ArchiveIndexBranch;
+use anyhow::{Context, anyhow};
 use chrono::Utc;
+use crates_io_github::{CreateCommit, parse_github_slug};
 use crates_io_worker::BackgroundJob;
+use oauth2::AccessToken;
+use secrecy::ExposeSecret;
 use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::process::Command;
 use std::sync::Arc;
 use std::time::Instant;
 use tracing::{info, instrument, warn};
+
+const MASTER_REF: &str = "refs/heads/master";
 
 #[derive(Serialize, Deserialize)]
 pub struct SquashIndex;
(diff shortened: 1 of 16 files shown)


Turbo87:
as mentioned in https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/index.20squashing/near/590951284, I've successfully tested this PR on our staging environment. the staging index is significantly smaller than the production index, but I'm fairly confident that the same logic should also work on production. if not, we can still fall back to the original background job for now.

src/worker/jobs/index/squash.rs
```diff
@@ -87,3 +89,106 @@ async fn enqueue_archive_job(env: &Environment, branch: &str) -> anyhow::Result<
...
+        let squash_start = Instant::now();
+        let input = CreateCommit {
+            message: &message,
+            tree: &tree_sha,
+            parents: &[],
+        };
+        let new_commit = github.create_commit(&owner, &repo, &input, &auth).await?;
+        let new_sha = new_commit.sha;
+        let duration = squash_start.elapsed().as_nanos();
+        info!(duration, "Squash commit created: {new_sha}",);
```
LawnGnome:
```suggestion
        info!(duration, "Squash commit created: {new_sha}");
```

src/worker/jobs/index/squash.rs
```diff
@@ -87,3 +89,106 @@ async fn enqueue_archive_job(env: &Environment, branch: &str) -> anyhow::Result<
...
+            .await?;
+
+        // Best-effort drift check. GitHub has no CAS for refs, so the
+        // `repository` queue is the real primary defense; this only
+        // shrinks the race window.
+        let current_head = github.get_ref(&owner, &repo, MASTER_REF).await?;
+        if current_head.object.sha != original_sha {
+            return Err(anyhow!(
+                "`master` drifted during squash (was {original_sha}, now {})",
+                current_head.object.sha
```
LawnGnome:
```suggestion
                "`{}` drifted during squash (was {original_sha}, now {})",
                MASTER_REF,
                current_head.object.sha
```

src/worker/jobs/index/squash.rs
```diff
@@ -87,3 +89,106 @@ async fn enqueue_archive_job(env: &Environment, branch: &str) -> anyhow::Result<
...
+        let token = github_app.installation_token().await?;
+        let auth = AccessToken::new(token.expose_secret().into());
+
+        let squash_start = Instant::now();
+        let input = CreateCommit {
+            message: &message,
+            tree: &tree_sha,
+            parents: &[],
+        };
+        let new_commit = github.create_commit(&owner, &repo, &input, &auth).await?;
```
LawnGnome, line 158: It took me a few read-throughs to understand exactly why this works — the fact tree objects are logically separate from commits is one of those things that you (well, I, at least) don't tend to think about much as an explicit part of the data model that underpins Git.

Perhaps we could have a brief comment somewhere in here explaining the theory for the benefit of future crates.io maintainers? (Or, more likely, me in six months.)

LawnGnome (approved):
There's quite a lot here. I _think_ I understand how it all fits together, but I don't think I'd want to have to pass an exam that deep dived into it.

If there's one thing that just worries me a tad here, it's the possibility of GitHub itself drifting in terms of API behaviour. The test coverage here is excellent, but relies on mocks, and this feels like a part of the GitHub API that is lower level than most users ever touch. I don't practically know how we'd test it without significant effort — we'd presumably need a smoke test type thing with a real repo in a known state that we could manipulate, and it's probably not worth it, but I like to write down my vague worries just in case.

Anyway, since this has been tested on staging already, I'm happy for us to give it a go.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13517/d23e9fa716d7afaa3fc763f3da37fd3ef8c72687..cca980e89a4c49fa41a8bfba3b49cfdfd849e55e)*

src/worker/jobs/index/squash.rs
Turbo87: I've added [097eeba](https://github.com/rust-lang/crates.io/pull/13517/commits/097eeba456105eb1a9766001ebfccf5fb620041f) which should hopefully make this clearer :)

PR #13518, opened 2026-04-27
Support regexes to block traffic by user agent

carols10cents:
So that we don't have to keep up with version numbers ;)

A few things that worry me:

- Splitting on commas is a problem regardless of this PR, if you wanted the comma to be part of the specified match. I'm worried with the addition of regex support we'll think "oh right, i'll just escape it and it won't be split on that comma" which... won't work, which at least I've documented in a test?
- ~~Deploying this will be a little tricky... we have the `BLOCKED_TRAFFIC` env var that we can use as a bit of indirection so that we could set an entirely new env var, deploy and maybe have everything broken for a second, then change the value of `BLOCKED_TRAFFIC` to use the new env var? Then after letting everything settle unset the old env var?~~
  ~~The main problem is a bunch of the current values have literal parens and periods that need to be escaped... maybe we could _duplicate_ those values _within_ the current env var's values (so we'd change a current value of `foo,bar (1.2.3)` to `foo,bar (1.2.3), bar \(1\.2\.3\)` [and then eventually to `foo,bar \([\d]+\.[\d]+\.[\d]+\)` to take advantage of the new capabilities] and just accept that some weird stuff is getting matched during the transition?~~

~~EDIT: Just checked that there's at least one value in the current env var value that contains a `+` that would panic with this PR, so I'm switching this to draft so that it doesn't get merged, but I would love a review and thoughts about how to merge+deploy this :)~~

The deploy worry has been mitigated by supporting both strings _and_ regex and you have to opt in to the regex behavior by surrounding the value with `/`, so all existing values should continue to be interpreted as the regular old string equality matches they are before this PR :)

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -280,12 +280,17 @@ fn blocked_traffic() -> Vec<(String, Vec<String>)> {
     parse_traffic_patterns(&pattern_list)
         .map(|(header, value_env_var)| {
             let value_list = dotenvy::var(value_env_var).unwrap_or_default();
-            let values = value_list.split(',').map(String::from).collect();
+            let values = parse_traffic_pattern_values(&value_list);
             (header.into(), values)
         })
         .collect()
 }
 
+/// Extract from the `BLOCKED_TRAFFIC` env var value a comma-separated list of pairs containing a
+/// header name, an equals sign, and the name of another environment variable that contains the
+/// values of that header that should be blocked. For example, if `BLOCKED_TRAFFIC` is set to
+/// `User-Agent=BLOCKED_UAS,custom-header=BLOCKED_CUSTOM`, this function will return the pairs
+/// (`User-Agent`, `BLOCKED_UAS`) and (`custom-header`, `BLOCKED_CUSTOM`).
 fn parse_traffic_patterns(patterns: &str) -> impl Iterator<Item = (&str, &str)> {
     patterns.split_terminator(',').map(|pattern| {
         pattern.split_once('=').unwrap_or_else(|| {
@@ -295,6 +300,13 @@ fn parse_traffic_patterns(patterns: &str) -> impl Iterator<Item = (&str, &str)> 
             )
         })
     })
+}
+
+/// After reading the value of an environment variable whose name was specified in the value of
+/// `BLOCKED_TRAFFIC`, parse a comma-separated list of values to be used as exact matches of the
+/// values of the header name specified in the `BLOCKED_TRAFFIC` pair.
+fn parse_traffic_pattern_values(value_list: &str) -> Vec<String> {
+    value_list.split(',').map(String::from).collect()
 }
 
 #[derive(Clone, Debug, Default)]
@@ -340,5 +352,20 @@ mod tests {
         assert_eq!(vec![("Baz", "QUX")], patterns_2);
 
         assert_none!(parse_traffic_patterns(pattern_string_3).next());
+    }
+
+    #[test]
+    fn parse_traffic_pattern_values_splits_on_comma_even_if_escaping_is_attempted() {
+        let pattern = "web-tool 1.2.3,fancy-crate\\, run by fancy-author v4.5.6";
+
+        let values = parse_traffic_pattern_values(pattern);
+        assert_eq!(
+            vec![
+                String::from("web-tool 1.2.3"),
+                String::from("fancy-crate\\"),
+                String::from(" run by fancy-author v4.5.6"),
+            ],
+            values
+        );
     }
 }
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -9,11 +9,10 @@ use super::base::Base;
 use super::database_pools::DatabasePools;
 use crate::config::CdnLogQueueConfig;
 use crate::config::cdn_log_storage::CdnLogStorageConfig;
-use crate::middleware::cargo_compat::StatusCodeConfig;
+use crate::middleware::{block_traffic::BlockCriteria, cargo_compat::StatusCodeConfig};
 use crate::storage::StorageConfig;
 use crates_io_env_vars::{list, list_parsed, required_var, var, var_parsed};
 use http::HeaderValue;
-use regex::Regex;
 use std::collections::{HashMap, HashSet};
 use std::convert::Infallible;
 use std::net::IpAddr;
@@ -49,7 +48,7 @@ pub struct Server {
     pub max_features: usize,
     pub rate_limiter: HashMap<LimitedAction, RateLimiterConfig>,
     pub new_version_rate_limit: Option<u32>,
-    pub blocked_traffic: Vec<(String, Vec<Regex>)>,
+    pub blocked_traffic: Vec<(String, Vec<BlockCriteria>)>,
     pub blocked_ips: HashSet<IpAddr>,
     pub max_allowed_page_offset: u32,
     pub excluded_crate_names: Vec<String>,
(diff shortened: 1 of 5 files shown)


syphar (commented):
My guess is that since the Fastly WAF is enabled for crates.io, that this would be the place for things like this? 

https://github.com/rust-lang/simpleinfra/blob/master/terragrunt/modules/crates-io/fastly-webapp-ngwaf.tf

( I might be wrong )

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13518/d23e9fa716d7afaa3fc763f3da37fd3ef8c72687..ced11b008914be958bdd30a9fdf80730abf90049)*

src/config/server.rs
```diff
@@ -297,6 +301,24 @@ fn parse_traffic_patterns(patterns: &str) -> impl Iterator<Item = (&str, &str)>
...
 }
 
+/// After reading the value of an environment variable whose name was specified in the value of
+/// `BLOCKED_TRAFFIC`, parse a comma-separated list of values to be used as full regex matches of
+/// the values of the header name specified in the `BLOCKED_TRAFFIC` pair.
+fn parse_traffic_pattern_values(value_list: &str) -> Vec<Regex> {
+    value_list
+        .split(',')
+        .map(|value| {
+            let anchored = format!("^{}$", value);
```
Turbo87: this prevents us from having unanchored regexes. I wonder if those might become useful at some point?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13518/d23e9fa716d7afaa3fc763f3da37fd3ef8c72687..ced11b008914be958bdd30a9fdf80730abf90049)*

src/config/server.rs
```diff
@@ -297,6 +301,24 @@ fn parse_traffic_patterns(patterns: &str) -> impl Iterator<Item = (&str, &str)>
...
+            Regex::new(&anchored).unwrap_or_else(|e| {
+                panic!(
+                    "BLOCKED_TRAFFIC values must be a valid regex after `^` and `$` are added, \
+                     got invalid regex {anchored}: {e}"
+                )
+            })
```
Turbo87: what do you think about supporting static strings and regexes at the same time? I could imagine `foo` being interpreted as a regular string, and `/foo/` as a regex. that's the notation that JS uses for inline regex. we could `let isRegex = s.startsWith('/') && s.endsWith('/');` or something like that 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13518/d23e9fa716d7afaa3fc763f3da37fd3ef8c72687..ced11b008914be958bdd30a9fdf80730abf90049)*

carols10cents:
> My guess is that since the Fastly WAF is enabled for crates.io, that this would be the place for things like this?
> 
> [rust-lang/simpleinfra@`master`/terragrunt/modules/crates-io/fastly-webapp-ngwaf.tf](https://github.com/rust-lang/simpleinfra/blob/master/terragrunt/modules/crates-io/fastly-webapp-ngwaf.tf)
> 
> ( I might be wrong )
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13518/d23e9fa716d7afaa3fc763f3da37fd3ef8c72687..ced11b008914be958bdd30a9fdf80730abf90049)_

This code is for API requests to the application, not downloads of crates from S3-- I believe we only have S3 going through Fastly, not API requests? Is that right @Turbo87 ?

src/config/server.rs
carols10cents: Yeah, i was mostly going for "keep the same behavior for the existing values", but if everyone who has access to these is aware of the behavior change, this is less important. 

Making a mental note that i should go update the ops guide when this goes in

src/config/server.rs
carols10cents: hmmmmm lemme give that a try :)

Turbo87:
> I believe we only have S3 going through Fastly, not API requests?

the API requests are also going through Fastly these days and we've been experimenting with the Fastly WAF a bit as a potential replacement for these API-level IP blocks, but at the moment we still primarily use the blocks in the API servers.

src/config/server.rs
carols10cents: @Turbo87 One new commit implementing this, I think i like it! This should get rid of the deployment worries I had too, because all the existing values will continue to be interpreted as strings instead of as regexes! great idea, wdyt of the implementation?

carols10cents: ok and another commit fixing some docs I missed 😅

src/middleware/block_traffic.rs
```diff
@@ -21,23 +22,82 @@ pub async fn middleware(
...
+    ///
+    /// This will panic if the specified string is interpreted as a `Regex` but the regex is
+    /// invalid.
+    pub fn new(s: &str) -> Self {
+        let is_regex = s.starts_with('/') && s.ends_with('/') && s.len() > 2;
+        if is_regex {
+            // Slicing is safe here because we checked the starting and ending characters and the
+            // length before entering this branch
+            Self::Regex(Regex::new(&s[1..s.len() - 1]).unwrap_or_else(|e| {
+                panic!(
```
LawnGnome: What happens if we `panic!`? Does the service fail to start in that case?

(Obviously we're all perfect on the crates.io team and would never, ever mess up a regex, but maybe a cosmic ray could flip something.)

LawnGnome (commented):
LGTM in general, and I like the test coverage, but one question on the behaviour:

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13518/e06e6a0d45644d2190089ccd767147f3645a785b..71a02b9781c0d7aecac60e775b104b3055e97b35)*

src/middleware/block_traffic.rs
Turbo87: > Does the service fail to start in that case?

looks like it, yes. Heroku has a "preboot" feature, which runs the before/after states in parallel for a couple of minutes until it is sure that the after state works as intended, but we are not using that feature. It would probably require us to be much more careful with the SQL migrations.

we are using panics in a lot of other places on startup, though in this instance I guess it might be better to log a warning/error instead and ignore that specific value 🤔

Turbo87: I've pushed a couple of commits to adjust the behavior to `warn!()` instead of `panic!()`.

Turbo87:
also rebased to resolve the merge conflict in the `Cargo.toml` file

src/middleware/block_traffic.rs
carols10cents: > I've pushed a couple of commits to adjust the behavior to `warn!()` instead of `panic!()`.

That's fine and let's go for it, but i do want to push back a tiny bit: 

I'm imagining someone getting paged in the middle of the night because crates.io is struggling because someone is hammering the API, and the person responding wants to block these requests and go back to bed. They look at user agent metrics in datadog, write up a regex that's accidentally invalid and add it to the env var, the app restarts but oh no the requests are still continuing!!! What would be likely to happen next? Would sleepy you think to go look in the logs for warnings about invalid regexes, or would you start panicking (pun intended) that something else is wrong and start doing other things that won't solve the problem? At least a panic would be very clear.

Turbo87: > At least a panic would be very clear.

sure, but a panic would also make every single other request fail. I'm not sure that I'd prefer that 😅

LawnGnome: Yeah, I was thinking about that balance too, and I think I want crates.io to at least try to be available.

The bigger problem is that we don't really have a way of signalling configuration issues in a way that is immediately obvious to an admin that doesn't also take the service down. Which we should probably also fix, but one problem at a time.

LawnGnome (approved):

src/middleware/block_traffic.rs
carols10cents: I will put a note in the ops guide at least! 🤞🏻

PR #13373, opened 2026-04-08
CI: Run Playwright tests against Svelte app

Turbo87:
This should allow us to run the Percy visual regression checks against the Svelte app, as a final check whether we have missed anything significant in our porting process.
### Related

- https://github.com/rust-lang/crates.io/issues/12515

PR #13531, opened 2026-04-30
e2e: Close pages before disabling MSW

Turbo87:
Without this, requests still in flight (or fired by late-running hydration code) bypass the mocks once `worker.disable()` removes the interception and hit the real network instead, surfacing as `[404]` log lines from the SvelteKit `vite preview` web server.

### DIFF
--- a/e2e/helper.ts
+++ b/e2e/helper.ts
@@ -68,6 +68,13 @@ export const test = base.extend<AppOptions & AppFixtures>({
       };
 
       await use({ worker, db, authenticateAs });
+
+      // Close pages before disabling MSW. Otherwise any request still in
+      // flight (or fired by late-running hydration code) bypasses the
+      // mocks once `worker.disable()` removes the interception, and hits
+      // the real network instead.
+      await Promise.all(context.pages().map(p => p.close()));
+
       await db.reset();
       await worker.disable();
     },


Turbo87:
@kettanaito is this something that should be mentioned in the MSW docs? (or maybe it already is, in that case apologies for the ping 😅)

kettanaito:
Hi, @Turbo87. We've already addressed that and the proper fix is coming in 3.0 as that requires a breaking change both in the `worker.stop()` API and the worker script. 

In fact, could you please give [this branch](https://github.com/mswjs/msw/pull/2692#issuecomment-4192279628) a try and see if your change is still necessary with it? Thanks.

Turbo87:
> could you please give [this branch](https://github.com/mswjs/msw/pull/2692#issuecomment-4192279628) a try and see if your change is still necessary with it?

I tried the beta from `pkg.pr.new/msw@2692` against our full test suite, but it looks like the leaks are still there.

We are using `@msw/playwright` in this case, instead of the raw `msw`, so maybe that is the reason for this not fixing the issue?

Our local fix is to close all context pages before calling `worker.disable()` in the fixture, which prevents Playwright's `Fetch.disable` from racing with late requests.

Turbo87:
damn... the change in the PR worked perfectly locally, but I've rebased https://github.com/rust-lang/crates.io/pull/13373 on top of it and CI still shows leaked requests 😵‍💫

Turbo87:
interestingly, removing the `network.disable()` (or in our case `worker.disable()`) seems to fix it on CI 🤔 

... but locally it still shows the occasional leak 😵‍💫

kettanaito:
We are in the process of updating `@msw/playwright` to use the new architecture so understandably the issue will still be reproducible there. It will take some time to that binding to adopt 3.0.

Turbo87:
yeah, no worries. it's not breaking anything here, just unintentionally leaking requests to the dev server. feel free to ping me once `@msw/playwright` is ready for testing :)

PR #13477, opened 2026-04-22
Genericize OAuth profiles

quinnjr:

PR #13547, opened 2026-05-01
Backfill users.name from oauth_github.login

quinnjr:
Most of the username groundwork was tangled up in #13477. Splitting it out so it can land on its own, per @carols10cents.

`users.name` is only set when a GitHub profile has a display name, so plenty of accounts have it as NULL. This backfills the NULLs from `oauth_github.login`. Walks `users.id` in 5000-row batches with a commit between each so we're not holding locks while logins are happening. Idempotent — re-running is a no-op once everyone has a name. Run it the same way as `data_oauth_github.sql`: `psql -f <file>`.

### DIFF
--- a/migrations/data_user_name_backfill.sql
+++ b/migrations/data_user_name_backfill.sql
@@ -1,0 +1,40 @@
+-- Backfill `users.name` from `oauth_github.login` for users whose `name` is
+-- still NULL. `users.name` is currently set only when the GitHub profile has
+-- a display name, so users without one have NULL there; this fills those in
+-- with the GitHub username so every account has a non-NULL name.
+--
+-- Iterates by `users.id` range and COMMITs between batches so row locks
+-- release between iterations and concurrent traffic isn't blocked. The DO
+-- block must be run outside an explicit transaction (`psql -f <file>` is
+-- fine; do NOT wrap with BEGIN/COMMIT, since COMMIT inside DO requires the
+-- block to be at the top level).
+--
+-- Idempotent: re-running the file is a no-op once every account has a name,
+-- because the UPDATE filters on `users.name IS NULL`.
+
+SET lock_timeout = '5s';
+SET statement_timeout = '60s';
+
+DO $$
+DECLARE
+    lo INT;
+    hi INT;
+    pos INT;
+    batch_size CONSTANT INT := 5000;
+BEGIN
+    SELECT MIN(id), MAX(id) INTO lo, hi FROM users WHERE name IS NULL;
+    IF lo IS NULL THEN RETURN; END IF;
+
+    pos := lo;
+    WHILE pos <= hi LOOP
+        UPDATE users
+        SET name = oauth_github.login
+        FROM oauth_github
+        WHERE oauth_github.user_id = users.id
+          AND users.name IS NULL
+          AND users.id >= pos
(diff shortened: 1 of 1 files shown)


carols10cents:
I'm confused... I don't remember seeing backfill of `users.name` in #13477. What you had [that I commented on](https://github.com/rust-lang/crates.io/pull/13477#discussion_r3134757267) was attempting to backfill `oauth_github`, which wasn't needed? Why is this PR described as being pulled out of #13477 when it never existed there as far as i remember?

In [the username RFC's unresolved questions](https://github.com/rust-lang/rfcs/pull/3946/changes#diff-3cc1c6b20bcc16abfcd05bea52fc42073980bd6b3f86d130554f89c3a9047efbR564-R570), I have been considering that we should get rid of `users.name` completely, rather than backfill it. Currently it's a display name, not the username. So this change may not be needed at all.

quinnjr:
Honestly I don't agree. Username/display name aren't standard across SSO providers (Github provides a username from Github, but Bitbucket/Google/etc is centered around email addresses). To keep current users in the position they would initially be in after signing up from Github, we need to backfill a users username/display name on the user table (and probably drop it from SSO providers to centralize around a core "Crates Username" identity for a single source of truth when you can potentially have a 1-N map of SSO providers to a single Crates username).

Further, keeping usernames as a SSO provided field makes lookups tougher on the database (it's an extra JOIN that doesn't need to be added).

Finally, tying username to account is easier on support if a user changes preferred SSO due to reasons outside Crate's scope. Say user John was a Github sign-up but wants toove everything over to GitLab because he is sick of Github's recent downtime statistics. Changing a single SSO preferred account but maintaining the Crate's username is a lot easier than needing to open a support ticket for manual database intervention.

carols10cents:
> (and probably drop it from SSO providers to centralize around a core "Crates Username" identity for a single source of truth when you can potentially have a 1-N map of SSO providers to a single Crates username).

Yes, this is what the entire https://github.com/rust-lang/rfcs/pull/3946 is about. Have you read it?

quinnjr:
> > (and probably drop it from SSO providers to centralize around a core "Crates Username" identity for a single source of truth when you can potentially have a 1-N map of SSO providers to a single Crates username).
> 
> Yes, this is what the entire https://github.com/rust-lang/rfcs/pull/3946 is about. Have you read it?

Yes I have read it. Which is exactly what the point of this PR is for (backfilling username based on current Github credentials already in the Database).

If the point of the RFC is to instead make usernames wholly unique and configurable via Crates, that would be addressed later in other work. For now, a backfill of current identity to the core user is what's needed to not break existing functionality and brand identity of users who have already signed up through Github.

Unless I'm missing something entirely.

carols10cents:
> Unless I'm missing something entirely.

Yes, I don't think my explanation is working, so let's try an approach where you go explore in crates.io to experience what I'm trying to explain for yourself:

- Find all the places in crates.io's codebase that use `users.name`, and collect links in GitHub to those lines of code. Explain what happens currently when `users.name` is not set.
- Explain what would change in the user experience, exactly, if we were to merge this PR.
- Explain how a user would change the value of `users.name` if we add other oauth providers.
- Explain why we need to make this change right now.

When you have those explanations, and if you think we still need this PR after doing that, you may reopen!

PR #13568, opened 2026-05-03
svelte: Strip test selectors from production builds

Turbo87:
Add a Svelte preprocessor that removes `data-test-*` and `data-testid` attributes from `.svelte` markup. Mirrors what `ember-test-selectors` did for the old Ember app: keep selectors in dev, vitest, and Playwright e2e builds; remove them from the shipped production bundle.

The preprocessor uses `svelte/compiler`'s `parse` to find `Attribute` nodes by name and `magic-string` to splice them out by character offsets. It is registered in `svelte.config.js` only when `NODE_ENV=production` and neither `PLAYWRIGHT` nor `VITEST` is set.

### DIFF
--- a/svelte/src/build/strip-test-selectors.js
+++ b/svelte/src/build/strip-test-selectors.js
@@ -1,0 +1,45 @@
+import MagicString from 'magic-string';
+import { parse } from 'svelte/compiler';
+import { walk } from 'zimmerframe';
+
+/**
+ * Removes `data-test-*` and `data-testid` attributes from `.svelte` markup.
+ *
+ * Returns a Svelte preprocessor that always strips when invoked. The decision
+ * to invoke it lives in `svelte.config.js`.
+ *
+ * @returns {import('svelte/compiler').PreprocessorGroup}
+ */
+export default function stripTestSelectors() {
+  return {
+    name: 'strip-test-selectors',
+
+    markup({ content, filename }) {
+      let ast = parse(content, { modern: true, filename });
+
+      /** @type {Array<{ start: number, end: number }>} */
+      let removals = [];
+
+      walk(/** @type {import('svelte/compiler').AST.SvelteNode} */ (ast), null, {
+        Attribute(node, ctx) {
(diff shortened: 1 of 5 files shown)


svelte/src/build/strip-test-selectors.js
```diff
@@ -0,0 +1,45 @@
...
+      /** @type {Array<{ start: number, end: number }>} */
+      let removals = [];
+
+      walk(/** @type {import('svelte/compiler').AST.SvelteNode} */ (ast), null, {
+        Attribute(node, ctx) {
+          if (node.name.startsWith('data-test-') || node.name === 'data-testid') {
+            removals.push({ start: node.start, end: node.end });
+          }
+          ctx.next();
+        },
```
eth3lbert, line 29: It would be great to support `SpreadAttribute` as well, but I believe it is already good enough for us for now 👍 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13568/a81c6e0586bc8d761f469c5b10803b426a216e0a..6915c9d50dddb7cf6472cdbb9489bd4c4000a4c0)*

Turbo87: can you give an example?

eth3lbert: e.g.:
```typescript
<script>
let attrs = { "data-test-foo": "bar" };
</script>

<div {...attrs}></div>
<div {...{"data-test-foo": "bar"}}></div>
```

The variable case is trickier because the AST only gives us the variable name. The inlined version works, but I don't think anyone actually writes code like that, so it feels non-idiomatic. I'd say it isn't worth the effort at this stage.

Turbo87: I see. yeah, these kinds of dynamic values would be hard to catch with just AST analysis. but as far as I remember we're not using this kind of pattern in the app (yet), so this is probably not much of an issue. and in the end all that would happen is that we won't strip these specific selectors. as long as we still strip the other 99% this is probably fine :)

PR #13570, opened 2026-05-03
Replace Docker Compose dev setup with a devcontainer

Turbo87:
Drops the standalone `docker-compose.yml`, `backend.Dockerfile`, `frontend.Dockerfile`, `docker_entrypoint.sh`, and the postgres init script in favor of a single full-stack devcontainer under `.devcontainer/` that provides Rust, Node.js 24, pnpm, Postgres 16, `diesel_cli`, and Playwright in one container. It works with VS Code "Reopen in Container", GitHub Codespaces, and the standalone `devcontainer` CLI, so contributors can get a working environment without installing the toolchain locally.

`postCreateCommand` initializes the local git index, creates the test database, runs migrations, installs JavaScript dependencies and the Playwright Chromium headless shell, pre-fetches Rust crates, and seeds `.env` from `.env.sample` if missing so the backend boots out of the box. `postStartCommand` reruns `diesel migration run` on every container start, and the `db` service has a `pg_isready` healthcheck so the app waits for Postgres to actually accept connections.

Named volumes keyed by `${devcontainerId}` persist the cargo target dir, cargo registry, pnpm store, and `local_uploads/` across rebuilds without bleeding into the host worktree. Backend, Svelte dev server, and Storybook ports are published on `127.0.0.1` via compose so non-VS-Code clients reach them too. The Vite config now honors the existing `DEV_DOCKER` env var and binds `0.0.0.0` when it is set, matching the backend's behavior, so the published Svelte port actually reaches a listener inside the container.

`docs/CONTRIBUTING.md` documents the new quick-start path and `.devcontainer/README.md` covers the details and troubleshooting. The existing manual setup instructions remain as the alternative.


LawnGnome:
I'm not going to do a full review on this right now, but I do want a flag a concern on removing the Docker Compose environment — dev containers are still pretty tied to the Microsoft ecosystem of developer tooling, and might not be as useful to people not using VSCode (or Zed, which apparently has some level of support). I have no problem adding a dev container environment as well, but I am interested in how others are developing crates.io.

(That said, I'll immediately undercut my own argument — I wrote my own Docker Compose environment for crates.io development because I didn't really like how the in-repo compose environment worked, so I'm not actually particularly tied to it either in its current state. I just don't want to increase friction for anyone wanting to make a drive-by contribution.)

Turbo87:
> I do want a flag a concern on removing the Docker Compose environment

the devcontainer setup is using docker-compose underneath too though, so what specifically do you think is missing or should be integrated differently?

.devcontainer/post-create.sh
```diff
@@ -0,0 +1,36 @@
...
+    script/init-local-index.sh
+
+psql -h db -U postgres <<'SQL'
+SELECT 'CREATE DATABASE cargo_registry_test'
+WHERE NOT EXISTS (SELECT FROM pg_database WHERE datname = 'cargo_registry_test')\gexec
+SQL
+
+diesel migration run
+
+pnpm install
```
ehuss: Just FYI, I tried this on my system and this step hung waiting for user input to confirm some action. I think I unblocked it with this change:


```suggestion
pnpm install --config.confirmModulesPurge=false
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13570/bce71d6b652f6b7f5c8fd29b104b2b1ee336eae7..4453d3fc8829890f14ef1da1667c7e89414ac991)*

docs/CONTRIBUTING.md
```diff
@@ -91,6 +90,30 @@ git clone https://github.com/rust-lang/crates.io.git
...
+- On GitHub: open a [Codespace] for the repository.
+- From the command line: `devcontainer up` (requires the
+  [`devcontainer` CLI]).
+
+The container's first start handles database creation, migrations, and
+dependency installation. See [`.devcontainer/README.md`](../.devcontainer/README.md)
+for details.
+
+The remaining sections describe how to set the same environment up
+manually if you prefer not to use containers.
```
ehuss: It's pretty unclear to me what to do next at this stage. It created the devcontainer, and there are some ports open, but loading them returns an empty page.

I was thinking maybe I need to manually start the server, but trying that fails:

```
vscode ➜ /workspaces/crates.io (devcontainer) $ pnpm dev:local
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "dev:local" not found
```

At this point, I'm pretty lost on how to get it running.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13570/bce71d6b652f6b7f5c8fd29b104b2b1ee336eae7..4453d3fc8829890f14ef1da1667c7e89414ac991)*

.devcontainer/post-create.sh
Turbo87: oh, good catch! I guess I didn't have a conflicting `node_modules` on my system that it needed to purge, so I didn't run into this issue 😅

docs/CONTRIBUTING.md
Turbo87: thanks for the review! :)

I've added another commit on top with cross links to the relevant docs sections, which should hopefully make this clearer.

> Command "dev:local" not found

the issue here is that this "script" is defined in the `svelte` subfolder, so you'd need to `cd svelte` before trying that, or use `pnpm --dir svelte dev:local` from the root.

ehuss: Nice, thanks!  I got it working now.

PR #13654, opened 2026-05-11
test_db: Use `STRATEGY = FILE_COPY` for test database creation

Turbo87:
The default `WAL_LOG` strategy streams every block of the template database through WAL, which serializes heavily across concurrent `CREATE DATABASE` calls. `FILE_COPY` copies at the filesystem level and pays two checkpoints per call, but writes far less WAL and scales much better when many tests race to spin up test databases in parallel.

Cuts `cargo test --workspace` wall-clock by roughly a third on local runs.


Turbo87:
Doesn't seem like it makes a huge difference on CI unfortunately, but at least it's not getting worse, so I guess this is still worth doing 😅

PR #13656, opened 2026-05-11
Use `pgcrypto` for `random_string()` randomness

Turbo87:
The PostgreSQL `random()` function is explicitly documented as not cryptographically secure, but `random_string()` is still used to generate e-mail verification and crate owner invitation tokens.

Switch to `pgcrypto`'s `gen_random_bytes()` with rejection sampling to keep the output uniform over the existing 62-character alphabet.

### Related

- https://github.com/rust-lang/crates.io/pull/13487#issuecomment-4422369698
- Closes https://github.com/rust-lang/crates.io/pull/13487

### DIFF
--- a/migrations/2026-05-11-183755-0000_use-pgcrypto-for-random-string/down.sql
+++ b/migrations/2026-05-11-183755-0000_use-pgcrypto-for-random-string/down.sql
@@ -1,0 +1,11 @@
+CREATE OR REPLACE FUNCTION random_string(int4) RETURNS text AS $$
+  SELECT (array_to_string(array(
+    SELECT substr(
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):
I'm kinda twitchy about this, honestly — I'd like to get this out of the database entirely and into something we have more direct knowledge of, and I liked the token changes I had in #13487.

But I'm thoroughly drowning at this point and that feels very "perfect is the enemy of good enough", so let's merge this and I'll get back to the token changes in a future PR in 2034 when I might have some free time.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13656/a27b5eb67c17e623e79d9d6d280bc2ca8586ceb7..fcc4aa8efad645e9d96dd64bfcd358429babc9df)*

PR #13668, opened 2026-05-13
test_db: Switch from per-database to per-schema isolation

Turbo87:
The integration test suite isolates each test in its own Postgres database via `CREATE DATABASE <name> TEMPLATE <template>`. That call takes a cluster-wide lock that serializes under parallelism, costing up to ~2s per test under contention.

This PR replaces per-test databases with per-test **schemas** inside a single shared test database (`TEST_DATABASE_URL`). The cluster-wide lock is gone, and the suite gets roughly twice as fast on my machine.

Reviewing this commit-by-commit is recommended for your own sanity :D

## Local benchmark

613 tests, default parallelism, 3 runs each, fresh test database:

| Variant | median | mean |
|---|---|---|
| `CREATE DATABASE … TEMPLATE` (baseline) | 73.3s | 72.5s |
| Schema + captured DDL replay | **35.8s** | **36.4s** |

These numbers are from local runs only. CI performance has not been measured yet; merging this PR is what will let us see whether CI sees a similar improvement.

## How

1. Once per process, `Management::new()` ensures a `test_template` schema exists with all Diesel migrations applied, then captures its structure with `pg_dump --schema=test_template --inserts` and caches the SQL in memory.
2. For each `TestDatabase::new()`, the cached SQL is replayed (with the template schema name rewritten to a fresh `test_<rand16>`) via a single `batch_execute()`.
3. Per-test scoping rides on the connection URL itself as `options=--search_path=<schema>,public`. Anything that opens a connection from `TestDatabase::url()` (diesel pools, worker test pools, shelled-out `pg_dump`, ChaosProxy-rewritten URLs) is automatically scoped, with no changes to `DbPoolConfig` / `ConnectionConfig`.
4. On `Drop`: `DROP SCHEMA <name> CASCADE`. On startup: a sweep removes any `test_<rand16>` schemas left behind by crashed prior runs.

## Migration changes (no-ops on production, where they're already applied)

- Strip `public.` qualifiers from references to user-defined types/tables (`public.semver_triple`, `public.crate_owner_invitations`). The explicit qualifier would defeat the per-test schema rewrite.
- Add `SCHEMA public` to the three `CREATE EXTENSION IF NOT EXISTS …` calls so extensions consistently install in `public` regardless of the active `search_path`.

These edits are syntactically equivalent against an already-migrated database, so production is unaffected.

## Tests that shell out to `pg_dump`

Two tests run `pg_dump` against the test DB themselves and would otherwise race with concurrent tests' `CREATE`/`DROP SCHEMA`:

- `test_dump_db_job()` now enqueues `DumpDb::for_schema(app.db_schema())`, so the job's `pg_dump` runs with `--schema=<x>`.
- `dump_db_and_reimport_dump()` dumps with `Some(test_db.schema())`, drops the source schema mid-test, then replays the dump against the same URL. The `search_path` resolves again as soon as the dump's `CREATE SCHEMA` recreates it. No second `TestDatabase` is needed.

The job-side mechanism is a new optional `DumpDb::schema: Option<String>` field (`#[serde(default)]` so queued payloads still deserialize after deploy). Production callers use `DumpDb::default()` and continue to dump every schema as before.

### DIFF
--- a/migrations/20170308140537_create_to_semver_no_prerelease/up.sql
+++ b/migrations/20170308140537_create_to_semver_no_prerelease/up.sql
@@ -12,6 +12,7 @@ CREATE FUNCTION to_semver_no_prerelease(text) RETURNS semver_triple IMMUTABLE AS
   )::semver_triple
   WHERE strpos($1, '-') = 0
   $$ LANGUAGE SQL
+  SET search_path FROM CURRENT
 ;
 
 CREATE INDEX versions_to_semver_no_prerelease_idx ON versions (crate_id, to_semver_no_prerelease(num) DESC NULLS LAST) WHERE NOT yanked;
--- a/migrations/20170308140537_create_to_semver_no_prerelease/up.sql
+++ b/migrations/20170308140537_create_to_semver_no_prerelease/up.sql
@@ -9,7 +9,7 @@ CREATE FUNCTION to_semver_no_prerelease(text) RETURNS semver_triple IMMUTABLE AS
     split_part($1, '.', 1)::int4,
     split_part($1, '.', 2)::int4,
     split_part(split_part($1, '+', 1), '.', 3)::int4
-  )::public.semver_triple
+  )::semver_triple
   WHERE strpos($1, '-') = 0
   $$ LANGUAGE SQL
 ;
--- a/src/worker/jobs/downloads/update_metadata.rs
+++ b/src/worker/jobs/downloads/update_metadata.rs
@@ -348,8 +348,7 @@ mod tests {
         let (_, version) = crate_and_version(&mut conn, user.id).await;
 
         // Wrap the test body in a transaction so `now` resolves to the same
-        // value across every query inside it. Committing at the end is fine:
-        // the test schema is dropped when the test finishes.
+        // value across every query inside it.
         conn.transaction(async |conn| {
             update(versions::table)
                 .set(versions::updated_at.eq(now.into_sql::<Timestamptz>() - 2.days()))
--- a/src/worker/jobs/downloads/update_metadata.rs
+++ b/src/worker/jobs/downloads/update_metadata.rs
@@ -347,43 +347,44 @@ mod tests {
         let user = user(&mut conn).await;
         let (_, version) = crate_and_version(&mut conn, user.id).await;
 
-        // This test is using a transaction to ensure `now` is the same for all
-        // queries within this test.
-        conn.begin_test_transaction().await.unwrap();
+        // Wrap the test body in a transaction so `now` resolves to the same
+        // value across every query inside it. Committing at the end is fine:
+        // the test schema is dropped when the test finishes.
+        conn.transaction(async |conn| {
+            update(versions::table)
+                .set(versions::updated_at.eq(now.into_sql::<Timestamptz>() - 2.days()))
+                .execute(conn)
+                .await?;
+            update(crates::table)
+                .set(crates::updated_at.eq(now.into_sql::<Timestamptz>() - 2.days()))
+                .execute(conn)
+                .await?;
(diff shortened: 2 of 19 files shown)


Turbo87:
not quite as much of an improvement on CI unfortunately, but I assume that is due to the CI machines running on lower parallelism

migrations/20170308140537_create_to_semver_no_prerelease/up.sql
```diff
@@ -9,7 +9,7 @@ CREATE FUNCTION to_semver_no_prerelease(text) RETURNS semver_triple IMMUTABLE AS
     split_part($1, '.', 1)::int4,
     split_part($1, '.', 2)::int4,
     split_part(split_part($1, '+', 1), '.', 3)::int4
-  )::public.semver_triple
+  )::semver_triple
```
eth3lbert, line 12: I ran into the same error as #12371 on PG 17.7, although I'm not sure if this is expected since I'm not using a devcontainer.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13668/8ea8e72767f87f47b1a130881b2aeef873260bfa..46de59f07f3db4c34620104fc24c54fd185c2167)*

src/worker/jobs/downloads/update_metadata.rs
```diff
@@ -347,43 +347,44 @@ mod tests {
...
-            .await;
-        let crates_changed = crates::table
-            .select(crates::updated_at.ne(now.into_sql::<Timestamptz>() - 2.days()))
-            .get_result(&mut conn)
-            .await;
-        assert_eq!(versions_changed, Ok(false));
-        assert_eq!(crates_changed, Ok(false));
+        // Wrap the test body in a transaction so `now` resolves to the same
+        // value across every query inside it. Committing at the end is fine:
+        // the test schema is dropped when the test finishes.
```
eth3lbert: nit: This comment `// the test schema is dropped when the test finishes.` wasn't very clear to me when I was reviewing the code commit-by-commit 😅 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13668/8ea8e72767f87f47b1a130881b2aeef873260bfa..46de59f07f3db4c34620104fc24c54fd185c2167)*

migrations/20170308140537_create_to_semver_no_prerelease/up.sql
Turbo87: damn... that's unfortunate :-/

I'll see if I can find a workaround for this 🤔

src/worker/jobs/downloads/update_metadata.rs
Turbo87: should we just remove that sentence?

eth3lbert: yeah, I think it's totally fine to remove it!

migrations/20170308140537_create_to_semver_no_prerelease/up.sql
Turbo87: I've found a workaround via `SET search_path FROM CURRENT` on the function. the function no longer exists in the database today, so this is unlikely to cause any issues in production, but should unblock this PR from getting used with pg17+

Turbo87: Adam was running into the same issue and confirmed that the fix worked for him. Also:

> re-ran the tests with the new changes. Went from about 3:30 to about 2:15

🎉

eth3lbert: I've tried this and it works flawlessly, or at least it does locally on my end.

PR #13672, opened 2026-05-14
Add a quantum computing category

AbeCoull:
The need arises from a project I am working on within the quantum space. Given quantum computing's increasing Rust ecosystem, adding this as a category would be beneficial to discovery. ~~If a more analytical look at a rough estimate of the number of packages that fall into this category is needed to warrant this change, I can gather that information.~~

---
The main motivation for adding this is that the quantum tooling ecosystem has started to take its own unique identity. It has robust simulations for quantum circuits, intermediate representations and parsers, client crates for interaction with the quantum devices, the list goes on. Quantum computing differs from existing categories as it represents a different way of thinking about and interacting programmatically with the hardware. Easier grouping and discoverability of the tools with a subcategory will allow better differentiation of the field. With more active areas of research, such as error correction, an increase of novel tools would be benefited by this aided discoverability.

There are over 250 crates that appear with the `#quantum-computing` or `#quantum` tag.
* http://crates.io/keywords/quantum
* http://crates.io/keywords/quantum-computing

Some of the more popular packages within those:
* https://crates.io/crates/quil-rs - more than 500,000 downloads. Used for parsing a quantum instruction language
* https://crates.io/crates/hugr - more than 145,000 downloads. A representation package for quantum circuits

### DIFF
--- a/src/boot/categories.toml
+++ b/src/boot/categories.toml
@@ -618,6 +618,14 @@ description = """
 Crates for processing biological sequences, including alignment, assembly, and annotation.
 """
 
+[science.categories.quantum-computing]
+name = "Quantum Computing"
+description = """
+Crates for quantum computing, including circuit construction,
+simulation, quantum algorithms, intermediate representations, and
+interfaces to quantum hardware backends.
+"""
+
 [security]
 name = "Security"
 description = """


Turbo87:
Hi @AbeCoull! did you see https://github.com/rust-lang/crates.io/blob/main/docs/categories.md? :)
 
particularly this part:

> For new categories, it's helpful to note in your PR description examples of crates that would fit in that category, and describe what distinguishes the new category from existing categories.

AbeCoull:
Hi @Turbo87, I apologize as I did miss that. The PR description is now updated.

Turbo87 (approved):
thanks, that seems reasonable :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13672/904e1085441161ace180148bc92d7a35ad4f0ef5..651d2232d92c3070b6bed46d6eccadd6bf2b6c5c)*

Issue #13720, opened 2026-05-20
Unable to receive email verification email


kwunlokng:
### Current Behavior

I kept getting error "A verified email address is required to publish crates to crates.io. Visit https://crates.io/settings/profile to set and verify your email address." but I cannot receive any emails from crates.io even after hitting resend, not even in spam.

### Expected Behavior

I should be either receiving email verification link or have a way to verify it without the resend.

### Environment

- Browser:
- OS:

Turbo87:
@rust-lang/infra could you check the mailgun logs? I see two `PUT /api/v1/users/396482/resend` requests in the Datadog logs, but both responded with `200 OK`, so the issue appears to be downstream. I don't have mailgun credentials, so I can't debug this any further.

ubiratansoares:
Hi had a look on logs, and that's is what I found

> No MX for fiireinbelly.com, lookup fiireinbelly.com. on .1:53: no such host"

Turbo87:
since the domain does not appear to exist, I assume that this is a case of a typo in the email address.

PR #13721, opened 2026-05-20
email: Log outcome of `Emails::send()`

Turbo87:
Adds tracing on `Emails::send()` so delivery problems can be debugged from Datadog. Each send emits an `info!` on success or a `warn!` on transport failure, with three structured fields:

* `email.message_id`: the join key against Mailgun's delivery logs
* `email.template`: the template that was sent (e.g. `user_confirm`)
* `email.recipient_domain`: useful as a facet to spot provider-wide problems

The full recipient address is kept out of the log on purpose. The Message-ID alone is enough to correlate with Mailgun (which retains the full address on its side), so duplicating it into Datadog would just be extra PII exposure.

### Related

- https://github.com/rust-lang/crates.io/issues/13720


LawnGnome (commented):
I understand the motivation behind wanting the domain, but it could still be PII for people with vanity domains.[^vanity] Maybe we could hash them so we have a way of enumerating, say, gmail.com, but can't trivially reverse back to myfulllegal.name?

Otherwise, LGTM.

[^vanity]: Hi, it's me, I'm the problem, it's me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13721/30e47d8c3d4d4a0bbf7ceebf951a61297fd91142..d83bb359fb6015872ef36afa232122d5c63d29f7)*

Turbo87:
I considered hashing the email address, but got LLM pushback telling me that it would also be easy with publicly leaked datasets to create the same hash and associate the PII that way. I don't have a strong opinion on this though. 😅

LawnGnome:
I think it's more about increasing the degree of difficulty — having straight up `adamharvey.name` in Datadog would be a problem in the case of a GDPR erasure request, but `ebf545d5875389ffdc81d9d37d7801297c1dd135b49dbe0e0d7e791942ec1d6a` feels like we'd be compliant (given it would at least be non-trivial to work backwards to a domain) without needing to figure out a way to scrub the logs, while still letting us correlate domains, at least.

Turbo87:
In that case, hash of the full address or only the domain portion?

LawnGnome:
> In that case, hash of the full address or only the domain portion?

I think the domain, since it lets us do the correlation and minimises the PII concerns.

Turbo87:
I've implemented the domain hashing in the structured field. I've implemented masking for the email address in the log message though. This seems significantly more useful to me than a) logging a hash or b) not logging the email address at all. from what I've found online so far, this level of masking should be sufficient to address GDPR concerns.

LawnGnome (approved):
Yep, that looks good to me as well. Nice work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13721/bac415bc0029198391c7459495a98942cc8d4569..cf8b27389162caf23b7dc75d35847e3f19d9b1cf)*

PR #13724, opened 2026-05-21
Replace Mozilla CDN with Fontsource packages

### LINKED ISSUE #13454: External resources on `code.cdn.mozilla.net` served only over legacy IPv4
### Current Behavior

External resources loaded from `code.cdn.mozilla.net` are only available over IPv4 and could cause issues with IPv6-only clients.

### Expected Behavior

All external resources should also be reachable over IPv6 for full compatibility.

### Steps To Reproduce

1. Load crates.io and notice external resources loaded from `code.cdn.mozilla.net` (a subdomain with only IPv4 A records)

### Environment

- Browser: Firefox 149.0.2
- OS: Linux Mint 22.3

### Anything else?

_No response_

Turbo87:
The `mozilla/Fira` repository was archived on 2026-05-13 and `code.cdn.mozilla.net` has uncertain long-term availability. The CDN is also only reachable over IPv4, which breaks IPv6-only clients.

This switches to `@fontsource/fira-sans` and `@fontsource/fira-mono` (both at 5.2.7) and imports the 8 weights currently used in the codebase. Vite bundles the WOFF2 and WOFF files into the static asset directory with content hashes, adding around 1.9 MB to the build output.

The `code.cdn.mozilla.net` entry is dropped from the CSP `style-src`, and the now-redundant `font-src` directive is removed since `default-src: 'self'` covers it.

Resolves #13454

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -134,6 +134,12 @@ importers:
       '@floating-ui/dom':
         specifier: 1.7.6
         version: 1.7.6
+      '@fontsource/fira-mono':
+        specifier: 5.2.7
+        version: 5.2.7
+      '@fontsource/fira-sans':
+        specifier: 5.2.7
+        version: 5.2.7
       chart.js:
         specifier: 4.5.1
         version: 4.5.1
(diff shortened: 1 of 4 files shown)


LawnGnome (commented):
I'm fine with this in general, but I do have a question: will the build output have the same cacheability as the Mozilla CDN — that is, does a user have to redownload the font files each time we rebuild and deploy?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13724/9078db5b7f1cb3374dacb437a461afe86232d72d..7baafb2d2fc429933142396611cebbac509de147)*

Turbo87:
> that is, does a user have to redownload the font files each time we rebuild and deploy?

no, Vite should generate the same exact files, with the same content hash in their file names, unless the font files actually change. the files in `/_app/immutable/` should all be served with the corresponding long-term cache headers.

LawnGnome (approved):

PR #13726, opened 2026-05-21
Add AI tools policy

Turbo87:
This PR proposes adopting the CPython devguide's AI tools policy as crates.io's own, as `docs/AI-TOOLS.md` (with one Python-specific testing sentence dropped). It is linked from `CONTRIBUTING.md` and `AGENTS.md`.

The CPython policy puts responsibility for content on the submitter, requires contributors to be able to explain their changes, and doesn't try to ban AI tooling outright. That matches how things are already done in this project and seems a reasonable starting point.

Project-wide AI policy for rust-lang is still being debated in two open RFCs (neither close to consensus), with a separate `rust-lang/rust`-scoped policy that explicitly excludes crates.io. If a project-wide policy is later adopted, we can align with it.

Open to feedback on whether this is the right policy and whether anything should be adjusted before merging.

### Related

- https://github.com/python/devguide/pull/1778
- https://github.com/rust-lang/rfcs/pull/3950
- https://github.com/rust-lang/rfcs/pull/3959
- https://github.com/rust-lang/rust-forge/pull/1040

### DIFF
--- a/docs/AI-TOOLS.md
+++ b/docs/AI-TOOLS.md
@@ -30,7 +30,6 @@ mind for the quality of your contribution:
   tests may be ensuring specific API behaviors are maintained.
 
 Pay close attention to AI generated recommendations for testing changes.
-Provide input about Python's testing principles when guiding an AI model.
 Always review the output before opening a pull request or issue,
 including proposed PR or issue titles and descriptions.
 
--- a/docs/AI-TOOLS.md
+++ b/docs/AI-TOOLS.md
@@ -1,0 +1,56 @@
+# Guidelines for using AI tools
+
+The person submitting an issue or PR is responsible for its content,
+regardless of whether AI tools were used in its creation. Generative AI
+tools can produce output quickly, but discretion, good judgment, and
+critical thinking are the foundation of all good contributions. We value
+good code, concise accurate documentation, and well scoped PRs without
+unneeded code churn.
+
+## Considerations for success
+
+Authors must review the work done by AI tooling in detail to ensure it
+actually makes sense before proposing it as a PR or filing it as an issue.
+
+We expect PR authors and those filing issues to be able to explain their
+proposed changes in their own words.
+
+Disclosure of the use of AI tools in the PR description is appreciated,
+while not required. Be prepared to explain how the tool was used and what
+changes it made.
+
+Whether you are using AI tools or not, keep the following principles in
+mind for the quality of your contribution:
+
+- Consider whether the change is necessary
+- Make minimal, focused changes
+- Follow existing coding style and patterns
+- Write tests that exercise the change
+- Keep backwards compatibility with prior releases in mind. Existing
+  tests may be ensuring specific API behaviors are maintained.
+
+Pay close attention to AI generated recommendations for testing changes.
+Provide input about Python's testing principles when guiding an AI model.
+Always review the output before opening a pull request or issue,
+including proposed PR or issue titles and descriptions.
+
+## Acceptable uses
+
+Some of the acceptable uses of generative AI include:
+
+- Assistance with writing comments, especially in a non-native language
+- Gaining understanding of existing code
+- Supplementing contributor knowledge for code, tests, and documentation
+
+## Unacceptable uses
+
+Maintainers may close issues and PRs that are not useful or productive,
+regardless of whether AI tools were used or not.
+
+If a contributor repeatedly opens unproductive issues or PRs, they may be
+blocked from contributing to the project because it is disruptive and
+disrespectful of the maintainers time.
+
+It is not acceptable to alter or bypass existing tests, or remove desired
+functionality, in order to make a failing test pass. Such changes are not
+a real fix.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -221,5 +221,6 @@ Before submitting:
 - `/docs/ARCHITECTURE.md` - System architecture and design decisions
 - `/docs/API-DESIGN.md` - JSON API conventions for new endpoints
 - `/docs/PR-REVIEW.md` - Guidelines for reviewing pull requests
+- `/docs/AI-TOOLS.md` - Guidelines for using AI tools when contributing
 - `/script/import-database-dump.sh` - Import production database dump for testing
 - `/.github/workflows/ci.yml` - CI pipeline definition and tool versions
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,6 +1,7 @@
 # Contributing to crates.io
 
 - [Attending the weekly team meetings](#attending-the-weekly-team-meetings)
+- [Using AI tools](#using-ai-tools)
 - [Finding an issue to work on](#finding-an-issue-to-work-on)
 - [Submitting a Pull Request](#submitting-a-pull-request)
 - [Reviewing Pull Requests](#reviewing-pull-requests)
@@ -16,6 +17,11 @@ on Zoom or [Zulip] (`#t-crates-io`) for our weekly team meeting, and we invite
 everyone who wants to contribute to crates.io to participate.
 
 [Zulip]: https://rust-lang.zulipchat.com/#narrow/stream/318791-t-crates-io/
+
+## Using AI tools
+
+If you use AI tools to help draft issues, PRs, or comments, please
+review [`docs/AI-TOOLS.md`](AI-TOOLS.md) for our guidelines.
 
 ## Finding an issue to work on
 


carols10cents (approved):

docs/AI-TOOLS.md
```diff
@@ -0,0 +1,55 @@
...
+
+Some of the acceptable uses of generative AI include:
+
+- Assistance with writing comments, especially in a non-native language
+- Gaining understanding of existing code
+- Supplementing contributor knowledge for code, tests, and documentation
+
+## Unacceptable uses
+
+Maintainers may close issues and PRs that are not useful or productive,
```
clarfonthey, line 46: So, I think it would be worth clarifying a bit more what "productive" means here, since it feels a little vague and it's not exactly clear what you're looking for. (I'm fine if you keep using the term "productive" as long as it's defined, just, maybe a few examples would help.)

Just guessing at what you mean, what might be worth mentioning:

* Typo fixes/aesthetic improvements that don't actually affect any public stuff: probably not helpful, can be submitted in large quantities by these tools
* Poorly justified changes, e.g. "I updated this page to be more readable" without any reasonable explanation
* Careless changes / author just wasn't paying attention and just clicked submit when they shouldn't have

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13726/9078db5b7f1cb3374dacb437a461afe86232d72d..7a2230e72d2419ae308ed837b650f22092493b57)*

docs/AI-TOOLS.md
```diff
@@ -0,0 +1,55 @@
...
+
+Authors must review the work done by AI tooling in detail to ensure it
+actually makes sense before proposing it as a PR or filing it as an issue.
+
+We expect PR authors and those filing issues to be able to explain their
+proposed changes in their own words.
+
+Disclosure of the use of AI tools in the PR description is appreciated,
+while not required. Be prepared to explain how the tool was used and what
+changes it made.
```
clarfonthey, line 20: So ultimately I don't expect crates.io to have a lot of PRs or issues so I think that ultimately this is just me giving advice that you're free to ignore. But at least from what I found, even projects which strongly support AI tools have been in favour of disclosing just so they know what they're dealing with, and I think that it's okay to have a hard rule with no punishments for forgetting it.

Like, I figure the rate of issues/PRs is so low that you have the time/resources to just individually talk to each person and figure out what they're working with anyway, but it can be helpful to ask for to see what people are using.

Just a few examples of other policies that feel relevant here:

* [curl](https://curl.se/dev/contribute.html#on-ai-use-in-curl) explicitly asks for AI disclosure in security reports
* [ghostty](https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md#ai-is-welcome-here) explicitly states "AI is welcome here" but still requests disclosure in basically the first bullet point

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13726/9078db5b7f1cb3374dacb437a461afe86232d72d..7a2230e72d2419ae308ed837b650f22092493b57)*

clarfonthey:
This is something I totally didn't even consider until now, but are there any other public repos managed by the crates.io team or is everything besides this just like, private infra accounts for controlling the actual deployed resources? I mostly ask since, this makes the most sense to just say "this is the policy of the crates.io team," but I don't actually know if there's anything else that would be controlled by you folks.

docs/AI-TOOLS.md
LawnGnome: > But at least from what I found, even projects which strongly support AI tools have been in favour of disclosing just so they know what they're dealing with, and I think that it's okay to have a hard rule with no punishments for forgetting it.

This would also be my preference.

I litigated this at some length within the Foundation when we were drafting our internal AI policy, so my apologies to @Turbo87 for the repeat, but the short version is that I believe that we should require disclosure for two primary reasons:

1. Legal concerns. While I suspect our US-focused legal advice that LLM contributions are not subject to copyright will eventually end up being codified into international copyright law, this isn't even fully settled in the US, and we also have to operate in other, slower moving jurisdictions. If this does not end up being the result, we will need to be able to account for the origin of every contribution. Even if that outcome is unlikely, I think the minimal cost of requiring disclosure now is a small price to pay to make it easier to handle that eventuality.
2. Review practicalities. While I don't think knowing LLMs have been used would meaningfully change how I review PRs from within the crates.io team, we do accept external contributions, and the sorts of things I expect to look for in review would change for a developer who used LLM assistance versus not, as would how I communicate with that contributor.

---

I don't want this to be some onerous process. I'd be happy with an `Assisted-By` or `Co-Authored-By` trailer in either the commit message or the PR description, much like [the Linux kernel policy](https://docs.kernel.org/process/coding-assistants.html#attribution). But I do think it's important that we require this.

LawnGnome (commented):
Approved is probably a strong word here, so I won't use it.

I have significant ethical concerns with the current state-of-the-art with regards to LLM assistants, and will not use them myself in any generative context. However, that is my own boundary, and I accept that everyone gets to make their own choice.

As discussed below, I think we should require disclosure. With that change, I would accept this policy.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13726/9078db5b7f1cb3374dacb437a461afe86232d72d..7a2230e72d2419ae308ed837b650f22092493b57)*

LawnGnome:
> This is something I totally didn't even consider until now, but are there any other public repos managed by the crates.io team or is everything besides this just like, private infra accounts for controlling the actual deployed resources? I mostly ask since, this makes the most sense to just say "this is the policy of the crates.io team," but I don't actually know if there's anything else that would be controlled by you folks.

That's a good call out, but in this case, this repo is the only public repo owned by the crates.io team.

I expect this policy — if approved — would also apply to our one private repo (our ops guide), but I'm not sure it's necessary to change this to explicitly include that. :shrug:

docs/AI-TOOLS.md
traviscross: It would be interesting to hear from counsel on the first point.  I would expect to hear a mixed set of pros and cons for each choice — widespread disclosure seems likely to carry some risks too for the Project.  Regarding the second point, the relevant comment in the Python discussions was [this one](https://discuss.python.org/t/i-am-concerned-about-llm-code-in-python/106691/82) — in short, reviewers need to assume LLM use anyway.

Disclosure has more than minimal cost:

- [The transparency dilemma](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=5205850) finding includes considerable data showing that, contrary to intuition, disclosure of LLM use erodes trust.
- Once primed about LLM use, people start to see LLM fingerprints that aren't actually there — we're creatures that see faces in clouds.  Disclosure can increase the risk of false accusations of more LLM use than was disclosed.
- We're operating in an environment where people are actively shaming others who use LLMs.  Highlighting one's use subjects one to the risk of this shaming.

Mandating disclosure has additional costs.  It becomes something we need to enforce fairly and consistently.  As part of a moderation policy, we'd be talking about banning people from the Project over this.  If it turns out the policy is unenforceable and that people choose not to disclose due to these costs, then we'll have created a kind of policy fiction.  That's not any good for us.

Similar to Python, we can encourage disclosure without incurring the costs that accompany mandating it.

clarfonthey:
Yeah, I figure that this kind of policy doesn't need to be clarified for private repos considering how it's mostly aimed at contributors who aren't on the team and thus don't have expectations set yet. So, with this being the only public crate for the repo *and* the repo receiving such a small amount of traffic externally as-is this policy is more than sufficient.

I don't know if there are any crates.io team members who aren't Foundation staff but I feel that regardless, just deferring to Foundation policy for internal stuff would be sufficient.

docs/AI-TOOLS.md
clarfonthey: So, it is worth reiterating on the legal point that generally the result of using LLMs is the *removal* of copyright protections, not the addition of extra burdens, which does not matter given the extremely permissive license crates.io uses. It *does* matter for proprietary code and copyleft code, but for open source, permissively licensed work, it seems fine.

Additionally, it seems incredibly likely that if any copyright questions came up, e.g. someone was upset that code verbatim was used without authorisation, there would be ample time to fix the issue (removing the offending code) rather than these being immediate repercussions.

Most of the projects that I've seen be concerned with copyright issues are copyleft and thus specifically rely on copyright to enforce licencing, which just isn't the case here. So, while I do think there are other points to discuss, I don't think the legal point is particularly compelling here, which also matches the existing advice given by Foundation counsel on the matter.

Turbo87: I think I do most of the PR reviews on crates.io at the moment (obviously excluding my own PRs…), and knowing whether AI was involved wouldn't change much how I review. The "explain your changes" requirement is already what I actually care about. For first-time contributors with zero context I agree disclosure can help, but that's a pretty different situation from someone opening a dozens of PRs each month.

On the legal point: the PSF landed on this exact wording without requiring disclosure. I don't know what their counsel actually said about it, so I'm only guessing, but I'd be a little surprised if a project that size shipped an AI policy without running it past their lawyers at some point. If that's roughly right, it makes me want to understand what's specifically different about our situation before treating the legal angle as a strong reason to go further than they did.

And I think Travis is right that disclosure has real downsides in the current climate. Adding a trailer to every PR also adds quite a bit of friction when LLM tooling is opening them, and getting that tooling to reliably add trailers is a whole separate problem.

That said, here are a few things I'd be happy to put in the policy, if it helps:

- Bump "appreciated, while not required" to "encouraged, especially from new contributors"
- Add something like "reviewers may ask about AI usage when it would help, and contributors should answer honestly"

Two other options to avoid the additional friction for regular contributors:

1. We can flip the framing: say explicitly that reviewers should assume AI may have been used, and make disclosure of *non-use* the optional opt-in. At this point the vast majority of recent crates.io PRs had LLM involvement and IMHO it's better to optimize for the 90% case.

2. One-time disclosure for regulars, recorded privately: contributors who use AI tooling regularly disclose it once (e.g. via a comment on a designated issue in the team's private ops guide repo) instead of repeating it on every PR. The record exists, so reviewers on the team have access and the legal/review-practicalities angle is covered, but the disclosure isn't sitting in a public list that anyone can scrape and turn into a target. For someone like me who opens a lot of PRs that drops the per-PR friction effectively to zero while still leaving the information available where it matters. First-time or external contributors would still get asked by a PR template or maybe by the reviewer.

docs/AI-TOOLS.md
Turbo87: Thanks, I see where you're coming from but I'd actually prefer to leave "productive" deliberately vague.

To be honest I'm not sure all your examples land for me. Typo fixes for instance often seem fine. The kind of thing I'd see as unproductive is more like a PR where the author clearly doesn't understand what the LLM produced and is just acting as a transport medium between the model and the reviewer. But that's just one example, and I'd really rather leave it to maintainer judgment than enumerate it.

My worry with putting cases in the policy is that they get read as an exhaustive list. A PR that doesn't fit any listed category becomes harder to push back on ("but the policy doesn't say this counts"), and we'd be locking ourselves into whatever taxonomy we write down today. The point of this section, as I read it, is to give maintainers cover to use judgment on cases that don't fit cleanly into rules.

Happy to revisit if it turns out to be confusing in practice.

Turbo87:
> That's a good call out, but in this case, this repo is the only public repo owned by the crates.io team.

Quick correction on the scope claim: there are actually a few other public repos in the crates.io team's orbit. The authoritative list is in the team-api (https://team-api.infra.rust-lang.org/v1/repos.json), but off the top of my head:

- `rust-lang/crates_io_og_image`, owned by the crates.io team, used for OG image rendering
- `rust-lang/crates-io-auth-action`, shared with infra I think
- `rust-lang/crates-io-heroku-metrics`, ownership a bit unclear, possibly infra
- `rust-lang/crates.io-index` and the related index repos, but the AI policy probably isn't very relevant there

For the repos where the policy is actually relevant (`crates_io_og_image` in particular), my preference would be to have the policy explicitly cover them rather than copying it into each repo. Maybe a sentence at the end of the policy like:

> This policy applies to all repositories owned by the crates.io team.

That way we don't have to manage copies, and as the team's repo list changes the policy follows automatically.

docs/AI-TOOLS.md
clarfonthey: I mean, ultimately, I do think that the small number of contributions is more than sufficient to prefer an ad hoc method of explaining/disclosing since, like I said, you probably have the bandwidth to just talk to everyone individually about their changes. I think that encouraging more transparency is good, but if you think the downsides outweigh the upsides here, that's fair.

clarfonthey:
Specifically for `crates.io-index` and co., I know that GitHub allows disabling both issues and PRs now so I would just disable both for the repo and any relevant related ones. It's not really a project to contribute to, just a dataset.

Might poke around later to see if the current infra we have allows automatically doing this via the `team` repo, since it is a new feature.

---

Edit: It appears that this is done manually: https://github.com/rust-lang/team/blob/main/repos/rust-lang/crates.io-index.toml

So, I'll open a thread on Zulip about it.

docs/AI-TOOLS.md
Turbo87: FYI for those that were not there: we discussed this policy proposal and the disclosure thread in particular in the team meeting last friday.

@rust-lang/crates-io I'm still open to feedback to my last comment in this thread. if I don't hear anything by the end of the week I'll assume that there are no objections to merging this as-is.

apiraino: In the opening comment here I read:
> Project-wide AI policy for rust-lang is still being debated [...] ]If a project-wide policy is later adopted, we can align with it.

The policy discussed here differs in many parts from [rust-forge#1040](https://github.com/rust-lang/rust-forge/pull/1040/) (for example the disclosure policy) and I appreciate you want something out sooner.

But I want to raise a more general question: this policy is more lenient than the other one, so I *think* that if you want to adopt [rust-forge#1040](https://github.com/rust-lang/rust-forge/pull/1040/), you might find yourself deciding to backpedal on some items that are now allowed (or not mentioned), like the [disclosure clause](https://github.com/rust-lang/rust-forge/pull/1040/files#diff-2d95c811a5fd15552f76562ec278135a5cc9ac9579df3a8ba53e6b7de8d1d855R42) which we feel strongly about and many others, like bots reviews, typos fixes, etc.

How does the t-crates.io team feels when (or if) comes the question about adopting the rust-lang/rust policy?

Turbo87: https://github.com/rust-lang/rust-forge/pull/1040/ is specifically only about rust-lang/rust and not about other subprojects/repositories. if that policy should influence the policies of other subprojects then that discussion needs to happen in a Project-wide RFC with buy-in from all teams.

apiraino: OK that partially answers my question, I'll try to fill in the blanks (correct me if I am wrong): you would not adopt [rust-forge#1040](https://github.com/rust-lang/rust-forge/pull/1040) if proposed to become a project-wide policy as it is, correct?

Turbo87: I honestly haven't followed the forge policy that closely explicitly because it was out of scope regarding crates.io. I have not checked what the current state is and where it might be incompatible with what we're doing on this side of the project. I was also not aware that there was even a discussion of this becoming a project-wide policy.

clarfonthey: I also was under the impression that the forge policy was explicitly not proposed as project-wide because it would then require the buy-in of all the teams, many of whom manage their repos differently (case in point: here). If a project-wide policy is adopted, team-wide policy can change to accommodate it, but for now, project-wide policy isn't adopted, so, I think it's totally fine for teams to adopt any policy since it helps remove ambiguity in enforcement.

PR #13759, opened 2026-05-26
Require `C` database collation for `semver_ord`

### LINKED ISSUE #13749: `semver_ord` produces non-semver ordering on databases with a non-`C` collation
### Current Behavior

On a database with a non-`C` collation, `semver_ord` orders prerelease identifiers wrong:
```
1.0.0-B > 1.0.0-alpha
```

Per spec:
> Identifiers with letters or hyphens are compared lexically in ASCII sort order.

`B` (0x42) sorts before `a` (0x61), so `1.0.0-B` should come before `1.0.0-alpha`, not after.

The official `postgres` image defaults to `en_US.utf8` (`docker run --rm postgres:16 printenv LANG`), so the devcontainer hits it too. `calculate_default_version` orders by `semver_ord`, so `default_version` can be wrong on any non-`C` deployment.

### Expected Behavior

`semver_ord` orders by ASCII regardless of the database collation:
```
1.0.0-alpha > 1.0.0-B
```

### Steps To Reproduce

Add an uppercase test case. `test_spec_order` only uses lowercase identifiers today, which sort the same under `C` and `en_US`, so it does not catch this:

```diff
diff --git a/crates/crates_io_database/tests/semver_ord.rs b/crates/crates_io_database/tests/semver_ord.rs
index 0a5dbadc1..607685a15 100644
--- a/crates/crates_io_database/tests/semver_ord.rs
+++ b/crates/crates_io_database/tests/semver_ord.rs
@@ -75,7 +75,8 @@ async fn test_spec_order() {
                     '1.0.0-beta.2',
                     '1.0.0-alpha.1',
                     '1.0.0-alpha.beta',
-                    '1.0.0-beta.11'
+                    '1.0.0-beta.11',
+                    '1.0.0-B'
                 ]) as num
             )
             select num
@@ -92,6 +93,7 @@ async fn test_spec_order() {
 
     insta::assert_debug_snapshot!(check("asc").await, @r"
     [
+        1.0.0-B,
         1.0.0-alpha,
         1.0.0-alpha.1,
         1.0.0-alpha.beta,
@@ -113,6 +115,7 @@ async fn test_spec_order() {
         1.0.0-alpha.beta,
         1.0.0-alpha.1,
         1.0.0-alpha,
+        1.0.0-B,
     ]
     ");
 }
```

```bash
export TEST_DATABASE_URL="postgres://postgres@localhost:55432/test"
```

Failing:
```bash
docker run --rm -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_DB=test -p 55432:5432 postgres:17
cargo test -p crates_io_database --test semver_ord test_spec_order
```

Works only with explicit `C`-collation:
```bash
docker run --rm -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_DB=test -e POSTGRES_INITDB_ARGS=--lc-collate=C -p 55432:5432 postgres:17
cargo test -p crates_io_database --test semver_ord test_spec_order
```

### Environment

n/a

### Anything else?

* The devcontainer should set up postgres with explicit `C` collation (`POSTGRES_INITDB_ARGS=--lc-collate=C`).
* `test_spec_order` should include an uppercase vs lowercase case.
* Ideally `semver_ord` should be collation independent.
* If `C` stays a requirement, `semver_ord` should document it.

Turbo87:
The `semver_ord()` function encodes prerelease identifiers as JSONB strings and orders versions via JSONB comparison. PostgreSQL compares JSONB strings using the database's default collation, so on any non-`C` collation (for example the `en_US.UTF-8` default of the official postgres image) the ASCII sort order required by SemVer spec item 11 is violated and e.g. `1.0.0-B` sorts after `1.0.0-alpha` instead of before it.

Production already runs on `C` collation, so this is currently invisible there, but local dev, the devcontainer, and contributors using the default postgres locale all hit it.

This pins the devcontainer's postgres to `C`, adds an uppercase identifier to `test_spec_order` so the bug surfaces in tests, adds `test_database_collation_is_c` which reports a misconfigured database with the exact `createdb` invocation to recreate it, and documents the requirement in `docs/CONTRIBUTING.md` and `AGENTS.md`.

/cc @danilobuerger

### Related

- Resolves https://github.com/rust-lang/crates.io/issues/13749

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -24,6 +24,10 @@ services:
     image: postgres:16
     environment:
       POSTGRES_DB: cargo_registry
+      # `semver_ord` stores prerelease identifiers as JSONB strings and
+      # relies on byte-wise ordering. Force `C` collation so JSONB string
+      # comparison matches the SemVer spec's ASCII sort order.
+      POSTGRES_INITDB_ARGS: '--lc-collate=C --lc-ctype=C'
       # Trust auth is fine because Postgres is only reachable on the
       # internal Docker network; no port is published to the host.
       POSTGRES_HOST_AUTH_METHOD: trust
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
 
     env:
       RUST_BACKTRACE: 1
-      TEST_DATABASE_URL: postgres://postgres:postgres@localhost/postgres
+      TEST_DATABASE_URL: postgres://postgres:postgres@localhost/cargo_registry_test
       RUSTFLAGS: '-D warnings -Cinstrument-coverage'
       MALLOC_CONF: 'background_thread:true,abort_conf:true,abort:true,junk:true'
 
@@ -178,6 +178,7 @@ jobs:
 
       - run: sudo systemctl start postgresql.service
       - run: sudo -u postgres psql -c "ALTER USER postgres WITH PASSWORD 'postgres'"
+      - run: sudo -u postgres createdb --lc-collate=C --lc-ctype=C -T template0 cargo_registry_test
 
       - run: cargo fetch --locked
       - run: cargo build --tests --workspace
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -94,7 +94,7 @@ cargo fmt --all --check                                # Formatting
 cargo clippy --all-targets --all-features --workspace  # Linting
 ```
 
-Test database setup: Set `TEST_DATABASE_URL` in `.env` to a separate database (e.g., `postgres://postgres@localhost/cargo_registry_test`). The test harness creates isolated databases and runs migrations automatically. Create the base test database once with `createdb cargo_registry_test`.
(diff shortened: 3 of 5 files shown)


LawnGnome (approved):
Let's party like it's 1979!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13759/6a3ac659fdd64bcb8eb792fcc2abbc3003e66891..f3e0ca394bdecb3c1694fdec9a94f3df8f27b1f7)*

Turbo87:
I'll keep this open for a bit to see if the `bytea` idea in https://github.com/rust-lang/crates.io/issues/13749 will lead somewhere. if we move to that then we don't need the collation changes at all.

eth3lbert (approved):
Both approaches LGTM! Since #13749 adds a `$` at the end of the regex to handle invalid cases like '1.0.0-' nicely, we can include it in `semver_ord` if we choose to only merge this PR. That said, it also makes sense to merge #13749 and this to ensure dev DB collation aligns with production, since we might hit other collation issues at some point in the future.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13759/6a3ac659fdd64bcb8eb792fcc2abbc3003e66891..f3e0ca394bdecb3c1694fdec9a94f3df8f27b1f7)*

Turbo87:
> That said, it also makes sense to merge #13749 and this to ensure dev DB collation aligns with production, since we might hit other collation issues at some point in the future.

alright, you convinced me :D

PR #13775, opened 2026-05-26
Add `semver_ord_v2` bytea sort key independent of database collation

danilobuerger:
Replaces the JSONB-based `semver_ord` with a bytea key that compares byte-wise, so ORDER BY matches semver precedence regardless of the database collation. The old key sorts `1.0.0-B` after `1.0.0-alpha` on non-`C` collations like en_US.utf8 (production runs on `C` today, but the devcontainer and other deployments that inherit the default `postgres` image collation do not).

Layout: enc(major) enc(minor) enc(patch) <prerelease>, where enc() is length-prefixed: [digit_count][ASCII digits]. The post-patch byte is 0x00 (prerelease end) < 0x01 (numeric ident) < 0x02 (alpha ident) < 0x03 (no prerelease), giving "fewer fields < more fields", "numeric < alpha", and "prerelease < release". Numeric idents are arbitrary precision up to 255 digits, removing the 10-part prerelease cap of the old jsonb encoding.

First step of the rollout discussed in #13749: the new column is nullable and filled by a trigger on insert. Backfill of existing rows will be done manually in batches; the old column and its readers stay in place until then.


migrations/2026-05-26-120000-0000_semver_ord_v2/up.sql
```diff
@@ -0,0 +1,104 @@
...
+-- over.
+
+-- Encode a semver numeric identifier as [digit_count][ASCII digits]. Byte order
+-- == numeric order: with no leading zeros, more digits means a larger number so
+-- the count byte dominates, and equal-length values compare digit-wise (ASCII
+-- order == numeric order for same-length decimals). It is arbitrary-precision up
+-- to 255 digits (the count byte's range).
+create or replace function semver_ord_num(digits text) returns bytea as $$
+    select set_byte('\x00'::bytea, 0, length(digits)) || convert_to(digits, 'UTF8');
+$$ language 'sql' immutable parallel safe strict;
```
Turbo87, line 18: would it be viable to encode it as a big-endian u64? the `semver` crate does not support more than u64 anyway, so that would be sufficient. then again, the majority of cases is probably single-digit which would be 2 bytes vs the 4 bytes for the u64. I'm not sure what is better 😅 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13775/6a3ac659fdd64bcb8eb792fcc2abbc3003e66891..a037e0ddf27c04b2782071bf1a7229e82cc3c4cb)*

danilobuerger: So this for sure could be optimized. My thinking was to use a variable padding as that will usually result in 2-3 bytes for a normal semver segment, vs a fixed big endian u64 needing 8 bytes

Turbo87:
@eth3lbert would love it if you could take a look as well :)

migrations/2026-05-26-120000-0000_semver_ord_v2/up.sql
Turbo87: oh yeah, it's 8 bytes. sorry, my head isn't working quite right today 🙈 

in that case the 2-3 bytes are definitely better 👍

eth3lbert (approved):
Looks solid to me! It also handles invalid cases like `'1.0.0-'` nicely, which is great. 👍

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13775/6a3ac659fdd64bcb8eb792fcc2abbc3003e66891..a037e0ddf27c04b2782071bf1a7229e82cc3c4cb)*

Turbo87 (approved):
alright, let's give this a try. nice work! 👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13775/6a3ac659fdd64bcb8eb792fcc2abbc3003e66891..221af8f6745c3e912260d9c90561af929ae224d9)*

danilobuerger:
@Turbo87 please let me know once you have done the manual batches so i can follow up with a PR.

Turbo87:
will do!

feel free to open the PR already in draft mode. we can merge it once the backfill is done. :)

Turbo87:
> please let me know once you have done the manual batches

done on both staging and prod. they all parsed clean without any errors 🎉

danilobuerger:
Hah! That was seriously fast @Turbo87 . I expected you to come back in a couple of weeks 😂. Well I better get on with making the next PR...

danilobuerger:
Follow up at #13782

PR #13781, opened 2026-05-27
Add nullable username column in users table

### LINKED ISSUE #13760: Create a new column `users.username` set to the value of `users.gh_login`
This depends on:

```
UPDATE users
SET username = gh_login;
```

MrSheerluck:
Fixes: #13760 

I've added a username column and kept it as nullable as informed in the issue. Also added up and down migration for the username column as well.

My only doubt is whether should I keep the `#[serde(skip)]` on username or not. I tested the `users` table on my local docker PostgreSQL and its running all migrations successfully including this new username migration.

Here's the screenshot of the users table:
Screenshot 2026-05-27 at 18 00 23


MrSheerluck:
@carols10cents I think I should remove the username from model to make the compilation successful. Let me know if you want me to do that or leave it as it is

carols10cents:
Hi, thank you for working on this! I think we should have the field on the model, and we can start setting the value of `users.username` whenever we set `users.gh_login`, to the same value (and then the backfill I'll run manually will set the rest).

MrSheerluck:
@carols10cents so I don't need to change anything in this PR right? As said in the issue, we should create another PR, am I right?

carols10cents:
You do need to change stuff in this PR to get it to compile with the field on the model :)

MrSheerluck:
In the models, I've already added the field username, can you please tell me what am I missing here? As far as I know, the action failures are coming from controllers

migrations/2026-05-27-120000_add_username_to_users/up.sql
hwittenborn, line 1: Is there any reason we can't seed the column with the existing GitHub usernames so the db field doesn't have to be nullable? Since this field isn't public yet I don't see the risk with doing otherwise.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13781/9b0658046aed41a568c3ef949644735d5810f918..ff371c19ee252cd48e3ad2ddf399b1ef893228c8)*

hwittenborn (changes requested):

migrations/2026-05-27-120000_add_username_to_users/up.sql
carols10cents: Migrations run during the deploy process, so they need to be definitely fast to run so that they don't hold up a deploy. Adding non-nullable columns with a non-constant default value can require a table rewrite.

carols10cents:
> In the models, I've already added the field username, can you please tell me what am I missing here? As far as I know, the action failures are coming from controllers

Yes, you will need to edit the controllers too, because they use the models.

Here's the error [from the backend/lint job](https://github.com/rust-lang/crates.io/actions/runs/26511438037/job/78076841402?pr=13781#step:7:47):

```
error[E0063]: missing field `username` in initializer of `crates_io_database::models::User`
  --> src/controllers/trustpub/emails.rs:56:9
   |
56 |         User {
   |         ^^^^ missing `username`
```

[That line is in a test function initializing a test `User` instance](https://github.com/rust-lang/crates.io/blob/def239ebc73a51156c9b884794e429995481f825/src/controllers/trustpub/emails.rs#L55-L56), so the `username` field needs to get a value in that initialization. It's fine to use the same value as `gh_login` gets: 

https://github.com/rust-lang/crates.io/blob/def239ebc73a51156c9b884794e429995481f825/src/controllers/trustpub/emails.rs#L58

Please give that a try!

MrSheerluck:
> > In the models, I've already added the field username, can you please tell me what am I missing here? As far as I know, the action failures are coming from controllers
> 
> Yes, you will need to edit the controllers too, because they use the models.
> 
> Here's the error [from the backend/lint job](https://github.com/rust-lang/crates.io/actions/runs/26511438037/job/78076841402?pr=13781#step:7:47):
> 
> ```
> error[E0063]: missing field `username` in initializer of `crates_io_database::models::User`
>   --> src/controllers/trustpub/emails.rs:56:9
>    |
> 56 |         User {
>    |         ^^^^ missing `username`
> ```
> 
> [That line is in a test function initializing a test `User` instance](https://github.com/rust-lang/crates.io/blob/def239ebc73a51156c9b884794e429995481f825/src/controllers/trustpub/emails.rs#L55-L56), so the `username` field needs to get a value in that initialization. It's fine to use the same value as `gh_login` gets:
> 
> https://github.com/rust-lang/crates.io/blob/def239ebc73a51156c9b884794e429995481f825/src/controllers/trustpub/emails.rs#L58
> 
> Please give that a try!

sure, I'll do that. I thought I only need to update the user's table. No worries

crates/crates_io_database/src/models/user.rs
```diff
@@ -25,6 +25,8 @@ pub struct User {
     pub account_lock_until: Option<DateTime<Utc>>,
     pub is_admin: bool,
     pub publish_notifications: bool,
+    #[serde(skip)]
```
carols10cents: I looked into this more, and I think we should remove the `#[serde(skip)]` here because we want the API responses that use this to start using `username` rather than `gh_login` (the place i found that uses the serialization is https://github.com/rust-lang/crates.io/blob/23554c310eeec13e7909895948dbccb4699ede21/src/controllers/trustpub/emails.rs#L17 and some other structs like that; `auth_user.gh_login` is used in places like https://github.com/rust-lang/crates.io/blob/23554c310eeec13e7909895948dbccb4699ede21/src/email/templates/trustpub_config_created/body.html.j2#L12-L15 and those will need to be updated to `auth_user.username` after this PR)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13781/9b0658046aed41a568c3ef949644735d5810f918..b0a1b38cdd4ce883990c1a87a39b8bbd548ef94c)*

crates/crates_io_database_dump/src/dump-db.toml
```diff
@@ -256,6 +256,7 @@ gh_login = "public"
 name = "public"
 gh_avatar = "public"
 gh_id = "public"
+username = "private"
```
carols10cents: I think we should go ahead and make this field `public` because we'll want anyone using the database dump to start using that column instead of `gh_login`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13781/9b0658046aed41a568c3ef949644735d5810f918..b0a1b38cdd4ce883990c1a87a39b8bbd548ef94c)*

migrations/2026-05-27-120000_add_username_to_users/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE users ADD COLUMN IF NOT EXISTS username VARCHAR;
```
carols10cents, line 1: Could you please add a comment for this field? If we do that, diesel will print the comment as documentation on the schema.

Here's an example migration that adds a comment to a new column on line 8 of the up migration: https://github.com/rust-lang/crates.io/commit/a1354d556cb9c02dd7edee1bb96235bb26734ab0#diff-262424d68a10590e54e60cf86669a4998364441a0b9202ac20d953dd91b21b42R8

and then once you've rerun the migration, the comment should show up in the changes to schema.rs, commit that change as well :)

The comment should say something like "Crates.io username, which may or may not be the same as any linked Oauth accounts' usernames" (please edit if that seems like too much info, not enough info, or confusing info!)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13781/9b0658046aed41a568c3ef949644735d5810f918..b0a1b38cdd4ce883990c1a87a39b8bbd548ef94c)*

carols10cents:
@Turbo87 brought up that [perhaps `login` is a better name for this column](https://github.com/rust-lang/crates.io/issues/13760#issuecomment-4562508568):

> I think I would advocate for naming the column login so that it matches the field name in [crates.io/api/v1/users/turbo87](https://crates.io/api/v1/users/turbo87)

but I think "username" is better because you'll never "log in" with this username, really.

@MrSheerluck what do you think about the field name? As someone relatively new to the codebase, which makes more sense to you, `username` or `login` (or something else)?

Turbo87:
> you'll never "log in" with this username, really

who knows... maybe one day you might 😉 

as I said in the other thread, my main reason for `login` would be consistency with our API responses.

MrSheerluck:
> @MrSheerluck what do you think about the field name? As someone relatively new to the codebase, which makes more sense to you, username or login (or something else)?

I lean username here. As someone newer to the codebase, username feels clearer to me for a crates.io-specific account name, while login suggests something auth-related, and users won’t actually log in with this value.

src/tests/util/test_app.rs
```diff
@@ -163,6 +163,7 @@ impl TestApp {
             account_lock_until: None,
             is_admin: false,
             publish_notifications: true,
+            username: None,
```
carols10cents: This spot I think should be:

```suggestion
            username: Some(new_user.gh_login.to_string()),
```

so that we don't have to update this spot again when we switch over to using this new field.

carols10cents (changes requested):
This PR is getting really close-- just one more little change I think!

@Turbo87 I would like to go with this PR once the last test helper function comment is fixed. what do you think?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13781/295c0ab7261e72e1b6d4dc6dd625840a66a70acf..73848ae9042c64b8514e8c0ae0fe7f292189d2c9)*

MrSheerluck:
@carols10cents updated as requested. Before merging just let me know what's the next issue I can pick up.

Turbo87:
> I would like to go with this PR once the last test helper function comment is fixed. what do you think?

I think we will need to adjust the `NewUser` struct and the `update_user_from_github` background job too.

Imagine this scenario:

- we merge and deploy this
- the migration adds the new column
- we backfill the column by copying the `gh_login` values into `username`
- someone new logs in
  - → we create the new `user` row with an empty `username`
- or an existing user logs in with a renamed GitHub login
  - → we update `gh_login`, but not `username`

in other words, we need to make sure that every future `gh_login` change also causes a `username` change until we explicitly stop the decoupling. we should probably also add a couple of tests for this, or adjust existing ones, if there are fitting ones.

we can also do all of that in another PR, but it would need to happen before we backfill the `username` column and start using it, unless we want to do that multiple times 😅

carols10cents:
> in other words, we need to make sure that every future `gh_login` change also causes a `username` change until we explicitly stop the decoupling. we should probably also add a couple of tests for this, or adjust existing ones, if there are fitting ones.

@Turbo87 Ah yes, great catch! I think we should take care of this in this PR.

@MrSheerluck I've added and edited some tests so that they fail now because `users.username` isn't being updated when `users.gh_login` is and `users.username` should be getting updated-- could you try getting them to pass? You'll need to edit code in https://github.com/rust-lang/crates.io/blob/6a2d432348847904e177b567c96e37fef4feff53/crates/crates_io_database/src/models/user.rs and https://github.com/rust-lang/crates.io/blob/6a2d432348847904e177b567c96e37fef4feff53/src/worker/jobs/update_user_from_github.rs and possibly other places depending on the changes you make to those two files.

Let me know if you have any questions!

carols10cents:
Whoops and I forgot to run `cargo fmt` before I pushed 😰 if you could fix that for me too while you're getting the tests to pass that would be great! ❤️

crates/crates_io_database/src/models/user.rs
```diff
@@ -84,6 +85,7 @@ impl User {
 pub struct NewUser<'a> {
     pub gh_id: i32,
     pub gh_login: &'a str,
+    pub username: Option<&'a str>,
```
carols10cents: I think `username` should be required when we're creating a new user, because we should always be setting it after this PR. That is, this line should be:


```suggestion
    pub username: &'a str,
```

carols10cents (changes requested):
Just one last thing, I think!!!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13781/295c0ab7261e72e1b6d4dc6dd625840a66a70acf..f1bb7a0c0596fb894c5104894d2926909cbeedde)*

crates/crates_io_database/src/models/user.rs
MrSheerluck: oh yeah makes sense

Turbo87 (approved):

carols10cents (approved):
Looks great!!!!!! Thank you so much for sticking with this!!

Next, could you send a pull request for the 2nd part of https://github.com/rust-lang/crates.io/issues/13760 that includes a migration to add a `NOT NULL` constraint and changes the `username` field's type to remove the `Option` where needed?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13781/295c0ab7261e72e1b6d4dc6dd625840a66a70acf..5cc8cd7246e6403f0404f9dea3d6aa639ad892bb)*

PR #13782, opened 2026-05-27
Switch default_version sort key to `semver_ord_v2` and require it

danilobuerger:
In #13775 we added `semver_ord_v2` as a nullable column populated by an insert trigger. Existing rows have since been backfilled manually, so the column is now fully populated and can be marked NOT NULL.

This commit:

- Adds a migration that sets `versions.semver_ord_v2 NOT NULL`.
- Exposes the column in `schema.rs` and adds a Diesel SQL function binding `semver_ord_v2(num: Text) -> Nullable<Bytea>`. The binding stays nullable because the underlying SQL function returns NULL for unparseable input.
- Switches the readers in `calculate_default_version` and `controllers/krate/versions::list` from `semver_ord` to `semver_ord_v2`, including the seek-based pagination filter.

For the "non-prerelease first" tier in default version selection, the JSONB introspection (`jsonb_typeof(retrieve_as_object(3)) = 'array'`) is replaced by a direct check of the last byte:
`get_byte(ord, octet_length(ord) - 1) <> 3`. The encoding guarantees the trailing byte is 0x03 for releases and 0x00 (the prerelease end marker) for prereleases.

The old `semver_ord` JSONB column, its SQL function, and the `semver_ord.rs` test are intentionally left in place and will be removed in a follow up once the new path is confirmed working.

Note: The `get_byte` expression is used as the previous available `semver_no_prerelease` was removed in #8867.

**Deployment Note:**
The migration uses a direct `ALTER COLUMN ... SET NOT NULL`, which briefly takes `ACCESS EXCLUSIVE` on `versions` and scans the table to verify the constraint.

### DIFF
--- a/migrations/2026-05-27-120000-0000_semver_ord_v2_not_null/up.sql
+++ b/migrations/2026-05-27-120000-0000_semver_ord_v2_not_null/up.sql
@@ -1,0 +1,8 @@
+-- safety-assured:start
+-- A `SELECT * FROM versions WHERE semver_ord_v2 IS NULL` against prod runs in
+-- around 3 seconds, so the validation scan that `SET NOT NULL` performs under
+-- ACCESS EXCLUSIVE is expected to be at least as fast. Not worth the extra
+-- CHECK constraint work for this table size.
+alter table versions
+    alter column semver_ord_v2 set not null;
+-- safety-assured:end
--- a/crates/crates_io_database/src/models/default_versions.rs
+++ b/crates/crates_io_database/src/models/default_versions.rs
@@ -1,7 +1,8 @@
 use crate::schema::{default_versions, versions};
 use crates_io_diesel_helpers::SemverVersion;
-use diesel::dsl::jsonb_typeof;
+use diesel::dsl::sql;
 use diesel::prelude::*;
+use diesel::sql_types::Bool;
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
 use tracing::{debug, instrument, warn};
 
@@ -126,10 +127,13 @@ async fn calculate_default_version(
         .order((
             // 1. Non-yanked first
             versions::yanked,
-            // 2. Non-prerelease first
-            jsonb_typeof(versions::semver_ord.retrieve_as_object(3)).eq("array"),
+            // 2. Non-prerelease first. The post-patch byte of `semver_ord_v2`
+            // is 0x03 for releases and 0x00 for prereleases.
+            sql::<Bool>(
+                "get_byte(versions.semver_ord_v2, octet_length(versions.semver_ord_v2) - 1) <> 3",
+            ),
             // 3. Higher semver first
-            versions::semver_ord.desc(),
+            versions::semver_ord_v2.desc(),
             // 4. Higher ID first as tie-breaker
             versions::id.desc(),
         ))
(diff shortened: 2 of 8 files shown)


danilobuerger:
@Turbo87 since CI complains about the `ACCESS EXCLUSIVE` lock I mentioned in my deployment notes: Do you want the check detour to save a table scan? I didn't think it was worth it (thats why I went directly with the set not null migration).

migrations/2026-05-27-120000-0000_semver_ord_v2_not_null/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table versions
+    alter column semver_ord_v2 set not null;
```
Turbo87, line 7: if we set this to `not null`, what happens in the trigger is a version number can't be parsed and the fn returns `null`? would it just reject the `INSERT`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13782/9b0658046aed41a568c3ef949644735d5810f918..1560602628777852082cf529b8badfc0b9a4db98)*

danilobuerger: Yes. Which can only happen if:
a) it has a number with more than 255 digits
b) it doesn't match the regex
both of which would have been denied before by the semver crate.

Turbo87: yeah, I think due to the Rust-based check in the publish flow this shouldn't be an issue in practice :)

Turbo87:
> Do you want the check detour to save a table scan?

good question. I think it would probably work well enough without it with the current table size, but at the same time, I guess it won't hurt either? if we try the `CONSTRAINT` route, I'm wondering if `diesel-guard` will still warn about the `SET NOT NULL` in the same migration. if it still warns and we have to silence the warning, then that `CONSTRAINT` is probably not worth it. if adding the temporary `CONSTRAINT` silences the warning then I guess we should prefer that over silencing the warning with a comment. what do you think?

danilobuerger:
> > Do you want the check detour to save a table scan?
> 
> good question. I think it would probably work well enough without it with the current table size, but at the same time, I guess it won't hurt either? if we try the `CONSTRAINT` route, I'm wondering if `diesel-guard` will still warn about the `SET NOT NULL` in the same migration. if it still warns and we have to silence the warning, then that `CONSTRAINT` is probably not worth it. if adding the temporary `CONSTRAINT` silences the warning then I guess we should prefer that over silencing the warning with a comment. what do you think?

It would probably have to be a different migration (so different PR), because you would first add it, then manually validate it and then apply the set null.

Turbo87:
right, because the migrations are blocking server startup... yeah, in that case let's silence the warning with a comment that explains our reasoning

`SELECT * FROM versions WHERE semver_ord_v2 IS NULL` on prod just took about 3 seconds, and I'm assuming that the `SET NOT NULL` check will probably be faster than this query, so this probably shouldn't be an issue.

danilobuerger:
Its updated. Fingers crossed 🤞

Turbo87:
I ran our smoke tests against the staging environment, where this is deployed now and things appear to be working okay. I'll keep it as-is for now and will promote this to production tomorrow (unless I forget... 😅)

Turbo87:
```
{"level":"INFO","target":"crates_admin::migrate","timestamp":"2026-05-28T10:48:14.579975Z","message":"Migrating the database"}
Running migration 2026-05-27-120000-0000_semver_ord_v2_not_null
{"level":"INFO","target":"crates_admin::migrate","timestamp":"2026-05-28T10:48:16.038170Z","message":"Synchronizing crate categories"}
```

The `NOT NULL` adjustment took about 1.5sec on production and so far hasn't caused any issues :)

PR #13786, opened 2026-05-27
Drop `semver_ord` JSONB column, function, and trigger

danilobuerger:
In #13782 the readers switched to `semver_ord_v2`, leaving the old JSONB-based `semver_ord` column, its SQL function, and the `set_semver_ord` insert trigger in place pending confirmation that the new path works. So this commit removes the old encoding entirely.

The `down.sql` is a verbatim copy of the original 2025-03-06 `semver_ord` migration's `up.sql`.

With `semver_ord_v2` being bytea / memcmp-based, the database `C` collation is no longer required, so the setup changes from #13759 (devcontainer `POSTGRES_INITDB_ARGS`, the `--lc-collate=C` flag in the CI `createdb`, and the related notes in `docs/CONTRIBUTING.md` and `AGENTS.md`) can be reverted as a follow up.


migrations/2026-05-27-130000-0000_drop_semver_ord/up.sql
```diff
@@ -0,0 +1,7 @@
+drop trigger if exists trigger_set_semver_ord on versions;
+drop function if exists set_semver_ord();
+-- safety-assured:start
```
Turbo87, line 3: what warning was this suppressing?

Turbo87 (approved):
> With `semver_ord_v2` being bytea / memcmp-based, the database `C` collation is no longer required, so the setup changes from #13759 (devcontainer `POSTGRES_INITDB_ARGS`, the `--lc-collate=C` flag in the CI `createdb`, and the related notes in `docs/CONTRIBUTING.md` and `AGENTS.md`) can be reverted as a follow up.

in https://github.com/rust-lang/crates.io/pull/13759#pullrequestreview-4369463169 it was mentioned that the C collation setup is probably still useful since it aligns us a bit more with the production setup. I guess it doesn't hurt too much to keep it around.

I'll keep this PR open for a couple of days until we have verified that the v2 impl works as expected in production :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13786/def239ebc73a51156c9b884794e429995481f825..52c3f101653bc31b2ceb8a1cb13d9795c89ee18e)*

migrations/2026-05-27-130000-0000_drop_semver_ord/up.sql
danilobuerger: Here is the failing run: https://github.com/rust-lang/crates.io/actions/runs/26539539455/job/78177315322

Turbo87: just for future reference, since CI logs are discarded at some point:

```
❌ DROP COLUMN  (line 3)

Problem:
  Dropping column 'semver_ord' from table 'versions' requires an ACCESS EXCLUSIVE lock, blocking all operations. This typically triggers a table rewrite with duration depending on table size.

Safe alternative:
  1. Mark the column as unused in your application code first.
  
  2. Deploy the application without the column references.
  
  3. (Optional) Set column to NULL to reclaim space:
     ALTER TABLE versions ALTER COLUMN semver_ord DROP NOT NULL;
     UPDATE versions SET semver_ord = NULL;
  
  4. Drop the column in a later migration after confirming it's unused:
     ALTER TABLE versions DROP COLUMN semver_ord;
  
  Note: Postgres doesn't support DROP COLUMN CONCURRENTLY. The rewrite is unavoidable but staging the removal reduces risk.
```

Turbo87: unless I'm missing something, "This typically triggers a table rewrite with duration depending on table size" seems wrong for Postgres. from what I've found so far, this should only be a table metadata change, so I'd say that is a bug in `diesel-guard` 🤔

danilobuerger: Yeah, I think so too.

Turbo87:
the `semver_ord_v2` seems to be working fine in production from what I can tell, so let's get the cleanup started :)

danilobuerger:
Thanks @Turbo87 for getting me to do these. It was fun!

Turbo87:
sure! thanks for working on this :)

if you're interested, there are probably a lot more DB optimization opportunities in the codebase :D

PR #13789, opened 2026-05-28
Rename `users.gh_login` column to `users.login`

### LINKED ISSUE #13760: Create a new column `users.username` set to the value of `users.gh_login`

### LINKED ISSUE #13772: Deprecate `users.gh_login` column
Depends on:

- https://github.com/rust-lang/crates.io/issues/13763
- https://github.com/rust-lang/crates.io/issues/13768
- possibly other sub-issues of the tracking issue as well

After the dependent issues are complete, ensure all remaining uses of `users.gh_login` are removed from the Rust code. Once there are no more uses, deprecate/remove the column from the database.

Turbo87:
The `users.login` column will eventually represent a first-class crates.io username, independent of the user's GitHub login (which is now stored separately in `oauth_github.login`). For now the two stay in sync. Renaming the column is a first step toward decoupling them.

The migration renames the `users.gh_login` column to `users.login` and the `lower_gh_login` index to `lower_login`. The Diesel schema, struct fields, builder methods, jinja email templates, and database dump configuration are updated to match.

The `create_or_update_github_team()` function in `controllers/krate/owners.rs` already had a `login: &str` parameter for the team slug, and its body previously bound `let gh_login = &req_user.gh_login;` for the user login. The local has been renamed to `user_login` so the two values do not collide after the rename.

### Related

- https://github.com/rust-lang/crates.io/issues/13760#issuecomment-4562254740
- Resolves https://github.com/rust-lang/crates.io/issues/13760
- Closes https://github.com/rust-lang/crates.io/issues/13772
- Closes https://github.com/rust-lang/crates.io/pull/13781


LawnGnome (approved):
LGTM.

It's convenient that the public API field was already `login`; good foresight, whoever that was!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13789/2787afaa5639481ab8cb69c5e2423601f3f8e9c9..f2fbbc71d54a5e050c6b83df7c05d08bce743ea9)*

carols10cents:
[it seems like Tobi has convinced himself this isn't a good idea](https://github.com/rust-lang/crates.io/issues/13760#issuecomment-4577569385): 

> The pure rename is probably a non-starter due to the server errors.

so i'm closing this. Reopen if you disagree!

PR #13790, opened 2026-05-28
Rename `users.gh_login` to `users.login` via compatibility view

Turbo87:
Renames the `users.gh_login` column to `users.login` as a first step toward giving crates.io accounts a first-class username independent of the user's GitHub identity. The GitHub login already lives in `oauth_github.login`. For now both columns stay in sync.

The four commits sit on the same branch so the strategy can be reviewed end to end. Commit 1 is standalone and can land on its own. Commits 2 and 3 are intended to ship in the same deploy: commit 2 adds the view, commit 3 switches application code, and the view's whole purpose is to bridge old dynos during the commit-3 rollout. Commit 4 lands in a later deploy once that rollout has stabilised.

1. `crates_io_database_dump: Restrict visibility check to base tables`. Defensive fix that makes the dump pipeline's `check_visibility_config` test ignore regular views. Materialized views were already ignored because PostgreSQL omits them from `information_schema`. No behavior change against the current schema.
2. `Rename \`users.gh_login\` to \`users.login\` via compatibility view`. Renames the underlying table to `users_v2`, renames the column to `login`, and adds a `users` view that exposes both `gh_login` and `login`. PostgreSQL's auto-updatable view rules forward INSERT/UPDATE/DELETE to the underlying table, so application code that still references `gh_login` keeps working unchanged. A `schema.patch` hunk keeps the Diesel-generated schema stable until the next commit lands.
3. `Switch application code to \`users.login\``. Renames every Rust, jinja, snapshot, and Diesel-schema reference from `gh_login` to `login`. Older dynos still reading and writing `users.gh_login` keep working through the view during the rollout.
4. `Drop \`users\` compatibility view and rename \`users_v2\` back`. Cleanup once every running deploy uses `login`. The schema returns to a single base table.

### Public dump format

Commit 2 changes the public database dump format. The `users.csv` entry becomes `users_v2.csv` for the period between commits 2 and 4, and commit 4 restores it. The dump pipeline runs `ALTER TABLE ... DISABLE TRIGGER ALL` and `TRUNCATE` against the dumped relation, neither of which PostgreSQL allows on a view, so the dump has to target the underlying table directly while the view exists. Downstream dump consumers would see one rename in each direction, if the database dump was generated in between commit 2 and 4.

### Related

- https://github.com/rust-lang/crates.io/issues/13760
- https://github.com/rust-lang/crates.io/pull/13789
- https://github.com/rust-lang/crates.io/pull/13781


crates/crates_io_database_dump/src/gen_scripts.rs
```diff
@@ -119,7 +119,6 @@ impl VisibilityConfig {
 mod tests {
```
carols10cents, line 119: I think this commit is good regardless of whether we end up using the rest of this PR

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13790/8f436f95905e79741dea51af7e9ba0568e037e4b..64830b49d7757b1807348d767c47264d12a00698)*

src/controllers/helpers/authorization.rs
```diff
@@ -46,7 +46,7 @@ impl Rights {
                     // the answer. If this is not the case, then we could accidentally leak
                     // private membership information here.
                     let is_team_member = match gh_client
-                        .team_membership(team.org_id, team.github_id, &user.gh_login, &token)
+                        .team_membership(team.org_id, team.github_id, &user.login, &token)
```
carols10cents, line 49: shouldn't this be changed to `oauth_github.login` instead?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13790/8f436f95905e79741dea51af7e9ba0568e037e4b..64830b49d7757b1807348d767c47264d12a00698)*

Turbo87:
- closing due to https://github.com/rust-lang/crates.io/issues/13760#issuecomment-4577850384, which is in favor of https://github.com/rust-lang/crates.io/pull/13781

PR #13792, opened 2026-05-28
Adjust `NewUser::insert()` fns to return only user IDs

Turbo87:
The `users.gh_*` columns are being migrated into the `oauth_github` table, which will make `User`'s selectable reference columns outside users and break `RETURNING` `User::as_returning()` from an insert against `users`. Narrowing the return type to `i32` sidesteps that without forcing the insert path to re-fetch through the joined query.

The `insert()` and `insert_or_update()` methods on `NewUser` both returned a fully populated `User`. The sole production caller (`controllers::session::create_or_update_user()`) only ever consumed `.id` from that value, and most test helpers were in the same shape: they held a `User` only to read its id. The `RETURNING` clause is now narrowed to `users::id`.

The change ripples outward through `controllers::session` and a few test helpers (`typosquat::faker::user()`, `add_team_to_crate()`, `tests::user()` in the download-metadata job). Each of those previously held a `User` only for its id, so they now take or return `i32` directly.

A few sites genuinely needed the full `User`. Test callers of `save_user_to_database()` in `src/tests/user.rs` and `src/tests/worker/update_user_from_github.rs` re-fetch the row via `User::find()`. `db_new_user()` keeps handing a `User` to `MockCookieUser` (to preserve `as_model()` for the many tests that read fields off it), synthesized from the `NewUser` builder and the returned id with database defaults for the remaining columns.

One spot worth a closer look during review: `create_or_update_user()` now sources `gh_login`, `gh_id`, and `gh_avatar` for the `NewOauthGithub` builder and the email confirmation template from the `NewUser` argument instead of the upserted `User`. The upsert's `SET` clause writes those same `excluded(...)` values, so the visible state matches.

### DIFF
--- a/crates/crates_io_database/src/models/user.rs
+++ b/crates/crates_io_database/src/models/user.rs
@@ -100,7 +100,7 @@ impl NewUser<'_> {
     }
 
     /// Inserts the user into the database, or updates an existing one.
-    pub async fn insert_or_update(&self, mut conn: &AsyncPgConnection) -> QueryResult<User> {
+    pub async fn insert_or_update(&self, mut conn: &AsyncPgConnection) -> QueryResult<i32> {
         diesel::insert_into(users::table)
             .values(self)
             // We need the `WHERE gh_id > 0` condition here because `gh_id` set
@@ -119,7 +119,7 @@ impl NewUser<'_> {
                 users::gh_avatar.eq(excluded(users::gh_avatar)),
                 users::gh_encrypted_token.eq(excluded(users::gh_encrypted_token)),
             ))
-            .returning(User::as_returning())
+            .returning(users::id)
             .get_result(&mut conn)
             .await
     }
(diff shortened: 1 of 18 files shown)


LawnGnome:
OK, I should have said I was reviewing this, but I wanted to sleep on it. Sorry.

I'm wondering a little if this is a code smell that the `User` struct has the wrong shape now that we're migrating to the per-provider OAuth tables. It might be premature to actually do anything about this, since this is all very much in flux, but I'm interested if you've thought about that at all, @Turbo87?

Turbo87:
> I'm wondering a little if this is a code smell that the `User` struct has the wrong shape now that we're migrating to the per-provider OAuth tables

it definitely has, yes. I consider this an intermediate state to unblock the decoupling, but this will definitely need further changes.

LawnGnome:
Cool. I'm especially glad to hear you say that since I got to #13800 right after writing that in my notifications, which only intensified that feeling. :laughing:

Issue #13758, opened 2026-05-26
Add the concept of a crates.io username for identity


carols10cents:
This is the main tracking issue for implementation of https://github.com/rust-lang/rfcs/pull/3946.

If you would like to help with any sub-issues, please comment on them that you plan to work on them and ask any clarifying questions! Do note there are some sub-issues that may have dependencies on other sub-issues; I will try to call those dependencies out.

Comment on this issue if it seems like something is missing from the sub-issues.

Turbo87:
I think there is a small conflict between:

- https://github.com/rust-lang/crates.io/issues/13761 and/or https://github.com/rust-lang/crates.io/issues/13762
- and https://github.com/rust-lang/crates.io/issues/13768

Once we consider `users.username` unique and stop updating it from GitHub then a user from GitHub with a username of an existing crates.io account would no longer be able to register their account until the new signup flow is implemented.

I'm not sure yet what the best way would be to work around that "deadlock".

Turbo87:
one more thing I noticed today: we still have the `gh_id`, `gh_avatar` and `gh_encrypted_token` columns on the `users` table. these are redundant now that we have the `oauth_github` table with the same values. it might be worth cleaning this up first before we add the additional account decoupling complexity.

carols10cents:
> one more thing I noticed today: we still have the `gh_id`, `gh_avatar` and `gh_encrypted_token` columns on the `users` table. these are redundant now that we have the `oauth_github` table with the same values. it might be worth cleaning this up first before we add the additional account decoupling complexity.

they're actually not redundant right this second-- the main reason being `oauth_github` records must have a `users` record to join to, so when you log in we have to find_or_create in users first. so as you noted in previous comments, these are all kind of tied together 😅

Turbo87:
> the main reason being `oauth_github` records must have a `users` record to join to, so when you log in we have to find_or_create in users first

can you expand this? if we dropped these columns from `users` then what blocks us from creating the user first and then the `oauth_github` row referencing the user?

carols10cents:
> can you expand this? if we dropped these columns from `users` then what blocks us from creating the user first and then the `oauth_github` row referencing the user?

[...]

Issue #13773, opened 2026-05-26
Implement avatar choice


carols10cents:
Allow users to pick which avatar they want shown in crates.io. Their choices will be:

- Their GitHub avatar
- A default avatar, TBD by the implementer of this issue (Ferris?)

This will be more useful someday when users can associate accounts from services other than GitHub as well, and they may have different avatars on different services and want to pick between them.

PR #13800, opened 2026-05-29
Read user avatar from `oauth_github` table via `LEFT JOIN`

Turbo87:
`User.gh_avatar` now sources from `oauth_github.avatar` via a `LEFT JOIN` in the model's `base_query`. The `users.gh_avatar` column is still written by `NewUser::insert_or_update()` but is no longer read by any of the affected query paths.

Each `User::as_select()` call site picks up an extra `.left_join(oauth_github::table)` so the selectable's reference to `oauth_github::avatar` resolves. `User::find()`, `Version::published_by()`, and the `crate_owner_invitations` accept path switched from `.find(id)` to `.filter(users::id.eq(id))` because the joined query doesn't support `find()`.

This will allow us to drop the `users.gh_avatar` column in the future.

### Related

- https://github.com/rust-lang/crates.io/issues/13758#issuecomment-4565431099
- #13792

### DIFF
--- a/crates/crates_io_database/src/models/action.rs
+++ b/crates/crates_io_database/src/models/action.rs
@@ -62,7 +62,7 @@ impl VersionOwnerAction {
 
         version_owner_actions::table
             .filter(version_id.eq(version.id))
-            .inner_join(users::table)
+            .inner_join(users::table.left_join(oauth_github::table))
             .select((VersionOwnerAction::as_select(), User::as_select()))
             .order(version_owner_actions::dsl::id)
             .load(&mut conn)
@@ -74,7 +74,7 @@ impl VersionOwnerAction {
         versions: &[&Version],
     ) -> QueryResult<Vec<Vec<(Self, User)>>> {
         Ok(Self::belonging_to(versions)
-            .inner_join(users::table)
+            .inner_join(users::table.left_join(oauth_github::table))
             .select((VersionOwnerAction::as_select(), User::as_select()))
             .order(version_owner_actions::dsl::id)
             .load(&mut conn)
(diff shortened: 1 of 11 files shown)


carols10cents:
So currently there are some user records that don't have `oauth_github` records, but most of them have a vestigial `users.gh_avatar` URL (that we should probably delete for PII reasons anyway since they've disassociated/deleted their github accounts).

There is only one user record in the database right now `where gh_avatar is null` (not going to post the username here but i know you have access to go get it if you want), and the UI just renders an `img` tag without a `src` and it doesn't look great 😂 

This change will put a bunch more users into the situation of their user JSON returning `avatar: null` and their user page looking gross, should we actually add a fallback avatar image (that we have plans to use with https://github.com/rust-lang/crates.io/issues/13773 anyway) and use that in this case?

Turbo87:
> should we actually add a fallback avatar image

yeah, that seems like a good idea! 👍

carols10cents (approved):
https://github.com/rust-lang/crates.io/pull/13804 resolved my concerns with this PR, i think this is ready to merge!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13800/8f436f95905e79741dea51af7e9ba0568e037e4b..c0aaa7697fc7aad343ad35074a7173518b140e5f)*

LawnGnome (approved):
LGTM.[^concern]

[^concern]: My concern in #13792 hasn't gone away, obviously, but I know you share it too and this is a necessary step towards a Better Future™.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13800/8f436f95905e79741dea51af7e9ba0568e037e4b..c0aaa7697fc7aad343ad35074a7173518b140e5f)*

PR #13812, opened 2026-05-31
Replace highlight.js with shiki for syntax highlighting

Turbo87:
Code blocks in rendered Markdown and the docs pages are now highlighted with [shiki](https://shiki.matsu.io/) instead of highlight.js. shiki uses the same TextMate grammars and themes as VS Code, which produces noticeably more accurate highlighting.

The highlighter is built synchronously with the pure-JavaScript regex engine, so no Oniguruma WebAssembly is shipped. Rust and the `github-light` and `github-dark` themes are bundled eagerly since crate sources are predominantly Rust, which lets Rust code highlight before the first paint. Every other grammar is loaded on demand in its own chunk.

Tokens carry inline `light-dark()` colors, so the hand-maintained `.hljs-*` theme block in `TextContent.svelte` is removed and color scheme switching needs no extra CSS.

This grows the eager highlighting payload from ~29 KB to ~58 KB gzipped (esbuild, minified). In return the non-Rust grammars that highlight.js shipped upfront are now lazy chunks of ~1 to 5 KB gzipped each, loaded only when a README needs them. The extra ~29 KB buys us a significantly better highlighting implementation:

### Before

Bildschirmfoto 2026-05-31 um 18 17 06

### After

Bildschirmfoto 2026-05-31 um 18 17 16


LawnGnome (approved):
Seems worth trying out; this is easy enough to revert if for some reason there's an issue.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13812/9a9e0ef8efac84ac1ddca18be934b6f06e196722..f7936ea716046487b6e1267966baf3510ce3fd2c)*

PR #13816, opened 2026-06-01
Remove "Browse All Crates" link from `Header`

Turbo87:
With 200k+ crates in the registry, browsing the full list is no longer a meaningful entry point, so the link no longer warrants a premium spot in the topbar. Search and the keyword listings are the practical ways to find crates now.

This drops the link from both the nav and the mobile menu, along with the now-unused `.sep` separator rule, and updates the affected end-to-end specs and ARIA snapshots accordingly.

### DIFF
--- a/e2e/acceptance/404.spec.ts-snapshots/aria.yml
+++ b/e2e/acceptance/404.spec.ts-snapshots/aria.yml
@@ -7,8 +7,6 @@
     - button "Search"
   - navigation:
     - 'button "Change color scheme. Current: system"'
-    - link "Browse All Crates":
-      - /url: /crates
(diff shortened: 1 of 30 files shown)


LawnGnome (approved):
I think this is fair enough. (There'll probably be someone who relies on that to get to the list of crates by recent download count, but that is only one more click away via the most downloaded header, I guess.)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13816/163859155ecc1ca9300f9c9c3315aa6cad894ed0..09d1050866c03ea45017af3c174f96c8d8d26391)*

PR #13725, opened 2026-05-21
Fix vertical alignment for header

yotamofek:
Before:
image

After:
image

Issue #3687, opened 2021-06-08
Login error covers login link


ehuss:
**Describe the bug**
When running `cargo login` to log in, it instructs you to visit https://crates.io/me.  Doing that, I see the following:

<img width="415" alt="image" src="https://user-images.githubusercontent.com/43198/121267716-78e40080-c871-11eb-837b-68b23092d828.png">

with no obvious clue what to do.  You have to wait for that error to go away to reveal the login link below it.  If you have a small-width display, the login link is further hidden behind a Menu drop-down.  Even then, it is not obvious what to do if you don't already know how crates.io authentication works.

**To Reproduce**
Be logged out.  Then visit https://crates.io/me.

**Expected behavior**
If you are not logged in, visiting https://crates.io/me should display a login page (or some sort of page to indicate you need to do something to proceed).  Another alternative is if you are already logged in to GitHub to automatically pass through, though I'm not sure how I feel about that. 

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Any

wangkirin:
I think this issue is caused by the adjustment of `header`  in PR #3681 
May be we need to adjust the position of notifications in sync.

Turbo87:
we've decided to revert the layout change for now until we've decided how we want to approach this issue

ehuss:
I think this was closed without being resolved (or maybe something changed in the meantime).  I still get the following behavior:

<img width="507" alt="image" src="https://user-images.githubusercontent.com/43198/147125717-3e5d41b4-84f9-4503-8ae9-111d02970ef8.png">

Covering the login link.

My expected behavior isn't about the positioning of the link or the error message.  It is that the entire page should be a clear message instructing the user how to log in (or, possibly skip that entire process if they have already authorized).  

That is, the current experience is:

1. Go to crates.io/me 
2. see home page with error message 
3. wait for error message to disappear 
4. determine that you need to click the sign in link
5. popup page to authorize with GitHub 
6. land in settings page.

I would expect it to be:

1. Go to crates.io/me 
2. page to authorize with GitHub 
3. land in settings page

Turbo87:
@ehuss yeah, I agree that I was too quick with resolving this. The primary issue reported here I think was caused by https://github.com/rust-lang/crates.io/pull/3681, which moved the login link to the far right side of the screen, which increase the chance of intersecting with that notification messages.

[...]

Issue #1559, opened 2018-11-22
Should the width for the content be increased?


MaikKlein:
First of all I am not a webdev.

I always felt that crates.io looks a bit cramped on my desktop machine.

For example this is how it currently looks

![](https://i.imgur.com/oeF0coM.png)

After I changed the width

![](https://i.imgur.com/lqyNWx0.png)

Now it has the same size as Github. I personally think this would be an improvement, what do you think?

carols10cents:
Having the README be the same width as GitHub, at least for crate show pages, would probably be nice. Not sure if that should extend to search pages or settings pages or the other pages we have, and if not, if it would be weird that the widths change...

moonheart08:
Maybe make it.. not be fixed width? Please? Screen sizes above 16:9 really suffer, and a lot of screen space is wasted. Pictured is a 21:9 display viewing crates.io.
![image](https://user-images.githubusercontent.com/7806367/73993186-11354280-4917-11ea-9a12-38bd7d7bd817.png)

miDeb:
There was already an issue for that (https://github.com/rust-lang/crates.io/issues/970), but if you go to the linked page https://crates.io/crates/colorparse today, it looks more like the "before" in the gif of the closing pr https://github.com/rust-lang/crates.io/pull/1064. Maybe that's even a regression?

ibaryshnikov:
Hi all! What's the state of this issue? Any desire to make a change for wider screens? It looks like here's the place responsible for the docs width
https://github.com/rust-lang/crates.io/blob/575b2fb97a492624e75e550fdebae209d1904906/app/styles/crate/version.module.css#L13-L19

Turbo87:
we're slowly working towards a state of the code where we should be able to increase the width without too many issues :)

ibaryshnikov:
Does it break something else when increased?

Turbo87:
@ibaryshnikov yes, see https://github.com/rust-lang/crates.io/issues/3687#issuecomment-857418731 for example 😅

hamidrezakp:
What if we support multiple fixed content widths? like Wikipedia, an standard and a wide option.

standard can be the current size but the wide one would fill the entire page.

This would be great in the `code` tab when we want to read codes and have files open next to it.

PR #13826, opened 2026-06-02
Use full page width for `Header`

Turbo87:
This drops the `width-limit` so the header spans the full viewport width instead of being constrained to the central content column. It caps the search box at 600px and centers it between two flexible side columns so it no longer stretches edge to edge on wide screens, and adds a column gap so the logo, search box, and nav do not crowd each other.

It also keeps the login button on a single line, since the nav column is now flexible and would otherwise let the label wrap.

This ports the header half of #3681 to the Svelte frontend and resolves half of #1559. The grid refactor and single search form from that PR already exist on `main`, and the dropdown alignment fixes landed separately in #13725.

#3681 was previously reverted in #3696 because of #3687, where logged-out users hitting a protected route saw a flash error overlapping the header login link, which the wider header made worse. #4315 has since fixed that root cause by rendering a proper "This page requires authentication" page instead of a flash error, so dropping the `width-limit` here cannot reintroduce that regression.

Screen Shot 2026-06-02 at 11 06 21

### Related

- #1559
- #3681
- #3687
- #3696
- #4315
- #13725

### DIFF
--- a/svelte/src/lib/components/Header.svelte
+++ b/svelte/src/lib/components/Header.svelte
@@ -36,7 +36,7 @@
 </script>
 
 <header class="header" class:hero>
-  <div class="header-inner width-limit">
+  <div class="header-inner">
     <a href={resolve('/')} class="index-link">
       <enhanced:img src="$lib/assets/cargo.png?w=38;76;114" role="none" alt="" class="logo" sizes="38px" />
       crates.io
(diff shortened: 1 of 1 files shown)


LawnGnome (approved):
LGTM!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13826/c00969f7789d9b562f2d66e21c752fce7f855e84..cc47be66af180a40984769039853e08ac89bede6)*

PR #13827, opened 2026-06-02
Show crate descriptions and version/downloads in front-page lists

Turbo87:
This reworks the front-page crate lists so each row shows the crate's description and a more useful trailing value in place of the chevron. New Crates and Just Updated now display the newest version on the trailing edge without the `v` prefix, while Most Downloaded and Most Recent Downloads show a compact download count with a download icon, the latter using the 90-day count. The description becomes the row subtitle, and the name, description, and version all truncate with an ellipsis when they get too long.

The compact counts come from `formatShortNum()`, which gains a `threshold` option and a billions (`B`) unit and now rolls over to the next unit as soon as a number reaches four digits, so `1,000,000` reads as `1.0M` rather than `1,000K`. The crate sidebar passes the previous `1500` threshold so its source-lines-of-code figure stays consistent with the generated OpenGraph image.

Each row stays a single link, so to keep the screen-reader links list scannable its accessible name is restricted to the crate name via `aria-labelledby`, with the description and trailing value exposed through `aria-describedby` and announced after the name. Popular Keywords and Popular Categories are unchanged and keep the chevron.

Bildschirmfoto 2026-06-02 um 12 46 13
Bildschirmfoto 2026-06-02 um 12 46 32

### DIFF
--- a/svelte/src/lib/components/Tooltip.svelte
+++ b/svelte/src/lib/components/Tooltip.svelte
@@ -8,6 +8,11 @@
 
   interface BaseProps {
     side?: 'top' | 'bottom' | 'left' | 'right';
+    /**
+     * Milliseconds to wait after the pointer enters the anchor before showing
+     * the tooltip. Defaults to `0`, which shows it immediately.
+     */
+    delay?: number;
   }
 
   interface TextProps extends BaseProps {
@@ -22,10 +27,11 @@
 
   type Props = TextProps | ChildrenProps;
 
-  let { text, children, side = 'top' }: Props = $props();
+  let { text, children, side = 'top', delay = 0 }: Props = $props();
 
   let anchorElement = $state<HTMLElement | null>(null);
   let visible = $state(false);
+  let showTimeout = $state<ReturnType<typeof setTimeout> | null>(null);
   let hideTimeout = $state<ReturnType<typeof setTimeout> | null>(null);
 
   function show() {
@@ -33,10 +39,21 @@
       clearTimeout(hideTimeout);
       hideTimeout = null;
     }
-    visible = true;
+
+    if (visible || showTimeout) return;
+
+    showTimeout = setTimeout(() => {
+      showTimeout = null;
+      visible = true;
+    }, delay);
   }
 
   function hide() {
+    if (showTimeout) {
+      clearTimeout(showTimeout);
+      showTimeout = null;
+    }
+
     hideTimeout = setTimeout(() => {
       visible = false;
     }, 100);
--- a/svelte/src/lib/components/Tooltip.svelte
+++ b/svelte/src/lib/components/Tooltip.svelte
@@ -13,6 +13,12 @@
      * the tooltip. Defaults to `0`, which shows it immediately.
      */
     delay?: number;
+    /**
+     * Only show the tooltip when the anchor element's content is horizontally
+     * truncated (e.g. clipped by `text-overflow: ellipsis`). Useful to avoid a
+     * redundant tooltip that merely repeats fully visible text.
+     */
+    onlyWhenTruncated?: boolean;
   }
 
   interface TextProps extends BaseProps {
@@ -27,7 +33,7 @@
 
   type Props = TextProps | ChildrenProps;
 
-  let { text, children, side = 'top', delay = 0 }: Props = $props();
+  let { text, children, side = 'top', delay = 0, onlyWhenTruncated = false }: Props = $props();
 
   let anchorElement = $state<HTMLElement | null>(null);
   let visible = $state(false);
@@ -35,6 +41,13 @@
   let hideTimeout = $state<ReturnType<typeof setTimeout> | null>(null);
 
   function show() {
+    // Sub-pixel rounding can leave `scrollWidth` 1px above `clientWidth`
+    // without any visible clipping, so require more than that before treating
+    // the content as truncated.
+    if (onlyWhenTruncated && anchorElement && anchorElement.scrollWidth - anchorElement.clientWidth <= 1) {
+      return;
+    }
+
     if (hideTimeout) {
       clearTimeout(hideTimeout);
       hideTimeout = null;
--- a/svelte/src/lib/components/frontpage/CrateLists.stories.svelte
+++ b/svelte/src/lib/components/frontpage/CrateLists.stories.svelte
@@ -9,10 +9,11 @@
     tags: ['autodocs'],
   });
 
-  function createMockCrate(name: string, version: string) {
+  function createMockCrate(name: string, version: string, description = `${name} is a useful Rust crate`) {
     return {
       id: name,
       name,
+      description,
       newest_version: version,
       max_version: version,
       downloads: 1_000_000,
--- a/svelte/src/lib/components/frontpage/ListItem.svelte
+++ b/svelte/src/lib/components/frontpage/ListItem.svelte
@@ -15,20 +15,34 @@
   }
 
   let { title, subtitle, href, version, downloads, class: className, ...restProps }: Props = $props();
+
+  // The accessible name is restricted to the crate name via `aria-labelledby`
+  // so the screen reader links list stays scannable; the subtitle and trailing
+  // value are exposed as a supplementary `aria-describedby` description that is
+  // announced after the name instead of being concatenated into it.
+  const uid = $props.id();
+  const titleId = `${uid}-title`;
+  const subtitleId = `${uid}-subtitle`;
+  const trailingId = `${uid}-trailing`;
+
+  let hasTrailing = $derived(Boolean(version) || downloads != null);
+  let describedBy = $derived(
+    [subtitle ? subtitleId : null, hasTrailing ? trailingId : null].filter(Boolean).join(' ') || undefined,
+  );
 </script>
 
 <!-- eslint-disable-next-line svelte/no-navigation-without-resolve -->
-<a {href} class={['box', className]} {...restProps}>
+<a {href} class={['box', className]} aria-labelledby={titleId} aria-describedby={describedBy} {...restProps}>
   <div class="left">
-    <div class="title">{title}</div>
-    {#if subtitle}<div class="subtitle">{subtitle}</div>{/if}
+    <div class="title" id={titleId} data-test-title>{title}</div>
+    {#if subtitle}<div class="subtitle" id={subtitleId} data-test-subtitle>{subtitle}</div>{/if}
   </div>
   {#if version}
-    <div class="version">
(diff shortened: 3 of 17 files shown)


LawnGnome:
(Not doing a full code review right now; let's agree on what we want to do design-wise first.)

I think the download count change is unambiguously good, and we should do it. Moving the version numbers to visually match then makes sense as well.

I'm less sure about showing the descriptions. On the one hand, they're potentially useful for the new crate/version columns. On the other hand, they're truncated so short that I'm not sure they're actually useful in a lot of cases, and they do add a bit of visual clutter.

At the very least, I think we'd need a tooltip for truncated descriptions so people don't have to click all the way through.

Interested in other thoughts from @rust-lang/crates-io.

Turbo87:
> At the very least, I think we'd need a tooltip for truncated descriptions so people don't have to click all the way through.

yeah, I think a tooltip might be nice. but... a) the shortening with ellipsis is done by CSS, so we don't know which cases need the tooltip and which don't (e.g. `syn` in the screenshot) and b) while you've hovering with the mouse waiting for the tooltip that doesn't appear Svelte already loads the necessary data in the background so once you decide to click the page basically loads instantly 😅

LawnGnome:
> so once you decide to click the page basically loads instantly 😅

Sure, but that still means the user has switched context, even if it is instant.

> a) the shortening with ellipsis is done by CSS, so we don't know which cases need the tooltip and which don't (e.g. `syn` in the screenshot)

I think we can detect this with `scrollWidth`, if we wanted to go that way.

Turbo87:
> I think we can detect this with `scrollWidth`, if we wanted to go that way.

hmm interesting idea. I'll play around with that and see if I can make that work 👍

eth3lbert:
> I think the download count change is unambiguously good, and we should do it. Moving the version numbers to visually match then makes sense as well.
> 
> I'm less sure about showing the descriptions. On the one hand, they're potentially useful for the new crate/version columns. On the other hand, they're truncated so short that I'm not sure they're actually useful in a lot of cases, and they do add a bit of visual clutter.

Yeah, I have the same feeling! The descriptions are truncated so short that I suspect anyone who is interested will just click to see the details anyway. Also, tooltip might not be useful on mobile, and I'm not sure how well they perform or how necessary they are for accessibility. But all the other parts make complete sense to me!

Turbo87:
> Also, tooltip might not be useful on mobile

true, but on mobile we also switch to a single column which increases the description length that is shown :D

> The descriptions are truncated so short that I suspect anyone who is interested will just click to see the details anyway

how do you know if you're interested if you only see the crate name? example: for `syn` I wouldn't know what this crate is about just from the name. the "Parser for Rust source code" description below it makes it a lot easier to understand that at first glance. and even if it had a longer description the truncated portion is often enough to get a first idea of whether it might be interesting or not, and for everything else we'll have the tooltips that I'm currently implementing :)

eth3lbert:
> how do you know if you're interested if you only see the crate name? example: for `syn` I wouldn't know what this crate is about just from the name. the "Parser for Rust source code" description below it makes it a lot easier to understand that at first glance. and even if it had a longer description the truncated portion is often enough to get a first idea of whether it might be interesting or not, and for everything else we'll have the tooltips that I'm currently implementing :)

Yeah, fair enough! I mostly rely on the search function to find crates, and it does indeed show truncated descriptions. So yeah, I have no objections to this since it makes a lot of sense functionally!!

Turbo87:
grafik

I've implemented it now in a way that it shows the tooltip only if the description is truncated. It shows after 350ms and below the truncated description to not hide the crate name itself.

svelte/src/lib/components/Tooltip.svelte
```diff
@@ -8,6 +8,17 @@
...
+     * Milliseconds to wait after the pointer enters the anchor before showing
+     * the tooltip. Defaults to `0`, which shows it immediately.
+     */
+    delay?: number;
+    /**
+     * Only show the tooltip when the anchor element's content is horizontally
+     * truncated (e.g. clipped by `text-overflow: ellipsis`). Useful to avoid a
+     * redundant tooltip that merely repeats fully visible text.
+     */
+    onlyWhenTruncated?: boolean;
```
eth3lbert, line 21: I haven't done a full review yet, but one thing I'm unsure about is adding this prop to the `Tooltip` component. To me, it makes more sense to conditionally determine whether to show the `Tooltip` from the parent container rather than inside the `Tooltip` itself, but that might just be me.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13827/751f0469562674d0e40046025e3a7291bc765f49..64379bbae44f5381e776bee4244a5e8eb4b3a7e1)*

Turbo87: I considered that, but from a DX perspective that seemed much more complicated to use. The `<Tooltip>` component only actually inserts the majority of the DOM elements when the anchor is hovered, so this shouldn't cause too many performance issues, I think.

svelte/src/lib/components/frontpage/CrateLists.stories.svelte
```diff
@@ -9,10 +9,11 @@
     tags: ['autodocs'],
   });
 
-  function createMockCrate(name: string, version: string) {
+  function createMockCrate(name: string, version: string, description = `${name} is a useful Rust crate`) {
```
LawnGnome, line 12: All Rust crates are useful!

_nervously shuffles a bunch of his own crates under a nearby rug_

svelte/src/lib/components/frontpage/ListItem.svelte
```diff
@@ -2,23 +2,60 @@
...
+  // value are exposed as a supplementary `aria-describedby` description that is
+  // announced after the name instead of being concatenated into it.
+  const uid = $props.id();
+  const titleId = `${uid}-title`;
+  const subtitleId = `${uid}-subtitle`;
+  const trailingId = `${uid}-trailing`;
+
+  let hasTrailing = $derived(Boolean(version) || downloads != null);
+  let describedBy = $derived(
+    [subtitle ? subtitleId : null, hasTrailing ? trailingId : null].filter(Boolean).join(' ') || undefined,
```
LawnGnome: You can say no, but this combination of an array literal and ternaries is kind of breaking my reading comprehension, so what about this as a gift to future me:

```suggestion
    let describedByIds = [];
    
    if (subtitle) {
      describedByIds.push(subtitleId);
    }
    if (hasTrailing) {
      describedByIds.push(trailingId);
    }
    
    if (describedByIds.length > 0) {
      return describedByIds;
    }
```

svelte/src/lib/components/Tooltip.svelte
LawnGnome: I don't love this either, honestly — it's unfortunate that `<Tooltip />` has to have knowledge of its parent at all, and more unfortunate that we're going to be extending that.

I'm OK with this as a pragmatic choice, but I'm also reserving the right to get mad about this and open a PR later. 😆

svelte/src/lib/components/Tooltip.svelte
```diff
@@ -22,21 +33,40 @@
...
   let anchorElement = $state<HTMLElement | null>(null);
   let visible = $state(false);
+  let showTimeout = $state<ReturnType<typeof setTimeout> | null>(null);
   let hideTimeout = $state<ReturnType<typeof setTimeout> | null>(null);
 
   function show() {
+    // Sub-pixel rounding can leave `scrollWidth` 1px above `clientWidth`
+    // without any visible clipping, so require more than that before treating
+    // the content as truncated.
+    if (onlyWhenTruncated && anchorElement && anchorElement.scrollWidth - anchorElement.clientWidth <= 1) {
```
LawnGnome, line 47: If we get any weirdness here, I'd actually be OK having the fudge factor be more than 1px, but let's see how we go.

svelte/src/lib/components/Tooltip.svelte
```diff
@@ -22,21 +33,40 @@
...
       hideTimeout = null;
     }
-    visible = true;
+
+    if (visible || showTimeout) return;
+
+    showTimeout = setTimeout(() => {
+      showTimeout = null;
+      visible = true;
+    }, delay);
```
LawnGnome, line 61: It won't matter for any of the uses we have today, but does this need to be `$derived` because of the use of `delay` (and other props, but that was the one that stood out to me)?

svelte/src/lib/utils/format-short-num.ts
```diff
@@ -1,17 +1,25 @@
-const THRESHOLD = 1500;
-const UNITS = ['', 'K', 'M'];
+// A four-digit mantissa rounds up to "1,000", so rolling over at this value
+// keeps the displayed number at three digits or fewer within each unit.
+const DEFAULT_THRESHOLD = 999.5;
+const UNITS = ['', 'K', 'M', 'B'];
```
LawnGnome, line 4: Resisting the urge to suggest a `T` entry with the rise in crates.io usage this year. 🙈

LawnGnome (approved):
I'm still a little concerned about the potential visual clutter aspect of this, but with the tooltip behaviour I think I'm now OK with us trying this out and seeing how it goes over.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13827/751f0469562674d0e40046025e3a7291bc765f49..64379bbae44f5381e776bee4244a5e8eb4b3a7e1)*

svelte/src/lib/components/Tooltip.svelte
Turbo87: > I'm also reserving the right to get mad about this and open a PR later. 😆

feel free. it's mostly implemented this way at the moment because during the migration I was trying to mirror the way the Ember.js addon worked. if we find better ways to do this then I'm not opposed :D

svelte/src/lib/components/Tooltip.svelte
Turbo87: technically, yes. practically, that would complicate the whole implementation quite a bit. 

IMHO it's fine to read `delay` as "the value of `delay` at the time of hovering the anchor parent element" and not as "the value of `delay` right now". 

so far `delay` is only used in this PR, and I'd be surprised if we ended up with a case where the `delay` argument intentionally changes while the `setTimeout` is already running 😅

svelte/src/lib/utils/format-short-num.ts
Turbo87: `syn` is at 1.7B now, and `hashbrown` is at 442M over the past 90 days, so I think it's probably fine for now 🤞

PR #13828, opened 2026-06-02
Replace the global `Vary` header with scoped cache headers

Turbo87:
The common-headers middleware applied `Vary: Accept, Accept-Encoding, Cookie` to almost every response. `Cookie` is high cardinality, so a shared CDN cache keys a separate entry per session and effectively stops caching, which defeated the ten-year `Cache-Control` on the content-hashed frontend assets and prevented public API responses from being cached. `Accept` has the same problem to a smaller degree, and neither value was needed on the bulk of responses.

This replaces the blanket header with values scoped to where they matter. Responses that depend on who is authenticated now set `Cache-Control: no-store`, so neither a shared cache nor a browser stores them. That covers the `/me` family, API token listing, crate owner invitations, trusted publishing configuration, the crate following status, the session routes, and crate search when it is filtered by `?following=`. `no-store` rather than `Vary: Cookie` is deliberate, because authentication can come from an API token as well as a session cookie, and because it also keeps one user's response out of the next person's browser cache on a shared computer.

The two endpoints that genuinely vary on the request `Accept` header, the crate download and readme redirects, now set `Vary: Accept` themselves. The static asset responses set `Vary: Accept-Encoding`, which `ServeDir` does not emit for precompressed variants yet ([tower-rs/tower-http#692](https://github.com/tower-rs/tower-http/pull/692)), so dropping the global header would otherwise let a shared cache hand a Brotli file to a client that only accepts gzip. With those cases handled, the global header is gone, and compressed responses still pick up `Vary: Accept-Encoding` from the compression layer.

The per-user set was found by auditing every GET handler for responses that change with the authenticated identity. A new `caching` integration module asserts these cases, along with a couple of public endpoints as smoke tests.

### Related

- https://github.com/rust-lang/crates.io/pull/910
- https://github.com/tower-rs/tower-http/pull/692

### DIFF
--- a/src/controllers/admin.rs
+++ b/src/controllers/admin.rs
@@ -4,8 +4,11 @@ use crate::{
     models::{OwnerKind, User},
     schema::*,
     util::errors::{AppResult, custom},
+    util::no_store,
 };
 use axum::{Json, extract::Path};
+use axum_extra::TypedHeader;
+use axum_extra::headers::CacheControl;
 use chrono::{DateTime, Utc};
 use diesel::{dsl::count_star, prelude::*};
 use diesel_async::RunQueryDsl;
@@ -63,7 +66,7 @@ pub async fn list(
     state: AppState,
     Path(username): Path<String>,
     req: Parts,
-) -> AppResult<Json<AdminListResponse>> {
+) -> AppResult<(TypedHeader<CacheControl>, Json<AdminListResponse>)> {
     let mut conn = state.db_read().await?;
 
     let auth = AuthCheck::default().check(&req, &mut conn).await?;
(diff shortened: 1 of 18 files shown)


LawnGnome (approved):
LGTM.

I have a very slight, overly paranoid voice in the back of my head wondering if we're accidentally relying on this behaviour to avoid stale responses somehow, but I think the easiest way to figure that out will be to do this, hopefully improve our CDN usage, and then see if anyone has problems.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13828/810dae2bc5e25acc2895626b61c2bda62c84ca7a..a46e0154555adaffec666c878823c40f62788fd5)*

Turbo87:
yeah... I share that feeling 😂

PR #13829, opened 2026-06-02
session: Accept `POST` on the OAuth endpoints

Turbo87:
The `/api/private/session/begin` and `/api/private/session/authorize` endpoints are served over `GET`, but both have side effects. `begin` stores a freshly generated CSRF token in the session, and `authorize` exchanges a single-use OAuth code, writes the user record, and logs the user in. Neither is safe or idempotent in the HTTP sense, and `authorize` additionally carries the single-use `code` in the query string, where it ends up in server logs, browser history, and `Referer` headers.

This adds `POST` variants as the canonical endpoints. `authorize` reads `code` and `state` from a JSON body so the code no longer leaks into URLs. The frontend (loading page and session store) and the login e2e tests move to `POST` in the same PR.

The `GET` variants are retained and marked deprecated in the OpenAPI document so that frontends already cached in browsers keep working across the deploy. A follow-up PR removes them once no live client relies on `GET`.

### DIFF
--- a/e2e/acceptance/login.spec.ts
+++ b/e2e/acceptance/login.spec.ts
@@ -35,11 +35,11 @@ test.describe('Acceptance | Login', { tag: '@acceptance' }, () => {
     let user = await msw.db.user.create({ name: 'John Doe' });
 
     msw.worker.use(
-      http.get('/api/private/session/authorize', async ({ request }) => {
-        let url = new URL(request.url);
-        expect([...url.searchParams.keys()]).toEqual(['code', 'state']);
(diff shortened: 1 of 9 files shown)


walterhpearce:
Lgtm

LawnGnome (approved):
Looks good to me as well. :+1:

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13829/c00969f7789d9b562f2d66e21c752fce7f855e84..119f71c0431dc31fca4ed06414ed8693300ccfa4)*

PR #13839, opened 2026-06-03
Fix dropdown trigger icon size in version list rows

Turbo87:
The UnoCSS icon migration left the actions dropdown trigger icon smaller than it was before. This sets it back to 2em to match the previous size.

### DIFF
--- a/svelte/src/lib/components/version-list/Row.svelte
+++ b/svelte/src/lib/components/version-list/Row.svelte
@@ -378,6 +378,11 @@
         color: var(--grey900);
         background-color: white;
       }
+
+      :global(.icon) {
+        width: 2em;
+        height: 2em;
+      }
     }
 
     :global(.menu) {


eth3lbert (commented):
Would it make more sense to control the size via `font-size`, given that the width and height are both 1em?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13839/751f0469562674d0e40046025e3a7291bc765f49..f267f89a506fd22f707a23650ecf90a244b9ed58)*

Turbo87:
that's actually a good idea and I'm a bit embarrassed that I haven't thought of that... 😂 

I think it might even be better if we used something like `<Icon size="2em">` and then potentially `font-size` inside to apply the `size` prop.

unless you have objections, I'd keep this PR as is though, for consistency reasons, since that is how the rest of the codebase currently scales the icons :)

eth3lbert (approved):
> unless you have objections, I'd keep this PR as is though, for consistency reasons, since that is how the rest of the codebase currently scales the icons :)

I was aware of the codebase situation when I left that comment. I have no objections, and we can easily retrofit this later on.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13839/751f0469562674d0e40046025e3a7291bc765f49..f267f89a506fd22f707a23650ecf90a244b9ed58)*

Turbo87:
- https://github.com/rust-lang/crates.io/pull/13841 :)

PR #13841, opened 2026-06-03
Size icons via `--icon-size` instead of `:global(.icon)`

Turbo87:
We had ~30 `:global(.icon)` blocks across the frontend whose only job was to override an icon's size. Reaching into the `Icon` component's `.icon` class with `:global` is awkward and easy to get wrong.

This adds an `--icon-size` custom property to `Icon`. The icon renders as a `1em` box (UnoCSS preset-icons default), so `.icon` now drives `font-size` from `var(--icon-size, 1em)`. Parents size an icon by setting the property on a containing element, which inherits through the component boundary without `:global` and works inside media queries.

Each call site is converted in its own commit. Where a `:global(.icon)` block only set the size, it is removed entirely. Where it also set non-inheritable properties (margin, opacity, vertical-align) or a `color` that would otherwise tint sibling text, those stay in `:global` and only the size moves to the property.

After the sweep no `:global(.icon)` block sets `width`/`height` anymore.

### DIFF
--- a/svelte/src/lib/components/ColorSchemeMenu.svelte
+++ b/svelte/src/lib/components/ColorSchemeMenu.svelte
@@ -57,11 +57,7 @@
   .color-scheme-menu {
     display: flex;
     align-items: center;
-
-    & :global(.icon) {
-      width: 1.4em;
-      height: 1.4em;
-    }
+    --icon-size: 1.4em;
 
     & :global(.trigger) {
       background: none;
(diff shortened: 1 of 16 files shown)


eth3lbert (approved):
Awesome!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13841/083c5c7144df45dba76cd44b504fd6f63ea58850..1c61c1b20e312addd4f50b0f73d38dec5e9975b5)*

PR #13855, opened 2026-06-04
Suggest standard-library replacements on crate pages

Turbo87:
When a crate's functionality is now available in the standard library, this shows a banner on the crate page and a small marker in dependency lists, each pointing the reader at the relevant `std` API. The suggestion is deliberately limited to the standard library and never compares third-party crates against each other, so every entry stays a checkable factual statement about `std` rather than an opinion.

The data lives in a curated module at `svelte/src/lib/data/native-replacements.ts`, which maps a crate name to one or more `std` targets. Each target records the Rust version that stabilized it and a documentation URL, and each entry carries a flag for whether `std` covers the crate fully or only its common case, alongside a short curated description. It ships seeded with `lazy_static`, `once_cell`, `matches`, and `num_cpus`. The rules for what qualifies sit right next to the data as a module-level doc comment, covering the limited scope, the bar for full and partial entries, and a step where the maintainer of a flagged crate can weigh in before the entry lands.

The banner appears on the crate page for every version of a flagged crate, and the dependency list shows a marker whose tooltip carries the same description and links straight to the `std` item. Both surfaces are visible in the `NativeReplacementBanner` and `dependency-list/Row` Storybook stories, and they are covered by component tests and end-to-end tests for the present and absent cases.

### Crate page screenshot
grafik

### Dependency list screenshot
grafik

### Related

- Original inspiration: https://npmx.dev/package/left-pad
- https://github.com/e18e/module-replacements
- https://github.com/npmx-dev/npmx.dev/blob/01c510088489b3988586342af3c6df0ad48d7cdb/app/components/Package/Replacement.vue
- https://nesbitt.io/2026/04/16/features-everyone-should-steal-from-npmx.html
- https://rust-lang.zulipchat.com/#narrow/channel/318791-t-crates-io/topic/std.20lib.20replacements/with/600131994

### DIFF
--- a/svelte/src/lib/data/native-replacements.ts
+++ b/svelte/src/lib/data/native-replacements.ts
@@ -58,8 +58,8 @@ export const nativeReplacements: Record<string, NativeReplacement> = {
   once_cell: {
     description:
       'Most of <code>once_cell</code> is now in <code>std</code>: <code>OnceCell</code>/<code>OnceLock</code> ' +
-      '(Rust 1.70) and <code>LazyCell</code>/<code>LazyLock</code> (Rust 1.80). A few APIs (e.g. fallible ' +
-      '<code>get_or_try_init</code>) are not yet stable in <code>std</code>.',
+      '(Rust 1.70) and <code>LazyCell</code>/<code>LazyLock</code> (Rust 1.80). The <code>race</code> module and ' +
+      'fallible <code>get_or_try_init</code> have no stable <code>std</code> equivalent yet.',
     url: 'https://blog.rust-lang.org/2024/07/25/Rust-1.80.0/',
   },
 
--- a/svelte/src/lib/data/native-replacements.ts
+++ b/svelte/src/lib/data/native-replacements.ts
@@ -1,3 +1,43 @@
+/**
+ * Native replacement policy
+ * =========================
+ *
+ * This dataset flags crates whose functionality has been absorbed into the
+ * Rust standard library, so a reader can see that `std` now provides what the
+ * crate offers. Every entry is a checkable factual statement about `std`, never
+ * an editorial judgement about a competing third-party crate.
+ *
+ * Scope
+ * -----
+ * Standard-library replacements only. The set deliberately does not include
+ * "prefer this nicer crate instead" style recommendations.
+ *
+ * Inclusion criteria
+ * ------------------
+ * Two kinds of entries qualify:
+ *
+ *   - Full: the crate's functionality is entirely available via a stable `std`
+ *     API.
+ *   - Partial: the bulk of the crate's common use case has moved to `std`, but
+ *     not all of it. The `description` must spell out what is still missing.
+ *
+ * Coverage is judged roughly, and only the dominant use case counts. If just a
+ * small slice of a crate's purpose lives in `std`, it does not qualify, since
+ * flagging it would be misleading and noisy. `itertools` is the canonical
+ * exclusion: a few adaptors have `std` equivalents, but the crate is
+ * overwhelmingly not replaced by `std`.
+ *
+ * Every entry's `description` must cite the stable `std` API(s) and the Rust
+ * version(s) that stabilized them.
+ *
+ * Maintainer notice-and-comment
+ * -----------------------------
+ * Entries should arrive as PRs judged against the criteria above. Before an entry
+ * lands, the maintainer(s) of the crate being flagged get a window to weigh in,
+ * either to object or to show that the replacement is less complete than claimed,
+ * in which case the entry becomes partial or is rejected.
+ */
+
 /** A crate whose functionality is (largely) available in std. */
 export interface NativeReplacement {
   /** HTML fragment describing the std replacement. */
--- a/svelte/src/lib/data/native-replacements.ts
+++ b/svelte/src/lib/data/native-replacements.ts
@@ -1,0 +1,38 @@
+/** A crate whose functionality is (largely) available in std. */
+export interface NativeReplacement {
+  /** HTML fragment describing the std replacement. */
+  description: string;
+  /** Representative docs URL (std docs or release notes) shown as a "Learn more" link. */
+  url: string;
+}
+
+export const nativeReplacements: Record<string, NativeReplacement> = {
+  lazy_static: {
+    description:
+      'The standard library provides <code>std::sync::LazyLock</code> (stable since Rust 1.80), which lets ' +
+      'you replace the <code>lazy_static!</code> macro with a plain <code>static</code> and remove the ' +
+      'dependency.',
+    url: 'https://doc.rust-lang.org/std/sync/struct.LazyLock.html',
+  },
+
+  once_cell: {
+    description:
+      'Most of <code>once_cell</code> is now in <code>std</code>: <code>OnceCell</code>/<code>OnceLock</code> ' +
+      '(Rust 1.70) and <code>LazyCell</code>/<code>LazyLock</code> (Rust 1.80). A few APIs (e.g. fallible ' +
+      '<code>get_or_try_init</code>) are not yet stable in <code>std</code>.',
+    url: 'https://blog.rust-lang.org/2024/07/25/Rust-1.80.0/',
+  },
+
+  matches: {
+    description: 'The <code>matches!</code> macro has been part of the standard library since Rust 1.42.',
+    url: 'https://doc.rust-lang.org/std/macro.matches.html',
+  },
+
+  num_cpus: {
+    description:
+      'For most uses, <code>std::thread::available_parallelism</code> (Rust 1.59) returns the parallelism ' +
+      'available to the process. (It does not distinguish physical vs logical cores the way ' +
+      '<code>num_cpus</code> does.)',
+    url: 'https://doc.rust-lang.org/std/thread/fn.available_parallelism.html',
+  },
+};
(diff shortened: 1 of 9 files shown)


svelte/src/lib/data/native-replacements.ts
```diff
@@ -0,0 +1,161 @@
...
+ * Every entry must cite the stable `std` API(s) and the Rust version(s) that
+ * stabilized them via the `targets` field.
+ *
+ * Maintainer notice-and-comment
+ * -----------------------------
+ * Entries should arrive as PRs judged against the criteria above. Before an entry
+ * lands, the maintainer(s) of the crate being flagged get a window to weigh in,
+ * either to object or to show that supersession is less complete than claimed,
+ * in which case the entry becomes partial or is rejected.
+ */
```
Turbo87, line 39: I've included a basic policy on what replacements should be added to the list.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13855/570ab4fc0a0aa71cbebb092c0e37b1411f53741f..33e0e9a2c4f8bf5215d41e1dd32169f719b86803)*

svelte/src/lib/data/native-replacements.ts
```diff
@@ -0,0 +1,161 @@
...
+ * otherwise `undefined`.
+ */
+export function primaryUrl(replacement: NativeReplacement): string | undefined {
+  if (replacement.url) return replacement.url;
+  if (replacement.targets.length === 1) return replacement.targets[0].url;
+  return undefined;
+}
+
+/** Keyed by exact crate name for O(1) lookup. */
+export const nativeReplacements: Record<string, NativeReplacement> = {
```
Turbo87, line 49: This contains a basic first set of replacements. I hope I characterized them correctly. Let me know if you find any mistakes!

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13855/570ab4fc0a0aa71cbebb092c0e37b1411f53741f..33e0e9a2c4f8bf5215d41e1dd32169f719b86803)*

svelte/src/lib/data/native-replacements.ts
```diff
@@ -0,0 +1,161 @@
...
+        path: 'std::sync::LazyLock',
+        stabilizedIn: '1.80.0',
+        url: 'https://doc.rust-lang.org/std/sync/struct.LazyLock.html',
+      },
+      {
+        path: 'std::cell::LazyCell',
+        stabilizedIn: '1.80.0',
+        url: 'https://doc.rust-lang.org/std/cell/struct.LazyCell.html',
+      },
+    ],
```
Turbo87: I wonder if this kind of structured data is actually needed for our first implementation, since we only use `description` and `url` anyway 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13855/570ab4fc0a0aa71cbebb092c0e37b1411f53741f..33e0e9a2c4f8bf5215d41e1dd32169f719b86803)*

LawnGnome:
I'd like to talk about this at tomorrow's meeting. (Or on Zulip, but since we have a meeting tomorrow anyway, let's start there.)

LawnGnome (approved):
No issues on the code.

If we're showing a tooltip on dep rows, should we also be showing a tooltip on search results?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13855/c163b371c13f1de49b7e17f5d13199a7a66d7c25..aa894de0dbe62e80643132810dbc75f2e0512231)*

Turbo87:
> If we're showing a tooltip on dep rows, should we also be showing a tooltip on search results?

I'm having a hard time expressing it properly, but I think the goals and target audiences are slightly different between these

svelte/src/lib/data/native-replacements.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const nativeReplacements: Record<string, NativeReplacement> = {
+  lazy_static: {
+    description:
+      'The standard library provides <code>std::sync::LazyLock</code> (stable since Rust 1.80), which lets ' +
+      'you replace the <code>lazy_static!</code> macro with a plain <code>static</code> and remove the ' +
+      'dependency.',
+    url: 'https://doc.rust-lang.org/std/sync/struct.LazyLock.html',
+  },
```
Turbo87, line 56: @cuviper 👋 since you released the last version of `lazy_static`: does this seem accurate to you? any objections?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13855/0e652117edbd859a570aa980f6a2955dc0b2be94..8f23fe4f8957aff58d81d4849fae47b7acaa2257)*

svelte/src/lib/data/native-replacements.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  once_cell: {
+    description:
+      'Most of <code>once_cell</code> is now in <code>std</code>: <code>OnceCell</code>/<code>OnceLock</code> ' +
+      '(Rust 1.70) and <code>LazyCell</code>/<code>LazyLock</code> (Rust 1.80). A few APIs (e.g. fallible ' +
+      '<code>get_or_try_init</code>) are not yet stable in <code>std</code>.',
+    url: 'https://blog.rust-lang.org/2024/07/25/Rust-1.80.0/',
+  },
```
Turbo87, line 64: @matklad 👋 does this seem accurate to you? any objections?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13855/0e652117edbd859a570aa980f6a2955dc0b2be94..8f23fe4f8957aff58d81d4849fae47b7acaa2257)*

svelte/src/lib/data/native-replacements.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  matches: {
+    description: 'The <code>matches!</code> macro has been part of the standard library since Rust 1.42.',
+    url: 'https://doc.rust-lang.org/std/macro.matches.html',
+  },
```
Turbo87, line 71: @SimonSapin 👋 does this seem accurate to you? any objections?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13855/0e652117edbd859a570aa980f6a2955dc0b2be94..8f23fe4f8957aff58d81d4849fae47b7acaa2257)*

svelte/src/lib/data/native-replacements.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  num_cpus: {
+    description:
+      'For most uses, <code>std::thread::available_parallelism</code> (Rust 1.59) returns the parallelism ' +
+      'available to the process. (It does not distinguish physical vs logical cores the way ' +
+      '<code>num_cpus</code> does.)',
+    url: 'https://doc.rust-lang.org/std/thread/fn.available_parallelism.html',
+  },
```
Turbo87, line 79: @seanmonstar 👋 does this seem accurate to you? any objections?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13855/0e652117edbd859a570aa980f6a2955dc0b2be94..8f23fe4f8957aff58d81d4849fae47b7acaa2257)*

svelte/src/lib/data/native-replacements.ts
SimonSapin: Accurate but incomplete: `assert_matches` and `debug_assert_matches` are also in std since 1.96.0:

* https://doc.rust-lang.org/std/macro.assert_matches.html
* https://doc.rust-lang.org/std/macro.debug_assert_matches.html

This covers everything in the crate

Turbo87: thank you, I'll fix the message :)

andrew:
@Turbo87 would you consider extracting the data in native-replacements.ts into a json file that others could consume as an api (or read out of the git repo if you don't want to commit to shipping it as an api)

Turbo87:
@andrew yeah, that is the goal, but we wanted to keep it simple for the PoC for now. once this is accepted and merged we can refactor it into static files :)

svelte/src/lib/data/native-replacements.ts
cuviper: Well I haven't really been involved in its maintenance -- I just published it on their behalf since I have T-libs access. @KodrAus is probably a better person to represent the crate itself.

But still, yes I think that's accurate, and FWIW there's a similar section in its README:
https://github.com/rust-lang-nursery/lazy-static.rs#standard-library

KodrAus: :+1: For `lazy_static` specifically I think this is accurate. I’m a little less sure about using this mechanism for cases that don’t have complete unambiguous replacements like `once_cell`.

svelte/src/lib/data/native-replacements.ts
matklad: Oh, hi, @Turbo87, it's been a while! 

The info looks most accurate, and I am very much in favor of surfacing this to the users in an unobtrusive way (so that `get_or_try_init` users aren't spammed, which I think the current implementation achieves). One more reason to use crates.io module is the `race` module for racy initialization (two threads run the `init` function, one wins), I don't think we have that in `std`.

Turbo87:
@andrew https://github.com/rust-lang/std-replacement-data (and https://rust-lang.github.io/std-replacement-data/all.json). I'm not comfortable yet giving hard stability guarantees, but for now this should be somewhat stable.

PR #13832, opened 2026-06-02
WIP removing more gh_ fields from user

carols10cents:
i'm getting an insta failure locally; running in CI to see if it's a personal problem or not

this PR is currently a mess; don't bother reviewing it until i've had a chance to clean up 😅

PR #13856, opened 2026-06-04
Switch session create to check `oauth_github` first, then get user

carols10cents:
This is what I was talking about [in this comment](https://github.com/rust-lang/crates.io/issues/13758#issuecomment-4578523078), that we can't use nice upserts but need to do multiple update/select/inserts in a transaction instead, so that we switch to using `oauth_github` as the source of truth of whether we're logging in an existing user or a new user.

I've extracted this from https://github.com/rust-lang/crates.io/pull/13832 and I think making this change first is needed before we can stop using `users.{gh_id, gh_avatar, gh_encrypted_token}`.


src/controllers/session.rs
```diff
@@ -210,28 +211,84 @@ async fn authorize(
...
+                // we know the associated user record.
+
+                // For now, update user display name, gh_login, and username. Eventually, we will
+                // get rid of `gh_login` and stop syncing `name` and `username` with GitHub.
+                diesel::update(users::table)
+                    .filter(users::id.eq(oauth_github.user_id))
+                    .set((
+                        users::name.eq(gh_user.name.as_ref()),
+                        users::gh_login.eq(&gh_user.login),
+                        // users::username.eq(&gh_user.login),
```
carols10cents: This will need to be uncommented if https://github.com/rust-lang/crates.io/pull/13781 gets in first

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/0f3189cc9df240ef372981d5b4d3931da61dfbe9..97aa9a7254c31378271910d1cd28e24e6aef4ba1)*

src/controllers/session.rs
```diff
@@ -210,28 +211,84 @@ async fn authorize(
...
+                // crates.io users to link more than one GitHub account to their crates.io account.
+                let new_user = NewUser::builder()
+                    .gh_id(gh_user.id)
+                    .gh_login(&gh_user.login)
+                    .maybe_name(gh_user.name.as_deref())
+                    .gh_encrypted_token(encrypted_token)
+                    .build();
+
+                let user_id =
+                    create_or_update_user(&new_user, gh_user.email.as_deref(), emails, conn)
```
carols10cents: `create_or_update_user` really should only be creating, i think. And it starts its own transaction, but i don't think diesel/postgres cares about nested transactions.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/0f3189cc9df240ef372981d5b4d3931da61dfbe9..97aa9a7254c31378271910d1cd28e24e6aef4ba1)*

Turbo87: > but i don't think diesel/postgres cares about nested transactions

it does:

> If there is already an open transaction, savepoints will be used instead.

(from https://docs.diesel.rs/1.4.x/diesel/connection/trait.Connection.html#method.transaction)

src/controllers/session.rs
```diff
@@ -210,28 +211,84 @@ async fn authorize(
...
+                    .execute(conn)
+                    .await?;
+
+                Ok(user_id)
+            }
+            Err(error) if is_read_only_error(&error) => {
+                // If we're in read only mode, we can't update their details
+                // just look for an existing user
+                find_user_by_gh_id(conn, gh_user.id).await?.ok_or(error)
+            }
```
Turbo87: this change has a hidden regression that becomes visible with https://github.com/rust-lang/crates.io/pull/13862.

previously the nested transactions caused the "read-only error -> use read-only query instead" case to work, but now the transaction is poisoned and won't work anymore for the read-only query either.

and in case you think I'm clever: nope, this was found by Claude 😅 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/0f3189cc9df240ef372981d5b4d3931da61dfbe9..97aa9a7254c31378271910d1cd28e24e6aef4ba1)*

Turbo87 (changes requested):

carols10cents:
@Turbo87 Ok, I rebased to pick up your new test and then thought a lot about the transactions to fix the problem. I _think_ the transactions don't need to be nested, but as I've noted in the commit message of the last commit and in comments, we're going to have to rethink how the transactions and errors and inserts and updates work when the users table bases its uniqueness on `users.username` rather than `users.gh_id`.

Let me know if you/Claude think this is correct now!

carols10cents:
@Turbo87 this is ready for re-review!

src/controllers/session.rs
```diff
@@ -136,26 +138,65 @@ pub async fn authorize_session(
...
+                ))
+                .get_result::<OauthGithub>(conn)
+                .await?;
+            diesel::update(users::table)
+                .filter(users::id.eq(oauth_github.user_id))
+                .set((
+                    users::name.eq(gh_user.name.as_ref()),
+                    users::gh_login.eq(&gh_user.login),
+                    users::username.eq(&gh_user.login),
+                ))
```
Turbo87: I think this is missing at least `gh_encrypted_token` (maybe `gh_avatar` too?), as long as that column still exists. I guess it might be valuable to have a test that asserts that all of these columns of the two tables are correctly updated.

src/controllers/session.rs
```diff
@@ -166,27 +207,47 @@ pub async fn save_user_to_database(
...
     emails: &Emails,
     conn: &mut AsyncPgConnection,
 ) -> QueryResult<i32> {
+    let new_user = NewUser::builder()
+        .gh_id(gh_user.id)
+        .gh_login(&gh_user.login)
+        .username(&gh_user.login)
+        .maybe_name(gh_user.name.as_deref())
+        .gh_encrypted_token(encrypted_token)
+        .build();
```
Turbo87: the old code path was including `.maybe_gh_avatar(user.avatar_url.as_deref())`, but this one apparently isn't. given that we don't read from that column anymore, this change might be fine, but IMHO this should be a deliberate change in a dedicated commit/PR and probably not mixed up in the session creation changes.

src/tests/user.rs
```diff
@@ -427,3 +427,35 @@ async fn existing_user_can_log_in_during_read_only_mode() -> anyhow::Result<()>
...
+        .await?;
+
+    // Logging in as a new user can't work in read-only mode.
+    let result = session::save_user_to_database(&gh_user, b"token", emails, &mut conn).await;
+
+    let error = assert_err!(result);
+    assert_eq!(
+        error.to_string(),
+        "cannot execute UPDATE in a read-only transaction"
+    );
```
Turbo87: I'd recommend `assert_snapshot!(error, @"cannot execute UPDATE in a read-only transaction");` to make it easier to adjust the assertion if PostgreSQL happens to change their error message.

src/controllers/session.rs
```diff
@@ -166,27 +207,47 @@ pub async fn save_user_to_database(
...
+
     conn.transaction(async |conn| {
+        // This currently inserts or updates based on `users::gh_id` being unique. When we switch
+        // to using `users::username` for uniqueness instead, this logic will need to change.
+        // See note on [`NewUser#insert_or_update`].
```
Turbo87:
```suggestion
        // See note on [`NewUser::insert_or_update`].
```

or does `#` also work? 🤔 

but also, this isn't actually a doc comment so the link won't be rendered anywhere anyway 😅

Turbo87 (changes requested):
I'm fine with merging this once the issues are fixed, but I'm wondering whether we should prioritize splitting login from registration in the UI in some way. if we had dedicated endpoints for both operations then this probably would be less messy 😅 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/f762fd818d0d7bcf43da3029d7515e2cf4aec42c..053f67403474e34892177c0f2012962499a15a77)*

src/controllers/session.rs
carols10cents: well `users::gh_id` still exists too? i'll put them all back back until we officially deprecate them

src/controllers/session.rs
carols10cents: LOL i was writing rubydoc i guess??? 🤦🏻‍♀️

carols10cents:
@Turbo87 Ready for re-re-review! I rebased and added new commits, but didn't change any of the commits you've already looked at.

> I'm fine with merging this once the issues are fixed, but I'm wondering whether we should prioritize splitting login from registration in the UI in some way. if we had dedicated endpoints for both operations then this probably would be less messy 😅

omg yes 😭 filing a new issue for that now-- it's sort of related to some of the username issues but can be done first as a straight refactoring i think

src/controllers/session.rs
Turbo87: > well `users::gh_id` still exists too?

sure, but that value will not change as long as we have the one-to-one relationship between the accounts 😉

src/controllers/session.rs
```diff
@@ -136,26 +138,69 @@ pub async fn authorize_session(
...
+                .get_result::<OauthGithub>(conn)
+                .await?;
+            diesel::update(users::table)
+                .filter(users::id.eq(oauth_github.user_id))
+                .set((
+                    users::name.eq(gh_user.name.as_ref()),
+                    users::username.eq(&gh_user.login),
+                    // These fields are soon to be deprecated.
+                    users::gh_login.eq(&gh_user.login),
+                    users::gh_id.eq(&gh_user.id),
```
Turbo87:
```suggestion
```

as mentioned in the other comment, the GitHub ID should not change between the logins. if it does then we have major problems

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/c61947f57302634e1c83e0b33091ec9af9003edc..cc4431e3e06ae7b8ba11c9ffc83d458a9f04d385)*

src/controllers/session.rs
```diff
@@ -136,26 +138,69 @@ pub async fn authorize_session(
...
+            // a one-to-one relationship; this will need to be changed if/when we allow
+            // crates.io users to link more than one GitHub account to their crates.io account.
+            let user_id = create_or_update_user(
+                gh_user,
+                encrypted_token,
+                gh_user.email.as_deref(),
+                emails,
+                conn,
+            )
+            .await?;
```
Turbo87: I'm still a bit confused by the "update-or-upsert" pattern. if the update of an existing record failed because the record does not exist, then shouldn't we fall back to a regular insert instead of an upsert? I feel like this would probably simplify the code a bit, but maybe I'm missing something 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/c61947f57302634e1c83e0b33091ec9af9003edc..cc4431e3e06ae7b8ba11c9ffc83d458a9f04d385)*

src/controllers/session.rs
carols10cents: ahhhhhhhhh i misunderstood. i get it now.

src/controllers/session.rs
carols10cents: i think i was just trying to minimize the diff. fixing!

carols10cents: actually, I know why-- if there's a race between two requests logging in the same user, and the update of an existing record failed in one of the requests because the record doesn't exist, then the other request creates the user, then the first request goes to create the user, the second creation (from the first request) will fail. It seemed better to do an upsert so that both requests would succeed? I'm going to add some comments to that effect; if I'm incorrect, please let me know and I'll edit/remove the comments.

carols10cents:
@Turbo87 ready for re-review!

src/controllers/session.rs
```diff
@@ -158,73 +160,130 @@ pub async fn save_user_to_database(
...
-        if let Some(user_email) = user.email.as_deref() {
-            let new_email = NewEmail::builder()
-                .user_id(user_id)
-                .email(user_email)
+    // First, try to update an existing `oauth_github` record with the specified GitHub ID and the
+    // associated `users` record in a transaction so that either both `oauth_github` and `users`
+    // get updated or neither do.
+    //
+    // For now, update user display name, gh_login, and username. Eventually, we will
+    // get rid of `gh_login` and stop syncing `name` and `username` with GitHub.
```
LawnGnome: Total nit, but could we move this comment block into the closure, since it's describing things happening in the closure, please?

src/controllers/session.rs
```diff
@@ -158,73 +160,130 @@ pub async fn save_user_to_database(
...
+                .set((
+                    oauth_github::encrypted_token.eq(excluded(oauth_github::encrypted_token)),
+                    oauth_github::login.eq(excluded(oauth_github::login)),
+                    oauth_github::avatar.eq(excluded(oauth_github::avatar)),
+                    oauth_github::last_sync.eq(Utc::now()),
+                ))
+                .execute(conn)
+                .await?;
+
+            // To send the user an account verification email
```
LawnGnome: Just since this will be more than a screen away for many people from the initial comment explaining that this is the insert path:

```suggestion
            // Since this is a new user, send an account verification email
```

LawnGnome (approved):
I have a couple of comment nits, but otherwise this makes sense to me behaviourally. Approved, although I suspect @Turbo87 will want to take another look too.

I do _kinda_ wonder if we should split `create_or_update_user` to have each path in a separate function for readability reasons, but I don't feel very strongly about it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05..6523bacbfa97b424e4e6390393aff1f3a6539b32)*

carols10cents:
> I do _kinda_ wonder if we should split `create_or_update_user` to have each path in a separate function for readability reasons, but I don't feel very strongly about it.

Yeah, I think that will be done soon as part of https://github.com/rust-lang/crates.io/issues/13907.

src/controllers/session.rs
```diff
@@ -158,73 +160,130 @@ pub async fn save_user_to_database(
...
-                .email(user_email)
+    match conn
+        .transaction(async |conn| {
+            // First, try to update an existing `oauth_github` record with the specified GitHub ID
+            // and the associated `users` record in a transaction so that either both
+            // `oauth_github` and `users` get updated or neither do.
+            //
+            // For now, update user display name, gh_login, and username. Eventually, we will
+            // get rid of `gh_login` and stop syncing `name` and `username` with GitHub.
+            let oauth_github = diesel::update(oauth_github::table)
```
Turbo87: > I do kinda wonder if we should split create_or_update_user to have each path in a separate function for readability reasons, but I don't feel very strongly about it.

I agree with Adam. I find the `create_or_update_user()` fn fairly hard to read. what do you think about extracting `create_user()` and `update_user()` fns and calling them within `create_or_update_user()`? I don't think we need to wait on #13907 for that.

something like this:

```rust
    let update_result = conn
        .transaction(async |conn| update_user(gh_user, encrypted_token, conn).await)
        .await;

    match update_result {
        Ok(user_id) => Ok(user_id),

        // If the update fails because the `oauth_github` record doesn't exist, this
        // currently means the `user` record doesn't exist either and we need to create
        // both. This assumption holds because crates.io and GitHub accounts currently have
        // a one-to-one relationship; this will need to be changed if/when we allow
        // crates.io users to link more than one GitHub account to their crates.io account.
        Err(diesel::result::Error::NotFound) => {
            create_user(gh_user, encrypted_token, emails, conn).await
        }

        Err(error) => Err(error),
    }
```

probably plus a transaction around the `create_user()` call (see below)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05..6c8e2f3fccd2b6273bf24a186ea46ffbd0d03e83)*

src/controllers/session.rs
```diff
@@ -158,73 +160,130 @@ pub async fn save_user_to_database(
...
+                .on_conflict(oauth_github::account_id)
+                .do_update()
+                .set((
+                    oauth_github::encrypted_token.eq(excluded(oauth_github::encrypted_token)),
+                    oauth_github::login.eq(excluded(oauth_github::login)),
+                    oauth_github::avatar.eq(excluded(oauth_github::avatar)),
+                    oauth_github::last_sync.eq(Utc::now()),
+                ))
+                .execute(conn)
+                .await?;
```
Turbo87: it looks like the "create" code path is no longer run in a transaction. this means if the `oauth_github` insert fails, then the `users` row is left orphaned. we should probably wrap the create path in a transaction too, right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05..6c8e2f3fccd2b6273bf24a186ea46ffbd0d03e83)*

src/controllers/session.rs
```diff
@@ -158,73 +160,130 @@ pub async fn save_user_to_database(
...
+            // crates.io users to link more than one GitHub account to their crates.io account.
+            let new_user = NewUser::builder()
+                .gh_id(gh_user.id)
+                .gh_login(&gh_user.login)
+                .username(&gh_user.login)
+                .maybe_name(gh_user.name.as_deref())
+                .gh_encrypted_token(encrypted_token)
                 .build();
 
-            if let Some(token) = new_email.insert_if_missing(conn).await? {
```
Turbo87, line 198: previously we were running this also on the "update" code path. I'm not sure if it's a big deal to drop that for now, but I wonder if that was an intentional choice?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05..6c8e2f3fccd2b6273bf24a186ea46ffbd0d03e83)*

src/controllers/session.rs
```diff
@@ -158,73 +160,130 @@ pub async fn save_user_to_database(
...
-                    }
-                };
+            // This currently inserts or updates based on `users::gh_id` being unique. When we switch
+            // to using `users::username` for uniqueness instead, this logic will need to change.
+            // See note on `NewUser::insert_or_update`.
+            //
+            // Do an upsert rather than an insert in case there are two requests racing to log
+            // in the same user in case a record for this user has been created between when the
+            // "update" transaction in this `match` expression ran and now.
+            let user_id = new_user.insert_or_update(conn).await?;
```
Turbo87: I'm not quite convince yet that this should be an upsert (vs. a regular insert). I think I'd rather see the second concurrent request fail due to the conflict instead of silently updating stuff in the "create" code path. In reality this is probably a rare occurrence anyway.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05..6c8e2f3fccd2b6273bf24a186ea46ffbd0d03e83)*

src/controllers/session.rs
```diff
@@ -158,73 +160,130 @@ pub async fn save_user_to_database(
...
+                    oauth_github::last_sync.eq(Utc::now()),
+                ))
+                .on_conflict(oauth_github::account_id)
+                .do_update()
+                .set((
+                    oauth_github::encrypted_token.eq(excluded(oauth_github::encrypted_token)),
+                    oauth_github::login.eq(excluded(oauth_github::login)),
+                    oauth_github::avatar.eq(excluded(oauth_github::avatar)),
+                    oauth_github::last_sync.eq(Utc::now()),
+                ))
```
Turbo87:
```suggestion
```

The same for this operation. I'm not convinced that we should use an upsert instead of insert here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05..6c8e2f3fccd2b6273bf24a186ea46ffbd0d03e83)*

crates/crates_io_database/src/models/user.rs
```diff
@@ -160,55 +165,3 @@ pub struct OauthGithub {
...
-    primary_key(account_id),
-    belongs_to(User),
-)]
-pub struct NewOauthGithub<'a> {
-    pub account_id: i64,           // corresponds to users.gh_id
-    pub avatar: Option<&'a str>,   // corresponds to users.gh_avatar
-    pub encrypted_token: &'a [u8], // corresponds to users.gh_encrypted_token
-    pub login: &'a str,            // corresponds to users.gh_login
-    pub user_id: i32,
-}
```
Turbo87, line 179: what was the reason for removing this?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05..6c8e2f3fccd2b6273bf24a186ea46ffbd0d03e83)*

carols10cents: I liked having the insert written out in diesel statements to parallel the upsert when they were in the same function 🤷🏻‍♀️ I think I reached for this in a different branch I'm working on, though, and the insert+upsert are in different functions now, so I'll put it back!

src/controllers/session.rs
carols10cents: Ok! But I'm going to put the transactions inside the `update_user` and `create_user` functions though, so that it's not possible to accidentally call them from outside a transaction.

src/controllers/session.rs
carols10cents: Good catch-- you're right, it was unintentional. But I think the direction we're moving with https://github.com/rust-lang/crates.io/issues/13907 is to sync less info with github, and rely more on explicit setting/confirmation in the new user case, and explicit editing of a canonical crates.io value in the update case? So now I am intentionally not putting it back :)

src/controllers/session.rs
carols10cents: Ok, i mean I don't think it hurts anything to return an error, the user will just hit refresh and then there shouldn't be a conflict 🤷🏻‍♀️

src/controllers/session.rs
carols10cents: Yup, you're right.

src/controllers/session.rs
Turbo87: yeah, I think that's fine. Just wanted it to be an intentional choice :)

carols10cents:
@Turbo87 @LawnGnome Ready for re-review! ❤️

src/controllers/session.rs
```diff
@@ -137,59 +138,120 @@ pub async fn authorize_session(
...
 
-/// Inserts the user into the database, or updates an existing one.
+/// Updates an existing user or inserts a new user into the database.
+async fn create_or_update_user(
+    gh_user: &GitHubUser,
+    encrypted_token: &[u8],
+    emails: &Emails,
+    conn: &mut AsyncPgConnection,
+) -> QueryResult<i32> {
+    let update_result = update_user(gh_user, encrypted_token, conn).await;
```
LawnGnome: It does feel _slightly_ weird to me that we're not doing this whole operation (update-and-maybe-create) in a transaction, although [I guess this is OK per this discussion](https://github.com/rust-lang/crates.io/pull/13856#discussion_r3549971057). I can't think of an actual problem, but I do think there are edge cases where this might feel a bit odd to a new user who is somehow doing two logins at the same time. (Although doing so is also odd.)

LawnGnome (approved):
These changes address my concerns from the first review, and I think the code generally reads better now, so thanks for that!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/37a6254482830358f72eb4caaabe0318f8fd185a..a5bbce64ac3be51747b58374f7f356e76f6709ca)*

src/controllers/session.rs
Turbo87: the problem is that if the update fails then the whole transaction is poisoned (or whatever the term is) and won't accept any further requests, so we can't run the create op in the same transaction as a fallback afterwards.

LawnGnome: Right, of course. Thanks!

Turbo87: thinking about this some more, I'm not sure if I believe my own statement anymore. I think this is true for insert-or-update order, but with update-or-insert order we could probably detect if we updated 1 or 0 rows and then go into the insert path for the latter 🤔

Turbo87: right, yeah, I was remembering the `is_read_only_error()` branch point in this flow. if the upsert failed then the read-only fallback to `find_user_by_gh_id()` wouldn't work anymore if it runs in the same transaction. for update-or-insert I think it should be possible within a single transaction.

src/tests/util/test_app.rs
```diff
@@ -165,8 +169,6 @@ impl TestApp {
 
         new_email.insert(&conn).await.unwrap();
 
-        let user = User::find(&conn, id).await.unwrap();
-
```
Turbo87, line 169: this change seems to be causing a lot of snapshot changes due to the hardcoded avatar in `OauthGithubBuilder::for_user()`, which makes the diff seem much larger than necessary. I wonder if we should either default to `None` for now in `OauthGithubBuilder::for_user()` or use `user.gh_avatar`. It seems that the only place that needs that hardcoded avatar is the new `fetch_crate_info()` test, which all the others are expecting `None`/`null` in the snapshots.

I realize that I probably should've caught this in #14175 instead. Sorry :-/

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/37a6254482830358f72eb4caaabe0318f8fd185a..a5bbce64ac3be51747b58374f7f356e76f6709ca)*

carols10cents: Yes, the diff is larger than necessary, but I do think this is more realistic than what we had before.

Using `user.gh_avatar` isn't a good idea because we're about to get rid of that 😂 I also think it's more likely that we get an avatar from GitHub, so setting an avatar by default in tests is more like production usage. 

I suppose we don't know how often we'll get avatars from future OAuth providers, and we haven't failed to catch any problems in the past because the test data was missing avatars on users, that I remember. So I will return this to its previous behavior, but I did just want to document that I don't think I'm _wrong_, strictly speaking 😉

src/controllers/session.rs
carols10cents: Ok. I moved the transaction up one function, added comments explaining why i didn't move it up two functions and which functions should be called in a transaction, and tests are passing!

carols10cents:
@Turbo87 @LawnGnome Ready for re-review! ❤️

src/tests/util/test_app.rs
Turbo87: no, I don't think you're wrong either, but this should IMHO be a dedicated refactoring and not mixed up with this behavior change PR :)

Turbo87 (approved):
thank you! :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13856/37a6254482830358f72eb4caaabe0318f8fd185a..9dceaa147c286871b66510d1c749ea0e54cffd3e)*

PR #13864, opened 2026-06-05
Use shallow clones for the worker index repository

Turbo87:
The worker's long-lived index clone only ever reads the current tip and commits/pushes on top of it, so it does not need the full index history.

This adds an opt-in `Repository::open_shallow()` fn that clones with `--depth 1` and keeps `reset_head()` fetches shallow. It is used from the worker's `lock_index()` fn. The full-history `Repository::open()` is unchanged so the `upload-index` admin tool's incremental mode keeps working.

### DIFF
--- a/crates/crates_io_index/src/repo.rs
+++ b/crates/crates_io_index/src/repo.rs
@@ -65,11 +65,21 @@ pub struct Repository {
     checkout_path: TempDir,
     repository: git2::Repository,
     credentials: Credentials,
+    /// Whether the local clone only contains the tip commit (`--depth 1`).
+    ///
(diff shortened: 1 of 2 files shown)


Turbo87:
I've tried this out on the staging server:

Bildschirmfoto 2026-06-05 um 14 29 28

→ https://github.com/rust-lang/staging.crates.io-index/commit/e1e8cf4a2e50e4eea24a0c74d649797004bd6b6e

but with 14 commits (now 15) on the staging index, that isn't quite representative in terms of performance. The commit creation and pushing works though, and we could still revert if this turns out to be a performance regression in production.

Issue #13771, opened 2026-05-26
Make `users.name` display name editable (rather than sync through GitHub)


carols10cents:
We currently have the concept of a "display name" that is associated with and managed through your GitHub "display name". The display name is currently used on crate pages for owners and user pages.

Along with the crates.io username changes to make crates.io more independent from GitHub, this issue is to change the display name field to get its initial value from GitHub when creating a new account (if the GitHub oauth data contains a display name), but otherwise allow crates.io users to set the value in crates.io (and not continue syncing the value with GitHub). Another way of saying this is that `users.name` should function more similarly to email addresses in the way it gets an initial value from GitHub if available, but doesn't get overwritten by changes from GitHub in the future. `users.name` won't be completely the same as emails because it's a column in the users table rather than its own table, and there's no verification needed like we do with email addresses.

So the pieces of solving this issue are:

- Make display name editable in crates.io rather than only by syncing from GitHub
  - Add a field to the `/settings/profile` UI 
  - Adjust the text on the profile settings page that starts with "To update your name and GitHub account" to be accurate
  - Change the [user update PUT API](https://github.com/rust-lang/crates.io/blob/0f3189cc9df240ef372981d5b4d3931da61dfbe9/src/controllers/user/update.rs) to save the user's display name in the database
- Audit everywhere in the UI that uses either display name or username; we're definitely inconsistent in where we use it.
  - Possibly use the display name in _more_ places? 
  - Or use display name + username in places that we're currently using only display name? 

## Rejected idea of getting rid of display name completely

Originally this issue proposed only showing crates.io usernames everywhere.

While the display name is a vector for possible impersonation (one that's technically possible today. It is against crates.io policies, and as far as I can remember we haven't had an account impersonation reported), people do like being able to choose the name that goes along with their username.

Turbo87:
tbh I think I'd prefer keeping the concept of a display name around on our side too. I generally consider this a useful thing and it seems that e.g. npm and PyPI also support this (aside from obviously GitHub itself).

carols10cents:
EDIT: I've updated this issue's description to be this TODO list now.

Ok, then in that case we need to:

- Make it editable in crates.io rather than the current way of editing which is only syncing from GitHub
- Audit everywhere in the UI that uses either display name or username; we're definitely inconsistent in where we use it.
  - Possibly use the display name in _more_ places? 
  - Or use display name + username in places?

jasonaowen:
[...]

Issue #13765, opened 2026-05-26
Add a table for reserved usernames similar to reserved crate names


carols10cents:
[In the same manner as `reserved_crate_names`](https://github.com/rust-lang/crates.io/blob/07920c13aa8d428105fa2ea6fc1cf74e803775b8/migrations/20170305095748_create_reserved_crate_names/up.sql), write a database migration that:

- creates a table named `reserved_usernames`
- creates a function named `ensure_username_not_reserved`
- uses the `ensure_username_not_reserved` function in a trigger on user record insert or update
- does NOT add initial values to the table; crates.io admins will take care of that

This table can be [publicly available in the database dump like `reserved_crate_names` is](https://github.com/rust-lang/crates.io/blob/07920c13aa8d428105fa2ea6fc1cf74e803775b8/crates/crates_io_database_dump/src/dump-db.toml#L196-L197).

Once this is deployed, crates.io admins should manually populate this table using the `reserved_crate_names` table as inspiration.

MrSheerluck:
I'll pick this

Turbo87:
reopening this because it will need a part two eventually. I think right now the database trigger would cause the corresponding endpoints to return "500 Internal Server Error", but they should probably return a 4xx error instead. We will also need tests for this. But since right now the users can't choose their own usernames yet anyway, this can/should probably be deferred until then.

carols10cents:
> reopening this because it will need a part two eventually. I think right now the database trigger would cause the corresponding endpoints to return "500 Internal Server Error", but they should probably return a 4xx error instead. We will also need tests for this. But since right now the users can't choose their own usernames yet anyway, this can/should probably be deferred until then.

oh cool cool looks like we don't have any tests for reserved _crate_ names (that I can find). And it looks like [we proactively check `reserved_crate_names` to return the nice error](https://github.com/rust-lang/crates.io/blob/main/src/controllers/krate/publish.rs#L464-L466), with the database trigger as extra insurance.

carols10cents:
Also before this can be closed, us admins need to seed the table with usernames we don't want to allow. I've added an item to our agenda to discuss this.

Also also, doing https://github.com/rust-lang/crates.io/issues/13907 before actually enforcing this will make it easier to return an error and put the user back at a page that lets them pick a different username.

carols10cents:
Ok, I have added the usernames we'd like to ban to the staging and prod databases, so I'm closing this issue. We can continue to tweak the values in that table.

Issue #13762, opened 2026-05-26
Add unique index to `users.username`


carols10cents:
We have [a postgres function `canon_crate_name`](https://github.com/rust-lang/crates.io/blob/07920c13aa8d428105fa2ea6fc1cf74e803775b8/migrations/20150319224700_dumped_migration_93/up.sql) that replaces hyphens with underscores and lowercases strings. This function is used for [the unique crate name index](https://github.com/rust-lang/crates.io/blob/main/migrations/20150319224701_dumped_migration_94/up.sql) and should be used for the unique username index as well.

I'm open to arguments about whether we should rename `canon_crate_name` to be something more general such as `canon_name` or `normalized_identifier` or similar, or whether we should keep the function name `canon_crate_name` even as we use it for usernames too.

The PR for this issue should include a migration that adds a unique username index as detailed above.

The admin that deploys this should verify that `users.username` is actually unique before deploying.

Turbo87:
~resolved by https://github.com/rust-lang/crates.io/commit/0d2adff3fd9d4044790f92fb8b5c769d8c5e1055~

no, it was not. the commit introduced an index, but not a *unique* one yet. that will have to wait until some of the other requirements are satisfied first.

PR #13872, opened 2026-06-06
Make `users.username` column `not null`

### LINKED ISSUE #13760: Create a new column `users.username` set to the value of `users.gh_login`

MrSheerluck:
Fixes: #13760 

Follow up PR of my last PR. Made username not null

### DIFF
--- a/migrations/2026-06-06-120000-0000_make_username_not_null/up.sql
+++ b/migrations/2026-06-06-120000-0000_make_username_not_null/up.sql
@@ -4,6 +4,4 @@
 -- performs under ACCESS EXCLUSIVE is expected to be fast.
 alter table users
     alter column username set not null;
-
-comment on column users.username is 'Username associated with the user''s crates.io account, independent of linked OAuth usernames.';
 -- safety-assured:end
--- a/migrations/2026-06-06-120000-0000_make_username_not_null/up.sql
+++ b/migrations/2026-06-06-120000-0000_make_username_not_null/up.sql
@@ -1,4 +1,13 @@
-alter table users
-    alter column username set not null;
+
+alter table users add constraint username_not_null check (username is not null) not valid;
+
+
+alter table users validate constraint username_not_null;
+
+
+alter table users alter column username set not null;
+
+
+alter table users drop constraint username_not_null;
 
 comment on column users.username is 'Username associated with the user''s crates.io account, independent of linked OAuth usernames. This column is now non-nullable following a data backfill from gh_login.';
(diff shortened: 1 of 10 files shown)


MrSheerluck:
@carols10cents can you please review this

migrations/2026-06-06-120000-0000_make_username_not_null/up.sql
```diff
@@ -0,0 +1,9 @@
+-- safety-assured:start
+-- All rows in the `users` table already have a non-null `username` value after
+-- the admin backfill from `gh_login`. The validation scan that `SET NOT NULL`
+-- performs under ACCESS EXCLUSIVE is expected to be fast.
+alter table users
+    alter column username set not null;
+
+comment on column users.username is 'Username associated with the user''s crates.io account, independent of linked OAuth usernames.';
```
Turbo87:
```suggestion
```

unless I'm missing something, this looks like the same comment as was previously set on the column, no?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13872/fd9432b295963e97f6e36f7d0a675af96c256b5e..8e18656c952e5b33b77fca43836feee0fb7b3807)*

MrSheerluck: I was actually in doubt whether I should add it or not. Let me know if I should keep it or change it to something else or remove it entirely

Turbo87: since the comment already exists in the database there is no point in assigning it again, so I think we should just remove it

MrSheerluck: sure, should I add something else instead of that. Otherwise I'll update the PR

carols10cents:
This looks great to me! We need to wait to merge it though, we haven't deployed the first set of changes yet.

Some ideas on what you could work on next, if you'd like:

- https://github.com/rust-lang/crates.io/issues/13762
- https://github.com/rust-lang/crates.io/issues/13765
- https://github.com/rust-lang/crates.io/issues/13771

MrSheerluck:
sure I can pick these one by one

Turbo87 (approved):

Turbo87:
> We need to wait to merge it though, we haven't deployed the first set of changes yet.

the `username` column has been backfilled by Carol yesterday, so I'd say this is good to go now :)

PR #13893, opened 2026-06-09
Add `reverse_dependencies` table and admin command

Turbo87:
This is the first of two PRs that move the reverse-dependencies endpoint off the per-request recomputation and onto a precomputed table. This one adds the table and the tooling to populate it without touching the endpoint itself, so it is a no-op for users.

The new `reverse_dependencies` table holds one row per `(target crate, dependent crate)` edge where the dependent crate's non-yanked default version depends on the target. With it in place the endpoint will be able to answer a page with a bounded index range scan instead of recomputing the full dependent set on every request. A set of triggers keeps the table in sync as `default_versions`, `versions`, `dependencies`, and `crate_downloads` change, and crate deletions ride on `ON DELETE CASCADE`.

Because those triggers only maintain the table on subsequent writes, the table starts out empty and needs an initial backfill. The new `crates-admin reverse-dependencies` command handles that: `update` rebuilds the edges and is safe to re-run, so it doubles as a repair tool, while `verify` recomputes them and compares against the stored rows.

That ordering is why this lands on its own. This PR has to be merged and deployed, and the table backfilled with `crates-admin reverse-dependencies update --all`, before the follow-up PR can switch the endpoint over to read from the table.


migrations/2026-06-05-120000-0000_create_reverse_dependencies/up.sql
```diff
@@ -0,0 +1,205 @@
...
+--
+-- The `integer` FKs match the referenced `crates.id`/`versions.id` `integer`
+-- primary keys, so the "widen to bigint" suggestion does not apply. The indexes
+-- are built on a brand-new, empty table inside this transaction, so the `SHARE`
+-- lock is instant and `CONCURRENTLY` is unnecessary (and cannot run here anyway).
+
+CREATE TABLE IF NOT EXISTS reverse_dependencies (
+    target_crate_id     integer NOT NULL REFERENCES crates (id) ON DELETE CASCADE,
+    dependent_crate_id  integer NOT NULL REFERENCES crates (id) ON DELETE CASCADE,
+    version_id          integer NOT NULL REFERENCES versions (id) ON DELETE CASCADE,
```
LawnGnome, line 23: I'm assuming having both the crate and version ID is for performance reasons?

crates/crates_io_database_dump/src/dump-db.toml
```diff
@@ -197,6 +197,15 @@ rendered_at = "private"
 [reserved_crate_names.columns]
 name = "public"
 
+# Derived cache table maintained from `default_versions`/`dependencies`; not
+# part of the public database dump.
```
LawnGnome, line 201: Should we include this in the dump? Sure, it's computable, but I don't think there's any harm in having it besides size, and I don't think the marginal difference would be problematic.

src/bin/crates-admin/reverse_dependencies.rs
```diff
@@ -0,0 +1,206 @@
...
+        .filter(crates::name.eq_any(&scope.crates))
+        .select((crates::id, crates::name))
+        .load(conn)
+        .await
+        .context("Failed to look up crates")?;
+
+    let found_names: HashSet<&str> = found.iter().map(|(_, name)| name.as_str()).collect();
+    for name in &scope.crates {
+        if !found_names.contains(name.as_str()) {
+            warn!("Crate `{name}` not found; skipping");
```
LawnGnome, line 134: I debated if this should error, but I think this is appropriate for something that is ultimately kinda transient.

LawnGnome (approved):
LGTM; I'm keen to see how this performs in prod.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13893/32ef3831723fec2f3a11fa2a3987b340108dff84..ff65a83b9ab06fce8fccb1f77d2ab41f9c9756ac)*

migrations/2026-06-05-120000-0000_create_reverse_dependencies/up.sql
Turbo87: yeah, one less redirection to resolve

crates/crates_io_database_dump/src/dump-db.toml
Turbo87: I'm a bit afraid of people then starting to use the derived table while we are still fiddling around with it. Once it's in the database dump we try to reduce the breaking changes at least a little bit.

PR #13901, opened 2026-06-10
jobs/generate_og_image: Read owner avatars from `oauth_github`

Turbo87:
`fetch_user_owners()` now sources avatars from `oauth_github.avatar` via a left-join instead of the `users.gh_avatar` column, matching how the `User` struct already reads the field (see #13800). This removes the one remaining read path of `users.gh_avatar`.

/cc @carols10cents 

### Related 

- #13800


carols10cents (approved):
🎉 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13901/5afd4943406627346173386f8a2ebe2ce14a1ab7..f1561aace9a6a410e018127c0fe26611e37f6e40)*

PR #13902, opened 2026-06-10
controllers/krate: Name a reverse dependency in delete error

### LINKED ISSUE #12881: Only crates without reverse dependencies can be deleted, except this one has none
### Current Behavior

Right now, I'm trying to delete `value`, an extremely old crate of mine which has seen minimal use and doesn't seem very useful. It seems unreasonable to effectively squat such a common word, but I can't delete it because it claims that it has reverse dependencies, even though it has none.

I think it may have had reverse-dependencies with another crate that I also deleted, but the database is inconsistent and thinks otherwise. 

### Expected Behavior

I should be able to delete it.

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

Turbo87:
When a crate deletion is blocked by reverse dependencies, the error now names an example dependent version (e.g. `bar@1.0.0`) so the caller can see what is blocking the deletion.

This replaces the `has_rev_dep()` boolean check with a `RevDep` struct queried via `for_crate()`, which fetches the dependent crate's name and version.

### Related

- Closes https://github.com/rust-lang/crates.io/issues/12881

### DIFF
--- a/src/controllers/krate/delete.rs
+++ b/src/controllers/krate/delete.rs
@@ -104,8 +104,11 @@ pub async fn delete_crate(
 
     // All crates with reverse dependencies are blocked from being deleted to avoid unexpected
     // historical index changes.
-    if has_rev_dep(&conn, krate.id).await? {
-        let msg = "only crates without reverse dependencies can be deleted";
+    if let Some(rev_dep) = RevDep::for_crate(krate.id, &conn).await? {
+        let msg = format!(
+            "only crates without reverse dependencies can be deleted (e.g. {}@{} depends on this crate)",
+            rev_dep.name, rev_dep.version
+        );
         return Err(custom(StatusCode::UNPROCESSABLE_ENTITY, msg));
     }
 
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):
This looks fine, so :+1:, but I'm curious: is there a reason to not display all the rdeps? It feels like we might encourage people into a game of whack-a-mole in some cases.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13902/5afd4943406627346173386f8a2ebe2ce14a1ab7..3746a1923a85139b19624adff50b515954576498)*

src/controllers/krate/delete.rs
```diff
@@ -185,19 +188,30 @@ pub fn max_downloads(age: &TimeDelta) -> u64 {
...
+    /// Queries the database for the first matching reverse dependency of a crate, if any.
+    async fn for_crate(crate_id: i32, mut conn: &AsyncPgConnection) -> QueryResult<Option<Self>> {
+        Self::query()
+            .filter(dependencies::crate_id.eq(crate_id))
+            // Ignore self-referencing dependencies, i.e. dependencies where the
+            // depending version belongs to the same crate that is being depended
+            // upon. Some crates depend on themselves (e.g. for backwards
+            // compatibility), and these should not block deletion.
+            .filter(versions::crate_id.ne(crate_id))
+            .first(&mut conn)
```
carols10cents, line 215: Should this have an `order_by` so that it doesn't change if you, say, reload the page (or have tests where there's more than one rev dep that could be flaky?) Since ordering without an `ORDER BY` is undefined in postgres.

I don't think it really matters _what_ the order is, so like `versions::id` would be fine I think. Just that there is some defined ordering.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13902/5afd4943406627346173386f8a2ebe2ce14a1ab7..3746a1923a85139b19624adff50b515954576498)*

src/tests/routes/crates/delete.rs
carols10cents, line 287: To test the ordering i've suggested above, should this publish another crate and/or another version so that the test can assert that the same one is returned every time?

And yanked versions should count too, should this test also have a yanked version? Or have another test that only has a yanked version depending on the crate being deleted?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13902/5afd4943406627346173386f8a2ebe2ce14a1ab7..3746a1923a85139b19624adff50b515954576498)*

src/controllers/krate/delete.rs
Turbo87: > I don't think it really matters _what_ the order is, so like `versions::id` would be fine I think. Just that there is some defined ordering.

yeah, I was gonna say that the ordering could be problematic in terms of performance, but if we order by a primary key just for result stability reasons then this might work 👍

Turbo87:
> is there a reason to not display all the rdeps? It feels like we might encourage people into a game of whack-a-mole in some cases.

your crate could easily have thousands of dependency edges (N crates x M versions). listing them all in the error message would feel a bit much 😅

PR #13903, opened 2026-06-10
Add `reserved_usernames` database table

### LINKED ISSUE #13765: Add a table for reserved usernames similar to reserved crate names

MrSheerluck:
closes #13765 

Added this new table and its functions and triggers as suggested in the issue

Screenshot 2026-06-10 at 19 02 47


migrations/2026-06-10-120000-0000_create_reserved_usernames/up.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE reserved_usernames (
+    username TEXT PRIMARY KEY
+);
+CREATE UNIQUE INDEX ON reserved_usernames (lower(username));
```
Turbo87: https://github.com/rust-lang/crates.io/issues/13762 suggests using `canon_crate_name()` instead of e.g. `lower()`. I'm personally not convinced that we should reuse `canon_crate_name()` for this, but something like `canon_username()` which turns the username to lower case and replaces `-` with `_` (or vice versa) would probably be a good idea. this should probably be implemented before the introduction of this index, possibly in a dedicated PR with corresponding tests.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13903/5afd4943406627346173386f8a2ebe2ce14a1ab7..0a66d11631d7a2ed3b56591bea8c6f1eb4ae2117)*

MrSheerluck: ok so should i change this to your suggestion or keep it as it is?

migrations/2026-06-10-120000-0000_create_reserved_usernames/up.sql
```diff
@@ -0,0 +1,28 @@
+CREATE FUNCTION canon_username(text) RETURNS text AS $$
+    SELECT replace(lower($1), '-', '_');
+$$ LANGUAGE SQL IMMUTABLE;
+
+-- safety-assured:start
```
Turbo87: if you add `safety-assured` comments, please explain in the comment what the issue was that this is working around and why you think that disabling the checks is safe

Turbo87 (commented):
this looks good to me in general. one minor nit about the safety comments, but once that is fixed I think this is good to go :)

migrations/2026-06-10-120000-0000_create_reserved_usernames/up.sql
MrSheerluck: ok I've added comments explaining the reasons. Please check and let me know if you want any alternate solution or you want any modifications to the comments

migrations/2026-06-10-120000-0000_create_reserved_usernames/up.sql
```diff
@@ -0,0 +1,36 @@
...
+-- Suppresses the "CREATE TABLE without IF NOT EXISTS" diesel-guard check.
+-- The table is brand new and `IF NOT EXISTS` is already used, so retrying a
+-- partially-failed migration is safe.
```
Turbo87: I don't understand this. the code below seems to use `IF NOT EXISTS` already, so why is diesel-check complaining about it?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13903/5afd4943406627346173386f8a2ebe2ce14a1ab7..7856cdc3d3af91eeaf72dba1a36ffe39ed6931e2)*

MrSheerluck: I'm sorry but I have no clue
 I just tried to fix this when I saw that I'm getting migration ci failure. 

Can you help me if there is another solution

Turbo87: I've fixed it. The `CREATE TABLE without IF NOT EXISTS` check was already fixed by adding `IF NOT EXISTS` to the statement, so the `safety-assured:start` instructions were essentially not having any effect. 

In the `down.sql` I'm not sure what `DROP TABLE` check was triggering for you. I ran `diesel-guard` locally and even without the `safety-assured:start` instruction the check passed fine.

PR #13924, opened 2026-06-12
test_db: strip pg_dump 17+ output that psql 16 can't handle

LawnGnome:
I guess this wouldn't be an issue if the dev container thing worked for me, but here we are.

### DIFF
--- a/crates/crates_io_test_db/src/lib.rs
+++ b/crates/crates_io_test_db/src/lib.rs
@@ -279,6 +279,8 @@ fn run_migrations(conn: &mut PgConnection) -> diesel::migration::Result<()> {
 ///   `batch_execute` can't run.
 /// - Lines starting with `\` (psql meta-commands like `\restrict`/
 ///   `\unrestrict`, emitted by `pg_dump` 17+) are stripped.
+/// - SETs for configuration that pg_dump 17+ emit that PostgreSQL 16 can't
+///   handle.
 /// - A trailing `RESET ALL` clears any session state the preamble's
 ///   `SET` / `set_config('search_path', …)` calls left behind so it
 ///   doesn't leak to the next checkout of the pooled connection.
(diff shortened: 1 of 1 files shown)


Turbo87:
does https://github.com/rust-lang/crates.io/blob/d6b22e88d6fb115cc8fe565894f961d9c55be160/.env.sample#L25-L29 not work for you?

> if the dev container thing worked for me

what about it isn't working for you?

LawnGnome:
> does
> 
> https://github.com/rust-lang/crates.io/blob/d6b22e88d6fb115cc8fe565894f961d9c55be160/.env.sample#L25-L29
> not work for you?

I missed that change going in.

That said, I also don't have easy access to PostgreSQL 16 binaries on Arch without installing from the AUR, and that's currently undergoing a massive malware attack, so I'm probably not going to do that right now. :slightly_smiling_face: 

What I'll probably end up doing if this PR isn't merged is to figure out some way of shimming the PostgreSQL commands into the Docker container that is running version 16.

> what about it isn't working for you?

OK, so this is drifting off topic, but you asked.

I had some breakage the first time I tried this, but that was in May, and I don't remember the details.

Before I made this change yesterday, I tried to use the dev container configuration to see if that was easier than yet another hour or two figuring out why the integration tests broke this time. My experience was:

1. The first time I tried it, I got an inscrutable error about port 5173 having already been bound. (I promise I wasn't running anything else listening on that port.)
2. The second time around, it got as far as starting, but host names wouldn't resolve inside the containers. (I promise every other Docker container works normally on my laptop.)
3. While debugging that, I realised that the terminal I was getting was within the container and had basically no useful tooling installed. I could probably figure out how to fix that, but I'd kind of hit my limit at that point.

Maybe this is Docker's fault. Maybe it's Zed's. Maybe it's mine. (OK, it's probably mine.) But, truthfully, I already have trouble keeping a crates.io environment that can run most of the integration tests running reliably, and adding a new set of variables wasn't helping, so I did this instead, since we already have prior art in the codebase to handle the major version mismatch.

I'm sure there's a happy path there for users on macOS with VS Code who don't rely much on the terminal when developing, but I'm not that user, unfortunately.

Turbo87:
alright, I guess this is small enough to not hurt too much

Issue #13867, opened 2026-06-05
Add `users.created_at`


LawnGnome:
We should track when a crates.io user is created.

## Steps

* [x] #13925 
* [x] ~~Use a batched background worker job to backfill values for existing users~~ Actually, I think I'm going to do this as a simple update script, given this is a one time problem.
* [ ] #14035

PR #13925, opened 2026-06-12
Add `users.created_at` to the database

LawnGnome:
This doesn't expose the field in any way other than the database dump; just adds it and sets setting it for new users.

Part one of (probably) three of #13867.


migrations/2026-06-11-213312-0000_users-created-at/up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE users
+ADD COLUMN IF NOT EXISTS created_at TIMESTAMP WITH TIME ZONE NULL DEFAULT NULL,
+ALTER COLUMN created_at
+SET DEFAULT NOW();
```
Turbo87, line 4: I was a little confused by this double-default pattern at first, but ended up in a TIL situation 😆 

for those that don't know: the `DEFAULT NULL` specifies the value that is going to be used for existing rows in the table (and new ones too), the `DEFAULT NOW()` then changes the value that is going to be used for any new rows, but keeps the `NULL` in the existing rows in place.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13925/d4063dbb474f9a2a7a602f414d2b33086befec5c..7926c5002c53485229337b45b2bcca31a3cc27e5)*

Turbo87 (approved):

migrations/2026-06-11-213312-0000_users-created-at/up.sql
LawnGnome: Sorry, I should have written a comment!

Turbo87: no worries, the intent was obvious, I just didn't know that it was possible like this 😄

Issue #5312, opened 2022-10-10
Most relevant results are listed last


jpryne:
### Current Behavior

When I search for "bevy_rapier" sorted by Relevance, I expect to see "bevy_rapier2d" & "bevy_rapier3d" at the top of the list. Instead, they are at the bottom of the results. The top result is "bevy-inspector-egui-rapier", which, while it is related, is not what I'm looking for.

### Expected Behavior

I expect to see "bevy_rapier2d" & "bevy_rapier3d" at the top of the list.

### Steps To Reproduce

1. Go to crates.io
2. Be sure "Most Relevant" is selected in the Search Results pulldown menu.
3. Search for "bevy_rapier"
4. Find "bevy_rapier2d" & "bevy_rapier3d" at the end of list list.

### Environment

- Browser: Chrome, latest
- OS: Windows 10 64

Turbo87:
thank you for the report. there are quite a few other reports about the suboptimal search results though, so I will close this one as a duplicate for now :)

Issue #2541, opened 2020-05-31
https://crates.io/search?q=truetype results are poorly ordered


jrmuizel:
The first result after the exact match is "stb_truetype_bugfix_19072017". This crate was last updated three years ago and has had only 432 downloads. "rusttype_bugfix_19072017" also ranks quite highly.

Turbo87:
@jrmuizel do you have any concrete suggestions on how to fix this? otherwise this issue isn't really actionable unfortunately.

jrmuizel:
Why does stb_truetype_bugfix_19072017 get listed before stb_truetype? Is there a description or documentation for how the relevance metric is computed?

Turbo87:
https://github.com/rust-lang/crates.io/issues/1270 has a bit more information on this. since that issue describes a situation fairly similar to yours I'll close this issue as a duplicate for now.

PR #13948, opened 2026-06-14
Bound relevance-sorted crate search to the most-downloaded candidates

Turbo87:
Sorting crate search by relevance ranked every crate matching the query with `ts_rank_cd()`. For short or common terms (`time`, `test`, `data`, ...) that is a large fraction of all crates, each requiring its large, TOAST-ed `tsvector` to be read just to be ranked and then discarded below the result limit, taking up to ~1-2s.

This restricts relevance ranking to the 1000 matching crates with the most recent downloads and ranks only those. The candidate selection in `relevance_candidate_ids()` sorts by an integer column rather than the `tsvector`, so the expensive `ts_rank_cd()` now runs over at most 1000 rows instead of 50k+. An exact name match is always kept among the candidates regardless of its download count, and a crate's score still reflects all of its searchable text within the candidate set.

It also breaks `ts_rank_cd()` ties by recent downloads before name. Common terms produce many equally-ranked matches, so previously the top results were effectively ordered alphabetically. With the tiebreaker the more popular crates surface first, and crates without recent downloads sort last. The `Seek::Relevance` pagination key gains a recent-downloads level so seek-based paging stays consistent with the ordering.

The reachable range covers what the frontend displays and the deepest non-bot API paging. An explicit page that starts beyond the cap now returns a `400` instead of a silently empty page, and seek pagination stops at the boundary on its own. The endpoint doc comment records the cap so it surfaces in the OpenAPI description and the generated TypeScript client.

The one visible behavior change worth flagging: a highly relevant but rarely downloaded crate no longer appears in relevance results, since ranking is bounded to the most-downloaded candidates. This is intentional. It demotes obscure README-keyword-dense crates in favor of widely-used ones for common terms. Other sorts (`new`, `alphabetical`, ...) stay unbounded.

### Related

- https://github.com/rust-lang/crates.io/issues/5312
- https://github.com/rust-lang/crates.io/issues/2541
- https://github.com/rust-lang/crates.io/issues/1370

### DIFF
--- a/packages/crates-io-api-client/schema.d.ts
+++ b/packages/crates-io-api-client/schema.d.ts
@@ -168,6 +168,9 @@ export interface paths {
          *     - Alphabetical listing of crates
          *     - List of crates under a specific owner
          *     - Listing a user's followed crates
+         *
+         *     When sorting by relevance, only the first 1000 results can be accessed, even
+         *     though `meta.total` may report a higher number of matching crates.
          */
         get: operations["list_crates"];
         put?: never;
--- a/src/controllers/krate/search.rs
+++ b/src/controllers/krate/search.rs
@@ -32,6 +32,10 @@ use crate::util::no_store;
 use crate::util::string_excl_null::StringExclNull;
 use crates_io_database::fns::{array_agg, canon_crate_name, lower};
 
+/// The maximum number of crates that are ranked by relevance for a search
+/// query (see [`FilterParams::relevance_candidate_ids`]).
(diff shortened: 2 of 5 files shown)


walterhpearce:
Lgtm 🎉

LawnGnome:
I'm nervous about the behaviour change, but let's see if anyone complains in practice now this is deployed.

Turbo87:
an alternative would be to use https://github.com/postgrespro/rum, but that isn't available on most systems/hosts, so while the benchmarks looked promising I don't think it's really an option unfortunately

PR #13974, opened 2026-06-17
Add `versions.tar_sha256` column and dual-write it

Turbo87:
This stores the crate file SHA256 as raw bytes in a new `versions.tar_sha256` (`bytea`) column instead of the hex `checksum` (`char(64)`) column, halving the storage for that data on one of the largest tables. All external representations stay hex (the API `checksum` field and the git index `cksum`).

The `versions` table is too large to rewrite inside a release-phase migration, so the change is spread across four sequential deploys that each work against both the old and the new code, with the table scans run manually on the live database between them:

1. (this deploy) `tar_sha256` is added as a nullable column with a `NOT VALID` length check and dual-written alongside `checksum`. Existing rows are then backfilled from `checksum` and the constraints validated.
2. `tar_sha256` becomes `NOT NULL` and the digest is read from it.
3. `checksum` writes stop and the column becomes nullable.
4. `checksum` is dropped.

Each deploy is fully rolled out before the next one ships.

This first deploy adds the column and the length check, both instant DDL. `NewVersion` writes the raw digest to `tar_sha256` in addition to the hex `checksum`, with both values supplied by the caller so the database layer stays free of hex encoding. Reads are unchanged, and the new column is included in the public database dump.

### Related

- https://github.com/rust-lang/crates.io/pull/13973


eth3lbert (approved):
The migration path looks safe to me 👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/13974/0db9c53908fad5ec2d2d419831682db04bd143d0..d204888ad44c543ac23bf56e1fc7139167a2305b)*

PR #13996, opened 2026-06-18
Make `versions.tar_sha256` non-null and read from it

Turbo87:
This is the second deploy in the migration of the crate file SHA256 from the hex `checksum` column to the binary `tar_sha256` column. It assumes the first deploy has rolled out and that existing rows have been backfilled and the constraints validated on the live database, so it cannot ship until that is confirmed.

The migration sets `tar_sha256` to `NOT NULL`. That is instant in production because a `versions_tar_sha256_present` check, added and validated manually after the backfill, already proves every row is non-null, so Postgres skips the table scan. The migration then drops that helper check.

The `Version` model gains `tar_sha256` and stops reading `checksum`, so the public API `checksum` field and the git index `cksum` now derive from `hex::encode(tar_sha256)`. The output stays byte-for-byte identical. Removing the `checksum` read keeps the model from selecting the column once a later deploy makes it nullable. `NewVersion` still writes `checksum` for the previous release's readers.

### Related

- https://github.com/rust-lang/crates.io/pull/13974
- https://github.com/rust-lang/crates.io/pull/13973


Turbo87:
> It assumes the first deploy has rolled out and that existing rows have been backfilled and the constraints validated on the live database, so it cannot ship until that is confirmed.

I ran the following on staging and production:

```sql
WITH batch AS (
    SELECT id FROM versions
    WHERE tar_sha256 IS NULL
    LIMIT 10000
    FOR UPDATE SKIP LOCKED
)
UPDATE versions v
SET tar_sha256 = decode(v.checksum, 'hex')
FROM batch
WHERE v.id = batch.id;
```

(until it said `UPDATE 0`)

```sql
ALTER TABLE versions VALIDATE CONSTRAINT versions_tar_sha256_len;

ALTER TABLE versions
    ADD CONSTRAINT versions_tar_sha256_present
    CHECK (tar_sha256 IS NOT NULL) NOT VALID;

ALTER TABLE versions VALIDATE CONSTRAINT versions_tar_sha256_present;
```

eth3lbert (approved):

PR #14005, opened 2026-06-19
Submit service metrics to Datadog from the background worker

Turbo87:
The background worker now pushes the `cratesio_service` gauges (`crates_total`, `versions_total`, `background_jobs`) to Datadog's `POST /api/v2/series` API every 5 seconds, gated on `DD_API_KEY`. The existing Prometheus scrape keeps running in parallel so the two can be compared before anything is removed.

It also fixes `ServiceMetrics::gather()`, which reset the `background_jobs` gauge and only set the combinations that currently have jobs, so a drained `(priority, job)` series stopped being reported. That is fine for Prometheus pull but freezes the value on a push backend, so `gather()` now zeroes drained combinations instead of dropping them.


Turbo87:
FYI I've tested this on the staging environment and it appears to be working as intended :)

Bildschirmfoto 2026-06-19 um 17 13 44

(ignore the 6 broken `build_zip_archive` jobs...)

PR #14012, opened 2026-06-20
pagerduty: Remove redundant `Authorization` header and API token

Turbo87:
The `Authorization: Token token=<api_token>` header appears to be a PagerDuty REST API authentication mechanism. `PagerdutyClient`, however, posts to the Events API v1 endpoint (`.../generic/2010-04-15/create_event.json`), which is not part of the REST API and authenticates solely via the `service_key` in the JSON request body. The header was therefore ignored.

This drops the header along with the now-unused `api_token` parameter, the `authorization` field, and the `PAGERDUTY_API_TOKEN` environment variable read in `monitor.rs`.

The Events API v1 docs list `service_key` as a body parameter and make no mention of an `Authorization` header, and the endpoint has no `401` response code:

- https://developer.pagerduty.com/docs/events-api-v1-overview
- https://developer.pagerduty.com/docs/send-v1-event

The `Token token=` header belongs to the REST API instead:

- https://developer.pagerduty.com/docs/authentication

### DIFF
--- a/crates/crates_io_pagerduty/examples/test_pagerduty.rs
+++ b/crates/crates_io_pagerduty/examples/test_pagerduty.rs
@@ -28,9 +28,6 @@ impl FromStr for EventType {
 #[derive(clap::Parser, Debug)]
 #[command(name = "test-pagerduty", about = "Send a test event to pagerduty")]
 struct Opts {
-    #[arg(long, env = "PAGERDUTY_API_TOKEN", hide_env_values = true)]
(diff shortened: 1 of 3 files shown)


Turbo87:
tested and confirmed through the `test_pagerduty` example bin (see https://rust-lang.zulipchat.com/#narrow/channel/346298-t-crates-io.2Fincident-response/topic/Incident.20tb_test.20.28.23850.29/with/605127261)

PR #14020, opened 2026-06-21
Add a source code viewer for crate versions

Turbo87:
This adds a "Code" route on crate version pages that browses a published version's source directly from the CDN, with no new backend. Individual files are range-fetched out of the seekable `.zip` archive and its `.zip.json` manifest and inflated client-side with `DecompressionStream`, so only the viewed file is downloaded. The file tree renders with `@pierre/trees` and a single highlighted file with `@pierre/diffs`.

It is reachable by direct URL (`/crates/<name>/<version>/code/...`). The crate navigation does not link to it yet in this PR so that we can run a public beta test for this feature first.

Bildschirmfoto 2026-06-21 um 16 04 38

### Related

- https://github.com/rust-lang/crates.io/pull/13978
- https://github.com/rust-lang/crates.io/pull/14019

### DIFF
--- a/e2e/acceptance/crate-code.spec.ts
+++ b/e2e/acceptance/crate-code.spec.ts
@@ -1,0 +1,73 @@
+import { expect, test } from '@/e2e/helper';
+
+const SOURCE_FILES = {
+  'Cargo.toml': '[package]\nname = "serde"\n',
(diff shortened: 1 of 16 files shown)


Turbo87:
since this is strictly additional and the new route isn't linked anywhere yet I'll go ahead and get this merged so that we can beta test this in production :)

after-the-fact reviews still very welcome though!

PR #14035, opened 2026-06-23
crates_io_api_types: add `created_at` to `EncodablePublicUser`

### LINKED ISSUE #13867: Add `users.created_at`

LawnGnome:
Now that we have `users.created_at` and have backfilled it, the final step is to expose it in the public API.

The only actual changes are in `crates/crates_io_api_types/src/lib.rs`; everything else is updating tests, mostly snapshots. I've converted some integration tests that predate our use of `insta` to use `insta` along the way, since that allows us to use insta's redaction functionality like other parts of the codebase.

Finishes, and therefore closes #13867.

### DIFF
--- a/crates/crates_io_api_types/src/lib.rs
+++ b/crates/crates_io_api_types/src/lib.rs
@@ -718,6 +718,13 @@ pub struct EncodablePrivateUser {
     /// Whether the user has opted in to receive publish notifications via email.
     #[schema(example = true)]
     pub publish_notifications: bool,
+
+    /// The date and time the user was created.
+    ///
+    /// For users created before June 19, 2026, the creation time will be the
+    /// time the user's GitHub account was created. If the GitHub account was
+    /// deleted before June 19, 2026, this field will be empty.
+    pub created_at: Option<DateTime<Utc>>,
 }
 
 impl EncodablePrivateUser {
@@ -735,6 +742,7 @@ impl EncodablePrivateUser {
             gh_avatar,
             is_admin,
             publish_notifications,
+            created_at,
             ..
         } = user;
         let url = format!("https://github.com/{gh_login}");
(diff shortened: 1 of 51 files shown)


crates/crates_io_api_types/src/lib.rs
```diff
@@ -718,6 +718,13 @@ pub struct EncodablePrivateUser {
     /// Whether the user has opted in to receive publish notifications via email.
     #[schema(example = true)]
     pub publish_notifications: bool,
+
+    /// The date and time the user was created.
+    ///
+    /// For users created before June 19, 2026, the creation time will be the
+    /// time the user's GitHub account was created. If the GitHub account was
+    /// deleted before June 19, 2026, this field will be empty.
+    pub created_at: Option<DateTime<Utc>>,
```
Turbo87, line 727: should we also make this not null in a separate PR?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14035/f9849b46011ac41ff34eb84f01ed0ec076aef93e..35dc9b2a3e3ece824a9534b199338165ccffb482)*

LawnGnome: We can't; there are users who we can't backfill because their GitHub account is deleted.

LawnGnome:
Thanks for doing the API client step I missed!

crates/crates_io_api_types/src/lib.rs
Turbo87: I guess we could look at the known `created_at` before and after that account and then choose either of them or something in the middle, but I don't have a strong opinion on this :D

Issue #13766, opened 2026-05-26
Change update user API to allow for crates.io username changes with some restrictions


carols10cents:
This issue feels somewhat large; this can be broken up into multiple PRs if desired.

This depends on the PR adding reserved crate names: https://github.com/rust-lang/crates.io/issues/13765

This issue is only for the API changes; there is a separate issue for changes to the UI to make use of this new API ability.

In a similar manner as [the `deleted_crates` table](https://github.com/rust-lang/crates.io/blob/cc661a3157abac2d9faa510167367f27ec0808bd/migrations/2024-11-12-125605_create-deleted-crates-table/up.sql#L2), add a `deleted_usernames` (or `abandoned_usernames`? because this will be for usernames that someone decided to stop using by changing to something else, not necessarily by deleting their account).

- Add to [`update_user`](https://github.com/rust-lang/crates.io/blob/07920c13aa8d428105fa2ea6fc1cf74e803775b8/src/controllers/user/update.rs#L50) a field for `username` in the JSON

- Similarly to the [`validate_crate_name` function](https://github.com/rust-lang/crates.io/blob/07920c13aa8d428105fa2ea6fc1cf74e803775b8/crates/crates_io_validation/src/lib.rs#L83), create a `validate_username` function that:
  - Only allows usernames to contain alphanumeric characters (`[a-zA-Z0-9]`), hyphens (`-`), and underscores (`_`).
  - Doesn't allow usernames to start with a hyphen or underscore.
  - Doesn't allow usernames to exceed 39 characters.
  - Returns explanatory errors if any restrictions are violated.

- Similarly to the [check of `deleted_crates` in crate publishing](https://github.com/rust-lang/crates.io/blob/cc661a3157abac2d9faa510167367f27ec0808bd/src/controllers/krate/publish.rs#L136-L150), verify that:
  - The new username is available to use
  - The current user changing their name has not changed their name in the past 30 days

- Similarly to [the check of `reserved_crate_names` in crate publishing](https://github.com/rust-lang/crates.io/blob/cc661a3157abac2d9faa510167367f27ec0808bd/src/controllers/krate/publish.rs#L454-L456), verify that the new username isn't reserved

avirshup:
Hi, first time contributor, saw some discussion on mastodon and wanted to help if possible.

[...]

PR #14069, opened 2026-06-25
WIP: Add abandoned_usernames table

avirshup:
This is the 1st of 2 prs to implement #13766. This probably shouldn't be merged to main until the full patch series is ready.

This PR adds the "abandoned_usernames" table, which will be used to track usernames that were once in "adopted" by an active user but were later "abandoned" (due to a username change or account deletion). The content of the table is based on the [`deleted_crates`](https://github.com/rust-lang/crates.io/blob/main/migrations/2024-11-12-125605_create-deleted-crates-table/up.sql) table, including an `avaliable_at` field for a cooldown period before the username can be used again.

(FYI, in the diff, it looks like `diesel print-schema` seems to have swapped the order of `readme_renderings` and `recent_crate_downloads` in `schema.rs`, I'm not quite sure why.)


migrations/2026-06-24-234807-0000_create_abandoned_usernames_table/up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS abandoned_usernames
+(
+    id           BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    username     VARCHAR     NOT NULL,
+    adopted_by   INTEGER
+        CONSTRAINT fk_abandoned_usernames_adopted_by
+            REFERENCES users
+            ON DELETE SET NULL,
+    adopted_at   TIMESTAMPTZ NOT NULL,
```
Turbo87: I'm not sure this can be `NOT NULL`. Where would this information come from? We recently gained a `users.created_at` column, but that is nullable too, because we couldn't derive the data for all of our users.

migrations/2026-06-24-234807-0000_create_abandoned_usernames_table/up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS abandoned_usernames
+(
+    id           BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    username     VARCHAR     NOT NULL,
+    adopted_by   INTEGER
+        CONSTRAINT fk_abandoned_usernames_adopted_by
+            REFERENCES users
+            ON DELETE SET NULL,
```
Turbo87: I'm not sure about the `adopted_by` naming. My first thought was "this is the user that has adopted the name after the `available_at` timeout". I wonder if we can find an alternative name for this. maybe something like `previous_user_id`?

migrations/2026-08-07-120000-0000_create_abandoned_usernames_table/up.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE IF NOT EXISTS abandoned_usernames
+(
+    id           BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    username     VARCHAR     NOT NULL,
+    adopted_by   INTEGER
+        CONSTRAINT fk_abandoned_usernames_adopted_by
+            REFERENCES users
+            ON DELETE SET NULL,
```
Turbo87, line 8: just for context: currently we don't delete any users, we only set their username to "ghost" and delete all the personal information from their account. I guess technically the `ON DELETE SET NULL` shouldn't be necessary since we never delete. I currently don't remember what the default behavior is when you don't specify it. Does it prevent the user from getting deleted then? 🤔

Turbo87 (commented):
> This probably shouldn't be merged to main until the full patch series is ready.

I'm a bit hesitant to merge a new database table without it being used by anything, so yeah, I agree that we should probably wait with this until there is an actual user for this table.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14069/083c31115195a2e128dba973aea36ad95b422d13..a38bf50c164ec691f4fd457c86a503d3ea41dfae)*

migrations/2026-06-24-234807-0000_create_abandoned_usernames_table/up.sql
avirshup: Oh yeah, 100% agreed - if `users.created_at` is nullable, then `adopted_at` needs to be too.

> Where would this information come from?

The plan i to have this information come from A) `user.created_at`, if the user has never changed their username, otherwise B) the timestamp of their last username change - which will need to be stored in a new db column somewhere, as part of the next PR for this issue ([see also my question in #13766](https://github.com/rust-lang/crates.io/issues/13766#issuecomment-4804327811)).

migrations/2026-06-24-234807-0000_create_abandoned_usernames_table/up.sql
avirshup: Yep looking at it now,  I agree that 'adopted_by' reads like it happened _after_ it was abandoned. My other thoughts were to call this `was_used_by` (which I thought was too vague) or even, just, `userid`. `previous_user_id` seems good and is hopefully explicit enough!

migrations/2026-08-07-120000-0000_create_abandoned_usernames_table/up.sql
avirshup: >  I currently don't remember what the default behavior is when you don't specify it. Does it prevent the user from getting deleted then?

Yeah, unless it has the `ON DELETE` clause, a foreign key constraint like this prevents dropping the referenced rows.

> currently we don't delete any users, we only set their username to "ghost" and delete all the personal information from their account.

Ah, got it, thanks. Would your preference be to take out the `ON DELETE SET NULL` here, since it shouldn't matter? Alternatively I could just put in a comment saying that this is never expected to happen in normal circumstances.

avirshup:
Thanks for your comments @Turbo87, I appreciate your time! I'll make the changes now, but will hold off on marking this as ready for review until discussion in the parent issue is complete, and I have the complete patch set ready.

carols10cents:
> (FYI, in the diff, it looks like `diesel print-schema` seems to have swapped the order of `readme_renderings` and `recent_crate_downloads` in `schema.rs`, I'm not quite sure why.)

HMMMM I have that problem too!

EDIT: Aha, it's the schema.patch, [i've fixed this before](https://github.com/rust-lang/crates.io/pull/12792/changes/b8306ac9ade3b51083bb71ddd62bd65c17adae2c)..... hmmmmmmm

Issue #13769, opened 2026-05-26
Implement username disambiguation in `cargo owner` param handling and error messages


carols10cents:
This should only _require_ changes to the crates.io owners API in how it handles parameters and what error messages it might return. However, if changes to `cargo` would make this change more user-friendly, a cargo issue should be opened to discuss those changes with the Cargo Team.

From [the RFC](https://github.com/rust-lang/rfcs/pull/3946): 

> If you run `cargo owner --add example_username` and the account's crates.io username differs from
> the GitHub username associated with the account, the command will error with a message similar to:
> 
> ```console
> $ cargo owner --add example_username
> error: username `example_username` is possibly ambiguous
> 
> Caused by:
>   The crates.io account `example_username` is associated with:
> 
>   - https://github.com/something_else
>   - [any other accounts once we have that ability]
> 
>   To confirm this is the account you want to add, please run one of the following:
> 
>   $ cargo owner --add cratesio:example_username
>   $ cargo owner --add github:something_else
> 
>   If this is not the account you want to add, verify the crates.io username of the account you want.
> ```
> 
> Returning an error and requesting the user re-run a command with a disambiguation prefix to confirm
> is the easiest way to maintain compatibility with existing versions of Cargo. With some additional
> work on Cargo, newer versions could be made that only require a `y` or `n` confirmation; see the
> "Prior Art" section on Keybase for one possibility.

---

> The current API request for inviting user owners or adding team owners consists of a `PUT` request
> to `/api/v1/crates/[crate name]/owners` with the following JSON (using a request to add user
> `some_user` and team `some_team` from the `some_org` GitHub organization as an example):
> 
> ```json
> {
>     "owners": [
>         "some_user",
>         "github:some_org:some_team"
>     ]
> }
> ```
> 
> The backend processes owner strings starting with `github` and containing two colons as an organization name and a team name; this behavior will be unchanged.
> 
> This request will begin to accept owners specified by strings containing one colon and starting
> with `cratesio`[^1], `github`, and any other OAuth service we eventually add. An owner specification
> of `cratesio:some_user` will only query `users.username` and not any other table. An owner
> specification of `github:some_user` will only query `oauth_github.login` and not any other table.
> As other services are added, we will add a prefix that can be used to only look up usernames in
> that service's table. If the username isn't found in the specified table (say, the `cratesio`
> prefix that specfies the `users` table), the request will return an error even if the username is
> in another table (such as the `oauth_github` table, for this example).
> 
[...]

PR #14072, opened 2026-06-26
Stop updating user.username when a user renames or deletes their github account

### LINKED ISSUE #13761: Stop updating `user.username` when a user renames or deletes their github account
This depends on https://github.com/rust-lang/crates.io/pull/13143 being merged.

Both in the background job that was added in https://github.com/rust-lang/crates.io/pull/13143 and on login, only `oauth_github.login` should be updated and not `user.username`.

moskirathe:
fixes https://github.com/rust-lang/crates.io/issues/13761

Stop updating a user's username when a user updates or deletes their github account. This update currently happens on background job added in https://github.com/rust-lang/crates.io/pull/13143 and on login.

### DIFF
--- a/crates/crates_io_database/src/models/user.rs
+++ b/crates/crates_io_database/src/models/user.rs
@@ -125,7 +125,6 @@ impl NewUser<'_> {
             .do_update()
             .set((
                 users::gh_login.eq(excluded(users::gh_login)),
-                users::username.eq(excluded(users::username)),
                 users::name.eq(excluded(users::name)),
                 users::gh_encrypted_token.eq(excluded(users::gh_encrypted_token)),
             ))
--- a/src/tests/worker/update_user_from_github.rs
+++ b/src/tests/worker/update_user_from_github.rs
@@ -82,7 +82,6 @@ impl UpdateTest {
         // For now, we want to update the `User` record too
         let user_after_update = User::find(&conn, user_id).await?;
         assert_eq!(user_after_update.gh_login, expected_username);
-        assert_eq!(user_after_update.username, expected_username);
 
         if job_result.is_err() {
             // The worker leaves failed rows in `background_jobs` so they can
--- a/src/tests/worker/update_user_from_github.rs
+++ b/src/tests/worker/update_user_from_github.rs
@@ -82,6 +82,8 @@ impl UpdateTest {
         // For now, we want to update the `User` record too
         let user_after_update = User::find(&conn, user_id).await?;
         assert_eq!(user_after_update.gh_login, expected_username);
+        // The user's username should not be updated
+        assert_eq!(user_after_update.username, existing_gh_user.login);
 
         if job_result.is_err() {
             // The worker leaves failed rows in `background_jobs` so they can
--- a/src/tests/user.rs
+++ b/src/tests/user.rs
@@ -51,6 +51,8 @@ async fn updating_existing_user_doesnt_change_api_token() -> anyhow::Result<()> 
     let user = assert_ok!(User::find(&conn, api_token.user_id).await);
 
     assert_eq!(user.gh_login, "bar");
+    // updating existing user should not change their username
+    assert_eq!(user.username, "foo");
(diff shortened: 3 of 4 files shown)


Turbo87:
sorry, but we can't do this yet without offering users a way to rename their accounts themselves and a way to disambiguate `cargo owner add` calls

moskirathe:
> sorry, but we can't do this yet without offering users a way to rename their accounts themselves and a way to disambiguate `cargo owner add` calls

@Turbo87 sounds good, we can keep this in draft until its unblocked. are there other issues in https://github.com/rust-lang/crates.io/issues/13758 you think i could help with in the meantime?

carols10cents:
> are there other issues in #13758 you think i could help with in the meantime?

The `cargo owner add` disambiguation is https://github.com/rust-lang/crates.io/issues/13769, would you like to work on that? Even though it's not currently possible to have a crates.io username that's different from your GitHub username in the live crates.io site, it should be possible to create a `user` record with a `username` value that's different from the associated `oauth_github.login` in tests.

src/tests/worker/update_user_from_github.rs
```diff
@@ -82,7 +82,6 @@ impl UpdateTest {
         // For now, we want to update the `User` record too
         let user_after_update = User::find(&conn, user_id).await?;
         assert_eq!(user_after_update.gh_login, expected_username);
-        assert_eq!(user_after_update.username, expected_username);
```
carols10cents, line 85: Rather than removing this assertion, I think it should change to assert that `user_after_update.username` still equals `existing_gh_user.login`.

crates/crates_io_database/src/models/user.rs
```diff
@@ -125,7 +125,6 @@ impl NewUser<'_> {
             .do_update()
             .set((
                 users::gh_login.eq(excluded(users::gh_login)),
-                users::username.eq(excluded(users::username)),
```
carols10cents, line 128: We should probably have a test for this change in behavior; I'm a little surprised that we don't have one that failed.

carols10cents (changes requested):
Some suggestions for test improvements I'd like to see :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14072/03b00e2439f73980f756caf4b2aca2b1f276a907..f4d2a978839187c988883c70189b268e5d91211f)*

crates/crates_io_database/src/models/user.rs
carols10cents: Aha, [this test would have failed](https://github.com/rust-lang/crates.io/blob/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05/src/tests/user.rs#L53), but it's checking `users.gh_login` rather than `users.username`. That test should be changed to `username`, which should pass without this change, then fail with the change, then the assertion should be changed to assert that `users.username` stays the same but that the `oauth_github.login` changes.

moskirathe:
> Some suggestions for test improvements I'd like to see :)
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14072/03b00e2439f73980f756caf4b2aca2b1f276a907..f4d2a978839187c988883c70189b268e5d91211f)_

thanks for the review @carols10cents! addressed in https://github.com/rust-lang/crates.io/pull/14072/commits/2a601cea8c5109774158c48223cb8bf955d363ce

moskirathe:
> > are there other issues in #13758 you think i could help with in the meantime?
> 
> The `cargo owner add` disambiguation is #13769, would you like to work on that? Even though it's not currently possible to have a crates.io username that's different from your GitHub username in the live crates.io site, it should be possible to create a `user` record with a `username` value that's different from the associated `oauth_github.login` in tests.

@carols10cents yes, i'd love to tackle https://github.com/rust-lang/crates.io/issues/13769. I'll get started on it this weekend.

PR #14075, opened 2026-06-26
storage: Add CDN cache-tag metadata on uploads

Turbo87:
This PR adds CDN cache-tag metadata to newly uploaded `static.crates.io` objects.

`StorageKey::cache_tags()` returns the tag string for taggable objects:

- `.crate`, `.zip`, `.zip.json`, and readme objects carry both `crate:{name}` and `release:{name}@{version}`
- the OG image and per-crate feed carry only `crate:{name}`
- global objects (site-wide feeds, db-dumps) stay untagged

`attributes()` emits the tags as the `x-amz-meta-cache-tags` user-metadata header, which CloudFront and Fastly will read at cache-fill time to index each object under its tags (requires further changes in the `simpleinfra` repo).

This is additive and changes no existing behavior. New objects gain the metadata and nothing consumes the tags yet. The local-filesystem dev backend continues to no-op attributes, so only S3 and the in-memory test backend carry it.

### DIFF
--- a/src/config/features.rs
+++ b/src/config/features.rs
@@ -11,16 +11,23 @@ pub struct FeaturesConfig {
     ///
     /// Read from the `ZIP_ARCHIVES_ENABLED` environment variable.
     pub zip_archives_enabled: bool,
+
+    /// Emit CDN cache tags as storage metadata on uploads.
+    ///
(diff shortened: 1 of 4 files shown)


Turbo87:
I've added a feature flag so that we can test this out on staging without blocking production deploys in the meantime.

PR #14079, opened 2026-06-26
Make the "Code" tab public

Turbo87:
The "Code" tab was gated behind admin access while the feature was being stabilized (see #14029). As discussed in the team meeting today, these two commits make it generally available and clean up the now-redundant sidebar link.

### Related

- #14029


weihanglo:
This is really a nice feature. Thank you Tobi and crates-io team!

PR #14112, opened 2026-06-30
Explicitly set UTF-8 in the dev database collation and ctype

LawnGnome:
Without this, trying to import a database dump using `import-database-dump.sh` fails due to encoding errors in the tsvector trigger.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -27,7 +27,7 @@ services:
       # `semver_ord` stores prerelease identifiers as JSONB strings and
       # relies on byte-wise ordering. Force `C` collation so JSONB string
       # comparison matches the SemVer spec's ASCII sort order.
-      POSTGRES_INITDB_ARGS: '--lc-collate=C --lc-ctype=C'
+      POSTGRES_INITDB_ARGS: '--lc-collate=C.UTF-8 --lc-ctype=C.UTF-8'
       # Trust auth is fine because Postgres is only reachable on the
       # internal Docker network; no port is published to the host.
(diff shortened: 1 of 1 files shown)


.devcontainer/docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
       # `semver_ord` stores prerelease identifiers as JSONB strings and
       # relies on byte-wise ordering. Force `C` collation so JSONB string
       # comparison matches the SemVer spec's ASCII sort order.
-      POSTGRES_INITDB_ARGS: '--lc-collate=C --lc-ctype=C'
+      POSTGRES_INITDB_ARGS: '--lc-collate=C.UTF-8 --lc-ctype=C.UTF-8'
```
Turbo87: I'v checked the `datcollate` and `datctype` on the production database and they are both raw `C`. the `encoding` column on `pg_database` however is indeed `UTF8`. can you try it the following also fixes it for you? :)

```suggestion
      POSTGRES_INITDB_ARGS: '--lc-collate=C --lc-ctype=C --encoding=UTF8'
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14112/69e632a0a7aadc17d861c2bd76e7071e2c5bf2e3..cfd96fc58036c9040a9cc9c0a814c54c1e2c1bbf)*

LawnGnome: Yep, that also appears to fix it. So I'm good if you'd prefer to handle it that way.

Turbo87 (approved):

PR #14114, opened 2026-06-30
CI: Disable Postgres durability

Turbo87:
The "Backend / Test" job recreates the test database from scratch each run, so `fsync/synchronous_commit/full_page_writes` provide no value. Turning them off removes an fsync per commit and DDL statement. Locally this cut the per-test schema setup from ~460ms to ~245ms, and it speeds up every write the tests perform.


Turbo87:
nevermind, turns out this isn't actually any faster on CI 😢

Issue #14113, opened 2026-06-30
`test_generate_og_image_job` is flaky


LawnGnome:
`test_generate_og_image_job` fails reliably for me on both of my development setups due to minor kerning differences in the snapshots.

I've been putting up with this for a while and will probably have to continue to do so, but it occurred to me that I should document it somewhere in case I ever have time to investigate this.

In repo:

<img width="1200" height="630" alt="Image" src="https://github.com/user-attachments/assets/c85bd77a-4743-4181-b71f-ba799cd9e9aa" />

My result:

<img width="1200" height="630" alt="Image" src="https://github.com/user-attachments/assets/5ad98cca-d43a-4f96-bde1-f1e27454e391" />

I'm assuming this is something about it using the specific font file installed on my computer.

Turbo87:
damn... I hoped that since it worked fine between my macOS locally and the Linux CI, this wouldn't cause issues :-/

any ideas what we can do about it? 🤔

LawnGnome:
No, I really haven't had a chance to investigate this at all, even though it's been happening for a while for me.

My completely unproven hypothesis is that it's picking up the system font file if it's installed and that has subtly different metrics somehow, but it could be anything, really.

LaxentaInc:
'-'

carols10cents:
Is this test worth it?

LawnGnome:
> Is this test worth it?

In theory, yes, I think so. The problem is that what we really want to assert is "was an image generated with the right content and general styling", rather than "is this a pixel perfect clone of what our server would generate", but that's somewhat harder to assert.

Turbo87:
for context: the real test suite for the image generation lives in https://github.com/rust-lang/crates_io_og_image these days. this specific test is more of a smoke test and a check on whether we pass the right things into the `crates_io_og_image` lib.

PR #14117, opened 2026-07-01
BUG/DO NOT MERGE: og_image snapshot differs locally

carols10cents:
When I run the tests locally, the snapshot that's part of `worker::generate_og_image::test_generate_og_image_job` differs and fails the test.

I've opened this as a PR rather than an issue to get the nice image diff... am I missing a font or something?


Turbo87:
> am I missing a font or something?

yeah, you need the Fira Sans font

https://github.com/rust-lang/crates.io/blob/4a3f839747dd7e2e1cb2fd4e4948f5fd27a66c05/.github/workflows/ci.yml#L146-L149

though it looks like Adam has issues with this despite having the font installed (https://github.com/rust-lang/crates.io/issues/14113)

PR #14134, opened 2026-07-02
Move GithubTokenEncryption to its own crate and make it more general

carols10cents:
This is sort of part of https://github.com/rust-lang/crates.io/issues/10611, in that these types etc had "github" or "gh" in their names but aren't actually github specific. The changes in this PR should only be moves and renames, though, so this should be fairly low risk. 🤞🏻 

I also wanted to use the token encryption code from another of the sub-crates but couldn't because they were defined in the main crate and that would make a circular dependency :(

I have already taken the liberty of setting a new env var `TOKEN_ENCRYPTION_KEY` to the same value as `GITHUB_TOKEN_ENCRYPTION_KEY` in both prod and staging, but please check my work before deploying. I'm happy to redo if we decide to go with a different name. I also could add a fallback to the old name in `from_environment` if anyone thinks that would be helpful to continue to support old dev envs or anyone running a fork at the expense of leaving cruft around in the codebase.


crates/crates_io_encryption/src/lib.rs
```diff
@@ -1,50 +1,51 @@
...
+    /// Creates a new [`TokenEncryption`] instance from the environment
     ///
-    /// Reads the `GITHUB_TOKEN_ENCRYPTION_KEY` environment variable, which
+    /// Reads the `TOKEN_ENCRYPTION_KEY` environment variable, which
     /// should be a 64-character hex string (32 bytes when decoded).
     pub fn from_environment() -> Result<Self> {
-        let gh_token_key = std::env::var("GITHUB_TOKEN_ENCRYPTION_KEY")
-            .context("GITHUB_TOKEN_ENCRYPTION_KEY environment variable not set")?;
+        let token_key = std::env::var("TOKEN_ENCRYPTION_KEY")
+            .context("TOKEN_ENCRYPTION_KEY environment variable not set")?;
```
Turbo87, line 33: we will have to set the `TOKEN_ENCRYPTION_KEY` env var on staging/prod to the same value as `GITHUB_TOKEN_ENCRYPTION_KEY` before we merge/deploy this. after the deploy we can remove the `GITHUB_TOKEN_ENCRYPTION_KEY` env var, but during deploy both need to be available to avoid issues.

Turbo87 (approved):
looks good at first glance 👍 

I was a bit worried at first about creating confusion between generic and GitHub-specific data, but since only the encryption mechanism is renamed, not the data, I think this is fine.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14134/e7718f8063a828574baf41219976d8ded5b866c2..65743cc9b84853f907fea06bd26a2b5333dfcb53)*

crates/crates_io_encryption/README.md
```diff
@@ -0,0 +1,13 @@
+# `crates_io_encryption`
+
+This package implements encryption of data at rest in the database.
+
+The current usage is to encrypt/decrypt Oauth tokens; the types are named to indicate they're for
```
LawnGnome, line 5:
```suggestion
The current usage is to encrypt/decrypt OAuth tokens; the types are named to indicate they're for
```

(I'm sorry.)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14134/e7718f8063a828574baf41219976d8ded5b866c2..65743cc9b84853f907fea06bd26a2b5333dfcb53)*

LawnGnome (approved):

eth3lbert (approved):

crates/crates_io_encryption/src/lib.rs
carols10cents: Please see the PR description where I said: 

> I have already taken the liberty of setting a new env var TOKEN_ENCRYPTION_KEY to the same value as GITHUB_TOKEN_ENCRYPTION_KEY in both prod and staging, but please check my work before deploying. I'm happy to redo if we decide to go with a different name. I also could add a fallback to the old name in from_environment if anyone thinks that would be helpful to continue to support old dev envs or anyone running a fork at the expense of leaving cruft around in the codebase.

😁

PR #14153, opened 2026-07-05
Replace `asdf` with `mise`

Turbo87:
see https://mise.jdx.dev/

This first changes the buildpack to use `mise` to read the `.tool-versions` file and install the tools (Node.js and pnpm). Then the second commit replaces the `.tool-versions` file with the native `mise.toml`.

In the future this will make it possible to also install/run/use version-pinned tools like `diesel_cli` or `diesel-guard` via `mise` too :)


Turbo87:
I've tested this on staging already, and it appears to work as planned :)

PR #14170, opened 2026-07-07
Extract User builder for tests to reduce duplication

carols10cents:
There are a bunch of places that created users for tests. Rather than do that more, I've extracted a `UserBuilder` to be like the `CrateBuilder` etc and be the one canonical way to create users in tests.

I have left the `new_user` function around for now but implemented it in terms of the `UserBuilder`. The `new_user` function is still only available in the top-level crate, so if you want to create users in tests in some other crate, you should reach for using `UserBuilder` directly. I thought about removing `new_user` to encourage use of `UserBuilder` directly, but I wasn't sure whether that was too much change or not.

This is why I wanted https://github.com/rust-lang/crates.io/pull/14134; now the `crates_io_test_utils` crate can create and use a `TokenEncryption::for_testing()`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2241,6 +2241,7 @@ dependencies = [
  "crates_io_api_types",
  "crates_io_cargo_toml",
  "crates_io_database",
+ "crates_io_encryption",
  "crates_io_tarball",
  "diesel",
  "diesel-async",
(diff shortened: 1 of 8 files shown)


Turbo87:
https://github.com/rust-lang/crates.io/pull/14171 should fix the RustSec issue. once it's merged you can rebase :)

Turbo87 (approved):

PR #14175, opened 2026-07-08
Test data passed to OG image generation instead of asserting on exact bytes

### LINKED ISSUE #14113: `test_generate_og_image_job` is flaky

carols10cents:
Fixes #14113.

Rather than asserting that the exact bytes/pixels of the rendered OG image are exactly the same, which isn't always the case on different systems, assert that we're passing the OG image generation the data we think it should be.

The first commit in this PR actually changes the snapshot, because the generated image didn't actually have all the data we expected to have 😂 The test created a crate with keywords, but the keywords didn't show up in the image because the test setup was only setting the keywords on the crate record and the og image generation gets keywords from the version record.

So I fixed the test setup and added other data to the test setup that og image generation uses, then added assertions to the parts of the og image generation that gather the og image data passed to the image generator.

I thought about creating and using a mock image generator, but that was too much of a pain, so the new test is asserting on the results of private functions that gather the data passed to the image generator.

Then once that test is in place, the value of the image snapshot assertion becomes even lower, so the last commit removes it.


crates/crates_io_test_utils/src/builders/version.rs
```diff
@@ -89,6 +93,18 @@ impl VersionBuilder {
         self
     }
 
+    /// Sets the crate's `description` value.
```
Turbo87:
```suggestion
    /// Sets the version's `description` value.
```

Turbo87 (approved):
one small copy-paste issue, but other than that this seems okay to me. we still have the `crates_io_og_image` test suite for the visual verification :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14175/cb8c86de03abd9d768801f0a602a3755609eaf05..067499c3cc7133492e9a05d00d68c0aa7a424f60)*

LawnGnome:
My local development setup thanks you! :bow: :tada:

PR #14188, opened 2026-07-09
Send messages to docs.rs after crate actions

### LINKED ISSUE #14111: Notify docs.rs when relevant actions occur
We discussed this at this year's All Hands: see [the meeting notes](https://github.com/rust-lang/all-hands-2026/issues/5#issuecomment-4508462642) for details.

Basically, the agreed upon path is that crates.io will send events to an SQS queue when a crate is published, deleted, yanked, or unyanked. Note that we do not intend to generalise this into a general notification mechanism for other users of crates.io — while we know there is some interest in that, this approach is something that is only sensible for a fellow rust-lang project.

@syphar has opened https://github.com/rust-lang/docs.rs/pull/3357 on the docs.rs with a draft shape for these messages. The next step is for me to use this to implement this on the crates.io side to ensure that we don't have any unexpected technical icebergs.

(I thought we had an issue for this in crates.io already, but I can't find it, so new issue it is. My apologies if I'm wrong about that.)

LawnGnome:
This is a sketch of how we might close #14111 by sending messages to docs.rs. It is best reviewed commit-by-commit.[^prs]

I do not expect to merge this in its current form; this draft PR is primarily to gather feedback.

## How it works

Very broadly, we need to do the following:

* Refactor the existing configuration logic used for `CdnLogQueueConfig` to be generic so we can reuse it for the docs.rs queue.
* Add a docs.rs queue.
* Add message sending functionality to `SqsQueue`.
* Change the `SyncToGitIndex` job to include the event that triggered the sync[^optional], which can then be turned into a message.
* Actually send the message.

## Open questions

I'm not particularly set on this particular implementation. Things I'd like feedback on include:

1. Should this live in the `SyncToGitIndex` job at all?
   1. If not, should this be a job triggered by `SyncToGitIndex`, or should this be decoupled from the Git index sync altogether?[^docsrs]
2. How do we alert on this when sending a message fails, and what happens then?
3. Are the new `SqsQueueImpl` tests actually useful?
4. Similarly, is the smoke test now too big and clunky to be useful in its current form?[^noise]

## Other prerequisites and testing

This depends on https://github.com/rust-lang/simpleinfra/pull/1099. Once that's merged, we could deploy this (or whatever comes out of this) to staging and start testing with a stub consumer to ensure that we get the messages we expect to see.

I'm unsure if the docs.rs team wants to actually publish the `docs_rs_crates_io` crate, but if they do, then that would also need to happen.

[^docsrs]: The main motivator behind tying this to the Git index sync is that docs.rs uses the Git index to pull crate metadata, so having the message only be sent after the Git index has already been updated means that docs.rs doesn't need to care about whether the update has occurred.
[^noise]: In particular, I don't love how big that function has now gotten with the usage of `MockSqsQueue`.
[^optional]: There are still cases where we won't have a trigger that's relevant to docs.rs, most notably manual syncs to the index initiated via `crates-admin`.
[^prs]: I expect that, in practice, we might want to split the SQS commits into a separate PR since they're technically a separate refactor.


src/sqs.rs
```diff
@@ -71,6 +98,21 @@ impl SqsQueue for SqsQueueImpl {
     }
 }
 
+#[async_trait]
+impl<T: SqsQueue + Send + Sync + ?Sized> SqsQueue for Arc<T> {
+    async fn receive_messages(&self, max_messages: i32) -> anyhow::Result<ReceiveMessageOutput> {
+        (**self).receive_messages(max_messages).await
+    }
+
+    async fn send_message(&self, message: String) -> anyhow::Result<SendMessageOutput> {
```
carols10cents, line 107: Nit: As of this commit, the `SqsQueue` trait doesn't have the `send_message` method on it. That is, if you check out this commit, I don't think it will build. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14188/34b702a48221df4e9ecc7459304ec702d80fc540..8d1389a2585c06ce1d2f6a8faa1e9d7c64398527)*

Cargo.toml
```diff
@@ -122,6 +122,7 @@ diesel = { version = "=2.3.10", features = ["postgres", "serde_json", "chrono",
 diesel-async = { version = "=0.9.2", features = ["async-connection-wrapper", "deadpool", "postgres"] }
 diesel_full_text_search = "=2.3.0"
 diesel_migrations = { version = "=2.3.2", features = ["postgres"] }
+docs_rs_crates_io = { git = "https://github.com/rust-lang/docs.rs.git", rev = "1604301c72411fbc63082434295fa45036d982b7", version = "0.1.0" }
```
carols10cents, line 125: just pointing out for other reviewers that the `docs_rs_crates_io` crate and this rev are from https://github.com/rust-lang/docs.rs/pull/3357

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14188/34b702a48221df4e9ecc7459304ec702d80fc540..8d1389a2585c06ce1d2f6a8faa1e9d7c64398527)*

carols10cents:
🤷🏻‍♀️ seems fine to me

src/sqs.rs
Turbo87: yeah, I think this was meant to be introduced two commits later in the "Add `SqsQueue::send_message`" commit

Cargo.toml
Turbo87: and I assume we won't rely on a git dependency once we merge this, right? 😅

src/worker/jobs/index/sync.rs
```diff
@@ -10,17 +12,93 @@ use serde::{Deserialize, Serialize};
...
 #[derive(Serialize, Deserialize)]
 pub struct SyncToGitIndex {
     krate: String,
+
+    /// The triggering event that led to the sync.
+    ///
+    /// Once the index has been synced, this event will be sent to docs.rs via
+    /// SQS.
+    #[serde(default)]
+    trigger: Option<Trigger>,
```
Turbo87, line 28: unfortunately I think this defeats the purposed of the `const DEDUPLICATED: bool = true;` below. we explicitly introduced the deduplication to avoid having to sync every update in isolation to the indexes because it was causing performance problems. if we include the triggers then either we don't deduplicate anymore, or we keep deduplicating but docs.rs might miss deduplicated events, or both.

it might also block us from potentially batching index updates in the future for related reasons.

does docs.rs really need this information? we can easily give them the added/updated/deleted information, but the yanked/unyanked/version-deleted would be slightly more complicated since we might combine multiple of these in a single commit.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14188/34b702a48221df4e9ecc7459304ec702d80fc540..8d1389a2585c06ce1d2f6a8faa1e9d7c64398527)*

src/worker/jobs/index/sync.rs
```diff
@@ -39,6 +117,7 @@ impl BackgroundJob for SyncToGitIndex {
 
         let crate_name = self.krate.clone();
         let mut conn = env.deadpool.get().await?;
+        let queue = env.docs_rs_queue.clone();
```
Turbo87, line 120: we should probably make this queue optional, so that it does not break the publishing flow when developing locally without having a queue configured. the default-to-mock behavior for the other queue was okayish in the past because the corresponding background job that used the queue had to be triggered manually anyway, but this one is triggered by the regular publish flow and yank/unyank. this might involve changing `QueueConfig::from_env()` to return `Result<Option<Self>>` or something like that.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14188/34b702a48221df4e9ecc7459304ec702d80fc540..8d1389a2585c06ce1d2f6a8faa1e9d7c64398527)*

src/worker/jobs/index/sync.rs
```diff
@@ -79,9 +158,27 @@ impl BackgroundJob for SyncToGitIndex {
...
+            // what failed so we can tell docs.rs.
+            //
+            // TODO: is this alertable? How can we notify docs.rs that rebuilds
+            // may be required?
+            error!(
+                krate.name = &self.krate,
+                trigger = ?self.trigger,
+                error = %e,
+                "error notifying docs.rs",
+            );
```
Turbo87, line 178: you could do something like `warn!(target: "docs_rs_queue", ...)` which makes it easy to filter on in Datadog

I'd also recommend to not use structured logging too much unless you plan on filtering by these fields, and including the fields in the log message too, so that not every log row on Datadog reads "error notifying docs.rs" without any context. something like "Failed to notify docs.rs about {trigger} update on {krate} crate: {error}" is much easier to comprehend at a glance without having to dig into the details of every log event (see https://github.com/rust-lang/crates.io/blob/main/docs/LOGGING.md) :)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14188/34b702a48221df4e9ecc7459304ec702d80fc540..8d1389a2585c06ce1d2f6a8faa1e9d7c64398527)*

Cargo.toml
syphar: @LawnGnome does this imply that the message schema is fine now? So we can merge / release in docs.rs? 

> and I assume we won't rely on a git dependency once we merge this, right? 😅

Just to have it written down: yes of course.

src/worker/jobs/index/sync.rs
```diff
@@ -79,9 +158,27 @@ impl BackgroundJob for SyncToGitIndex {
...
 
-            Ok(())
+            anyhow::Ok(())
         })
-        .await?
+        .await??;
+
+        // This is implemented inline for now, but we may want to enqueue
+        // another job to handle sending the actual messages.
+        if let Err(e) = self.notify_docs_rs(queue).await {
```
syphar, line 167: do the crates.io jobs have retries? 

That would probably help catching the most temporary issues. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14188/34b702a48221df4e9ecc7459304ec702d80fc540..8d1389a2585c06ce1d2f6a8faa1e9d7c64398527)*

src/worker/jobs/index/sync.rs
```diff
@@ -79,9 +158,27 @@ impl BackgroundJob for SyncToGitIndex {
...
+            // We won't consider this fatal right now, but we do need to know
+            // what failed so we can tell docs.rs.
+            //
+            // TODO: is this alertable? How can we notify docs.rs that rebuilds
```
syphar, line 171: generally I like "self-healing" designs, 

in this case perhaps a scheduled "repair" job based on the git index, or the crates.io DB? 

( or relying less on crates.io data / sync generally). 

I'm not sure yet how I would handle the "live" SQS events vs a (less often / hourly) updated git index, and the repair job. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14188/34b702a48221df4e9ecc7459304ec702d80fc540..8d1389a2585c06ce1d2f6a8faa1e9d7c64398527)*

src/worker/jobs/index/sync.rs
Turbo87: > do the crates.io jobs have retries?

yes

src/worker/jobs/index/sync.rs
Eh2406: Could the live SQS event include the commit hash from the get index? "Version 1.2.3 of Crate foo Was published into the git index at commit hash 1243534534". Then it should be possible To write an accurate repair job.
1. Pause reading real time events.
2. Check out the index as of the most recently processed real time event.
3. Check that everything in the index (or all the changes since the last repair) have been correctly ingested.
4. Go back to reading real time events.

I can dredge up some horror stories of what it's actually like trying to run a service that has both real time ingestion from registries while also having occasional repair tasks from the same registries when there is no way to correlate between the real time events and their snapshots. But I'd rather not have to put them in writing. I'd prefer to hop on a call.

src/worker/jobs/index/sync.rs
syphar: But I assume they wouldn't kick in if the error is consumed like it is right now?

syphar: Could be a good reason for the separate job for the msg publish later

src/worker/jobs/index/sync.rs
syphar: aren't crates.io logs json logs? Datadog makes using them (also making then nicely viewable) quite easy.

syphar: ah, sorry, the logging.md explains more

src/worker/jobs/index/sync.rs
syphar: From what I remember, at some point the git index will be updated less often (hourly?) and we will still get the SQS events directly from crates.io.

Cargo.toml
syphar, line 125: This is released now: 
https://docs.rs/docs_rs_crates_io/latest/docs_rs_crates_io/events/index.html

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14188/34b702a48221df4e9ecc7459304ec702d80fc540..8d1389a2585c06ce1d2f6a8faa1e9d7c64398527)*

Issue #13763, opened 2026-05-26
Update the Find/Show User API to query users.username


carols10cents:
~~Depends on https://github.com/rust-lang/crates.io/issues/13762. Once `users.username` is unique, that can be used for lookup.~~ 

We can change this API to look for the username from the URL in `users.username` instead of `users.gh_login` and continue to use the `order` to resolve ambiguity as done here: https://github.com/rust-lang/crates.io/blob/a313be6dfb6d67791f208fedf6c2129a3b47ccce/src/controllers/user/other.rs#L41

The rest of the tasks have been moved to sub-issues.

PR #14213, opened 2026-07-12
Cargo owner username disambiguation

### LINKED ISSUE #13769: Implement username disambiguation in `cargo owner` param handling and error messages

moskirathe:
fixes: https://github.com/rust-lang/crates.io/issues/13769


crates/crates_io_database/src/models/oauth_github.rs
```diff
@@ -0,0 +1,98 @@
+use bon::Builder;
```
moskirathe: i moved the implementation in this file from `users.rs` (see https://github.com/rust-lang/crates.io/pull/14213/files#diff-968fca8c098c2d6ddbef45dd59a130d21595aeae288e6bfe6232931e62e5618aL137-L214) to decouple the model definitions.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..4e89b5a61053e0397a106b880a53d50407fe7425)*

crates/crates_io_database/src/models/owner.rs
```diff
@@ -104,7 +104,7 @@ impl Owner {
 
     pub fn login(&self) -> &str {
         match self {
-            Owner::User(user) => &user.gh_login,
+            Owner::User(user) => &user.username,
```
moskirathe, line 107: i updated _most_ applicable uses of `user.gh_login` to use `user.username` instead, except for a few cases where we explicitly need `user.gh_login` such as when we check for github team membership. i implemented this change in https://github.com/rust-lang/crates.io/pull/14213/commits/d719b95abe558d47a71a2f94d0e724846affe61c but let me know if it would be better to update these in a separate issue.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..4e89b5a61053e0397a106b880a53d50407fe7425)*

crates/crates_io_database/src/models/oauth_github.rs
Turbo87: FWIW that makes the review significantly harder since it couples refactorings with behavior changes in the same PR

crates/crates_io_database/src/models/owner.rs
moskirathe: split these refactoring changes from this pr, to be tacked in https://github.com/rust-lang/crates.io/issues/13772 or https://github.com/rust-lang/crates.io/issues/13763

crates/crates_io_database/src/models/oauth_github.rs
moskirathe: removed. this refactoring change can probably be tackled as part of https://github.com/rust-lang/crates.io/issues/13772

src/controllers/krate/owners.rs
```diff
@@ -323,26 +325,130 @@ async fn add_owner(
...
+
+        let username = chunks.next().unwrap();
+        return match prefix {
+            "github" => Ok(Login::GitHub(username)),
+            "cratesio" => Ok(Login::CratesIo(username)),
+            _ => Err(bad_request(
+                "unsupported username prefix, only github and cratesio prefixes are supported",
+            )),
+        };
+    }
```
Turbo87: Could we reject owner specifications unless the parser consumes the complete input? `github:alice:team:extra` misses the team regex, then this branch reads only `github` and `alice` and ignores `team:extra`. If GitHub user `alice` exists, add invites that user and remove can remove that user instead of rejecting malformed input.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/controllers/krate/owners.rs
```diff
@@ -323,26 +325,130 @@ async fn add_owner(
...
+/// Disambiguate a login string
+async fn disambiguate_login<'a>(
+    conn: &mut AsyncPgConnection,
+    login: &'a str,
+    add: bool,
+) -> Result<Login<'a>, BoxedAppError> {
+    // Team login: exactly two colons (e.g. github:org:team)
+    if Regex::new(r"^[^:]+:[^:]+:[^:]+$").is_ok_and(|r| r.is_match(login)) {
+        return Ok(Login::Team);
+    }
```
Turbo87: I'm not sure we need a regex for this. We could use something like `login.split(':')` and `iter.next()` to get the individual parts and validate them.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/controllers/krate/owners.rs
```diff
@@ -323,26 +325,130 @@ async fn add_owner(
...
+enum Login<'a> {
+    /// A team login e.g `github:org:team`
+    Team,
+    /// A disambiguated `github:username` resolved via `oauth_github`.
+    GitHub(&'a str),
+    /// A disambiguated `cratesio:username` resolved via users table.
+    CratesIo(&'a str),
+    /// An unambigous username (`users.username` == `oauth_github.login`)
+    Username(User),
+}
```
Turbo87, line 528: I wonder if we should either adjust this enum or create another one that just handles the login string parsing and representation. I'd imagine something like this:

```suggestion
/// Parsed login string representation
enum Login<'a> {
    /// GitHub organization team (e.g `github:org:team`).
    GitHubTeam { org: &'a str, team: &'a str },
    /// GitHub user (e.g. `github:username`).
    GitHub(&'a str),
    /// crates.io user (`crates.io:username`).
    CratesIo(&'a str),
    /// Unprefixed username (`username` without any prefix)
    Unprefixed(&'a str),
}
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/controllers/krate/owners.rs
```diff
@@ -323,26 +325,130 @@ async fn add_owner(
...
+        .select(OauthGithub::as_select())
+        .first(conn)
+        .await
+        .optional()?;
+
+    let command = if add { "add" } else { "remove" };
+    let username = user.username.to_owned();
+
+    if let Some(oauth_github) = oauth_github
+        && oauth_github.login != user.username
```
Turbo87: The surrounding username lookups are case-insensitive, but this comparison is case-sensitive. A crates.io username `Alice` associated with GitHub login `alice` would be reported as ambiguous even though both namespaces treat those names as equal. Could we compare using the same canonicalization and add a case-only regression test? `_` vs. `-` might also be relevant for this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

crates/crates_io_database/src/models/user.rs
```diff
@@ -161,6 +164,19 @@ pub struct OauthGithub {
     pub user_id: i32,
 }
 
+impl OauthGithub {
+    pub async fn find_by_login(
+        mut conn: &AsyncPgConnection,
+        login: &str,
+    ) -> QueryResult<OauthGithub> {
+        oauth_github::table
+            .filter(lower(oauth_github::login).eq(login.to_lowercase()))
```
Turbo87: `oauth_github.login` is not unique (yet), so this unordered `.first()` can resolve `github:<name>` to an arbitrary crates.io account. The previous `User::find_by_login()` explicitly ordered by descending GitHub ID. Could we preserve the intended deterministic tie-breaker, or reject multiple matches, and cover duplicate case-insensitive logins in a regression test? 🙏 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/controllers/krate/owners.rs
Turbo87: This ownership check runs before the prefixed input is resolved. For an existing owner with crates.io username `alice` and GitHub login `alice-gh`, both `cratesio:alice` and `github:alice-gh` bypass this comparison and create a pending invitation for the existing owner. Could we resolve the specification first, then compare the resulting owner kind and ID, with regression tests for both prefixes?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/controllers/krate/owners.rs
```diff
@@ -323,26 +325,130 @@ async fn add_owner(
...
+) -> Result<(), OwnerRemoveError> {
+    match disambiguate_login(conn, login, false)
+        .await
+        .map_err(OwnerRemoveError::from)?
+    {
+        Login::Team => krate.owner_remove_with_username(conn, login).await,
+        Login::GitHub(login) => krate.owner_remove_with_gh_login(conn, login).await,
+        Login::CratesIo(username) => krate.owner_remove_with_username(conn, username).await,
+        Login::Username(_) => krate.owner_remove_with_username(conn, login).await,
+    }
```
Turbo87: Removal needs to resolve the input against the crate’s current owners, rather than reuse the global add resolver. In `test_remove_ambiguous_user`, `user2` is the only current owner matching that name, so issue #13769 says the unprefixed removal should succeed despite its linked GitHub name differing. The model queries also update every owner sharing the supplied service login, which is unsafe because those columns are not unique. Could we resolve the specification to exactly one current owner first, reject zero or multiple matches, and delete that owner by kind and ID?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/controllers/krate/owners.rs
```diff
@@ -524,13 +630,20 @@ impl From<BoxedAppError> for OwnerAddError {
     }
 }
 
+impl From<BoxedAppError> for OwnerRemoveError {
+    fn from(value: BoxedAppError) -> Self {
+        Self::AppError(value.to_string())
+    }
+}
```
Turbo87: This conversion loses the response semantics carried by `BoxedAppError`. A database read-only error that should remain a 503, or an internal Diesel failure that should become a logged 500, is stringified here and later rebuilt with `bad_request()`, potentially exposing the internal message as a 400. Could we retain the boxed error, as `OwnerAddError` does, and potentially add a test that preserves a non-400 status through the remove path?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

crates/crates_io_database/src/models/user.rs
```diff
@@ -42,9 +42,12 @@ impl User {
...
     }
 
-    pub async fn find_by_login(mut conn: &AsyncPgConnection, login: &str) -> QueryResult<User> {
+    pub async fn find_by_username(
+        mut conn: &AsyncPgConnection,
+        username: &str,
+    ) -> QueryResult<User> {
         User::query()
-            .filter(lower(users::gh_login).eq(login.to_lowercase()))
+            .filter(lower(users::username).eq(username.to_lowercase()))
```
Turbo87: Both new lookup paths filter on `lower(...)`, but there are no matching functional indexes: the existing user index covers `lower(gh_login)`, while the OAuth table only has an index on `user_id`. A request can resolve up to ten owners, causing repeated full scans of these identity tables. Could we add indexes for `lower(users.username)` and `lower(oauth_github.login)` as part of this change? Or given https://github.com/rust-lang/crates.io/pull/13903 I think we might actually want `canon_username(users.username)` and `lower(oauth_github.login)`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/controllers/krate/owners.rs
```diff
@@ -323,26 +325,130 @@ async fn add_owner(
...
+        let gh_login = &oauth_github.login;
+        let error = format_args!(
+            "error: username {username} is possibly ambiguous.\n\n\
+                Caused by: \n  \
+                The crates.io account `{username}` is associated with GitHub user `{gh_login}`.\n  \
+                To confirm this is the account you want to {command}, please run one of the following:\n\n  \
+                $ cargo owner --{command} cratesio:{username}\n  \
+                $ cargo owner --{command} github:{gh_login}\n\n  \
+                If this is not the account you want to {command}, verify the crates.io username of the account you want.",
+        );
```
Turbo87: I think `cargo` already adds its own `error:` prefix, operation context, and `Caused by:` chain around registry API failures. Including those markers in the JSON detail produces a nested diagnostic such as `... responded with an error: error: username ... Caused by: ...`. Could the API return only the semantic message and guidance, matching the JSON example in issue #13769?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/bin/crates-admin/delete_crate.rs
Turbo87:
```suggestion
    /// Your crates.io username.
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

src/controllers/krate/owners.rs
Turbo87: The public schema still documents only bare user names and `github:org:team`, so it does not expose the new `cratesio:<username>` and `github:<username>` forms or explain when a bare name requires disambiguation. I think we should update the field documentation and example to explain this and update the corresponding OpenAPI schema snapshot and generated JS API client accordingly.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/b791849c1c64bae97e3237a31eae3b3b8deb25a2..c39e10ab4c949dcb86982b446294bbbf80451626)*

Turbo87 (changes requested):

crates/crates_io_database/src/models/user.rs
moskirathe: good catch. i've added functional indexes for both columns.

crates/crates_io_database/src/models/user.rs
moskirathe: updated to preserve original functionality of explicitly ordering by Github ID.

src/controllers/krate/owners.rs
moskirathe: updated enum 👍

moskirathe:
@Turbo87 thanks for the review! I've addressed your comments. the pr is ready for re-review.

crates/crates_io_database/src/models/user.rs
```diff
@@ -22,6 +22,9 @@ pub struct User {
     pub name: Option<String>,
     pub gh_id: i32,
     pub gh_login: String,
+    // Rename this field to gh_login when gh_login is removed from this struct.
+    #[diesel(select_expression = oauth_github::login.nullable())]
+    pub gh_username: Option<String>,
```
moskirathe, line 83: this is the same as `gh_login` but reading from `oauth_github` instead of the users table. I considered a change to stop reading from `users.gh_login` and start reading from `oauth_github.login` throughout the codebase instead (similar to https://github.com/rust-lang/crates.io/pull/14258), but I think that would be a much bigger change that would warrant being tackled in a separate pr

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/887e23cd9b2b1f2a5299c0e544682b47bc4a63d4..8a44eaea6a02da08e920a3debc6044c4c3ee49b7)*

moskirathe:
@carols10cents would also love your review on this when you have some time

src/controllers/krate/owners.rs
Turbo87: Thanks, the case-only comparison and regression tests look fixed. One part is still inconsistent with the new canonical username behavior, though: this comparison uses lowercase only, so `foo-bar` and `foo_bar` are still treated as different. Since the new index and #13903 use `canon_username()`, could we use the same canonicalization here and add a separator regression test?

crates/crates_io_database/src/models/user.rs
Turbo87: Thanks, ordering by descending `account_id` restores the deterministic selection. Could we also add the duplicate case-insensitive login regression test requested here? That is the state which makes the ordering observable and protects it from being removed again later.

src/controllers/krate/owners.rs
Turbo87: Parsing before the check fixes the prefix bypass in the simple cases, but the identity is still not resolved before comparison. With two OAuth rows sharing a login, an older account can already own the crate while `find_by_login()` would select a newer, different account. The current `owners.iter().any(...)` rejects that as already owned. Could we resolve the owner kind and ID first, then compare that identity against the current owners?

crates/crates_io_database/src/models/user.rs
Turbo87: The indexes themselves now have the intended expressions, but the users lookup still filters on `lower(users::username)`. PostgreSQL cannot use the `canon_username(username)` index for that expression, and the lookup does not canonicalize `-` and `_`. Could we change the query to use `canon_username()` consistently on both sides and add coverage for the separator case?

src/controllers/krate/owners.rs
```diff
@@ -321,28 +341,166 @@ async fn add_owner(
...
+enum Login<'a> {
+    /// GitHub organization team (e.g `github:org:team`). the original login is preserved as a convenience to avoid rebuilding it.
+    GitHubTeam {
+        login: &'a str,
+        org: &'a str,
+        team: &'a str,
+    },
+    /// GitHub user (e.g. `github:username`).
+    GitHub(&'a str),
+    /// crates.io user (`crates.io:username`).
```
carols10cents, line 524: I think this comment should be:


```suggestion
    /// crates.io user (`cratesio:username`).
```

based on the rest of this PR, right?

carols10cents (changes requested):
Took a quick look and saw one little thing, and I agree with the comments @Turbo87 made.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/90c9cca210ce8d1dbf59204491c5b3661342db22..4c528c395b2d703899ba03461aeee3a1d9f66560)*

src/controllers/krate/owners.rs
Turbo87: I think I'd prefer `crates.io`. is there a technical reason why we would need to drop the `.`?

crates/crates_io_database/src/models/user.rs
moskirathe: updated to use `canon_username` and added coverage for separator cases. thanks!

src/controllers/krate/owners.rs
moskirathe: i fixed this in previous commits but let me know if you think i missed something

src/controllers/krate/owners.rs
moskirathe: right! i've updated the implementation to resolve potentially ambiguous logins first before checking for ownership. i also added regression tests for this.

src/controllers/krate/owners.rs
moskirathe: there is no technical reason for dropping the `.` here, but i updated this to be consistent with all other comments and error messages that include `cratesio:username`. i don't have a strong preference for either `cratesio` or `crates.io`, so happy to update this to `crates.io` everywhere if you think it would be better.

moskirathe:
@Turbo87 @carols10cents ready for re-review. thanks!

migrations/2026-07-18-120000-0000_add_users_username_index/up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX CONCURRENTLY IF NOT EXISTS index_users_username ON users (canon_username(username));
```
moskirathe: noticed we added this index in https://github.com/rust-lang/crates.io/pull/14417, will remove from here once https://github.com/rust-lang/crates.io/pull/14417 is merged.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/0584a2df65968a4e9e6859ef46bbed430408a3f1..c1091af697afd17d29c913707891aca43bca939a)*

src/controllers/krate/owners.rs
Turbo87: yeah, I think unless there is a technical reason why we would need to avoid the `.` character we should use our usual brand name and adjust it in the other "comments and error messages" too :)

moskirathe: sounds good. updated in 065fdec41c50cf06e1c32e3bc1b82bb380532c5e

moskirathe:
@Turbo87 @carols10cents still looking for a review on this one when you have some time :)

Turbo87:
don't worry about the conflicts, I'll resolve them :)

crates/crates_io_database/src/models/krate.rs
```diff
@@ -243,7 +244,47 @@ impl Crate {
             WHERE crate_owners.crate_id = crate_owners_with_login.crate_id
                 AND crate_owners.owner_id = crate_owners_with_login.owner_id
                 AND crate_owners.owner_kind = crate_owners_with_login.owner_kind
-                AND lower(crate_owners_with_login.login) = lower($2);"#,
+                AND canon_username(crate_owners_with_login.login) = canon_username($2);"#,
```
carols10cents, line 247: I just noticed that the change to this query, because `crate_owners_with_login` is selecting both `teams.login` and `users.username` as `crate_owners_with_login.login`, makes it so that we're comparing _team names_ using the `canon_username` function (which also normalizes hyphen to underscore), whereas before we were only comparing team names using `lower`.

I don't think we want to change anything about how we handle team names right now. That's under GitHub's control (and we don't have plans to change the way teams work yet).

Looking at the production database, I don't see any team names that differ only by hyphen/underscore today, but I do see some team names with hyphens and some with underscores, so GitHub does allow it. While I don't think it would be a good idea for anyone to do this, I just confirmed that it's possible to create two teams in GitHub, one named `crates-io` and one named `crates_io`. I think the change to this query would mean you couldn't manage them independently. If they both existed in crates.io, then you ran `cargo owner --add github:rust-lang:crates-io`, both the `crates-io` and the `crates_io` team would get added, if my reasoning is correct.

Given that we now have this query in `owner_remove_with_username` and a slightly different query in `owner_remove_with_gh_login`, maybe we should split `owner_remove_with_username` into `owner_remove_with_username` and `owner_remove_with_team_name`?

crates/crates_io_database/src/models/user.rs
```diff
@@ -216,6 +220,20 @@ pub struct OauthGithub {
...
 }
 
+impl OauthGithub {
+    pub async fn find_by_login(
+        mut conn: &AsyncPgConnection,
+        login: &str,
+    ) -> QueryResult<OauthGithub> {
+        oauth_github::table
+            .filter(canon_username(oauth_github::login).eq(canon_username(login)))
+            .filter(oauth_github::account_id.ne(-1))
```
carols10cents, line 230: The `.filter(oauth_github::account_id.ne(-1))` isn't necessary-- users that have `gh_id` set to -1 in the `users` table just don't have any records in the `oauth_github` table. The `-1` values were always sort of a hack and I'm looking forward to getting rid of it with this transition, not continue propagating it :)

crates/crates_io_database/src/models/user.rs
```diff
@@ -245,3 +263,53 @@ impl NewOauthGithub<'_> {
...
+    use super::*;
+    use crates_io_test_db::TestDatabase;
+
+    async fn insert_user(
+        conn: &AsyncPgConnection,
+        username: &str,
+        gh_login: &str,
+        gh_id: i32,
+    ) -> QueryResult<i32> {
+        let user_id = NewUser::builder()
```
carols10cents, line 278: Are you able to use [`UserBuilder`](https://github.com/rust-lang/crates.io/blob/a9da2aee0fd04332b9756f2f2382f413292c8dd6/crates/crates_io_test_utils/src/builders/user.rs#L23) here and...

crates/crates_io_database/src/models/user.rs
```diff
@@ -245,3 +263,53 @@ impl NewOauthGithub<'_> {
...
+            .gh_id(gh_id)
+            .gh_login(gh_login)
+            .username(username)
+            .build()
+            .insert(conn)
+            .await?;
+
+        NewOauthGithub::builder()
```
carols10cents, line 286: ... [`OauthGithubBuilder::for_user`](https://github.com/rust-lang/crates.io/blob/a9da2aee0fd04332b9756f2f2382f413292c8dd6/crates/crates_io_test_utils/src/builders/user.rs#L103) here?

These test builders might not have existed when you started this PR, but they should make other changes smaller, like the `User.gh_login` change you mentioned we should do in another comment :)

migrations/2026-07-18-120001-0000_add_oauth_github_login_index/up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX CONCURRENTLY IF NOT EXISTS index_oauth_github_login ON oauth_github (lower(login));
```
carols10cents, line 1: Based on a resolved comment thread on this PR that I'm having trouble linking to and the current state of the query in `owner_remove_with_gh_login` that has `AND canon_username(crate_owners_with_gh_login.login) = canon_username($2)`, I expected this index to be `ON oauth_github (canon_username(login))` rather than `lower`? Maybe the change got lost in a rebase?

packages/crates-io-api-client/schema.ts
```diff
@@ -3293,9 +3293,14 @@ export interface operations {
                      *
                      *     For users, use just the username (e.g., `"octocat"`).
```
carols10cents, line 3294: Tiny nit that I think will make this documentation clearer: The note this PR adds about disambiguation of usernames should be added up here with users rather than after the note about teams, and the "just" in "use _just_ the username" should be removed because that's not necessarily true now.

packages/crates-io-api-client/schema.ts
```diff
@@ -3358,9 +3363,14 @@ export interface operations {
...
```
carols10cents, line 3364: Same here with moving the disambiguation note up here and removing "just"

src/controllers/krate/owners.rs
```diff
@@ -192,7 +195,10 @@ pub struct ChangeOwnersRequest {
     ///
     /// For users, use just the username (e.g., `"octocat"`).
```
carols10cents: Same here with moving the disambiguation note up here and removing "just"

src/controllers/krate/owners.rs
```diff
@@ -257,21 +263,36 @@ async fn modify_owners(
...
                 for login in &logins {
-                    let login_test =
-                        |owner: &Owner| owner.login().to_lowercase() == *login.to_lowercase();
+                    let parsed_login = Login::parse(login)?;
+                    let owner = resolve_unprefixed_login(conn, &parsed_login).await?;
+
+                    let login_test = |owner: &Owner| -> bool {
+                        match parsed_login {
+                            Login::GitHubTeam(_) => {
+                                canon_username(owner.username()) == canon_username(login)
```
carols10cents: Similarly to my comment on `owner_remove_with_username`, I think this is changing the way we handle team names in a way we don't want to change in this PR. Team names should still only use `lower`.

src/controllers/krate/owners.rs
```diff
@@ -344,41 +366,191 @@ async fn modify_owners(
...
+    Ok(Some(user))
+}
+
+async fn find_user_by_username(conn: &mut AsyncPgConnection, username: &str) -> QueryResult<User> {
+    users_by_username(username)
+        .left_join(oauth_github::table)
+        .select(User::as_select())
+        .first(conn)
+        .await
+}
```
carols10cents, line 419: Note for myself: this is where i stopped reviewing and saved the review

carols10cents (changes requested):
I haven't quite been through the whole thing but I have to run - there are a few things that need to change and I will look at the rest sometime in the next 24 hours :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14213/89d60cada6219056609483da3b7df2f415772a24..270d425551733b19b7a79208eb8fb31eebc4064f)*

migrations/2026-07-18-120001-0000_add_oauth_github_login_index/up.sql
Turbo87: sorry, I wasn't done with refactorings in this branch yet 🙈 

I think we should actually do the opposite and keep GitHub account comparisons as only lower(), since hyphen and underscore are not equivalent on the GitHub side, even if we plan on treating them that way for the crates.io usernames.

carols10cents: oh whoops! sorry-- i'll hold off with the rest of my review til you say you're done :)

sounds good with the github normalization!

crates/crates_io_database/src/models/krate.rs
moskirathe: yes, thanks for catching this! i did a partial extraction from this pr and did the split in https://github.com/rust-lang/crates.io/pull/14596. I'll rebase when https://github.com/rust-lang/crates.io/pull/14596 is merged.

PR #14228, opened 2026-07-14
fastly: Add surrogate-key purging

Turbo87:
Path purges cannot invalidate every object associated with a crate without enumerating its URLs. Fastly surrogate keys provide a service-scoped grouped invalidation primitive.

This adds a `purge_surrogate_key()` fn to implement such purges. The the `purge` example binary is extended to support these too via `--service-id <ID> <KEY>`.

### DIFF
--- a/crates/crates_io_fastly/src/lib.rs
+++ b/crates/crates_io_fastly/src/lib.rs
@@ -73,6 +73,18 @@ impl Fastly {
         Ok(())
     }
 
+    /// Invalidates all objects associated with a surrogate key on a Fastly service.
+    ///
+    /// The key uses Fastly's native representation without CloudFront's leading `#` marker.
+    ///
+    /// More information on Fastly's APIs for cache invalidations can be found here:
+    /// <https://developer.fastly.com/reference/api/purging/>
+    #[instrument(skip(self))]
+    pub async fn purge_surrogate_key(&self, service_id: &str, key: &str) -> Result<(), Error> {
+        let url = format!("{}/service/{service_id}/purge/{key}", self.api_base_url);
+        self.send_purge_request(url).await
+    }
+
     /// Invalidates a path on Fastly
     ///
     /// This method takes a domain and path and invalidates the cached content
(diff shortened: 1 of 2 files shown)


crates/crates_io_fastly/src/lib.rs
```diff
@@ -73,6 +73,18 @@ impl Fastly {
...
 
+    /// Invalidates all objects associated with a surrogate key on a Fastly service.
+    ///
+    /// The key uses Fastly's native representation without CloudFront's leading `#` marker.
+    ///
+    /// More information on Fastly's APIs for cache invalidations can be found here:
+    /// <https://developer.fastly.com/reference/api/purging/>
+    #[instrument(skip(self))]
+    pub async fn purge_surrogate_key(&self, service_id: &str, key: &str) -> Result<(), Error> {
+        let url = format!("{}/service/{service_id}/purge/{key}", self.api_base_url);
```
Turbo87, line 84: FYI I've tested it manually: the `key` does not need URL encoding for `+` characters in version numbers.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14228/c12b91a0580dadeb0368a7c93f2134ac3b79572b..a085f4d24b1884ca48182d0cb1bc43d29ee8a487)*

LawnGnome: I guess the question I have is: do we need the surrogate keys to be human-readable? We could hash and base 16/32 encode the name and version and then sidestep any issues of encoding for non-alphanumerics ever, at the cost of not being able to as quickly construct a surrogate key from scratch in a support or incident situation.

LawnGnome (approved):
I'm assuming this (and #14229) will have follow ups to actually set these, so LGTM.

One other thought: it feels like the example purge binary has logic that we might want to pair with the Cloudfront implementation and put into a crates-admin command at some point.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14228/c12b91a0580dadeb0368a7c93f2134ac3b79572b..a085f4d24b1884ca48182d0cb1bc43d29ee8a487)*

crates/crates_io_fastly/src/lib.rs
Turbo87: yeah, I guess that would've been an option. having them be somewhat readable is also kinda nice though 😅

since the cache tags have already been applied to the S3 objects it's probably easiest to keep it that way unless we run into any major issues with it in the future.

PR #14229, opened 2026-07-14
cloudfront: Support cache-tag invalidations

Turbo87:
CloudFront represents cache-tag invalidations as entries beginning with `#`. Prefixing those entries with `/` changes them into path invalidations and prevents the requested tag purge.

This PR changes the implementation to preserve both `/` paths and `#` tags during normalization, and cover
mixed invalidation requests.

### DIFF
--- a/src/cloudfront.rs
+++ b/src/cloudfront.rs
@@ -42,10 +42,13 @@ pub struct CloudFront {
     static_distribution_id: String,
 }
 
-/// Ensures CloudFront invalidation paths start with a slash.
+/// Normalizes CloudFront invalidation entries.
+///
+/// Path invalidations must start with `/`. Cache-tag invalidations already start with `#`
+/// and must remain unchanged so CloudFront recognizes them as cache tags.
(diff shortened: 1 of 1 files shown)


src/cloudfront.rs
```diff
@@ -42,6 +42,15 @@ pub struct CloudFront {
     static_distribution_id: String,
 }
 
+/// Ensures CloudFront URL invalidation paths start with a slash while preserving cache tags.
```
LawnGnome: I think we might want to reproduce more of the "why" that's in the PR description here as well for future us.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14229/c12b91a0580dadeb0368a7c93f2134ac3b79572b..e64d732c91f229c99f2dec0d574e5d56e481548a)*

LawnGnome (approved):

src/cloudfront.rs
Turbo87: I've adjusted the comment 👍

PR #14238, opened 2026-07-15
Use `UserBuilder` in more tests

carols10cents:
As we remove `gh_*` fields from user records, using the `UserBuilder` in more places now means fewer places will need to be updated to remove the `gh_*` fields when we deprecate those since the `UserBuilder` consolidates setting the `gh_*` fields on the test user records to one place.

There are still a few places that I've deliberately *not* used `UserBuilder`, such as:

- sync admins test; we only want to support setting `is_admin` in the sync admins job, not the regular user flow
- tests deliberately exercising what happens when `users.gh_id` is -1, which is still a thing in production

### DIFF
--- a/crates/crates_io_database/Cargo.toml
+++ b/crates/crates_io_database/Cargo.toml
@@ -28,6 +28,7 @@ utoipa = { version = "=5.5.0", features = ["chrono"] }
 [dev-dependencies]
 claims = "=0.8.0"
 crates_io_test_db = { path = "../crates_io_test_db" }
+crates_io_test_utils = { path = "../crates_io_test_utils" }
 googletest = "=0.14.3"
 insta = { version = "=1.48.0", features = ["filters", "json"] }
 tokio = { version = "=1.52.3", features = ["macros", "rt"] }
--- a/crates/crates_io_database/Cargo.toml
+++ b/crates/crates_io_database/Cargo.toml
@@ -28,7 +28,6 @@ utoipa = { version = "=5.5.0", features = ["chrono"] }
 [dev-dependencies]
 claims = "=0.8.0"
 crates_io_test_db = { path = "../crates_io_test_db" }
-crates_io_test_utils = { path = "../crates_io_test_utils" }
 googletest = "=0.14.3"
 insta = { version = "=1.48.0", features = ["filters", "json"] }
 tokio = { version = "=1.52.3", features = ["macros", "rt"] }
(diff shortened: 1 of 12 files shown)


crates/crates_io_database/Cargo.toml
```diff
@@ -28,6 +28,7 @@ utoipa = { version = "=5.5.0", features = ["chrono"] }
 [dev-dependencies]
 claims = "=0.8.0"
 crates_io_test_db = { path = "../crates_io_test_db" }
+crates_io_test_utils = { path = "../crates_io_test_utils" }
```
Turbo87: hmm... I'm feeling a bit uneasy about adding this to the database crate since it pulls in:

```toml
crates_io_api_types = { path = "../crates_io_api_types" }
crates_io_cargo_toml = { path = "../crates_io_cargo_toml" }
crates_io_database = { path = "../crates_io_database" }
crates_io_encryption = { path = "../crates_io_encryption" }
crates_io_linecount = { path = "../crates_io_linecount" }
crates_io_tarball = { path = "../crates_io_tarball", features = ["builder"] }
```

that creates quite a spaghetti in the dependency tree 😅 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14238/fdac0bf7f4d65b98aefa233b08c4567a099482e4..69f10da346df080637f7c9de8bf551e5c97d21f7)*

carols10cents: ok 🤷🏻‍♀️  i put it back the way it was!

crates/crates_io_test_utils/src/builders/user.rs
```diff
@@ -22,6 +22,7 @@ static ENCRYPTED_TOKEN: LazyLock<Vec<u8>> = LazyLock::new(|| {
 /// If you want to test logic that happens as part of signing up or logging in,
 pub struct UserBuilder<'a> {
```
LawnGnome, line 23: Wondering at what point we would want to pull `bon` in, but probably not at two fields.

crates/crates_io_test_utils/src/builders/user.rs
```diff
@@ -30,18 +31,26 @@ impl<'a> UserBuilder<'a> {
...
+            ..self
+        }
     }
 
     pub fn build(self) -> User {
         User {
             id: 1,
             gh_login: self.username.into(),
-            name: Some("The Octocat".into()),
+            name: self.display_name.map(ToString::to_string),
```
LawnGnome, line 53: This is technically a change in behaviour, but since the CI is green I'm assuming we weren't relying on it anywhere.

LawnGnome (approved):
Makes sense to me!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14238/fdac0bf7f4d65b98aefa233b08c4567a099482e4..50211803bd8a5b3990d37c8c369f51597ec47f1e)*

PR #14246, opened 2026-07-15
Consistently order rev dep version info

carols10cents:
Otherwise, postgres doesn't guarantee any order and the test asserting this snapshot can be flaky.

I have a branch locally that changes the schema and somehow changes the undefined order that this snapshot's data gets returned 🤷🏻‍♀️ it fails consistently without this change and passes consistently with the change.

The versions get associated with the dependencies in the UI based on version ID, so changing the order doesn't change any user-visible behavior.


Turbo87 (approved):
thanks! :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14246/242c0db1450ced71de66812ab5701acdcc32707c..a05b78e4bd36a868cb7e4be7aaad2032e829257d)*

Issue #14245, opened 2026-07-15
Republished crates still uses previous README


OctopusET:
### Current Behavior

I removed the this crates and re-published https://crates.io/crates/azo
But it still uses previous README

### Expected Behavior

New README should be shown

### Environment

- Browser:
- OS:

Turbo87:
this sounds related to https://github.com/rust-lang/crates.io/issues/14219.

I've opened https://github.com/rust-lang/crates.io/pull/14228 and https://github.com/rust-lang/crates.io/pull/14229 as required changes before we can properly fix the cache invalidations, but those PRs are still in review phase.

Turbo87:
FYI I've purge the cache keys for the crate mentioned above, so hopefully the issue should be resolved. We are still investigating a by-URL invalidation issue with Fastly though, which could influence cache invalidation for things like the RSS feeds or OpenGraph images.

Issue #14219, opened 2026-07-13
Checksum of `onslaught_counters-0.0.1.crate` doesn't match.


malaire:
### Current Behavior

According to `https://index.crates.io/on/sl/onslaught_counters` checksum of 0.0.1 version is `81cb786475db5df71539dec140f309c4f119bf93c0c92b1909e47eefe23f2e55` but when I download the file from `https://crates.io/api/v1/crates/onslaught_counters/0.0.1/download` the checksum of the downloaded file is `13650473d4b53177322632756bd5e08f6994604e711b828fe69761eb3b5b23a7`.

### Expected Behavior

Checksums should match.

### Steps To Reproduce

1. Download `https://index.crates.io/on/sl/onslaught_counters`
2. Download `https://crates.io/api/v1/crates/onslaught_counters/0.0.1/download`
3. Compare crate checksum to the checksum mentioned in index.

Alternatively run following commands:
```
cargo new testing
cd testing
cargo add onslaught_counters@0.0.1
cargo build
```
This results in error `error: failed to verify the checksum of onslaught_counters v0.0.1`.

Turbo87:
I can’t reproduce this currently. The downloaded crate hashes to the index checksum, `81cb7864…e55`.

A plain curl request currently receives a 403 JSON response because curl’s default user agent is blocked. Without `--fail`, curl may save that response as the crate. Its checksum changes because the response contains a unique request ID. Could you retry with `curl -fL -A 'checksum-check' <URL>` and confirm the result?

malaire:
I can confirm that `curl -fL -A 'checksum-check' https://crates.io/api/v1/crates/onslaught_counters/0.0.1/download | sha256sum` also returns `13650473d4b53177322632756bd5e08f6994604e711b828fe69761eb3b5b23a7`.

I also get same result if I download via Firefox browser or via `reqwest` crate, and as I mentioned `cargo` also gives checksum error.

I'm located in Finland, in case that matters.

malaire:
Should I attach the downloaded `onslaught_counters-0.0.1.crate` here? It's 32147 bytes and a valid archive (`tar xzf *.crate` will uncompress it without errors).

Turbo87:
> Should I attach the downloaded `onslaught_counters-0.0.1.crate` here? It's 32147 bytes.

yeah, if GitHub lets you attach it then that might help debug this

malaire:
[...]

Issue #14062, opened 2026-06-25
CDN invalidation paths use literal `+` instead of `%2B` for build-metadata versions


Turbo87:
The S3 object key for a build-metadata version contains a literal `+` byte (e.g. `crates/foo/foo-1.0.0+foo.crate`). Over HTTP it must be accessed as `%2B`, since a literal `+` in a path is ambiguous (see #4891). So everywhere we hand a path to an external HTTP system we normalize `+` → `%2B`: client redirect URLs via `Storage::crate_location()` and friends (`src/storage.rs:212-251`), and both CDNs rewrite literal `+` → `%2B` at the viewer stage, meaning cached objects are keyed under `%2B`.

CDN invalidation is the exception. `InvalidateCdns` receives `object_store::Path` values and calls `path.to_string()` (`src/worker/jobs/invalidate_cdns.rs`). `object_store` does not percent-encode `+` (its `INVALID` set omits it), so the paths keep the literal `+` and get sent that way to Fastly purge (`crates/crates_io_fastly/src/lib.rs:79`) and the CloudFront invalidation queue (`src/worker/jobs/invalidate_cdns.rs:54-65`). Since the cached objects are keyed under `%2B`, these invalidations don't match.

Turbo87:
this was essentially resolved by https://github.com/rust-lang/crates.io/pull/14250

PR #14250, opened 2026-07-16
Add feature-flagged cache-tag invalidation

Turbo87:
Cache-tag invalidation is selected when CDN invalidation jobs are enqueued through the new `CACHE_TAG_INVALIDATIONS_ENABLED` flag. Crate and version deletions enqueue cache tags when enabled and retain the existing path payloads when disabled.

The `InvalidateCdns` background job provides explicit constructors for path-only and cache-tag-only jobs while continuing to process mixed payloads in path-first order. Fastly purges tags through the service ID (`FASTLY_SERVICE_ID_STATIC`), while CloudFront uses the existing `#` marker for tagged invalidations.

### Related

- https://github.com/rust-lang/crates.io/issues/14219
- https://github.com/rust-lang/crates.io/issues/14245
- #14229 
- #14228
- https://github.com/rust-lang/crates.io/issues/14062

### DIFF
--- a/src/bin/crates-admin/delete_version.rs
+++ b/src/bin/crates-admin/delete_version.rs
@@ -2,9 +2,9 @@ use crate::dialoguer;
 use anyhow::Context;
 use crates_io::models::update_default_version;
 use crates_io::schema::crates;
-use crates_io::storage::{Storage, StorageKey};
+use crates_io::storage::{Storage, StorageKey, release_cache_tag};
 use crates_io::worker::jobs;
-use crates_io::{db, schema::versions};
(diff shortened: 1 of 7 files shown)


Turbo87:
merging this so that we can test it out on staging. it is feature flagged anyway, so the production impact should be minimal.

OctopusET:
Thanks! Let me check if my issue is fixed (#14245)

Turbo87:
it is not yet deployed to our production environment 😉

OctopusET:
Ah got it I thought there might be a way to use staging server. Anyways time will fix this issue for me. Thank you again!

PR #14258, opened 2026-07-16
Stop reading from users.gh_encrypted_token

carols10cents:
Continue writing to it, but always select the value from oauth_github instead.

This is an incremental step towards fully removing the `users.gh_encrypted_token` column. This keeps all writes but moves all reads to `oauth_github.encrypted_token`, so that we can easily revert.

I've opened a PR to show the future commits that I won't submit until there's only one commit left: https://github.com/rust-lang/crates.io/pull/14259

### DIFF
--- a/crates/crates_io_database/src/models/user.rs
+++ b/crates/crates_io_database/src/models/user.rs
@@ -24,8 +24,9 @@ pub struct User {
     pub gh_login: String,
     #[diesel(select_expression = oauth_github::avatar.nullable())]
     pub gh_avatar: Option<String>,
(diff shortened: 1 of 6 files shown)


Turbo87 (approved):
seems reasonable. thanks :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14258/785b3c2c9c160e695e0dad857ccd8f07421c75a6..79141a128c12d2cbec1a6b9443133fb5f2ed5558)*

moskirathe (approved):

PR #14261, opened 2026-07-16
Fix README image distortion by preserving aspect ratio

### LINKED ISSUE #14133: Add `height: auto` to README images
### Current Behavior

Images in READMEs currently get `max-width: 100%`. For wider-than-container images, this squishes images together, distorting their aspect ratio.

Image
  
The common fix to this is adding `height: auto` to the image CSS. I've confirmed that this works for crates.io as well by adding that one line in the Dev Tools CSS inspector.

Small fix, but makes the READMEs look better and aligns with how GitHub and others does it:

Image

Example of squished image: https://crates.io/crates/rerun

### Environment

- Browser: Chrome 147
- OS: MacOS 26.3.1

fabianbernhart:
Adds `height: auto` to README images to preserve their aspect ratio when constrained by max-width: 100%. This prevents wider-than-container images from being vertically squished and improves README rendering for images with larger dimensions.

Fixes #14133

### DIFF
--- a/svelte/src/lib/components/CrateVersionPage.svelte
+++ b/svelte/src/lib/components/CrateVersionPage.svelte
@@ -213,11 +213,6 @@
     border-radius: var(--space-3xs);
     box-shadow: var(--shadow);
 
-    /* see: https://github.com/rust-lang/crates.io/issues/14133 */
-    :global(img) {
-      height: auto;
-    }
-
     @media only screen and (max-width: 550px) {
       margin-left: calc(var(--main-layout-padding) * -1);
       margin-right: calc(var(--main-layout-padding) * -1);
--- a/svelte/src/lib/components/CrateVersionPage.svelte
+++ b/svelte/src/lib/components/CrateVersionPage.svelte
@@ -213,6 +213,11 @@
     border-radius: var(--space-3xs);
     box-shadow: var(--shadow);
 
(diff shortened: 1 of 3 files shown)


svelte/src/lib/components/CrateVersionPage.svelte
```diff
@@ -213,6 +213,11 @@
     border-radius: var(--space-3xs);
     box-shadow: var(--shadow);
 
+    /* see: https://github.com/rust-lang/crates.io/issues/14133 */
+    :global(img) {
+      height: auto; 
+    }
```
Turbo87: this should probably be on the `TextContent` component, like all the other global CSS overrides.

Turbo87 (changes requested):
could you add an example to the Playwright test that renders a dummy readme? that way we can verify it via visual regression testing (Percy)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14261/e9c29a333bcb7c14e82adb59309e0220a771648a..3d4e78d766295852d52d4945d96147a63d1397ae)*

Turbo87 (approved):
perfect, thank you! :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14261/e9c29a333bcb7c14e82adb59309e0220a771648a..ccc12edd78ca58c10a82fc9c31080a0232846279)*

PR #14293, opened 2026-07-20
svelte: Navigate to index when there is no previous page

kidonng:
Title. Currently the "Go Back" button does nothing when e.g. navigating directly to a not found page like https://crates.io/404. One could always use the top left header link but this is mo intuitive.

### DIFF
--- a/e2e/acceptance/404.spec.ts
+++ b/e2e/acceptance/404.spec.ts
@@ -13,11 +13,15 @@ test.describe('Acceptance | 404', { tag: '@acceptance' }, () => {
   });
 
   test('go back navigates to index when there is no previous page', async ({ page }) => {
-    // Can't use page.goto which adds a history entry
-    await page.evaluate(() => location.replace('/unknown-route'));
+    await page.goto('/unknown-route');
     await expect(page.locator('[data-test-go-back]')).toBeVisible();
     await page.click('[data-test-go-back]');
-    await expect(page).toHaveURL('/');
+    // Svelte doesn't update URL during tests for some reason, so the following assertion only works in actual browsers
+    // await expect(page).toHaveURL('/');
+    // Instead, assert we are no longer on 404 page...
+    await expect(page.locator('[data-test-404-page]')).toBeHidden();
+    // ...but on index page instead
+    await expect(page).toHaveTitle('crates.io: Rust Package Registry');
   });
 
   test('go back navigates to previous page when history exists', async ({ page }) => {
--- a/e2e/acceptance/404.spec.ts
+++ b/e2e/acceptance/404.spec.ts
@@ -13,15 +13,16 @@ test.describe('Acceptance | 404', { tag: '@acceptance' }, () => {
   });
 
   test('go back navigates to index when there is no previous page', async ({ page }) => {
-    await page.goto('/unknown-route');
-    await expect(page.locator('[data-test-go-back]')).toBeVisible();
-    await page.click('[data-test-go-back]');
-    // Svelte doesn't update URL during tests for some reason, so the following assertion only works in actual browsers
-    // await expect(page).toHaveURL('/');
-    // Instead, assert we are no longer on 404 page...
-    await expect(page.locator('[data-test-404-page]')).toBeHidden();
-    // ...but on index page instead
-    await expect(page).toHaveTitle('crates.io: Rust Package Registry');
+    // Pre-warm index page
+    await page.goto('/');
+
+    // Open popup to get a fresh history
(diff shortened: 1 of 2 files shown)


Turbo87:
interesting, thanks!

could you fix the lint issue and add playwright tests for this? 🙏

e2e/acceptance/404.spec.ts
```diff
@@ -11,4 +11,24 @@ test.describe('Acceptance | 404', { tag: '@acceptance' }, () => {
...
+  test('go back navigates to index when there is no previous page', async ({ page }) => {
+    await page.goto('/unknown-route');
+    await expect(page.locator('[data-test-go-back]')).toBeVisible();
+    await page.click('[data-test-go-back]');
+    // Svelte doesn't update URL during tests for some reason, so the following assertion only works in actual browsers
+    // await expect(page).toHaveURL('/');
+    // Instead, assert we are no longer on 404 page...
+    await expect(page.locator('[data-test-404-page]')).toBeHidden();
+    // ...but on index page instead
+    await expect(page).toHaveTitle('crates.io: Rust Package Registry');
```
kidonng: Sorry! Tests took longer than I expected, due to this quirk I documented here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14293/ee5873ed182f3b53cb906c751240690223e5a56c..4a493c5d495a4de120069a6b693f6d544b774436)*

svelte/src/routes/+error.svelte
```diff
@@ -14,6 +16,12 @@
   }
 
   function goBack() {
+    // window.navigation is only available in modern browsers
+    let canGoBack = window.navigation?.canGoBack ?? history.length > 1;
```
Turbo87: `navigation.canGoBack` is not equivalent to “there is a previous browser history entry.” The Navigation API only includes contiguous same-origin entries, so this is false when someone follows a link from another site to a missing crates.io URL. In that case the button navigates to the crates.io index instead of returning to the previous page. Since the fallback is intended for a new tab/direct visit with no preceding entry, I think we should use `history.length <= 1` here without checking `navigation.canGoBack`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14293/ee5873ed182f3b53cb906c751240690223e5a56c..2d6dc50b092b1f09d8bebc16d4c4493a00d796e2)*

e2e/acceptance/404.spec.ts
```diff
@@ -11,4 +11,24 @@ test.describe('Acceptance | 404', { tag: '@acceptance' }, () => {
     await percy.snapshot();
     await expect(page).toMatchAriaSnapshot({ name: 'aria.yml' });
   });
+
+  test('go back navigates to index when there is no previous page', async ({ page }) => {
+    await page.goto('/unknown-route');
```
Turbo87: This setup does not seem to actually create the no-history condition named by the test. Playwright starts the page at `about:blank`, and `page.goto('/unknown-route')` leaves `history.length === 2`. The test passes in Chromium because `navigation.canGoBack` excludes that earlier entry, so it does not protect the `history.length` fallback used by browsers without the Navigation API.

Could we create a page with `history.length === 1`, assert that precondition, click Go Back, and assert `toHaveURL('/')`?

We could for example open `/unknown-route` in a popup. In local testing I noticed that Svelte/Vite HMR can cause issues with cold starts though. What worked reliably is first loading `/` and then opening the popup.

I think the comments saying Svelte does not update the URL can then be removed because the URL assertion passes when the intended branch is actually exercised.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14293/ee5873ed182f3b53cb906c751240690223e5a56c..2d6dc50b092b1f09d8bebc16d4c4493a00d796e2)*

svelte/src/routes/+error.svelte
kidonng: This could go both ways but I will respect your choice here!

e2e/acceptance/404.spec.ts
kidonng: Right, this currently relies on `navigation.canGoBack`. I tried but can't find a way to let Playwright start on a page instead of `about:blank`. The popup idea seems interesting! I will try it

Turbo87 (approved):
perfect, thank you! :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14293/ee5873ed182f3b53cb906c751240690223e5a56c..dc11a6619368be700ff8dd5696a09851ca0edd47)*

Issue #13907, opened 2026-06-10
Separate session controller authorize into "signup" and "login"


carols10cents:
Will make these easier:

- https://github.com/rust-lang/crates.io/issues/13771
- https://github.com/rust-lang/crates.io/issues/13768
- https://github.com/rust-lang/crates.io/issues/13773

[In `authorize_session`](https://github.com/rust-lang/crates.io/blob/c61947f57302634e1c83e0b33091ec9af9003edc/src/controllers/session.rs#L87), it's either creating a new user record or updating an existing record and then logging in that user. This makes the database queries complicated (as in https://github.com/rust-lang/crates.io/pull/13856), and actually we want the behavior to be different based on whether the user is existing or new:

- For existing users that have the relevant records in the database, update any info that changed that should be synced with the oauth info and log them in (as the code does currently).
- For new users, display a page with a form to collect or confirm any additional information. Carry the oauth info through the requests, and don't create anything in the database until they've submitted this form. Then once everything's been created, log them in.
  - For this issue, only show a field for email address, which we currently just use if we happen to get it from github and then they have to go edit profile later if they didn't share their email address or wanted to use a different email address. eventually this will also confirm username, avatar, and display name as in the issues linked above

PR #14310, opened 2026-07-21
WIP separate signup and login 

carols10cents:
Connects to #13907. Not ready for review yet.

### DIFF
--- a/crates/crates_io_github/src/lib.rs
+++ b/crates/crates_io_github/src/lib.rs
@@ -418,7 +418,7 @@ impl From<reqwest::Error> for GitHubError {
     }
 }
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Serialize)]
 pub struct GitHubUser {
     pub avatar_url: Option<String>,
     pub email: Option<String>,
--- a/src/controllers/session.rs
+++ b/src/controllers/session.rs
@@ -7,7 +7,7 @@ use crate::schema::{oauth_github, users};
 use crate::util::diesel::is_read_only_error;
 use crate::util::errors::{AppResult, bad_request, server_error};
 use crate::util::oauth::ReqwestClient;
-use crate::views::EncodableMe;
+use crate::views::{EncodableMe, EncodablePrivateUser};
 use axum::Json;
 use chrono::Utc;
 use crates_io_github::{GitHubAuth, GitHubUser};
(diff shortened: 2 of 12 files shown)


carols10cents:
@Turbo87 and anyone else who feels like looking at this, i'm not super happy with this and i'm wondering if i'm going about everything all wrong. i thought of a bunch of different ways to do this and didn't like any of them, which usually means i'm missing something easy 😅 

and of course i'm flailing at svelte hard, after reading (checks notes) exactly 0 pages of svelte documentation 🤣 but i don't really want to spend time hitting my head against the frontend if i'm going to have to redo the backend anyway... so... 

how bad is it?

Turbo87:
👋  I took a closer look at this. I think the overall direction is correct. Separating existing-user login from new-user signup is the right boundary, especially once signup needs user-confirmed fields.

The main concern is that the current approach represents someone who has completed GitHub OAuth, but has not created a crates.io account, as an authenticated user with ID `-1`. That makes `/api/v1/me` handle pending registrations and risks masking unrelated authentication errors. The new signup path also calls the account-creation inserts without their previous transaction, so a failure can leave a partially created account. Finally, the pending cookie values are removed before creation succeeds, which prevents a safe retry in case the insertion fails for some reason (e.g. read-only mode).

I think a cleaner design would have `/session/authorize` return either `signed_in` with a real user or `signup_required` with no user object. For the latter, the backend would store one short-lived, structured pending-signup value in the signed cookie. A dedicated `GET`, `POST`, and `DELETE /api/private/session/signup` resource would load the form data, create the account transactionally, or cancel the attempt. The frontend would remain logged out and navigate to a dedicated signup page. That allows `/api/v1/me` to keep its existing meaning.

I have mostly stayed out of the account untangling work because I did not want to step on anyone’s toes. If this alternative sounds useful, I would be happy to implement it.

carols10cents:
> I have mostly stayed out of the account untangling work because I did not want to step on anyone’s toes. If this alternative sounds useful, I would be happy to implement it.

And I didn't want to add to your workload! 😅 I would be grateful if you would take this one over... what you've written sounds fine, parts like

> The frontend would remain logged out and navigate to a dedicated signup page.

i'm not at all sure how to implement, and it would probably be more effort for us to go back and forth on me messing it up, you explaining, etc than if you would just do it. 

But I do think this is blocking things like letting the user choose their username on signup, do you think you'll have time for this in the next week or two? If not, I'll give it a try. Closing this!

Turbo87:
depends on when the egg will hatch, but I'll try to get started at least 😄

PR #14349, opened 2026-07-25
WIP: add unique index for usernames

### LINKED ISSUE #13762: Add unique index to `users.username`

moskirathe:
fixes: https://github.com/rust-lang/crates.io/issues/13762

adds a unique username index for the `username` column in the `users` table.


Turbo87:
not sure why you reopened the PR, but this was already done in https://github.com/rust-lang/crates.io/commit/0d2adff3fd9d4044790f92fb8b5c769d8c5e1055 :)

Turbo87:
oh, wait, this is about uniqueness... sorry 🤦‍♂️

PR #14353, opened 2026-07-26
jobs/invalidate_cdns: Percent-encode `+` in invalidation paths

### LINKED ISSUE #14062: CDN invalidation paths use literal `+` instead of `%2B` for build-metadata versions

Booyaka101:
Fixes #14062.

`InvalidateCdns::paths()` receives `object_store::Path` values and renders them with `to_string()`. `object_store` omits `+` from its `INVALID` set, so the literal byte survives.

Everywhere else a path is handed to an external HTTP system it goes through `StorageKey::cdn_path()`, which encodes `+` as `%2B`, and both CDNs rewrite literal `+` to `%2B` at the viewer stage. Cached objects for build-metadata versions (e.g. `crates/foo/foo-1.0.0+foo.crate`) are therefore keyed under `%2B`.

CDN invalidation was the exception. The purge request targeted a path that was never cached, so it succeeded while invalidating nothing. This affects both consumers — Fastly via `purge_both_domains()` and the CloudFront invalidation queue.

Crate files are stored with `CACHE_CONTROL_IMMUTABLE` (`max-age=31536000`), so for a build-metadata version the practical consequence is that the `.crate` file stays served from the CDN for up to a year after the version was deleted, rather than until the next short TTL window.

## Changes

- Extract the `+` to `%2B` normalization out of `StorageKey::cdn_path()` into `storage::encode_cdn_path()`, so the two places that need this rendering share one implementation.
- Apply it in `InvalidateCdns::paths()`, which is the single point both callers (`DeleteCrateFromStorage` and the `delete-version` admin command) funnel through.

Cache tags are surrogate keys rather than paths, so they are deliberately left unencoded; there is a test asserting that.

## Testing

Two new unit tests in `invalidate_cdns`. `paths_percent_encode_plus` also asserts the underlying `object_store::Path` behaviour, so it documents the cause rather than just the symptom.

Verified the test actually pins the bug by reverting the fix and re-running:

```
assertion `left == right` failed
  left:  ["crates/foo/foo-1.0.0+foo.crate",   "readmes/foo/foo-1.0.0+foo.html"]
  right: ["crates/foo/foo-1.0.0%2Bfoo.crate", "readmes/foo/foo-1.0.0%2Bfoo.html"]
```

The 17 existing `storage::` tests pass unchanged, including `storage::tests::locations`, which covers `cdn_path()` and confirms the extraction is behaviour-neutral. `cargo clippy --lib --all-features` and `cargo fmt --check` are clean.

## AI tool disclosure

Per `docs/AI-TOOLS.md`: Claude was used to help trace the encoding divergence through `storage.rs`, `invalidate_cdns.rs` and `crates_io_fastly`, and to draft the tests and this description. I reviewed the change and ran the tests, including the negative control above.

### DIFF
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -27,6 +27,17 @@ const CACHE_CONTROL_INDEX: &str = "public,max-age=600";
 const CACHE_CONTROL_README: &str = "public,max-age=604800";
 const CACHE_CONTROL_OG_IMAGE: &str = "public,max-age=86400";
 
+/// Renders a storage path for use with an external HTTP system, percent-encoding
+/// `+` as `%2B`.
+///
+/// A literal `+` in a URL path is ambiguous (see #4891), so the CDNs cache and
+/// serve objects under the `%2B` form. Any path handed to a CDN — whether in a
+/// redirect URL or a cache invalidation — has to use the same form to refer to
+/// the same object.
+pub fn encode_cdn_path(path: &str) -> String {
+    path.replace('+', "%2B")
+}
+
 /// Builds the cache tag shared by every object belonging to a crate.
 pub fn crate_cache_tag(name: &str) -> String {
     format!("crate:{name}")
@@ -428,7 +439,7 @@ impl<'a> StorageKey<'a> {
     /// [`Self::path()`] rendered for a public URL, with `+` percent-encoded as
     /// `%2B`.
     pub fn cdn_path(&self) -> String {
-        self.path().as_ref().replace('+', "%2B")
+        encode_cdn_path(self.path().as_ref())
     }
 
     /// The content-type the file should be stored with, or `None` to rely on
--- a/src/worker/jobs/invalidate_cdns.rs
+++ b/src/worker/jobs/invalidate_cdns.rs
@@ -5,6 +5,7 @@ use crates_io_database::models::{CloudFrontDistribution, CloudFrontInvalidationQ
 use crates_io_worker::BackgroundJob;
 use serde::{Deserialize, Serialize};
 
+use crate::storage::encode_cdn_path;
 use crate::worker::Environment;
 use crate::worker::jobs::ProcessCloudfrontInvalidationQueue;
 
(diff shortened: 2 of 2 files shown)


Turbo87:
are you sure that https://github.com/rust-lang/crates.io/issues/14062 is still relevant now that we (mostly) switched to invalidation via cache tags?

Booyaka101:
Yeah, fair it won't fire if that flag's on. I went and checked:

`InvalidateCdns::paths()` only has two callers left, `DeleteCrateFromStorage` and the `delete-version` admin command, and both sit in the `else` branch of `features.cache_tag_invalidations_enabled`. That flag defaults to `false` and needs `CACHE_TAGS_ENABLED` too, so with both set on prod this is unreachable.

It's still the default for anyone who hasn't opted in though, and it's what you'd fall back to if the flag went off again, that's really the only reason I'd bother making the paths correct.

That said, if the URL-based branch is going away anyway then ripping it out beats fixing the escaping inside it. Happy to close this if that's the plan, your call.

Turbo87:
yeah, we have it enabled in both environments, and since it also fixed https://github.com/rust-lang/crates.io/issues/14219 we will likely keep it enabled. that specific issue is not closed yet because we are still investigating a general issue around URL-based invalidations on Fastly, which is unrelated to the `+` encoding. I think the only remaining URL-based invalidations are not version-specific, so the `+` encoding bug should not be relevant anymore.

Turbo87:
sorry for the wasted time/tokens. I'll close this one in favor of https://github.com/rust-lang/crates.io/pull/14250, but still thanks for the PR :)

Issue #14380, opened 2026-07-30
Add a `github_username_matches` field to `Encodable{Public|Private}Api`


carols10cents:
This field should be set to `true` if the user's `oauth_github.login` equals their `users.username` and false otherwise. 

This field's value will be used to decide whether to show an indication in the UI next to a username to let people know this username isn't the GitHub username as they might be expecting; the UI work is https://github.com/rust-lang/crates.io/issues/13764.

PR #14384, opened 2026-07-30
api: Extract `PublicUser` database model

Turbo87:
This introduces a `PublicUser` database model for data used by public user and owner responses. It updates the affected queries and serializers to use the narrower type while preserving existing JSON and OAuth join behavior.

The dedicated projection keeps private user fields out of public response paths and makes the data requirements of public API responses explicit. It will also make it easier in the future to introduce the planned [`github_username_matches`](https://github.com/rust-lang/crates.io/issues/14380) field.

/cc @carols10cents

### DIFF
--- a/crates/crates_io_database/src/models/user.rs
+++ b/crates/crates_io_database/src/models/user.rs
@@ -11,6 +11,33 @@ use crate::fns::lower;
 use crate::models::{Crate, CrateOwner, Email, OwnerKind};
 use crate::schema::{crate_owners, emails, oauth_github, users};
 
+/// Public data for a crates.io user.
+#[derive(Clone, Debug, HasQuery, Serialize)]
+#[diesel(
+    table_name = users,
+    base_query = users::table.left_join(oauth_github::table),
+)]
+pub struct PublicUser {
+    pub id: i32,
+    pub name: Option<String>,
+    pub gh_login: String,
+    #[diesel(select_expression = oauth_github::avatar.nullable())]
+    pub gh_avatar: Option<String>,
+    pub username: String,
+    pub created_at: Option<DateTime<Utc>>,
+}
+
+impl PublicUser {
+    pub async fn owning(krate: &Crate, mut conn: &AsyncPgConnection) -> QueryResult<Vec<Self>> {
+        CrateOwner::by_owner_kind(OwnerKind::User)
+            .inner_join(users::table.left_join(oauth_github::table))
+            .select(PublicUser::as_select())
+            .filter(crate_owners::crate_id.eq(krate.id))
+            .load(&mut conn)
+            .await
+    }
+}
+
 /// The model representing a row in the `users` database table.
 #[derive(Clone, Debug, HasQuery, Identifiable, Serialize)]
 #[diesel(
--- a/src/controllers/krate/owners.rs
+++ b/src/controllers/krate/owners.rs
@@ -3,7 +3,7 @@
 use crate::controllers::helpers::authorization::Rights;
 use crate::controllers::krate::CratePath;
 use crate::models::krate::OwnerRemoveError;
-use crate::models::{Crate, Owner, Team, User};
+use crate::models::{Crate, Owner, PublicUser, Team, User};
 use crate::models::{
     CrateOwner, NewCrateOwnerInvitation, NewCrateOwnerInvitationOutcome, NewTeam,
     krate::NewOwnerInvite, token::EndpointScope,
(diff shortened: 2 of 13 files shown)


crates/crates_io_database/src/models/team.rs
```diff
@@ -52,18 +52,14 @@ impl NewTeam<'_> {
...
-    pub async fn owning(krate: &Crate, mut conn: &AsyncPgConnection) -> QueryResult<Vec<Owner>> {
-        let base_query = CrateOwner::belonging_to(krate).filter(crate_owners::deleted.eq(false));
-        let teams = base_query
+    pub async fn owning(krate: &Crate, mut conn: &AsyncPgConnection) -> QueryResult<Vec<Team>> {
+        CrateOwner::by_owner_kind(OwnerKind::Team)
             .inner_join(teams::table)
             .select(Team::as_select())
-            .filter(crate_owners::owner_kind.eq(OwnerKind::Team))
+            .filter(crate_owners::crate_id.eq(krate.id))
+            .order(crate_owners::owner_id)
```
carols10cents: I didn't expect to see a change of the `order` of the team owner query based on the description of the PR; were tests failing? 

Also the changes to the queries to return `Team` and `User` rather than `Owner` would have been better made in a separate commit, if not a separate PR-- while that change may have been driven by the addition of `PublicUser`, it seems like it could be independent, right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14384/a313be6dfb6d67791f208fedf6c2129a3b47ccce..f43b06235d9d7e9d16ddc1ede893ba9d307065b0)*

src/controllers/krate/owners.rs
```diff
@@ -44,11 +44,14 @@ pub async fn list_owners(state: AppState, path: CratePath) -> AppResult<Json<Use
 
     let krate = path.load_crate(&conn).await?;
 
-    let users = krate
-        .owners(&conn)
-        .await?
+    let (users, teams) = tokio::try_join!(
```
carols10cents: I'm a little bit concerned that there's now duplication/potential divergence between the queries that power `list_owners` and the other uses of `krate.owners`. I realize this spot is the way it is because it no longer needs to go through the `Owner` enum... also I'm not seeing unit tests for `krate.owners`, `PublicUser::owning`, and `Team::owning`. Do these things cause you any concern? Why or why not?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14384/a313be6dfb6d67791f208fedf6c2129a3b47ccce..f43b06235d9d7e9d16ddc1ede893ba9d307065b0)*

crates/crates_io_database/src/models/user.rs
```diff
@@ -11,6 +11,40 @@ use crate::fns::lower;
...
+    pub gh_avatar: Option<String>,
+    pub username: String,
+    pub created_at: Option<DateTime<Utc>>,
+}
+
+impl PublicUser {
+    /// Loads active user owners for `krate`, ordered by owner ID.
+    ///
+    /// Team owners and deleted ownership records are excluded.
+    pub async fn owning(
```
carols10cents: It looks like there's now only one remaining usage of `User::owning`: https://github.com/rust-lang/crates.io/blob/3ee7f5c72fb5855a924a395a00a515ef922818d9/src/controllers/krate/owners.rs#L292

and that spot doesn't need anything from `User` that isn't in `PublicUser`. Wdyt about changing that spot to `PublicUser::owning` and removing `User::owning` to reduce duplication?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14384/a313be6dfb6d67791f208fedf6c2129a3b47ccce..f43b06235d9d7e9d16ddc1ede893ba9d307065b0)*

crates/crates_io_database/src/models/team.rs
Turbo87: > if not a separate PR

fair :)

-> https://github.com/rust-lang/crates.io/pull/14402

src/controllers/krate/owners.rs
Turbo87: > I'm a little bit concerned that there's now duplication/potential divergence between the queries that power `list_owners` and the other uses of `krate.owners`. I realize this spot is the way it is because it no longer needs to go through the `Owner` enum...

yeah, exactly. I played around with generics today to try and make this use the same code paths, but I'm not sure it's worth the complexity to make this work properly 😅 

> also I'm not seeing unit tests for `krate.owners`, `PublicUser::owning`, and `Team::owning`. Do these things cause you any concern? Why or why not?

I think our integration tests are already covering a decent chunk of the relevant cases, so I'm not too concerned at the moment.

crates/crates_io_database/src/models/user.rs
Turbo87: in https://github.com/rust-lang/crates.io/pull/14402 I've actually gone the other way and made `Crate::owners()` rely on `User::owning()` to reduce the duplication between them, which had the nice side-effect of simplifying `Crate::owners()` quite significantly :)

Turbo87:
rebased on top of https://github.com/rust-lang/crates.io/pull/14402 now

carols10cents (approved):
Looks great now, thank you!!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14384/d4d53b62d86e9ed7e93e1575cf97d704dc9f4910..6334c42ed292b0715e7bb6043ee2757b10694967)*

PR #14259, opened 2026-07-16
Fully remove the `users.gh_encrypted_token` column

carols10cents:
This drops the `users.gh_encrypted_token` that has been completely migrated to `oauth_github.encrypted_token`.

Extracted PRs that have been deployed already:

- https://github.com/rust-lang/crates.io/pull/14258
- https://github.com/rust-lang/crates.io/pull/14357
- https://github.com/rust-lang/crates.io/pull/14387

PR #14387, opened 2026-07-31
Stop writing `users.gh_encrypted_token`

carols10cents:
This value is no longer being read anywhere, so now we can stop writing it.

The value is also being written to the `oauth_github.encrypted_token` field; if we need to revert this, we can backfill any missed values in `users.gh_encrypted_token` using `oauth_github.encrypted_token`.

Part of https://github.com/rust-lang/crates.io/pull/14259.

### DIFF
--- a/crates/crates_io_database/src/models/user.rs
+++ b/crates/crates_io_database/src/models/user.rs
@@ -97,7 +97,6 @@ pub struct NewUser<'a> {
     pub gh_login: &'a str,
     pub username: &'a str,
     pub name: Option<&'a str>,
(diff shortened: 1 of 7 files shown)


Turbo87 (approved):
👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14387/3ee7f5c72fb5855a924a395a00a515ef922818d9..9bb54e6f2dc8aa02cc47bc683d1c8d6534c68796)*

eth3lbert (approved):

PR #14388, opened 2026-07-31
Add backend support for explicit signup

Turbo87:
This introduces an `explicit_signup_enabled` feature flag and tagged authorization outcomes so existing GitHub-linked users continue to sign in while new users enter a bounded pending-signup session. Until the signup page lands and if the feature flag is enabled, the frontend reports that signup is temporarily unavailable without marking the user as logged in.

Best reviewed commit-by-commit :)

/cc @carols10cents 

### Related

- #13907

### DIFF
--- a/src/controllers/session.rs
+++ b/src/controllers/session.rs
@@ -10,7 +10,7 @@ use crate::util::errors::{AppResult, bad_request, server_error};
 use crate::util::oauth::ReqwestClient;
 use crate::views::EncodableMe;
 use axum::Json;
-use chrono::Utc;
+use chrono::{DateTime, Utc};
 use crates_io_github::{GitHubAuth, GitHubUser};
 use crates_io_session::SessionExtension;
 use diesel::prelude::*;
@@ -21,6 +21,8 @@ use oauth2::{AuthorizationCode, CsrfToken, Scope, TokenResponse};
 use secrecy::ExposeSecret;
 use serde::{Deserialize, Serialize};
 use tracing::{error, warn};
+
+const PENDING_SIGNUP_KEY: &str = "pending_signup";
 
 #[derive(Debug, Serialize, utoipa::ToSchema)]
 pub struct BeginResponse {
@@ -51,6 +53,8 @@ pub struct BeginResponse {
     ),
 )]
 pub async fn begin_session(app: AppState, session: SessionExtension) -> Json<BeginResponse> {
+    session.remove(PENDING_SIGNUP_KEY);
+
     let (url, state) = app
         .github_oauth
         .authorize_url(oauth2::CsrfToken::new_random)
@@ -84,12 +88,35 @@ pub enum AuthorizeResponse {
     SignupRequired,
 }
 
+/// GitHub account details retained while a user completes signup.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct PendingSignup {
+    /// GitHub-controlled account details captured during OAuth authorization.
+    pub github_user: GitHubUser,
+    /// Encrypted GitHub access token used when the account is created.
+    pub encrypted_token: Vec<u8>,
+    /// Creation time used to determine when this pending signup expires.
+    pub created_at: DateTime<Utc>,
+}
+
+impl PendingSignup {
+    /// Creates pending signup state with the current time.
+    fn new(github_user: GitHubUser, encrypted_token: Vec<u8>) -> Self {
+        Self {
+            github_user,
+            encrypted_token,
+            created_at: Utc::now(),
+        }
+    }
+}
+
 /// Complete authentication flow.
 ///
 /// This route is called from the GitHub API OAuth flow after the user accepted or rejected
 /// the data access permissions. It will check the `state` parameter and then call the GitHub API
-/// to exchange the temporary `code` for an API token. The response indicates whether the
-/// corresponding user was signed in or needs to complete signup.
+/// to exchange the temporary `code` for an API token. Existing crates.io users are signed in.
+/// New users must complete signup when explicit signup is enabled and are created immediately
+/// otherwise.
 ///
 /// see <https://developer.github.com/v3/oauth/#github-redirects-back-to-your-site>
 #[utoipa::path(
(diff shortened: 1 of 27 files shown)


src/controllers/session.rs
```diff
@@ -70,12 +81,174 @@ pub struct AuthorizeBody {
...
+    responses((status = 200, description = "Successful Response", body = inline(EncodableMe))),
+)]
+pub async fn complete_pending_signup(
+    app: AppState,
+    session: SessionExtension,
+    Json(body): Json<CompletePendingSignupRequest>,
+) -> AppResult<Json<EncodableMe>> {
+    let pending_signup = load_pending_signup(&session)?;
+    let mut github_user = pending_signup.github_user;
+    github_user.email = Some(body.signup.email.to_string());
```
eth3lbert, line 241: FWIW, from what I've learned and experienced, relying entirely on the client side can easily lead to potential security issues like replay attacks. Even though a malicious user might not be able to do anything harmful through these routes, I still have a slight concern about whether this is the most suitable approach. I would probably choose to track the signup state and temporary signup session in the DB instead. I'm open to discussion though, and would love to hear what others think!

src/controllers/session.rs
```diff
@@ -70,12 +81,174 @@ pub struct AuthorizeBody {
...
+/// GitHub account details retained while a user completes signup.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct PendingSignup {
+    /// GitHub-controlled account details captured during OAuth authorization.
+    pub github_user: GitHubUser,
+    /// Encrypted GitHub access token used when the account is created.
+    pub encrypted_token: Vec<u8>,
+    /// Time when GitHub authorization completed and the absolute lifetime began.
+    pub created_at: DateTime<Utc>,
+}
```
eth3lbert, line 107: Although this looks good when only considering the GitHub provider, this schema might vary once we take multiple IdPs into account. I'm wondering if we should make this more generic by adding a provider field, or if we should make the name more explicit by using a Github prefix. What do you think?

eth3lbert (commented):
I just did a first review and left some nit comments, but it would be great if others could also take a look!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14388/415422e1b9d1f4a886ed53669a12e2769e3b2584..405aa22902fc7c0d76a4018014ae3e61f09b2bc0)*

src/controllers/session.rs
Turbo87: > relying entirely on the client side can easily lead to potential security issues like replay attacks

that's why the signup struct is only valid for 30min and uses a signed cookie thing, where we can verify on the server that the server has create that cookie.

> `github_user.email = Some(body.signup.email.to_string());`

the email specifically should be overridable by the user, which is why we read it from the request body JSON object. not sure if you explicitly included that line for your comment?

src/controllers/session.rs
Turbo87: I think we can do that once we actually implement support for other providers and know what we need. the pending signup objects only have a lifetime of 30min, so if we want to change the schema later, it shouldn't be too much of a problem, I think.

src/controllers/session.rs
eth3lbert: Yeah, I know that's the point of a short-lived session cookie. I'm just wondering if we should stop users who already finished signing up from hitting that route. They'd fail to create a user anyway because of the duplicate `gh_id` and wouldn't gain anything, so maybe it's just overengineering and not worth it.

However, IIRC, we'll likely have session control in the database in the future anyway, so I wonder if this should also be tracked and controllable from the backend?

src/controllers/session.rs
eth3lbert: fair enough!

src/controllers/session.rs
Turbo87: if they finished signing up then the pending signup cookie should have been removed and the `load_pending_signup()` call would fail, so they would already not be able to use that route.

> However, IIRC, we'll likely have session control in the database in the future anyway, so I wonder if this should also be tracked and controllable from the backend?

we'd still have to save something in the session cookie to be able to find the database entry, so I'm not sure how much we'd gain from that. for the regular authentication sessions I think it makes sense because those are long-lived and it would be good to be able to revoke sessions. but for the short-lived data in the pending signup I'm not sure it's worth the complexity.

eth3lbert: > if they finished signing up then the pending signup cookie should have been removed and the `load_pending_signup()` call would fail, so they would already not be able to use that route.

Oh, yeah, I was referring more to a malicious user who captures and saves the cookie.


> but for the short-lived data in the pending signup I'm not sure it's worth the complexity.

Perhaps @LawnGnome could put his security hat on and add a few words :D

src/tests/util/test_app.rs
```diff
@@ -600,6 +600,7 @@ fn simple_config() -> config::Server {
         disable_token_creation: None,
         banner_message: None,
         features: FeaturesConfig {
+            explicit_signup_enabled: true,
```
carols10cents, line 603: Just wanted to note that always having `explicit_signup_enabled` set to true in integration tests (and i don't see any that set it to false) while set to false in production means we're testing different behavior than we're deploying, which runs a risk of breaking production 😅 So I hope the length of time we need to have this feature flag in the code will be short!

src/controllers/session.rs
```diff
@@ -70,12 +81,174 @@ pub struct AuthorizeBody {
...
+
+/// GitHub account details retained while a user completes signup.
+#[derive(Debug, Deserialize, Serialize)]
+pub struct PendingSignup {
+    /// GitHub-controlled account details captured during OAuth authorization.
+    pub github_user: GitHubUser,
+    /// Encrypted GitHub access token used when the account is created.
+    pub encrypted_token: Vec<u8>,
+    /// Time when GitHub authorization completed and the absolute lifetime began.
```
carols10cents: When I first read this comment, I was confused about the absolute lifetime of... _what_? my first guess was of the user being created, like this was going to be used as the `users.created_at` value or something. One of the commit messages after this comment was introduced cleared it up for me-- so I think this would be clearer as:


```suggestion
    /// Time when GitHub authorization completed and the absolute lifetime of the pending signup state in the session began.
```

src/controllers/session.rs
```diff
@@ -311,13 +525,205 @@ async fn find_user_by_gh_id(mut conn: &AsyncPgConnection, gh_id: i32) -> QueryRe
...
+            save_user_to_database(false, &github_user(), &[1, 2, 3], &emails, &mut conn).await
+        );
+
+        assert_some!(user_id);
+        assert_eq!(count_users(&mut conn).await, 1);
+        assert_eq!(count_oauth_github(&mut conn).await, 1);
+    }
+
+    #[tokio::test]
+    async fn explicit_signup_allows_new_user_during_read_only_mode() {
```
carols10cents: this test name confused me at first because i thought by "allows new user during read-only mode" it meant "_creates_ new user during read-only mode", but after reading and thinking more, I see that it means "doesn't error during read-only mode (but doesn't create a new user either)". 

I'm not sure how best to clarify in a concise way... maybe...


```suggestion
    async fn explicit_signup_allows_identifying_new_user_during_read_only_mode() {
```

... because it can query to see that this user would be new, were we able to create it?

or...

```suggestion
    async fn explicit_signup_doesnt_error_for_new_user_during_read_only_mode() {
```

?

src/controllers/session.rs
```diff
@@ -70,12 +81,174 @@ pub struct AuthorizeBody {
...
+}
+
+/// Loads valid pending-signup state from the signed session cookie.
+fn load_pending_signup(session: &SessionExtension) -> AppResult<PendingSignup> {
+    let Some(value) = session.get(PENDING_SIGNUP_KEY) else {
+        return Err(bad_request(PENDING_SIGNUP_ERROR));
+    };
+
+    let pending_signup: PendingSignup =
+        serde_json::from_str(&value).map_err(|_| bad_request(PENDING_SIGNUP_ERROR))?;
```
carols10cents: Should a pending-signup value that can't be parsed as JSON also be removed, as an expired pending-signup value is? Otherwise how will the user be able to recover if they're in this case?

src/controllers/session.rs
```diff
@@ -70,12 +81,174 @@ pub struct AuthorizeBody {
...
+pub struct CompletePendingSignupData {
+    /// The email address to associate with the new account.
+    #[schema(value_type = String, format = Email, example = "new-user@example.com")]
+    email: Address,
+}
+
+/// Loads valid pending-signup state from the signed session cookie.
+fn load_pending_signup(session: &SessionExtension) -> AppResult<PendingSignup> {
+    let Some(value) = session.get(PENDING_SIGNUP_KEY) else {
+        return Err(bad_request(PENDING_SIGNUP_ERROR));
```
carols10cents, line 159: Currently, whether I have `explicit_signup` turned on or off, I'm signed in as an existing user locally, and I navigate to http://localhost:5173/api/private/session/signup, I see this error from this code path. Which is fine (this code path isn't doing anything), although the error isn't exactly accurate? Do we care if users go to this URL manually before we've enabled the feature flag? 

That is, should this instead check `if !explicit_signup_enabled` and return, say, a 404 error instead? And then if explicit signup _is_ enabled but there's a `user_id` in the session, should this return still a 400 but an error message that's something like "you can't sign up, you're already logged in"?

carols10cents (commented):
A few small things, but I tested this out locally with and without the feature enabled, with existing and new users, and it behaved as described! Thank you!!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14388/415422e1b9d1f4a886ed53669a12e2769e3b2584..405aa22902fc7c0d76a4018014ae3e61f09b2bc0)*

src/tests/util/test_app.rs
Turbo87: > note that always having `explicit_signup_enabled` set to true in integration tests (and i don't see any that set it to false)

yeah, that's true. I initially planned to use `explicit_signup_enabled: false` for the tests that check the old behavior, but since we haven't implemented a mock OAuth abstraction (yet) we don't have any tests that check the endpoint through it's HTTP interface. instead, we use unit tests for the fns called by this endpoint (e.g. `save_user_to_database()`). not a great long-term solution IMHO, but good enough for now.

src/controllers/session.rs
Turbo87: done 👍

src/controllers/session.rs
Turbo87: when the user restarts the OAuth flow the broken `PendingSignup` would have been overwritten, so recovery would be possible, but I agree that there is no point in keeping the broken record aroung 👍

src/controllers/session.rs
Turbo87: renamed ✅

src/controllers/session.rs
Turbo87: yeah, I agree that the doc comment wasn't particularly helpful without context. I've adjusted it to hopefully be more helpful now :)

carols10cents (approved):
LGTM now, thank you! I think we should merge this without waiting for @LawnGnome or anyone else more security-minded (I also don't know how to go about trying to replay sessions or edit cookies), because the feature flag is off in production. Who knows, maybe someone donating their LLM tokens will find any problems with this for us! 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14388/7bef82cebb702b89ec8d3f13facf67a83bc7d090..725eee76b65019bf735808c3a349cf2dc7462a4a)*

PR #14396, opened 2026-08-02
Generate db-dump.tar.gz that supports multithreaded extract

dtolnay:
As part of optimizing https://github.com/dtolnay/db-dump, I came up with a way that db-dump.tar.gz would be able to support reading files in parallel and efficiently skipping files. See https://github.com/dtolnay/chapter-tgz.

On my machine, reading the existing db-dump.tar.gz files currently takes about 6.5 seconds using the following program.

```rust
use std::fs::File;
use std::io::Cursor;

// 6.5 seconds
fn main() -> anyhow::Result<()> {
    let file = File::open("db-dump-original.tar.gz")?;
    let data = unsafe { memmap2::Mmap::map(&file)? };
    let mut tgz = tar::Archive::new(flate2::read::GzDecoder::new(Cursor::new(&*data)));
    for _entry in tgz.entries()? {}
    Ok(())
}
```

With the new approach, the above program continues to take 6.5 seconds, but it becomes possible to extract different data files simultaneously on different threads, which takes only 3.1 seconds. Instead of decompressing everything serially as usually required by tgz, the runtime becomes governed by the largest file in the tgz, which is currently dependencies.csv (43.6% of the total csv size &mdash; that closely matches the ratio of the speedup).

```rust
use std::fs::File;
use std::io::{self, Cursor};
use std::thread;

// 3.1 seconds
fn main() -> anyhow::Result<()> {
    let file = File::open("db-dump-with-chapters.tar.gz")?;
    let data = unsafe { memmap2::Mmap::map(&file)? };
    let tgz = chapter_tgz::TgzReader::open(Cursor::new(&*data))?;
    let mut chapters = Vec::new();
    for i in 0..tgz.chapters() {
        let chapter = tgz.independent_read_chapter(i)?;
        chapters.push(chapter);
    }

    thread::scope(|scope| {
        for mut chapter in chapters {
            scope.spawn(move || {
                if let Err(err) = (|| -> io::Result<()> {
                    for _entry in chapter.entries()? {}
                    Ok(())
                })() {
                    eprintln!("Error: {err}");
                }
            });
        }
    });

    Ok(())
}
```

But most uses of `db_dump` do not require all of the data files. The API of `db_dump` was already designed from the beginning to stop reading as soon as all of the files needed in the user's code were found. However, it was not possible to _jump ahead_ in the tgz directly to the needed files, so a prefix of the tgz always had to be read, including files that were not needed. The new approach supports jumping directly to specific files, as in the following code which reads users.csv in 0.01 seconds.

```rust
use anyhow::Context as _;
use std::fs::File;
use std::io::{BufRead as _, BufReader, Cursor};

// 0.01 seconds
fn main() -> anyhow::Result<()> {
    let file = File::open("db-dump-with-chapters.tar.gz")?;
    let data = unsafe { memmap2::Mmap::map(&file)? };
    let mut tgz = chapter_tgz::TgzReader::open(Cursor::new(&*data))?;
    while let Some(mut chapter) = tgz.next_chapter() {
        let mut entries = chapter.entries()?;
        let entry = entries.next().context("expected at least one entry in chapter")??;
        let path = entry.path()?;
        let file_name = path.file_name().context("tar entry should have filename")?;
        if file_name == "users.csv" {
            println!("{} lines", BufReader::new(entry).lines().count());
        }
    }
    Ok(())
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1248,6 +1248,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "chapter-tgz"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "71d3b546f2d916ddf609e8f8e271c93c9dcd66d7500a87ee8d81d4c159839d7c"
+dependencies = [
+ "flate2",
+ "tar",
+]
+
+[[package]]
 name = "chinese-number"
 version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1876,6 +1886,7 @@ name = "crates_io_database_dump"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "chapter-tgz",
  "chrono",
  "crates_io_env_vars",
  "crates_io_test_db",
--- a/crates/crates_io_database_dump/Cargo.toml
+++ b/crates/crates_io_database_dump/Cargo.toml
@@ -9,6 +9,7 @@ workspace = true
 
 [dependencies]
 anyhow = "=1.0.104"
+chapter-tgz = "0.1"
 chrono = { version = "=0.4.45", default-features = false, features = ["clock", "serde"] }
 crates_io_version = { path = "../crates_io_version" }
 flate2 = "=1.1.9"
--- a/crates/crates_io_database_dump/src/lib.rs
+++ b/crates/crates_io_database_dump/src/lib.rs
@@ -1,6 +1,7 @@
 #![doc = include_str!("../README.md")]
 
 use anyhow::{Context, anyhow};
+use chapter_tgz::TgzWriter;
 use serde::Serialize;
 use std::fs;
 use std::fs::File;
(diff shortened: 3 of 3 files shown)


Turbo87:
I haven't looked at this in detail yet, but doesn't the db dump zip file already solve some of these issues?

dtolnay:
To be honest I didn't remember that a zip is available until coming to edit this code in crates.io and seeing the ZipWriter being written.

I tested db-dump.zip and the performance for each of the use cases is similar to the new tgz.

Extract everything on one thread: 6.9 seconds

```rust
use std::fs::File;
use std::io::{self, Cursor};

// 6.9 seconds
fn main() -> anyhow::Result<()> {
    let file = File::open("db-dump.zip")?;
    let data = unsafe { memmap2::Mmap::map(&file)? };
    let mut zip = zip::ZipArchive::new(Cursor::new(&*data))?;
    for i in 0..zip.len() {
        let mut file = zip.by_index(i)?;
        io::copy(&mut file, &mut io::sink())?;
    }
    Ok(())
}
```

Extract data files on separate threads: 3.4 seconds

```rust
use std::fs::File;
use std::io::{self, Cursor};
use std::thread;

// 3.4 seconds
fn main() -> anyhow::Result<()> {
    let file = File::open("db-dump.zip")?;
    let data = unsafe { memmap2::Mmap::map(&file)? };
    let mut zip = zip::ZipArchive::new(Cursor::new(&*data))?;
    let mut csv_indices = Vec::new();
    for i in 0..zip.len() {
        let entry = zip.by_index(i)?;
        if entry.is_file() && entry.name().starts_with("data/") {
            csv_indices.push(i);
        }
    }

    thread::scope(|scope| {
        for i in csv_indices {
            scope.spawn({
                let data = &*data;
                move || {
                    if let Err(err) = (|| -> io::Result<()> {
                        let mut zip = zip::ZipArchive::new(Cursor::new(data))?;
                        let mut entry = zip.by_index(i)?;
                        io::copy(&mut entry, &mut io::sink())?;
                        Ok(())
                    })() {
                        eprintln!("Error: {err}");
                    }
                }
            });
        }
    });

    Ok(())
}
```

Extract single small file: 0.01 seconds

```rust
use std::fs::File;
use std::io::{BufRead as _, BufReader, Cursor};

// 0.01 seconds
fn main() -> anyhow::Result<()> {
    let file = File::open("db-dump.zip")?;
    let data = unsafe { memmap2::Mmap::map(&file)? };
    let mut zip = zip::ZipArchive::new(Cursor::new(&*data))?;
    let entry = zip.by_name("data/users.csv")?;
    println!("{} lines", BufReader::new(entry).lines().count());
    Ok(())
}
```

Turbo87:
> I tested db-dump.zip and the performance for each of the use cases is similar to the new tgz.

in that case, are you okay with closing the PR? I'd prefer not to increase the complexity of our tarball generation if the zipfile can already cover these needs adequately :)

dtolnay:
Sounds good. Thanks!

PR #14402, opened 2026-08-03
Refactor crate owner queries

Turbo87:
This extracts a set of owner query changes from #14384 into a dedicated PR and builds on them with related cleanup. `User::owning()` and `Team::owning()` now return their concrete model types, explicit `EncodableOwner` constructors handle serialization, and `Crate::owners()` reuses and pipelines both queries.

Owner ordering moves out of the database queries and into the API handlers. This preserves the established ordering of `/owners` while introducing deterministic ordering for `/owner_user` and `/owner_team`, which previously had no explicit ordering. `Crate::owners()` users that don't need the deterministic ordering no longer have to pay the price for it.

The owner endpoint tests now use JSON snapshots and cover multiple user and team owners across all three endpoints.

### Related

- #14384

### DIFF
--- a/crates/crates_io_database/src/models/krate.rs
+++ b/crates/crates_io_database/src/models/krate.rs
@@ -205,8 +205,10 @@ impl Crate {
     }
 
     pub async fn owners(&self, conn: &AsyncPgConnection) -> QueryResult<Vec<Owner>> {
-        let users = User::owning(self, conn).await?.into_iter().map(Owner::User);
-        let teams = Team::owning(self, conn).await?.into_iter().map(Owner::Team);
+        let (users, teams) = tokio::try_join!(User::owning(self, conn), Team::owning(self, conn))?;
+
+        let users = users.into_iter().map(Owner::User);
+        let teams = teams.into_iter().map(Owner::Team);
 
         Ok(users.chain(teams).collect())
     }
--- a/crates/crates_io_database/src/models/krate.rs
+++ b/crates/crates_io_database/src/models/krate.rs
@@ -1,6 +1,6 @@
 use crate::fns::canon_crate_name;
 use crate::models::version::TopVersions;
-use crate::models::{CrateOwner, Owner, OwnerKind, User, Version};
+use crate::models::{CrateOwner, Owner, User, Version};
 use crate::schema::*;
 use chrono::{DateTime, Utc};
 use diesel::associations::Identifiable;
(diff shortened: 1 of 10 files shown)


crates/crates_io_database/src/models/team.rs
```diff
@@ -53,11 +53,10 @@ impl NewTeam<'_> {
 
 impl Team {
     pub async fn owning(krate: &Crate, mut conn: &AsyncPgConnection) -> QueryResult<Vec<Self>> {
-        let base_query = CrateOwner::belonging_to(krate).filter(crate_owners::deleted.eq(false));
-        base_query
+        CrateOwner::by_owner_kind(OwnerKind::Team)
```
carols10cents, line 56: I didn't even notice that `Team::owning` was constructing queries differently from `User::owning` before; this is way nicer!

src/tests/snapshots/integration__owners__check_ownership_one_crate-2.snap
```diff
@@ -0,0 +1,40 @@
+---
+source: src/tests/owners.rs
+expression: response.json()
+---
+{
+  "users": [
```
carols10cents, line 6: kinda weird that the response for `/owners` uses the key "users" and also returns teams; i see that it didn't change though :( and now we have to be backwards compatible :( oh well. just wanted to note that i noticed it.

crates/crates_io_database/src/models/krate.rs
```diff
@@ -204,24 +204,9 @@ impl Crate {
...
-            .filter(crate_owners::crate_id.eq(self.id))
-            .inner_join(teams::table)
-            .select(Team::as_select())
-            .load(&mut conn)
-            .await?
-            .into_iter()
-            .map(Owner::Team);
+    pub async fn owners(&self, conn: &AsyncPgConnection) -> QueryResult<Vec<Owner>> {
+        let users = User::owning(self, conn).await?.into_iter().map(Owner::User);
+        let teams = Team::owning(self, conn).await?.into_iter().map(Owner::Team);
```
carols10cents, line 209: 🎉

carols10cents (approved):
so much nicer, and easier to review as a separate PR :) thank you!!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14402/25102112aed12fbfbbea034dd1eb82f802bd2918..a0d5ace3910cfdbcae266e422c5c0fcfcce6f5c7)*

src/tests/snapshots/integration__owners__check_ownership_one_crate-2.snap
Turbo87: yeah, I assume this is a backward compatibility thing, back from when teams were introduced. probably not worth a v2 endpoint yet though 😅

PR #14417, opened 2026-08-04
Use crates.io usernames for user lookups and responses

### LINKED ISSUE #13763: Update the Find/Show User API to query users.username

### LINKED ISSUE #14379: Change `Encodable{Public|Private}User` to set `login` to `users.username` instead of `users.gh_login`
For use by the user API responses.

`EncodablePrivateUser` change needs to be made here: https://github.com/rust-lang/crates.io/blob/a313be6dfb6d67791f208fedf6c2129a3b47ccce/crates/crates_io_api_types/src/lib.rs#L756

`EncodablePublicUser` change needs to be made here: https://github.com/rust-lang/crates.io/blob/a313be6dfb6d67791f208fedf6c2129a3b47ccce/crates/crates_io_api_types/src/lib.rs#L812

and the associated documentation for those API types should be changed to say "crates.io username".

Turbo87:
User API lookups now resolve canonical crates.io usernames, while response login fields expose `users.username`. GitHub profile URLs continue to use the linked GitHub login, and team responses remain unchanged.

A new database index is introduced to keeps canonical username lookups efficient. The MSW models and handlers now represent crates.io and GitHub identities independently. Production currently still synchronizes both values, so this should currently not result in any visible changes.

### Related

- Resolves https://github.com/rust-lang/crates.io/issues/13763
- Resolves https://github.com/rust-lang/crates.io/issues/14379


carols10cents (approved):
LGTM!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14417/7bef82cebb702b89ec8d3f13facf67a83bc7d090..7faecfe42b0a70626955c5d700f9c7c6ddefde9f)*

moskirathe (approved):

Issue #13764, opened 2026-05-26
Add an indication on crate and user pages when crates.io username doesn't match GitHub username


carols10cents:
Depends on https://github.com/rust-lang/crates.io/issues/13763 adding the `github_username_matches` field; the UI changes here should use the value in that field. If `github_username_matches` is true, that's the usual case that doesn't get any changes. If `github_username_matches` is false, that's the situation we want to communicate to the viewer of these pages.

The icon could be :warning: or something more neutral such as :information_source:. I think using emoji is fine; I'm open to arguments that we should use a custom image instead.

Hovering (on desktop) or tapping (on mobile) should display text in some manner that says something like "This user's crates.io username is different from their GitHub username." Should we say something like "See their user page for details" or keep it short and sweet? Should we link to documentation about crates.io usernames?

PR #14437, opened 2026-08-07
Add `github_username_matches` field to user/owner responses

### LINKED ISSUE #14380: Add a `github_username_matches` field to `Encodable{Public|Private}Api`

Turbo87:
This adds `github_username_matches` to the shared public `User` schema and the user variant of `Owner`. The value comes from an exact, case-sensitive check across linked GitHub accounts.

Using the shared `PublicUser` struct keeps version publishers, audit users, invitations, crate owners, and Find User responses consistent, while private user responses avoid the additional query. The MSW models mirror the same linked-account behavior.

I opted to not add it to the private user responses, since the authenticated user already knows if they have a matching GitHub username or not and if they look at their profile page then the public user response would be used anyway.

Best reviewed commit-by-commit :)

### Related

- Resolves #14380
- https://github.com/rust-lang/crates.io/issues/13764

### DIFF
--- a/crates/crates_io_api_types/src/lib.rs
+++ b/crates/crates_io_api_types/src/lib.rs
@@ -549,13 +549,20 @@ pub struct EncodableOwner {
     /// The avatar URL of the team or user.
     #[schema(example = "https://avatars2.githubusercontent.com/u/1234567?v=4")]
     pub avatar: Option<String>,
+
+    /// Whether a linked GitHub username exactly matches the crates.io username.
+    ///
+    /// This field is present only for user owners.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub github_username_matches: Option<bool>,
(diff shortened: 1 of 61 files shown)


carols10cents (approved):
LGTM!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14437/709506b938313aa8df51f2e0fce133bc0e7b37f7..5a7941678c7993d91caacda4b2c8cefdeeb3d71c)*

PR #14487, opened 2026-08-20
crate-backup: store reverse dependencies

marcoieni:
## Test

Run `python3 script/crate-backup.py poetry-book`

outputs the file `reverse-dependencies.csv`:

```
crate,dependent_crate,dependent_version,requirement,kind,target,optional,default_features,features,downloads
poetry-book,poetry-book-cli,0.1.4,^0.1.3,normal,,False,True,[],6929
poetry-book,poetry-book-web,0.1.0,^0.1.1,normal,,False,True,[],1759
```

it reflects the [UI](https://crates.io/crates/poetry-book/reverse_dependencies):
image

## AI disclosure

I used GPT5.6-Sol with the codex harness to generate this change. I reviewed its output and changed it where necessary.

### DIFF
--- a/script/crate-backup.py
+++ b/script/crate-backup.py
@@ -8,6 +8,21 @@ from pathlib import Path
 
 USER_AGENT = "crate-backup (https://github.com/rust-lang/crates.io)"
 DOWNLOADS_CSV = "downloads.csv"
+REVERSE_DEPENDENCIES_CSV = "reverse-dependencies.csv"
+PER_PAGE = 100
+DOWNLOADS_CSV_HEADER = ["crate", "version", "downloads"]
+REVERSE_DEPENDENCIES_CSV_HEADER = [
+    "crate",
+    "dependent_crate",
+    "dependent_version",
+    "requirement",
+    "kind",
+    "target",
+    "optional",
+    "default_features",
+    "features",
+    "downloads",
+]
 
 
 def fetch(url):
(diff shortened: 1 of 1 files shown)


Turbo87:
hmm... something like `serde` has 115975 reverse dependencies. I'm not sure if it's really viable to iterate through them all. also the API only shows the reverse dependencies where the "default version" of crate A depends on crate B, but not for all previous versions. if we want to record the full set of reverse dependencies for all depending versions then we would probably have to query the database directly.

marcoieni:
should I query the db from python? Or at this point is better to copy paste a query in the runbook?

Turbo87:
@marcoieni since I just noticed, please use your personal repo fork for PRs in the future 🙏

Turbo87:
> should I query the db from python? 

the python script does not have access to the database at the moment and I think it's better to keep it that way for now, so that people without db access can still use that part of the script.

> Or at this point is better to copy paste a query in the runbook?

yeah, I think that's easiest for now

marcoieni:
ooops, sorry, should we have a ruleset that prevents this from happening?

Turbo87:
iirc I've had some cases in the past where it was beneficial to work directly on a branch in the upstream repo due to how CI works. might have been when working on https://github.com/rust-lang/crates.io/blob/main/.github/workflows/update-cdn-ip-ranges.yml, but I don't remember exactly. I think convention is probably sufficient.

PR #14488, opened 2026-08-20
Add button to copy dependency version only

daviddanielng:
### Motivation

Sometimes users just need the version number on its own  e.g. to update an existing dependency line, whether it's outdated because of an old snippet, an LLM suggestion, or just stale copy-pasted code. This avoids retyping or editing a full snippet just to grab the version.

### Changes

- Added `versionOnly`, stripping build metadata (`+...`) like the existing `tomlSnippet`.
- Added a "Or copy just the version:" copy button using the existing `CopyButton` component, with clipboard-fallback support.

### Testing

- `pnpm --filter crates.io-svelte test` — 190 tests pass.


Turbo87:
Bildschirmfoto 2026-08-20 um 14 41 26

IMHO it seems a bit cluttered

PR #14514, opened 2026-08-24
controllers/user: Return requested linked accounts

### LINKED ISSUE #14382: Implement `include=linked_accounts` for user API requests
Much like the current crate API allows for opt-in of returning related data, the API requests that return user information should take an `include` parameter that indicates whether the response should return all linked account information for this user (once crates.io supports more services, right now this will only be associated information from `oauth_github`).

This info should be optionally included in this API request because to get it will require querying all OAuth tables. The crates.io frontend will add this parameter to the users API requests it makes on user pages, but not on users API requests it makes on crate pages or search result pages, etc, to avoid adding lots of load to the database.

Turbo87:
The Find User endpoint now accepts `include=linked_accounts` and returns public GitHub account details. Requests without the parameter keep the existing response shape, while users without linked accounts receive an empty array.

The OpenAPI schema, API client types, and MSW handler expose the same optional query and response data.

### Related

- Resolves https://github.com/rust-lang/crates.io/issues/14382

### DIFF
--- a/crates/crates_io_api_types/src/lib.rs
+++ b/crates/crates_io_api_types/src/lib.rs
@@ -853,6 +853,42 @@ impl From<PublicUser> for EncodablePublicUser {
     }
 }
 
+/// This type contains public data for an external account of a crates.io user.
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq, utoipa::ToSchema)]
+#[schema(as = LinkedAccount)]
+#[serde(tag = "provider")]
+pub enum EncodableLinkedAccount {
+    /// A linked GitHub account.
+    #[serde(rename = "github")]
+    GitHub {
+        /// The stable GitHub account ID.
+        account_id: String,
+
+        /// The GitHub username.
+        login: String,
+
+        /// The GitHub avatar URL.
+        #[schema(required)]
+        avatar: Option<String>,
+
+        /// The GitHub profile URL.
+        url: String,
+    },
+}
+
+impl EncodableLinkedAccount {
+    /// Creates a public linked GitHub account.
+    pub fn github(account_id: i64, login: String, avatar: Option<String>) -> Self {
+        let url = format!("https://github.com/{login}");
+        Self::GitHub {
+            account_id: account_id.to_string(),
+            login,
+            avatar,
+            url,
+        }
+    }
+}
+
 #[derive(Deserialize, Serialize, Debug, utoipa::ToSchema)]
 pub struct EncodableAuditAction {
     /// The action that was performed.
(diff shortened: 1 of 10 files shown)


crates/crates_io_api_types/src/lib.rs
```diff
@@ -853,6 +853,42 @@ impl From<PublicUser> for EncodablePublicUser {
...
+
+        /// The GitHub profile URL.
+        url: String,
+    },
+}
+
+impl EncodableLinkedAccount {
+    /// Creates a public linked GitHub account.
+    pub fn github(account_id: i64, login: String, avatar: Option<String>) -> Self {
+        let url = format!("https://github.com/{login}");
```
Turbo87: now that I think about it, do we actually need this? we could just generate it on the frontend instead 🤔 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14514/fe3702444d8f02203627f4d4d3b5aa66744a946a..33341dc66e9efd4ecc16d23f78d49ebcecfef483)*

eth3lbert: Yeah, I agree! We can always add it later if a need comes up.

src/controllers/user/other.rs
```diff
@@ -46,7 +76,56 @@ pub async fn find_user(
...
+        let accounts = accounts
+            .into_iter()
+            .map(|(account_id, login, avatar)| {
+                EncodableLinkedAccount::github(account_id, login, avatar)
+            })
+            .collect();
+        Some(accounts)
+    } else {
+        None
+    };
```
eth3lbert: I guess this part could be refactored to pipelined queries 🤔?

eth3lbert (approved):
LGTM! Just a minor suggestion regarding the pipelined queries.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14514/fe3702444d8f02203627f4d4d3b5aa66744a946a..33341dc66e9efd4ecc16d23f78d49ebcecfef483)*

src/controllers/user/other.rs
Turbo87: good point, I think that should be possible. let me give that a try 👍

Turbo87: I missed the `.filter(oauth_github::user_id.eq(user.id))` part. in other words: the second query depends on the result of the first, so we can't easily pipeline the queries without running the first one twice on the database side.

I think more relevant will be pipelining the queries to read the linked accounts for all the supported providers, once we have more than just GitHub. but for now we only support one provider, so no need for pipelining a single query just yet :)

eth3lbert: Oh, right, I missed that as well. I overlooked the fact that it passes `userid` rather than `username`. We could probably use a subselect to get the `userid`  here, but I'm not sure if it's worth the effort at this point.

Turbo87: > We could probably use a subselect to get the `userid` here

yeah, that's what I meant above:

> we can't easily pipeline the queries without running the first one twice on the database side.

probably not worth it in this case :)

PR #14516, opened 2026-08-25
markdown: Resolve `.` and `..` segments in relative URLs

### LINKED ISSUE #5318: Workspaced crate READMEs do not display relative images correctly
### Current Behavior

Viewing the README of a crate that is originally from a workspace (take [imgui-rs](https://crates.io/crates/imgui) as an example) the image is not found. This is specifically member crates that share a README from the parent folder, and relatively references images in that location.

### Expected Behavior

The image should be loaded correctly.

### Steps To Reproduce

1. Go to the [imgui-rs](https://crates.io/crates/imgui) crates.io page.

### Environment

- Browser: Safari 16.0
- OS: macOS 12.6

### Anything else?

I believe I know the crux of this issue. Take a look at this snippet I have:

At around [line 206 in lib.rs](https://github.com/rust-lang/crates.io/blob/8be845c314905f95fc972419c67eba57084dae2c/cargo-registry-markdown/lib.rs#L206):
```rust
new_url += if is_media { "raw/HEAD" } else { "blob/HEAD" };
if !self.base_dir.is_empty() {
    new_url += "/";
    new_url += &self.base_dir;
}
```

For reference, the URL generated currently is this:

`https://github.com/imgui-rs/imgui-rs/raw/HEAD/../hello_world.png`

Obviously, it's trying to append `base_dir` (which because this is a workspace crate, I'm assuming is `..`) but GitHub clearly doesn't like that. Removing the `..` and just writing:

`https://github.com/imgui-rs/imgui-rs/raw/HEAD/hello_world.png`

Does work! So I wonder if this routine should be modified, so we collapse `..` if needed. See imgui's Cargo.toml for how this happened:

```toml
[package]
name = "imgui"
readme = "../README.markdown"
```

I can write up a quick PR to fix this myself, but I wanted to know if there's any kind of weird edge case this code was supposed to be handling.

jakobjung10:
Relative URLs were appended to the README's directory verbatim, so a link like `../img/logo.png` in a crate that lives in a subdirectory of its repository produced a `raw/HEAD/subdir/../img/logo.png` URL that GitHub serves a 404 for. Resolve the segments before building the URL, clamping at the repository root.

SVGs were unaffected because appending the `sanitize=true` query reparses the URL, which collapses the segments as a side effect.

Closes #5318


Turbo87:
sorry for the long radio silence. I completely forgot about this PR 🙈

PR #14517, opened 2026-08-25
Extract user and team page header components

Turbo87:
This PR extracts the user and team page headers into dedicated `UserPageHeader` and `TeamPageHeader` components. I've also added a couple of Storybook stories to provide baseline coverage for default, missing-avatar, missing-name, and long-identity states.

This also revealed a few layout bugs that I'll fix in one of the follow-up PRs:

Bildschirmfoto 2026-08-25 um 19 43 49

### DIFF
--- a/svelte/src/lib/components/TeamPageHeader.stories.svelte
+++ b/svelte/src/lib/components/TeamPageHeader.stories.svelte
@@ -1,0 +1,61 @@
+<script module lang="ts">
+  import type { ComponentProps } from 'svelte';
+
+  import { defineMeta } from '@storybook/addon-svelte-csf';
+
+  import TeamPageHeader from './TeamPageHeader.svelte';
+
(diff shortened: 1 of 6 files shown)


eth3lbert (approved):
The only thing I'm unsure about is whether we should wrap long content that overflows, since it ends up hiding the provider icon 🤔. Other than that, LGTM!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14517/fe3702444d8f02203627f4d4d3b5aa66744a946a..e3a4a35f5d8cdd5f865e3a4facfc7fb1cbc833f4)*

Turbo87:
> The only thing I'm unsure about is whether we should wrap long content that overflows, since it ends up hiding the provider icon 🤔.

we should, and I already have a series of follow-up PRs prepared that do so 😉

PR #14523, opened 2026-08-25
jobs/archive_index_branch: Split pushes into commit chunks

Turbo87:
GitHub enforces a 2 GiB limit for each push. This job now force-pushes 100,000-commit checkpoints to a `temp` branch before publishing the snapshot branch.

### DIFF
--- a/src/worker/jobs/index/archive.rs
+++ b/src/worker/jobs/index/archive.rs
@@ -12,6 +12,33 @@ use tracing::{info, instrument, warn};
 use url::Url;
 
 const REMOTE_NAME: &str = "archive";
+const TEMP_BRANCH: &str = "temp";
+const COMMITS_PER_PUSH: usize = 100_000;
+
+/// A contiguous commit range ending at the checkpoint pushed to the archive.
+#[derive(Debug, PartialEq)]
+struct ArchivePushChunk {
+    start: usize,
+    end: usize,
+    commit: String,
+}
+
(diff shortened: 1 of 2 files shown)


src/worker/jobs/index/archive.rs
```diff
@@ -116,6 +143,67 @@ impl BackgroundJob for ArchiveIndexBranch {
...
+        let commit_count = chunks.last().map_or(0, |chunk| chunk.end);
+        let chunk_count = chunks.len();
+        info!(
+            "Found {commit_count} commits in {chunk_count} chunks for snapshot branch ({branch})",
+            branch = self.branch,
+        );
+
+        for chunk in chunks {
+            let ArchivePushChunk { start, end, commit } = chunk;
+            info!(
```
jlizen, line 172: nit: wouldn't this be a bit noisy? i would have expected only the top-level one to be info level

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14523/2004dfd43516e132c38c044c3fdb43a6684e0894..67172d32935a77e0f58458707a295babc3582281)*

jlizen (approved):

src/worker/jobs/index/archive.rs
Turbo87: it would be if this would run a lot, but since the job is queued manually every couple of weeks I'd rather have the detailed logs :)

PR #14548, opened 2026-08-28
Add an admin `yank-version --undo` command option

carols10cents:
We had `crates-io yank-version`; this PR adds an `--undo` flag that will unyank a version.

### DIFF
--- a/src/bin/crates-io/admin/yank_version.rs
+++ b/src/bin/crates-io/admin/yank_version.rs
@@ -15,8 +15,11 @@ use diesel_async::{AsyncConnection, AsyncPgConnection, RunQueryDsl};
 pub struct Opts {
     /// Name of the crate
     crate_name: String,
-    /// Version number that should be deleted
+    /// Version number that should be yanked or unyanked
     version: String,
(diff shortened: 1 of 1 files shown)


Turbo87:
since this is largely a copy of the yank-version command, what are your thoughts on adding a `--unyank` option to the existing command instead?

carols10cents:
> since this is largely a copy of the yank-version command, what are your thoughts on adding a `--unyank` option to the existing command instead?

I'm cool with that generally, but `crates-io yank-version --unyank` seems a bit weird. What about mirroring Cargo, which does `cargo yank --undo`, so this would be `crates-io yank-version --undo`?

carols10cents:
I went ahead with `--undo`, wydt?

Turbo87 (approved):
sounds good 👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14548/6e1154a10953511db929508277eda633252be7a6..d38f5314a0f1297b28c49c9384052c2ced735d23)*

PR #14571, opened 2026-09-02
Hide the report button for logged-out users

Turbo87:
The "Report crate" button is now only shown to logged-in users as an experiment to reduce spam reports. The support page and its footer link remain accessible to everyone.

This extracts the button visibility change from #13023. I was skeptical of the extra friction at the time, but the spam situation seems to have gotten worse, so I've changed my mind about trying this narrower change.

### Related

- #13023

### DIFF
--- a/e2e/acceptance/crate.spec.ts-snapshots/aria-latest.yml
+++ b/e2e/acceptance/crate.spec.ts-snapshots/aria-latest.yml
@@ -48,8 +48,6 @@
     - button "nanomsg = \"0.6.1\""
     - heading "Owners" [level=2]
     - list
-    - link "Report crate":
-      - /url: /support?crate=nanomsg&inquire=crate-violation
   - heading "Stats Overview" [level=3]
   - text: /\d+,\d+ Downloads all time 2 Versions published/
   - heading /Downloads over the last \d+ days/ [level=4]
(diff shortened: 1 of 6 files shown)


eth3lbert (commented):
I'm fine with this approach, but instead of only updating the visibility, how about requiring users to be logged in to report, as suggested in that PR? wdyt?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14571/c2a8915966df9ad7ced51c3ceacee6923ad123bf..5e8b8af53d11952dd6dbbb42da19fc9c7de693d0)*

Turbo87:
> but instead of only updating the visibility, how about requiring users to be logged in to report, as suggested in that PR? wdyt?

the thing is, I don't want to make it impossible for regular users to report crates or send us support requests. I only want to hide the button for the unauthenticated users that happen to find a crates.io link in some license texts, click it, and report a crate without even knowing what a crate is. we had a couple of cases lately where we got support requests asking us how they can remove something from their phone, because apparently some crate was mentioned somewhere in a license text or something like that.

eth3lbert:
> > but instead of only updating the visibility, how about requiring users to be logged in to report, as suggested in that PR? wdyt?
> 
> the thing is, I don't want to make it impossible for regular users to report crates or send us support requests. I only want to hide the button for the unauthenticated users that happen to find a crates.io link in some license texts, click it, and report a crate without even knowing what a crate is. we had a couple of cases lately where we got support requests asking us how they can remove something from their phone, because apparently some crate was mentioned somewhere in a license text or something like that.

Fair enough! Although I think that suggestion could also apply just to that button, I'm happy to approve this since it's simple and lightens the burden for the support team. We can always retrofit things later if others ask for it.

eth3lbert (approved):

PR #14575, opened 2026-09-02
WIP removing `users.gh_login`

carols10cents:
Will be submitting these as separate PRs safe to deploy

Connects to https://github.com/rust-lang/crates.io/issues/13772

PR #14576, opened 2026-09-02
Use crates.io username in admin contexts

carols10cents:
As part of the transition to crates.io usernames, change admin commands and APIs to use `users.username` rather than `users.gh_login`.

End users of crates.io should not see any observable effects of this PR.

Admins will need to remember that:

- Looking up a user by their token with the `verify-token` command will return the user's crates.io username, which currently always matches their GitHub username, but that will not necessarily be the case in the future.
- Using the `list_admin` API will look up users by their crates.io username (I think I might be the only one who uses this)
- Usernames in some log messages will now be crates.io usernames (I added "crates.io username" to the log messages as a clarification)

Extracted from https://github.com/rust-lang/crates.io/pull/14575. Shouldn't conflict with https://github.com/rust-lang/crates.io/pull/14213.

### DIFF
--- a/src/bin/crates-io/admin/verify_token.rs
+++ b/src/bin/crates-io/admin/verify_token.rs
@@ -6,9 +6,9 @@ use crates_io::{db, models::User};
 #[derive(clap::Parser, Debug)]
 #[command(
     name = "verify-token",
-    about = "Look up a username by API token.",
-    long_about = "Look up a username by API token. Used by staff to verify someone's identity \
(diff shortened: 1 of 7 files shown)


Turbo87 (approved):
👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14576/62456ba14115e5c8905f7c3ccfba920e783af146..0de7f53a2efe2a309626c623bb694ffd716a002c)*

PR #14577, opened 2026-09-03
In tests, where appropriate, use `users.username`

carols10cents:
Instead of `users.gh_login`.

For now, these fields should always be the same, so there is no change in the test or implementation behavior coverage.

When we fully enable crates.io usernames as being independent of GitHub usernames, these tests should be using the crates.io username.

Extracted from https://github.com/rust-lang/crates.io/pull/14575. Might conflict with https://github.com/rust-lang/crates.io/pull/14213 a tiny bit.

### DIFF
--- a/src/tests/owners.rs
+++ b/src/tests/owners.rs
@@ -175,7 +175,7 @@ async fn create_and_add_owner(
 async fn owners_can_remove_self() {
     let (app, _, user, token) = TestApp::init().with_token().await;
     let mut conn = app.db_conn().await;
-    let username = &user.as_model().gh_login;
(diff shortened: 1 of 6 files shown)


Turbo87 (approved):
technically the add/remove owner endpoints still expect their inputs to be GitHub account names, but since the setup in these tests ensures that they are equivalent, I think this is fine :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14577/2e1b870971b57caddc4c10d6be52189e04ba06d9..285708dbad99e6d27c388d3d60cade233a5cbb30)*

PR #14579, opened 2026-09-03
middleware: Restrict legacy `/svelte` redirects

Turbo87:
The frontend move is permanent now, so this PR changes the legacy `/svelte` redirects to use HTTP 308 to signal that search engines should prefer the unprefixed URLs.

Non-root redirect targets must now also start with an ASCII letter after the leading slash, preventing destinations that clients could interpret as another host. I've also excluded targets starting with `/api`.

### Related

- #13542


fereidani:
Thanks for the quick fix! I've published a public report of the issue here: https://fereidani.com/cratesio-svelte-middleware-backslash-open-redirect

PR #14581, opened 2026-09-03
In owner invite messages, use crates.io username

carols10cents:
These are messages that a user sees when they've invited other users to own a crate. Right now, there shouldn't be any confusion because `users.username` should always equal `users.gh_login`, but I've added "Crates.io user" to the message to hopefully clarify for the future when we have crates.io usernames independent of GitHub usernames.

Extracted from https://github.com/rust-lang/crates.io/pull/14575; will probably conflict with https://github.com/rust-lang/crates.io/pull/14213 a little bit.

### DIFF
--- a/packages/crates-io-api-client/schema.ts
+++ b/packages/crates-io-api-client/schema.ts
@@ -3312,7 +3312,7 @@ export interface operations {
                     "application/json": {
                         /**
                          * @description A message describing the result of the operation.
-                         * @example user ghost has been invited to be an owner of crate serde
+                         * @example Crates.io user ghost has been invited to be an owner of crate serde
                          */
                         msg: string;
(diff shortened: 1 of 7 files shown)


moskirathe (approved):
i'll update https://github.com/rust-lang/crates.io/pull/14213 to resolve the minor conflict this introduces. Should be easy enough to propagate the username type (`crates.io` or `github` currently) so that its included in the invite messages.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14581/4878f150daca72da01b24ca075dcb88b392df375..db57efd970195b72693d3bf6ebb142763859284f)*

Turbo87 (approved):

PR #14589, opened 2026-09-04
controllers/owners: Compare user owners by ID

Turbo87:
User invitations now resolve the requested account before checking existing ownership and compare crates.io user IDs instead of login strings. This prevents a reused GitHub login from making a different account appear to already own the crate while preserving the existing team checks and lookup error precedence.

The test now also cover highest-ID account selection and removal of all owners matching a reused GitHub login. This work was partially extracted from #14213 so the identity fix can be reviewed independently from the username-prefix and disambiguation changes.

### Related

- #14213
- #13769

### DIFF
--- a/src/controllers/krate/owners.rs
+++ b/src/controllers/krate/owners.rs
@@ -357,7 +357,8 @@ async fn add_owner(
 ) -> Result<NewOwnerInvite, OwnerAddError> {
     match Login::parse(login)? {
         Login::GitHubTeam(team) => {
-            let login_test = |owner: &Owner| owner.login().to_lowercase() == *login.to_lowercase();
+            let lowercase_login = login.to_lowercase();
+            let login_test = |owner: &Owner| owner.login().to_lowercase() == lowercase_login;
             if owners.iter().any(login_test) {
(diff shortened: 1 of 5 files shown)


moskirathe (approved):

src/controllers/krate/owners.rs
```diff
@@ -358,15 +352,34 @@ async fn add_owner(
     conn: &mut AsyncPgConnection,
     req_user: &User,
     krate: &Crate,
+    owners: &[Owner],
     login: &str,
 ) -> Result<NewOwnerInvite, OwnerAddError> {
     match Login::parse(login)? {
         Login::GitHubTeam(team) => {
+            let login_test = |owner: &Owner| owner.login().to_lowercase() == *login.to_lowercase();
```
eth3lbert: I think we could extract `login.to_lowercase()` outside the closure.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14589/66d2657105d3e976994cd5f2c4620e40b4d0ec56..02d85b60b22a1ba99d241717c302603689447f47)*

eth3lbert (approved):

PR #14596, opened 2026-09-05
controllers/owners: split owner remove query to remove by username and remove by team name

moskirathe:
This PR is a partial extraction from https://github.com/rust-lang/crates.io/pull/14213. 

It splits the `owner_remove` query to `owner_remove_by_username` and `owner_remove_by_team_name`. This will allow us to change the behaviour of `ower_remove_by_username` so that we can compare usernames names using the canon_username function (which also normalizes hyphen to underscore), but keep comparing team names using `lower` - github allows separate team names that only differ by underscore and hyphens to exist.

**Tests:**
Owner remove is now routed through `Login:parse`, which rejects malformed logins before any DB lookup happens. `Login::parse` currently treats *any* login containing `:` as a team login and hands it to `GitHubTeamLogin::parse`, which rejects malformed ones with a `bad_request`.

Previously `remove_owners` passed the raw string straight to `krate.owner_remove`, which did the lookup, matched nothing (stored logins never contain `:`), and returned `OwnerRemoveError::not_found(login)`.

So the tests, which assert the old fall-through behavior, now get a parse error instead.

### DIFF
--- a/crates/crates_io_database/src/models/krate.rs
+++ b/crates/crates_io_database/src/models/krate.rs
@@ -213,7 +213,53 @@ impl Crate {
         Ok(users.chain(teams).collect())
     }
 
-    pub async fn owner_remove(
+    pub async fn owner_remove_with_username(
+        &self,
+        mut conn: &AsyncPgConnection,
+        login: &str,
+    ) -> Result<(), OwnerRemoveError> {
+        let query = diesel::sql_query(
+            r#"WITH crate_owners_with_login AS (
+                SELECT
+                    crate_owners.*,
+                    CASE WHEN crate_owners.owner_kind = 1 THEN
+                         teams.login
+                    ELSE
+                         users.gh_login
+                    END AS login
+                FROM crate_owners
+                LEFT JOIN teams
+                    ON crate_owners.owner_id = teams.id
+                    AND crate_owners.owner_kind = 1
+                LEFT JOIN users
+                    ON crate_owners.owner_id = users.id
+                    AND crate_owners.owner_kind = 0
+                WHERE crate_owners.crate_id = $1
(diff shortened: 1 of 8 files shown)


Turbo87 (commented):
sorry, but I think this is the wrong direction 😅 

I think what we want is to:
1. load the owners of the crate
2. resolve the supplied names into `(owner_type, owner_id)` tuples and reject ambiguities
3. remove all selected pairs with one `UPDATE`, scoped by `crate_id`

this should also allow us to get rid of the raw SQL queries. I have a local branch here already that implements it that way.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14596/66d2657105d3e976994cd5f2c4620e40b4d0ec56..1106345aa77155b549f231cc41137131d7516f87)*

moskirathe:
> sorry, but I think this is the wrong direction 😅
> 
> I think what we want is to:
> 
> 1. load the owners of the crate
> 2. resolve the supplied names into `(owner_type, owner_id)` tuples and reject ambiguities
> 3. remove all selected pairs with one `UPDATE`, scoped by `crate_id`
> 
> this should also allow us to get rid of the raw SQL queries. I have a local branch here already that implements it that way.
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14596/66d2657105d3e976994cd5f2c4620e40b4d0ec56..1106345aa77155b549f231cc41137131d7516f87)_

@Turbo87 ah ok! your approach makes sense. i'll close this pr and hold off on making any changes to https://github.com/rust-lang/crates.io/pull/14213 until your refactors are in. lmk if there is anything i can help with in the meantime.

PR #14628, opened 2026-09-09
Support explicit prefixes in owner invitations

Turbo87:
Owner invitations now accept `crates.io:username` and `github:username` to select an account explicitly by its crates.io username or linked GitHub login. GitHub lookups use a new case-insensitive index and keep the existing highest-ID-wins behavior.

This extracts the prefix support for invitations from #14213. Unprefixed invitations and owner removals keep their current behavior for now to limit the review scope of this PR. Best reviewed commit-by-commit :)

### Related

- #14213
- #13769


crates/crates_io_database/src/models/oauth_github.rs
```diff
@@ -32,6 +33,20 @@ pub struct OauthGithub {
     pub user_id: i32,
 }
 
+impl OauthGithub {
+    /// Finds a linked GitHub account by its case-insensitive login.
+    /// If several accounts match, returns the one with the highest account ID.
+    /// Accounts with the invalid legacy ID `-1` are excluded.
```
carols10cents: There really shouldn't be any rows in the `oauth_github` table with a `-1` ID -- I just checked and there aren't any results for `SELECT * FROM oauth_github WHERE account_id = -1` in prod or staging. `account_id` is also the primary key of this table, so it would only be possible to have one record with `-1` anyway.

This value comes from the `GitHubUser.id` without modification: https://github.com/rust-lang/crates.io/blob/2df080c06de8a9e8e559ecf4989794e66e70086b/src/controllers/session.rs#L505

It'd be nice to not have this hack in the codebase at all once we've completed the transition off the `users` fields... is there a particular concern you're addressing by having this caveat and filter?

crates/crates_io_database/tests/oauth_github.rs
```diff
@@ -0,0 +1,72 @@
...
+    let conn = test_db.async_connect().await;
+    let user_id = insert_user(&conn, "newer-user", 2, "shared").await;
+    insert_user(&conn, "older-user", 1, "SHARED").await;
+
+    let account = OauthGithub::find_by_login(&conn, "shared").await.unwrap();
+    assert_eq!(account.account_id, 2);
+    assert_eq!(account.user_id, user_id);
+}
+
+/// Accounts with the invalid legacy ID cannot be resolved by login.
```
carols10cents: As stated above, I don't think this test is valid because I don't think this is possible in the actual app.

src/controllers/krate/owners.rs
```diff
@@ -366,36 +373,82 @@ async fn add_owner(
...
+                bad_request(message)
+            })?;
+            Some(PublicUser::find(conn, account.user_id).await?)
         }
+        Login::Unprefixed(username) => {
+            PublicUser::find_by_login(conn, username).await.optional()?
+        }
+    };
+    let user =
+        user.ok_or_else(|| bad_request(format_args!("could not find user with login `{login}`")))?;
```
carols10cents: It feels a little redundant to have this `bad_request` handling out here in _addition_ to the ones within the `Login::CratesIo` arm and the `Login::GitHub` arm? I like that the arms have more specific error messages including the type of username that was used for lookup.

What do you think about moving this into the `Login::Unprefixed` arm and then just returning a `PublicUser` rather than `Option<PublicUser>` from the `match`?

carols10cents (changes requested):
I have a few things for addressing, either with changes or explanation 😁 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14628/8dde01fa95a05d89d1ac6f981e12444e71a155ab..59116cfdf3576a5ea614ae64f76684b4d2a17ec7)*

crates/crates_io_database/src/models/oauth_github.rs
Turbo87: good catch, I confused this with the `users.gh_id` behavior. I'll fix it 👍

src/controllers/krate/owners.rs
```diff
@@ -366,37 +372,60 @@ async fn add_owner(
...
             let github = &*app.github;
             let encryption = &app.config.token_encryption;
-            add_github_team_owner(github, conn, req_user, krate, team, encryption).await
+            return add_github_team_owner(github, conn, req_user, krate, team, encryption).await;
         }
-        Login::Unprefixed(login) => {
-            let user = PublicUser::find_by_login(conn, login).await.optional()?;
+        Login::CratesIo(username) => {
+            let user = users_by_username(username)
+                .left_join(oauth_github::table)
```
moskirathe, line 380: curious, why left_join here? because the username here is already disambiguated, shouldn't the lookup only happen on the `users` table? 

or is the plan to remove the left join in a separate pr once we add the logic to check if a username is ambiguous?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14628/8dde01fa95a05d89d1ac6f981e12444e71a155ab..59116cfdf3576a5ea614ae64f76684b4d2a17ec7)*

Turbo87: the `PublicUser::github_account_matches` field (or whatever it was called 😅) requires the left join to be present.

src/controllers/krate/owners.rs
```diff
@@ -366,37 +372,60 @@ async fn add_owner(
...
+    /// crates.io username, such as `crates.io:octocat`.
+    CratesIo(&'a str),
     /// User login without a service prefix.
     Unprefixed(&'a str),
 }
 
 impl<'a> Login<'a> {
     /// Parses an owner login.
     fn parse(login: &'a str) -> Result<Self, BoxedAppError> {
+        if let Some(username) = login.strip_prefix("crates.io:") {
```
moskirathe: i think this wont match for mixed / uppercase logins (e.g `Crates.io:username`). should we lowercase `login` before stripping the prefix?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14628/8dde01fa95a05d89d1ac6f981e12444e71a155ab..59116cfdf3576a5ea614ae64f76684b4d2a17ec7)*

src/controllers/krate/owners.rs
```diff
@@ -644,6 +674,13 @@ mod tests {
         assert_eq!(username, "octocat");
     }
 
+    /// A crates.io prefix must be followed by exactly one nonempty username.
+    #[test]
+    fn rejects_malformed_crates_io_login() {
+        assert_err!(Login::parse("crates.io:"));
+        assert_err!(Login::parse("crates.io:octocat:extra"));
+    }
```
moskirathe: could also add some mixed case / uppercase tests here for login parse (unless you plan to handle that elsewhere)

src/controllers/krate/owners.rs
Turbo87: IMHO there is no reason for us to support case-insensitivity for the prefix.

moskirathe: why not? i'm thinking if someone tries to add `crates.io:Some_Username` (which we'll support) they are just as likely to type `Crates.io:Some_Username` 😄

Turbo87: if we don't support it then there is no reason for them to spell it that way 🤷‍♂️ 

for account names things are a bit different because of impersonation concerns, but for these prefixes I don't see why we should allow other spellings.

moskirathe: ok got it 👍

carols10cents: we can always add support for any case of `crates.io` later if people complain, but I don't recall anyone asking for support for different cases of `github` in the teams owner `github:org:teams` syntax.

src/controllers/krate/owners.rs
Turbo87: turns out that was incorrect. `github_username_matches` uses a subquery, so it wouldn't need the left join, but the `gh_avatar` field reads from the `oauth_github` table, which then still makes this require the left join :)

Turbo87:
rebased on top of https://github.com/rust-lang/crates.io/pull/14633 now and slightly adjusted to address the feedback. let me know if this is better now :)

moskirathe (approved):

carols10cents (approved):
LGTM now! Thank you!!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14628/5c8e472dd2ce0603c57110ab21abddec50ba986c..413373d4ff7572fea5e04ccf5c3f6acc324f5b19)*

PR #14636, opened 2026-09-10
database_dump: Include all users in database dumps

Turbo87:
The database dump is currently inconsistent: `crate_owners`, `deleted_crates`, and `oauth_github` contain foreign keys referencing users that are missing from the dump. The user filter only includes current crate owners and publishers of existing versions, so users referenced elsewhere can be left out.

The filter was introduced in https://github.com/rust-lang/crates.io/issues/630 to keep accounts private until they published a crate or became an owner. However, the public API exposes user profiles without requiring either action, and the dump already includes GitHub account records for users excluded by the filter. Filtering only the `users` table no longer provides a consistent privacy boundary. Including all users resolves these missing references and avoids maintaining a separate list of relationships that qualify a user for export. Email addresses, authentication tokens, and other private fields remain excluded.

The alternative would be to adjust the filter, but then we'd also need to introduce a similar but probably even more complex filter to the `oauth_github` table and keep these updated when we add new `user_id` columns in other tables. IMHO that does not seem worth it, given that our API makes all accounts public anyway.

### DIFF
--- a/crates/crates_io_database_dump/src/dump-db.toml
+++ b/crates/crates_io_database_dump/src/dump-db.toml
@@ -255,13 +255,6 @@ jti = "private"
 used_at = "private"
 expires_at = "private"
 
-[users]
-filter = """
-id in (
-    SELECT owner_id AS user_id FROM crate_owners WHERE NOT deleted AND owner_kind = 0
-    UNION
-    SELECT published_by as user_id FROM versions
-)"""
 [users.columns]
 id = "public"
 gh_login = "public"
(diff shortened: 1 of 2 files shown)


moskirathe (approved):

carols10cents (approved):
LGTM! 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14636/990b75280f1876796211688c9ed52bb7796e005a..525b3d94b03aa1af8b47ab60f65700b47d76ea84)*

PR #14638, opened 2026-09-10
database_dump: Add schema scripts for staged imports

Turbo87:
The dump now includes `schema-before.sql` and `schema-after.sql`, so consumers can load data before creating indexes and validating foreign keys. The two triggers that populate fields omitted from the CSV files remain in the before script. All three schema scripts come from the same schema-only `pg_dump` output, and the complete `schema.sql` remains available for existing consumers.

A prototype restore took about 6m22s with indexes deferred for all tables, including index creation and foreign-key validation, compared with 8m27s for an already partially optimized restore that deferred download indexes. Download records account for most rows in the dump, so that baseline already avoided much of the index maintenance during loading. A regular import maintains those indexes too and takes even longer.


carols10cents (approved):
Sounds good!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14638/c8a3cdacff4ffe17c806e216d1edfdb688e0926c..aa9285b1c79164460e91f5fc99ee76f7fb727069)*

PR #14685, opened 2026-09-15
Reject missing Cargo target source files

Turbo87:
This PR adjusts tarball processing to derive metadata for each build script, library, and binary from the `Cargo.toml` manifest. I've adjusted the publishing endpoint to to reject any target declaration whose path is missing from the tarball.

The `TargetMetadata` struct might look a bit excessive for this specific use case, but in the follow-up PRs I'm planning to save it to the database and expose it to the frontend, to enable more features.


jlizen:
I'm concerned that this fails closed if targets are not statically included in the tarball, but rather generated with build.rs. Example: protobuf schema is used to codegen a types crate.

I assume it's a *pretty* rare pattern, usually you codegen modules inside of statically defined crate, but I feel cautious about enforcing this given all the cursed things people do with build.rs. If we want to get stricter with build.rs, sounds great, but it should be a broader RFC or similar probably?

Anyway seems like we could fail open if we can't parse metadata instead and sidestep possible issues?

Turbo87:
I had Codex validate whether this scenario is actually possible with Cargo. This is the response:

----

No. The failure mode described in the comment is not supported by Cargo.

I verified this with Cargo 1.98.1:

- A `[lib]` or `[[bin]]` whose `path` does not exist fails target resolution before `build.rs` runs.
- `cargo package` warns that such a target is not included in the package.
  - If it is the only target, packaging fails with “no targets specified.”
  - If another valid target exists, Cargo removes the missing target from the normalized `Cargo.toml`.
- Cargo’s package verification also rejects build scripts that modify the packaged source tree outside `OUT_DIR`.

Cargo explicitly defines a target’s `path` as its source file, and its documentation says build scripts must write generated artifacts into `OUT_DIR`: [Cargo targets](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#configuring-a-target), [build scripts](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script).

The supported protobuf/codegen pattern is:

```rust
// src/lib.rs, included in the tarball
include!(concat!(env!("OUT_DIR"), "/generated.rs"));
```

`build.rs` generates `generated.rs` inside `OUT_DIR`. I confirmed that this pattern builds and passes `cargo package` verification.

Therefore, this PR would not reject a valid code-generated crate:

- The static target entry point, such as `src/lib.rs`, must already be present.
- Generated modules behind that entry point do not need to be in the tarball.
- A target entry point generated by `build.rs` cannot work as proposed.

So failing closed on a missing declared target source is consistent with Cargo’s existing behavior and should not require a broader build-script RFC. The comment appears to conflate generated modules with generated Cargo targets.

PR #14699, opened 2026-09-16
Remove legacy metrics after OpenTelemetry migration

Turbo87:
This PR implements the cleanup phase of the migration from Prometheus and custom Datadog metric submissions to OpenTelemetry. It removes direct Datadog metric submission, instance metrics log forwarding, and the private Prometheus endpoint, along with the legacy instrumentation and configuration. Datadog monitor service checks remain unchanged for now.

The original draft covered both expansion and cleanup. The expansion phase was extracted and merged in #14700, introducing OpenTelemetry while retaining the legacy paths for rollout comparison. Only the cleanup commits remain here.

### Related

- #14700

PR #14700, opened 2026-09-16
Add `opentelemetry` metrics support

Turbo87:
This PR adds `opentelemetry` metrics support to the application.

It can be enabled by setting the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable and all of the other necessary environment variables read by the `opentelemetry` crate family.

The implementation is intentionally not coupled to Datadog and can easily be configure to point to a different OTLP server. Note that once `OTEL_EXPORTER_OTLP_ENDPOINT` is enabled, the custom Datadog metrics exporter is disabled to avoid conflicting reports.

### Related

- Full implementation with the cleanup phase: https://github.com/rust-lang/crates.io/pull/14699


Turbo87:
I'm gonna try this out on staging. The code is essentially disabled unless the relevant env vars are set, so even if this gets deployed to production it shouldn't have an immediate impact.

Turbo87:
staging test was successful. everything appears to be working as intended.

PR #14696, opened 2026-09-16
Support explicit prefixes in owner removals

Turbo87:
Owner removals now accept `crates.io:username` and `github:username` to select current owners by their crates.io username or linked GitHub login. Unprefixed names keep their existing GitHub login matching behavior for now.

The crate settings page now uses the `crates.io:` prefix when removing individual owners, while GitHub team identifiers remain unchanged.

This was partially extracted from the original implementation in #14213.

### Related

- Original implementation: #14213
- Tracking issue: #13769
- Invitation prefix support: #14628
- Removal resolver preparation: #14681

PR #14681, opened 2026-09-14
controllers/owners: Resolve removal owner names in Rust

Turbo87:
The removal handler now resolves owner names against the initial owner list in Rust before passing user and team identities to a single database update. This separates name resolution from row removal, simplifies the SQL query and prepares the handler to support distinct `crates.io:` and `github:` namespaces in the next PR.

### Related

- #14213
- #13769
- #14628

PR #14633, opened 2026-09-09
controllers/owners: Improve `OwnerLogin` parser

Turbo87:
Owner login parsing now rejects empty components and more than three colon-separated components, preventing trailing input from being silently ignored. Prefixed usernames receive a dedicated unsupported error instead of being parsed as GitHub teams.

### Related

- https://github.com/rust-lang/crates.io/issues/13769
- #14628

PR #14625, opened 2026-09-09
crate.settings: Recognize user prefixes in owner invitations

Turbo87:
The owner form now only treats names containing two colons as teams. User identifiers such as `crates.io:alice` and `github:alice` contain only one colon and receive the invitation notification after a successful request. Previously, any colon caused the form to announce that a team was added, which was incompatible with the introduction of these new identifiers.

### Related

- https://github.com/rust-lang/crates.io/issues/13769

PR #14703, opened 2026-09-17
controllers/owners: Disambiguate unprefixed owner names

### LINKED ISSUE #13769: Implement username disambiguation in `cargo owner` param handling and error messages

Turbo87:
Unprefixed owner invitations now resolve the supplied name as a crates.io username and confirm that the corresponding GitHub login belongs to the same account. Ambiguous or mismatched names return actionable `crates.io:` and `github:` alternatives instead of silently selecting an account.

Unprefixed removals resolve both namespaces among the crate's current owners. The request proceeds when both names identify the same owner and otherwise requires an explicit prefix, including for reused GitHub logins.

### Related

- Resolves #13769
- RFC: rust-lang/rfcs#3946
- Original implementation: Closes #14213
- Invitation groundwork: #14619, #14625, #14628, #14633
- Removal groundwork: #14681, #14696


src/controllers/krate/owners.rs
```diff
@@ -308,26 +308,52 @@ pub async fn remove_owners(
...
+                        .select(oauth_github::user_id)
+                        .distinct()
+                        .load::<i32>(&mut conn)
+                        .await?
+                };
+
+                if !crates_io.is_empty() && !github.is_empty() {
+                    // Compare every matching owner, including reused GitHub logins.
+                    let same_owners = crates_io.len() == github.len()
+                        && crates_io.iter().all(|id| github.contains(id));
```
eth3lbert:
```suggestion
                        && crates_io.iter().zip(github.iter()).all(|(&a, &b)| a == b);
```

src/controllers/krate/owners.rs
```diff
@@ -308,26 +308,52 @@ pub async fn remove_owners(
...
+                } else {
+                    oauth_github::table
+                        .filter(oauth_github::user_id.eq_any(&user_ids))
+                        .filter(lower(oauth_github::login).eq(lower(username)))
+                        .filter(oauth_github::account_id.ne(-1))
+                        .select(oauth_github::user_id)
+                        .distinct()
+                        .load::<i32>(&mut conn)
+                        .await?
+                };
```
eth3lbert: This could likely be refactored to use pipelined queries, but there's no need to implement it in this PR. 😄

src/controllers/krate/owners.rs
eth3lbert: I think it would be better to sort both and compare them one by one. Even better, we could convert one into a `HashSet` and perform lookups.

eth3lbert (approved):
LGTM! Just a few nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14703/00877699ec218d83dd05ba787f29c403e9c4da6d..9c0c4778b24e8d6329bb3bda719b64507148492a)*

moskirathe (approved):
thanks! 🚀 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14703/00877699ec218d83dd05ba787f29c403e9c4da6d..9c0c4778b24e8d6329bb3bda719b64507148492a)*

src/controllers/krate/owners.rs
Turbo87: > I think it would be better to sort both and compare them one by one.

that's a great idea. it's much more readable this way. 👍

src/controllers/krate/owners.rs
Turbo87: pipelining implementation. wasn't actually that much more code :)

PR #14704, opened 2026-09-17
Persist `versions.target_metadata` in the database

Turbo87:
New versions now store build-script, library, and binary metadata in the nullable `versions.target_metadata` JSONB column. Existing versions remain distinguishable as unanalyzed via SQL `NULL`.

This is supposed to replace the existing `has_lib`, `lib_name` and `bin_names` columns, which remain populated for now until we have backfilled `target_metadata`. The additional metadata allows us to eventually display whether crates contain a build script, or if they implement proc-macros.

### Related

- #14685


Turbo87:
Since this is an experimental column and not publicly exposed yet I'll merge this to try it out on staging and production. If we notice any issues we can still revert without public API breakage. :)

PR #14741, opened 2026-09-21
Align `github_username_matches` with owner invitations

Turbo87:
The flag now uses the same case-insensitive GitHub lookup as owner invitations, preferring the highest GitHub account ID. This avoids reporting mismatches for case-only differences and reporting matches when a reused GitHub login resolves to another crates.io account.

### Related

- #13769
- #14703


Turbo87:
@rust-lang/crates-io this has implications to how we display mismatches on the frontend, so I'm not entirely sure about this yet. I guess we also should deal with the fact that we could return a linked GitHub account with an outdated username, where we have newer `oauth_github` row already with the same `username`. I'm open to suggestions here. This whole GitHub syncing thing seems like a mess. I'm glad we're cleaning this up a bit now :)

PR #14744, opened 2026-09-21
Prioritize GitHub syncs for conflicting logins

Turbo87:
GitHub sync batches now prioritize accounts whose case-insensitive username matches another stored account with a higher GitHub ID, so outdated usernames get refreshed sooner. Both priority groups retain oldest-first sync ordering within the existing batch limit.

### Related

- https://github.com/rust-lang/crates.io/pull/13143


Turbo87:
btw I did measure the performance impact of this change. relatively speaking it is not insignificant, but small enough in absolute terms to not matter too much in practice. this query does not run in an API endpoint anyway and is only triggered once per hour at the moment.

De-ASI-INTERFACE (changes requested):
Ensure no identical credentials at all. 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14744/e5e8c07914cd3505f54ff950929ed66fe7112b27..04b321f5749980c7d844b4bc1862762283a283a3)*

De-ASI-INTERFACE (changes requested):
Prioritize eradicating them 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14744/e5e8c07914cd3505f54ff950929ed66fe7112b27..04b321f5749980c7d844b4bc1862762283a283a3)*

De-ASI-INTERFACE (approved):

crates/crates_io_database/src/models/oauth_github.rs
```diff
@@ -34,6 +35,24 @@ pub struct OauthGithub {
...
+            .field(oauth_github::account_id)
+            .gt(oauth_github::account_id);
+        let conflicting_login = exists(newer_account.filter(login_matches).filter(higher_id));
+
+        oauth_github::table
+            .order((conflicting_login.desc(), oauth_github::last_sync.asc()))
+            .limit(limit)
+            .load(&mut conn)
+            .await
+    }
```
eth3lbert: 🙋‍♂️ One quick question: can `conflicting_login` occur with enterprise managed users? If that's the case, since they won't be modified (only `last_sync` will change), won't they just keep getting selected in future runs due to the priority?

eth3lbert (approved):
Makes sense to me 👍 I just have a quick question related to enterprise managed users.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14744/e5e8c07914cd3505f54ff950929ed66fe7112b27..04b321f5749980c7d844b4bc1862762283a283a3)*

crates/crates_io_database/src/models/oauth_github.rs
Turbo87: ugh... I hate these enterprise accounts...

but good catch! I've excluded the enterprise accounts from the prioritization for now.

carols10cents (approved):
LGTM!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/crates.io/14744/dff5ec433b2e8ef0c70990b5d5f515396be7b736..38a9fc4809e1ff8883b6f0ea592d7c077b0fffb5)*

